Issue #8, opened 2018-12-13
[BUG][NV] CMakeLists.txt structure inconsistent with cuDF, cuML


mt-jones:
* The test infrastructure in `cugraph` places a `CMakeLists.txt` in every test directory.
* Gunrock is added as an external project, and it cloned to a location along-side `cugraph` rather than in a directory inside of `cugraph/build`, meaning deleting `build` does not restore the the compiled project to its initial state
* Locations for libraries like `libcudf.so` are hard-coded, rather than searched in system locations and Conda locations, etc.

rgsl888:
I have posted a PR for this.

PR #11, opened 2018-12-13
[REVIEW] Resolves issues of Cmake Files mentioned in rapidsai/cugraph#8

rgsl888:
Resolves [ rapidsai/cugraph#8](https://github.com/rapidsai/cugraph/issues/8),

i.e
> The test infrastructure in cugraph places a CMakeLists.txt in every test directory. 

These are removed

> Gunrock is added as an external project, and it cloned to a location along-side cugraph rather than in a directory inside of cugraph/build, meaning deleting build does not restore the the compiled project to its initial state

Gets built in build

> Locations for libraries like libcudf.so are hard-coded, rather than searched in system locations and Conda locations, etc.

Finds path for library and include path


GPUtester:

Issue #17, opened 2018-12-19
[FEA][NV] Return identifiers along with results at Python API level


afender:
Currently, we return a cuDF series that implicitly matches the adjacency list layout.
We should return a cuDF data frame containing enough cuDF series to explicitly identify whose this result belongs to. 

For instance : 
- (vertex, pagerank) for PageRank
- (source, destination, jaccard_similarity) for Jaccard
- (vertex, cluster_assignment) for Louvain
- etc.

afender:
The feature was added along with:
*Merge pull request #41 from Iroy30/jaccard_grmat_graph_bindings*

@Iroy30 : Can we test this at the python level in Pagerank and Jaccard tests? NetworkX does return these identifiers, so the comparison should be pretty straightforward.

Iroy30:
Currently we are returning dataframes instead of cudf series. The python tests verify the 'pagerank' column and the 'jaccard_coeff' column of the corresponding dataframes  with networkx results. Issue can be closed.

afender:
Do you test that identifiers are properly returned?

Iroy30:
Do you want the vertex ids from 0 to n-1 for pagerank and the source. destination for jaccard to be compared as well? These are not verified in the tests currently

afender:
Exactly. Please upgrade the test to check that : 
- vertex identifiers are properly returned in Pagerank, BFS, SSSP, louvain and spectral clustering.
- edge identifiers are properly returned in Jaccard

PR #20, opened 2018-12-28
[REVIEW] Retrieve vertex/edge indices from adjacency list 

afender:
This addresses the C++/CUDA part for #17. 
I also added a feature allowing conversion from adjacency list to edge list.


GPUtester:

PR #21, opened 2018-12-28
[REVIEW] Fixed addAdjList and addTransposed

afender:
These functions were using the size of the graph as the size of the offset array. 
Improved the tests for gdf_add_adj_list and gdf_add_transpose to catch this.


GPUtester:

Issue #22, opened 2019-01-03
[NV][BUG] Fix utilities/error_utils.h dependency


afender:

afender:
Closing. Fixed by :
*Fixed error_utils.h dependency #23*

PR #23, opened 2019-01-03
Fixed error_utils.h dependency

afender:
Issue #22

Fixed `Cmakelist.txt` to find `error_utils.h`.
Removed `error_utils.h` from `cugraph.h` so that `error_utils.h` does not need to be installed


afender:
Tested by @Iroy30

PR #32, opened 2019-01-04
Refactor fixes

afender:
3 fixes :

1. Reduced the size of grmat tests because of timeout issue on Jenkins. We still test it on graphs with more than 2M edges.
2. Fixed intermittent segfault in cudart UnregisterBinaryUtil. This issue was impacting cuda9.2. 
3. Re-enabled nvgraph plugin and plugin tests (jaccard, louvain, sssp)


afender:
This fixes build and tests till the Python step.

Issue #27, opened 2019-01-03
[FEA][NV] I/O : support edge lists using any numerical identifier as vertex label


afender:
Currently, sources and destinations must be 32-bit integers in the range [0, V-1].
We need to support 64-bit integers without range.

afender:
The CUDA part has been addressed by Chuck in PR#38. 
Next step is integration in python bindings and CAPI.

afender:
The next PR should also add changes for RMM compatibility: 
1. #include <rmm_utils.h>
2. replace `cudaMalloc/free` by `ALLOC_MANAGED_TRY / ALLOC_FREE_TRY`

ChuckHastings:
PR #234 adds a python interface for integer lookups.  I believe this completes this issue.

afender:
We need to add documentation in the python API.
I understand you have a notebook already, correct?. If so, it would be great to add it to the notebook repo.

ChuckHastings:
PR #247 adds python interface.

https://github.com/rapidsai/notebooks/pull/127 adds the notebook to the notebook repo.

PR #38, opened 2019-01-14
[REVIEW] Renumber vertices

ChuckHastings:
Created function to renumber source and dest vertices for a graph.  Addresses #27 

Takes an array of src vertices and dst vertices (assumed to be integer types) and generates a densely packed version of the vertex arrays and a map translating the new vertex ids back to the original vertex ids.


GPUtester:

kkraus14:
add to whitelist

afender:
1. Would it be a lot of work to have `template <typename T_in, typename T_out>` in _renumber.cuh:107_? Since all our algorithms currently only support 32bits input that would allow loading/converting 64bit input. 
2. It seems to me that 32 `threadPerBlock` and `threadBlock` lead to underoccupancy issues. Can you double check these numbers? Unless I’m missing something, 32 threadPerBlock covers only 1 warp, which means warps are not overlapped within an SM and 32 threadBlock covers only the first 32 SM.

ChuckHastings:
> * Would it be a lot of work to have `template <typename T_in, typename T_out>` in _renumber.cuh:107_? Since all our algorithms currently only support 32bits input that would allow loading/converting 64bit input.

  Not a lot of work.  I actually pondered that but didn't want to make it too complex to start.  But that's an easy use case to justify.  I think it's just a simple editing job to work through which things would be T_in and which would be T_out.

> * It seems to me that 32 `threadPerBlock` and `threadBlock` lead to underoccupancy issues. Can you double check these numbers? Unless I’m missing something, 32 threadPerBlock covers only 1 warp, which means warps are not overlapped within an SM and 32 threadBlock covers only the first 32 SM.

  I should have added a comment about those being for testing purposes.  I was thinking that those parameters should not be constants.  I wasn't sure if we wanted to expose them as parameters or have some heuristic computation based on the number of edges in the graph.

afender:
In the renumbered output, should we test that `min(src U dest) = 0` and `max(src U dest) = V-1` ?

PR #39, opened 2019-01-15
[Review]RMM placeholders in cuGraph

aschaffer:
This is preparation integration work for when RMM library in RAPIDS will be stable. It provides hooks for integration with RMM, so that only minimal changes will be necessary when RMM lib will be ready (ideally only updating a header + CMakeLists.txt).

### DIFF
--- a/include/rmm_utils.h
+++ b/include/rmm_utils.h
@@ -1,0 +1,75 @@
+#pragma once
+
+#define DEBUG_NO_RMM
+
+#include <sstream>
+#include <stdexcept>
+
+#define RMM_TRY_THROW( call )  if ((call)!=RMM_SUCCESS) \
+    {                                                   \
+      std::stringstream ss;                             \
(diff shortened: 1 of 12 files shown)


include/rmm_utils.h
```diff
@@ -0,0 +1,75 @@
...
+
+#define RMM_TRY_THROW( call )  if ((call)!=RMM_SUCCESS) \
+    {                                                   \
+      std::stringstream ss;                             \
+      ss << "ERROR: RMM runtime call  " << #call        \
+         << cudaGetErrorString(cudaGetLastError());     \
+      throw std::runtime_error(ss.str());               \
+    }
+
+#ifdef DEBUG_NO_RMM
```
harrism: Most of this is redundant with functionality already in RMM. RMM is ready to use. Please use it and file RMM issues for problems you run into.

afender:
Closing as this work has been addressed in PR #45 and #49.

Issue #24, opened 2019-01-03
[FEA][NV] Add Python bindings for SSSP


afender:
I already implemented the CUDA/C++ part to access nvgraph features using cugraph graph structure and SSSP wrapper in [src/plugin/nvgraph/nvgraph_gdf.cu](https://github.com/rapidsai/cugraph/blob/master/src/plugin/nvgraph/nvgraph_gdf.cu). The task consist in adding python bindings for `gdf_error gdf_sssp_nvgraph`

Subtasks checklist:

- [ ]  Python bindings and API (may include some adjustments of the C++/CUDA part)
- [ ]  Pytest testing against NetworkX
- [ ]  Documentation
     - [ ]  Doxygen (C++)
     - [ ]  Sphinx (Python)
     - [ ]  python code example
- [ ]  Check headers and licenses
- [ ]  Validate everything in Jenkins

archenroot:
as for other graph algos I will do Java binding

afender:
Closing after merging : Fea python bindings sssp #44

PR #44, opened 2019-01-30
[REVIEW] Fea python bindings sssp

ChuckHastings:
Create python bindings to call the nvgraph implementation of SSSP.  Addresses #24


afender:
I think we should update the python API to return a Data Frame containing <vertex_ID, sssp> so that it is consistent with other algorithms. Just need to call g.adjList.get_vertex_identifiers(...) to get the IDs, there is an example in PageRank feature: https://github.com/rapidsai/cugraph/blob/master/python/pagerank/pagerank_wrapper.pyx .

PR #47, opened 2019-01-31
BFS imported and python bindings added.

jwyles:


GPUtester:

afender:
@jwyles, since RMM hooks are now in master, it would be great to make this code RMM ready. For bfs, I think you only need to :
1. #include <rmm_utils.h>
2. replace cudaMalloc/free by ALLOC_MANAGED_TRY / ALLOC_FREE_TRY

Please sync-up with @aschaffer if needed

jwyles:
RMM hooks have been added to BFS

PR #48, opened 2019-02-01
[REVIEW] [BUG] ABI fixes

ChuckHastings:
Couldn't build master due to ABI issues.  This PR adds back some cmake entries that had previously been removed to allow for the user to turn the CXX11 ABI changes on or off universally for everything built with cugraph.

Had to make a change to one of the unit tests which failed when built using the old ABI.


afender:
@Iroy30 can you verify if that fixes your problem too?

afender:
Thanks. Merging then.

PR #70, opened 2019-02-12
Make everything build with the new abi - remove backward compatibility

ChuckHastings:
This PR removes code in the CMake files for supporting backward compatibility of the C++ ABI change.

At this point, all of the conda dependencies are built with the new ABI, there is no reason to maintain backward compatibility with version 0.6.

This PR will eliminate the need to explicitly compile with the new ABI, the default build will use the new ABI.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -28,7 +28,10 @@ include(CTest)
 ###################################################################################################
 # - compiler options ------------------------------------------------------------------------------
 
-#option(CMAKE_CXX11_ABI "Enable the GLIBCXX11 ABI" OFF)
+### NOTE:  This flag is set to ensure the RMM is compiled with the ABI on.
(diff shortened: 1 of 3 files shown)


BradReesWork:
run build

GPUtester:

BradReesWork:
run test

aschaffer:
Ack-ed. Will clone latest branch-0.6 today and give it a try. I can already confirm that everything builds w/ "cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DCMAKE_CXX11_ABI=on".

afender:
This PR looks good for the new ABI, which sholud be the default one. 

After slack discussions yesterday it seems that we would also need a way to build with the old ABI. Feel free to add it to this PR today if you can or we will do another one (whatever is simpler). 
I think that supporting the old ABI is somewhat less important than having the new ABI working so it should not block this PR for too long.

Issue #87, opened 2019-02-21
[BUG] Readme.md : add a note on how to set the ABI flag


afender:
CuDF flag is `OFF` by default for custom builds but conda packages are built with this flag `ON`
Cugraph is `ON` by default and releases conda packages with the flag `ON`

Every time there is a mismatch this creates cryptic bugs. This happened to several devs so we should document this in the Readme.

We should also double check with cudf team if this flag should be turned `ON` by default there for consistency.

afender:
Closed by #88

PR #88, opened 2019-02-21
add instructions on building with the old ABI

ChuckHastings:
Addresses #87 

Added a section to the README about the C++ ABI.  Hopefully this will clarify how to build with the different ABI issues.


afender:
closes #87

PR #97, opened 2019-02-26
[REVIEW] Speeding up python tests

### LINKED ISSUE #83: [BUG] Python tests take too long
Python tests take more than 30 min to run. This is blocking the testing platform for too long. Users may also think that something is not working correctly. 

We should identify what takes so long (afaik, most of the time is spent in NextworkX to get the reference result) and try to reduce the total time as much as possible.

jwyles:
This PR modifies the python tests to run faster. By ensuring we are not doing elementwise GPU copies during comparisons and also by using a smaller "big" dataset which the networkx algorithms don't choke on. Close #83


GPUtester:

PR #109, opened 2019-03-05
[REVIEW] Fixed BFS to generate the adjacency list if it's not present when called

### LINKED ISSUE #108: [BUG] BFS crashes when using an edge list
The cugraph BFS algorithm crashes when the input data is an edge list 

G = cugraph.Graph()
G.add_edge_list(gdf["src"], gdf["dst"])
df = cugraph.bfs(G,0)

jwyles:
Close #108


GPUtester:

PR #113, opened 2019-03-05
[REVIEW] Fixed Louvain to generate adjacency list if only edge list present

### LINKED ISSUE #110: [BUG] Louvain crashed when using an edge list
cugraph Louvain crashed when the input data is an edge list

# create a Graph 
G1 = cugraph.Graph()
G1.add_edge_list(gdf["src"], gdf["dst"])
df1, mod1 = cugraph.nvLouvain(G1)

jwyles:
Close #110


GPUtester:

Issue #112, opened 2019-03-05
[BUG] cuGraph needs to use latest version of RMM


BradReesWork:
current version of cuGraph is using an old version of RMM.  User will not know that they need to do a git checkout of an old RMM version.  

cuGraph needs to be updated to use the latest version.  This including updating code to the latest API

PR #115, opened 2019-03-06
[REVIEW] Bug use cudf 0.6

ChuckHastings:
Addresses #112 and a bit more.

This PR actually includes several things encountered while trying to build cugraph with cudf/rmm version 0.6.  Biggest changes here:

1.  RMM was removed as a thirdparty reference.  Instead, RMM is picked up from the cudf installation.  Added instructions in the cmake file to discover RMM wherever we discovered cudf and removed the instructions to build it from source.

1.  Created a second conda environment file that points to the nightly build server.  This should make it easier for developers of a new release to point to new releases - while users of a specific release can point to the published artifacts.

1.  One of the unit tests assumed that cuda memory usage would go up and down as memory is allocated.  Now that we're using RMM that is no longer true.  I commented out the code rather than deleting it, perhaps there is an RMM call that would be reasonable if we think this should be tested.

1.  Made some changes in c_graph.pxd and c_graph.pyx to accommodate a change in the function signature of one of the cudf functions we call.  Started a dialog about the longer-term solution, but that will be addressed in 0.7.


CHANGELOG.md
```diff
@@ -28,6 +28,7 @@
 - PR #48 ABI Fixes
 - PR #72 Bug fix for segfault issue getting transpose from adjacency list
 - PR #110 Bug fix for segfault calling Louvain with only edge list
+- PR #115 Fixes for changes in cudf 0.6, pick up RMM from cudf instead of thirdpary
```
BradReesWork: File is out of date, will have to sync on merge

cpp/src/tests/gdf_graph/gdf_graph.cu
```diff
@@ -243,21 +243,21 @@ TEST(gdf_delete_adjacency_list, success1)
       
   gdf_graph G;
   gdf_column col_off, col_ind, col_w;
-  size_t free, free2, total;  
-  cudaMemGetInfo(&free, &total);
+  //size_t free, free2, total;  
+  //cudaMemGetInfo(&free, &total);
```
BradReesWork: why comment out lines whether than removing them?

ChuckHastings: In my description above I mentioned that it was unclear to me whether we want to not check this information at all, or if there was some RMM calls we could make that would be equivalent.  I left the commented lines in to know where we had those error checks.

I'm find deleting them instead, I just didn't want to delete something from the unit tester that someone really thinks we should be checking.

PR #129, opened 2019-03-11
[REVIEW] New README 

BradReesWork:


README.md
```diff
@@ -1,310 +1,320 @@
...
+
+
+
+
+
+
+## Getting cuGraph
+### Intro
+There are 4 ways to get cuGraph :
+1. [Quick start with Docker Demo Repo (Coming Soon)](#quick)
```
ChuckHastings: We say (Coming Soon) on 3 of these.  Is the expectation that this will be available for 0.6?  How soon?  It seems like some sort of time frame (less ambiguous than soon) would be helpful here.

README.md
```diff
@@ -1,310 +1,320 @@
...
+  ```bash
+  # Set the localtion to cuGraph in an environment variable CUGRAPH_HOME 
+  export CUGRAPH_HOME=$(pwd)/cugraph
+
+  # Download the cuGraph repo
+  git clone https://github.com/rapidsai/cugraph.git $CUGRAPH_HOME
+
+  # Next load all the submodules
+  cd $CUGRAPH_HOME
+  git submodule update --init --remote --recursive
```
ChuckHastings: I'm still not sure about --remote when we start branching.  It feels like --remote should be used when you're on the development branch.  We should - assuming we continue to do the thirdparty submodule route - pin the versions on a branch, in which case this command does not need the --remote option.

However, more importantly for version 0.6 - this step is unnecessary.  I removed the only submodule that we were using.  We may add it back in version 0.7.  So my inclination would be to just cut the submodule step from this README for now and add it back in once we need it.

README.md
```diff
@@ -1,310 +1,320 @@
...
+
+
+
+​	B) Create the conda development environment `cugraph_nightly`  
+
+If you are  on the latest development branch then you must use the `cugraph_nightly` environment.  The latest cuGraph code uses the latest cuDF features that might not yet be in the master branch.  To work off of the latest development branch, which could be unstable, use the nightly build environment.  
+
+```bash
+# create the conda environment (assuming in base `cugraph` directory)
+conda env create --name cugraph_nightly --file conda/environments/cugraph_nightly.yml
```
ChuckHastings: We should probably add a coda 10 version of the cugraph_nightly.yml file.  Perhaps we should add an issue for this?  This could be deferred until 0.7, since the nightly builds won't be necessary to build with 0.6 once we release.

README.md
```diff
@@ -1,310 +1,320 @@
...
+# activate the environment
+conda activate cugraph_nightly 
+
+```
+
+
+  - The environment can be updated as development includes/changes the dependencies. To do so, run:  
+
+    ```bash
+    # for CUDA 9.2
```
ChuckHastings: This only covers the 9.2 version.

Would it be better to list the yml files and provide the structure for the command:

conda env update --name <profilename> --file <environment_yml_file>

README.md
```diff
@@ -1,310 +1,320 @@
...
+
+
+
+
+#### Run tests
+
+6. Run either the standalone tests or the Python tests with datasets
+  - **C++ stand alone tests** 
+
+    From the build directory : `gtests/gdfgraph_test`
```
ChuckHastings: Why do we need this line?  If we think we need it, gdfgraph_test is now GDF_GRAPH_TEST.  But it feels like the next section of code describes this.

README.md
```diff
@@ -1,310 +1,320 @@
...
+
+    ```bash
+    # Run the cugraph tests
+    cd $CUGRAPH_HOME
+    cd cpp/build
+    gtests/GDFGRAPH_TEST		# this is an executable file
```
ChuckHastings: I typically run make test at this point.  make test runs ALL of the unit tests.  tests/GDFGRAPH_TEST only tests the C++ side of the python to C++ bindings.

README.md
BradReesWork: Those will be there for 0.6 - I believe that we should include them now but I defer to @taureandyernv

README.md
BradReesWork: I'll update and remove `-remote`

README.md
BradReesWork: Let's push it since it only impact new feature developers who can edit the file

README.md
BradReesWork: yep :(

README.md
BradReesWork: That should have been updated to the correct syntax

README.md
BradReesWork: Removed

README.md
BradReesWork: new yml file created

README.md
BradReesWork: updated to new yml file

README.md
BradReesWork: I can replace gtests/GDFGRAPH_TEST with make test

BradReesWork:
I am going to removed the "Coming Soon" references.   Then push towards then end of the week so that it is part of the big roll-out.   Since we stress that non-master branches are unstable, I think it is fine that the links do not work

Issue #153, opened 2019-03-15
[FEA] Try a sort-free coo2csr to reduce coo2csr memory requirement on SNSG


afender:
**Is your feature request related to a problem? Please describe.**
We identified that cugraph’s peak memory requirement is in coo2csr. The current implementation sorts the edge list and then builds the CSR matrix

**Describe the solution you'd like**
The sort needs a temporary buffer equivalent ~2 edge lists. Andrei has a sort-free GPU implementation that uses a simple form of hashing. 

**Describe alternatives you've considered**
- UVM in RMM
- Manual memory management + sort that can stream data (much more complex to develop)
- Use Multiple GPUs (long term solution - need something sooner)

**Task List**

- [ ] get a memory profile 
- [ ] get performance results on large datasets 
- [ ] compare (time and memory) to what we have currently in cuGraph

(if better than what we have in cugraph)
- [ ] integrate in cugraph
- [ ] run all tests and verify that there is no unexpected regression

ChuckHastings:
Current status of my coo2csr exploration.  For my arithmetic below, n = size (the number of vertices - technically the maximum vertex id), m = nnz (the number of edges).

The original code:
 * makes a copy of source and destination arrays (since we can't corrupt the original)
 * sorts the copies to order the COO, then computes the starting point for the CSR.

This uses ~ m + 3n extra memory (one of the n's is an estimate guessed from what the CUB call seems to require) that is freed at the end of the call.


I implemented a coo2csr that does not require sorting the COO.  This prevents the need to copy the source and destination array.  My algorithm constructs the CSR in a non-deterministic way directly from the original arrays.  It then does a segmented sort of the resulting array (segmented sort should be faster than a full sort - theoretically).

Construction of the unsorted CSR is markedly faster than the original algorithm.  If we don't need the edge lists to be sorted, the technique works very well.  When adding the segmented sort - at least for now - the total cost is comparable (about 10-20% slower).

The one exception seems to be the rmat graph that I generated for testing.  In this case both the creation of the unsorted CSR and the sorting are markedly slower.

There is a performance enhancement for constructing the CSR that I haven't put in yet.  The segmented sort still requires tuning.

My algorithm only requires an additional (1 + n) memory, everything else is done in place - in the output arrays so that the original input is preserved.

It seems like my algorithm would translate to SNMG, but I haven't specifically laid out how.

BradReesWork:
dropped - new OPG version will address memory usage

PR #165, opened 2019-03-18
[WIP] Fea sort free coo to csr

ChuckHastings:
Implementation of COO To CSR without sorting.

If you replace calls to ConvertCOOtoCSR with calls to ConvertCOOtoCSR_unsorted the unit tests all succeed and use the new algorithm.


afender:
issue #153

afender:
Since the col Indices array is populated in a non-deterministic way, I suppose we would need an extra step to sort it for each row/vertex?

ChuckHastings:
Oded mentioned the idea of either adding a flag to the graph structure indicating whether the result is sorted or not, or apparently he has code that will sort a CSR that is not sorted.

afender:
We should account for this sort step then. I am not aware of a single cuGraph feature that has been tested on unsorted inputs. We should check if merge-path spmv can handle this. I know for sure that some features like Jaccard rely on the property that indices are sorted.

cpp/src/COOtoCSR.cuh
```diff
@@ -67,6 +67,81 @@ __global__ void offsetsKernel(T runCounts, T* unique, T* counts, T* offsets) {
...
+               temp_array, temp_array + result.size + 1, result.rowOffsets);
+
+  //
+  //  Now populate the colIndices array (non-deterministic)
+  //
+  thrust::for_each(thrust::cuda::par(allocator).on(stream),
+                   thrust::make_counting_iterator<int64_t>(0),
+                   thrust::make_counting_iterator<int64_t>(nnz),
+                   [temp_array, sources, destinations, result]
+                   __device__ (int64_t id) {
```
afender: Do I understand correctly that the use of `int64_t` here allows to run on graphs that need 64bit for the number of edges but where vertices Ids can be 32bits?

cpp/src/COOtoCSR.cuh
```diff
@@ -67,6 +67,81 @@ __global__ void offsetsKernel(T runCounts, T* unique, T* counts, T* offsets) {
...
+                     T offset = atomicAdd(temp_array + sources[id], T{1});
+                     result.colIndices[offset] = destinations[id];
+                   });
+
```
afender: We should add the local sorts for each vertex here as per the earlier conversations. This can be an option but should be enabled by default for now. This would also allow doing an apple to apple comparison against the existing solution.

afender:
I like the clear thrust-based implementation of this coo2csr.

It would be good to know more about how this compares to the existing code regarding memory usage. For instance, it would be great to identify the largest graph it can handle.

Since there are many atomics I'd also recommend double checking how it scales (in term of execution time).

Finally, in order to replace the existing coo2csr we would also need to add :
- sorted indices
- support for the values/weights

cpp/src/COOtoCSR.cuh
ChuckHastings: I literally copied James' function and adapted it, and he used an int64_t for that parameter.

Yes, I believe that an int64_t here would allow us to have > 2 billion edges with < 2 billion vertices.

afender:
Chuck, do you want to add the latest version on top of this PR? Otherwise, I can close it if you are planning on creating a new one.

Issue #167, opened 2019-03-19
[BUG] view_adj_list() results in crash on weighted grapha during jaccard and sssp algorithms


afender:
**Describe the bug**
View_adj_list() and view_tanspose_adj_list() result in crash (sssp and jaccard) only when the graph has weights.
The crash happens in an RMM_FREE but in different ways as SSSP hangs and jaccard doesn't.

**To Reproduce**
I used cugraph_time.py : https://drive.google.com/file/d/1nIoL6ORhbvf2qzvEdvj-nW-OCCdPmpUJ/view?usp=sharing

```
python cugraph_time.py /datasets/networks/netscience.mtx --algo jaccard --auto_csr 0
python cugraph_time.py /datasets/networks/netscience.mtx --algo sssp --auto_csr 0
```

**Expected behavior**
We expect this to pass

**Screenshots**
```
(cugraph_dev) afender@strata3:~/sw$ python cugraph_time.py /datasets/networks/netscience.mtx --algo sssp --auto_csr 0
280.143,0.001,Traceback (most recent call last):
  File "cugraph_time.py", line 110, in <module>
    sssp_time = sssp_call(G, args.source)
  File "cugraph_time.py", line 48, in sssp_call
    df = cugraph.sssp(G,start)
  File "cugraph/sssp/sssp_wrapper.pyx", line 25, in cugraph.sssp
  File "cugraph/sssp/sssp_wrapper.pyx", line 71, in cugraph.sssp
  File "cudf/bindings/cudf_cpp.pyx", line 216, in cudf.bindings.cudf_cpp.check_gdf_error
  File "cudf/bindings/cudf_cpp.pyx", line 234, in cudf.bindings.cudf_cpp.check_gdf_error
cudf.bindings.GDFError.GDFError: CUDA ERROR. b'cudaSuccess': b'no error'
terminate called after throwing an instance of 'std::runtime_error'
  what():  ERROR: RMM runtime call  RMM_FREE( (col->data), (nullptr) )an illegal memory access was encountered
^C^C^C^C^C^C^C^CAborted (core dumped)
(cugraph_dev) afender@strata3:~/sw$ python cugraph_time.py /datasets/networks/netscience.mtx --algo jaccard
ERROR:Call to cuMemcpyHtoD results in UNKNOWN_CUDA_ERROR
0.63850998878479,0.003723621368408203Traceback (most recent call last):
  File "cugraph_time.py", line 114, in <module>
[...]

PR #181, opened 2019-03-22
[REVIEW] fix crash/rmm free error when edge values present

Iroy30:
Fixes Issue #167


afender:
The code looks good. Just need to add a test.

Issue #170, opened 2019-03-20
[FEA] cuGraph needs to provide an option to deep copy cudf dataframe objects when creating a cugraph instance


seunghwak:
**Is your feature request related to a problem? Please describe.**
cuGraph creates a graph from cudf dataframe objects which store graph data in column arrays. A created cuGraph instance shares the memory locations with cudf dataframe objects. If cudf dataframe objects are modified, the cugraph instance's data get changed as well. This behavior is not always desirable.

Pandas dataframe provide an option to create a new instance by either deep copying or shallow copying (https://stackoverflow.com/questions/44993846/pandas-dataframe-mutability). cuGraph needs to provide this option as well.

**Describe the solution you'd like**
cudf class Column already provides the copy function with input parameter deep. This issue may better be addressed together with https://github.com/rapidsai/cugraph/issues/168

**Task List**
1. First solve https://github.com/rapidsai/cugraph/issues/168
2. Then, update cugraph's API similar to that of Pandas and properly set the input parameter deep in creating column arrays.

afender:
CuGraph still needs to be able to reuse cuDF's memory directly without a copy. This should remain the default behavior.

seunghwak:
Got it, and copy is set to False by default in Pandas as well (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.html). So I think we should set copy to False by default as well.

afender:
Closing after #190

Issue #168, opened 2019-03-20
[BUG] cugraph's column data become invalid if cudf dataframe objects are deleted or garbage collected.


seunghwak:
**Describe the bug**
cuGraph's python class Graph creates a graph instance from cudf dataframe objects which hold graph data in column arrays. The Graph class currently uses C++ struct gdf_graph internally and passes pointers to the column arrays' data to a gdf_graph instance. However, this C++ struct gdf_graph does not participate in python's reference counting based garbage collection, and when the cudf dataframe objects are deleted or garbage collected, the memories assigned to the column arrays get reclaimed (python has no idea these memories are still used by C++); the pointers to the column arrays (in C++ struct gdf_graph) are no longer valid.

**To Reproduce**
Accessing invalid memory locations can lead to undefined behaviors.

**Expected behavior**
cuGraph instances should participate in python's reference counting in garbage collection for column arrays. cudf provides python class Column. If we use this Column to store column arrays, this problem problem gets solved. This maximizes code reuse as well and this is better software engineering practice in general than recreating our own solution.

The Column class provides the copy function with input parameter deep(=True). If deep is set to False, this performs shallow copy and shares memory similar to the current cuGraph implementation. If deep is set to True, deep copy is performed and creates an independent copy.

afender:
closing after PR #190

PR #190, opened 2019-03-27
[REVIEW] Bug cudf & cugraph interoperability issue

### LINKED ISSUE #158: [BUG] Get a warning during cugraph/python install from source
During `python setup.py install` when installing cuGraph from source, I get the following deprecation warning:
```
/conda/envs/rapids/lib/python3.7/site-packages/Cython/Compiler/Main.py:367: FutureWarning: Cython directive 'language_level' not set, using 2 for now (Py2). This will change in a later release! File: /cugraph/python/cugraph/cugraph.pyx
  tree = Parsing.p_module(s, pxd, full_module_name)
```

@dantegd fyi

seunghwak:
close #158, #168, #170

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,13 @@
 ## Improvements
 - PR #157 Removed cudatoolkit dependency in setup.py
 - PR #185 Update docs version
-- PR #190 Added a copy optoin in graph creation
+- PR #190 Added a copy option in graph creation
 
 ## Bug Fixes
 - PR #169 Disable terminal output in sssp 
 - PR #181 Fixed crash/rmm free error when edge values provided
 - PR #190 Fixed a memory reference counting error between cudf & cugraph
-- PR #190 Fixed a langugae level warning (cython)
+- PR #190 Fixed a language level warning (cython)
 
 # cuGraph 0.6.0 (22 Mar 2019)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,13 @@
 ## Improvements
 - PR #157 Removed cudatoolkit dependency in setup.py
 - PR #185 Update docs version
+- PR #190 Added a copy optoin in graph creation
 
 ## Bug Fixes
 - PR #169 Disable terminal output in sssp 
 - PR #181 Fixed crash/rmm free error when edge values provided
+- PR #190 Fixed a memory reference counting error between cudf & cugraph
+- PR #190 Fixed a langugae level warning (cython)
 
 # cuGraph 0.6.0 (22 Mar 2019)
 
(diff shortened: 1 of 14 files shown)


raydouglass:
rerun tests

CHANGELOG.md
```diff
@@ -7,10 +7,13 @@
 ## Improvements
 - PR #157 Removed cudatoolkit dependency in setup.py
 - PR #185 Update docs version
+- PR #190 Added a copy optoin in graph creation
 
 ## Bug Fixes
 - PR #169 Disable terminal output in sssp 
 - PR #181 Fixed crash/rmm free error when edge values provided
+- PR #190 Fixed a memory reference counting error between cudf & cugraph
+- PR #190 Fixed a langugae level warning (cython)
```
ChuckHastings: Typo here (language)

cpp/include/functions.h
```diff
@@ -85,7 +85,7 @@ gdf_error gdf_add_adj_list(gdf_graph *graph);
  */
 /* ----------------------------------------------------------------------------*/
 
-gdf_error gdf_add_transpose(gdf_graph *graph);
+gdf_error gdf_add_transposed_adj_list(gdf_graph *graph);
```
ChuckHastings: Question... would this be better named gdf_add_transposed_from_adj_list (same comment applies to all)?  My first "read" of this function name suggests that it is adding an adj_list... which is in fact the opposite of what it is doing.

ChuckHastings: By the same token, if we go the route I suggest should gdf_add_edge_list be renamed to gdf_add_edge_list_from_adj_list?

CHANGELOG.md
seunghwak: Thanks, I'd better run spelling checker next time :-)

cpp/include/functions.h
seunghwak: Thank you very much for raising this issue. I had similar doubts. So this is related to a bigger API design issue.

First there are naming inconsistencies between python functions and C++/CUDA functions; e.g. add_edge_list in python add_edge_list to the graph from a user provided edge list but gdf_add_edge_list does conversion. And add_edge_list calls C++/CUDA gdf_edge_list_view (not gdf_edge_list).

Second some functions may surprise users; e.g. view_edge_list, view_adj_list, and num_vertices can (implicitly) involve graph conversion.

So we may need to review APIs related to class Graph and C++/CUDA counter part, but this affects multiple parts, I guess we need to solicit options from the entire group. I will submit a separate issue to discuss this.

afender:
@Iroy30 any feedback on this? 
@ChuckHastings do you approve the PR after the changes?

ChuckHastings:
I'm OK tackling the broader issue in #192

Issue #164, opened 2019-03-18
[BUG] segfault when values array is not set


afender:
**Describe the bug**
Segfault in Louvain when the value array is not set. Currently, the user has to manually fill a column with `1.0` and attach it to the graph in order to run on unweighted graphs.
==> This also applies to spectral clustering
**To Reproduce**
Remove values from the input graph in the test.

**Expected behavior**
Python bindings should provide an array of `ones` to the C++ API when they don't already exist in the input graph.

Iroy30:
can be closed

afender:
done after #193

PR #193, opened 2019-03-29
[REVIEW] fix segfault when edgevals not provided

Iroy30:
Fixes issue #164

### DIFF
--- a/cpp/src/nvgraph_gdf.cu
+++ b/cpp/src/nvgraph_gdf.cu
@@ -157,6 +157,7 @@ gdf_error gdf_createGraph_nvgraph(nvgraphHandle_t nvg_handle,
 					return GDF_UNSUPPORTED_DTYPE;
 			}
 		}
+
 	}
 	else {
 		// convert edgeList to adjList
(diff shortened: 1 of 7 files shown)


cpp/src/nvgraph_gdf.cu
```diff
@@ -278,15 +279,45 @@ gdf_error gdf_balancedCutClustering_nvgraph(gdf_graph* gdf_G,
...
 	// Ensure that the input graph has values
 	GDF_TRY(gdf_add_adj_list(gdf_G));
-	GDF_REQUIRE(gdf_G->adjList->edge_data != nullptr, GDF_INVALID_API_CALL);
+	//GDF_REQUIRE(gdf_G->adjList->edge_data != nullptr, GDF_INVALID_API_CALL);
 
 	// Initialize Nvgraph and wrap the graph
 	nvgraphHandle_t nvg_handle = nullptr;
 	nvgraphGraphDescr_t nvgraph_G = nullptr;
+        cudaDataType_t settype;
+        Vector<double> d_val;
```
afender: Why do we need double instead of float?

python/cugraph/spectral_clustering/test_balanced_cut.py
```diff
@@ -53,6 +53,25 @@ def random_call(G, partitions):
 @pytest.mark.parametrize('graph_file', DATASETS)
 @pytest.mark.parametrize('partitions', PARTITIONS)
 def test_modularity_clustering(graph_file, partitions):
+    # Read in the graph and get a cugraph object
+    M = read_mtx_file(graph_file).tocsr()
+    row_offsets = cudf.Series(M.indptr)
+    col_indices = cudf.Series(M.indices)
+    G = cugraph.Graph()
+    G.add_adj_list(row_offsets, col_indices)
```
afender: Can we also test on edge list?

cpp/src/nvgraph_gdf.cu
afender: Synced up with Ishika onsite, f64 is more stable. Keeping 64 bit by default.

Issue #33, opened 2019-01-04
[FEA][NV] Move nvgraph repo in cugraph repo


afender:
This task will result in open sourcing nvGraph. NvGraph contains the core CUDA code for Jaccard, Louvain, BFS, SSSP and more. 

The goal is to open source all at once to facilitate the integration and minimize the refactoring effort.

Before moving nvGraph repo here, the following tasks must be done in nvGraph:
- [x] Use cmake infrastructure (see #34)
- [x] Review for IP
- [x] Integrate recent changes from perforce (if any)

Then : 
- [x] Copy nvgraph code in cugraph repo:  code should be in a nvgraph folder under cpp/nvgraph
- [x] Add build option to produce nvgraph library
- [x] rename nvgraph stand alone library `nvgraph_st` into `nvgraph_rapids`
- [x] Update cugraph infrastructure to build the newly added nvpraph code

Important improvement
- [ ] Remove multiple CUB dependencies (3 versions of cub in cugraph after transferring nvgraph...)
- [x] Reduce build time (nvgraph takes minutes to build, this will negatively impact cugraph development)

afender:
Open sourcing now would reduce the deployment overhead and remove an unnecessary complexity layer in cugraph. Currently, nvgraph has to be updated on anaconda.com each time something change which is a source of bugs and slows down development.

afender:
I was able to move nvgraph in cugraph repo and successfully completed the subtasks above. Nvgraph is automatically built with cugrph but can also be built as a stand-alone. It still takes at least 1 minute to build after optimizations. This works locally but Jenkins returns errors inside cub currently

afender:
closed by #194

PR #194, opened 2019-03-29
[review] Open source nvgraph in cugraph repository

afender:
This is the PR for #33 

Contains:
- [x] Copy nvgraph code in cugraph repo under cpp/nvgraph
- [x] Integrate recent changes from perforce and minor cleanup after reviews
- [x] Add a path to produce nvgraph library only
- [x] Add a path to automatically build nvpraph code when building cugraph
- [x] rename nvgraph stand-alone library `nvgraph_st` into `nvgraph_rapids`
- [x] Reduce nvgraph build time by disabling undocumented features when built from cugraph (got 3x speedup)
- [x] update readme(s)


afender:
Hi @raydouglass ,

I can repro the error I see on Jenkins only when I get CUB submodule with the command: `git submodule update --init --recursive --remote cpp/nvgraph/cpp/thirdparty/cub` .

It works fine when I get cub using `git submodule update --init --recursive` or even `git clone --recurse-submodules  git@github.com:afender/cugraph.git`. 

That sounds like an easy fix we could do in CI, correct?

raydouglass:
> Hi @raydouglass ,
> 
> I can repro the error I see on Jenkins only when I get CUB submodule with the command: `git submodule update --init --recursive --remote cpp/nvgraph/cpp/thirdparty/cub` .
> 
> It works fine when I get cub using `git submodule update --init --recursive` or even `git clone --recurse-submodules git@github.com:afender/cugraph.git`.
> 
> That sounds like an easy fix we could do on Jenkins, correct?

Unfortunately not. Jenkins uses the former method to update submodules.

You should be able to resolve this by add branch information to `.gitmodules` like cuDF: https://github.com/rapidsai/cudf/blob/branch-0.7/.gitmodules
You likely want to use the same cub repository and branch as cuDF.

afender:
> You should be able to resolve this by add branch information to .gitmodules 

Thanks, I added branches to submodules and C++ builds pass now.

However, we have another problem on Jenkins for the python part. Cugraph now build and installs nvgraph (under `${CMAKE_INSTALL_PREFIX}`) but the python setup cannot find it on Jenkins later on (it searches in `CONDA_PREFIX/lib` and `LD_LIBRARY_PATH`).

@raydouglass  how should we update the `setup.py` so that it can find libnvgraph_rapids.so on Jenkins?

afender:
@raydouglass ⬆️

raydouglass:
@afender I opened a PR against your branch: https://github.com/afender/cugraph/pull/8
I think that will resolve the problem.

afender:
@raydouglass I was just talking to @rlratzel about conda packaging changes resulting from this PR in 0.7 release. I lost track if the updates have been done or not. There are basically two things I wanted to double check with you:

1. We should keep releasing nvgraph as a requirement. We should just make sure that `conda install nvgraph` reflects what's in https://github.com/rapidsai/cugraph/tree/branch-0.7/cpp/nvgraph rather than https://gitlab-master.nvidia.com/RAPIDS/nvgraph now.

2. Now that nvgraph lives inside cugraph and is built along with it, should we update the way the cugraph conda packadge is built in 0.7? I suspect there is something to do so that `conda install cugraph` also installs nvgraph lib and headers?

PR #195, opened 2019-03-29
2 hop jaccard

### LINKED ISSUE #136: [FEA] Add option to Jaccard to support 2-hop comparisons
Currently Jaccard only compute a score if the vertices are connected by an edge.  However there are analytics user cases where the user wants to find similarities between unconnected vertices (max is 2-hops).  Computing the 2-hop comparison can be computationally expensive, therefore having it as an option might be preferred.

jwyles:
Moved jaccard code from Nvgraph into cugraph, added get_two_hop_neighbors() method to graph object which returns a list of vertex pairs such that each pair is connected by a two-hop path in the graph, added a new C-side method to compute jaccard coefficients given a list of vertex pairs, modified jaccard and weighted Jaccard to accept optional 'first' and 'second' arguments giving the vertex pairs to compute for. Added tests for Graph.get_two_hop_neighbors() and using list arguments (with two hop neighbors) for Jaccard. close #136


GPUtester:

cpp/src/graph_utils.cuh
```diff
@@ -41,489 +40,554 @@ namespace cugraph
...
-        //notice that no __threadfence or __syncthreads are needed in this implementation
-    }
-    //get the value of the last thread (to all threads)
-    last = shfl(sum, blockDim.x-1, blockDim.x);
-
-    return last;
-}
+	}
+
+	template<typename IdxType, typename ValType>
```
afender: It seems that the encoding or indentation has been overwritten in this file

cpp/include/cugraph.h
```diff
@@ -25,4 +25,6 @@
 
 #include "functions.h"
 
+#include "algorithms.h"
```
afender: Why do we need `algorithms.h` to be separated from `functions.h`? 
This may lead to an inconsistency with cuDF which is using `functions.h` only.

python/cugraph/jaccard/jaccard_wrapper.pyx
```diff
@@ -34,7 +33,7 @@ cpdef nvJaccard(input_graph):
```
afender: We should document the newly added parameters

python/cugraph/jaccard/test_jaccard.py
```diff
@@ -145,3 +145,26 @@ def test_jaccard_edgevals(graph_file):
...
+
+@pytest.mark.parametrize('graph_file', DATASETS)
+def test_jaccard_two_hop(graph_file):
+    M = read_mtx_file(graph_file)
+    M = M.tocsr()
+    Gnx = nx.DiGraph(M).to_undirected()
+    G = cugraph.Graph()
+    row_offsets = cudf.Series(M.indptr)
+    col_indices = cudf.Series(M.indices)
+    G.add_adj_list(row_offsets, col_indices, None)
```
afender: We need to test on edge lists too in 0.7.

python/cugraph/jaccard/wjaccard_wrapper.pyx
```diff
@@ -11,18 +11,19 @@
...
 import cudf
 from libgdf_cffi import libgdf
 from librmm_cffi import librmm as rmm
 import numpy as np
+from numpy.core.numeric import result_type
 
 gdf_to_np_dtypes = {GDF_INT32:np.int32, GDF_INT64:np.int64, GDF_FLOAT32:np.float32, GDF_FLOAT64:np.float64}
 
-def nvJaccard_w(input_graph, weights):
+cpdef jaccard_w(input_graph, weights, first=None, second=None):
```
afender: Comments in jaccard.pyx also apply to this file

python/cugraph/graph/c_graph.pyx
```diff
@@ -249,19 +249,58 @@ class Graph:
...
         # g.transposedAdjList.offsets.data and g.transposedAdjList.indices.data
         # are not owned by this instance, so should not be freed here (this
         # will lead to double free, and undefined behavior).
 
         return cudf.Series(offsets_data), cudf.Series(indices_data)
+
+    def get_two_hop_neighbors(self):
+        """
+        Return a dataframe containing vertex pairs such that each pair of vertices is 
+        connected by a path of two hops in the graph.
```
afender: We should mention if this is sorted or not.

cpp/include/cugraph.h
jwyles: The reasoning is to avoid having a too large file with everything in it. functions.h covers functions on the graph object and algorithms covers algorithms which act on a graph object.

python/cugraph/jaccard/jaccard_wrapper.pyx
jwyles: Will do.

python/cugraph/jaccard/test_jaccard.py
jwyles: I will add a test.

python/cugraph/graph/c_graph.pyx
jwyles: Done, it is sorted.

python/cugraph/jaccard/wjaccard_wrapper.pyx
jwyles: Done.

python/cugraph/jaccard/test_jaccard.py
jwyles: Done.

python/cugraph/jaccard/jaccard_wrapper.pyx
jwyles: Done.

python/cugraph/graph/c_graph.pyx
afender: Thanks for clarifying. Is it sorted by "src" and then by "dest" or is this an unstblabe sort? As far as I remember, jaccard implementation needed both to be sorted becasue of the binary search implementation.

If Jaccard needs pairs to be sorted, we should also say it in Jaccard doc. If not, we may need test on unsorted data.

afender:
@raydouglass I cannot find Jenkins results for this PR. What should we do you run that?

python/cugraph/graph/c_graph.pyx
jwyles: The inputs don't need to be sorted for Jaccard to work. Jaccard only depends on the adjacency of the graph being sorted, which it always is.

cpp/include/cugraph.h
afender: Ok, let's keep `algorithms.h`

raydouglass:
rerun tests

jwyles:
rerun tests

jwyles:
@raydouglass Can you make the tests run again?

jwyles:
rerun tests

python/cugraph/jaccard/jaccard_wrapper.pyx
```diff
@@ -47,43 +46,53 @@ cpdef nvJaccard(input_graph):
...
+        cudf.bindings.cudf_cpp.check_gdf_error(err)
+        df = cudf.DataFrame()
+        df['source'] = first
+        df['destination'] = second
+        df['jaccard_coeff'] = result
+        return df
 
-    err = gdf_jaccard(<gdf_graph*>g, <void*>c_gamma, <gdf_column*>NULL, <gdf_column*>weight_j_col_ptr)
-    cudf.bindings.cudf_cpp.check_gdf_error(err)
+    elif first is None and second is None:
```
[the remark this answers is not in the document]
jwyles: This has been documented, it is the default behaviour.

cpp/src/graph_utils.cuh
jwyles: I think I re-formatted it. Does it need to be undone?

python/cugraph/jaccard/jaccard_wrapper.pyx
```diff
@@ -20,9 +21,7 @@ from librmm_cffi import librmm as rmm
 import numpy as np
 from cython cimport floating
 
-gdf_to_np_dtypes = {GDF_INT32:np.int32, GDF_INT64:np.int64, GDF_FLOAT32:np.float32, GDF_FLOAT64:np.float64}
-
-cpdef nvJaccard(input_graph):
+cpdef jaccard(input_graph, first=None, second=None):
```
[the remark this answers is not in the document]
jwyles: Yes, the notebooks for Jaccard and Weighted Jaccard should be updated to reflect the renaming.

cpp/src/graph_utils.cuh
afender: It is best to avoid re-formatting because it is hard to see what changed and prevents auto-merging in case someone was editing this file too. 
If you have a quick way of undoing it, great. If not, I don't think you should spend a lot of time on that.

afender: Seems it is not straight forward to revert. We'll just keep the new version

python/cugraph/jaccard/jaccard_wrapper.pyx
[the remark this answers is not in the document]
afender: Opened #201 for this

PR #203, opened 2019-04-02
[Review] Added small datasets directly in the repo

afender:
- [x] added the datasets in a directory and deleted the tar.gz file
- [x] updated python paths to be relative
- [x] disabled tests on dblp (too long)
- [x] disabled test_two_hop_neighbors on netscience
- [x] updated readme


afender:
@raydouglass we can now remove the step that untar the datasets in `/datasets`.

As this PR removed the tar file we get the following error on Jenkins 
`tar (child): cpp/src/tests/datasets.tar.gz: Cannot open: No such file or directory`

raydouglass:
@afender Just remove this line from the gpu build script I think: https://github.com/rapidsai/cugraph/blob/branch-0.7/ci/gpu/build.sh#L78

afender:
Same Jenkins issue than PR #206 
@raydouglass could this be a CI problem?

README.md
```diff
@@ -238,17 +238,14 @@ python setup.py install    # install cugraph python bindings
     # Run the cugraph tests
     cd $CUGRAPH_HOME
     cd cpp/build
-    make test
-    
-    # alternatively, you can run individual test
     gtests/GDFGRAPH_TEST		# this is an executable file
```
seunghwak: If we're running just gtests/GDFGRAPH_TEST, shouldn't we better remove other tests? (Sorry for bringing this issue on this pull request's review... This may be a separate issue, but just in case if you can quickly fix this). cpp/src/tests/nvgraph_plugin/nvgraph_gdf_sssp.cpp and cpp/src/test/pagerank/pagerank_test.cu still have absolute paths to the test files.
e.g.
pagerank_test.cu
```cpp
188    ::testing::Values(  Pagerank_Usecase("/datasets/networks/karate.mtx", "")
```
gtests/PAGERANK_TEST leads to Segmentation fault. This has been like this for quite a while, and if we are not planning to fix this, I think it's better to just remove these tests.

python/cugraph/bfs/test_bfs.py
```diff
@@ -74,10 +74,10 @@ def base_call(M, start_vertex):
...
 
 
-DATASETS = ['/datasets/networks/dolphins.mtx',
-            '/datasets/networks/karate.mtx',
-            '/datasets/networks/polbooks.mtx',
-            '/datasets/golden_data/graphs/dblp.mtx']
+DATASETS = ['../datasets/dolphins.mtx',
+            '../datasets/karate.mtx',
+            '../datasets/polbooks.mtx',
+            '../datasets/netscience.mtx']
```
seunghwak: So this works if we run pytest under python, but this will not work if we run this under python/cugraph or python/cugraph/bfs. I sometimes do this to run just a subset of tests. This is my biggest concern on using a relative path. I thought about using an environmental variable (e.g. something like RAPIDS_CUGRAPH_DATASET_PATH) to override the default path for such cases.

python/cugraph/louvain/test_louvain.py
```diff
@@ -80,9 +80,9 @@ def networkx_call(M):
     return parts
 
 
-DATASETS = ['/datasets/networks/karate.mtx',
-            '/datasets/networks/dolphins.mtx',
-            '/datasets/networks/netscience.mtx']
+DATASETS = ['../datasets/karate.mtx',
+            '../datasets/dolphins.mtx',
+            '../datasets/netscience.mtx']
```
seunghwak: Absolute paths are still used in this file.

```python
110 DATASETS = ['/datasets/networks/karate.mtx',
111             '/datasets/networks/dolphins.mtx']
```

README.md
afender: This is a separate issue. 
`make test` should not be documented until we can provide access to the C++ datasets.

python/cugraph/bfs/test_bfs.py
afender: That correct but this is a separate issue.
@rlratzel is currently looking into adding paths as a parameter.

afender: In the meantime, you can still run one particular test with the pystest command from the main python dir.

python/cugraph/louvain/test_louvain.py
afender: Good catch! I fixed it with ced91aa

README.md
afender: I think it still worth running `gtests/GDFGRAPH_TEST` as this verifies that libcugraph.so is not corrupted (independently of any python things).

python/cugraph/bfs/test_bfs.py
afender: Yeah it seems Jenkins run it from another directory as well :( 
I need to fix that

afender: Fixed after 0d7f1d5
@rlratzel is currently working on a better solution that allows datasets to be anywhere by looking at an environment variable.

PR #207, opened 2019-04-04
[review] mtx2csv script

afender:


python/utils/mtx2csv.py
```diff
@@ -0,0 +1,46 @@
...
+import argparse
+
+
+parser = argparse.ArgumentParser(description='Convert the sparsity pattern \
+                                 of a MatrixMarket file into a CSV file. \
+                                 Each directed edge is explicitely stored, \
+                                 edges are unsorted, IDs are 0-based.')
+parser.add_argument('file', type=str,
+                    help='Path to the MatrixMarket file')
+parser.add_argument('--csv_separator_name', type=str, default="space",
```
raydouglass: You can add another parameter `choices=["space","tab","comma"]` and the parser will enforce it.

python/utils/mtx2csv.py
```diff
@@ -0,0 +1,46 @@
...
+import time
+from scipy.io import mmread
+import argparse
+
+
+parser = argparse.ArgumentParser(description='Convert the sparsity pattern \
+                                 of a MatrixMarket file into a CSV file. \
+                                 Each directed edge is explicitely stored, \
+                                 edges are unsorted, IDs are 0-based.')
+parser.add_argument('file', type=str,
```
raydouglass: If you use `type=argparse.FileType()` then the `file` arg will be a file-like object. And the parser will check that it exists and is readable, plus support using `-` for bash piping.

python/utils/mtx2csv.py
afender: addressed in f3a16b4

python/utils/mtx2csv.py
afender: addressed in f3a16b4

PR #218, opened 2019-04-12
Update c_graph.pyx

afender:
Fixing a typo reported by Ishika


afender:
> The change looks good to me but the build is failing (asking to update CHANGELOG.md). This is so trivial, so I am OK about skipping this, but it seems like the build system is not happy about this :-)

Yes, I updated it while you were writing your comment ;)

PR #220, opened 2019-04-15
Triangle count

### LINKED ISSUE #52: [FEA] Fix and Add triangle counting
The first step is to add the C++ code to access nvgraph TC features when using cugraph graph structure.  Since Triangle Counting code is quite self-contained, it can be refactored to get rid of nvgraph dependencies and moved inside cugraph. In this case, the CUDA code needs to be updated to support RMM. Alternatively, it is possible to just add a C++ driver in cugraph to call nvgraph API by using the helper functions provided in `nvgraph_gdf.cu`.

The second step consists of adding python bindings for the TC feature

Subtasks checklist for python:

- [ ]  Python bindings and API (may include some adjustments of the C++/CUDA part)
- [ ]  Pytest testing against NetworkX
- [ ]  Documentation
     - [ ]  Doxygen (C++)
     - [ ]  Sphinx (Python)
     - [ ]  python code example
- [ ]  Check headers and licenses
- [ ]  Validate everything in Jenkins

jwyles:
Added in python bindings for Nvgraph triangle counting. Fixed bugs in Nvgraph triangle counting that were causing our results to differ from Networkx results. Close #52


cpp/nvgraph/cpp/src/triangles_counting.cpp
```diff
@@ -58,7 +58,7 @@ TrianglesCount<IndexType>::~TrianglesCount()
 template <typename IndexType>
 void TrianglesCount<IndexType>::tcount_bsh()
 {
-    //printf("TrianglesCount: %s\n", __func__); fflush(stdout);
+    printf("TrianglesCount: %s\n", __func__); fflush(stdout);
```
afender: we should disable internal prints (also applies to line 90, 138, 179, 197)

jwyles: Will do. Thanks for catching that.

jwyles: Done.

PR #232, opened 2019-04-19
[REVIEW] Bug ext create column

### LINKED ISSUE #186: [BUG] gdf_col_delete frees only data (does not free valid)
**Describe the bug**
gdf_column has three pointers data, valid, and col_name, and these pointers can point dynamically allocated memory.

gdf_col_delete in cpp/src/cugraph.cu frees col->data but does not free col->valid and col->col_name.

```C++
void gdf_col_delete(gdf_column* col) {
  if (col) {
    col->size = 0;
    if(col->data) {
      ALLOC_FREE_TRY(col->data, nullptr);
    }
    ...
}
```

col_name may be removed in the future (https://github.com/rapidsai/cudf/issues/1297), so we may not need to free col_name but we need to check col->valid and free col->valid if col->valid is not nullptr.

And we may not create our own version of gdf_col_delete but better use cudf's gdf_column_free (but we cannot use it as is because it does not check col->data and col->valid are not nullptr).

### LINKED ISSUE #189: [BUG] Memory leaks after invoking create_column
**Describe the bug**
create_column function is defined in python/graph/c_graph.pyx. Inside this function, the following function heap allocates memory of size sizeof(gdf_column) and returns this memory. The caller is responsible for freeing this memory.

```cython
cdef gdf_column * c_col = < gdf_column *> malloc(sizeof(gdf_column))
```

create_column is called in wjaccard_wrapper.pyx, jaccard_wrapper.pyx, spectral_clustering.pyx, sssp_wrapper.pyx, bfs_wrapper.pyx, louvain_wrapper.pyx, pagerank_wrapper.pyx (in addition to c_graph.pyx), but they don't free memory allocated by create_column. This leads to memory leaks.

**Expected behavior**
Option 1) call delete_column defined in c_graph.pyx
Option 2) instead of returning raw pointer, return unique_ptr to transfer ownership (see https://dmtn-013.lsst.io/ for using std::unique_ptr in cython).
Option 3) gdf_column is not big (it will not blow stack) and there is no strong reason to heap allocate this variable (don't get confused by the sizes of the arrays pointed by the pointer member variables in gdf_column, they can be big and they need to be heap allocated). Update create_column to return gdf_column (not pointer to gdf_column). We may need to rename create_column if create_... implies a factory function with memory allocation in RAPIDS's naming convention.

seunghwak:
Closes #186, Closes #189.

### DIFF
--- a/python/cugraph/graph/c_graph.pyx
+++ b/python/cugraph/graph/c_graph.pyx
@@ -375,6 +375,13 @@ class Graph:
 
         return cudf.Series(offsets_data), cudf.Series(indices_data)
 
+    def delete_transposed_adj_list(self):
+        """
+        Delete the transposed adjacency list.
+        """
+        cdef uintptr_t graph = self.graph_ptr
+        err = gdf_delete_transposed_adj_list(<gdf_graph*> graph)
+        cudf.bindings.cudf_cpp.check_gdf_error(err)
 
     def get_two_hop_neighbors(self):
         """
@@ -414,22 +421,6 @@ class Graph:
             df['second'] = second_out
 
         return df
-
-    def delete_adj_list(self):
-        """
-        Delete the adjacency list.
-        """
-        cdef uintptr_t graph = self.graph_ptr
-        err = gdf_delete_adj_list(<gdf_graph*> graph)
-        cudf.bindings.cudf_cpp.check_gdf_error(err)
-
-    def delete_transposed_adj_list(self):
-        """
-        Delete the transposed adjacency list.
-        """
-        cdef uintptr_t graph = self.graph_ptr
-        err = gdf_delete_transposed_adj_list(<gdf_graph*> graph)
-        cudf.bindings.cudf_cpp.check_gdf_error(err)
 
     def num_vertices(self):
         """
--- a/python/cugraph/graph/c_graph.pyx
+++ b/python/cugraph/graph/c_graph.pyx
@@ -171,21 +171,6 @@ class Graph:
                                  c_value_col_ptr)
         cudf.bindings.cudf_cpp.check_gdf_error(err)
 
-    def num_vertices(self):
-        """
-        Get the number of vertices in the graph
-        """
-        cdef uintptr_t graph = self.graph_ptr
-        cdef gdf_graph * g = <gdf_graph*> graph
-        if g.adjList:
-            return g.adjList.offsets.size - 1
-        elif g.transposedAdjList:
-            return g.transposedAdjList.offsets.size - 1
-        else:
-            err = gdf_add_adj_list(g)
-            cudf.bindings.cudf_cpp.check_gdf_error(err)
-            return g.adjList.offsets.size - 1   
-
     def view_edge_list(self):
         """
         Display the edge list. Compute it if needed.
@@ -451,10 +436,15 @@ class Graph:
         Get the number of vertices in the graph
         """
         cdef uintptr_t graph = self.graph_ptr
-        cdef gdf_graph* g = <gdf_graph*> graph
-        err = gdf_add_adj_list(g)
-        cudf.bindings.cudf_cpp.check_gdf_error(err)
-        return g.adjList.offsets.size - 1
+        cdef gdf_graph * g = <gdf_graph*> graph
+        if g.adjList:
+            return g.adjList.offsets.size - 1
+        elif g.transposedAdjList:
+            return g.transposedAdjList.offsets.size - 1
+        else:
+            err = gdf_add_adj_list(g)
+            cudf.bindings.cudf_cpp.check_gdf_error(err)
+            return g.adjList.offsets.size - 1   
 
     def in_degree(self, vertex_subset = None):
         """
--- a/python/cugraph/graph/c_graph.pyx
+++ b/python/cugraph/graph/c_graph.pyx
@@ -356,7 +356,7 @@ class Graph:
         cudf.bindings.cudf_cpp.check_gdf_error(err)
 
         offset_col_size = self.num_vertices() + 1
-        inex_col_size = g.transposedAdjList.indices.size
+        index_col_size = g.transposedAdjList.indices.size
 
         cdef uintptr_t offset_col_data = <uintptr_t> g.transposedAdjList.offsets.data
         cdef uintptr_t index_col_data = <uintptr_t> g.transposedAdjList.indices.data
@@ -366,7 +366,7 @@ class Graph:
                                      dtype=np.int32)  # ,
                                      # finalizer=rmm._make_finalizer(offset_col_data, 0))
         indices_data = rmm.device_array_from_ptr(index_col_data,
-                                     nelem=inex_col_size,
+                                     nelem=index_col_size,
                                      dtype=np.int32)  # ,
                                      # finalizer=rmm._make_finalizer(index_col_data, 0))
         # g.transposedAdjList.offsets.data and g.transposedAdjList.indices.data
--- a/python/cugraph/graph/c_graph.pyx
+++ b/python/cugraph/graph/c_graph.pyx
@@ -318,7 +318,7 @@ class Graph:
         err = gdf_add_adj_list(g)
         cudf.bindings.cudf_cpp.check_gdf_error(err)
 
-        col_size_off = g.adjList.offsets.size
+        col_size_off = self.num_vertices() + 1
         col_size_ind = g.adjList.indices.size
 
         cdef uintptr_t offset_col_data = <uintptr_t> g.adjList.offsets.data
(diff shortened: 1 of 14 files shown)


python/cugraph/graph/c_graph.pyx
```diff
@@ -280,56 +269,52 @@ class Graph:
...
+        cdef gdf_column c_offset_col = get_gdf_column_view(self.adj_list_offset_col)
+        cdef gdf_column c_index_col = get_gdf_column_view(self.adj_list_index_col)
+        cdef gdf_column c_value_col
+        cdef gdf_column * c_value_col_ptr
         if value_col is None:
-            value = 0
+            c_value_col_ptr = NULL
         else:
-            value = create_column(self.adj_list_value_col)
+            c_value_col = get_gdf_column_view(self.adj_list_value_col)
```
jwyles: Is there a scoping issue here? You are creating an object on the stack inside of a code block and assigning a pointer to it which is used outside of the code block.

python/cugraph/graph/c_graph.pyx
```diff
@@ -280,56 +269,52 @@ class Graph:
...
         """
         cdef uintptr_t graph = self.graph_ptr
-        cdef gdf_graph * g = < gdf_graph *> graph
+        cdef gdf_graph * g = <gdf_graph*> graph
         err = gdf_add_adj_list(g)
         cudf.bindings.cudf_cpp.check_gdf_error(err)
 
-        col_size_off = g.adjList.offsets.size
-        col_size_ind = g.adjList.indices.size
+        offset_col_size = self.num_vertices() + 1
```
jwyles: I think it's better to just take the size of the offsets column directly rather than calling self.num_vertices() it's less complicated.

python/cugraph/graph/c_graph.pyx
```diff
@@ -358,38 +343,45 @@ class Graph:
...
         """
         cdef uintptr_t graph = self.graph_ptr
-        cdef gdf_graph * g = < gdf_graph *> graph
+        cdef gdf_graph * g = <gdf_graph*> graph
         err = gdf_add_transposed_adj_list(g)
         cudf.bindings.cudf_cpp.check_gdf_error(err)
 
-        off_size = g.transposedAdjList.offsets.size
-        ind_size = g.transposedAdjList.indices.size
+        offset_col_size = self.num_vertices() + 1
```
jwyles: Same here, getting the size of the column directly is simpler.

python/cugraph/graph/c_graph.pyx
```diff
@@ -358,38 +343,45 @@ class Graph:
...
+        inex_col_size = g.transposedAdjList.indices.size
```
jwyles: I think you may have meant 'index_col_size' here.

python/cugraph/graph/c_graph.pyx
seunghwak:
```python
283         cdef gdf_column c_value_col
```

c_value_col is defined in line 238, and it's not used after

```python
291         err = gdf_adj_list_view(<gdf_graph*> graph,
292                                 &c_offset_col,
293                                 &c_index_col,
294                                 c_value_col_ptr)
```

Please note that the actual column array data are stored in self.adj_list_value_col (a python object), and c_value_col is a temporary C++ wrapper used to pass the column data in python object to a C++ function expecting C++ data type.

python/cugraph/graph/c_graph.pyx
seunghwak: I may not agree with this. We'd better hide C++ data structure (gdf_graph) internals from the functions using gdf_graph. If we use num_verticies(), it does not matter whether we change gdf_graph internals or not as long as we have num_verticies(). If some change is necessary, we just need to update num_verticies(). We are about to update graph data structures to match NetworkX, and in this case, we need to search the entire code base and replace every g.adjList.offsets.size -1 (and g.transposedAdjList.offset.size if g.adjList is not available and g.transposedAdjList is available) to match the new data structure. This will become more and more problematic as we add more and more graph functions; there basically are more places we need to track.

See Hiding data (and code) in https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/

Hiding data structure internals is a key principle in object oriented programming.

python/cugraph/graph/c_graph.pyx
seunghwak: I disagree for the same reason.

python/cugraph/graph/c_graph.pyx
seunghwak: Great catch!!! I made a fix.

python/cugraph/graph/c_graph.pyx
seunghwak: And cudf folks are working on replacing gdf_column with cudf::column to provide better abstraction (https://github.com/rapidsai/cudf/issues/1443).

We need to do something similar for gdf_graph; and I think eventually most (if not all) gdf_graph member variables should become private and should not be directly accessed outside gdf_graph.

python/cugraph/jaccard/wjaccard_wrapper.pyx
```diff
@@ -74,29 +74,29 @@ cpdef jaccard_w(input_graph, weights, first=None, second=None):
...
-    cdef uintptr_t result_ptr
-    cdef uintptr_t weight_ptr
-    cdef uintptr_t first_ptr
-    cdef uintptr_t second_ptr
-    cdef uintptr_t indices_ptr
+    cdef gdf_column c_result_col
+    cdef gdf_column c_weight_col
+    cdef gdf_column c_first_col
+    cdef gdf_column c_second_col
+    cdef gdf_column c_indices_col
```
Iroy30: Should this be c_index_col

python/cugraph/graph/c_graph.pyx
jwyles: I still disagree. While you are right that it is good to hide the internals of the graph object, the issue is that we aren't looking for the number of vertices in the graph, but for the size of the offsets array. Using `self.num_vertices() + 1` to get that is still relying on information on the graph object's internals, just doing it in a more roundabout way.

seunghwak: Thank you for your reply!!!

See "The array IA is of length m + 1." from https://en.wikipedia.org/wiki/Sparse_matrix (here, m is the number of vertices).

self.num_vertices() + 1 (or m + 1 if I use notation from the above link) is from the general knowledge about the CSR format, not from the graph structure internals. Hope this helps to address your concern.

python/cugraph/jaccard/wjaccard_wrapper.pyx
seunghwak: Thanks for finding this!!! and I fixed this.

python/cugraph/graph/c_graph.pyx
afender: While I don't think this should block the PR, I'm mostly concerned about consistency here. It is now using a mix of direct structure access and accessors. If `self.num_vertices()+1` is used to set the size of the offset array, it is best to use a `self.num_edges()` for indices arrays too or keep the previous, consistent, solution imo.

seunghwak: Yes, I have the same concern. The thing is that we do not have the num_edges() function, and we are planning to do some major restructuring to match the NetworkX API (for the 0.8 release I guess), so I was a bit hesitant to implement the num_edges() function that will be replaced in one month. I decided to defer this work for the 0.8 release but if this inconsistency really bothers (it somewhat bothers me but I can endure if this will last for only short time), I can do the work for this pull request.

python/cugraph/graph/c_graph.pyx
afender: It seems to me that `c_value_col` is still in the scope when used in `gdf_adj_list_view`.
However, there is also a small consistency issue here as `c_offset_col` and `c_index_col` are accessed from the stack objects while a pointer to the stack object is created for the values.

python/cugraph/graph/c_graph.pyx
seunghwak: Actually, I think I'd better do this now.

I will rename num_vertices() to number_of_nodes() and num_edges() to number_of_edges() to follow NetworkX (https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.DiGraph.number_of_nodes.html and https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.Graph.number_of_edges.html).

This will not be too much work and not gonna be entirely replaced.

afender: I can also live with that as long as these changes are tracked in 0.8 release ;)

python/cugraph/graph/c_graph.pyx
seunghwak: The inconsistency is because value is optional (it can be NULL). I can either use value_col_ptr or replicate the gdf_edge_list_view call for the two different cases (if value_col is None and else). I picked to use value_col_ptr.

Issue #219, opened 2019-04-12
[FEA] CSV testing


afender:
Current python tests are based on Matrix Market files and SciPy reader. We want to change that for CSV files and cuDF CSV reader. 

This will better reflect cugraph's e2e use cases. It will also be faster.

Tasks:
- [x] convert mtx files in `datasets/` with the python script in `python/utils`
- [x] update all tests to use cuDF's csv reader

afender:
Most of this was done in #233.

It would be great to switch to csv input for Nx in 0.8 as well. Testing stability should be improved as discussed here https://github.com/rapidsai/cugraph/pull/233#issuecomment-486445796

PR #233, opened 2019-04-19
[WIP] edit tests for reading csv

Iroy30:


afender:
@Iroy30 Merging #232 PR created a conflict in this PR. Can you please take a look at it?

afender:
Other than that the code looks good. This is the PR for #219.
Question: do we really need to keep `.mtx` files/tests?

afender:
Just talked to Ishika, mtx files are needed for Scipy reader for Nx, we will use Panda's CSV reader and remove MTX files + Scipy reader.

afender:
@Iroy30 is it technically possible to remove the mtx files in the end or should I merge this PR?

Iroy30:
Removing mtx would have the following issues:
- Reading from csv into graph doesn't have information about nodes not present in the csv. This leads to column mismatches( for example, col length of pagerank and their comparisions)
- cugraph provides result corresponding to sorted edgelist (for example in jaccard_coeff) while networkx provides the results in the order of edges (which could be unsorted). This calls for either sorting edgelist initially before input as edges to networkx jaccard api or creating a dictionary of edgelist vs nx coefficients in networkx_call and reference that.
- Some tests check correctness of adding adjacency list or working on adjacency list input. For this, we would need a csr format which was easier in scipy using M.tocsr(). pandas/cudf dataframes cannot store mismatched column lengths, therefore, reading the csv and then having a coo to csr conversion is additionally required.

afender:
It seems to me that :
(1) will work when we will have renumbering. cuGraph should do the same thing as Nx after that.
(2) both solutions you proposed improve the current testing solution which is currently making big assumptions on sorted input.
(3) we cannot get rid of this one.

So we should get (1) and (2) at some point but we will still have to keep mtx files for (3). It is ok to keep mtx files as NetworkX input for now.

PR #234, opened 2019-04-19
[REVIEW] Fea renumbering phase2

ChuckHastings:
Additions to the renumbering logic including:
- Some refactoring based on ideas from the segmented sort logic
- Use RMM for allocating device memory
- Add python interface for renumbering
- Add python unit test for renumbering


seunghwak:
So, multiple checks are failing (including missing change log & flake8 style test), so I think these need to be fixed first (before going to more thorough code review).

ChuckHastings:
I have fixed the CHANGELOG and python formatting issue (that's a new check since the last time I checked in python code).

I can't make progress on the gpu-test failing issue.  I have done a clean checkout/build and it works properly on my system.  The test I added shouldn't require any special data, it has some hard-coded data in the test script.

seunghwak:
> I have fixed the CHANGELOG and python formatting issue (that's a new check since the last time I checked in python code).
> 
> I can't make progress on the gpu-test failing issue. I have done a clean checkout/build and it works properly on my system. The test I added shouldn't require any special data, it has some hard-coded data in the test script.

Isn't this something we can ask help to the CI/CD people?

afender:
Seunghwa is correct, tests should pass before we can review and merge it. 
@ChuckHastings please work with CI to identify and fix the problem on Jenkins.

ChuckHastings:
rerun tests

cpp/include/functions.h
```diff
@@ -15,6 +15,31 @@
  */
 #pragma once 
 
+/* ----------------------------------------------------------------------------*/
+
+/**
+ * @Synopsis Renumber source and destination indexes to be a dense numbering.
+ *
```
afender: Not sure if "dense renumbering" is standard. I'd recommend adding "The resulting renumbered output contains GDF_INT32 or GDF_INT64 indexes between 0 and the number of vertices minus 1".

cpp/src/cugraph.cu
```diff
@@ -113,6 +115,109 @@ gdf_error gdf_adj_list::get_source_indices (gdf_column *src_indices) {
...
+    //  NOTE: At the moment, we force the renumbered graph to use
+    //        32-bit integer ids.  Since renumbering is going to make
+    //        the vertex range dense, this limits us to 2 billion
+    //        vertices.
+    //
+    //        The renumbering code supports 64-bit integer generation
+    //        so we can run this with int64_t output if desired...
+    //        but none of the algorithms support that.
+    //
+    printf("in renumber, 64-bit, src->dtype = %d\n", src->dtype);
```
afender: A couple of `printf` was added to this file

cpp/src/heap.cuh
```diff
@@ -0,0 +1,225 @@
...
+      
+    /**
+     * @brief Reorder an existing array of elements into a heap
+     *
+     * @param[in, out]   array   - the existing array
+     * @param[in]        size    - the number of elements in the existing array
+     * @param[in]        compare - the comparison function to use
+     *
+     */
+    template <typename T, typename S, typename C>
```
afender: What are `typename T, typename S, typename C`? It would be great to document or rename with more explicit names.

cpp/include/functions.h
```diff
@@ -15,6 +15,31 @@
...
+ *    Assumptions:
+ *       * source and dest have same size and type
+ *       * source and dest are either GDF_INT32 or GDF_INT64
```
afender: Is it possible to provide GDF_INT64 as input and get GDF_INT32 on output if the dataset is small enough with this API? The cuda code seems to support it. If so, we should document it in the API.

cpp/src/tests/renumber/renumber_test.cu
```diff
@@ -260,24 +266,39 @@ TEST_F(RenumberingTest, Random100KVertexSet)
...
+  size_t cnt = 0;
   for (size_t i = 0 ; i < num_verts ; ++i) {
     min_id = min(min_id, tmp_results[i]);
     max_id = max(max_id, tmp_results[i]);
-    EXPECT_EQ(tmp_map[tmp_results[i]], src_data[i]);
+    if (tmp_map[tmp_results[i]] != src_data[i])
+      ++cnt;
+
+    if (cnt < 20)
+      EXPECT_EQ(tmp_map[tmp_results[i]], src_data[i]);
```
afender: I suspect this `if` was added for debug. It seems to me that when cnt is greater than 0 the data is corrupted, correct?

cpp/src/tests/renumber/renumber_test.cu
```diff
@@ -260,24 +266,39 @@ TEST_F(RenumberingTest, Random100KVertexSet)
...
   EXPECT_EQ(cudaMemcpy(tmp_results, dst_d, sizeof(uint64_t) * num_verts, cudaMemcpyDeviceToHost), cudaSuccess);
   for (size_t i = 0 ; i < num_verts ; ++i) {
     min_id = min(min_id, tmp_results[i]);
     max_id = max(max_id, tmp_results[i]);
-    EXPECT_EQ(tmp_map[tmp_results[i]], dst_data[i]);
+    if (tmp_map[tmp_results[i]] != dst_data[i])
+      ++cnt;
+
+    if (cnt < 20)
+      EXPECT_EQ(tmp_map[tmp_results[i]], dst_data[i]);
```
afender: Same here, I suspect this if was added for debug. It seems to me that when cnt is greater than 0 the data is corrupted, correct?

cpp/src/tests/renumber/renumber_test.cu
afender: I mean should we have a `EXPECT_EQ(0, cnt);`?

afender: Nevermind I think I get the logic it should capture the failure properly

cpp/src/renumber.cuh
```diff
@@ -193,7 +199,8 @@ namespace cugraph {
 			      T_in ** numbering_map,
 			      int max_threads_per_block = CUDA_MAX_KERNEL_THREADS,
 			      int max_blocks = CUDA_MAX_BLOCKS,
-			      uint32_t hash_size = 8191) {
+			      detail::hash_type hash_size = 8191) {
```
afender: Is this default size working for the 100k vertices test on GV100 only or do you expect it to be pretty good for all sizes/cards?

cpp/include/functions.h
ChuckHastings: I left the word dense in, but added your words as well.  Actually put the GDF_INT32 comment in the definition of the output rather than in the synopsis.

python/cugraph/graph/test_graph.py
```diff
@@ -268,3 +271,37 @@ def test_degree_functionality(graph_file):
     assert err_in_degree == 0
     assert err_out_degree == 0
     assert err_degree == 0
+
+
+def test_renumber():
+    source_list = ['192.168.1.1',
```
afender: That's a pretty cool test. 

I would just recommend adding another python test with more than 1025 vertices. The reason is that smaller tests will run in only one block in the GPU. Many problems can remain silent if we test just that. I noticed the 100k test on the C++ side but only python tests are being run in Jenkins currently. We use `netscience` dataset at the python level to cover this. I was thinking you could apply a simple transformation to it, call the renumbering, and check that it corresponds to the original dataset.

cpp/include/functions.h
ChuckHastings: The current output will ALWAYS be GDF_INT32 in the python API, since only GDF_INT32 works in most of the algorithms.  Check the new text (including the new return value in case the renumbering doesn't fit in a 32-bit integer.

cpp/src/cugraph.cu
ChuckHastings: Cleaned up in next push.

afender: This should be disabled in the final phase (it is ok to keep them in renumbering for now if you plan to do another PR later)

cpp/src/heap.cuh
ChuckHastings: Suggestions on how to document this?  I don't see anything in the doxygen documentation about how to properly do that.  I guess I could rename them to ArrayT, SizeT and Compare if that seems more clear to you.

afender:
Overall the code is in good shape but I recommend a few changes that I think should be done before the release. I was not sure if that was the final PR or not and did the review assuming it was. In case this is just a checkpoint, I am completely fine with approving this PR if the review is addressed in another one.

cpp/src/renumber.cuh
ChuckHastings: I don't know.  I expect we'll want to do some more extensive experimentation.  In my testing thus far it seems to be a good value.  But I expect that using strings will change the dynamic a bit, and I'm hoping that I can optimize the segmented sort a bit more and use thet here - which may also change the optimal hash size.

cpp/src/heap.cuh
afender: I think just a regular comment would work or the renaming you suggested. It is just to help with readability for other developers if it needs to be modified or re-used elsewhere in the future.

cpp/src/renumber.cuh
afender: We should probably create an issue to keep this on the radar. It would be great to have some heuristics to set it automatically based on the input. Renumbering is not cheap and this can impact perf by over 2x.

cpp/src/cugraph.cu
```diff
@@ -113,6 +115,105 @@ gdf_error gdf_adj_list::get_source_indices (gdf_column *src_indices) {
...
+				gdf_column **numbering_map) {
+
+  GDF_REQUIRE( src->size == dst->size, GDF_COLUMN_SIZE_MISMATCH );
+  GDF_REQUIRE( src->dtype == dst->dtype, GDF_UNSUPPORTED_DTYPE );
+  GDF_REQUIRE( ((src->dtype == GDF_INT32) || (src->dtype == GDF_INT64)), GDF_UNSUPPORTED_DTYPE );
+  GDF_REQUIRE( src->size > 0, GDF_DATASET_EMPTY ); 
+
+  *src_renumbered = new gdf_column;
+  *dst_renumbered = new gdf_column;
+  *numbering_map = new gdf_column;
```
seunghwak: So do we really need these memory allocations here? If we ignore the fact that this function is called by cython and also ignore that we are returning error code (as I know we will switch to exception based error handling), I think safer implementation is to do

```cpp
gdf_column src_renumbered;
gdf_column dst_renumbered;
gdf_column numbering_map;
...
return std::make_tuple(src_renumbered, dst_renumbered, numbering_map);
```

Now, we have our limitation and also this function needs to maintain consistency with other functions in this file. Then, I think it's better to do in cython

```
cdef gdf_column src_renumbered
cdef gdf_column dst_renumbered;
cdef gdf_column numbering_map;
...
err = gdf_renumber_vertices(..., &src_renumbered, &dst_renumbered, &numbering_map)
```

and modify this function's signature to take gdf_column* (instead of gdf_column**) for src_renumbered, dst_renumbered, and numbering_map. Then, we don't need to worry about explicitly freeing these memory allocations (and also avoid heap allocation overhead). I will comment in the cython file again, but the cython caller did not indeed free these allocations; and this leads to memory leak.

python/cugraph/graph/c_graph.pyx
```diff
@@ -94,6 +113,41 @@ class Graph:
...
+        src_renumbered_array = rmm.device_array_from_ptr(<uintptr_t> src_renumbered.data,
+                                     nelem=src_renumbered.size,
+                                     dtype=dtypes_inv[src_renumbered.dtype])
+        dst_renumbered_array = rmm.device_array_from_ptr(<uintptr_t> dst_renumbered.data,
+                                     nelem=dst_renumbered.size,
+                                     dtype=dtypes_inv[dst_renumbered.dtype])
+        numbering_map_array = rmm.device_array_from_ptr(<uintptr_t> numbering_map.data,
+                                     nelem=numbering_map.size,
+                                     dtype=dtypes_inv[numbering_map.dtype])
+
```
seunghwak: So this function is responsible for freeing src_renumbered, dst_renumbered, and numbering_map allocated by gdf_renumber_vertices, but fail to do that. As I mentioned in the implementation of gdf_renumber_vertices, a better way is to avoid memory allocations.

Don't get confused between allocating memory for the gdf_column structure and allocating actual column array. The gdf_column structure has fixed size and is not really that large, so no need to worry about blowing out stack. Also, src_renumbered.data, dst_renumbered.data, and numbering_map.data should not be freed in this function but src_renumbered, dst_renumbered, and numbering_map are used only temporarily inside this function.

cpp/include/functions.h
seunghwak: What about using "contiguous" instead of "dense"? If I am not mistaken, renumbering makes vertex ids to be contiguous, and I guess this is more specific than dense.

cpp/src/cugraph.cu
ChuckHastings: OK.  I think I started this work before some of the cleanup you did for memory leaks.  Probably copied this from somewhere else.  My next push should address this.

cpp/src/cugraph.cu
```diff
@@ -375,23 +478,27 @@ gdf_error gdf_add_adj_list(gdf_graph *graph) {
...
 
   GDF_REQUIRE( graph->edgeList != nullptr , GDF_INVALID_API_CALL);
-  GDF_REQUIRE( graph->adjList == nullptr , GDF_INVALID_API_CALL);
+  GDF_REQUIRE( graph->edgeList->src_indices->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE );
 
   if (graph->edgeList->edge_data != nullptr) {
     switch (graph->edgeList->edge_data->dtype) {
-      case GDF_FLOAT32:   return gdf_add_adj_list_impl<float>(graph);
-      case GDF_FLOAT64:   return gdf_add_adj_list_impl<double>(graph);
+      case GDF_FLOAT32:   return gdf_add_adj_list_impl<int, float>(graph);
```
seunghwak: So we're mixing int32_t and int in this code, so, if "int" here intends "int32_t", shouldn't we better use int32_t here? (and many other places in this file, if the intention is really to use 32 bit integer (not just some integral number that will not become very big)).

python/cugraph/graph/c_graph.pyx
ChuckHastings: Changed gdf_column to be local, so that should (I think) take care of that.  Should be in next push.

cpp/src/cugraph.cu
```diff
@@ -343,7 +445,8 @@ gdf_error gdf_pagerank_impl (gdf_graph *graph,
   ALLOC_MANAGED_TRY((void**)&d_val, sizeof(WT) * nnz , stream);
   ALLOC_MANAGED_TRY((void**)&d_pr,    sizeof(WT) * m, stream);
 
-  cugraph::HT_matrix_csc_coo(m, nnz, (int*)graph->transposedAdjList->offsets->data, (int*)graph->transposedAdjList->indices->data, d_val, d_leaf_vector);
+  //  The templating for HT_matrix_csc_coo assumes that m, nnz and data are all the same type
+  cugraph::HT_matrix_csc_coo(m, nnz, (int *)graph->transposedAdjList->offsets->data, (int *)graph->transposedAdjList->indices->data, d_val, d_leaf_vector);
```
seunghwak: If I am not mistaken, typename T is added to specify the type for graph->tranposedAdjList->offsets->data, so should we cast to (T*)?

cpp/src/heap.cuh
```diff
@@ -0,0 +1,225 @@
...
+	for (S i = size - 1 ; i > 0 ; --i) {
+	  S p = detail::parent(i);
+
+	  {
+	    S lc = detail::left_child(p);
+	    S rc = detail::right_child(p);
+	    if (i == lc) {
+	    } else if (i == rc) {
+	    } else {
+	      printf("in heapify... i = %ld, p = %ld, lc = %ld, rc = %ld\n", i, p, lc, rc);
```
seunghwak: Are we going to keep this printf? I guess this is for debugging and cuGraph users don't need this information.

cpp/src/heap.cuh
```diff
@@ -0,0 +1,225 @@
...
+    template <typename T, typename S, typename C>
+    inline void __host__ __device__ check_heap(T *array, S size, C compare) {
+      static_assert(std::is_integral<S>::value, "Index must be of an integral type");
+
+      for (S i = 0 ; i < size ; ++i) {
+	S lc = detail::left_child(i);
+	S rc = detail::right_child(i);
+
+	if ((rc < size) && (compare(array[rc], array[i]))) {
+	  printf("error... array[%ld] = %ld, array[%ld] = %ld, size = %ld\n", rc, array[rc], i, array[i], size);
```
seunghwak: Again, are we going to keep this printf?

cpp/src/heap.cuh
```diff
@@ -0,0 +1,225 @@
...
+	} else if ((lc < size) && (compare(array[lc], array[i]))) {
+	  printf("error... array[%ld] = %ld, array[%ld] = %ld, size = %ld\n", lc, array[lc], i, array[i], size);
```
seunghwak: printf here as well.

cpp/src/cugraph.cu
ChuckHastings: Only spacing difference in my change.  I had started down the path of adding typename T to functions when I discovered that most of the algorithms only work on int32_t anyway.  This particular function does not have a typename T associated with it.

cpp/src/heap.cuh
ChuckHastings: They were for debugging... forgot they were still there.  I'll clean them up.

cpp/src/cugraph.cu
seunghwak: And c type casting is dangerous and generally not advised in C++. static_cast<T*>(ptr_to_cast) is much safer in general.

cpp/src/heap.cuh
ChuckHastings: This is a debugging function.  I'm deleting the whole function.

cpp/src/cugraph.cu
ChuckHastings: As I didn't actually change this code... should I be fixing those things that were already wrong, or should we create a different issue for this?

seunghwak: It's up to you, I am not going to block this PR for this, and I guess this isn't the only place for this, it will be a huge amount of work to remove all C type casting in cugraph. What I recommend is to remove C type casting one by one when you see those.

cpp/src/heap.cuh
ChuckHastings: Just pushed this change.

cpp/src/cugraph.cu
ChuckHastings: I'd suggest that we make this a priority going forward.  I'm happy to try stomping them out as the opportunity arises.  Not really comfortable making those changes at the end of a PR.

cpp/src/renumber.cuh
afender: #243

PR #239, opened 2019-04-23
[REVIEW] Bug ext two graphs

### LINKED ISSUE #187: [BUG] If both add_edge_list and add_adj_list of the class Graph (c_graph.pyx) are called, a single graph instance can store two different graphs.
**Describe the bug**
add_edge_list of class Graph (in c_graph.pyx) calls gdf_edge_list_view and add_adj_list calls gdf_adj_list_view.

gdf_edge_list_view does not check there already exists an adjacency list representation and gdf_adj_list_view does not check there already exists an edge list representation. If both are called representing two different graphs, it is unclear what num_vertices of class Graph should return (current it returns the number of vertices in the adjacency list).

**Expected behavior**
There are two options. If the graph instance already holds a graph, 1) we may allow only conversion between formats. 2) Alternatively, we may delete an existing graph (including graph representations in different formats) before adding a new one.

seunghwak:
Close #187

### DIFF
--- a/cpp/src/cugraph.cu
+++ b/cpp/src/cugraph.cu
@@ -70,12 +70,16 @@ void cpy_column_view(const gdf_column *in, gdf_column *out) {
 
 gdf_error gdf_adj_list_view(gdf_graph *graph, const gdf_column *offsets,
                                  const gdf_column *indices, const gdf_column *edge_data) {
+  //This function returns an error if this graph object has at least one graph
+  //representation to prevent a single object storing two different graphs.
+  GDF_REQUIRE( ((graph->edgeList == nullptr) && (graph->adjList == nullptr) &&
+    (graph->transposedAdjList == nullptr)), GDF_INVALID_API_CALL);
+  GDF_REQUIRE( (graph->transposedAdjList == nullptr) , GDF_INVALID_API_CALL);
   GDF_REQUIRE( offsets->null_count == 0 , GDF_VALIDITY_UNSUPPORTED );
   GDF_REQUIRE( indices->null_count == 0 , GDF_VALIDITY_UNSUPPORTED );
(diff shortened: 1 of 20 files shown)


cpp/src/cugraph.cu
```diff
@@ -70,12 +70,16 @@ void cpy_column_view(const gdf_column *in, gdf_column *out) {
 
 gdf_error gdf_adj_list_view(gdf_graph *graph, const gdf_column *offsets,
                                  const gdf_column *indices, const gdf_column *edge_data) {
+  //This function returns an error if this graph object has at least one graph
+  //representation to prevent a single object storing two different graphs.
+  GDF_REQUIRE( ((graph->edgeList == nullptr) && (graph->adjList == nullptr) &&
+    (graph->transposedAdjList == nullptr)), GDF_INVALID_API_CALL);
+  GDF_REQUIRE( (graph->transposedAdjList == nullptr) , GDF_INVALID_API_CALL);
```
afender: redundant

python/cugraph/graph/c_graph.pyx
```diff
@@ -432,27 +445,51 @@ class Graph:
             return g.adjList.offsets.size - 1
         elif g.transposedAdjList:
             return g.transposedAdjList.offsets.size - 1
-        else:
+        elif g.edgeList:
+            # This code needs to be revisited when updating gdf_graph. Users
+            # may expect numbrer_of_vertcies() as a cheap query but this
```
afender: Agreed, I created #241 for this. I don't expect many users to upload an edge list and directly ask for the number of vertices before actually running any analytics. So it would be nice to have but I don't see it as a top priority.

cpp/src/cugraph.cu
seunghwak: Thanks for finding this and I made a fix!

PR #246, opened 2019-04-25
[REVIEW] Subgraph extraction python bindings

### LINKED ISSUE #60: [FEA] [NV] Fix and Add Python bindings for Subgraph Extraction
Following Alex's  CUDA/C++ method for accessing nvgraph features using cugraph graph structure, wrapper Subgraph Extraction  [src/plugin/nvgraph/nvgraph_gdf.cu]. 

Subtasks checklist:
- [ ] Pythion bindings and API
- [ ] Pytest testing against NetworkX
- [ ] Documentation
- [ ] licenses info
- [ ] Validate

jwyles:
Adding in the python bindings for subgraph extraction after the fix in nvgraph was put in. Closes #60


ChuckHastings:
Looks good if you can make make the checks pass.

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

python/cugraph/subgraph_extraction/subgraph_extraction.pyx
```diff
@@ -0,0 +1,68 @@
...
+from libc.stdint cimport uintptr_t
+from libc.stdlib cimport calloc, malloc, free
+from libc.float cimport FLT_MAX_EXP
+import cudf
+from librmm_cffi import librmm as rmm
+import numpy as np
+
+cpdef subgraph(G, vertices):
+    """
+    Compute a clustering/partitioning of the given graph using the spectral balanced
```
afender: needs to be updated to reflect subgraph extraction

python/cugraph/subgraph_extraction/subgraph_extraction.pyx
```diff
@@ -0,0 +1,68 @@
...
+    cut method.
+    
+    Parameters
+    ----------
+    G : cuGraph.Graph                  
+       cuGraph graph descriptor
```
afender: does it work both on directed and undirected graphs? We may re-use nvgraph doc
https://docs.nvidia.com/cuda/nvgraph/index.html#function-nvgraphextractsubgraphbyvertex

ChuckHastings: Yes, it works for both directed and undirected graphs.  It doesn't actually traverse the edges, so direction is immaterial.

cpp/include/nvgraph_gdf.h
```diff
@@ -132,6 +132,16 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
                                                    float* score);
 
+/**
+ * Wrapper function for Nvgraph extract subgraph by vertices
+ * @param gdf_G Pointer to GDF graph object, this is the input graph
+ * @param result Pointer to GDF graph object, this is the output must be a valid pointer
+ * @param vertices Pointer to GDF column object which contains the list of vertices to extract
+ * @return Error code
+ */
+gdf_error gdf_extract_subgraph_vertex_nvgraph(gdf_graph* gdf_G,
```
seunghwak: So as I long as I understand based on comments,

gdf_graph* gdf_g and gdf_column* vertices are input parameters and gdf_graph* result is an output parameter. A function input parameter list having (input, output, input) is unorthodox. It seems like other functions in this file places input parameters before output parameters in the function input parameter lists, we should follow that convention. Also, if this function is not modifying gdf_G and vertices, we should enforce "const correctness".

So the input parameter should be
(const gdf_graph* gdf_G, const gdf_column* vertices, gdf_graph* result).

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
 	return GDF_SUCCESS;
 }
 
+
+gdf_error gdf_extract_subgraph_vertex_nvgraph(gdf_graph* gdf_G,
+																							gdf_graph* result,
```
seunghwak: Indentation is weird, here. The input parameter list should be updated to match the suggested changes in the header file.

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+	NVG_TRY(nvgraphCreate(&nvg_handle));
+	GDF_TRY(gdf_createGraph_nvgraph(nvg_handle, gdf_G, &nvgraph_G, false));
+
+	// Create an Nvgraph graph descriptor for the result and initialize
+	nvgraphGraphDescr_t nv_result = nullptr;
+	NVG_TRY(nvgraphCreateGraphDescr(nvg_handle, &nv_result));
+
+	// Call Nvgraph function to get subgraph (into nv_result descriptor)
+	NVG_TRY(nvgraphExtractSubgraphByVertex(	nvg_handle,
+																					nvgraph_G,
```
seunghwak: Weird indentation.

seunghwak: weird indentation

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+																					vertices->size));
+
+	// Get the vertices and edges of the created subgraph to allocate memory:
+	nvgraphCSRTopology32I_st topo;
+	topo.source_offsets = nullptr;
+	topo.destination_indices = nullptr;
+	nvgraphTopologyType_t TT = NVGRAPH_CSR_32;
+	NVG_TRY(nvgraphGetGraphStructure(nvg_handle, nv_result, (void*)&topo, &TT));
+	if (TT != NVGRAPH_CSR_32)
+		return GDF_C_ERROR;
```
seunghwak: And this is not a request to make change, but it seems like you're using tab for indentation. cugraph mostly uses 2 space or 4 space for indentation in C++. We really need a rule for this, but right at this moment, I am not sure I should ask you to use either 2 space or 4 space... so you may just keep using tab or switch to 2 space or 4 space :-(

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+	int num_verts = topo.nvertices;
+	int num_edges = topo.nedges;
+	result->adjList = new gdf_adj_list;
+	result->adjList->offsets = new gdf_column;
+	result->adjList->indices = new gdf_column;
+	result->adjList->ownership = 0;
+	int *offsets, *indices;
+	CUDA_TRY(cudaMallocManaged((void**)&offsets, sizeof(int) * (num_verts + 1)));
```
seunghwak: We should not directly invoke cudaMalloc/cudaMallocManaged. We are allocating memory through RMM, so we need to use RMM_ALLOC instead.

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+	nvgraphGraphDescr_t nvgraph_G = nullptr;
+	NVG_TRY(nvgraphCreate(&nvg_handle));
+	GDF_TRY(gdf_createGraph_nvgraph(nvg_handle, gdf_G, &nvgraph_G, false));
+
+	// Create an Nvgraph graph descriptor for the result and initialize
+	nvgraphGraphDescr_t nv_result = nullptr;
+	NVG_TRY(nvgraphCreateGraphDescr(nvg_handle, &nv_result));
+
+	// Call Nvgraph function to get subgraph (into nv_result descriptor)
+	NVG_TRY(nvgraphExtractSubgraphByVertex(	nvg_handle,
```
seunghwak: Sorry to complain about this trivial thing... but in function calls, cugraph code base (mostly) does not place space after opening parenthesis. So better be "(nvg_handle" instead of "( nvg_handle"

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+																							gdf_graph* result,
+																							gdf_column* vertices) {
+	GDF_REQUIRE(gdf_G != nullptr, GDF_INVALID_API_CALL);
+	GDF_REQUIRE((gdf_G->adjList != nullptr) || (gdf_G->edgeList != nullptr), GDF_INVALID_API_CALL);
+	GDF_REQUIRE(vertices != nullptr, GDF_INVALID_API_CALL);
+	GDF_REQUIRE(vertices->data != nullptr, GDF_INVALID_API_CALL);
+	GDF_REQUIRE(!vertices->valid, GDF_VALIDITY_UNSUPPORTED);
+
+	// Initialize Nvgraph and wrap the graph
+	nvgraphHandle_t nvg_handle = nullptr;
```
seunghwak: and is there really a reason we are mixing nv_, nvg_, and nvgraph_ in variable names (e.g. nv_result, nvg_handle, and nvgraph_G)? I am guessing all three are for nvgraph data type variables.

cpp/src/nvgraph_gdf.cu
```diff
@@ -515,6 +515,69 @@ gdf_error gdf_AnalyzeClustering_ratio_cut_nvgraph(gdf_graph* gdf_G,
...
+	nvgraphTopologyType_t TT = NVGRAPH_CSR_32;
+	NVG_TRY(nvgraphGetGraphStructure(nvg_handle, nv_result, (void*)&topo, &TT));
+	if (TT != NVGRAPH_CSR_32)
+		return GDF_C_ERROR;
+	int num_verts = topo.nvertices;
+	int num_edges = topo.nedges;
+	result->adjList = new gdf_adj_list;
+	result->adjList->offsets = new gdf_column;
+	result->adjList->indices = new gdf_column;
+	result->adjList->ownership = 0;
```
seunghwak: If ownership is set to 0, who will free the memory allocated in line 560 and 561?

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cugraph
+import cudf
+import pytest
+import numpy as np
+import networkx as nx
+from scipy.io import mmread
+
```
seunghwak: Unless you have a strong reason to import in this order, we may better follow import orders in other files.

Import python standard modules first, then thirdparty modules, then rapids modules.

We may alphabetically sort within a group. And if NetworkX folks hasn't fixed this, importing networkx generates deprecation warnings and to avoid this, we need to do

# Temporarily suppress warnings till networkX fixes deprecation warnings
# (Using or importing the ABCs from 'collections' instead of from
# 'collections.abc' is deprecated, and in 3.8 it will stop working) for
# python 3.7.  Also, this import networkx needs to be relocated in the
# third-party group once this gets fixed.
import warnings
with warnings.catch_warnings():
    warnings.filterwarnings("ignore", category=DeprecationWarning)
    import networkx as nx

See other test files (e.g. test_sssp.py) for an example.

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+
+def compareEdges(cg, nxg, verts):
+    src, dest = cg.view_edge_list()
+    if (len(src) != nxg.size()):
```
seunghwak: why not
```
assert len(src) == nxg.size()
```
?

Similarly in line 27 & 28,

```
assert nxg.has_edge(verts[src[i]], verts[dest[i]]) == True
```
Looks clearer to me.

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+    cols = cudf.Series(M.col)
+    G.add_edge_list(rows, cols, None)
+    cu_verts = cudf.Series(verts)
+    Sg = cugraph.subgraph(G, cu_verts)
+    return Sg
+
+
+def nx_call(M, verts):
+    G = nx.DiGraph(M)
+    Sg = nx.subgraph(G, verts)
```
seunghwak: See the following for PEP8 naming rules.

https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

Sg and cu_Sg violates

"Function names should be lowercase, with words separated by underscores as necessary to improve readability.

Variable names follow the same convention as function names."

unless "mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility." and I don't think mixed case is already the prevailing style in cugraph.

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+datasets = ['/datasets/networks/karate.mtx',
+            '/datasets/networks/dolphins.mtx',
+            '/datasets/networks/netscience.mtx']
+
+
+@pytest.mark.parametrize('graph_file', datasets)
+def test_subgraph_extraction(graph_file):
+    M = ReadMtxFile(graph_file)
```
seunghwak: ReadMtxFile also violates the PEP8 naming rule.

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
```diff
@@ -0,0 +1,67 @@
...
+    verts = np.zeros(3, dtype=np.int32)
+    verts[0] = 0
+    verts[1] = 1
+    verts[2] = 17
+    cu_Sg = cugraph_call(M, verts)
+    nx_Sg = nx_call(M, verts)
+    assert compareEdges(cu_Sg, nx_Sg, verts)
```
seunghwak: compareEdges also violate the PEP8 naming rule.

cpp/src/nvgraph_gdf.cu
ChuckHastings: Converted tabs to spaces in the file.  Should address the weird indentation.

cpp/src/nvgraph_gdf.cu
ChuckHastings: Output of the analytic is a graph.  This space is owned by the graph, the user will need to delete the graph when they are done with it.

cpp/src/nvgraph_gdf.cu
ChuckHastings: Changed to use RMM

cpp/include/nvgraph_gdf.h
ChuckHastings: Changed the order.

cpp/src/nvgraph_gdf.cu
ChuckHastings: fixed.

cpp/src/nvgraph_gdf.cu
ChuckHastings: changed everything to use nvg_

cpp/include/nvgraph_gdf.h
ChuckHastings: Can't make this const gdf_graph* gdf_G.  The graph can be changed if there is no CSR representation.

ChuckHastings: Changing vertices to const makes sense.  However, that would require changing the implementation in nvgraph.  It seems to me like this is more than we want to address now.  In fact, we probably should talk about how much we want to change nvgraph other than bug fixes.

python/cugraph/subgraph_extraction/test_subgraph_extraction.py
ChuckHastings: Fixed these naming issues

Issue #248, opened 2019-04-26
[FEA] Add flake8 to conda?


ChuckHastings:
Would it make sense to add flake8 to the conda environment so that developers can easily check their python code?

dillon-cullinan:
Flake8 is already used as a style check in CI. See here: https://github.com/rapidsai/cugraph/blob/branch-0.7/ci/checks/style.sh

ChuckHastings:
I guess that's the conversation.  In order to create a PR that passes the CI checks, the code needs to pass the flake8 style check.  While technically not necessary to build, I would argue that it is necessary if you're going to create a PR and get the system to update.  I suppose you could create the PR, wait for the error messages and keep pushing updates until CI runs the flake8 checker without an error.

So the question is do we want the conda dev environment contain the minimal tools necessary to compile, or do we want the conda environment to contain the tools necessary to effectively participate in development.

I'm fine with either answer.  I'm new to open source development and not accustomed to the conventions.  If I have to custom add a set of tools to my conda environment every time I create a new one so I can do effective development I can do that.

dillon-cullinan:
I've pushed a local build script to cuDF that allows devs to mimic the CI test on their local machine. This will be coming to all repos soon. I started my PR for cuGraph already. You will be able to use flake8 in that container if you would like. You bring up a good point though, I think this is something worth adding to the scripts README. :)

rlratzel:
Thank you @ChuckHastings for bringing this up. The script @dillon-cullinan is referring to is an excellent resource to do exactly what you and other devs need to do: validate overall correctness pre-PR and shorten the feedback loop.
As Dillon mentioned, we'll be adding these and other tools to other repos to help with this, and a larger (still to be defined) education initiative will be part of it.

PR #250, opened 2019-04-26
[REVIEW] Local build script to mimic gpuCI

dillon-cullinan:
Local script for developer to use to copy the gpuCI environment in a container.


dillon-cullinan:
Could be considered a solution to https://github.com/rapidsai/cugraph/issues/248

PR #261, opened 2019-05-01
[REVIEW] Add docs build script to cuGraph [skip ci]

dillon-cullinan:
Script for nightly doc building


ci/docs/build.sh
```diff
@@ -0,0 +1,79 @@
...
+logger "Check environment..."
+env
+
+logger "Check GPU usage..."
+nvidia-smi
+
+logger "Activate conda env..."
+source activate gdf
+conda install -c nvidia/label/cuda$CUDA_REL -c rapidsai/label/cuda$CUDA_REL -c rapidsai-nightly/label/cuda$CUDA_REL -c numba -c conda-forge \
+    cudf=$CUDF_VERSION rmm=$RMM_VERSION nvgraph networkx python-louvain sphinx sphinx_rtd_theme \
```
afender: @dillon-cullinan and @raydouglass I am still worried about this `nvgraph` dependency I keep seeing around. Recall that the current nvgraph conda package is not going to work for 0.7 release.

raydouglass: nvgraph is still listed as a dependency in the conda recipes so that's why is it included here. If it needs to be removed everywhere, can you open an issue? Any dev can make that change.

afender: Sure, I commented a week ago on nvgraph open sourcing about that https://github.com/rapidsai/cugraph/pull/194#issuecomment-486789449. I'll create a separate issue if that works better.

afender: I created https://github.com/rapidsai/cugraph/issues/266 to keep track of this.

Issue #275, opened 2019-05-06
[BUG] cugraph is importing libgdf_cffi, resulting in an import error


rlratzel:
**Describe the bug**
Various cugraph modules are importing libgdf_cffi, which no longer exists. *This needs to be resolved for 0.7*

**To Reproduce**
`py.test cugraph/python`
```
ImportError while importing test module '/rapids/cugraph/python/cugraph/triangle_count/test_triangle_count.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
cugraph/triangle_count/test_triangle_count.py:18: in <module>
    import cugraph
cugraph/jaccard/jaccard_wrapper.pyx:19: in init cugraph
    from libgdf_cffi import libgdf
E   ModuleNotFoundError: No module named 'libgdf_cffi'
```

**Expected behavior**
All tests pass

jakirkham:
Appears to be solved with PR ( https://github.com/rapidsai/cugraph/pull/276 ).

BradReesWork:
closed by PR #278

PR #278, opened 2019-05-07
Drop unused `libgdf_cffi` import

jakirkham:
Appears this import is no longer being used for anything. So just go ahead and drop it from all files that have it.

Fixes https://github.com/rapidsai/cugraph/issues/275


jakirkham:
Oops. Missed that PR ( https://github.com/rapidsai/cugraph/pull/276 ) already did this. Closing.

PR #284, opened 2019-05-08
comment out renumber test that's failing.  We'll retry after they fix…

ChuckHastings:
Comment out the renumber test that fails on T4's due to a cudf bug.

The renumber logic is not an issue... it fails in cudf before even getting there.

We can add this back in 0.8 once the cudf issue is resolved.


afender:
Is there a way to still test renumbering? 
Dropping all renumbering testing just because there is an intermittent cuDF failure on T4 seems a bit extreme to me.

PR #288, opened 2019-05-09
[REVIEW] Add boost conda dependency

raydouglass:


conda/recipes/libcugraph/meta.yaml
```diff
@@ -28,10 +28,12 @@ requirements:
     - libcudf=0.7*
     - cython
     - cudatoolkit {{ cuda_version }}.*
+    - boost
```
rlratzel: Is `boost` the same as `boost-cpp`?  I see the library in question provided by `boost-cpp` on another system:
```
/conda/pkgs/boost-cpp-1.68.0-h11c811c_1000/lib/libboost_system.so.1.68.0
```

raydouglass: @rlratzel I went with `boost` because that's what is in the conda environment files.
https://github.com/rapidsai/cugraph/blob/branch-0.8/conda/environments/cugraph_dev.yml#L22

rlratzel: sounds good to me.

cpp/CMakeLists.txt
```diff
@@ -136,7 +136,7 @@ include(ConfigureNvgraph)
 
 ###################################################################################################
 # - Find and add different modules and supporting repos -------------------------------------------
-find_package(Boost 1.45.0 COMPONENTS system)
+find_package(Boost REQUIRED)
```
dantegd: @afender would the change of not getting only the system COMPONENTS affect anything negatively? Also I believe the change removing 1.45 is just removing the minimum version cmake looks for

afender: Considering how close to the release we are, I'm fine as long as build and test are passing.

PR #300, opened 2019-05-24
[REVIEW] Personalized Page Rank

kaatish:


python/cugraph/pagerank/test_pagerank.py
```diff
@@ -116,18 +140,22 @@ def networkx_call(M, max_iter, tol, alpha):
 MAX_ITERATIONS = [500]
 TOLERANCE = [1.0e-06]
 ALPHA = [0.85]
+PERSONALIZATION_PERC = [0, 10, 50]
```
afender: I think we should add a comment explaining how these parameters impact the way the personalization is generated in `networkx_call `.

python/cugraph/pagerank/test_pagerank.py
```diff
@@ -80,6 +91,19 @@ def networkx_call(M, max_iter, tol, alpha):
     if M.shape[0] != M.shape[1]:
         raise TypeError('Shape is not square')
 
+    personalization = None
+    if personalization_perc != 0:
```
afender: 1. We should explain how the personalization vector is set. I'm not completely sure about what's in there in the end.
2. In cyber, some users don't need values. They just have a set of vertex they want to parametrize. We would need to add a test for that where `val[i] = 1.0/num_personalized_vertices` if `i` is parametrized and `0.0` otherwise.

cpp/include/algorithms.h
```diff
@@ -20,27 +20,31 @@
...
- *
- * @Param[out] *pagerank          The PageRank : pagerank[i] is the PageRank of vertex i.
- *
- * @Returns                       GDF_SUCCESS upon successful completion.
+ * @Param[in] has_guess              This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user doesn't have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
+ *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+ *                                   The initial guess must not be the vector of 0s. Any value other than 1 or 0 is treated as an invalid value.
+ * @Param[in] pagerank (optional)    Initial guess if has_guess=true
+ * @Param[in] personalization_subset (optional) Vertices for running personalized page rank
+ * @Param[in] personalization_values (optional) Values associated with personalization_subset vertices
```
afender: The size is arbitrary correct?

afender: I mean a user could potentially personalize everything (size=V) or a single vertex (size=1)?

cpp/src/link_analysis/pagerank.cu
```diff
@@ -164,16 +170,38 @@ int pagerank(IndexType n, IndexType e, IndexType *cscPtr, IndexType *cscInd, Val
...
+                      float alpha = 0.85,
                       float tolerance = 1e-4, int max_iter = 200,
                       bool has_guess = false) {
+  bool has_personalization = false;
+  int *prsVtx = nullptr;
+  WT  *prsVal = nullptr;
+  int prsLen = 0;
+  GDF_REQUIRE((personalization_subset == nullptr) == (personalization_values == nullptr), GDF_INVALID_API_CALL);
+  if (personalization_subset != nullptr) {
+    has_personalization = true;
```
afender: What happens when personalization_subset != nullptr but its size is 0? Does it run the regular PageRank?

cpp/src/link_analysis/pagerank.cu
```diff
@@ -114,7 +115,12 @@ int pagerank(IndexType n, IndexType e, IndexType *cscPtr, IndexType *cscInd, Val
     copy(n, pagerank_vector, tmp);
   }
 
-  fill(n, b, randomProbability);
+  if (has_personalization) {
+    fill(n, b, static_cast<ValueType>(0));
+    scatter(prsLen, prsVal, b, prsVtx);
```
afender: Should `b` sum to one from the mathematical perspective of the problem? 
If so, what if the user's values don't? Should we normalize by nrm1 when it is larger or fill the rest of the vector with the correct uniform value when it is smaller? We could also return an error.

cpp/include/algorithms.h
kaatish: Yes, that's right

cpp/src/link_analysis/pagerank.cu
kaatish: Good catch. Right now it will still try to run personalized page rank. This needs to be fixed to revert to normal pagerank.

cpp/src/link_analysis/pagerank.cu
afender: Discussed on slack with Aatish and Haekyu. We may normalize the input.

```
if |Q| == 1.0: continue
else if Q == 0.0 <- uniform dist for all nodes
else Q <- Q / |Q| 
```

cpp/include/algorithms.h
```diff
@@ -20,27 +20,31 @@
...
- * @Param[in] max_iter            The maximum number of iterations before an answer is returned. This can be used to limit the execution time and do an early exit before the solver reaches the convergence tolerance.
- *                                If this value is lower or equal to 0 cuGRAPH will use the default value, which is 500.
- *
- * @Param[out] *pagerank          The PageRank : pagerank[i] is the PageRank of vertex i.
- *
- * @Returns                       GDF_SUCCESS upon successful completion.
+ * @Param[in] has_guess              This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user doesn't have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
+ *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+ *                                   The initial guess must not be the vector of 0s. Any value other than 1 or 0 is treated as an invalid value.
+ * @Param[in] pagerank (optional)    Initial guess if has_guess=true
```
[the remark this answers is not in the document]
afender: Discussed on slack. Plan :
Since we run cugraph after Nx we could provide Nx output as guess to cugraph and expect to converge in 1 or 2 iterations.

python/cugraph/pagerank/test_pagerank.py
kaatish: Fixed [here](https://github.com/rapidsai/cugraph/pull/300/files?file-filters%5B%5D=.h&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.sh&owned-by%5B%5D=#diff-83a484342317d80b4509f185eee9a5bbR85)

python/cugraph/pagerank/test_pagerank.py
kaatish: 1. Fixed [here](https://github.com/rapidsai/cugraph/pull/300/files?file-filters%5B%5D=.h&file-filters%5B%5D=.md&file-filters%5B%5D=.py&file-filters%5B%5D=.sh&owned-by%5B%5D=#diff-83a484342317d80b4509f185eee9a5bbR85)
2. That case is already tested in personalization=None

cpp/src/link_analysis/pagerank.cu
kaatish: Fixed [here](https://github.com/rapidsai/cugraph/pull/300/files?file-filters%5B%5D=.cu&file-filters%5B%5D=.py&owned-by%5B%5D=#diff-c774f1d383f54e52fdc1f235c24c0251R120)

cpp/src/link_analysis/pagerank.cu
kaatish: Fixed [here](https://github.com/rapidsai/cugraph/pull/300/files?file-filters%5B%5D=.cu&file-filters%5B%5D=.py&owned-by%5B%5D=#diff-c774f1d383f54e52fdc1f235c24c0251R123)

PR #301, opened 2019-05-24
Added build.sh script, updated CI scripts and documentation

rlratzel:
This PR adds a new top-level script named build.sh that allows a contributor to build from source using the same commands and options the CI system, container builds, doc builds, nightly unit test runs, and (eventually) ASV build & test iterations use. The relevant CI scripts have been updated to call this script and all related documentation should be updated as well.

This PR allows a maintainer to update the common build steps in one place while also simplifying the build process for all contributors and other build-related operations.

The changes were tested using the gpuCI container via the `ci/local/build.sh` testing script, where cugraph branch-0.8 was successfully built and tested without errors.


build.sh
```diff
@@ -0,0 +1,118 @@
...
+
+################################################################################
+# Configure, build, and install libcugraph
+if (( ${NUMARGS} == 0 )) || hasArg libcugraph; then
+
+    mkdir -p ${LIBCUGRAPH_BUILD_DIR}
+    cd ${LIBCUGRAPH_BUILD_DIR}
+    cmake -DCMAKE_INSTALL_PREFIX=${INSTALL_PREFIX} \
+          -DCMAKE_CXX11_ABI=${BUILD_ABI} \
+          -DNVG_PLUGIN=True \
```
afender: `DNVG_PLUGIN` is not used anymore in cugraph

build.sh
```diff
@@ -0,0 +1,118 @@
...
+          -DCMAKE_BUILD_TYPE=${BUILD_TYPE} ..
+    make -j${PARALLEL_LEVEL} VERBOSE=${VERBOSE} ${INSTALL_TARGET}
+fi
+
+# Build and install the cugraph Python package
+if (( ${NUMARGS} == 0 )) || hasArg cugraph; then
+
+    cd ${REPODIR}/python
+    if [[ ${INSTALL_TARGET} != "" ]]; then
+	python setup.py build_ext --inplace
```
afender: Not sure about these python options, I never tried them. We typically just call `python setup.py`

build.sh
```diff
@@ -0,0 +1,118 @@
...
+BUILD_DIRS="${LIBCUGRAPH_BUILD_DIR} ${CUGRAPH_BUILD_DIR}"
+
+# Set defaults for vars modified by flags to this script
+VERBOSE=""
+BUILD_TYPE=Release
+INSTALL_TARGET=install
+
+# Set defaults for vars that may not have been defined externally
+#  FIXME: if INSTALL_PREFIX is not set, check PREFIX, then check
+#         CONDA_PREFIX, but there is no fallback from there!
```
afender: When one would use `PREFIX` or `INSTALL_PREFIX` instead of `CONDA_PREFIX`? 
cugraph devs use `CONDA_PREFIX` almost exclusively. Should  `CONDA_PREFIX`  be the default? It seems to me that cugraph may not be the only product interested in setting `PREFIX` or `INSTALL_PREFIX` env variables. That may lead to an unwanted install location if they are preferred to  `CONDA_PREFIX` .

build.sh
```diff
@@ -0,0 +1,118 @@
...
+INSTALL_PREFIX=${INSTALL_PREFIX:=${PREFIX:=${CONDA_PREFIX}}}
+PARALLEL_LEVEL=${PARALLEL_LEVEL:=""}
```
afender: Should this default to the number of CPU core of the machine? cugraph build takes a while otherwise.

afender: This is given by `nproc` or `cat /proc/cpuinfo | grep processor | wc -l`

build.sh
rlratzel: conda-build uses $PREFIX in the recipes so we still need to support that, but I believe I can safely change the precedence (it can check CONDA_PREFIX first, then PREFIX). INSTALL_PREFIX is a more self-documenting alternative being offered here, but I can remove that one altogether.

build.sh
rlratzel: This is part of an effort to standardize with how cudf and (now) cuml are doing. the separate `build_ext --inplace` and `install` lines allow the script to support the `-n` option, which might also be nice as a RAPIDS-wide build option. I think that the `--record` and `--single-version-externally-managed` may not be as important to cugraph (I believe cudf did need those though).

build.sh
rlratzel: The default results in a `make -j`, which builds in parallel with "no limit", and was the default for the other repos. I can change it to use `nproc` if you prefer though.

build.sh
rlratzel: after our in-person discussion, I'll remove `INSTALL_PREFIX` but leave the precedence of `PREFIX` over `CONDA_PREFIX` as-is to allow someone to override conda's location if necessary.

build.sh
rlratzel: I'll remove this.

PR #306, opened 2019-05-28
[REVIEW] Cmake warnings, tighter nvgraph integration, pytest with different RMM configurations

### LINKED ISSUE #293: [BUG] openmp runtime library conflict
Cmake cannot generate a safe runtime search for OpenMP

This is a stability issue that should be addressed before the release. This may cause bugs in all SNMG features.

On my machine, `libcugrpah.so` ends up picking the one in anaconda (which is installed as part of another package since I don’t see any openmp package when I do `conda list`).
Since we use the system compiler we may want to try using the openmp on the system too. Perhaps it is related to the intermittent nightly failures?

```CMake Warning at src/tests/CMakeLists.txt:48 (add_executable):
  Cannot generate a safe runtime search path for target <...> because
  files in some directories may conflict with libraries in implicit
  directories:

    runtime library [libgomp.so.1] in /usr/lib/gcc/x86_64-linux-gnu/5 may be hidden by files in:
      /home/afender/anaconda3/envs/cugraph_dev/lib

  Some of these libraries may not be found correctly.```

### LINKED ISSUE #298: [FEA] Tighter cugraph - nvgraph integration
Just a place holder for the work Sunghwa is currently doing. The plan was discussed on slack already. Let's continue the discussion here if needed.

Building nvgraph as a separate shared library is no longer a requirement. We should :

**move**
- nvgraph/cpp/src to cugraph/cpp/src/nvgraph 
- nvgraph/cpp/include to cugraph/cpp/src/nvgraph
- nvgraph/external to cugraph/thirdparty 

**remove**
- code in nvgraph that is useless to cugraph

**merge** 
- the two shared libraries (cmake files) to one

**improve** 
- the way we access cub in cugraph. Ideally, we should clone cub only once at the top cugraph level, but gunrock clones its own. We may keep gunrock as is, but just not use this version (remove include path to gunrock's cub)

### PR DESCRIPTION
**Pull request by seunghwak**
Close #293 Close #298

### CODE REVIEW

File: cpp/CMakeLists.txt [hunk 1]
```diff
@@ -159,75 +126,149 @@ else()
...
-###################################################################################################
-# - library paths ---------------------------------------------------------------------------------
-
-link_directories("${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}" # CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES is an undocumented/unsupported variable containing the link directories for nvcc
-                 "${CMAKE_BINARY_DIR}/lib"
-                 "${GTEST_LIBRARY_DIR}")
-
-###################################################################################################
-# - library targets -------------------------------------------------------------------------------
+else(NVGRAPH_LIGHT MATCHES True)
```
afender: We could get rid of this `else` branch I think

cpp/cmake/Modules/ConfigureNvgraph.cmake
```diff
@@ -20,7 +20,7 @@ endif(NOT CMAKE_CXX11_ABI)
 file(MAKE_DIRECTORY "${NVGRAPH_ROOT}/cpp/build")
 #file(MAKE_DIRECTORY "${NVGRAPH_ROOT}/install")
 
-execute_process(COMMAND ${CMAKE_COMMAND} -G ${CMAKE_GENERATOR} .. -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DNVGRAPH_LIGHT=${NVGRAPH_LIGHT}
+execute_process(COMMAND ${CMAKE_COMMAND} -G ${CMAKE_GENERATOR} .. -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DCMAKE_INSTALL_PREFIX=${CMAKE_INSTALL_PREFIX} -DNVGRAPH_LIGHT=${NVGRAPH_LIGHT}
```
afender: I don't see `include(ConfigureNvgraph)` being called in the main cmakelists. Should we delete this file?

cpp/CMakeLists.txt
seunghwak: Got it.

cpp/cmake/Modules/ConfigureNvgraph.cmake
seunghwak: Nice catch, and I will delete the file.

.gitmodules
```diff
@@ -1,8 +1,8 @@
-[submodule "cpp/nvgraph/cpp/thirdparty/cub"]
-	path = cpp/nvgraph/cpp/thirdparty/cub
-	url = https://github.com/NVlabs/cub.git
-	branch = 1.8.0
-[submodule "cpp/nvgraph/external/cusp"]
-	path = cpp/nvgraph/external/cusp
+[submodule "thirdparty/cusp"]
+	path = thirdparty/cusp
```
afender: Is this really needed when using the "NVGRAPH_LIGHT"? I don't recall any feature exposed in cugraph needing cusp. I think it worth trying to removing it.

cpp/CMakeLists.txt
```diff
@@ -244,51 +285,150 @@ add_library(cugraph SHARED
...
-else()
-    message(AUTHOR_WARNING "Google C++ Testing Framework (Google Test) not found: automated tests are disabled.")
-endif()
+###   Use ${OpenMP_CXX_LIB_NAMES} instead of OpenMP::OpenMP_CXX to avoid the following warnings.
+###
+###   Cannot generate a safe runtime search path for target TARGET_NAME
+###   because files in some directories may conflict with libraries in implicit
+###   directories:
+###   ...
+###
```
afender: That was a tricky one! Questions : 
1. In practice, which openmp is loaded for cuGraph now?
2. Are compiler and OpenMP versions interdependent? If not, is it safe to use a system compiler and a conda OpenMP and vice versa?
3. How safe is that for deployment? I mean are openMP headers and lib backward/forward compatible? If we build libcugraph.so using a very old/recent OpenMP, will the binary work for all openMP versions or should we add a note in the readme?

.gitmodules
seunghwak: Yes, this compiles without thirdparty/cusp, and I removed this.

cpp/CMakeLists.txt
```diff
@@ -244,51 +285,150 @@ add_library(cugraph SHARED
...
+###   soft links) in the search paths (implicit directoires + BUILD_RAPTH). There is one in the
+###   path included in BUILD_RPATH when ${CUDF_LIBRARY} and ${RMM_LIBRARY} are added; this one can
+###   potentially hide the one in the provided full path and cmake generates a warning (and RPATH
+###   is searched before the directories in /etc/ld.so/conf; ld.so.conf does not coincide but
+###   overlaps with implicit directories).
+###
+###   If we provide just the library names (gomp;pthread), cmake does not generate warnings (we
+###   did not specify which libgomp.so should be loaded in runtime), and the one first found in
+###   the search order is loaded (we can change the load library by setting LD_LIBRARY_PATH or
+###     manually editing BUILD_RPATH).
```
afender: Are there cases where a developer/user would want to manually set it? If so, should we document this in CONTRIBUTING.md and expose a cmake variable?

cpp/CMakeLists.txt
seunghwak: 1. So there are two answers: one for build time and another for install time. For the build time, assuming that LD_LIBRARY_PATH does not include a path to openmp, the conda one will be loaded (because the path to libcudf.so and librmm.so will go to BUILD_RPATH). For the install time, no RPATH(or RUNPATH) is specified, and libraries are searched in the following order 1) LD_LIBRARY_PATH 2) /etc/ld.so.conf 3) builtin directories - basically /lib and /usr/lib (https://gitlab.kitware.com/cmake/community/wikis/doc/cmake/RPATH-handling).

If you want to run with the system openmp in build time testing (without specifying LD_LIBRARY_PATH), we need to set BUILD_RPATH accordingly. I provided the instruction in the comment.

```
###   set(TARGET_BUILD_RPATH "")
###   foreach(TMP_VAR_FULLPATH IN LISTS OpenMP_CXX_LIBRARIES)
###       get_filename_component(TMP_VAR_DIR ${TMP_VAR_FULLPATH} DIRECTORY)
###       string(APPEND TARGET_BUILD_RPATH "${TMP_VAR_DIR};")
###       get_filename_component(TMP_VAR_REALPATH ${TMP_VAR_FULLPATH} REALPATH)
###       get_filename_component(TMP_VAR_DIR ${TMP_VAR_REALPATH} DIRECTORY)
###       # cmake automatically removes duplicates, so skip checking.
###       string(APPEND TARGET_BUILD_RPATH "${TMP_VAR_DIR};")
###   endforeach()
###   string(APPEND TARGET_BUILD_RPATH "${CONDA_PREFIX}/lib")
###   message(STATUS "TARGET_BUILD_RPATH=${TARGET_BUILD_RPATH}")
###   set_target_properties(target PROPERTIES
###       BUILD_RPATH "${TARGET_BUILD_RPATH}")
```

2. Newer compilers are necessary to use newer OpenMP features. If you use new OpenMP features (compiled using a newer compiler) and use an old library, it will not run (e.g. according to https://www.openmp.org/resources/openmp-compilers-tools/, "From GCC 6.1, OpenMP 4.5 is fully supported for C and C++."), but this affects only build time testing. On deployment, users should have a proper OpenMP library and that should appear first in the search order (than the outdated one if exists, see the answer for 1 for the search order).

3. This has no impact on deployment. No changes are made for installed one (cmake removes RUNPATH on installation unless we override the default option).

cpp/CMakeLists.txt
seunghwak: See the answer for 1 for your previous question. I guess it will rarely be a case (unless we are using new OpenMP features and conda comes with old OpenMP library), but I included the instruction in comments for the case this actually becomes a problem (it took me a quite good amount of effort to figure this out, and I wanted to properly document this in a place we can easily find once this becomes necessary).

cpp/CMakeLists.txt
afender: Perfect. I checked the recent OpenMP releases note and I doubt we use any >4.0 features so we should be fine for any realistic configuration.

cpp/CMakeLists.txt
afender: Ok, that sounds safe enough. Let's keep it like this for now. The documentation is very useful, we can quickly provide a more advanced solution if needed.

PR #315, opened 2019-06-01
[REVIEW] Fea ext new weak cc

aschaffer:
New PR for Weakly Connected Components adapted from cuML. This is the C++ part. Python bindings in a later PR.


cpp/src/components/weak_cc.cuh
```diff
@@ -0,0 +1,344 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+///#include "cuda_utils.h"
+///#include "array/array.h"
+
```
BradReesWork: unused code should be removed

Issue #192, opened 2019-03-28
[ENH] class Graph API (python) design needs improvements


seunghwak:
**Is your feature request related to a problem? Please describe.**
Related: https://github.com/rapidsai/cugraph/issues/187

class Graph (defined in c_graph.pyx) stores a graph and provides functions to initialize, delete, convert (between different graph formats), and query.

* initialize: add_edge_list, add_adj_list
* delete: delete_edge_list, delete_adj_list, delete_transposed_adj_list
* convert: add_transposed_adj_list, (view_edge_list, view_adj_list, num_vertices).
* query: view_edge_list, view_adj_list, num_vertices

There are corresponding C++/CUDA level functions.

* initialize: gdf_edge_list_view, gdf_adj_list_view
* delete: gdf_delete_edge_list, gdf_delete_adj_list, gdf_delete_transposed_adj_list
* convert: gdf_add_edge_list, gdf_add_adj_list, gdf_add_transposed_adj_list
* query: N/A

There exist some inconsistencies in naming, and some functions may surprise users. In general, surprising users is bad in API design. If a user expects a function to do A, but the function silently does B, or (B + A), it may leads to a hard-to-fix bug till the user finds that the function is doing something unexpected.

So inconsistencies. In the python level, add_edge_list and add_adj_list initialize a graph while add_transposed_adj_list performs conversion. In CUDA/C++, gdf_add_...\_list performs conversion but add_edge_list and add_adj_list perform initialization in python. In CUDA/C++, gdf_..._list_view functions perform initialization. Calling both add_edge_list and add_adj_list can result in a single graph instance storing two different graphs (issue #187). And we have num_vertices, why no num_edges?

And the surprise part. view_edge_list, view_adj_list, and num_vertices implicitly perform conversion. This requires more thoughts as there is a design trade-off. Implicit (or automatic) conversion allows users to type less. But this can require a significant amount of computing and also more memory for large graphs. It can be quite surprising for users to see memory allocation error after long wait when they simply query the number of vertices.

**Describe the solution you'd like**
So we may clarify the API functions for initialization, deletion, conversion, and query and also need to fix inconsistencies between the python API and C++/CUDA API.

For initialization:
* Option 1) initialize in instance creation (similar to cudf dataframe or series).
e.g. G = cugraph.Graph(col_0, col_1, col_2, format) # col_0 and col_1 do not sound like good names but col_0 and col_1 can have different meanings in COO and CSR
* Option 2) Initialize using view
[...]

PR #321, opened 2019-06-06
[REVIEW] First step for NetworkX compatibility

seunghwak:
First step to address https://github.com/rapidsai/cugraph/issues/192

added

add_edge_from(), add_weighted_edges_from(), from_cudf_edgelist(), clear(), and number_of_nodes()


python/cugraph/graph/convert_matrix.pyx
```diff
@@ -0,0 +1,50 @@
...
+# .py and should be located outside the python/cugraph/bindings directory.
+
+import cudf
+
+from cugraph import Graph
+
+
+def from_cudf_edgelist(df, source='source', target='target', weight=None):
+    """
+    Add edges to a graph. Currently, self should be an empty graph.
```
ChuckHastings: I'm confused.  The comment suggests that an error will be raised if self is not empty.  The code does nothing with self, creates an empty graph, adds edges and returns the new graph object.  This seems inconsistent.

python/cugraph/graph/test_graph.py
```diff
@@ -75,6 +76,40 @@ def compare_offsets(offset0, offset1):
     return True
 
 
+def compare_graphs(nx_graph, cu_graph):
```
aschaffer: compare_graphs() is not exactly correct. You can have 2 identical graphs, one with vertex indices permuted wrt the other and they may have different row_offsets and column_indices. This is the problem of graph isomorphism and does _not_ currently have (such) a trivial solution (it's active research). Perhaps this function should be renamed and comments added to reflect that is a very simplistic heuristics for graph comparison.

ChuckHastings: To clarify... this compare_graphs will return the correct value for it's intended purpose.  If you take the same set of edges, with the same numbering of vertices and create a graph, this function will return true if the resulting graphs match and false if they don't.

The issue @aschaffer is mentioning is a broader graph comparison issue.

seunghwak: Thank you very much for the comments, and yes, this is not to test two graphs are isomorphic but to test two graphs are automorphic (isomorphic + vertex mapping between two graphs is an identity mapping). I will add comments to make this point clear.

python/cugraph/graph/convert_matrix.pyx
seunghwak: A great catch, I forgot to update comments after moving this from inside the Graph class (as a member function) to outside the Graph class. I will update the comments.

seunghwak: And I should remove add_edges_from() and add_weighted_edges_from() as discussed in the email, so no need to review those functions (I didn't know this PR will be reviewed at this point, I was waiting for cudf's fix so all the tests can be passed before starting a review).

seunghwak:
Why's this PR in the Waiting on Author state? 

python/cugraph/graph/test_graph.py

is not marked as outdated may be due to my force-push (I should do that because I accidentally added hornet integration related updates to this branch) but I addressed the comments in more recent commits.

aschaffer:
Is there any action for me here? I thought I've already approved this.

afender:
> Is there any action item for me here? I thought I've already approved this.

It is best to select "approve" when you approve the PR so that others can see that you approved. In this case, your status is on "commented" for instance

aschaffer:
The intention was to comment _without_ blocking the merge (i.e., the comment is merely a suggestion and not a reason to prevent the merge). The "Comment" option, in Review, doesn't prevent merging, does it?

PR #324, opened 2019-06-12
[REVIEW] Fea ext weakcc

aschaffer:
Fixed crash in Weakly CC for larger graphs; added src/topology/toipology.cuh for various topology-related tasks; added adj. matrix symmetry check and enforced the check in WeakCC.


cpp/src/components/utils.h
```diff
@@ -122,8 +122,8 @@ class Exception : public std::exception {
  */
 template <typename Type>
 void copy(Type *dst, const Type *src, size_t len, cudaStream_t stream) {
-    CUDA_CHECK(cudaMemcpyAsync(dst, src, len * sizeof(Type),
```
afender: It is important to check the return code of cudaMemcpyAsync. I noticed CUDA_CHECK has been commented. There are 3 occurrences in this file.

aschaffer: Ok, done.

PR #325, opened 2019-06-14
[REVIEW] Implement SSSP with predecessors (cuda/C++)

pgera:
The current SSSP implementation in cugraph computes only distances, not
predecessors. The predecessors are necessary for extracting the actual paths.
This commit provides a new SSSP implementation which can compute both distances
and predecessors.

The implementation is based on frontier expansion similar to BFS. The primary
differences compared to BFS are that:
 - Vertices can be visited multiple times during the course of the algorithm
 - Distances and weights are floats/doubles. Hence, in a given frontier
expansion, a vertex's distance may be updated mutiple times by multiple parents.
In order to avoid duplicates in the frontier and to assign the right predecessors,
SSSP needs a two phase approach. The current implementation strikes a balance
between performance and the additional memory requirements. Further optimisations
that trade off space or time are possible.

Since this implementation shares code with BFS, common functions from BFS have
been factored out.

FREEBIE:
Fixed one memory overflow bug in the common BFS code

TODO:
Python bindings will follow in a separate PR


GPUtester:

afender:
rerun test

cpp/CMakeLists.txt
```diff
@@ -218,6 +218,7 @@ add_library(cugraph SHARED
     src/structure/cugraph.cu
     src/link_analysis/pagerank.cu
     src/traversal/bfs.cu
+    src/traversal/sssp.cu
```
seunghwak: Does this replace the previous src/traversal/nvgraph_sssp.cu? If yes, can we remove the outdated nvgraph_sssp.cu?

cpp/src/tests/sssp/.clang-format
```diff
@@ -0,0 +1,12 @@
+---
```
afender: I am not against adding a clang format file, but I missed the discussion on setting these parameters. Is this something standard that we are using elsewhere?

afender:
rerun tests

dillon-cullinan:
add to whitelist

cpp/CMakeLists.txt
pgera: Yes, it replaces the nvgraph version. I haven't added the python bindings yet. Once the python bindings are in place, they will naturally call the new code.

cpp/src/tests/sssp/sssp_test.cu
```diff
@@ -0,0 +1,419 @@
...
+                     "--normalized --quiet",
+                     0),
+        SSSP_Usecase(RMAT,
+                     "grmat --rmat_scale=12 --rmat_edgefactor=8 --device=0  "
+                     "--normalized --quiet",
+                     10),
+        SSSP_Usecase(MTX, "test/datasets/dblp.mtx", 100),
+        SSSP_Usecase(MTX, "test/datasets/wiki2003.mtx", 100000),
+        SSSP_Usecase(MTX, "test/datasets/karate.mtx", 1)
+        //,SSSP_Usecase(MTX, "test/datasets/cit-Patents.mtx", 100)
```
afender: 2 questions : 
- why cit-Patents.mtx is commented? We may want to delete or uncomment.
- In the previous sssp test suite, we were comparing against reference results stored in files on s3 for a couple more datasets in order to capture different scenario. Can you just double check that we are still testing both directed and undirected?

cpp/src/tests/sssp/.clang-format
```diff
@@ -0,0 +1,12 @@
...
+BasedOnStyle: chromium 
+AlignOperands: false
+AlwaysBreakAfterReturnType: None
+AlwaysBreakAfterDefinitionReturnType: None
+ColumnLimit: 79
+BinPackArguments: false
+BinPackParameters: false
+Standard: Cpp11
+...
+
```
seunghwak: cuGraph currently does not use clang-format, so this is out of sync with the cuGraph repository.

That being said, cuDF and cuML are incorporating clang-format(
https://github.com/rapidsai/cudf/issues/540
https://github.com/rapidsai/cuml/pull/616
)

We should consider incorporating this as well at the cuGraph level (@BradReesWork, @afender).

cpp/src/utilities/graph_utils.cuh
```diff
@@ -517,5 +517,45 @@ namespace cugraph
...
         cudaCheckError();
     }
+    
+    template<typename DistType>
+	    bool has_negative_val(DistType* arr, size_t n){
+		    // custom kernel with boolean bitwise reduce may be 
+		    // faster.
+#if 0
+		    // cub throws errors with double in cuda-memcheck
+		    // switch to thrust until resolved
```
afender: Should we fill an issue in CUB repo?

cpp/src/tests/sssp/.clang-format
afender: Yes, that caught my attention too. 
Let's add it @pgera. Just make sure it is consistent with the rest of RAPIDS.

cpp/src/tests/sssp/sssp_test.cu
pgera: I commented it since the validation runs (serial textbook cpu code) take a while, particularly for validating predecessors since it needs to create the min_edge_weight map. I can uncomment it if runtime is not a huge concern. I can optionally add other larger graphs in a separate case that does only performance benchmarks, but not validation. I ran other cases too and they all pass (with cuda-memcheck).

cpp/src/tests/sssp/.clang-format
pgera: This was committed accidentally. It should have been in gitignore. That said, I just picked the params after some basic trial and error. Also, I didn't attempt any deeper integration with clang's tools. This was just a quick fix to get some basic formatting.

afender:
2 more things : 
- Changelog needs to be manually updated for each PR. 
- It seems that the build failure is related to the gunrock problem that was reported this morning on slack.

cpp/src/tests/sssp/.clang-format
afender: Let's remove it for now and address this as part of another PR then

cpp/CMakeLists.txt
seunghwak: OK, please don't forget to delete the dead/unused/outdated files once you finished the job. I am still reviewing your code, but you did a great job in writing clean code while keeping the code in sync with the (somewhat convoluted) cuGraph code base. Very impressive!!!

cpp/src/utilities/graph_utils.cuh
pgera: I think so. I didn't investigate it too much though. This was reported only by cuda-memcheck. There were no other side-effects.

cpp/CMakeLists.txt
afender: Great, yes please make sure to delete the dead code once the python bindings are in place.

cpp/src/tests/sssp/sssp_test.cu
afender: Sounds good. The current test solution is fine if it captures all common cases.

cpp/include/types.h
```diff
@@ -19,13 +19,16 @@ void gdf_col_delete(gdf_column* col);
...
+typedef enum gdf_prop_type{GDF_PROP_UNDEF, GDF_PROP_FALSE, GDF_PROP_TRUE} GDFPropType;
+
 struct gdf_graph_properties {
   bool directed;
   bool weighted;
   bool multigraph;
   bool bipartite;
   bool tree;
-  gdf_graph_properties() : directed(false), weighted(false), multigraph(false), bipartite(false), tree(false){}
+  GDFPropType has_negative_edges;
```
seunghwak: Does has_negative_edges need to be gdf_graph_properties's member variable?

Currently, this is used only for SSSP, but we are running has_negative_val() for every gdf_add_adj_list() call. This incurs additional performance overhead to every graph algorithm (and currently, cuGraph C++ API has two mechanisms to initialize graphs (add_edge_list_view and add_adj_list_view) and three mechanisms to convert graph formats (add_edge_list, add_adj_list, and add_transposed_adj_list... I know the names are misnomers and can be very confusing... We're planning to fix the gdf_graph structure...). If we are going to update has_negative_edges for every graph object, we need to consider all these possibilities, but I think this will add too much performance overhead for other graph analytics if they do not use has_negative_edges.

An alternative is too call has_negative_val() inside the SSSP routine; this adds unnecessary overhead if we call SSSP or other algorithms using has_negative_edges on the same graph multiple times but I think this is less overhead than incurring the overhead of has_negative_val() for every graph analytics.

cpp/src/tests/CMakeLists.txt
```diff
@@ -139,6 +139,14 @@ set(PAGERANK_TEST_SRC
 
 ConfigureTest(PAGERANK_TEST "${PAGERANK_TEST_SRC}")
 
+###################################################################################################
+#- native sssp tests -- ---------------------------------------------------------------------------
+set(NATIVE_SSSP_TEST_SRCS
+	"${CMAKE_SOURCE_DIR}/../thirdparty/mmio/mmio.c"
+	"${CMAKE_CURRENT_SOURCE_DIR}/sssp/sssp_test.cu")
+
+ConfigureTest(NATIVE_SSSP_TEST "${NATIVE_SSSP_TEST_SRCS}")
```
seunghwak: And once you remove the outdated SSSP_TEST, you can rename this to just SSSP_TEST :-)

cpp/src/tests/sssp/.clang-format
afender: https://github.com/rapidsai/cugraph/issues/326

cpp/src/tests/sssp/.clang-format
afender: https://github.com/rapidsai/cugraph/issues/326

seunghwak: @pgera and just FYI, RAPIDS is officially using C++14.

cpp/include/types.h
pgera: I had the exact same discussion with @afender yesterday. I originally had it inside the SSSP code behind a flag, and my plan was to use some other persistent storage to set the flag on first run but not on subsequent runs. That makes it somewhat kludgy though.

seunghwak: OK, may be one way to resolve this is (and you already defined GDF_PROP_UNDEF) to maintain this as a some optional graph state variable which may be set to UNDEFINED in most cases, but at the beginning of any algorithm requiring this property, you may check whether it is UNDEFINED, then run has_negative_val)() and set this value. If set, then, just reuse.

If we do this, we don't need to run has_negative_val() everytime we initialize graphs.

We will consider doing this in a more elegant way in the planning gdf_graph redesign, but as a temporary solution, you may update this in the SSSP call instead of add_adj_list().

pgera: Yes, fair point. I'll move the negative edges check inside sssp.

afender: So, in the long run, we want a safe mode where the graph input is validated and a performance mode where we skip these checks. This should be controlled by the cugraph handle. Having these checks in one place at loading time allows preserving the analytics call performance even when running in safe mode.

I agree that most cugraph algorithms don't care about negative weights but for the majority that's because they just don't use weights at all. In this case, this check does not happen and performance is not impacted.

seunghwak: And don't forget to reset has_negative_edge to UNDEFINED when the graph is reinitialized (gdf_edge_list_view & gdf_adj_list_view).

cpp/src/utilities/graph_utils.cuh
afender: I tend to trust cuda memcheck, some real bug can be silent for a long time. It would be ideal if you could find a quick repro to fill a CUB issue after you are done with the rest of the SSSP project.

cpp/include/types.h
seunghwak: Yeah... I think Alex's comment makes sense. Then, we should make sure has_negative_edge is properly updated in every possible path of graph initialization (not just add_adj_list which coverts an existing graph to the CSR format).

afender: For completeness, spectral clustering and Louvain are other algorithms that would benefit from that since they support positive weights only in theory. I think other algo don't use weights at all.

seunghwak: We are using CSR for (AFAIK) every graph algorithm (minus PageRank? which uses CSC instead), so doing this in add_adj_list is safe for most cases, but I don't think it's future proof.

afender: Agreed. I thought it was in gdf_edge_list_view actually. It is probably better to add it for the edge list since this is supposed to be the place where we first encounter the input (in data science workloads).

cpp/src/tests/sssp/sssp_test.cu
```diff
@@ -0,0 +1,419 @@
...
+      create_gdf_column(cooRowInd, &col_src);
+      create_gdf_column(cooColInd, &col_dest);
+      create_gdf_column(cooVal, &col_weights);
+      v = m;
+      e = nnz;
+    } else {
+      ASSERT_TRUE(0);
+    }
+
+    // std::cout << "v = " << v << "e = " << e << "\n";
```
seunghwak: Better delete temporary printouts.

cpp/src/tests/sssp/sssp_test.cu
```diff
@@ -0,0 +1,419 @@
...
+                static_cast<DistType>(RAND_MAX);
+          }
+        }
+      }
+
+      ASSERT_EQ(fclose(fpin), 0);
+      // gdf columns
+      create_gdf_column(cooRowInd, &col_src);
+      create_gdf_column(cooColInd, &col_dest);
+      create_gdf_column(cooVal, &col_weights);
```
seunghwak: Where are you releasing memory allocated in the above three lines?

I know it is very confusing, but in gdf_edge_list_view(), ownership is set to 0, the resulting graph does not release memory pointed by the gdf_column's data. So you need to manually release memory once the graph object is destroyed and you're done. Also, note that there is another version of create_gdf_column() which returns std::unique_ptr and releases memory when the std::unique_ptr object becomes out of scope (in this case, you need to make sure that this object out-lives the graph object).

cpp/src/traversal/bfs.cu
```diff
@@ -17,9 +17,10 @@
 #include "rmm_utils.h"
 
 #include "utilities/graph_utils.cuh"
+#include "traversal_common.cuh"
 #include "bfs_kernels.cuh"
 
-using namespace bfs_kernels;
+//using namespace bfs_kernels;
```
seunghwak: Better just delete the dead code.

cpp/src/traversal/bfs.cu
```diff
@@ -270,20 +271,20 @@ namespace cugraph {
 
               //We need to prepare the switch back to top down
               //We couldnt keep track of mu during bottom up - because we dont know what mf is. Computing mu here
-              count_unvisited_edges(unvisited_queue,
+	      bfs_kernels::count_unvisited_edges(unvisited_queue,
```
seunghwak: @BradReesWork @afender

What is cuGraph's policy on namespaces?

If we're following cudf (https://github.com/rapidsai/cudf/blob/branch-0.8/cpp/docs/TRANSITIONGUIDE.md),

We should use cugraph:: for public API, cugraph::detail:: for cugraph internal implementations, and anonymous namespace for implementations restricted to a single translation unit.

cpp/src/traversal/sssp.cu
```diff
@@ -0,0 +1,430 @@
...
+  traversal::flag_isolated_vertices(
+      n, isolated_bmap, row_offsets, vertex_degree, d_nisolated, stream);
+
+  cudaMemcpyAsync(&nisolated,
+                  d_nisolated,
+                  sizeof(IndexType),
+                  cudaMemcpyDeviceToHost,
+                  stream);
+
+  // We need nisolated to be ready to use
```
seunghwak: What is nisolated???

cpp/src/traversal/sssp.cu
```diff
@@ -0,0 +1,430 @@
...
+                  stream);
+
+  // Number of vertices in the frontier and number of out-edges from the
+  // frontier
+  IndexType mf, nf;
+  nf = 1;
+  int iters = 0;
+
+  while (nf > 0) {
+#if 0
```
seunghwak: Better delete temporary code for debugging

afender:
rerun tests

cpp/src/traversal/sssp.cu
pgera: nisolated is the number of isolated vertices, where isolated is defined as vertices with zero out-degree. The naming is a bit misleading, but I kept it to keep it consistent with the bfs codebase. I can also change it uniformly to something else across bfs and sssp.

seunghwak: Ah... OK, I mistakenly thought it's a typo.

seunghwak:
@pgera And don't forget to update https://github.com/rapidsai/cugraph/blob/branch-0.8/CHANGELOG.md

To merge, all the tests should pass, and this requires updating the change log (and that's the only failing check).

cpp/src/traversal/sssp.cu
pgera: Ack. Done in next version.

cpp/src/traversal/bfs.cu
pgera: Ack. Done in next version.

cpp/src/tests/sssp/sssp_test.cu
pgera: Ack.

cpp/src/tests/sssp/sssp_test.cu
pgera: Ack. Added free in revised commit.

cpp/include/types.h
pgera: I have moved the checks to adj and edge list view functions now. So they should be called on all normal paths.

cpp/src/traversal/bfs.cu
afender: makes sense. let's add this to the c++ refactoring task.

cpp/CMakeLists.txt
pgera: Done in https://github.com/rapidsai/cugraph/pull/341

cpp/src/tests/CMakeLists.txt
pgera: Done in https://github.com/rapidsai/cugraph/pull/325

Issue #61, opened 2019-02-07
[FEA] [NV] Renumbering support for String vertex labels


BradReesWork:
Update Renumbering to accept a string column (nvString) so that users can specify a vertices with string labels

- [ ] add nvStrings as a third party library
- [ ] accept a string column

**Blocked until cuDF has string as a column type**

BradReesWork:
replaced by issue: #636

PR #332, opened 2019-06-18
[REVIEW] Fea renumbering with strings

ChuckHastings:
This PR addresses #61 

 * Added NVSTRINGS, NVCATEGORY and NVTEXT libraries to cmake files - although currently only use NVSTRINGS
 * Modified C++ call signature to support a hashing function and a comparison function to support other data types
 * Created a hashing and comparison function for strings on the GPU
 * Added C++ unit tests for strings
 * Implemented python bindings - but they don't work because the cudf functions we rely upon from the C++ side (namely gdf_column_view) don't support strings yet.


seunghwak:
@ChuckHastings If you add "close #61" in your PR (instead of "This PR addresses #61"), #61 gets automatically closed once this PR gets merged, and AFAIK, I think we are expected to do this.

ChuckHastings:
> @ChuckHastings If you add "close #61" in your PR (instead of "This PR addresses #61"), #61 gets automatically closed once this PR gets merged, and AFAIK, I think we are expected to do this.

Understood.  I believe the discussion yesterday was that this issue isn't officially closed until the python bindings are resolved.  Although if @BradReesWork  or @afender want to have this close with the C++ work and define a new issue to add python bindings once the cudf support for strings is ready then we could take that approach also.  I have no strong opinion either way.

seunghwak:
> > @ChuckHastings If you add "close #61" in your PR (instead of "This PR addresses #61"), #61 gets automatically closed once this PR gets merged, and AFAIK, I think we are expected to do this.
> 
> Understood. I believe the discussion yesterday was that this issue isn't officially closed until the python bindings are resolved. Although if @BradReesWork or @afender want to have this close with the C++ work and define a new issue to add python bindings once the cudf support for strings is ready then we could take that approach also. I have no strong opinion either way.

OK, I thought this PR will close the issue, but if not, no complaint!

cpp/src/converters/renumber.cu
```diff
@@ -34,79 +44,176 @@ gdf_error gdf_renumber_vertices(const gdf_column *src, const gdf_column *dst,
...
+      int32_t *tmp;
+
+      ALLOC_TRY((void**) &tmp, sizeof(int32_t) * src->size, stream);
+      gdf_column_view(src_renumbered, tmp, src->valid, src->size, src->dtype);
+
+      ALLOC_TRY((void**) &tmp, sizeof(int32_t) * src->size, stream);
+      gdf_column_view(dst_renumbered, tmp, dst->valid, dst->size, dst->dtype);
+
+      gdf_error err = cugraph::renumber_vertices(src->size,
+                                                 (const int32_t *) src->data,
```
seunghwak: I recommend using C++ style type casting (static_cast in general and reinterpret_cast only when it is absolutely necessary, and if we need to use reinterpret_cast, then we may add comment to explain why it is OK), static_cast is safer than C style casting which works like reinterpret_cast.

cpp/src/converters/renumber.cu
```diff
@@ -34,79 +44,176 @@ gdf_error gdf_renumber_vertices(const gdf_column *src, const gdf_column *dst,
...
+      ALLOC_TRY((void**) &tmp, sizeof(int32_t) * src->size, stream);
+      gdf_column_view(src_renumbered, tmp, src->valid, src->size, GDF_INT32);
+
+      ALLOC_TRY((void**) &tmp, sizeof(int32_t) * src->size, stream);
+      gdf_column_view(dst_renumbered, tmp, dst->valid, dst->size, GDF_INT32);
+
+      NVStrings *srcList = reinterpret_cast<NVStrings*>(src->data);
+      NVStrings *dstList = reinterpret_cast<NVStrings*>(dst->data);
+
+      typedef std::pair<const char *, size_t>   pair_t;
```
seunghwak: Are you using pair_t? I can see std::pair<const char *, size_t> right below this code, so if not used, I recommend deleting this (and I generally do not prefer excessive use of typedef or something like "using pair_t = std::pair<const char*, size_t>" to save little bit of typing, I recommend using aliases only when it significantly improves code readability, if I see pair_t, I need additional lookup to figure out pair_t actually means std::pair<const char*, size_t>, but this is just recommendation).

cpp/src/converters/renumber.cuh
```diff
@@ -31,134 +31,111 @@
...
+#include <nvstrings/NVStrings.h>
 #include <cuda_runtime_api.h>
 
 #include "utilities/error_utils.h"
 #include "utilities/graph_utils.cuh"
-#include "utilities/heap.cuh"
+#include "sort/bitonic.cuh"
 #include "rmm_utils.h"
 
+namespace thrust {
```
seunghwak: Should this be in thrust's namespace? If this is just for cugraph, it may better be in cugraph::detail namespace (or anonymous namespace if used only within a single translation unit following cudf's convention). Someone seeing thrust::swap(pair0, pair1) may mistakenly think this is defined in thrust.

cpp/src/sort/bitonic.cuh
```diff
@@ -0,0 +1,561 @@
...
+                         //
+                         // NOTE: If size is 0 or 1 then no
+                         //       sorting is required, so we'll
+                         //       eliminate those bins here
+                         //
+                         if (size > 1)
+                           atomicAdd(d_lrb + __clz(size), 1);
+                       });
+      
+#ifdef DEBUG
```
seunghwak: May better delete temporary debug code?

cpp/CMakeLists.txt
```diff
@@ -155,6 +155,47 @@ if(BUILD_TESTS)
...
+find_path(NVSTRINGS_INCLUDE "nvstrings"
+          HINTS "$ENV{NVSTRINGS_ROOT}/include"
+                "$ENV{CONDA_PREFIX}/include/nvstrings"
+                "$ENV{CONDA_PREFIX}/include")
+
+find_library(NVSTRINGS_LIBRARY "NVStrings"
+             HINTS "$ENV{NVSTRINGS_ROOT}/lib"
+                   "$ENV{CONDA_PREFIX}/lib")
+
+find_library(NVCATEGORY_LIBRARY "NVCategory"
```
seunghwak: Are we using NVCATEGORY_LIBRARY AND NVTEXT_LIBRARY? I may have missed but just double check these are actually used.

ChuckHastings: Not yet.  I copied this code from another project.  Brad and I discussed NVCATEGORY being next (although we might be able to do them together), not sure whether NVTEXT is appropriate - I'm unfamiliar with that part of cudf.

cpp/src/converters/renumber.cu
ChuckHastings: Agreed.  One of the intermediate forms of this code made more full use of this, now it is reduced to a typing saver... I will fix this in my next push.

cpp/src/converters/renumber.cuh
ChuckHastings: I will check and address in my next push.  It strikes me it needs to be in the thrust namespace because it is overloading a templated function that is defined in the thrust namespace.  But I will see if I can put it somewhere else, my last exploration of swap was back a few iterations of the C++ standard.

ChuckHastings: I can't get any other mechanism to compile, although if you know another way to override the thrust::swap method (the default method generates an incorrect answer) I'd be happy to change this.

BradReesWork: The issues has been marked as blocked and moved to release 0.10, so it should *NOT* be closed.

BradReesWork:
The issues has been marked as blocked and moved to release 0.11, so it should NOT be closed.

PR #333, opened 2019-06-18
[Review] Various general improvements at the library level

### LINKED ISSUE #273: [FEA] Support null values
Null values in cudf::Series/gdf_columns are not supported currently in cuGraph. Support for null values should be added in the future

Options:
1. Check for nulls in cudf::Series edge list input and create a copy without null values. There might be a function for that un cuDF. That should happen before renumbering as well. This is faster to implement but will likely double the memory requirements.
2. Support null values natively in any algorithms using the edge list. That requires checking the null mask each time an element is accessed. In this case, we cannot rely on the size of the column for the number of edges anymore so we may consider changing the API.

### LINKED ISSUE #304: [FEA] Add __version__ attribute
Hi!

I have tried to double check current cugraph version in my container by running the following code:

```
import cugraph
print(cugraph.__version__)
```

It returns the following error:

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-11-39ef4ed12d29> in <module>
      1 import cugraph
      2 
----> 3 print(cugraph.__version__)

AttributeError: module 'cugraph' has no attribute '__version__'
```

cudf, cuml, dask_cudf, dask_cudf already implement '\_\_version\_\_' attribute.

It would be great if cugraph could also implement it.

Hope it helps!

Regards,
Miguel

### LINKED ISSUE #329: [BUG] [DOC] Jaccard doc is not being generated
The code looks ok, further investigation is needed.
I also noticed that the links on the left navigation panel are not working for most algos.

https://docs.rapids.ai/api/cugraph/nightly/

afender:
Close #273 by returning errors for null masks at the python level
Close #304 by adding files from versioneer.py
Close #329 by fixing doc syntax (should be verified on the actual website after it is merged)

#231: improved input checking and error handling by looking for unsupported index values in the input edge list. Improved doc and added usage warnings. Error handling should be improved further as part of the refactoring.

Fixed test warnings


cpp/src/converters/COOtoCSR.cuh
```diff
@@ -98,7 +98,15 @@ gdf_error ConvertCOOtoCSR(T* sources, T* destinations, int64_t nnz, CSR_Result<T
     CUDA_TRY(cudaMemcpy(&maxId2, maxId_it, sizeof(T), cudaMemcpyDefault));
     maxId = maxId > maxId2 ? maxId : maxId2;
     result.size = maxId + 1;
-
+    // Sending a warning rather than an error here as this may be intended and suported.
+    if (result.size > nnz ) {
+        std::cerr<< "WARNING: there are more vertices than edges in the graph ";
+        std::cerr<< ": V=" << result.size <<", E="<<nnz <<". ";
+        std::cerr<< "Sometime this is not intended and may cause performace and stability issues. ";
+        std::cerr<< "Vertex identifieres must be in the range 0-V where V is the number of vertices. ";
```
seunghwak: 0-V can be confusing (are 0 and V included in the range?), and I guess what you actually meant is 0, 1, ... , V-1.

[0, V) or 0, ... , V-1 might be clearer?

cpp/src/structure/cugraph.cu
```diff
@@ -71,19 +71,18 @@ void cpy_column_view(const gdf_column *in, gdf_column *out) {
 }
 
 gdf_error gdf_adj_list_view(gdf_graph *graph, const gdf_column *offsets,
-                            const gdf_column *indices,
+                            const gdf_column *indices,
```
seunghwak: Just for curiosity, why did you add an additional trailing whitespace here? Is this intended?

cpp/src/utilities/validation.cuh
```diff
@@ -0,0 +1,59 @@
...
+    if (minId2 < 0 )
+        return GDF_INVALID_API_CALL; 
+
+    minId = minId < minId2 ? minId : minId2;
+
+    // warning when smallest vertex is not 0
+    if (minId > 0 ) {
+        std::cerr<< "WARNING: the smallest vertex index in the edge set is "<<minId<<". ";
+        std::cerr<< "Cugraph supports 0-based indexing. ";
+        std::cerr<< "Vertex 0 to "<<  minId <<" will be created." << std::endl;
```
seunghwak: I cannot understand what "Vertex 0 to minId will be created" actually means. Can this warning be little more specific?

cpp/src/converters/COOtoCSR.cuh
ChuckHastings: I'd vote for [0, V), it's compact and precise - but I was trained as a mathematician.

cpp/src/utilities/validation.cuh
ChuckHastings: Perhaps another place where [0, minId) would be appropriate.  Or is this [0, minId]?

cpp/src/converters/COOtoCSR.cuh
afender: Yes, plus we already use [0, V) elsewhere in the doc. I'll fix that.

cpp/src/structure/cugraph.cu
afender: Not intended

cpp/src/utilities/validation.cuh
afender: @seunghwak here are more details : 
The smallest vertex is always 0 in cugraph. When 0 does not have any neighbor (ie. 0 cannot be found in src or dst) we will show a warning saying that vertex [0, <smallest ID found in the edge list>) still exist. In most cases, that's expected. However, some user may be using a 1-based indexing or random indexes starting with several digits. This warning will help them identify that they should use the renumbering feature prior to loading the edge list.

I updated the comment to make this clearer while trying to keeping it concise.

cpp/src/converters/COOtoCSR.cuh
afender: done

cpp/src/structure/cugraph.cu
afender: updated

afender:
@ChuckHastings I cannot comment or click resolve on your suggested items. Not sure why. I addressed it in the last commit.

PR #338, opened 2019-06-20
[REVIEW] Integrating Hornet Katz Centrality Part1

seunghwak:
This PR brings Hornet libraries and header files to the thirdparty/hornet directory. This PR is to separate mere copies from Hornet from actual changes (which will appear in the next PR).

And this is a temporary solution to integrate Hornet to cuGraph. We plan to discuss how to properly integrate Hornet to cuGraph (and how to maintain Hornet) in the future.


afender:
Some binary files were added. I suspect this is not intended.

ogreen:
I think that integrating via submodule will be much simpler than doing a "full" copy/integration.

seunghwak:
> Just had a quick look but why do we copy hornet instead of adding it as a submodule? I suspect we want to edit things in Hornet but wouldn't that be better to add them directly in Hornet repo? Since both repos are being developed I worry that we will hit a lot of integration issues in the long run.

The main reason is to avoid committing temporary solutions to the Hornet main repo. We have not decided how to maintain Hornet and before that gets decided, I don't want to push temporary changes to the Hornet main repo.

seunghwak:
> Just had a quick look but why do we copy hornet instead of adding it as a submodule? I suspect we want to edit things in Hornet but wouldn't that be better to add them directly in Hornet repo? Since both repos are being developed I worry that we will hit a lot of integration issues in the long run.

And also note that this solution will only be short-lived. Brad told me that we will discuss how to integrate Hornet later this month once Oded is back.

seunghwak:
> Some binary files were added. I suspect this is not intended.

libhornet.a and libhornetAlg.a are intended. These are built outside cuGraph (with different CMake configurations, e.g. set_target_properties(hornetAlg PROPERTIES POSITION_INDEPENDENT_CODE ON), I am not sure whether these changes should be pushed to Hornet or not as we are not sure how we will integrate/maintain Hornet). Few files under thirdparty/hornet/include/externals-cpp-btree-master may be unnecessary, but this is an external module even for hornet, so I thought I'd better just bring this module as a whole rather than decide which one to include (e.g. COPYING files stating copyright) and which one to not.

seunghwak:
Closed this PR based on our discussion today, we will wait till Hornet becomes ready for proper integration.

PR #354, opened 2019-06-26
[REVIEW] Bug cugraph debug mode

### LINKED ISSUE #337: [BUG] cugraph fails to compile in debug mode
**Describe the bug**
If we run cmake with -DCMAKE_BUILD_TYPE=Debug,

cugraph fails to compile with the following error message (compiler: CUDA10, gcc 7.4.0)

```
/usr/local/cuda-10.0/bin/nvcc  -Dcugraph_EXPORTS -I/usr/local/cuda-10.0/include -I/home/seunghwak/Program/anaconda3/envs/cuda10/include -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/cub -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/hornet/include/hornet -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/hornet/include/hornetsnest -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/hornet/include/primitives -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/hornet/include/externals/cpp-btree-master -I/home/seunghwak/RAPIDS/cugraph/cpp/../thirdparty/hornet/include/xlib -I/home/seunghwak/RAPIDS/cugraph/cpp/src -I/home/seunghwak/RAPIDS/cugraph/cpp/build/gunrock -I/home/seunghwak/RAPIDS/cugraph/cpp/build/gunrock/externals/moderngpu/include -I/home/seunghwak/RAPIDS/cugraph/cpp/include  -gencode=arch=compute_60,code=sm_60 -gencode=arch=compute_70,code=sm_70 -gencode=arch=compute_70,code=compute_70 --expt-extended-lambda -Werror cross-execution-space-call -Wno-deprecated-declarations -Xptxas --disable-warnings -Xcompiler -Wall,-Wno-error=sign-compare,-Wno-error=unused-but-set-variable -DCUB_WRAPPER -DRMM_WRAPPER -G -Xcompiler=-fopenmp -Xcompiler=-fPIC   -std=c++14 -x cu -c /home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cu -o CMakeFiles/cugraph.dir/src/converters/renumber.cu.o
/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(240): error: calling a __host__ function("std::pair<const char *, unsigned long> ::operator =") from a __device__ function(" const") is not allowed

/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(240): error: identifier "std::pair<const char *, unsigned long> ::operator =" is undefined in device code

/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(248): error: calling a __host__ function("std::pair<const char *, unsigned long> ::operator =") from a __device__ function(" const") is not allowed

/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(248): error: identifier "std::pair<const char *, unsigned long> ::operator =" is undefined in device code

/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(339): error: calling a __host__ function("std::pair<const char *, unsigned long> ::operator =") from a __device__ function(" const") is not allowed

/home/seunghwak/RAPIDS/cugraph/cpp/src/converters/renumber.cuh(339): error: identifier "std::pair<const char *, unsigned long> ::operator =" is undefined in device code

6 errors detected in the compilation of "/tmp/tmpxft_0000365e_00000000-7_renumber.compute_70.cpp1.ii".
```

ChuckHastings:
close #337 

This is a hack.  Best solution I could find.  Here is the fundamental problem:

1. nvstrings method create_index REQUIRES that you use std::pair
2. std::pair does not understand the concept of a device or host function, so the compiler doesn't build device versions of the functions
3. thrust::pair - which is structurally the same as std::pair (they are derived from the same original source) understand all of the device/host methods - so thrust::pair does everything that we need - except there is no nvstrings method that uses thrust::pair
4. I believe that when compiled not in debug mode the function is inlined and the fact that it is not a device function is irrelevant - but in debug mode it needs to actually call the function.

This solution uses the old c-style cast to convert the thrust::pair to an std::pair for the nvstrings calls.  This is a very fragile solution - if either std::pair or thrust::pair are changed this could break... but at the moment it will work.

Once cudf and nvstrings have resolved the using string columns I imagine this will be resolved in some way and we can rework this code.

One side effect - by using thrust::pair allows me to delete the overriding of thrust::swap - as thrust::swap already has an implementation for thrust::pair.


ChuckHastings:
rerun tests

BradReesWork:
Please updated to branch-0.9

PR #353, opened 2019-06-25
[REVIEW] Snmg python wrapper edit

Iroy30:

PR #357, opened 2019-06-27
[REVIEW] Pagerank cpp

afender:
- Connected all cpp building blocks
- Added Pagerank C/C++ API
- Upgraded CPP tests to run the coo2csr+pagerank
- API documentation
- fix to enable peer device access only once for the whole pipeline
- removed namespace in API bindings
- cuda-memcheck passed

Note
2 data sets trigger intermittent issues in coo2csr. Need further investigation. The same datasets pass when running only coo2csr tests and in the CSR PageRank test. It also passes if this is the first test to run in the edge list PageRank test. Does not repro on one GPU. This should not block testing the python bindings.


afender:
Waiting on #353 to be merged before fixing this one

Iroy30:
Looks good to me

PR #358, opened 2019-06-28
[REVIEW] Remove boost from meta.yaml & linking

raydouglass:
cuGraph does not need to link against boost. Doing so has caused some strange build issues in the conda packages.

### DIFF
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -28,12 +28,10 @@ requirements:
     - libcudf=0.8*
     - cython
     - cudatoolkit {{ cuda_version }}.*
-    - boost
   run:
     - libcudf=0.8*
     - cython
     - {{ pin_compatible('cudatoolkit', max_pin='x.x') }}
-    - boost
 
 #test:
 #  commands:
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -28,10 +28,12 @@ requirements:
     - libcudf=0.8*
     - cython
     - cudatoolkit {{ cuda_version }}.*
+    - boost
   run:
     - libcudf=0.8*
     - cython
     - {{ pin_compatible('cudatoolkit', max_pin='x.x') }}
+    - boost
 
 #test:
 #  commands:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# cuGraph 0.8.1 (28 June 2019)
+
+## Bug Fixes
+- PR #358 Remove boost linking dependency
+
+
 # cuGraph 0.8.0 (27 June 2019)
 
 ## New Features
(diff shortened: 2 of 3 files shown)


afender:
Just scanned the slack discussion, it seems @rlratzel and @seunghwak were able to build without it so perhaps this is just a CI / testing issue?

In the eventuality boost is really needed for CI to pass, I doubt cuGraph needs a specific version. Any version should work.

conda/recipes/libcugraph/meta.yaml
```diff
@@ -28,12 +28,10 @@ requirements:
     - libcudf=0.8*
     - cython
     - cudatoolkit {{ cuda_version }}.*
-    - boost
```
rlratzel: Do we still need it for build to ensure the needed headers are present, or are you relying on boost being installed at the system level?

afender: my local build seems to get boost's headers from the conda env

```
/cpp/build/CMakeFiles/cugraph.dir/depend.internal:18: /home/afender/anaconda3/envs/cugraph_dev/include/boost/config.hpp
...
```

BradReesWork:
What is the proposed solution?  Should be kill the PR and quickly redo the code or define a version of boost to get CI and docker to build without error?

seunghwak:
https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/gpuCI%2Fcugraph%2Fprb%2Fcugraph-cpu-build/detail/cugraph-cpu-build/1360/pipeline

../../libcugraph.so: undefined reference to `boost::system::system_category()'

../../libcugraph.so: undefined reference to `boost::system::generic_category()'

collect2: error: ld returned 1 exit status

make[2]: *** [src/tests/CMakeFiles/LOUVAIN_TEST.dir/build.make:102: gtests/LOUVAIN_TEST] Error 1

make[1]: *** [CMakeFiles/Makefile2:1083: src/tests/CMakeFiles/LOUVAIN_TEST.dir/all] Error 2

So it seems like boost library needs to be linked in some system configurations.

seunghwak:
https://stackoverflow.com/questions/9723793/undefined-reference-to-boostsystemsystem-category-when-compiling

See the second answer with 50 votes. It seems like this is a sort of bug for boost before 1.66

rlratzel:
> https://stackoverflow.com/questions/9723793/undefined-reference-to-boostsystemsystem-category-when-compiling
> 
> See the second answer with 50 votes. It seems like this is a sort of bug for boost before 1.66

So if I'm understanding it right, we should be able to add a `build` dependency on `boost>=1.66` and avoid having to link against the boost lib.

seunghwak:
> > https://stackoverflow.com/questions/9723793/undefined-reference-to-boostsystemsystem-category-when-compiling
> > See the second answer with 50 votes. It seems like this is a sort of bug for boost before 1.66
> 
> So if I'm understanding it right, we should be able to add a `build` dependency on `boost>=1.66` and avoid having to link against the boost lib.

I think that should work (in theory). Another option might be adding -DBOOST_SYSTEM_NO_DEPRECATED. Not sure which one is better.

rlratzel:
I worked with @raydouglass in parallel on a different solution and he asked me to open PR #359 after I saw it build and pass tests on a local CI instance as well as on a `runtime` container that was previously failing.

PR #361, opened 2019-07-01
[REVIEW] Define the prototype of what the SNMG sort would look like

ChuckHastings:
This PR defines sample SNMG sort functions.  Once approved I will fill in the details from the cusort repository.

### DIFF
--- a/cpp/src/sort/sort.cuh
+++ b/cpp/src/sort/sort.cuh
@@ -10,60 +10,62 @@ namespace cusort {
    * @brief  Sort key value pairs distributed across multiple GPUs
    *
    *   This sort function takes arrays of keys and values distributed
-   *   around multiple GPUs 
+   *   around multiple GPUs, redistributes them so that GPU 0 contains
+   *   the smallest elements, GPU 1 the next smallest elements, etc.
    *
-   * @param[in]   input_keys_d     The unsorted keys, stored in device arrays.
+   * @param[in]   d_input_keys     The unsorted keys, stored in device arrays.
    *                               input_keys_d[i] is the array of keys on GPU i
-   * @param[in]   input_values_d   The unsorted values, stored in device arrays.
+   * @param[in]   d_input_values   The unsorted values, stored in device arrays.
    *                               input_values_d[i] is the array of values on GPU i
-   * @param[in]   input_length_h   Host array containing the number of elements on
+   * @param[in]   h_input_length   Host array containing the number of elements on
    *                               each GPU in the input key/value arrays.
-   * @param[out]  output_keys_d    The sorted keys, stored in device arrays.
+   * @param[out]  d_output_keys    The sorted keys, stored in device arrays.
    *                               output_keys_d[i] is the array of keys on GPU i
-   * @param[out]  output_values_d  The sorted values, stored in device arrays.
+   * @param[out]  d_output_values  The sorted values, stored in device arrays.
    *                               output_values_d[i] is the array of values on GPU i
-   * @param[out]  output_length_h  Host array containing the number of elements on
+   * @param[out]  h_output_length  Host array containing the number of elements on
    *                               each GPU in the output key/value arrays.
    * @param[in]   num_gpus         The number of GPUs
    *
    * @return   GDF_SUCCESS upon successful completion
    */
   template <typename Key_t, typename Value_t, typename Length_t>
-  gdf_error sortKeyValue(Key_t **input_keys_d,
-                         Value_t **input_values_d,
-                         Length_t *input_length_h,
-                         Key_t **output_keys_d,
-                         Value_t **output_values_d,
-                         Length_t *output_length_h,
-                         int num_gpus) {
+  gdf_error sort_key_value(Key_t **d_input_keys,
+                           Value_t **d_input_values,
+                           Length_t *h_input_length,
+                           Key_t **d_output_keys,
+                           Value_t **d_output_values,
+                           Length_t *h_output_length,
+                           int num_gpus) {
 
     return GDF_SUCCESS;
   }
   
   /**
-   * @brief  Sort key value pairs distributed across multiple GPUs
+   * @brief  Sort keys distributed across multiple GPUs
    *
-   *   This sort function takes arrays of keys and values distributed
-   *   around multiple GPUs 
+   *   This sort function takes an array of keys distributed
+   *   around multiple GPUs, redistributes them so that GPU 0 contains
+   *   the smallest elements, GPU 1 the next smallest elements, etc.
    *
-   * @param[in]   input_keys_d     The unsorted keys, stored in device arrays.
+   * @param[in]   d_input_keys     The unsorted keys, stored in device arrays.
    *                               input_keys_d[i] is the array of keys on GPU i
-   * @param[in]   input_length_h   Host array containing the number of elements on
+   * @param[in]   h_input_length   Host array containing the number of elements on
    *                               each GPU in the input key/value arrays.
-   * @param[out]  output_keys_d    The sorted keys, stored in device arrays.
+   * @param[out]  d_output_keys    The sorted keys, stored in device arrays.
    *                               output_keys_d[i] is the array of keys on GPU i
-   * @param[out]  output_length_h  Host array containing the number of elements on
+   * @param[out]  h_output_length  Host array containing the number of elements on
    *                               each GPU in the output key/value arrays.
    * @param[in]   num_gpus         The number of GPUs
    *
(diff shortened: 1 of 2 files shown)


cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+
+    return GDF_SUCCESS;
+  }
+  
+  /**
+   * @brief  Sort key value pairs distributed across multiple GPUs
+   *
+   *   This sort function takes arrays of keys and values distributed
+   *   around multiple GPUs 
+   *
```
seunghwak: Is this correct? Does this function sort key-value pairs? or just keys?

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+   * @param[in]   input_length_h   Host array containing the number of elements on
+   *                               each GPU in the input key/value arrays.
+   * @param[out]  output_keys_d    The sorted keys, stored in device arrays.
+   *                               output_keys_d[i] is the array of keys on GPU i
+   * @param[out]  output_values_d  The sorted values, stored in device arrays.
+   *                               output_values_d[i] is the array of values on GPU i
+   * @param[out]  output_length_h  Host array containing the number of elements on
+   *                               each GPU in the output key/value arrays.
+   * @param[in]   num_gpus         The number of GPUs
+   *
```
seunghwak: AFAIK, using h_ for host variables and d_ for device variables is more common than using _h and _d.

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
+// -*-c++-*-
+
+#pragma once
+
+#include <cudf/types.h>
+
+namespace cusort {
```
seunghwak: If we are initially placing these functions under cugraph, shouldn't we put this in the cugraph namespace?

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+   *                               output_values_d[i] is the array of values on GPU i
+   * @param[out]  output_length_h  Host array containing the number of elements on
+   *                               each GPU in the output key/value arrays.
+   * @param[in]   num_gpus         The number of GPUs
+   *
+   * @return   GDF_SUCCESS upon successful completion
+   */
+  template <typename Key_t, typename Value_t, typename Length_t>
+  gdf_error sortKeyValue(Key_t **input_keys_d,
+                         Value_t **input_values_d,
```
seunghwak: And aren't we (mostly) using _ instead of camel case in function names? (e.g. sort_key_value instead of sortKeyValue?)

cpp/src/sort/sort.cuh
ChuckHastings: I imagined that putting them in the cusort namespace would reduce the number of code changes later when we move it, but I'm find putting it in either cugraph or cusort.

cpp/src/sort/sort.cuh
ChuckHastings: OK, I'll change these.

cpp/src/sort/sort.cuh
ChuckHastings: I always get confused since we haven't finalized a naming standard.  I'll switch to _

cpp/src/sort/sort.cuh
ChuckHastings: Copy paste error, I'll fix it.

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+#pragma once
+
+#include <cudf/types.h>
+
+namespace cusort {
+
+  /**
+   * @brief  Sort key value pairs distributed across multiple GPUs
+   *
+   *   This sort function takes arrays of keys and values distributed
```
afender: We could add more documentation, for instance :
- if this is expected to be called inside a parallel section or by a single thread
- are there any assumptions such as the Cuda device being is already set to a specific id ?
- does this work only if the gpus are fully connected (ie each device must enable cuda access to all peers)?

I would recommend addressing this in the description so that users can double check this before filling issues.

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+   *
+   * @param[in]   input_keys_d     The unsorted keys, stored in device arrays.
+   *                               input_keys_d[i] is the array of keys on GPU i
+   * @param[in]   input_values_d   The unsorted values, stored in device arrays.
+   *                               input_values_d[i] is the array of values on GPU i
+   * @param[in]   input_length_h   Host array containing the number of elements on
+   *                               each GPU in the input key/value arrays.
+   * @param[out]  output_keys_d    The sorted keys, stored in device arrays.
+   *                               output_keys_d[i] is the array of keys on GPU i
+   * @param[out]  output_values_d  The sorted values, stored in device arrays.
```
afender: I suppose the output is allocated by this function. I would recommend adding a note mentioning that.
Also, I recall that one key cannot be present on two GPU in the output. This is a feature that worth documenting I think.

cpp/src/sort/sort.cuh
afender: Depends on the intent. In cugraph only internal functions are in cugraph namespace. The public API is not encapsulated in a namespace. It is a C API legacy thing. Adding it under cugraph namespace really indicates it is a cugraph internal function.

cpp/src/sort/sort.cuh
```diff
@@ -0,0 +1,70 @@
...
+   *                               output_values_d[i] is the array of values on GPU i
+   * @param[out]  output_length_h  Host array containing the number of elements on
+   *                               each GPU in the output key/value arrays.
+   * @param[in]   num_gpus         The number of GPUs
+   *
+   * @return   GDF_SUCCESS upon successful completion
+   */
+  template <typename Key_t, typename Value_t, typename Length_t>
```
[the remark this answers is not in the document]
ChuckHastings: We could hard-code size_t.  But that would unnecessarily restrict the caller.  If they had an array of int32_t that contained the sizes they would need to create an array of size_t and copy the data.  I templated it to avoid that extra work.

The renumber function is also templated.  When I integrated it into a python binding I simply passed the correct arguments into the function call.

cpp/src/sort/sort.cuh
seunghwak: OK, if we are planning to move to the C++ API, and anything here is temporary, I have no complaint.

cpp/src/sort/sort.cuh
ChuckHastings: I will add some additional documentation tomorrow - but to answer your questions:
1. This is expected to be called by a single thread.  It uses openmp to start threads on each GPU
1. No assumptions on the cuda device, the openmp parallel region sets the device
1. The assumption here is SNMG.  We do cudaMemcpy calls (device to device) to transfer data between GPUs.  The function assumes that the peer access has already been enabled.

cpp/src/sort/sort.cuh
ChuckHastings: Will add to documentation tomorrow.

cpp/src/sort/sort.cuh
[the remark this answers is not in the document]
ogreen: For the initial use case (Coo2Csr), we probably don't need templating as much as we have the exact data types. In the longer term, we may want to change the key and value types.

cpp/src/sort/sort.cuh
ogreen: There is no requirements that the GPUs be fully connected for the algorithm to work. However, performance will probably go down if they are not fully connected.

ChuckHastings:
@ogreen - agreed.  I left those parameter out, expecting to add them as optional parameters later when we need them.

cpp/src/sort/sort.cuh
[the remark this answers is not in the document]
afender: Sounds good

cpp/src/sort/sort.cuh
afender: Ah yes I think mentioning that peer access must be enabled by the user before calling this function is very important.

cpp/src/sort/sort.cuh
BradReesWork: Since the long-term plan is to move sorting into RAFT (ml-graph primitives repo) the namespace might need to change.  As Alex's pointed out, since it is not going to be a cugrapn internal function, maybe 'cugraph' is not the right namespace.  I'm okay leaving it as 'cusort' for now.

PR #366, opened 2019-07-03
In out degree

### LINKED ISSUE #363: Add function to graph object which returns both in and out degree [FEA]
Add the function compute_degree as a member method of the graph object. It should take no arguments, and should return a dataframe with three columns: 'vertex', 'in_degree', and 'out_degree'.

jwyles:
Closes #363

### DIFF
--- a/python/cugraph/graph/c_graph.pyx
+++ b/python/cugraph/graph/c_graph.pyx
@@ -20,7 +20,6 @@ import cudf
 from librmm_cffi import librmm as rmm
 
 from c_graph cimport *
-from pyasn1.type.char import GraphicString
 
 
 dtypes = {np.int32: GDF_INT32, np.int64: GDF_INT64, np.float32: GDF_FLOAT32, np.float64: GDF_FLOAT64}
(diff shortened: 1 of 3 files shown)


python/cugraph/graph/c_graph.pyx
```diff
@@ -795,3 +795,72 @@ class Graph:
             del degree_col
 
         return df
+
+    def compute_degree(self, vertex_subset=None):
```
afender: Would "in_out_degree" or "degrees" be more consistent than "compute_degree" when comparing with the existing "in_degree", "out_degree", "degree" functions?

afender:
I recall that @BradReesWork mentioned a performance problem with previous degree functions. 
@jwyles were you able to check the performance of this new feature?

python/cugraph/graph/c_graph.pyx
BradReesWork: @Alex you raise an interesting question.  Spark GraphX just has a structure that is created when a Graph is created with inDegreem outDegree, and degree - the use doesn't have to explicitly call a functions.   NetworkX also has separate in and out calls - which is what James currently has.  This function just get in and out in one call.  
I like "degrees" just because I don't think we need the "compute_" part

jwyles: I like calling the function 'degrees' as well. If there aren't any objections I will go ahead and change it to that.

PR #375, opened 2019-07-09
[REVIEW] Added a function to initialize gdf columns in cugraph

afender:
This fixes 
```ERROR: RMM runtime call  RMM_FREE( (col->valid), (stream) )invalid device pointer```
Initiated by dask-cugraph https://github.com/rapidsai/dask-cugraph/issues/23 after the first fix https://github.com/rapidsai/dask-cugraph/pull/24


afender:
The problem was that col->valid was not automatically set to null as gdf columns are C structures.
That resulted in a wrong evaluation of `if (col->valid != nullptr)` and an attempt to free a null pointer

Iroy30:
looks good to me

PR #380, opened 2019-07-15
[REVIEW] Fea cusort merge

### LINKED ISSUE #295: [FEA] Integrate snmg sort in cugraph
- [ ] The content of https://github.com/ogreen/cugraph/blob/fea-ext-cusort/cpp/include/cusort/sort.cuh should be added to `cugraph/cpp/src/`

- [ ] Tests should be implemented in `cugraph/cpp/src/test`. They should compare against the output of the existing naive C++ implementation for a couple of datasets.

ChuckHastings:
Base implementation of cusort.  Implements SNMG sorting of key/value pairs.

Does not include a capability for sorting keys only, that will be added as a separate feature.

close #295


ogreen:
All the 
  unsigned long long  h_input_offsets[3];
  unsigned long long  h_output_offsets[3];
Should be
  unsigned long long  h_input_offsets[MAX_NUM_GPUS+1];
  unsigned long long  h_output_offsets[MAX_NUM_GPUS+1];

cpp/src/tests/sort/sort_test.cu
```diff
@@ -0,0 +1,443 @@
...
+
+TEST_F(SortTest, Random100MPerDevice_uint64_t)
+{
+  cudaStream_t stream{nullptr};
+  
+  uint64_t           *d_input[MAX_NUM_GPUS];
+  uint64_t           *d_input_values[MAX_NUM_GPUS];
+  uint64_t           *d_output[MAX_NUM_GPUS];
+  uint64_t           *d_output_values[MAX_NUM_GPUS];
+  unsigned long long  h_input_offsets[3];
```
ogreen: unsigned long long  h_input_offsets[MAX_NUM_GPUS+1];
  unsigned long long  h_output_offsets[MAX_NUM_GPUS+1];

BradReesWork: why use "unsigned long long" versus "uint64_t" ?

ogreen: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicaddA lot of the class is templated and uses atomicAdds in various locations. unit64_t can cause compile time warnings and errors.
--

BradReesWork: oops, I do remember that about atomics.  good catch

cpp/src/tests/sort/sort_test.cu
```diff
@@ -0,0 +1,443 @@
...
+
+#include "sort/sort.cuh"
+#include "rmm_utils.h"
+#include "test_utils.h"
+
+#include <chrono>
+
+#include <curand_kernel.h>
+
+#define MAX_NUM_GPUS 16
```
seunghwak: I am not an assigned reviewer, but seeing the email thread on this PR, this worries me. Is this future proof? What happens if NVIDIA releases a new system with more than 16 GPUs? (Or if another company custom builds a system with more than 16 GPUs in a single node?) A bug due to hard-coding MAX_NUM_GPUS happened in hornet, and this can happen here as well in the future. I think we should better query the number of GPUs in the target system than hard-coding MAX_NUM_GPUS.

seunghwak: Or we should at least add a check routine comparing the queried number of GPUs and MAX_NUM_GPUS and if # GPUs > MAX_NUM_GPUS, abort the execution with error message to avoid undefined behaviors which can be hard to debug.

cpp/src/sort/sort_impl.cuh
```diff
@@ -0,0 +1,497 @@
...
+    template <typename Key_t>
+    struct CountLeadingZeros<Key_t, 8> {
+      __inline__ __device__ int operator()(Key_t k) {
+        return __clzll(k);
+      }
+    };
+  }
+  
+  template <typename Key_t,typename Value_t, typename Length_t,
+            int MAX_NUM_GPUS = 16, int BIN_SCALE = 16,
```
ogreen: @seunghwak , the cusort class itselft doesn't make any "hardcoded" assumptions. These are runtime configurable.

cpp/src/tests/sort/sort_test.cu
ogreen: The number of GPUs is set a template parameter for the class and can be changed based on needs. This is different than Hornet that was defining it as a macro (which it shouldn't have...)

seunghwak: @ogreen yes, it's a template parameter elsewhere, but in this file, I see "#define MAX_NUM_GPUS 16" (sort_test.cu line 32).

and later (sort_test.cu, from line 217), this can lead to writing passing the array boundary.

```
  for (int i = 0 ; i < n_gpus ; ++i) {
    cudaSetDevice(i);

    EXPECT_EQ(RMM_FREE(d_input[i], stream), RMM_SUCCESS);
    EXPECT_EQ(RMM_FREE(d_output[i], stream), RMM_SUCCESS);
  }
```

cpp/src/sort/sort_impl.cuh
seunghwak: I got this but I am not sure this applies to sort_test.cu line 32

```
#define MAX_NUM_GPUS 16
```

as well.

```
TEST_F(SortTest, Random10MPerDevice_uint64_t)
{
  cudaStream_t stream{nullptr};
  
  uint64_t           *d_input[MAX_NUM_GPUS];<= MAX_NUM_GPUS hard coded to 16
  uint64_t           *d_input_values[MAX_NUM_GPUS];
  uint64_t           *d_output[MAX_NUM_GPUS];
  uint64_t           *d_output_values[MAX_NUM_GPUS];
  unsigned long long  h_input_offsets[3];
  unsigned long long  h_output_offsets[3];

  const int num_elements = 10000000;
  const int seed = 43;
  int n_gpus = 0;

  CUDA_RT_CALL(cudaGetDeviceCount(&n_gpus));<= n_gpus can be larger than 16 in future systems.

  for (int i = 0 ; i < (n_gpus + 1) ; ++i)
    h_input_offsets[i] = i * num_elements;<=h_input_offsets's size is set to three, so isn't this a bug as well?

  omp_set_num_threads(n_gpus);

  generate_random(d_input, d_input_values, h_input_offsets, n_gpus, seed, stream);

  //
  //  Initialize Peer-to-peer communications
  //
  cusort::initialize_snmg_communication(n_gpus);

  // NOTE:  could vary numBins, binScale, useThrust
  cusort::sort_key_value(d_input, d_input_values, h_input_offsets,
                         d_output, d_output_values, h_output_offsets,
                         n_gpus);

  verify_sorted_order(d_output, d_output_values, h_output_offsets, n_gpus, stream, true);

  for (int i = 0 ; i < n_gpus ; ++i) {
    cudaSetDevice(i);

    EXPECT_EQ(RMM_FREE(d_input[i], stream), RMM_SUCCESS);<=d_input[i] may not be valid if n_gpus > MAX_NUM_GPUS
    EXPECT_EQ(RMM_FREE(d_output[i], stream), RMM_SUCCESS);
  }
}
```

See the comments starting with <= in the above code.

cpp/src/tests/sort/sort_test.cu
ogreen: This is just the tester file. We need to define a maximum for testing purpose.

seunghwak: I think using std::vector and resizing std::vector based on n_gpus is a safer coding practice, but this is just a recommendation. But I think we should at least check n_gpus <= MAX_NUM_GPUS to avoid surprising undefined behaviors; and sort_impl.cuh is doing this check as well.

cpp/src/tests/sort/sort_test.cu
ChuckHastings: I tried to minimize the places where unsigned long long was used to only those places that get used by atomics.  It would be nice to have the cuda atomic add fixed to support all of the various spellings of the type names properly.

I will shortly push a fix to the hardcoded 3... that was careless.

cpp/src/tests/sort/sort_test.cu
ChuckHastings: I have added a check in the test to ensure that the number of opus on the system <= MAX_NUM_GPUS.  Will add more commentary below.

cpp/src/sort/sort_impl.cuh
ChuckHastings: See above discussion.

cpp/src/tests/sort/sort_test.cu
ChuckHastings: I agree that using std::vector would be a safer coding practice.  I actually had a different solution in mind - but felt like I had done enough refactoring of the code thus far to go ahead and get something into the code base - so I made MAX_NUM_GPUS a template parameter for now and used the fixed array sizes.

I can add a refactoring of this logic to the backlog and try and get to that later.

ChuckHastings:
rerun tests

PR #395, opened 2019-07-19
[REVIEW] run omp_get_num_threads in a parallel context

### LINKED ISSUE #394: [BUG] SNMG sort corrupting number of OMP threads
A last minute change to PR #380 introduced a bug.

The function omp_get_num_threads is called to compute the number of openmp threads... except it is called in a serial context - which always returns 1.  The function only returns the desired value if called in a parallel context.

ChuckHastings:
close #394


ogreen:
Do you need to get the number of threads also inside an omp section?

ChuckHastings:
@ogreen - yes, that's the change here.

PR #410, opened 2019-07-23
Snmg coo2csr debug

jwyles:
Fixed overflow bug in SNMG COO2CSR


jwyles:
Not sure what the failure in CI is about, it looks like a timeout of some kind. Could we try running it again?

afender:
rerun tests

PR #416, opened 2019-07-26
[REVIEW] Added initial implementation of db_object class

jwyles:
Added the initial implementation of the db_object class along with associated lesser classes.


cpp/src/db/db_object.cu
```diff
@@ -0,0 +1,327 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cugraph.h>
+#include <rmm_utils.h>
+#include <db/db_object.cuh>
+#include <cub/device/device_run_length_encode.cuh>
+
```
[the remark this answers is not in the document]
jwyles: No reason. I will move it inside the namespace.

cpp/src/db/db_object.cu
```diff
@@ -0,0 +1,327 @@
...
+  void db_table<idx_t>::addEntry(db_pattern<idx_t>& pattern) {
+    if (!pattern.isAllConstants())
+      throw new std::invalid_argument("Can't add an entry that isn't all constants");
+    if (!pattern.getSize() != columns.size())
+      throw new std::invalid_argument("Can't add an entry that isn't the right size");
+    inputBuffer.push_back(pattern);
+  }
+
+  template<typename idx_t>
+  void db_table<idx_t>::rebuildIndices() {
```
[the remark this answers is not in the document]
jwyles: Okay.

cpp/src/db/db_object.cu
```diff
@@ -0,0 +1,327 @@
...
+      thrust::exclusive_scan(rmm::exec_policy(nullptr)->on(nullptr),
+                             offsets,
+                             offsets + maxId + 2,
+                             offsets);
+      ALLOC_FREE_TRY(tempColumn, nullptr);
+      ALLOC_FREE_TRY(unique, nullptr);
+      ALLOC_FREE_TRY(counts, nullptr);
+      ALLOC_FREE_TRY(runCount, nullptr);
+
+      // Assign new offsets array and indirection vector to index
```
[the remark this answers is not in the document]
jwyles: How would you use gdf_column_view in this context. I am all for making this part of things simpler if there are existing methods to do it.

cpp/src/db/db_object.cuh
```diff
@@ -0,0 +1,92 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cugraph.h>
+#include <vector>
+#include <map>
+#include "utilities/graph_utils.cuh"
+
```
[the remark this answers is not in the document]
jwyles: Will add in #pragma once.

cpp/src/db/db_object.cuh
```diff
@@ -0,0 +1,92 @@
...
+namespace cugraph {
+  template <typename idx_t>
```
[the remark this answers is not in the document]
jwyles: Will do.

cpp/src/db/db_object.cu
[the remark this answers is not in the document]
afender: It seems to me that 
```      gdf_col_set_defaults(offsetsCol);
      offsetsCol->data = offsets;
      offsetsCol->size = maxId + 2;
      if (std::is_same<idx_t, int32_t>::value)
        offsetsCol->dtype = GDF_INT32;
      if (std::is_same<idx_t, int64_t>::value)
        offsetsCol->dtype = GDF_INT64;
```
could be replaced by something like
```
gdf_col_set_defaults(offsetsCol);  // perhaps this is not needed when using gdf_column_view?
gdf_column_view(offsetsCol, offsets, nullptr, maxId + 2,
                std::is_same<idx_t, int32_t>::value?GDF_INT32:GDF_INT64); 
```

what do you think?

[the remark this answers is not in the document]
jwyles: I have updated.

cpp/src/db/db_object.cuh
```diff
@@ -0,0 +1,123 @@
...
+    std::vector<db_pattern_entry<idx_t>> entries;
+  public:
+    db_pattern();
+    db_pattern(const db_pattern<idx_t>& other);
+    int getSize() const;
+    const db_pattern_entry<idx_t>& getEntry(int position) const;
+    void addEntry(db_pattern_entry<idx_t>& entry);
+    bool isAllConstants();
+  };
+
```
seunghwak: Before jumping to the more detailed review, I have a question, and how do this db_pattern & db_pattern_entry classes map to cypher queries?

Say for example, in the following query,

MATCH (:Person {name: 'Jennifer'})-[:WORKS_FOR]->(company:Company)
RETURN company

I guess this has three steps: 1) find the vertex with a label set to person and the name property set to 'Jennifer', 2) investigate the edges of the found vertex to find the edge with its label set to 'WORKS_FOR', and 3) traverse the edge to reach the company vertex.

This query is stored in a db_pattern object and each step becomes a db_pattern_entry? Or am I mis-guessing? Any corrections?

jwyles: You are mostly right. In your query we would first find all vertices with label:Person, then from those vertices find ones that have name:'Jennifer', then from those vertices find edges with type:WORKS_FOR, then from the ending vertices of the last find vertices with label:Company and return them. Each step of the query would be represented by a db_pattern object.

seunghwak: Got it, so each step will be mapped to a db_pattern object (not a db_pattern_entry). I think it will be greatly helpful to add comments explaining this mapping in code readability.

cpp/src/db/db_object.cuh
```diff
@@ -0,0 +1,123 @@
...
+namespace cugraph {
+  /**
+   * Class for representing an entry in a pattern, which may either be a variable or constant value
+   */
+  template <typename idx_t>
+  class db_pattern_entry {
+    bool is_var;
+    idx_t constantValue;
+    std::string variableName;
+  public:
```
seunghwak: So here, idx_t constantValue is used for vertex indices and std::string variableName is used for label/property names we want to match? Adding comments explaining how this maps to cypher queries will be really helpful in understanding this code.

cpp/src/db/db_object.cu
```diff
@@ -0,0 +1,328 @@
...
+  void db_pattern<idx_t>::addEntry(db_pattern_entry<idx_t>& entry) {
+    entries.push_back(entry);
+  }
+
+  template<typename idx_t>
+  bool db_pattern<idx_t>::isAllConstants() {
+    bool is_all_constants = true;
+    for (size_t i = 0; i < entries.size(); i++)
+      if (entries[i].isVariable())
+        is_all_constants = false;
```
seunghwak: you can just return false at this point without finishing the loop.

cpp/src/db/db_object.cu
```diff
@@ -0,0 +1,328 @@
...
+      cudaMemcpy(&runCount_h, runCount, sizeof(idx_t), cudaMemcpyDefault);
+      idx_t* offsets;
+
+      // Allocating the new offsets array
+      ALLOC_TRY(&offsets, (maxId + 2) * sizeof(idx_t), nullptr);
+
+      // Filling values in offsets array from the encoded run lengths
+      int threadsPerBlock = 1024;
+      int numBlocks = (runCount_h + threadsPerBlock - 1) / threadsPerBlock;
+      offsetsKernel<<<numBlocks, threadsPerBlock>>>(runCount_h, unique, counts, offsets);
```
seunghwak: May consider using thrust::permutation_iterator (as this code is mostly composed of thrust and cub calls).

cpp/src/db/db_object.cuh
jwyles: No, in a pattern you have entries that are either a variable or a constant. In the case you are referring to the label or property names would be constants and represented by an idx_t (since all constants are dictionary-encoded to ids before getting stored in the table.

jwyles: I have added some more comments explaining this further.

cpp/src/db/db_object.cu
jwyles: It doesn't seem like using thrust::permutation_iterator will simplify things much. I will leave this as is.

cpp/src/db/db_object.cu
jwyles: Will make the change.

PR #420, opened 2019-07-29
[REVIEW] snmg drop duplicates, add conditional read_csv wait in pagerank.py

### LINKED ISSUE #406: [BUG] [MG] Duplicates in hibench files
We found that HiBench files contain duplicates.
Nx removes these duplicates and runs with fewer edges while cugraph does not. This causes 2% of the result to exceed the tolerance threshold when comparing the outputs.

Proposed solution: remove duplicates in the test
Dask cudf does not provide the drop_duplicates feature but cuDF does. We can wait for dask cudf or try a hack with cudf directly. Indeed, we can use a local `df = df.drop_duplicates()` when the edge list sorted to have the same effect than a global drop duplicate. Notice that hibench input is sorted by source.

### LINKED ISSUE #409: [BUG] [MG] Improve the wait mechanism after dask csv reader
It seems that, by default, the reader is lazy and exits before finishing reading. This resulted in longer PageRank runtime since we waited for dask tasks to complete before calling CUDA functions. To bypass that problem and have accurate timings we added a way to wait for the csv reader before entering  cugraph.dask.pagerank. The current solution does :
```
    ddf = dask_cudf.read_csv(...)
    y = ddf.to_delayed()
    x = client.compute(y)
    wait(x)
    pr = dcg.pagerank(x, ...)
```

I wish there was a parameter in dask / dask_cudf or a sync function. So that we would just call: 
```
    ddf = dask_cudf.read_csv(...)
    pr = dcg.pagerank(ddf, ...)
```

Iroy30:
Close #406 
Close #409

### DIFF
--- a/python/cugraph/dask/pagerank/pagerank.py
+++ b/python/cugraph/dask/pagerank/pagerank.py
@@ -14,6 +14,7 @@ def to_gpu_array(df):
     Get the gpu_array pointer to the data in columns of the
     input dataframe.
     """
+    df.drop_duplicates(inplace=True)
     start_idx = df.index[0]
     stop_idx = df.index[-1]
     gpu_array_src = df['src']._column._data.mem
@@ -191,12 +192,6 @@ def pagerank(edge_list, alpha=0.85, max_iter=30):
     return ddf
 
 
-def find_dev(df):
-    gpu_array_src = df['src']._column._data.mem
-    dev = device_of_devicendarray(gpu_array_src)
-    return dev
-
-
 def _get_mg_info(ddf):
     # Get gpu data pointers of columns of each dataframe partition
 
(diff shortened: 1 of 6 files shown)


python/cugraph/dask/pagerank/pagerank.py
```diff
@@ -7,13 +7,15 @@
...
 from toolz import first
+import dask.dataframe as dd
 
 
 def to_gpu_array(df):
     """
     Get the gpu_array pointer to the data in columns of the
     input dataframe.
     """
+    df = drop_duplicates(df)
```
afender: Can we call `drop_duplicates` in the hibench test instead of here?
I worry that calling it here adds an unnecessary overhead on edge lists that don't have duplicates.

Iroy30: edited

PR #423, opened 2019-07-30
[REVIEW] Fea ext SCC

aschaffer:
This addresses the following additions and updates:
1. Strongly Connected Components (SCC) for either directed or undirected graphs;
2. Updates in the Weakly CC C++ API: pass a `cudf::table` to the API, containing the labels and vertices columns;
3. Similar C++ API for SCC;
4. Corresponding Cython/Python API and tests;
5. SCC C++ test;


python/cugraph/components/connectivity_wrapper.pyx
```diff
@@ -42,9 +41,39 @@ def weakly_connected_components(graph_ptr, connect_type=CUGRAPH_WEAK):
 
     df = cudf.DataFrame()
     df['labels'] = cudf.Series(np.zeros(num_verts, dtype=np.int32))
-    cdef gdf_column c_labels = get_gdf_column_view(df['labels'])
+    df['vertices'] = cudf.Series(np.zeros(num_verts, dtype=np.int32))
+    
+    cdef cudf_table* tbl = <cudf_table*> table_from_dataframe(df)
```
shwina: After you are done with `tbl`, you need to `del tbl` corresponding to the `new` in `table_from_dataframe`. Note that this will not actually free any of the columns of the table, so `df` will be unaffected

aschaffer: Fixed it! Thanks.

aschaffer:
Hi @shwina, I see the following error on some platforms, following your recommendation to `del tbl`:

Error compiling Cython file:
------------------------------------------------------------
...
    cdef cugraph_cc_t connect_type=CUGRAPH_STRONG

    err = gdf_connected_components(g, <cugraph_cc_t>connect_type, tbl)

    cudf.bindings.cudf_cpp.check_gdf_error(err)

    del tbl
       ^
------------------------------------------------------------
cugraph/components/connectivity_wrapper.pyx:81:8: Deletion of non-Python, non-C++ object

Any idea what might be happening?

aschaffer:
So, per slack thread between @shwina and dev-ops, CI is not picking up latest fixed in cudf-0.9 that provide the `cudf::table` fixes. Not clear, yet, how / who should address that.

cpp/src/components/connectivity.cu
```diff
@@ -16,27 +19,38 @@
...
  * @tparam TPB_X the threads to use per block when configuring the kernel
  * @param graph input graph; assumed undirected for weakly CC [in]
- * @param labels gdf_column for the output labels [out]
- * @param connectivity_type CUGRAPH_WEAK or CUGRAPH_STRONG
- * @param stream the cuda stream
+ * @param table of 2 gdf_columns: output labels and vertex indices [out]
+ * @param connectivity_type CUGRAPH_WEAK or CUGRAPH_STRONG [in]
+ * @param stream the cuda stream [in]
  */
 template<typename IndexT,
```
seunghwak: The comment above (connectivity.cu from line 20 to 43) is redundant to the comment in cpp/include/algorithms.h. And AFAIK the standard way (in cugraph and in general as well) is to include this only in the header file. I haven't checked for every cugraph functions, but this is at least true for gdf_pagerank.

cpp/src/components/connectivity.cu
```diff
@@ -131,40 +146,85 @@ gdf_connected_components_impl(gdf_graph *graph,
...
+ * (combine, reduce) == (&, |) (bitwise ops)
+ * Then: X = C & transpose(C); and finally, apply get_labels(X);
+ *
  *
- 
  * @param graph input graph; assumed undirected for weakly CC [in]
- * @param connectivity_type CUGRAPH_WEAK, CUGRAPH_STRONG  [in]
- * @param labels gdf_column for the output labels [out]
+ * @param connectivity_type CUGRAPH_WEAK or CUGRAPH_STRONG [in]
+ * @param table of 2 gdf_columns: output labels and vertex indices [out]
```
seunghwak: Same here, we have the same comments describing this function in three different places (algorithms.h, gdf_connected_components, and gdf_connected_components_impl). I think this should appear only in the header file

cpp/src/components/connectivity.cu
```diff
@@ -131,40 +146,85 @@ gdf_connected_components_impl(gdf_graph *graph,
...
+      n2 = n2*(nrows*sizeof(ByteT) + sizeof(IndexT)) + 1;
+
+      int device;
+      cudaDeviceProp prop;
+      
+      cudaGetDevice(&device);
+      cudaGetDeviceProperties(&prop, device);
+
+      if( n2 > prop.totalGlobalMem )
+        {
```
seunghwak: What's the purpose of this check (we need to check memory allocation results anyway)?

"n2 > prop.totalGlobalMem" does not guarantee that we have enough memory to finish this computation (e.g. most of the totalGlobalMem can be already allocated to other GPU dataframes unrelated to this function).

cpp/src/components/connectivity.cu
aschaffer: There's nothing wrong w/ duplicating comments in headers and implementation files. It just makes information more readily available. AFAIK we didn't standardize the number of comments across source files. We cannot possibly standardize _everything_. Having _more_ comments cannot hurt. It helped the developer (me) when he wrote this; it will help the developers again when they need to update / add to the file.

cpp/src/components/connectivity.cu
aschaffer: ...and same answer as above.

cpp/src/components/scc_matrix.cuh
```diff
@@ -0,0 +1,260 @@
...
+    return p_d_c_i_;
+  }
+  
+  //protected: cannot have device lambda inside protected memf
+  void init(void)
+  {    
+    //assert( d_Cprev.size() == nrows_ * nrows_ );
+    
+    //init d_Cprev to identity:
+    //
```
seunghwak: Better delete the temporary commented out code.

cpp/src/components/connectivity.cu
aschaffer: The purpose of the check is to ensure _minimal_ memory requirements. Since this implementation of SCC is very memory consuming (O(|V|^2), the algorithm may fail. Without this check it would fail silently without apparent reason. This was a requirement that was specifically agreed upon during the numerous talks about GEMM SCC implementation and specifically required by @BradReesWork .

seunghwak: Does this mean this can still silently fail if memory allocation fails?

cpp/src/components/connectivity.cu
aschaffer: And in general it looks we need a file capturing all these "standardization" rules, many of which seem to be made up as we go. If there's a specific standardization please refer to the document stipulating it.

seunghwak: OK, just a recommendation, not something deserves a debate.

cpp/src/components/scc_matrix.cuh
aschaffer: Ok, removed dead code.

python/cugraph/components/connectivity.py
```diff
@@ -48,3 +49,39 @@ def weakly_connected_components(G):
...
+      an edge list (edge weights are not used for this algorithm). The graph
+      can be either directed or undirected where an undirected edge is
+      represented by a directed edge in both directions.
+      The adjacency list will be computed if not already present.
+      The number of vertices should fit into a 32b int.
+
+    Returns
+    -------
+    df : cudf.DataFrame
+      df['labels'][i] gives the label id of the i'th vertex
```
seunghwak: Need an empty line after this line for this to be properly formatted. If not this line and the next line appears in the same line in the generated html file.

aschaffer: I'm sorry, if it passes flake8 tests, it's compliant. I don't see why you feel the need to enforce more rules than flake8.

python/cugraph/components/connectivity.py
```diff
@@ -48,3 +49,39 @@ def weakly_connected_components(G):
...
+      df['vertices'][i] gives the vertex id of the i'th vertex
+
+    Examples
+    --------
+    >>> M = read_mtx_file(graph_file)
+    >>> sources = cudf.Series(M.row)
```
seunghwak: @afender requested to replace mtx file read in the examples with csv file read.

And to be consistent with the remainder of cugraph, this should be
```
    >>> M = cudf.read_csv('datasets/karate.csv', delimiter=' ',
    >>>                   dtype=['int32', 'int32', 'float32'], header=None)
    >>> sources = cudf.Series(M['0'])
    >>> destinations = cudf.Series(M['1'])
```

python/cugraph/components/connectivity.py
```diff
@@ -48,3 +49,39 @@ def weakly_connected_components(G):
...
+    >>> destinations = cudf.Series(M.col)
+    >>> G = cuGraph.Graph()
```
seunghwak: cuGraph.Graph() should be cugraph.Graph()

This code implicitly assumes "import cugraph" (not "import cugraph as cuGraph") is executed before this example.

python/cugraph/components/connectivity.py
```diff
@@ -48,3 +49,39 @@ def weakly_connected_components(G):
...
+    >>> G.add_edge_list(sources,destinations,none)
```
seunghwak: "none" should be "None".

python/cugraph/components/connectivity.py
```diff
@@ -48,3 +49,39 @@ def weakly_connected_components(G):
...
+    >>> df = cuGraph.strongly_connected_components(G)
```
seunghwak: cu"G"raph.strongly... should be cugraph.strongly...

python/cugraph/components/connectivity.py
aschaffer: Ok, fixed that.

python/cugraph/components/connectivity.py
aschaffer: Fixed.

python/cugraph/components/connectivity.py
aschaffer: Fixed.

python/cugraph/components/connectivity.py
seunghwak: This is not about python code formatting. This is about automatic document generation.

After running python setup.py install, if you go to cugraph/doc and run make html. This will automatically generate html documentation under build/html.

If you open the html file, this will appear as
```
df['labels'][i] gives the label id of the i'th vertex df['vertices'][i] gives the vertex id of the i'th vertex
```
instead of
```
df['labels'][i] gives the label id of the i'th vertex
df['vertices'][i] gives the vertex id of the i'th vertex
```
An empty line is necessary to get the latter (if you have used latex, this works similar to latex).

python/cugraph/components/connectivity.py
aschaffer: Fixed.

aschaffer:
@seunghwak 
"Looks good to me except for

1. the possibility of silent failure on memory allocation error (if you have checked that this does not happen, then, I am fine).

2. and my complaints about Examples in the python docstring."

1. it does not _because_ of the check I put in place;
2. theses issues have been addressed ;

seunghwak:
> 3. it does not _because_ of the check I put in place;

I guess "it does not" only if you run only strongly connected components. But if a user runs multiple different analytics (and they still take a significant amount of the GPU memory), and then, the user runs strongly connected components, the fact that GPU has enough physically installed DRAM does not guarantee that memory allocation will not fail. You may test this with first creating a large dummy dataframe and so some works, then run strongly connected components.

This is my concern, but if you have checked that this code will not silently fail even in this scenario, then, I will approve this PR.

aschaffer:
@seunghwak , I don't quite understand what is your point about the memory check. SCC checks the _minimal_ memory requirements _before_ it runs the scc implementation. If the minimal requirements are not met it bails out w/ an explicit error message (without which it would fail silently). Otherwise, it runs. Are you saying that between the time of the check and the time of the run other analytics could have kicked in and deplete the memory? Checking that would require protecting the memory pool w/ a mutex / semaphore, which is beyond the scope of this PR.

@BradReesWork, please moderate this.

seunghwak:
> @seunghwak , I don't quite understand what is your point about the memory check. SCC checks the _minimal_ memory requirements _before_ it runs the scc implementation. If the minimal requirements are not met it bails out w/ an explicit error message (without which it would fail silently). Otherwise, it runs. Are you saying that between the time of the check and the time of the run other analytics could have kicked in and deplete the memory? Checking that would require protecting the memory pool w/ a mutex / semaphore, which is beyond the scope of this PR.
> 
> @BradReesWork, please moderate this.
AFAIK, totalGlobalMem is the amount of physically installed DRAM (not the available DRAM at the point of the call, if I am properly interpreting the following link).
https://www.cs.cmu.edu/afs/cs/academic/class/15668-s11/www/cuda-doc/html/structcudaDeviceProp_983c292e2078dd5a4240f49c41d647f3.html#983c292e2078dd5a4240f49c41d647f3

If totalGlobalMem is the available memory, then, you're right. So, are you sure totalGlobalMem is the available memory?

Issue #422, opened 2019-07-30
[BUG] Update env


afender:
- rename the cugraph_dev to cugraph_dev_cuda9.2.yml
- update cugraph_dev env to use 0.9 packages

seunghwak:
@afender @BradReesWork Could you assign this to me? I need this to work on https://github.com/rapidsai/cugraph/issues/371

PR #438, opened 2019-08-02
[REVIEW] Fix versions of packages in build script and conda yml

dillon-cullinan:


afender:
Should we address https://github.com/rapidsai/cugraph/issues/422 in a separate PR or do everything at once?

PR #439, opened 2019-08-02
[REVIEW] FEA number of vertices from edge list

### LINKED ISSUE #241: [FEA] number_of_vertices from edge list
The current implementation computes the adj_list (if not present) in order to return the number of vertices. We should compute and store `max(max(src),max(dst))` when there is only the edge list.

ChuckHastings:
Closes #241 

Compute number of vertices more efficiently if we haven't generated CSR from the COO edge list.

1. Add an numberOfVertices element to the graph structure.  It is initialized to 0.
1. Populate it when we generate CSR
1. If the number_of_vertices method is called, do the following:
  a. If numberOfVertices is still 0, do a reduce on the COO arrays to compute the max and set numberOfVertices to this value
  b. otherwise return the value of numberOfVertices


kaatish:
I think that having an uninit val as -1 for numberOfVertices might be a better choice.

ChuckHastings:
I pondered using -1.  The only downside of using 0 that I could see is if a user could add a list of edges with 0 entries.  But that is currently prohibited - the code checks and fails if the user attempts to add a 0 length COO.

PR #445, opened 2019-08-06
[Review] Updates python graph view methods to return edge weights in addition

### LINKED ISSUE #319: [FEA] view_edge_list, view_adj_list, and view_transposed_adj_list do not return edge weights for weighted graphs
Currently, users have no way to retrieve edge weights from a cuGraph object.

seunghwak:
Close #319

### DIFF
--- a/python/cugraph/structure/graph_wrapper.pyx
+++ b/python/cugraph/structure/graph_wrapper.pyx
@@ -22,6 +22,7 @@ from libcpp cimport bool
 from libc.stdint cimport uintptr_t
 from libc.stdlib cimport calloc, malloc, free
 
+from cudf.bindings.cudf_cpp import gdf_to_np_dtype
 import cudf
 from librmm_cffi import librmm as rmm
 import numpy as np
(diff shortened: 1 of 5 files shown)


python/cugraph/structure/graph_wrapper.pyx
```diff
@@ -193,20 +233,40 @@ def view_transposed_adj_list(graph_ptr):
...
+                       nelem=offset_col_size,
+                       dtype=np.int32)  # ,
+                       # finalizer=rmm._make_finalizer(offset_col_data, 0))
+    offset_col = cudf.Series(offset_data)
+
+    index_data = rmm.device_array_from_ptr(
+                     index_col_data,
+                     nelem=index_col_size,
+                     dtype=np.int32)  # ,
+                     # finalizer=rmm._make_finalizer(index_col_data, 0))
```
kaatish: Is this dead code? Can this be removed?

seunghwak: Yeah... I'd better move that to the comment saying this column data should not be freed by when python object becomes out-of-scope.

seunghwak: Fixed.

PR #448, opened 2019-08-07
Import filter_unreachable

### LINKED ISSUE #392: [BUG] fix filter_unreachable
1. We need a small smoke test that calls the feature to make sure it is alive. It seems it was present in 0.8 but not in 0.9 after the refactoring.

2. __init__.py should be updated to import `filter_unreachable` : https://github.com/rapidsai/cugraph/blob/branch-0.9/python/cugraph/traversal/__init__.py

kaatish:
Fixes #392 
Add smoke test for filter_unreachable


afender:
Should we add a comment there https://github.com/kaatish/cugraph/blob/8e2fbab0e192a9fb0cb1bc6747428f6afcd22e04/python/cugraph/tests/test_sssp.py#L130 about the 32bit/64bit testing?

python/cugraph/tests/test_filter_unreachable.py
```diff
@@ -0,0 +1,79 @@
...
+
+    print('cugraph Solving... ')
+    t1 = time.time()
+
+    df = cugraph.sssp(G, source)
+
+    t2 = time.time() - t1
+    print('Time : '+str(t2))
+
+    cugraph.filter_unreachable(df)
```
seunghwak: Shouldn't we have some code to validate filter_unreachable is working as expected? (Need to check the outcome of filter_unreachable).

seunghwak: Besides this, this code looks good to me.

afender: Yes. We could just add a sanity check that verifies the returned DF is not empty and that it does not contain an infinite value.
For more advanced filtering feature testing, it is reasonable to rely on cudf tests in this case as this function is just wrapping one cudf call.

PR #450, opened 2019-08-07
[Review] Added a multi-GPU section in cuGraph documentation.

seunghwak:


seunghwak:
@afender As this update is about documentation on SNMG Pagerank, could you review this?

docs/source/dask-cugraph.rst
```diff
@@ -0,0 +1,15 @@
+~~~~~~~~~~~~~~~~~~~
+Multi-GPU with Dask
```
afender: Multi-GPU with Dask -> Multi-GPU

docs/source/dask-cugraph.rst
```diff
@@ -0,0 +1,15 @@
...
+~~~~~~~~~~~~~~~~~~~
+
+cuGraph is a single-GPU library. For Multi-GPU cuGraph solutions we use Dask (https://dask.org), which is able to scale cuGraph across multiple GPUs on a single machine, or in future releases, multiple GPUs across many machines in a cluster.
```
afender: I would suggest the following content (md format) : 
```
CuGraph's multi-GPU features leverage [Dask](https://dask.org). Dask is a flexible library for parallel computing in Python that makes scaling out your workflow smooth and simple. cuGraph also uses other RAPIDS projects based on Dask such as [dask-cuda](https://github.com/rapidsai/dask-cuda). 

The current solution is able to scale across multiple GPUs on a single machine.  Distributing the graph and computation lets you analyze datasets far larger than a single GPU’s memory without running into out of memory errors. With cuGraph and Dask, whether you’re using a single NVIDIA GPU or using all 16 NVIDIA V100 GPUs in a DGX-2, your RAPIDS workflow will run smoothly, intelligently distributing the workload across the resources available. 

###When to use multiple GPUs in cuGraph
If your graph comfortably fits in memory on a single GPU, you would want to use the single-GPU version of cuGraph. If you want to distribute your workflow across multiple GPUs and have more data than you can fit in memory on a single GPU, you would want to use cuGraph's multi-GPU solutions.
```

seunghwak: This edit looks great but my only concern is the "md format" part.

I picked rst instead of md to automatically generate documentation for SNMG analytics (currently only pagerank) from python files (AFAIK, markdown does not support this, but I am far from an expert in this regard, please correct me if it is possible).

I can embed markdown inside rst, but this requires installing additional m2r package (so this will involve some additional coordination with the CI team).

Another option is to add one .md file (for your description) and one .rst file from index.rst, but in this case, it will look like
```
cuGraph API Reference
    ...
Multi-GPU (markdown)
    ...
Multi-GPU : Supported Analytics (rst file)
    ...
```
Any thoughts?

afender: Markdown is just what I already had for this content. I didn't mean that we should keep it like this. We should convert that to rst if that's simpler. 

The m2r option you suggested would be interesting in the long run since we use markdown for readmes, GitHub descriptions and notebooks already. With the m2r setup we could easily reuse that content "as is" for the doc too. Probably something to consider during the refactoring release.

seunghwak: Make sense. I am little worried about introducing m2r in the burndown stage and see the CI people complaining about document build failure, but we should consider this in the restructuring phase.

Just FYI, here's the Stack Overflow link explaining how to include markdown in rst: https://stackoverflow.com/questions/46278683/include-my-markdown-readme-into-sphinx

afender: Sounds good. I added that to the refactor list, it seems that the only thing that can be problematic is the header level.

seunghwak: Made a fix.

I made few edits in the text and collapsed "Multi-GPU" header with "Multi-GPU with cuGraph" header (these two sound a bit redundant).

afender: perfect

Issue #456, opened 2019-08-14
[BUG] Potential race in SSSP


pgera:
I was going over the SSSP code, and I realized that there is a potential race. 
In this snippet,

```
// Try to relax non-masked edges                                    
              if (!edge_mask || edge_mask[edge]) {                                
                DistType* old_addr = &distances[dst_id];                          
                DistType old_val = *old_addr;                                     
                DistType new_val = distances[src_id] + edge_weights[edge];        
                if (new_val < old_val) {                                                                              
                  old_val = traversal::atomicMin(old_addr, new_val);   
```
the distances array is read as well as written. So we can end up with some corner cases where this check later in populate_frontier_and_preds:
```
if (expected_val == dst_val)
```
will fail.

The fix for this is to create a separate distances array. i.e., Have two arrays, curr_distances and next_distances similar to curr_frontier and next_frontier

PR #458, opened 2019-08-14
[REVIEW] Fix race in SSSP

pgera:
Fixes https://github.com/rapidsai/cugraph/issues/456


afender:
Thanks for fixing this @pgera ! I just changed the branch to 0.10 as we started packaging 0.9 already.

PR #464, opened 2019-08-15
[REVIEW] Consider limiting tests' initial pool size

trxcllnt:
Since RMM allocates half the available GPU memory by default, the pytests are forced to run sequentially. I'm opening this PR to discuss capping RMM's initial pool size in each test, so they can be run in parallel with `pytest-xdist`.

Here's the execution time pytest reports with and without `pytest-xdist`:
```shell
$ pytest .
== 460 passed, 1 warnings in 426.15 seconds ==

$ pytest -n auto .
== 460 passed, 2 warnings in 82.64 seconds ==
```

I'm not familiar enough with the tests to know whether the 2GB cap in this PR is too high or low. The tests also passed when I set it to 8, 128, and 512mb.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Improvements
 - PR #466 Add file splitting test; Update to reduce dask overhead
 - PR #468 Remove unnecessary print statement
+- PR #464 Limit initial RMM pool allocator size to 128mb so pytest can run in parallel
 
 ## Bug Fixes
 - PR #458 Fix potential race condition in SSSP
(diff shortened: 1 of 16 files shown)


afender:
We should be able to lower it to 128MB. The test that requires the most memory will be the strongly connected component one.

seunghwak:
Umm... and AFAIK, RMM dynamically increases its pool size as necessary, so if we run this sequentially, pytest will succeed regardless of the initial allocation size. So, if we are planning to run 32 GB / 128 MB tests in parallel, we need to make sure RMM pool size actually remains at 128 MB.

afender:
ok then let's try with 128MB. We'll also need to update changelog and check-style.

trxcllnt:
@afender alright, I've updated to use 128mb and add a changelog entry :-)

afender:
Thanks, @trxcllnt!

PR #466, opened 2019-08-15
snmg pagerank: add bigdatax8 test, edit returned dask-cudf

Iroy30:


python/cugraph/dask/pagerank/pagerank.py
```diff
@@ -278,3 +282,32 @@ def get_chunksize(input_path):
...
+              dtype=['int32', 'int32']):
+    df = []
+    for f in input_files:
+        df.append(cudf.read_csv(f, delimiter=delimiter, names=names,
+                                dtype=dtype))
+    df_concatenated = cudf.concat(df)
+    return df_concatenated
+
+
+def read_split_csv(input_files):
```
afender: Can we add a short doc at the beginning since it is exposed? It would be useful for users to know about the use case and limitations.

python/cugraph/dask/pagerank/pagerank.py
```diff
@@ -278,3 +282,32 @@ def get_chunksize(input_path):
...
+    for i, w in enumerate(client.has_what().keys()):
+        files_per_gpu = input_files[i*n_files_per_gpu: (i+1)*n_files_per_gpu]
+        worker_map.append((files_per_gpu, w))
+    t0 = time.time()
+    new_ddf = [client.submit(_read_csv, part, workers=[worker])
+               for part, worker in worker_map]
+
+    wait(new_ddf)
+    t1 = time.time()
+    print("Reading Csv time: ", t1-t0)
```
afender: the timer should not be inside the API function

python/cugraph/dask/pagerank/pagerank.py
```diff
@@ -278,3 +282,32 @@ def get_chunksize(input_path):
         size = [os.path.getsize(_file) for _file in input_files]
         chunksize = max(size)
     return chunksize
+
+
+def _read_csv(input_files, delimiter='\t', names=['src', 'dst'],
```
afender: should `delimiter` be a parameter in `read_split_csv`?

python/cugraph/dask/pagerank/pagerank.py
Iroy30: moved it in the test

python/cugraph/dask/pagerank/pagerank.py
Iroy30: added delimiter, names and dtype as parameters user can control

python/cugraph/dask/pagerank/pagerank.py
Iroy30: added

Issue #408, opened 2019-07-22
[FEA] Support int weights in SSSP


pgera:
Currently, BFS uses int32 for distances whereas SSSP uses floats or doubles for weights and distances. If the graph is unweighted, SSSP creates unit float weights (1.0f). In such cases, SSSP should fall back to BFS since BFS would be more efficient (more so for undirected graphs due to bottom up optimisations). Further, floats have limited precision for this use case which would be better served by int types. 

In order to do that, SSSP needs to support int weights at the API level first, so that it can seamlessly call BFS. It would be straightforward to add support for int types, but we need to think through some over/underflow issues before adding it. Since SSSP relies on edge relaxations, we want to avoid int over or under flows. One option is to add a conservative check at the beginning that checks that the weights in the graph are in the range (int_min/(|V|-1), int_max/(|V|-1) so that even in the worst case with a path of length |V|-1, we would not have over/under flow issues.

Iroy30:
addressed in PR #481. Instead of supporting int weights in sssp, we internally call bfs if the graph is unweighted for sssp. This roughly gives an improvement of 2.25x for a 3GB input csv

Issue #344, opened 2019-06-21
[BUG] weakly_connected_components(G, connect_type = CUGRAPH_WEAK) should not take connect_type


seunghwak:
**Describe the bug**

    connect_type : cugraph_cc_t
      Weak (CUGRAPH_WEAK), or Strong (CUGRAPH_STRONG) (not implemented, yet)

weakly_connected_compnents(G, connect_type=CUGRAPH_STRONG) does not make sense. In this case, do we get weakly connected components or strongly connected components? The function name already implies that this will compute weakly connected components.

Iroy30:
Can be closed

PR #481, opened 2019-09-09
[WIP] call bfs on weighted graphs when sssp is called

Iroy30:
Addresses #408 #344

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - PR #466 Add file splitting test; Update to reduce dask overhead
 - PR #468 Remove unnecessary print statement
 - PR #464 Limit initial RMM pool allocator size to 128mb so pytest can run in parallel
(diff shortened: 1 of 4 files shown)


afender:
rerun tests

ChuckHastings:
Code looks good to me.

The bug you fixed (344) is unrelated to the other code here.  Do we generally mix unrelated code into a single PR?  I've generally tried to avoid that in case a change needs to be rolled back, or a bug fix needs to be migrated into a different version.

I don't object, but was curious what the accepted practice was.

afender:
You are right, we should avoid mixing unrelated code addressing critical bugs and/or new features. 
However, we often see small enhancements or typos that can be fixed by just a one-line change. The overhead of filling a PR and starting a review process for these small issues is huge compared to the actual work of fixing it. It is more efficient to bundle those with a bigger PR as long as it is clearly specified and they have different commits IDs (so that they can be rolled back or migrated if needed).

PR #477, opened 2019-09-04
[REVIEW] Add cuHornet as a submodule

kaatish:
Fixes #431

PR #483, opened 2019-09-10
[REVIEW] Katz Centrality

kaatish:
Implemented Katz Centrality feature


cpp/include/algorithms.h
```diff
@@ -278,3 +278,20 @@ gdf_error gdf_snmg_pagerank (
             const size_t n_gpus, 
             const float damping_factor, 
             const int n_iter);
+/**                                                                             
+ * @Synopsis   Compute the Katz centrality for the nodes of the graph G
+ *                                                                              
+ * @Param[in] *graph                 cuGRAPH graph descriptor with a valid edgeList or adjList
+ *                                                                              
+ * @Param[out] *katz_centrality      If set to a valid column, this is populated by the katz centrality of every vertex in the graph
+ *
```
seunghwak: Where are the descriptions about the remaining parameters (alpha, max_iter, tol, has_guess, normalized)?

cpp/src/tests/centrality/katz_centrality_test.cu
```diff
@@ -0,0 +1,131 @@
...
+
+std::vector<int>
+getGoldenTopKIds(std::string file, int k = 10) {
+  std::vector<int> vec;
+  std::ifstream fin(file);
+  int val;
+  while (fin>>val) {
+    vec.push_back(val);
+  }
+  vec.resize(k);
```
seunghwak: Why not just read first k?

seunghwak: fin.close() missing here?

cpp/src/tests/centrality/katz_centrality_test.cu
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+  CoreNumber_Usecase_t& operator=(const CoreNumber_Usecase_t& rhs) {
+    matrix_file = rhs.matrix_file;
+    result_file = rhs.result_file;
+    return *this;
+  }
+} CoreNumber_Usecase;
+
+class Tests_CoreNumber : public ::testing::TestWithParam<CoreNumber_Usecase> {
```
seunghwak: Just for curiosity, why did you name this class as Tests_"CoreNumber"?

cpp/src/tests/centrality/katz_centrality_test.cu
```diff
@@ -0,0 +1,131 @@
...
+ public:
+  Tests_CoreNumber() {}
+  static void SetupTestCase() {}
+  static void TearDownTestCase() {}
+  virtual void SetUp() {}
+  virtual void TearDown() {}
+
+  static std::vector<double> SSSP_time;
```
seunghwak: What is this? Are we using this anywhere?

cpp/src/tests/centrality/katz_centrality_test.cu
```diff
@@ -0,0 +1,131 @@
...
+
+      ASSERT_EQ(gdf_edge_list_view(G.get(), col_src.get(), col_dest.get(), nullptr),0);
+
+      status = gdf_katz_centrality(G.get(), col_katz_centrality.get(), 0.1, 100, 1e-5, false, true);
+      EXPECT_EQ(status,0);
+
+      std::vector<int> top10CUGraph = getTopKIds(std::move(col_katz_centrality));
+      std::vector<int> top10Golden  = getGoldenTopKIds(param.result_file);
+
+      EXPECT_THAT(top10CUGraph, ::testing::ContainerEq(top10Golden));
```
seunghwak: Just to double check, due to the differences between NetworkX and cuGraph in computing Katz Centrality scores, we are trying to match just relative rankings, not actual scores?

python/cugraph/__init__.py
```diff
@@ -21,6 +21,7 @@
     subgraph,
     triangles
 )
+from cugraph.centrality import katz_centrality
```
seunghwak: Just to double check, in an older email thread, we discussed to expose the underlying framework to python users (e.g. cugraph.thirdparty.hornet.katz_centrality instead of cugraph.katz_centrality), but have we changed this to hide the underlying framework to users?

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+    Compute the Katz centrality for the nodes of the graph G
+
+    Parameters
+    ----------
+    graph : cuGraph.Graph
+        cuGraph graph descriptor with connectivity information. Edge weights,
+        if present, should be single or double precision floating point values.
+    alpha : float
+        Attenuation factor with a default value of 0.1. Alpha is set to
+        1/(lambda_max) if it is greater where lambda_max is the maximum degree
```
seunghwak: Just to double check, it is "greater" or it is smaller (i.e. shouldn't alpha <= 1 / lambda_max to converge? so, if alpha is smaller than 1/lambda_max, no need to change alpha but if alpha > 1 / lambda_max, shouldn't we reset alpha to 1 / lambda_max?). And if alpha is too large, shouldn't we better throw exceptions rather than silently (even though documented) resetting alpha to another value?

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+        The maximum number of iterations before an answer is returned. This can
+        be used to limit the execution time and do an early exit before the
+        solver reaches the convergence tolerance.
+        If this value is lower or equal to 0 cuGraph will use the default
+        value, which is 100.
+    tolerance : float
+        Set the tolerance the approximation, this parameter should be a small
+        magnitude value.
+        The lower the tolerance the better the approximation. If this value is
+        0.0f, cuGraph will use the default value which is 1.0E-5.
```
seunghwak: 1.0E-5=>1.0e-5 to be consistent with line 20.

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+        Setting too small a tolerance can lead to non-convergence due to
+        numerical roundoff. Usually values between 0.01 and 0.00001 are
```
seunghwak: 1e-2 and 1e-5 might be better for readability than 0.01 and 0.00001 (one needs to count zeros in 0.00001).

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+        acceptable.
+    nstart : cudf.Dataframe
+        GPU Dataframe containing the initial guess for katz centrality.
+    normalized : bool
+        If True normalize the resulting katz centrality values
+
+    Returns
+    -------
+    df : cudf.DataFrame
+        df['vertex'][i] gives the vertex id of the i'th vertex.
```
seunghwak: IIRC, there should be an empty line between line 57 and 58 to prevent cugraph document generator to place line 57 and 58 in the same line (you can run make doc under cugraph/docs and see the resulting document in build to confirm this).

python/cugraph/structure/graph.py
```diff
@@ -393,7 +393,7 @@ def number_of_edges(self):
 
     def in_degree(self, vertex_subset=None):
         """
-        Compute veretx in-degree. Vertex in-degree is the number of edges
+        Compute vertex in-degree. Vertex in-degree is the number of edges
```
seunghwak: Great catch!!!!!!!

python/cugraph/tests/dask/test_hibench_small.py
```diff
@@ -16,6 +17,7 @@
     import networkx as nx
 
 
+@pytest.mark.skip(reason="ongoing dask issue")
```
seunghwak: Isn't this a part of the previously merged PR (https://github.com/rapidsai/cugraph/pull/477/files)? Why does this appear again?

cpp/src/tests/centrality/katz_centrality_test.cu
kaatish: > fin.close() missing here?

Destructor of `std::basic_ifstream` calls the close function.

cpp/src/tests/centrality/katz_centrality_test.cu
kaatish: Yes, that is correct. NetworkX algorithm is different that the one implement here.

python/cugraph/centrality/katz_centrality.py
kaatish: > Alpha is set to 1/(lambda_max) if it is greater

I am implying that Alpha = min(alpha, 1/lambda_max)

> And if alpha is too large, shouldn't we better throw exceptions rather than silently (even though documented) resetting alpha to another value?

That could be done. We are computing out_degree's of the graph to reset this value. If we give up this responsibility to the user that would mean lesser computation for us.

python/cugraph/tests/dask/test_hibench_small.py
kaatish: Since this branched off from before that PR was merged, I had the option to merge it or add this line for the tests to pass. I could undo this and rebase the whole PR.

python/cugraph/__init__.py
BradReesWork: cuHornet is not a thirdparty library,  so it should be cugraph.cuhornet.katz_centrality.  
BUT ...   we also need a cugraph.katz_centrality function.  That top level call should pick the best sub-solution, which in this case is just the cuhornet version

python/cugraph/centrality/katz_centrality.py
seunghwak: I'm not saying we should give up this responsibility, but once we detected that alpha is too large, we'd better throw exception instead of silently setting alpha to a new value as this may surprise the user.

seunghwak: We still need to check that alpha is small enough for convergence, so I don't think this will reduce the amount of computing.

python/cugraph/tests/dask/test_hibench_small.py
seunghwak: Not a big deal as long as this gets properly merged.

cpp/CMakeLists.txt
```diff
@@ -327,6 +328,10 @@ target_include_directories(cugraph
     "${CUDF_INCLUDE}"
     "${CMAKE_CURRENT_SOURCE_DIR}/../thirdparty"
     "${CMAKE_CURRENT_SOURCE_DIR}/../thirdparty/cub"
+    "${CMAKE_CURRENT_SOURCE_DIR}/../thirdparty/cuhornet/hornet/include"
+    "${CMAKE_CURRENT_SOURCE_DIR}/../thirdparty/cuhornet/hornetsnest/include"
```
afender: Are we moving HornetNest inside cuGraph? If so, should `hornetsnest/include` be in cuGraph?

python/cugraph/__init__.py
afender: Agreed. @kaatish proposal is pretty user-friendly based on what we have. As a user what I would really like to do is `from cugraph.centrality import katz_centrality` and I don't need to be aware of the backend details especially if there's a single backend.

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+                    tol=1.0e-5,
+                    nstart=None,
+                    normalized=True):
+    """
+    Compute the Katz centrality for the nodes of the graph G
+
+    Parameters
+    ----------
+    graph : cuGraph.Graph
+        cuGraph graph descriptor with connectivity information. Edge weights,
```
afender: Should we talk about directed/undirected support?

python/cugraph/centrality/katz_centrality.py
afender: Why 1e-5? Nx uses 1.0e-6

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+    alpha : float
+        Attenuation factor with a default value of 0.1. Alpha is set to
+        1/(lambda_max) if it is greater where lambda_max is the maximum degree
+        of the graph.
+    max_iter : int
+        The maximum number of iterations before an answer is returned. This can
+        be used to limit the execution time and do an early exit before the
+        solver reaches the convergence tolerance.
+        If this value is lower or equal to 0 cuGraph will use the default
+        value, which is 100.
```
afender: Nx uses 1000. Not sure it is a good idea though. What do you think?

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+                    nstart=None,
+                    normalized=True):
+    """
+    Compute the Katz centrality for the nodes of the graph G
+
+    Parameters
+    ----------
+    graph : cuGraph.Graph
+        cuGraph graph descriptor with connectivity information. Edge weights,
+        if present, should be single or double precision floating point values.
```
afender: I suspect we temporarily set edge weights to something if they are not present? If so we may want to mention it here.

python/cugraph/centrality/katz_centrality.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def katz_centrality(G,
+                    alpha=0.1,
+                    max_iter=100,
+                    tol=1.0e-5,
+                    nstart=None,
+                    normalized=True):
+    """
+    Compute the Katz centrality for the nodes of the graph G
```
afender: We should add a note here for Nx compatibility regarding the missing `beta` parameter in our API.

python/cugraph/centrality/katz_centrality.py
ogreen: We can definitely use 1000. NetworkX uses a different convergence technique than what we use. So this parameter is slightly different. 1000 will increase the execution by 10X without seeing a big change in performance. I won't be surprised if 10 or 20 iterations are enough.

python/cugraph/centrality/katz_centrality.py
ogreen: The algorithm should work for both types.

python/cugraph/centrality/katz_centrality.py
afender: I see. Does it worth mentioning that we use a different convergence technique in the doc then? Or we could just say something like "For most cases, the ranking does not change much after 20 interations with our implementation"

python/cugraph/centrality/katz_centrality.py
afender: Should an undirected edge be stored as a directed edge in both directions in the CSR matrix?

python/cugraph/centrality/katz_centrality.py
ogreen: Good question. Maybe we can run some experiments to figure out what a good iteration count is for the average case.

python/cugraph/centrality/katz_centrality.py
ogreen: That is what we do. We don't have two arrays with in and out. Only out edges. So each (u,v) appears twice. Once in the out of u and once in the out of v.

python/cugraph/centrality/katz_centrality.py
ogreen: I don't think that we need edges weights at all. Katz centrality doesn't use weights by default. So edge weights will be ignored.

python/cugraph/centrality/katz_centrality.py
afender: That's consistent with what other cuGraph algo do. I don't see it in the doc though. We may want to clarify it here.

python/cugraph/centrality/katz_centrality.py
afender: Sounds good, let's update the doc to reflect that.

ogreen:
Do we want to add a pointer to the academic paper that explains this algorithm in additional detail?

python/cugraph/centrality/katz_centrality.py
seunghwak: and if we're going to document hornet (gunrock, graphBLAS) algorithms as other cuGraph algorithms, we need to add

```
Centrality
=======

Katz Centrality
---------------

.. automodule:: cugraph.centrality.katz_centrality (or cugraph.cuhornet.centrality.katz_centrality based on where we place this in the python directory).
    :members:
    :undoc-members:
```

to

https://github.com/rapidsai/cugraph/blob/branch-0.10/docs/source/api.rst

python/cugraph/centrality/katz_centrality.py
kaatish: I have mentioned the different.

python/cugraph/centrality/katz_centrality.py
kaatish: I have mentioned the undirect/directness of the input graph.

python/cugraph/centrality/katz_centrality.py
kaatish: I have followed example in `spectral_clustering.py`

python/cugraph/centrality/katz_centrality.py
kaatish: Done

cpp/src/tests/centrality/katz_centrality_test.cu
kaatish: Removed

cpp/src/tests/centrality/katz_centrality_test.cu
kaatish: Fixed

cpp/src/tests/centrality/katz_centrality_test.cu
kaatish: Fixed

cpp/include/algorithms.h
kaatish: Added

python/cugraph/centrality/katz_centrality.py
kaatish: Fixed

python/cugraph/centrality/katz_centrality.py
kaatish: Fixed. `gdf_katz_centrality` will return `GDF_CUDA_ERROR` along with a message notifying the user of the largest acceptable `alpha` value for invalid user input.

kaatish:
> Do we want to add a pointer to the academic paper that explains this algorithm in additional detail?

Added in `katz_centrality.py`

PR #496, opened 2019-09-22
[WIP] [REVIEW] Fea ext gunrock sm

aschaffer:
This branch is for review of C++ API only (a new one will be created for the actual functional code). Currently this branch has a lot of issues (waiting for some clarifications from Gunrock side). Our API wrapper of Gunrock SM is in `cpp/src/matching/subg_match.cu`.


cpp/src/matching/subg_match.cu
```diff
@@ -0,0 +1,159 @@
...
+#include "utilities/graph_utils.cuh"
+#include "utilities/error_utils.h"
+#include <cugraph.h>
+#include <algo_types.h>
+
+#include <iostream>
+#include <type_traits>
+#include <cstdint>
+
+extern double sm(const int            num_nodes,
```
ChuckHastings: I assume this definition is a work-around for their problem of not exposing this call in the current version.

cpp/src/matching/subg_match.cu
```diff
@@ -0,0 +1,159 @@
...
+    return G->adjList->edge_data->dtype;
+  };
+
+  
+  auto subg_dtype = subgraphs->dtype;
+  auto ro_dtype   = row_offsets_t_(graph);
+  auto ci_dtype   = col_indices_t_(graph);
+  auto v_dtype    = values_t_(graph);
+  
+  GDF_REQUIRE( subg_dtype == ci_dtype, GDF_INVALID_API_CALL);//currently Gunrock's API requires that graph's col indices and subgraphs must be same type;
```
ChuckHastings: Do we think this is an unreasonable assumption?  If so are they going to fix this, or do we need to worry about working around this?

cpp/src/tests/CMakeLists.txt
```diff
@@ -125,19 +125,19 @@ set(KATZ_TEST_SRC
 ###################################################################################################
 # - gdfgraph tests --------------------------------------------------------------------------------
 
-set(GDFGRAPH_TEST_SRC
-    "${CMAKE_CURRENT_SOURCE_DIR}/gdf_graph/gdf_graph.cu")
+#set(GDFGRAPH_TEST_SRC
+#    "${CMAKE_CURRENT_SOURCE_DIR}/gdf_graph/gdf_graph.cu")
```
afender: There are many stand-alone tests in there that don't need GRMAT

cpp/src/tests/CMakeLists.txt
```diff
@@ -150,29 +150,29 @@ ConfigureTest(PAGERANK_TEST "${PAGERANK_TEST_SRC}" "")
 
 ###################################################################################################
 # - SSSP tests ------------------------------------------------------------------------------------
-set(SSSP_TEST_SRCS
```
afender: Same for jaccard and louvain, the GRMAT dependency is limited to a specific google test. Disabling only these tests would allow other tests to run.
For SSSP it is a feature of the test app that can be deleted.

cpp/src/matching/subg_match.cu
aschaffer: I doubt it's reasonable. But that's what they have in their code. There are a few other issues that I addressed on their Github. Specifically, indexing type for tow offsets is `SizeT` rather than `VertexT`. But...we must comply with their API.

cpp/src/matching/subg_match.cu
aschaffer: Yep. I was hoping I can bypass that and just include the header with the `template<...> sm(...)` but when I try to instantiate it (same way they do in their code) I get compiler errors in their code. You may actually see that `#include <...sm_app.cuh>` is commented out.

cpp/src/tests/CMakeLists.txt
aschaffer: I'm not sure: what seems to be the issue here? I commented out GRMAT tests and `grmat.cu`, as requested by @BradReesWork.

cpp/src/tests/CMakeLists.txt
aschaffer: Yes. This is what I have done.

cpp/src/tests/CMakeLists.txt
afender: By commenting tests apps here you also got rid of all non-grmat tests. If you check https://github.com/rapidsai/cugraph/blob/branch-0.10/cpp/src/tests/gdf_graph/gdf_graph.cu you will see that it contains only 3 tests using `GRMAT` and 22 tests that don't.

cpp/src/tests/CMakeLists.txt
afender: I am not seeing this in the PR.

cpp/src/tests/CMakeLists.txt
aschaffer: Ah, okay, I understand now. Okay, will address this.

cpp/src/tests/CMakeLists.txt
aschaffer: Again, what you're saying is to selectively disable the GRMAT-using tests, _only_ for Louvain, Jaccard, SSSP. Correct? I can do so, but that's changing a lot of code (potentially) that we want to revive, anyway. This was just a temporary measure until RMAT is operational in Gunrock again. Which, in fact, already is (I'm told). It's just a matter of rewriting our GRMAT tests to use their new API (why they canhed their RMAT API...that I don't know). So, I doubt there's much value in spending time changing five tests so that we change them back in the new release. We might as well disable them for now.

aschaffer: I actually would have been in favor of _not disabling_ any test, because that to me is tantamount to a regression; but, fix the RMAT first. But that was overruled in favor of a faster fix for this release, and taking care of RMAT in the next.

afender:
Seems like python still has some grmat dependencies : `undefined symbol: _Z13gdf_grmat_gen`
https://gpuci.gpuopenanalytics.com/blue/organizations/jenkins/gpuCI%2Fcugraph%2Fprb%2Fcugraph-gpu-build/detail/cugraph-gpu-build/1326/pipeline

.gitmodules
```diff
@@ -5,3 +5,7 @@
 [submodule "thirdparty/cuhornet"]
 	path = thirdparty/cuhornet
 	url = https://github.com/rapidsai/cuhornet.git
+[submodule "thirdparty/gunrock"]
+	path = thirdparty/gunrock
+	url = https://github.com/gunrock/gunrock
+	branch = dev
```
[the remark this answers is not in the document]
aschaffer: Because the master is behind. It doesn't have the SM template fixes. Nor the new RMAT.

cpp/src/matching/subg_match.cu
```diff
@@ -0,0 +1,159 @@
...
+ * @param  subgraphs   Return number of subgraphs [out]
+ * @param  stream the cuda stream [in / optional]
+ */
+template <typename VertexT,
+          typename SizeT,
+          typename GValueT>
+gdf_error gdf_subgraph_matching_impl(gdf_graph *graph,
+                                     VertexT* subgraphs,
+                                     cudaStream_t stream = nullptr)
+{
```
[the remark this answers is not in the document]
aschaffer: I changed the API in the meantime. It takes 2 graphs now.

cpp/src/matching/subg_match.cu
```diff
@@ -0,0 +1,159 @@
...
+          //more...
+          default:
+            break;//warning eater
+          }
+      }
+    //more...
+    default:
+      break;//warning eater
+    }
+  */
```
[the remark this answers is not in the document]
aschaffer: This is placeholder code for when SM is ready. Look at it like the slot to put the gunrock code in when it will be ready.

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

rlratzel:
rerun tests

rlratzel:
@aschaffer I believe the failing CI checks are because the `gunrock` submodule you wrote code against is not up-to-date, and the CI check will always update it (using `git submodule update --init --recursive --remote thirdparty/gunrock`).  If you run with the above `--remote` command, you'll see the file CI is complaining about has been moved/renamed.  I'm just guessing here, but is it now inline code moved to a header file?
```
rratzel@sm01:~/Projects/cugraph$ ls -l thirdparty/gunrock/externals/moderngpu/src/mgpucontext.cu
ls: cannot access 'thirdparty/gunrock/externals/moderngpu/src/mgpucontext.cu': No such file or directory

rratzel@sm01:~/Projects/cugraph$ ls -l thirdparty/gunrock/externals/moderngpu/src/moderngpu/context.hxx
-rw-rw-r-- 1 rratzel rratzel 6162 Oct  4 02:19 thirdparty/gunrock/externals/moderngpu/src/moderngpu/context.hxx
```

https://git-scm.com/book/en/v2/Git-Tools-Submodules

cc @raydouglass

rlratzel:
> I'm not sure what's the fix. One thing I can say is that I am (was) cloning their `dev` (rather than `release`) branch because that was the branch that had what we needed from them. As far as their `moderngpu` dependency goes: I don't know. We don't have it anymore (specifically dropped our dependency on `moderngpu`). So, what's the action item for me? Or, is the PR now mergeable?

Not mergeable, unfortunately.  I'm not very familiar with git submodules either to be of much help yet.  Were you able to take a look at the gunrock source when you add the --remote option?  I'm guessing your cmake and other code may need to be updated to accommodate the gunrock updates, but I'm not sure.  We should talk to @dantegd since he's handled this in cuml before.

I should also mention that the best suggestion may be to have cmake pull in the dependency instead of using a git submodule, simply because this is the way cuml is now doing it, and I think cudf will be too.

aschaffer:
@rlratzel, so I couldn't repro the issue, and you mentioned you couldn't either, initially. Just so I understand the issue correctly, it seems there's a specific CI setup step that triggers this problem to become manifest. Is it the `--remote` setting in the `git submodule` command? 

We typically use (and recommend to the users in the documentation: https://github.com/rapidsai/cugraph/blob/branch-0.10/CONTRIBUTING.md ) `git submodule update --init --recursive`; i.e., without `--remote`. Is `--remote` critical for CI? If not, can it be dropped? Or, can it be selectively applied to dependencies that need it (and _not_ applied to those that might not need it)?

BradReesWork:
rerun tests

python/cugraph/__init__.py
```diff
@@ -28,7 +28,7 @@
 from cugraph.link_prediction import jaccard, overlap, jaccard_w, overlap_w
 from cugraph.structure import Graph, from_cudf_edgelist, renumber, symmetrize, symmetrize_df
 from cugraph.traversal import bfs, sssp, filter_unreachable
-from cugraph.utilities import grmat_gen
+# from cugraph.utilities import grmat_gen
```
rlratzel: I remember us discussing this approach (just comment out all places where this is imported) vs. removing the (temporarily) broken function from the cython interface. This way is easier, but it allows someone to still import `grmat_gen` and crash.  Would it be better to remove the API from cython instead, so graph users cannot import a broken API? Or another option could be to re-write `grmat_gen` in cython to raise `NotImplementedError`?

rlratzel: I suppose there's no easy way to remove an API, and either approach will result in a traceback, so I'm fine with it either way. I guess we just have to decide what type of exception we'd want a user to see. @afender ?

aschaffer: We're temporarily suspending RMAT. We intend to bring it back in 0.11 (I think). So, possible subsequent bugs that might be opened by users trying to run GRMAT from Cython (which I think is the only way to access it, since direct Python dependencies on it have been removed) will get automatically addressed in next release. Which might be a better scenario than silently removing a feature that people would ask were it went.

PR #499, opened 2019-09-24
[REVIEW] - PR #499 Moved small katz golden data to test file

kaatish:


BradReesWork:
Is there a way to indicate that the datafiles contain the Katz results and not the graph?   Just looking at the file name gets confusing since there is another file with the same name

kaatish:
> Is there a way to indicate that the datafiles contain the Katz results and not the graph? Just looking at the file name gets confusing since there is another file with the same name

Looks like this PR is not needed at all since the build was fixed.

PR #520, opened 2019-09-26
[REVIEW] Core number and K Core

kaatish:
- Core Number and KCore implementation
- Python tests
- Corrected wrong Katz Centrality documentation


cpp/src/core/core_number.cu
```diff
@@ -0,0 +1,141 @@
...
+    core_number_ptr = coreNumber.data().get();
+    err = core_number_impl(in_graph, core_number_ptr);
+    GDF_REQUIRE(err, GDF_SUCCESS);
+  } else {
+    core_number_ptr = static_cast<int*>(core_number->data);
+    vertex_identifier_ptr = static_cast<int*>(vertex_id->data);
+    vLen = vertex_id->size;
+  }
+
+  auto cnPtr = thrust::device_pointer_cast(core_number_ptr);
```
ChuckHastings: Not a big fan of using auto to define the type of a variable that's going to be used for a while.  Isn't this just int *?  That would be a bit more readable.

cpp/src/core/core_number.cu
```diff
@@ -0,0 +1,141 @@
...
+  if (k == -1) {
+    k = thrust::reduce(
+        rmm::exec_policy(stream)->on(stream),
+        cnPtr, cnPtr + vLen, -1, thrust::maximum<int>());
+  }
+
+  rmm::device_vector<int> filteredVertices(vLen);
+  auto vInPtr = thrust::device_pointer_cast(vertex_identifier_ptr);
+  auto ptr = thrust::remove_copy_if(
```
ChuckHastings: Ditto above.  Aren't vInPtr and ptr also just int *?

I'm a little less pressed about this since you only use the variable in the very next statement.  But I think int * here (or whatever the type really is if not that) is a little more clear for someone coming along later to understand or modify the code.

python/cugraph/core/core_number.py
```diff
@@ -0,0 +1,52 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cugraph.core import core_number_wrapper
+
+
+def core_number(G):
+    """
+    Compute the core numbers for the nodes of the graph G.
```
ChuckHastings: Maybe a bit more description here.  NetworkX has a couple of sentences describing k-core and what the core number really is.  I couldn't remember and had to go look it up.

python/cugraph/core/k_core.py
```diff
@@ -0,0 +1,61 @@
...
+
+from cugraph.core import k_core_wrapper
+from cugraph.structure.graph import Graph
+
+
+def k_core(G,
+           k=None,
+           core_number=None):
+    """
+    Compute the core numbers for the nodes of the graph G.
```
ChuckHastings: Better description here.  In fact, this is misleading.  k_core is going to compute the core numbers, and then it's going to create and return a new graph that is the k-core of this graph.  Perhaps a copy/paste error.

cpp/src/centrality/katz_centrality.cu
```diff
@@ -39,6 +39,8 @@ gdf_error gdf_katz_centrality(gdf_graph *graph,
     return err;
   GDF_REQUIRE(graph->adjList->offsets->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);
   GDF_REQUIRE(graph->adjList->indices->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);
+  GDF_REQUIRE(katz_centrality->dtype == GDF_FLOAT64, GDF_UNSUPPORTED_DTYPE);
```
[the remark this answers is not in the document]
ogreen: This is not really a limitation of the implementation. Rather it was a design decision to ensure that the results are more accurate due to insatiability of floating point. Specifically, if float32 is used instead of float64\double, then the conversion might happen in too few iterations to ensure correctness. However, it could be that it might be enough.

cpp/src/core/core_number.cu
kaatish: I can replace `auto` with `thrust::device_ptr<int>` here.

cpp/src/core/core_number.cu
kaatish: I can replace `auto` with `thrust::device_ptr<int>` here for `vInPtr`.

`ptr` however is `thrust::detail::normal_iterator<thrust::device_ptr<int>>`. Its a bit too long and unclear so that's why I chose `auto`.

python/cugraph/core/k_core_wrapper.pyx
```diff
@@ -0,0 +1,57 @@
...
+    cdef gdf_graph* rg = <gdf_graph*>rGraph
+
+    kVal = -1
+    if k is not None:
+        kVal = k
+
+    cdef gdf_column c_vertex
+    cdef gdf_column c_values
+
+    if core_number is None:
```
Iroy30: Should the if else be reversed? if core number is None, it can't subscript core['vertex'] or core['values']

cpp/src/core/core_number.cu
```diff
@@ -0,0 +1,144 @@
+/*
```
afender: Can we rename `core` directory into `cores`? Mostly because `src/core` can be misleading as one would expect "core" files for the whole project to live there. As a nice side effect Nx also uses `cores`

python/cugraph/centrality/katz_centrality.py
```diff
@@ -33,7 +33,7 @@ def katz_centrality(G,
 
     Parameters
     ----------
-    graph : cuGraph.Graph
+    G : cuGraph.Graph
         cuGraph graph descriptor with connectivity information. The graph can
         contain either directed or undirected edges where undirected edges are
         represented as directed edges in both directions.
```
afender: We could add a word on weights here btw.

python/cugraph/core/core_number.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Parameters
+    ----------
+    graph : cuGraph.Graph
+        cuGraph graph descriptor with connectivity information. The graph can
+        contain either directed or undirected edges where undirected edges are
+        represented as directed edges in both directions.
+
```
afender: 1. That's pretty obvious but it doesn't harm to clarify that weights are not used in this algo.
2. I read in Nx that self-loops are not supported in their implementation, is this ok in yours? If not we should say it here.

python/cugraph/core/k_core.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
afender: would recommend renaming the directory here as well as `python/cugraph/core` can mean something else.

python/cugraph/core/k_core.py
```diff
@@ -0,0 +1,65 @@
...
+    Parameters
+    ----------
+    G : cuGraph.Graph
+        cuGraph graph descriptor with connectivity information. The graph can
+        contain either directed or undirected edges where undirected edges are
+        represented as directed edges in both directions.
+    k : int, optional
+        Order of the core. If set to None, the main core is returned.
+    core_number : cudf.DataFrame, optional
+        Precomputer core number of the nodes of the graph G containing two
```
afender: Precomputer -> Precomputed

python/cugraph/core/k_core.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Examples
+    --------
+    >>> M = cudf.read_csv('datasets/karate.csv', delimiter=' ',
+    >>>                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> sources = cudf.Series(M['0'])
+    >>> destinations = cudf.Series(M['1'])
+    >>> G = cugraph.Graph()
+    >>> G.add_edge_list(sources, destinations, None)
+    >>> KCoreGraph = cugraph.core_number(G)
```
afender: Seems like the example need to be updated to reflect  `k_core` function

cpp/src/core/core_number.cu
```diff
@@ -0,0 +1,144 @@
...
+
+  LessThan(int _k) : k(_k) {}
+
+  __host__ __device__
+    bool operator()(const int core) {
+      return core < k;
+    }
+};
+
+gdf_error gdf_k_core(gdf_graph *in_graph,
```
afender: This function is doing more than just calling k-core from hornet. What is the motivation for that? 
I though implementations were supposed to be in Hornet as headers and cuGraph would just instantiate and call them. It would be great if the integration pattern was similar to Katz and CoreNumber.

afender:
+ changelog

cpp/src/core/core_number.cu
kaatish: Hornet is being used for core_number per vertex call as seen in `gdf_core_number`. This function calculates the relevant subgraph using those results. The subgraph can definitely be calculated in a better way in Hornet however that would require an extra copy and a few other steps. Given that gdf_k_core is already expected to accept filtered vertices as a column, it made more sense to make use of the functions in cugraph.

python/cugraph/core/k_core_wrapper.pyx
kaatish: Yes, I believe that's an error here.

python/cugraph/core/core_number.py
kaatish: 1. Fixed
2. NetworkX does not support self loops and parallel edges. Hornet does not support this as well. However, the difference is that our implementation does not check for this while NetworkX does before running core_number/kcore. Should there be a function implemented that checks for this error?

afender: It would be useful to have a helper function for checking self-loops in cuGraph. Once it is available we can add this to k-core pipelines and others. I don't want to block this PR for that though. https://github.com/rapidsai/cugraph/issues/536

afender: What really matters for this PR is that not supporting self loops is clearly mentioned in the documentation.

afender:
rerun tests

cpp/src/core/core_number.cu
afender: I think that having hornet-based implementations in hornet would be ideal for long term maintenance. Having said that I agree that the subgraph extraction from cugraph is part of the k-core feature and can make things a little bit more challenging. I noticed thrust calls before and after calling the core function, what would be the trade-off if we were to wrap just these? 

Also, can we add a few comments in `gdf_k_core `?

Issue #482, opened 2019-09-09
[BUG] Pagerank shows different results between networkx and cugraph


inc0:
**Describe the bug**
We have tried to run pagerank on cugraph and networkx on same dataset. NetworkX shown different order than cugraph, and it made much more sense for us. 

**To Reproduce**
1. Download this dataset: http://snap.stanford.edu/data/soc-Epinions1.html
2. Create cugraph and networkx graphs
3. Run pagerank
4. Compare top results

**Expected behavior**
Relative pagerank should be similar in both libs

**Additional context**
Internally we ran pagerank on repositories and it's dependencies for javascript. In cugraph, top pagerank were consistently small repos vs in networkx it was `babel/babel`, react etc, which makes much more sense. I assume networkx produces better pagerank so it might be the case in this dataset too.

ogreen:
Could you clarify, the following "Compare top results":
Are you comparing the PageRank values or the rankings?

inc0:
ranking ordered by pagerank value, as in, `cugraph.link_analysis.pagerank(G).sort_values(by="pagerank", ascending=False).to_pandas()`

afender:
I had a look at the file. It says there are 75879 vertices. However, I see some vertex indices over 75887 in that file. It is possible that Nx silently renumbers this dataset while cuGraph creates all vertices using the edge list labels (from 0 to max_label). A simple way to check that would be to ask the actual number of vertices to each framework and check if they match.

BradReesWork:
Let's download the dataset, give it a try with and without renumbering

afender:
@inc0 It would be helpful to see the code you are using. As far as I can see cuGraph and Nx outputs match when the same input is provided to PageRank.

As @BradReesWork suggested, I downloaded the dataset and compared cuGraph output to Nx with and without renumbering. I realized that loading the graph in Nx using `nx.read_edgelist(file)` creates an undirected graph and NetworkX ends up analyzing 811480 directed edges instead of 508837. Is this what you are using? If that configuration produces better results in your application, you could transform your graph by symmetrizing it in cuGraph too by using `symmetrize_df(df, src_name, dst_name)`.

Alternatively, when I used Scipy reader for networkX I was able to get the proper DiGraph (508837 edges) in Nx and got the same result as cuGraph. I attached a notebook for it in the zip file :  `pr_epinions_mtx.ipynb`. 

[...]

PR #527, opened 2019-09-27
[REVIEW] Support testing with asymmetric graph

### LINKED ISSUE #317: [FEA] Needs python tests for asymmetric directed graphs.

All our python test input graphs are symmetric. All csv files store symmetric graphs. Mtx format assumes that directed graphs (can be asymmetric) are converted to undirected graphs (automatically add reverse edges on read).

Our cugraph python tests do not test cugraph algorithms work properly for asymmetric directed graphs.

ChuckHastings:
Closes #317 

Added new utility function that allows us to run the networkx tests and the cugraph tests from just the csv files (rather than needing both csv and mtx formats).

Modified a bunch of tests to use this new utility function.

The file email-Eu-core.csv contains a directed graph (asymmetric) that was added for symmetrization tests.  I added this test data set to several of the algorithms to test how they work on an asymmetric graph.

State of this work:
- Modified all of the python tests to use csv as input for both CPU and GPU tests
- Tried adding the email-Eu-core.csv test to each of the unit tests.
- Works for:  bfs, connectivity, modularity, sssp, subgraph extraction
- Works for overlap and woverlap, but the tests take too long, so it is commented out
- The following tests get different answers from networkx on the asymmetric graph:  triangle counting, jaccard, wjaccard, katz_centrality, louvain.  These should be investigated further
- balanced cut does not get the same answer.  Internet search suggests that spectral clustering on directed graphs is not well understood, so perhaps we shouldn't try it on a directed graph
- page rank mostly works but needs some exploration.  The email-Eu-core graph requires more iterations to converge.  If I raise the number of iterations high enough everything converges but we end up with a fair number of differences with networkx (enough for a few of the tests to fail).  Makes me suspect that something might be a little different in the computation or the numerical instability is introducing noise in the computation.

I suggest we create stories for verifying behavior on a directed graph with the following algorithms:
- page rank
- triangle counting
- jaccard and wjaccard
- katz
- louvain

We might want to consider if there is a more efficient way of testing overlap/woverlap to validate results.  The GPU version runs very quickly.  There is no networkx implementation, we test with some python code written in the unit test.  I did a simple optimization of the test code, but it was insufficient.


python/cugraph/tests/test_pagerank.py
```diff
@@ -159,16 +158,16 @@ def test_pagerank(managed, pool, graph_file, max_iter, tol, alpha,
...
-    M = utils.read_mtx_file(graph_file+'.mtx')
+    M = utils.read_csv_for_nx(graph_file)
     networkx_pr, networkx_prsn = networkx_call(M, max_iter, tol, alpha,
                                                personalization_perc)
 
     cu_nstart = None
     if has_guess == 1:
         cu_nstart = cudify(networkx_pr)
-        max_iter = 5
+        max_iter = 500
```
BradReesWork: 500 seems like a lot of iterations.  I thought 50 or 100 was more common

python/cugraph/tests/test_pagerank.py
```diff
@@ -177,7 +176,7 @@ def test_pagerank(managed, pool, graph_file, max_iter, tol, alpha,
     err = 0
     assert len(cugraph_pr) == len(networkx_pr)
     for i in range(len(cugraph_pr)):
-        if(abs(cugraph_pr[i][1]-networkx_pr[i][1]) > tol*1.1
+        if(abs(cugraph_pr[i][1]-networkx_pr[i][1]) > tol*2.1
```
BradReesWork: The initial check was that the two values needed to be within 10%.  It looks like you changed that to be within 210%

afender:
> balanced cut does not get the same answer. Internet search suggests that spectral clustering on directed graphs is not well understood, so perhaps we shouldn't try it on a directed graph

The spectral clustering theory relies on finding the eigenpairs of an SPD matrix (Laplacian). One way of dealing with nonsymmetric input is to symmetrize it before calling spectral clustering. Passing a nonsymmetric input is expected to lead to undefined behavior and errors.

afender:
For Pagerank, I'd suggest checking that the graph is properly loaded in Nx (ie. in a way that it is the same data as in cuGraph). Nx often creates an undirected graph by default (ie. a `Graph` instead of DiGraph) and silently gets rid of singletons. This was the problem in 
https://github.com/rapidsai/cugraph/issues/482 for instance.

A good way to check that would be to add a test for the `read_csv_for_nx` function. You could then ask for the number of edges and vertices in Nx and CuGraph and make sure they match.

afender:
> The following tests get different answers from networkx on the asymmetric graph: triangle counting, jaccard, wjaccard, katz_centrality, louvain. These should be investigated further.

Jaccard and Louvain implementations are known to work only on undirected graphs. Doc is up to date https://docs.rapids.ai/api/cugraph/stable/. We can disable the nonsymmetric test for them.

afender:
> Works for: bfs, connectivity, modularity, sssp, subgraph extraction

This is surprising from Modularity. It shouldn't work for the same reasons as for Balanced cut. When the graph is nonsymmetric, the Courant-Fischer theorem is no longer applicable "as is" and the objective function no longer corresponds to the largest eigenpairs of the Modularity matrix.

python/cugraph/tests/test_modularity.py
```diff
@@ -42,10 +42,10 @@ def random_call(G, partitions):
...
 
 
-DATASETS = [
-    '../datasets/karate',
-    '../datasets/dolphins',
-    '../datasets/netscience']
+DATASETS = ['../datasets/karate.csv',
+            '../datasets/dolphins.csv',
+            '../datasets/netscience.csv',
+            '../datasets/email-Eu-core.csv']
```
afender: I'd also recommend deleting directed input testing here as it is not supposed to work on undirected input

python/cugraph/tests/test_pagerank.py
ChuckHastings: That's residue from trying to get it to work with the directed data set.  I didn't find a good number, but 50 wasn't sufficient.  500 converged but got different answers between networkx and cugraph for some combinations.  But I turned the directed data sets off so I'll set it back to 5.

A successful call exits once we converge to the tolerance, so most of the runs were just as fast as long as max_iter was large enough.

python/cugraph/tests/test_pagerank.py
ChuckHastings: Same as above.  I was trying to see what it would take to get the directed graph to pass.  Changed it back.

python/cugraph/tests/test_modularity.py
ChuckHastings: done

ChuckHastings:
> > page rank mostly works but needs some exploration.
> 
> For Pagerank, I'd suggest checking that the graph is properly loaded in Nx (ie. in a way that it is the same data as in cuGraph). Nx often creates an undirected graph by default (ie. a `Graph` instead of DiGraph) and silently gets rid of singletons. This was the problem in
> #482 for instance.
> 
> A good way to check that would be to add a test for the `read_csv_for_nx` function. You could then ask for the number of edges and vertices in Nx and CuGraph and make sure they match.

ChuckHastings:
> > page rank mostly works but needs some exploration.
> 
> For Pagerank, I'd suggest checking that the graph is properly loaded in Nx (ie. in a way that it is the same data as in cuGraph). Nx often creates an undirected graph by default (ie. a `Graph` instead of DiGraph) and silently gets rid of singletons. This was the problem in
> #482 for instance.
> 
> A good way to check that would be to add a test for the `read_csv_for_nx` function. You could then ask for the number of edges and vertices in Nx and CuGraph and make sure they match.

I added a test_read_csv_for_nx.  I wasn't sure the best way to accomplish this.  For now I just compare to the old mmread call.  Both functions return a scipy.sparse.coo_matrix.  I compare the size and shape of the matrices.  Because we do floating point parsing differently, I compute the differences between the floating point values and compare them to epsilon (min value / 1000.0 is what I chose).  They come out identical.

I suspect the page rank issue is more subtle than that.  We can create a story for some subsequent release to test the page rank on the directed graph.  As I recall (I did the page rank testing last week), the basic page rank worked... but some of the personalization options got different results than networkx on the directed graph.

ChuckHastings:
Based on the comments, I think that leaves us with future stories for:
 * test page_rank with asymmetric graph
 * test katz centrality with asymmetric graph
 * test triangle counting with asymmetric graph

ChuckHastings:
Added #533, #534, #535 to address tests that still need to be done.

afender:
rerun tests

PR #547, opened 2019-10-07
[WIP] remove adjacency list conversions

Iroy30:
Remove graph data conversions from cpp. Include data validation/checking in python


python/cugraph/structure/graph_wrapper.pyx
```diff
@@ -74,6 +74,16 @@ def add_edge_list(graph_ptr, source_col, dest_col, value_col=None):
     cdef uintptr_t graph = graph_ptr
     cdef gdf_graph * g = <gdf_graph*> graph
 
+    # Checks in python
```
Iroy30: Migrating checks from cpp to python:
-> Checks if src col and dest col lengths are same
-> Typecasts dtype of src and dest to int32 if not already int32
-> Typecasts dtype of value(weights) to float32 if not already float32 or float64
-> Null_count checking to be added?

python/cugraph/structure/graph_wrapper.pyx
```diff
@@ -144,6 +154,13 @@ def add_adj_list(graph_ptr, offset_col, index_col, value_col=None):
     cdef uintptr_t graph = graph_ptr
     cdef gdf_graph * g = <gdf_graph*> graph
 
+    if offset_col.dtype.type is not np.int32:
```
Iroy30: Similar checks as adding edgelist.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -208,7 +208,7 @@ gdf_error gdf_pagerank_impl (gdf_graph *graph,
     GDF_REQUIRE(personalization_subset->null_count == 0 , GDF_VALIDITY_UNSUPPORTED );
     GDF_REQUIRE(personalization_values->null_count == 0 , GDF_VALIDITY_UNSUPPORTED );
   }
-  GDF_REQUIRE( graph->edgeList != nullptr, GDF_VALIDITY_UNSUPPORTED );
+  /*GDF_REQUIRE( graph->edgeList != nullptr, GDF_VALIDITY_UNSUPPORTED );
```
Iroy30: -> Remove edgelist checks since user input could be in the form of adjacency list(edgelist might be nullptr in that case)
-> Remove pagerank column checks since the column is initialized in the pagerank_wrapper.pyx with pre-set specifications(size, dtype, data initialized to zeros, etc) which is redundantly verified in cpp.
Note: Commented lines to be removed after initial review.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -217,15 +217,16 @@ gdf_error gdf_pagerank_impl (gdf_graph *graph,
...
-  int m=pagerank->size, nnz = graph->edgeList->src_indices->size, status = 0;
+  */
+  //int m=pagerank->size, nnz = graph->edgeList->src_indices->size, status = 0;
+  int m=pagerank->size, nnz = graph->transposedAdjList->indices->size, status = 0;
   WT *d_pr, *d_val = nullptr, *d_leaf_vector = nullptr;
   WT res = 1.0;
   WT *residual = &res;
 
-  if (graph->transposedAdjList == nullptr) {
+  /*if (graph->transposedAdjList == nullptr) {
```
Iroy30: Tansposed adjacency list already added in python wrapper before calling cpp pagerank

cpp/src/link_analysis/pagerank.cu
Iroy30: -> Remove edgelist checks since user input could be in the form of adjacency list(edgelist might be nullptr in that case)
-> Remove pagerank column checks since the column is initialized in the pagerank_wrapper.pyx with pre-set specifications(size, dtype, data initialized to zeros, etc) which is redundantly verified in cpp.
Note: Commented lines to be removed after initial review.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -276,14 +277,17 @@ gdf_error gdf_pagerank(gdf_graph *graph, gdf_column *pagerank,
   //
   //  If csr doesn't exist, create it.  Then check type to make sure it is 32-bit.
   //
-  GDF_REQUIRE(graph->adjList != nullptr || graph->edgeList != nullptr, GDF_INVALID_API_CALL);
+  /*GDF_REQUIRE(graph->adjList != nullptr || graph->edgeList != nullptr, GDF_INVALID_API_CALL);
```
Iroy30: -> Remove adding adjacency list since transposed adjacency list is already added/called in python wrapper and this data is required by and passed to cugraph::HT_matrix_csc_coo and cugraph::pagerank calls.
-> Remove adjacency list checks. The offsets and indices are ensured to be int32 since it is checked and typecasted while adding the graph itself (within add_edge_list and add_adj_list functions in graph_wrapper.pyx)

cpp/src/traversal/bfs.cu
```diff
@@ -471,14 +471,16 @@ namespace cugraph {
 } // end namespace cugraph
 
 gdf_error gdf_bfs(gdf_graph *graph, gdf_column *distances, gdf_column *predecessors, int start_vertex, bool directed) {
-  GDF_REQUIRE(graph->adjList != nullptr || graph->edgeList != nullptr, GDF_INVALID_API_CALL);
-  gdf_error err = gdf_add_adj_list(graph);
+  //GDF_REQUIRE(graph->adjList != nullptr || graph->edgeList != nullptr, GDF_INVALID_API_CALL);
+
+  /*gdf_error err = gdf_add_adj_list(graph);
```
Iroy30: -> Remove adding adj list, already added in wrapper before calling gdf_bfs.
-> Remove int32 checks as it is ensured when adding graph itself.
-> Remove distance and predecessor column checks since the column is initialized in the bfs_wrapper.pyx with pre-set specifications(size, dtype as int32, data initialized to zeros, etc) which is redundantly verified in cpp.

cpp/src/traversal/sssp.cu
```diff
@@ -314,7 +316,7 @@ gdf_error gdf_sssp(gdf_graph* gdf_G,
                 GDF_INVALID_API_CALL);
   }
 
-  gdf_error err = gdf_add_adj_list(gdf_G);
+  /*gdf_error err = gdf_add_adj_list(gdf_G);
```
Iroy30: -> Remove adding of adjacency list (added in sssp wrapper) and int32 checks (ensured when adding graph data)

cpp/src/link_analysis/pagerank.cu
afender: I agree regarding dropping edge list checks but I think we should keep PageRank ones in case someone calls it from the c++ API. It will also help debugging if we introduce an issue at the python level.

cpp/src/link_analysis/pagerank.cu
afender: agreed

cpp/src/link_analysis/pagerank.cu
afender: lgtm

cpp/src/traversal/sssp.cu
afender: it seems we should keep GDF_REQUIRE(gdf_G->adjList, GDF_INVALID_API_CALL);

cpp/src/traversal/bfs.cu
afender: We should keep distance and predecessor column checks in case someone calls it from the c++ API. It will also help debugging if we introduce an issue at the python level.
```
  GDF_REQUIRE(distances->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);
  GDF_REQUIRE(predecessors->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);	  GDF_REQUIRE(predecessors->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);```

cpp/src/link_analysis/pagerank.cu
afender: it seems we should have a GDF_REQUIRE(gdf_G->transposedAdjList, GDF_INVALID_API_CALL);

python/cugraph/structure/graph_wrapper.pyx
afender: -> Typecasts dtype of value(weights) to float32 if not already float32 or float64
We should allow othertype, the cast should happen at the time an algo requiring fp32/64 is called

afender: -> Typecasts dtype of src and dest to int32 if not already int32
we may want to leverage renumbering for that

afender: Overall, I'd recommend to first update the python graph structure before adding more automation at the python level

Iroy30: The edge weights reside in the form of a cpp gdf_column in the graph structure. If we save the graph with user-input datatypes into our gdf_G structure and only typecast it when algo requiring fp32/64 is called, since the api takes in only pointer to gdf_graph, typecasting could either be done by manipulating the graph edge weights itself (which is, changing the user-input in saved graph, would require a cpp function to be called) or sending in edge weights as a separate column (not ideal)

Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

PR #554, opened 2019-10-10
Namespace back-end

afender:
added
cugraph::snmg
cugraph::db
cugraph::detail

remains
cugraph (for the API)
fix some files that didn't have a namespace


afender:
rerun tests

Iroy30:
is it cugraph::sort or cugraph::detail::sort?

Iroy30:
lgtm

afender:
> is it cugraph::sort or cugraph::detail::sort?

I think I went with `cugraph::sort` because the sort is being hosted in cuGraph without really being a cuGraph internal implementation. I thought it would be simpler in case it moves to raft (so we would just switch to  `raft::sort`).

afender:
Blocked by #573

PR #559, opened 2019-10-13
[BUG] Added RMM init check so RMM free APIs are not called if not initialized

rlratzel:
Only call RMM free if the RMM session has been initialized.

This resolves almost all the cugraph notebook test failures (still an issue with `Renumbering-test`, may have to handle separately).

I'm not sure if this is the right approach as a client of RMM, hoping @kkraus14 has some thoughts.  Also unsure if this is the best place to add this check, hoping @afender and @seunghwak have feedback.

### DIFF
--- a/cpp/include/rmm_utils.h
+++ b/cpp/include/rmm_utils.h
@@ -37,6 +37,11 @@
   RMM_TRY_THROW( RMM_REALLOC((ptr), (sz), (stream)) ) \
 }
 
+// TODO: temporarily wrapping RMM_FREE in a rmmIsInitialized() check to work
+// around the RMM session being finalized prior to this call. A larger
+// refactoring will need to be done to eliminate the need to do this, and
+// calling RMM APIs directly should likely also be removed in favor of working
+// with a higher-level abstraction that manages RMM properly (eg. cuDF?)
 #define ALLOC_FREE_TRY(ptr, stream){              \
   if(rmmIsInitialized((rmmOptions_t*) NULL)) {    \
     RMM_TRY_THROW( RMM_FREE( (ptr), (stream) ) )  \
--- a/cpp/include/rmm_utils.h
+++ b/cpp/include/rmm_utils.h
@@ -18,7 +18,7 @@
 #include <sstream>
 #include <stdexcept>
 
-#define RMM_TRY_THROW( call )  if ((call)!=RMM_SUCCESS) \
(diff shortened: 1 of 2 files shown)


seunghwak:
I think this resembles the problem I have encountered in testing graph algorithms with all four RMM configurations.

https://github.com/rapidsai/cugraph/blob/branch-0.11/python/cugraph/tests/test_bfs.py

```
# Test all combinations of default/managed and pooled/non-pooled allocation
@pytest.mark.parametrize('managed, pool',
                         list(product([False, True], [False, True])))
@pytest.mark.parametrize('graph_file', DATASETS)
def test_bfs(managed, pool, graph_file):
    gc.collect()

    rmm.finalize()
    rmm_config.use_managed_memory = managed
    rmm_config.use_pool_allocator = pool
    rmm_config.initial_pool_size = 2 << 27
    rmm.initialize()
```

Garbage collection may or may not happen when every object gets out-of-scope, and without explicit call to gc.collect() before rmm.finalize(), actual free happened after rmm.finalize(), and this leads to weird errors in the previous case. I'm suspecting that this is indeed a case here (explicitly invoking del G serves a similar purpose).

rlratzel:
> Garbage collection may or may not happen when every object gets out-of-scope, and without explicit call to gc.collect() before rmm.finalize(), actual free happened after rmm.finalize(), and this leads to weird errors in the previous case. I'm suspecting that this is indeed a case here (explicitly invoking del G serves a similar purpose).

I agree, and I think the `gc.collect()` pattern used in the unit tests are why we never saw CI failures. The graph notebooks do not do this, hence the issue.

cpp/include/rmm_utils.h
```diff
@@ -37,6 +37,8 @@
   RMM_TRY_THROW( RMM_REALLOC((ptr), (sz), (stream)) ) \
 }
 
-#define ALLOC_FREE_TRY(ptr, stream){            \
-  RMM_TRY_THROW( RMM_FREE( (ptr), (stream) ) )  \
+#define ALLOC_FREE_TRY(ptr, stream){              \
+  if(rmmIsInitialized((rmmOptions_t*) NULL)) {    \
```
seunghwak: Could you just add comments stating this is a temporary solution for this release and this needs to be revisited as we finalize our longer term strategy to address this issue.

rlratzel: Yes, good idea, will do.

PR #562, opened 2019-10-15
[REVIEW] Remove pyarrow dependency so we inherit the one cudf uses

trxcllnt:
This is related to cuDF's [upgrade to Arrow-0.15.0](https://github.com/rapidsai/cudf/pull/3084). I'm not sure how much cuGraph is relying on arrow-cpp/pyarrow, but I ran the tests after rebuilding and they seem to pass.


trxcllnt:
CI seems to be failing because https://github.com/rapidsai/cugraph/pull/564

afender:
rerun tests

kkraus14:
@afender Can we merge this to unblock https://github.com/rapidsai/cudf/pull/3084?

PR #564, opened 2019-10-16
[REVIEW] Update cudf  type aliases to the new names

trxcllnt:
Updates the type aliases from cudf following the merger of https://github.com/rapidsai/cudf/pull/3029

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## New Features
 
 ## Improvements
+- PR #564 Update cudf type aliases
 
 ## Bug Fixes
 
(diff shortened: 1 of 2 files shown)


seunghwak:
Is this for the cython part only?

gdf_size_type is used in cpp files as well.
e.g.
https://github.com/rapidsai/cugraph/blob/branch-0.11/cpp/src/cores/core_number.cu#L118

trxcllnt:
Yes this is only for Cython, since https://github.com/rapidsai/cudf/pull/3029 keeps the `gdf_*` aliases in `cudf/types.h`.

PR #569, opened 2019-10-24
[REVIEW] Exceptions

### LINKED ISSUE #486: [FEA] Add MACROS to support exception based error handling
RAPIDS is switching from error code based error handling to exception based error handling. cuGraph does not have utility functions to support this, so individual programmers need to develop their own way to throw exceptions: this requires additional coding effort and leads to inconsistencies.

cuDF already developed a solution for this (https://github.com/rapidsai/cudf/blob/branch-0.10/cpp/src/utilities/error_utils.hpp).

```
#define STRINGIFY_DETAIL(x) #x
#define CUDF_STRINGIFY(x) STRINGIFY_DETAIL(x)

struct logic_error : public std::logic_error {
  logic_error(char const* const message) : std::logic_error(message) {}

  logic_error(std::string const& message) : std::logic_error(message) {}

  // TODO Add an error code member? This would be useful for translating an
  // exception to an error code in a pure-C API
};

/**---------------------------------------------------------------------------*
 * @brief Macro for checking (pre-)conditions that throws an exception when  
 * a condition is violated.
 * 
 * Example usage:
 * 
 * @code
 * CUDF_EXPECTS(lhs->dtype == rhs->dtype, "Column type mismatch");
 * @endcode
 *
 * @param[in] cond Expression that evaluates to true or false
 * @param[in] reason String literal description of the reason that cond is
 * expected to be true
 * @throw cudf::logic_error if the condition evaluates to false.
 *---------------------------------------------------------------------------**/
#define CUDF_EXPECTS(cond, reason)                           \
  (!!(cond))                                                 \
      ? static_cast<void>(0)                                 \
      : throw cudf::logic_error("cuDF failure at: " __FILE__ \
                                ":" CUDF_STRINGIFY(__LINE__) ": " reason)

/**---------------------------------------------------------------------------*
 * @brief Error macro that throws an exception
 * 
 * Example usage:
 * 
 * @code
 * CUDF_FAIL("Non-arithmetic operation is not supported");
 * @endcode
 *
 * @param[in] reason String literal description of the reason
 * @throw cudf::logic_error if the condition evaluates to false.
 *---------------------------------------------------------------------------**/
#define CUDF_FAIL(reason)      				     \
    throw cudf::logic_error("cuDF failure at: " __FILE__     \
                            ":" CUDF_STRINGIFY(__LINE__) ": " reason)
```

We need to bring this cugraph, replace CUDF and cudf with CUGRAPH and cugraph, respectively, and start using this in any new code development.

I haven't considered any other approaches.

### LINKED ISSUE #498: [FEA] Exception based error handling
RAPIDS is switching from error code based error handling to exception-based error handling.

- [x] Add #486
- [x] Check if some cugraph-specific error helpers are needed
- [x] The new C++ API should have an exception-based error handling from the start
- [x] Add `except +` keyword to the cython wrappers
- [ ] Upgrade the documentation

### Note on CUDF update

The `gdf_error` codes are deprecated. Instead, `cudf` functions should prefer throwing exceptions.

In place of the `GDF_REQUIRES` convenience macro, the `CUDF_EXPECTS` macro is provided in `cudf/cpp/src/utilities/error_utils.hpp`. 

Similar to `GDF_REQUIRES`, `CUDF_EXPECTS` will throw an exception if a condition is violated. The first argument is the conditional expression that can be evaluated as a `boolean` expression and is expected to resolve to `true` under normal conditions. If the conditional evaluates to `false`, then an error has occurred and an instance of `cudf::logic_error` is thrown. The second argument to `CUDF_EXPECTS` is a short description of the error that has occurred if the conditional is false. 

Example usage: 

```
CUDF_EXPECTS(lhs->dtype == rhs->dtype, "Column type mismatch");
```

There are times where a particular code path, if reached, should indicate an error no matter what. For example, in the `default` case of a `switch` statement where the only valid code paths are in one of the `case` statements. 

For these cases, the `CUDF_FAIL` convenience macro should be used. This is effectively the same as doing `CUDF_EXPECTS(false, reason)`. 

Example:

```
CUDF_FAIL("This code path should not be reached.");
```

Checking for the succesful completion of CUDA runtime API functions should continue to be done via the `CUDA_TRY` macro. Note that this macro has been updated to throw an exception of type `cudf::cuda_error` if the return value of the CUDA API does not return `cudaSuccess`. The thrown exception will include a description of the CUDA error code that occurred in it's `what()` message.

Example:

```
CUDA_TRY( cudaMemcpy(&dst, &src, num_bytes) );

afender:
Close #498 
Close #486 

RAPIDS is switching from error code based error handling to exception-based error handling.

The design is closely following cuDF solution. 
C++ exception type and message are propagated to the python layer, cuda and cugraph calls have specific macros simplifying the process. Expected test failures (C++ and python) have been updated to check exceptions instead of error codes. 

File and line, as well as error messages, are properly transiting through all layers. For instance, a misusage of the python API triggering a C++ exception will result in :
`RuntimeError: CUGRAPH failure at: /home/afender/cugraph/cpp/src/structure/cugraph.cu:175: Invalid API parameter`

I will do another PR with the API changes (after confirming the depreciation plan):
- remove gdf_error returns in API
- upgrade API doc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## New Features
 
 ## Improvements
+- PR #569 Exceptions
 
 ## Bug Fixes
 
--- a/cpp/src/centrality/katz_centrality.cu
+++ b/cpp/src/centrality/katz_centrality.cu
@@ -33,14 +33,14 @@ gdf_error gdf_katz_centrality(gdf_graph *graph,
                               double tol,
                               bool has_guess,
                               bool normalized) {
-  GDF_REQUIRE(graph->adjList != nullptr || graph->edgeList != nullptr, GDF_INVALID_API_CALL);
+  CUGRAPH_EXPECTS(graph->adjList != nullptr || graph->edgeList != nullptr, "Invalid API parameter");
   gdf_error err = gdf_add_adj_list(graph);
   if (err != GDF_SUCCESS)
     return err;
-  GDF_REQUIRE(graph->adjList->offsets->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);
-  GDF_REQUIRE(graph->adjList->indices->dtype == GDF_INT32, GDF_UNSUPPORTED_DTYPE);
-  GDF_REQUIRE(katz_centrality->dtype == GDF_FLOAT64, GDF_UNSUPPORTED_DTYPE);
-  GDF_REQUIRE(katz_centrality->size == graph->numberOfVertices, GDF_COLUMN_SIZE_MISMATCH);
+  CUGRAPH_EXPECTS(graph->adjList->offsets->dtype == GDF_INT32, "Unsupported data type");
+  CUGRAPH_EXPECTS(graph->adjList->indices->dtype == GDF_INT32, "Unsupported data type");
+  CUGRAPH_EXPECTS(katz_centrality->dtype == GDF_FLOAT64, "Unsupported data type");
+  CUGRAPH_EXPECTS(katz_centrality->size == graph->numberOfVertices, "Column size mismatch");
 
   const bool isStatic = true;
   using HornetGraph = hornet::gpu::HornetStatic<int>;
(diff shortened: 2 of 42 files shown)


afender:
Blocked by #573

afender:
rerun tests

PR #575, opened 2019-10-29
Added updated benchmark files that use new func wrapper pattern and asvdb

rlratzel:
Adding initial benchamark updates needed for nightly ASV E2E runs. This is still somewhat of a WIP and may get updated soon and frequently as our benchmarking tools evolve.

### DIFF
--- a/python/utils/benchmark.py
+++ b/python/utils/benchmark.py
@@ -1,0 +1,94 @@
+import sys
+#from time import process_time_ns   # only in 3.7!
+from time import clock_gettime, CLOCK_MONOTONIC_RAW
+
+class Nop:
+    def __getattr__(self, attr):
+        return Nop()
+    def __getitem__(self, key):
+        return Nop()
+    def __call__(self, *args, **kwargs):
+        return Nop()
+nop = Nop()
+
+
+# wrappers
+def noStdoutWrapper(func):
+    def wrapper(*args):
+        prev = sys.stdout
+        sys.stdout = open(os.devnull, 'w')
+        try:
+            retVal = func(*args)
+            sys.stdout = prev
+            return retVal
+        except Exception:
+            sys.stdout = prev
+            raise
+    return wrapper
+
+
+def logExeTime(func):
+    def wrapper(*args):
+        retVal = None
+        perfData = logExeTime.perfData
+        try:
+            #st = process_time_ns()
+            st = clock_gettime(CLOCK_MONOTONIC_RAW)
+            retVal = func(*args)
+        except Exception as e:
+            perfData.append((func.__name__, "ERROR: %s" % e))
+            return
+        #exeTime = (process_time_ns() - st) / 1e9
+        exeTime = clock_gettime(CLOCK_MONOTONIC_RAW) - st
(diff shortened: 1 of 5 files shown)


python/utils/run_benchmarks.py
```diff
@@ -1,82 +1,85 @@
...
-                {"obj": G,
-                 },
-                }
-    return algoData
-
-
-def loadDataFile(file_name, file_type, delimiter=' '):
+def getBenchmarks(G, edgelist_gdf, args):
+    benches = [
+        Benchmark(name="pagerank",
```
afender: So, when adding a new feature in cuGraph, we just need to :
1. add it in `getBenchmarks` in order to see it in the benchmark runs
2. update `parseCLI` with relevant parameters specific to the new feature

python/utils/run_benchmarks.py
```diff
@@ -180,71 +148,84 @@ def parseCLI(argv):
...
+    return list(getBenchmarks(nop, nop, nop).keys())
 
 ###############################################################################
 if __name__ == "__main__":
     perfData = []
     args = parseCLI(sys.argv[1:])
-    delimiter = {"space": ' ', "tab": '\t'}[args.delimiter]
 
+    # set algosToRun based on the command line args
     allPossibleAlgos = getAllPossibleAlgos()
```
afender: How would you run the same algo twice with different parameters? Let's say we want to benchmark SSSP a few times changing the `source`. Does this involve running the whole script twice using `parseCLI`or is there a smarter way?

python/utils/benchmark.py
```diff
@@ -0,0 +1,94 @@
...
+        for wrapper in runWrappers:
+            self.func = wrapper(self.func)
+            self.func.__name__ = self.name
+
+    def run(self):
+        return self.func(*self.args)
+
+
+class Benchmark(WrappedFunc):
+    wrappers = [logExeTime, printLastResult]
```
afender: Is this where we will add a `logMemPeak` in the future? Or is memory going to be tracked outside of this script?

python/utils/run_benchmarks.py
afender: Correct? I think I'll add that to the checklist we should go through when adding a new algo.

rlratzel: yes, that's correct

python/utils/run_benchmarks.py
rlratzel: I think right now you'd run the script multiple times with different `--source` like you said. You can also have it run only the SSSP algo using the `--algo` arg, so that way you wouldn't need to re-run every algo just to get multiple SSSP runs.  I think I can improve this use case in a future PR though.

python/utils/benchmark.py
rlratzel: I'm planning on adding the other metrics in a separate PR, but yes I think this is where we'll add additional wrappers to measure mem usage, etc.  The idea behind the wrappers is that they can be added/removed easily without modifications to the benchmarks themselves, so adding those should hopefully be quick.

PR #580, opened 2019-11-05
[BUG] Changed hardcoded test output file to a generated tempfile file name

rlratzel:
Changed hardcoded `~/pagerank.csv` test output file to a generated tempfile to avoid permission problems for certain test environments that cannot write to ~ (such as the nightly unit test runs).

I'm not sure if users expect the resulting output file to remain (since the previous version never deleted it after the test ended), but that can be something done here as well if necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - PR #570 Temporarily disabling 2 DB tests
 - PR #573 Fix pagerank test and symmetrize for cudf 0.11
 - PR #574 dev env update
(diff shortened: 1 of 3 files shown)


rlratzel:
rerun tests

rlratzel:
update: @mike-wendt just informed me that the `conda install` timeout we're experiencing on the failed GPU test is a problem on Anaconda's side, so that check may have to be re-run later when it gets resolved.

rlratzel:
rerun tests

rlratzel:
rerun tests

rlratzel:
rerun tests

Issue #515, opened 2019-09-24
Refactor Graph


BradReesWork:
- [ ] Error Handling
- [ ] C++ API
- [ ] Style

Python Class:

- as a Python class
-- Handle:
-- Undirected
-- Directed
-- Bipartite
-- Dynamic
-- Multi
- Renumbering
- Symmetrizing
- Strings

BradReesWork:
EPIC being closed - too vague

Issue #307, opened 2019-05-28
[FEA] [EPIC] New C++ API 


afender:
Cugraph should have a modern C++ API instead of a C API. 
 
- [x] Encapsulate the API in `namespace cugraph` (the backend needs to move to another namespace) #540 
- [x] get rid of `gdf_error` #498 
- [x] remove `gdf_` prefix
- [x] library handle (from RAFT) #523 
- [x] remove `gdf_column` #538
- [x] redo the graph object  
- [x] transition guide

afender:
From Seungwha, about namespaces : 
If we're following cudf (https://github.com/rapidsai/cudf/blob/branch-0.8/cpp/docs/TRANSITIONGUIDE.md),

We should use cugraph:: for public API, cugraph::detail:: for cugraph internal implementations, and anonymous namespace for implementations restricted to a single translation unit.

afender:
From @BradReesWork, about cugraph handle :
get the device attributes and add a required statement on supported gpu types.

afender:
#772 for the graph object

afender:
Closing,  all of the linked pieces are done.

PR #584, opened 2019-11-07
[WIP][skip-ci] Fea-ext-ktruss-max - adding a wrapper to finding maximal k-truss of a graph

ogreen:
This WIP is intended to let others see the progress on the wrapping and integration of finding the maximal k-truss of a graph using cuHornet.

Closes #577

PR #587, opened 2019-11-11
[REVIEW] API upgrade 

### LINKED ISSUE #540: [FEA] Namespaces
### `cugraph` Namespace

Following libcudf C++ style, all external APIs should be placed in the `cugraph` namespace. For example:

```
namespace cugraph{

  void some_function(...);

} // namespace cugraph

```

### `cugraph::detail` Namespace

Utility functions that are for use only in cugraph internals, but can be reused in many places should be placed in the `cugraph::detail` namespace. For example:

```
namespace cugraph{
namespace detail{
void reusable_utility_function(...);
} // namespace detail
} // namespace cugraph
```

### Anonymous Namespaces

All functions specific to a translation unit should be placed in an anonymous namespace.

Example:
```
namespace {
void some_specific_helper_function(...);
} // namespace
```

afender:
### C++ changes
- Got rid of `gdf_error` as C++ API return type : now using exceptions for errors instead
- C++ API is encapsulated in namespace `cugraph::`
- `gdf_graph` is now `Graph`: we will make `Graph` better in future PRs, but the API of analytics won't change much.
- got rid of `gdf_` prefix in C++ API

```c
// depreciated API 
gdf_error err = gdf_pagerank(<gdf_graph>, ...) 
// new API 
cugraph::pagerank(<cugraph::Graph>, ...) 
```

### Collateral
- Python bindings use the new API.
- C++ tests use the new API
- upgraded c++ documentation

### Extras

- simplified public headers : `cugraph.h` is now the only header needed. It aggregates a smaller number of public sub header
- made `style.sh` executable

Close #540 #498 
New milestone for #307  and #515


afender:
@ogreen / @rlratzel can one of you help reviewing this so we can merge it and unblock #588 and #584?

cpp/src/converters/COOtoCSR.cuh
```diff
@@ -143,12 +141,12 @@ gdf_error ConvertCOOtoCSR(T* sources, T* destinations, int64_t nnz, CSR_Result<T
     ALLOC_FREE_TRY(unique, stream);
```
ogreen: Will an error be thrown by ALLOC_FREE_TRY? If so, then this is fine.
Otherwise, we need to revisit its usage.

ogreen:
Looks good. Just a few minor comments and I am good to accepting it.

cpp/src/converters/COOtoCSR.cuh
afender: Yes, this is a wrapper from `rmm_utils.h`. It will end up calling `cugraph::detail::throw_rmm_error(status, __FILE__, __LINE__);` when RMM returns a failure.

cpp/src/tests/pagerank/pagerank_test.cu
```diff
@@ -113,26 +113,26 @@ class Tests_Pagerank : public ::testing::TestWithParam<Pagerank_Usecase> {
     col_dest = create_gdf_column(cooColInd);
     col_pagerank = create_gdf_column(pagerank);
 
-    CUGRAPH_TRY(gdf_edge_list_view(G.get(), col_src.get(), col_dest.get(), nullptr));
-    CUGRAPH_TRY(gdf_add_transposed_adj_list(G.get()));
+    cugraph::edge_list_view(G.get(), col_src.get(), col_dest.get(), nullptr);
```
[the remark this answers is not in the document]
afender: Every time a function was returning an error it should now throw one.
In this case, `CUGRAPH_EXPECTS`calls in `edge_list_view` will throw errors when user provider parameters of this API function are not correct.

afender:
Addressed 
- error handling
- C++ API 

In #500, #501, #502, #504, #505, #506, #507, #508, #509, #510, #511, #512,# 513, #514

PR #588, opened 2019-11-13
[REVIEW] Python graph class

### LINKED ISSUE #581: [FEA] auto renumbering and unrenumbering
Currently, users need to call renumbering when their vertices are not between [0,  V). This relies too much on the assumption that the user is familiar with the renumbering concept and internal representation. 

CuGraph should:

- automatically renumber (and store) the user's input at the time an edge list is loaded
- automatically unrenumber vertex IDs when returning any analytics results to the user

Sugested API change : 
```
G.add_edges( ..., renumber=True)
optional renumbering parameter True by default
```

### LINKED ISSUE #583: [BUG] Better support DF in the API
cuGraph uses mostly Series in the API while the rest of RAPIDS APIs leverage DataFrame more. It makes end-to-end workflow awkward. A good example of this is `cudf.read_csv` returning a dataFrame and `add_edge_list` taking only Series which requires a cast.

We should use a dataFrame in cuGrpah API each time we accept/return 2 Series or more of the same size.

### LINKED ISSUE #590: [FEA] Data type conversion
Users should be allowed to store a variety of edge attributes types (int 32 / 64, float 32, 64 etc.) in the graph at the Python level. Problem : the vast majority of internal CUDA kernels and C++ algorithms work for specific data types. In the long run, we may add more supported type at the CUDA/C++ layer. A good temporary solution is to simply do a type conversion (cast) for now.

At the time a user calls an analytics, the python layer is aware of both the user data type(s) and the supported type(s) by the requested analytics. It should automatically try to do the proper type of conversion in order to run it.

We should relax the type check at the python level on the edge list/adj list and add proper data type conversions in analytics

 #579 will help to automate this policy but remains independent

### LINKED ISSUE #594: [BUG] Depreciation for add_adjList and add_edgeList
These are not matching any Nx function and will be replaced by the newer and greater `from_cudf_edgelist(df, source='u', destination='v', edge_attr=None, create_using=None, renumber=True, property=None)`

1. Add a warning in these functions saying it will be depreciated after 0.11 release.
2. Upgrade all internal code and tests to call the new functions
3. Remove the old ones during 0.12

Iroy30:
first steps on python side for #515 and #192 
closes #581
closes #583 
closes #594
closes #590


python/cugraph/structure/graph.py
```diff
@@ -37,32 +58,25 @@ def __init__(self):
...
         """
-        self.delete_edge_list()
-        self.delete_adj_list()
-        self.delete_transposed_adj_list()
+        self.edgelist = None
+        self.adjlist = None
+        self.transposedadjlist = None
 
-    def add_edge_list(self, source_col, dest_col, value_col=None, copy=False):
+    def add_edge_list(self, input_df):
```
afender: We need something closer to this : https://networkx.github.io/documentation/stable/reference/generated/networkx.convert_matrix.from_pandas_edgelist.html#networkx.convert_matrix.from_pandas_edgelist

In addition, we should accept a renumbering parameter (`True` by default to call renumbering before storing the edge list) and `property` info (to set the flag seen at 61-69).

python/cugraph/structure/graph.py
```diff
@@ -171,191 +171,35 @@ def view_edge_list(self):
...
-        graph_wrapper.delete_edge_list(self.graph_ptr)
-
         # decrease reference count to free memory if the referenced objects are
         # no longer used.
-        self.edge_list_source_col = None
-        self.edge_list_dest_col = None
-        self.edge_list_value_col = None
-
-    def add_adj_list(self, offset_col, index_col, value_col=None, copy=False):
-        """
```
afender: I think documentation got accidentally deleted

python/cugraph/structure/graph.py
```diff
@@ -171,191 +171,35 @@ def view_edge_list(self):
...
-        self.adj_list_index_col = tmp_index_col
-        self.adj_list_value_col = tmp_value_col
+    def add_adj_list(self, offset_col, index_col, value_col=None):
+        if self.edgelist is not None or self.adjlist is not None:
+            raise Exception('Graph already has values')
+        self.adjlist = Graph.AdjList(offset_col, index_col, value_col)
 
     def view_adj_list(self):
-        """
-        Display the adjacency list. Compute it if needed.
```
afender: I think documentation got accidentally deleted

python/cugraph/structure/graph.py
```diff
@@ -171,191 +171,35 @@ def view_edge_list(self):
...
-            graph_wrapper.view_adj_list(self.graph_ptr)
-
-        return offset_col, index_col, value_col
+        if self.adjlist is None:
+            graph_wrapper.view_adj_list(self)
+        return self.adjlist.offsets, self.adjlist.indices, self.adjlist.weights
 
     def delete_adj_list(self):
-        """
-        Delete the adjacency list.
```
afender: I think documentation got accidentally deleted

python/cugraph/structure/graph.py
```diff
@@ -20,13 +20,34 @@ def null_check(col):
...
-
 class Graph:
+
+    class EdgeList:
+        def __init__(self, source, dest, value=None, renumber_map=None):
+            self.renumber_map = renumber_map
+            df = cudf.DataFrame()
+            df['src'] = source
+            df['dst'] = dest
+            self.weights = False
```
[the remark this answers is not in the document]
afender: Or are we going to have another class fo this?

python/cugraph/community/subgraph_extraction.py
```diff
@@ -52,11 +52,11 @@ def subgraph(G, vertices):
 
     null_check(vertices)
 
-    result_graph = Graph()
+    result_graph = DiGraph()
```
afender: Can you please point me to where DiGraph is implemented?

Iroy30: https://github.com/Iroy30/cugraph/blob/763aa86c45b945cc59b1e34e02d37368d65a8b29/python/cugraph/structure/graph.py#L426

python/cugraph/structure/graph.py
Iroy30: yes, i would be adding.updating documentation

python/cugraph/structure/graph.py
```diff
@@ -20,13 +20,34 @@ def null_check(col):
     if col.null_count != 0:
         raise ValueError('Series contains NULL values')
 
-
 class Graph:
+
+    class EdgeList:
+        def __init__(self, source, dest, value=None, renumber_map=None):
```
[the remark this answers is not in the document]
Iroy30: are you referring to the api for adding or the name of storage variable in self?

python/cugraph/structure/graph.py
Iroy30: i thought we were following https://networkx.github.io/documentation/stable/reference/classes/generated/networkx.Graph.add_edges_from.html#networkx.Graph.add_edges_from 
with ebunch_to_add being our dataframe

afender: The `from_panda_edgelist` style looks more suited for cudf dataframe and cuGraph's main workflow. In case the users have a DF with 10 columns they can still pass it to cuGraph by telling what columns are relevant.

python/cugraph/structure/graph.py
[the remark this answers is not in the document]
afender: Ideally, everything should be consistent but the most important part is the API. We can have `src` and `dst` or `source` and `destination` but mixing is not ideal. For the wegiths (values) `edge_attr` looks more generic and matches Nx.

Iroy30:
rerun tests

python/cugraph/structure/graph.py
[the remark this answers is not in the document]
Iroy30: will be edited as a part of #589

python/cugraph/structure/graph.py
```diff
@@ -20,13 +20,34 @@ def null_check(col):
...
+            if value is not None:
+                self.weights = True
+                df['weights'] = value
+            self.edgelist_df = df
+
+    class AdjList:
+        def __init__(self, offsets, indices, value=None):
+            self.offsets = offsets
+            self.indices = indices
+            self.weights = value #Should de a daftaframe for multiple weights
```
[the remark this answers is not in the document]
Iroy30: will be edited as a part of #589

python/cugraph/structure/graph.py
```diff
@@ -112,39 +126,25 @@ def add_edge_list(self, source_col, dest_col, value_col=None, copy=False):
...
-        # Create temporary references first as the member variables should not
-        # be updated on failure.
-        if copy is False:
-            tmp_source_col = source_col
-            tmp_dest_col = dest_col
-            tmp_value_col = value_col
+
+        if self.edgelist is not None or self.adjlist is not None:
+            raise Exception('Graph already has values')
+        source_col = input_df[input_df.columns[0]]
```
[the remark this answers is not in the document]
Iroy30: fixed

python/cugraph/structure/graph.py
Iroy30: fixed. Renumber is set to False by default. User needs to set it to True for renumbering.

python/cugraph/structure/graph.py
Iroy30: updated

python/cugraph/structure/graph.py
Iroy30: updated

python/cugraph/structure/graph.py
Iroy30: updated

python/cugraph/structure/graph.py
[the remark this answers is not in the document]
Iroy30: updated

python/cugraph/structure/graph.py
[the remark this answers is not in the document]
Iroy30: updated

PR #591, opened 2019-11-15
[REVIEW - ENH] Adding initial GPU metrics to benchmark utils

rlratzel:
Adding initial GPU metrics to benchmark utils


rlratzel:
rerun tests

python/utils/gpu_metric_poller.py
```diff
@@ -0,0 +1,168 @@
...
+# hack - get actual device ID somehow
+devObj = smi.nvmlDeviceGetHandleByIndex(0)
+memObj = smi.nvmlDeviceGetMemoryInfo(devObj)
+utilObj = smi.nvmlDeviceGetUtilizationRates(devObj)
+initialMemUsed = memObj.used
+initialGpuUtil = utilObj.gpu
+
+while not self.__stop:
+    time.sleep(0.01)
+
```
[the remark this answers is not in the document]
afender: Although .001 might be too fast for python, we may want do some experiment to find out the minimum we can use.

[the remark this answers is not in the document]
rlratzel: I agree.  I can look into parameterizing the polling interval to make it easy to experiment with as well as using bigger graphs and multiple runs averaged.

python/utils/gpu_metric_poller.py
```diff
@@ -0,0 +1,168 @@
...
+                break
+            controlStr = self.__waitForInput(childReadPipe)
+
+        smi.nvmlShutdown()
+        childReadPipe.close()
+        childWritePipe.close()
+
+    def run(self):
+        (parentReadPipeFileNo, childWritePipeFileNo) = os.pipe2(os.O_NONBLOCK)
+        (childReadPipeFileNo, parentWritePipeFileNo) = os.pipe2(os.O_NONBLOCK)
```
BradReesWork: parentReadPipeFileNo and childReadPipeFileNo will be the same number based on the syntax of this call.  so why not just have readPipeNo and not have the second call?

rlratzel: Those calls create a pair of one-way pipes, each with "ends" that need to be used by the parent and the child. I create two of them for the two-way communication (so the parent can issue a poll and the child can send the numbers back), and each end is a unique file descriptor (I checked in the debugger):
```
(Pdb) p (parentReadPipeFileNo, childWritePipeFileNo)
(4, 5)
(Pdb) p (childReadPipeFileNo, parentWritePipeFileNo)
(6, 7)
```
unless, were you thinking the two-way communication isn't necessary? ...which in thinking about it now, I can just have the child constantly stream back numbers with only a single pipe.

BradReesWork: I was thinking that each call to os.pipe2 would return the same IDs.

PR #599, opened 2019-11-26
[REVIEW] Pregel-based BFS

BradReesWork:
pregel BFS code and test

### DIFF
--- a/python/cugraph/pregel/traversal/bfs_bsp.py
+++ b/python/cugraph/pregel/traversal/bfs_bsp.py
@@ -21,6 +21,9 @@ def bfs_df_pregel(df, start, src_col='src', dst_col='dst', copy_data=True):
     This function executes an unwieghted Breadth-First-Search (BFS) traversal
     to find the distances and predecessors from a specified starting vertex
 
+    NOTE: Only reachable vertices are returned
+    NOTE: data is not sorted
+
     Parameters
     ----------
     df : cudf.dataframe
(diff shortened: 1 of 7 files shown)


python/cugraph/pregel/traversal/bfs_bsp.py
```diff
@@ -0,0 +1,141 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import cudf
+from collections import OrderedDict
+
+
+def bfs_df_pregel(df, start, src_col='src', dst_col='dst', copy_data=True):
```
afender: Naming : 
- `pregel` is tightly connected to google's system. Should we use `bsp` or something more generic?
- is `_df` necessary?

API: the `Graph` object takes care of renumbering, casting and input checking. Since we are not using it in this analytics, we may want to add similar features here in the long run. 
Do we expect all pregerl/bsp functions to operate on a DF edge list directly instead of a `(Di)Graph`?

python/cugraph/pregel/traversal/bfs_bsp.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    This function executes an unwieghted Breadth-First-Search (BFS) traversal
+    to find the distances and predecessors from a specified starting vertex
+
+    NOTE: Only reachable vertices are returned
+    NOTE: data is not sorted
```
afender: stored -> copied ?

python/cugraph/tests/test_bfs_pregel.py
```diff
@@ -0,0 +1,101 @@
...
+    assert(rmm.is_initialized())
+
+    M = utils.read_csv_for_nx(graph_file)
+    cu_M = utils.read_csv_file(graph_file)
+
+    base_vid, base_dist = base_call(M, 0)
+    cugraph_vid, cugraph_dist = cugraph_call(cu_M, np.int32(0))
+
+    # Calculating mismatch
+    num_dist = np.count_nonzero(base_dist != _int_max)
```
afender: Not checking/testing content?

python/cugraph/pregel/traversal/bfs_bsp.py
BradReesWork: I would rather use BSP.  Pregel did start as a Google thing but the name is not used within Spark.

I would like to add an additional function (wrapper) that takes the new Graph object.

BradReesWork: It should be _df to avoid collisions

python/cugraph/tests/test_bfs_pregel.py
BradReesWork: No.  The issues is that I would need to populate the result with every vertex and then sort the results.  I do have benchmark code that does check the vertex-distance pairs, but only when there are no unreachable.  

Redoing the full test suit is part of the next release when the API gets redone

python/cugraph/pregel/traversal/bfs_bsp.py
BradReesWork: not sure the question?

afender: About: `NOTE: data is not sorted` I wondered if you meant `NOTE: data is not duplicated/copied`

PR #601, opened 2019-11-26
[REVIEW] sssp test for type conversion, remove adj list from createGraph_nvgraph

### LINKED ISSUE #582: [BUG] c++ add adj list policies fixes
Some changes are still needed after #578 
 - Katz, two hops neighbors and `createGraph_nvgraph` still initiate format conversions at the C++ level

Iroy30:
closes #582

### DIFF
--- a/cpp/src/converters/nvgraph.cu
+++ b/cpp/src/converters/nvgraph.cu
@@ -43,9 +43,8 @@ void createGraph_nvgraph(nvgraphHandle_t nvg_handle,
   // setup nvgraph variables
   if (use_transposed) {
     // convert edgeList to transposedAdjList
-    if (gdf_G->transposedAdjList == nullptr) {
-      cugraph::add_transposed_adj_list(gdf_G);
-    }
+    CUGRAPH_EXPECTS(gdf_G->transposedAdjList != nullptr,
+              "Invalid API parameter");
     // using exiting transposedAdjList if it exisits and if adjList is missing
     TT = NVGRAPH_CSC_32;
     nvgraphCSCTopology32I_st topoData;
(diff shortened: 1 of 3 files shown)


cpp/src/converters/nvgraph.cu
```diff
@@ -43,9 +43,8 @@ void createGraph_nvgraph(nvgraphHandle_t nvg_handle,
   // setup nvgraph variables
   if (use_transposed) {
     // convert edgeList to transposedAdjList
-    if (gdf_G->transposedAdjList == nullptr) {
-      cugraph::add_transposed_adj_list(gdf_G);
-    }
+    CUGRAPH_EXPECTS(gdf_G->transposedAdjList != nullptr,
+              "Invalid API parameter");
```
BradReesWork: I'm not sure this error message makes sense.  It does not tell the user which parameter is causing the error.
Maybe something like:  "Invalid API parameter - transposed data is missing"

cpp/src/converters/nvgraph.cu
```diff
@@ -81,10 +80,8 @@ void createGraph_nvgraph(nvgraphHandle_t nvg_handle,
 
   }
   else {
-    // convert edgeList to adjList
-    if (gdf_G->adjList == nullptr) {
-      cugraph::add_adj_list(gdf_G);
-    }
+    CUGRAPH_EXPECTS(gdf_G->adjList != nullptr,
+              "Invalid API parameter");
```
BradReesWork: same here

cpp/src/converters/nvgraph.cu
Iroy30: This is following the errors we are throwing in other algorithms currently. 
Better error messages could be a completely separate PR with changes for all such ambiguities.

PR #610, opened 2019-12-02
[WIP][skip-ci] Python renaming

### LINKED ISSUE #369: [FEA] We need consistent python cython file naming scheme
Based on the comments from @dantegd

https://github.com/rapidsai/cugraph/pull/362#discussion_r301101812

We have multiple file types in python/cython

.pxd files wrapping C headers (e.g. structure/c_graph.pxd)
.pxd files to reuse cython code in other cython files (e.g. utilities/column_utils.pxd)
.pyx files calling C routines (e.g. components/connectivity_wrapper.pyx)
.pyx files providing common utility functions for cython implementations (e.g. utilities/column_utils.pyx)
.py files defining python user API (e.g. traversal/bfs.py)
.py files providing test suites (e.g. traversal/test_bfs.py)

@dantegd pointed out that c_ may be unnecessary for .pxd files wrapping C headers (and cudf does not use c_ though in cudf these files are located in the separate bindings directory) and naming schemes for .pxd files wrapping C headers and .pyx files calling C routines need improvements.

We should discuss proper naming scheme and update file names.

afender:
Close #369


afender:
Having  <xyz>.pxd/pyx/py in the same folder confuses python. The C++ function and the python function often share the same names as the files which confuse it further.

Our existing file name prefix and suffix scheme allows to :
(1) have consistent names across C++ and python functions.
(2) keep all files of a given analytics in a same directory

We should not trade 1. or 2. just to improve file names. I will try to find some middle ground in another PR.

PR #627, opened 2019-12-09
Master

vrushang25:


GPUtester:

afender:
Hello Vrushang, it seems to me that you are requesting a feature? If so, I'd recommend using the following link: https://github.com/rapidsai/cugraph/issues/new/choose. We usually use Pull Requests to review new code.

BradReesWork:
What is the purpose of this PR since it is merging into an old branch (0.6)?

PR #628, opened 2019-12-09
[REVIEW] multigraph_changes

### LINKED ISSUE #589: [FEA] Add a (Di)Graph constructor that accepts Multi(Di)Graph and one edge_attr
Users may want to call an existing single-dimension algorithm on a graph with multiple edge attributes (`Multi(Di)Graph`). That raises the question of which edge attribute should be used in this case. 

We should have a `(Di)Graph` constructor that accepts : 
- a `Multi(Di)Graph`
- valid `src` and `dst` columns within this `Multi(Di)Graph`. 
- one optional user-specified `edge_attribute` within this `Multi(Di)Graph`. 

We should avoid copying of the GPU DataFrame though. The `(Di)Graph` would just point to the relevant `Multi(Di)Graph` data.

The user can then pass this `(Di)Graph` to any existing analytics.

Iroy30:
This PR enables user to create (Di)Graph from Multi(Di)Graph with an option of providing any one column of the multigraph as the edge_attr.
Closes #589

### DIFF
--- a/python/cugraph/structure/graph.py
+++ b/python/cugraph/structure/graph.py
@@ -125,7 +125,7 @@ class Graph:
         source : str
             source argument is source column name
         destination : str
-            target argument is destination column name.
+            destination argument is destination column name.
         edge_attr : str
             edge_attr argument is the weights column name.
         renumber : bool
@@ -176,7 +176,7 @@ class Graph:
  Use from_cudf_edgelist instead')
         input_df = cudf.DataFrame()
         input_df['source'] = source
-        input_df['target'] = destination
+        input_df['destination'] = destination
         if value is not None:
             input_df['weights'] = value
             self.from_cudf_edgelist(input_df, edge_attr='weights')
--- a/python/cugraph/structure/graph.py
+++ b/python/cugraph/structure/graph.py
@@ -80,8 +80,10 @@ class Graph:
             if edge_attr is None:
                 raise Exception('edge_attr not provided')
             else:
-                self.from_cudf_edgelist(m_graph.edgelist.edgelist_df, source='src',
-                                        destination='dst', edge_attr=edge_attr)
+                self.from_cudf_edgelist(m_graph.edgelist.edgelist_df,
+                                        source='src',
+                                        destination='dst',
+                                        edge_attr=edge_attr)
         # self.number_of_vertices = None
 
     def clear(self):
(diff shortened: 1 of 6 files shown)


afender:
Please add relevant flags in the PR title

Iroy30:
rerun tests

python/cugraph/structure/graph.py
```diff
@@ -72,6 +76,14 @@ def __init__(self, symmetrized=False, bipartite=False, multi=False,
         self.edgelist = None
         self.adjlist = None
         self.transposedadjlist = None
+        if m_graph is not None:
+            if edge_attr is None:
+                raise Exception('edge_attr not provided')
```
afender: It should be ok to have graphs without edge attributes / weights. Several algos don't use weights at all.

python/cugraph/structure/graph.py
```diff
@@ -113,7 +125,7 @@ def from_cudf_edgelist(self, input_df, source='source',
         source : str
             source argument is source column name
         destination : str
-            destination argument is destination column name.
+            target argument is destination column name.
```
afender: We should use `destination` consistently

python/cugraph/structure/graph.py
```diff
@@ -158,7 +176,7 @@ def add_edge_list(self, source, destination, value=None):
  Use from_cudf_edgelist instead')
         input_df = cudf.DataFrame()
         input_df['source'] = source
-        input_df['destination'] = destination
+        input_df['target'] = destination
```
afender: destination

python/cugraph/structure/graph.py
Iroy30: fixed with better exception throwing and removing edge_attr constraint for graph constructor with multigraph

Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

python/cugraph/structure/graph.py
Iroy30: fixed

python/cugraph/structure/graph.py
Iroy30: fixed

python/cugraph/structure/graph.py
Iroy30: Added test to test the feature

PR #629, opened 2019-12-09
[REVIEW] get rid of gdf_column in pagerank

### LINKED ISSUE #538: [FEA] Get rid of gdf_column
cuGraph uses `gdf_column` as a device array container in the C++ API. As CuDF is depreciating `gdf_column`, we need a new device array container. There are hundreds of occurrences in cuGraph.

- Replace `gdf_column` in the API
- Update all tests
- Update bindings
- Update doc

- [ ] katz_centrality.cu
- [ ] nvgraph.gdf.cu
- [ ] ECG.cu
- [ ] connectivity.cu
- [ ] core_number.cu
- [ ] db_object.cu
- [ ] jaccard.cu
- [ ] overlap.cu
- [ ] two_hop_neighbors.cu

afender:
The first step for #538 

This PR shows how to replace gdf_column in C++, bindings and tests


python/cugraph/link_analysis/pagerank_wrapper.pyx
```diff
@@ -65,22 +66,27 @@ def pagerank(input_graph, alpha=0.85, personalization=None, max_iter=100, tol=1.
...
     
-    if personalization is None:
-        c_pagerank.pagerank(g, &c_pagerank_col, <gdf_column*> NULL, <gdf_column*> NULL,
-                <float> alpha, <float> tol, <int> max_iter, has_guess)
-    else:
-        c_pers_vtx = get_gdf_column_view(personalization['vertex'])
-        c_pers_val = get_gdf_column_view(personalization['values'])
-        c_pagerank.pagerank(g, &c_pagerank_col, &c_pers_vtx, &c_pers_val,
-                <float> alpha, <float> tol, <int> max_iter, has_guess)
+    if (df['pagerank'].dtype is np.float32):
```
Iroy30: Right now it is going to the else loop since if (df['pagerank'].dtype is np.float32): is False and hence giving incorrect results with double. It should be changed to "if (df['pagerank'].dtype == np.float32):"

afender: Done

cpp/src/link_analysis/pagerank.cu
```diff
@@ -183,44 +183,30 @@ template int pagerankSolver<int, double> (  int n, int e, int *cscPtr, int *cscI
...
-  WT  *prsVal = nullptr;
   int prsLen = 0;
-  CUGRAPH_EXPECTS((personalization_subset == nullptr) == (personalization_values == nullptr), "Invalid API parameter");
-  if (personalization_subset != nullptr) {
-    has_personalization = true;
-    prsVtx = reinterpret_cast<int*>(personalization_subset->data);
-    prsVal = reinterpret_cast<WT* >(personalization_values->data);
-    prsLen = reinterpret_cast<int >(personalization_subset->size);
-    CUGRAPH_EXPECTS(pagerank->dtype == personalization_values->dtype, "Invalid API parameter");
-    CUGRAPH_EXPECTS(personalization_subset->dtype == GDF_INT32, "Unsupported data type");
```
Iroy30: Do we need to forgo these checks? Or should they still be kept as sanity checks verifying dtypes, in case someone uses the CPP api

cpp/include/algorithms.h
```diff
@@ -25,35 +25,36 @@ namespace cugraph {
...
+ * @Param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a transposed adjacency list (CSR). Edge weights are not used for this algorithm.
  * @Param[in] alpha                  The damping factor alpha represents the probability to follow an outgoing edge, standard value is 0.85.
- Thus, 1.0-alpha is the probability to “teleport” to a random vertex. Alpha should be greater than 0.0 and strictly lower than 1.0.
- * @Param[in] has_guess              This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user doesn't have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
- *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+                                     Thus, 1.0-alpha is the probability to “teleport” to a random vertex. Alpha should be greater than 0.0 and strictly lower than 1.0.
  *                                   The initial guess must not be the vector of 0s. Any value other than 1 or 0 is treated as an invalid value.
```
Iroy30: I think the doc changed. This should be under has_guess definition

cpp/src/link_analysis/pagerank.cu
afender: Yes. I had mostly C++ users in mind for those. It is also helpful for cugrah's python developer and debugging as it will tell directly if they passed the wrong data type.

cpp/include/algorithms.h
afender: Fixed. Good catch.

PR #634, opened 2019-12-12
[REVIEW] renumber vertex ids

### LINKED ISSUE #592: [BUG] Accept user's notation when passing vertices ID to analytics
Now that we have auto renumbering (when loading a graph) and unrenumbering (when returning an analytics result). We should also update analytics that accepts vertex IDs so that they match the user's scope and not our own internal representation.

- Update Jaccard to accept unremembered IDs for arbitrary pairs of vertices specified by the user.
- Update Personalized PageRank to accept unremembered IDs for the personalization parameter
- Pagerank guess should be a key-value DF of size V like in nstart for Katz.
- Double-check the API to make sure we are not assuming renumbered input anywhere.
- Katz
- Subgraph extraction
- sssp
- BFS
- K-core
- degrees functions

Iroy30:
This PR takes user notations into consideration when vertices are passed as arguments for analytics. If the graph is renumbered, passed vertices are also auto-renumbered according to the renumbering map used for the graph.
Changed first and second cudf.Series (to provide vertex pairs for computation of coefficient) in jaccard, wjaccard, overlap and woverlap to  a cudf.DataFrame input since it is more intuitive.
closes #592

### DIFF
--- a/python/cugraph/link_prediction/jaccard.py
+++ b/python/cugraph/link_prediction/jaccard.py
@@ -16,7 +16,7 @@ from cugraph.structure.graph import null_check
 import cudf
 
 
-def jaccard(input_graph, first=None, second=None):
+def jaccard(input_graph, vertex_pair=None):
     """
     Compute the Jaccard similarity between each pair of vertices connected by
     an edge, or between arbitrary pairs of vertices specified by the user.
@@ -36,12 +36,9 @@ def jaccard(input_graph, first=None, second=None):
         graph should be undirected where an undirected edge is represented by a
         directed edge in both direction. The adjacency list will be computed if
         not already present.
-    first : cudf.Series
-        Specifies the first vertices of each pair of vertices to compute for,
+    vertex_pair : cudf.DataFrame
+        Specifies the pair of vertices to compute for,
         must be specified along with second.
-    second : cudf.Series
-        Specifies the second vertices of each pair of vertices to compute for,
-        must be specified along with first.
 
     Returns
     -------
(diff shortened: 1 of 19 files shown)


afender:
These algorithms also accept V IDs and should be upgraded:
- Katz
- Subgraph extraction
- sssp
- BFS
- K-core
- degrees functions

Iroy30:
rerun tests

python/cugraph/link_prediction/jaccard.py
```diff
@@ -16,7 +16,7 @@
 import cudf
 
 
-def jaccard(input_graph, first=None, second=None):
+def jaccard(input_graph, vertex_pair=None):
```
afender: Please add a note for this API change in the description of the PR

python/cugraph/link_prediction/jaccard.py
```diff
@@ -36,12 +36,8 @@ def jaccard(input_graph, first=None, second=None):
...
         directed edge in both direction. The adjacency list will be computed if
         not already present.
-    first : cudf.Series
-        Specifies the first vertices of each pair of vertices to compute for,
-        must be specified along with second.
-    second : cudf.Series
-        Specifies the second vertices of each pair of vertices to compute for,
-        must be specified along with first.
+    vertex_pair : cudf.DataFrame
+        Specifies the pair of vertices to compute for
```
afender: Doc needs improvements (see slides 13 to 30 from RAPIDS review and doc talk)

python/cugraph/link_prediction/overlap.py
```diff
@@ -34,12 +34,8 @@ def overlap(input_graph, first=None, second=None):
...
         as an edge list (edge weights are not used for this algorithm). The
         adjacency list will be computed if not already present.
-    first : cudf.Series
-        Specifies the first vertices of each pair of vertices to compute for,
-        must be specified along with second.
-    second : cudf.Series
-        Specifies the second vertices of each pair of vertices to compute for,
-        must be specified along with first.
+    vertex_pair : cudf.DataFrame
+        Specifies the pair of vertices to compute for
```
afender: Doc should be improved

python/cugraph/link_prediction/jaccard.py
Iroy30: added

afender:
Let's update the notebook for the API change

Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
Corresponding vertex_pair api update in rapidsai/notebooks PR 292

afender:
Notebook's PR link for github : https://github.com/rapidsai/notebooks/pull/292

PR #636, opened 2019-12-13
[REVIEW] multi-colum renumbering

### LINKED ISSUE #635: [FEA] Multi-Column Renumbering
Renumbering should allow for multi-colum source and destination.  The column data types can be anything and any combination

BradReesWork:
This feature allows users to combine multiple columns together to form the vertex IDs
This is useful for things like geo data where the vertices are Lat + log pairs.

Closes #635


python/cugraph/structure/renumber.py
```diff
@@ -63,3 +66,105 @@ def renumber(source_col, dest_col):
...
+    df : cudf.DataFrame
+        The dataframe containing the source and destination columans
+    source_cols_names : List
+        This is a list of source column names
+    dest_cols_names : List
+        This is a list of destination column names
+
+    Returns
+    ---------
+    src_ids : cudf.Series
```
afender: A Dataframe instead of 2 Series would be ideal since this is what we accept in `from_cudf_edgelist`

BradReesWork: I need to do some performance testing before we drop the old version.  There is a cudf merge issues where it does not guarantee order, so there are two extra sorts that might need to be improved

python/cugraph/structure/renumber.py
```diff
@@ -63,3 +66,105 @@ def renumber(source_col, dest_col):
...
+    del _s
+    del _d
+
+    _src_ids = _tmp_df_src.merge(
+        _tmp_df, on=_vals, how='left').drop(_vals).sort_values(by='index')
+
+    _dst_ids = _tmp_df_dst.merge(
+        _tmp_df, on=_vals, how='left').drop(_vals).sort_values(by='index')
+
+    return _src_ids['id'], _dst_ids['id'], _tmp_df
```
Iroy30: the returned numbering_df has 'id' column which can maybe be dropped after the merge, as it is same as index values

BradReesWork: true, but then it would be tough to be a bulk `merge` to translate results back to columns.

Iroy30: I'm not sure what you mean by merge to translate results back to columns. Are you referring to the merge to get _src_ids_ and _dst_ids_?
Dropping ids by returning _tmp_df.drop('ids') at 170 could be more user friendly. Or if it has some implications later, we could leave it as is.

BradReesWork: without the `id` column you just have the list of unique column values.  To un-renumber the data, you need to have a field that can be used join the graph results data with the numbering dataframe.  

unnumbered = results.merge(number_df, left_on='`vertex`, right_on='id', how='left')

BradReesWork: the iterative way we do it for testing is okay for testing, but not for performance.  and if the number datafrane order is changed, then it won't work.
for i in len(results):
    col = numbering[results['vertex'][i]

Iroy30: agreed, for unrenumbering later, the merge on id option is better suited for multi-column

PR #651, opened 2019-12-23
[REVIEW] fix cudf_error in Katz wrapper

Iroy30:
-> Fix cudf error for scattering function call in katz wrapper
-> Fix personalization bug due to premature garbage collection
-> Fix bug in transposed_adj_list_view
-> Edit pagerank to renumber nstart

### DIFF
--- a/python/cugraph/centrality/katz_centrality_wrapper.pyx
+++ b/python/cugraph/centrality/katz_centrality_wrapper.pyx
@@ -61,25 +61,26 @@ def katz_centrality(input_graph, alpha=0.1, max_iter=100, tol=1.0e-5, nstart=Non
     df['vertex'] = cudf.Series(np.zeros(num_verts, dtype=np.int32))
     cdef gdf_column c_identifier_col = get_gdf_column_view(df['vertex'])
     df['katz_centrality'] = cudf.Series(np.zeros(num_verts, dtype=np.float64))
-    cdef gdf_column c_katz_centrality_col = get_gdf_column_view(df['katz_centrality'])
 
     cdef bool has_guess = <bool> 0
     if nstart is not None:
+        if len(nstart) != num_verts:
+            raise ValueError('nstart must have initial guess for all vertices')
         if input_graph.renumbered is True:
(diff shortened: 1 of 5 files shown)


Iroy30:
rerun tests

python/cugraph/centrality/katz_centrality_wrapper.pyx
```diff
@@ -65,9 +65,17 @@ def katz_centrality(input_graph, alpha=0.1, max_iter=100, tol=1.0e-5, nstart=Non
...
-                                            [df['katz_centrality']._column])
+        if input_graph.renumbered is True:
+            renumber_series = cudf.Series(input_graph.edgelist.renumber_map.index,
+                                          index=input_graph.edgelist.renumber_map)
+            nstart_vertex_renumbered = cudf.Series(renumber_series.loc[nstart['vertex']], dtype=np.int32)
+            cudf.bindings.copying.apply_scatter([nstart['values']._column],
+                                                nstart_vertex_renumbered._column,
+                                                [df['katz_centrality']._column])
+        else:
+            cudf.bindings.copying.apply_scatter([nstart['values']._column],
```
afender: We can see this as a sparse eigenvalue problem of size V. So, if nstart has less than V entries it should throw an error. Also, I don’t think 0 would be a valid guess. To me, a valid guess makes sense as a whole set of size V.

The alternative is to document the default value and type of normalization (as it probably won’t work if not normalized) when the user provided a subset of the guess. There should also be a test in this case. But this looks like a stretch / corner case to me. It is not a priority at this point.

Accepting vertices / value pairs in an arbitrary order in nstart is important though.

Iroy30: I believe we should go with the first option and work on the second in further iterations(if required)

afender: agreed

afender:
Looks like we are ready to enable renumbering by default in `from_cudf_edgelist`

Iroy30:
> Looks like we are ready to enable renumbering by default in `from_cudf_edgelist`

This could probably be a separate PR along with fixing relevant tests

PR #666, opened 2020-01-14
[REVIEW] multi column renumbering changes for python graph structure

### LINKED ISSUE #604: [FEA] Multi (Di)Graph
This describes how this is stored and exposed to the user

The data frame will have E rows representing the topology and as many weight columns as edge properties/dimensions/or source/destination candidates. For a topology to be valid, each source-destination pair must appear once (duplicate should be properly handled with cuDF functions before reaching cuGraph).

Here are some examples :
```
src, dst,  distance,  avg_speed,  toll_cost
0    1     18.2        65                0    -> notice that there might be some sparsity here
0    2     20.2        75                5.87
``` 

```
src_ip,   src_port,  dst_ip,  dst_port ,  time
0           0              1          1                13:30
0           0              2         2                5:10
```

Related to #589

Iroy30:
With Multi-column renumbering #636 , python bindings need to be updated for a renumber_map with multiple columns as a dataframe. The PR addresses the following:
--> Updates in python API (Graph and MultiGraph) to accept multiple source and destination columns and call auto renumbering
--> Updates all algorithms returning vertices to un-renumber to multiple columns as per users notation.
Closes #604 
Further followup is needed as described in #678

### DIFF
--- a/python/cugraph/centrality/katz_centrality_wrapper.pyx
+++ b/python/cugraph/centrality/katz_centrality_wrapper.pyx
@@ -20,6 +20,7 @@ cimport cugraph.centrality.katz_centrality as c_katz
 from cugraph.structure.graph cimport *
 from cugraph.structure import graph_wrapper
 from cugraph.utilities.column_utils cimport *
+from cugraph.utilities.unrenumber import unrenumber
 from libcpp cimport bool
 from libc.stdint cimport uintptr_t
 from libc.stdlib cimport calloc, malloc, free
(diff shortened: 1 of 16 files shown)


Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

python/cugraph/centrality/katz_centrality_wrapper.pyx
```diff
@@ -85,6 +85,11 @@ def katz_centrality(input_graph, alpha=0.1, max_iter=100, tol=1.0e-5, nstart=Non
     c_katz.katz_centrality(g, &c_katz_centrality_col, alpha, max_iter, tol, has_guess, normalized)
 
     if input_graph.renumbered:
-        df['vertex']=input_graph.edgelist.renumber_map[df['vertex']]
+        if isinstance(input_graph.edgelist.renumber_map, cudf.DataFrame):
+            unrenumered_df = df.merge(input_graph.edgelist.renumber_map, left_on='vertex', right_on='id', how='left').drop(['id', 'vertex'])
```
BradReesWork: line is to long

Iroy30: Added an utility function to be called in for most of the bindings for unrenumbering in new commit. However a few bindings need different code within repeating codeblock. That could be simplified in consequent PRs addressing #678 as code could possibly change.

PR #667, opened 2020-01-14
[REVIEW] Drop `cython` from run requirements

jakirkham:
As `cython` is only needed to build the Python bindings and does not contain any needed runtime libraries, drop it from the run requirements.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 - PR #641 Add codeowners
 - PR #646 Skipping all tests in test_bfs_bsp.py since SG BFS is not formally supported
 - PR #652 Remove gdf_column in BFS
+- PR #667 Drop `cython` from run requirements in conda recipe
+
 ## Bug Fixes
 - PR #634 renumber vertex ids passed in analytics
 - PR #649 Change variable names in wjaccard and woverlap to avoid exception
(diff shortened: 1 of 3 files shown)


jakirkham:
Thanks Mike! 😄

mike-wendt:
@jakirkham looks like we need Cython https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cugraph/job/prb/job/cugraph-cpu-build/2325/consoleFull#L1,447

jakirkham:
That's odd. It is pointing to that coming from the `setup.py`. Why would that be called during runtime? 🤔

jakirkham:
Ah! It's because the [Python package doesn't list the requirement]( https://github.com/rapidsai/cugraph/blob/fa98f611c4762800963324c5db87522d319cb80a/conda/recipes/cugraph/meta.yaml#L25-L29 ) so it is expect the C++ package to supply the dependency. Will push a fix.

jakirkham:
Have pushed a fix. Let's see what CI thinks 🙂

jakirkham:
It looks like it builds. However there are unit test failures ( likely due to the cuDF refactor https://github.com/rapidsai/cudf/pull/3254 ). This can safely wait until things are updated.

BradReesWork:
rerun tests

BradReesWork:
rerun tests

jakirkham:
Thanks! 😄

PR #680, opened 2020-01-22
[REVIEW] disable renumbering for already renumbered inputs in problematic tests

afender:
Fixing tests that were renumbering already renumbered input and causing failures.

This is a temporary patch to fix the build. The root cause should be investigated further.


afender:
rerun tests

afender:
Closing as duplicate of #681

PR #681, opened 2020-01-22
[REVIEW] Add index reset

Iroy30:


ayushdg:
Not really a blocker but good to have.
`reset_index().drop('index')` in the future could be replaced by `.reset_index(drop=True)`. This will drop the original index column and prevent an additional deep copy under the hood incurred by the additional `drop('index')` function call.

Issue #476, opened 2019-08-29
[BUG] avoid using cudaGetDeviceProperties (use cudaDeviceGetAttribute instead)


seunghwak:
**Describe the bug**
Base on

https://devblogs.nvidia.com/cuda-pro-tip-the-fast-way-to-query-device-properties/

cudaGetDeviceProperties is much more expensive than cudaDeviceGetAttribute as the first queries the entire set of device properties whether you need it or not and getting some attributes are expensive.

```
(cuda10) seunghwak@seunghwak-linux:~/RAPIDS/cugraph$ grep -rHn cudaGetDeviceProperties .
./cpp/src/tests/components/scc_test.cu:155:    cudaGetDeviceProperties(&prop, device);
./cpp/src/nvgraph/triangles_counting.cpp:38:    cudaGetDeviceProperties(&m_dev_props, m_dev_id);
./cpp/src/components/connectivity.cu:160:      cudaGetDeviceProperties(&prop, device);
./thirdparty/cub/test/test_util.h:277:            error = CubDebug(cudaGetDeviceProperties(&deviceProp, dev));
./thirdparty/cub/experimental/histogram/histogram_gmem_atomics.h:151:    cudaGetDeviceProperties(&props, 0);
./thirdparty/cub/experimental/histogram/histogram_smem_atomics.h:161:    cudaGetDeviceProperties(&props, 0);
```

We're using cudaGetDeviceProperties in few places just to get the total global memory size or the shared memory size per SM. We'd better establish a habit of using cudaDeviceGetAttribute (unless we indeed need a big percentage of the entire set of device attributes) as advised by the pro tip.

seunghwak:
Partially addressed in https://github.com/rapidsai/cugraph/pull/694

cudaGetDeviceProperties calls in SCC are unnecessary and should be fixed in https://github.com/rapidsai/cugraph/issues/693

Other cudaGetDeviceProperties calls are from third-party libraries (cub and cuhornet) and should be addressed there.

BradReesWork:
Use of "cudaGetDeviceProperties" is acceptable in test files but should not be used in production code.

PR #694, opened 2020-02-03
[REVIEW] Call cudaDeviceGetAttribute instead of cudaGetDeviceProperties in triangle counting

seunghwak:
Partially address #476 

Based on
https://devblogs.nvidia.com/cuda-pro-tip-the-fast-way-to-query-device-properties/

cudaGetDeviceProperties can be much slower than cudaDeviceGetAttribute.

This PR replaces cudaGetDeviceProperties with cudaDeviceGetAttribute in triangle counting that results in over 100x speed up in device property query time in triangle counting (the overall speed up in triangle counting should be much smaller as device query time takes only a small percentage of triangle counting execution time).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## Improvements
 
 - PR #688 Cleanup datasets after testing on gpuCI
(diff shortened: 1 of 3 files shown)


seunghwak:
rerun test

seunghwak:
rerun tests

seunghwak:
rerun tests

PR #722, opened 2020-02-14
[REVIEW] Refactor graph to remove gdf_column

ChuckHastings:
Minimize dependence on cudf column.  Also updated to use new cudf++ column in the interface.

Split the Graph object into 3 templated objects:  GraphCOO, GraphCSR, GraphCSC.  This simplifies the C++ interaction with the graphs - many of the error checks the algorithms used to do become compile time checks instead of runtime checks.

All 3 templated objects allow for templating based on the vertex id type and the edge weight type.

Updated the page rank analytic (C++ only so far) to use the new graph structure.


cpp/include/algorithms.hpp
```diff
@@ -0,0 +1,61 @@
...
+ *                                   If this value is lower or equal to 0 cuGRAPH will use the default value, which is 500.
+  * @Param[in] has_guess             (optional) This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user does not have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
+ *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+ * @Param[out] *pagerank             The PageRank : pagerank[i] is the PageRank of vertex i. Memory remains provided and owned by the caller.
+ *
+ * @throws                           cugraph::logic_error with a custom message when an error occurs.
+ */
+/* ----------------------------------------------------------------------------*/
+template <typename VT, typename WT>
+void pagerank(experimental::GraphCSC<VT,WT> const &graph,
```
afender: Should we expose the graph structure in the algo API? We could just accept a "Graph" so that we can change the algorithm implementation without touching the API? Otherwise, we will have an entry point for each implemented format.

cpp/CMakeLists.txt
```diff
@@ -261,9 +261,9 @@ link_directories(
     "${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
 
 add_library(cugraph SHARED
-	src/db/db_object.cu
-	src/db/db_parser_integration_test.cu
-	src/db/db_operators.cu
+    src/db/db_object.cu
```
afender: why?

cpp/include/graph.hpp
```diff
@@ -0,0 +1,218 @@
...
+ * @tparam WT   Type of weight (defaults to float)
+ */
+template <typename VT = int, typename WT = float>
+class GraphCOO {
+public:
+  VT const *src_indices;   ///< rowInd
+  VT const *dst_indices;   ///< colInd
+  WT const *edge_data;     ///< edge weight
+
+  GraphProperties          prop;
```
afender: graph prop, V and E are shared for all graph formats, they could be in a parent `Graph` class.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -1,286 +1,278 @@
...
-template void pagerank<int, float>(Graph *graph,float* pagerank,
-  size_t personalization_subset_size, int* personalization_subset,float* personalization_values,
-  float alpha, float tolerance, int max_iter, bool has_guess);
-template void pagerank<int, double>(Graph *graph,double* pagerank,
-  size_t personalization_subset_size, int* personalization_subset,double* personalization_values,
-  float alpha, float tolerance, int max_iter, bool has_guess);
-
-} //namespace cugraph 
\ No newline at end of file
+/*
```
afender: The git diff messed up. Have you changed encoding?
It is a bit hard to review the changes, would be great to fix.

cpp/CMakeLists.txt
ChuckHastings: This was just a formatting change... distracts me when things are indented inconsistently.

cpp/include/algorithms.hpp
ChuckHastings: Exposing the graph structure in the algorithm API allows us to check the validity at compile time.  If you tried using a GraphCOO to call pagerank then you wouldn't be able to compile.

When we have an analytic that can run off of different types of graphs, we end up with a bunch of if statements that  switch us into the correct configuration.  These are, IMO, simpler if you just have two entry points.  They can still call common underlying functions to avoid duplicating code.

cpp/include/graph.hpp
ChuckHastings: I thought about that, could definitely add a parent graph class and reduce some duplication here.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,218 @@
...
+  {}
+};
+
+/**
+ * @Synopsis    A graph stored in CSC (Compressed Sparse Column) format.
+ *
+ * @tparam VT   Type of vertex (defaults to int)
+ * @tparam WT   Type of weight (defaults to float)
+ */
+template <typename VT = int, typename WT = float>
```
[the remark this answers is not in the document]
ChuckHastings: I like the idea of having a GraphCSC primarily for doing compile time type checking.  Let me see if I can do something to reduce the code duplication here.

cpp/src/link_analysis/pagerank.cu
ChuckHastings: I did change the encoding.  This file was the only file I've encountered that is encoded as a DOS file.  I could switch it back if you like.

cpp/CMakeLists.txt
afender: Oh ok thanks for catching that

cpp/src/link_analysis/pagerank.cu
afender: Ugh.. ok. Let's keep it like this. It should not have been DOS in the first place.

cpp/include/graph.hpp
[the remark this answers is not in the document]
ChuckHastings: Just pushed an update with a new pair of base classes that clean up the code replication issue, I think.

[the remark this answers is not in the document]
afender: We could have a `transposed` flag in the CSR structure to indicate that it contains the transposed view. The typedef trick could redirect to the CSR constructor with the transposed flag on.

cpp/src/tests/pagerank/pagerank_test.cu
```diff
@@ -115,34 +111,31 @@ class Tests_Pagerank : public ::testing::TestWithParam<Pagerank_Usecase> {
...
+     CSR_Result_Weighted<int,T>   result;
+     ConvertCOOtoCSR_weighted(&cooColInd[0], &cooRowInd[0], &cooVal[0], nnz, result);
 
-    cudaDeviceSynchronize();
-    if (PERF) {
-      hr_clock.start();
-      for (int i = 0; i < PERF_MULTIPLIER; ++i) {
-       cugraph::pagerank<int,T>(G.get(), d_pagerank);
-       cudaDeviceSynchronize();
+     cugraph::experimental::GraphCSC<int,T> G(result.rowOffsets, result.colIndices, result.edgeWeights, m, nnz);
```
ChuckHastings: Notice here... we don't have to create columns.  The graph_test.cu test demonstrates that we can operate correctly from cudf::column objects.  But if we have other graph data we can operate on the arrays directly without translating into columns and back out.

cpp/src/tests/test_utils.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include <cudf/utilities/traits.hpp>
+#include <rmm/device_buffer.hpp>
+
+#include <cudf/column/column.hpp>
+
+#include <thrust/distance.h>
+
+namespace detail {
+
+template <typename Element, typename InputIterator>
```
ChuckHastings: I'm thinking this might be better to to put these utilities directly in the graph_test.cu file and having that file be the ONLY place where we test with cudf::column.

cpp/include/algorithms.hpp
afender: While I acknowledge the benefits of compile-time checks, multiplying the number of specific API entries has side costs in terms of documentation, testing, and user-friendliness. In general, I tend to prefer (a reasonable amount of) abstractions as a way of future-proofing the API and controlling maintenance. 

@seunghwak @BradReesWork any preference here?

ChuckHastings: I guess the question is how often do we expect to have algorithms that have more than one implementation?  Do we expect to have COO and CSR implementations of many of these algorithms?

seunghwak: I think we should better abstract this out. Instead of algorithms directly accessing graph internals,

We should have something like

graph.vertices_begin(), graph.vertices_end(), <= for the entire set of vertices
graph.edges_begin(), graph.edges_end(), <= for the entire set of edges
graph.edges_begin(v), graph.edges_end(v), <= for the edges of vertex v
graph.(incoming or)outgoing_edges_begin(v), graph.(incoming_or)outgoing_edges_end(v) <= for the (incoming or)outgoing edges of vertex v

This way, we can separate graph data structure internals from algorithm developments and the same algorithm will work whether we have static CSR or dynamically updatable CSR or other data types.

seunghwak: We should not write new algorithms for every new graph representation.

If I bring a quote from "A tour of C++ by Bjarne Stroustrup (the inventor of C++)"

"An algorithm operates on its data through iterators and knows nothing about the container in which the elements are stored. Conversely, a container knows nothing about the algorithms operating on its elements; all it does is to supply iterators upon request (e.g., begin() and end()). This model of separation between data storage and algorithm delivers very general and flexible software"

I think the recent libcudf porting achieved this goal, and we should follow this as well.

seunghwak: And vertex.begin() may be better than vertices_begin()... need more thinking about the exact API, but I think this is the direction we should move.

cpp/include/graph.hpp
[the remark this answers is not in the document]
afender: https://github.com/rapidsai/cugraph/pull/722/commits/09adaa80b619c811ced0ab08f977237aebca6c5d is a very good tradeoff

cpp/include/graph.hpp
afender: addressed in https://github.com/rapidsai/cugraph/pull/722/commits/09adaa80b619c811ced0ab08f977237aebca6c5d

cpp/include/algorithms.hpp
ChuckHastings: I would agree to the principal here... with the following caveat.  cugraph (at least as it exists today) is largely about integrating a collection of disparate graph libraries (nvgraph, hornet, gunrock, some homegrown analytics).  As long as cugraph integrates libraries of analytics developed by different developers, we need to have an interface that will be common among them.  At this point, I think the lowest common denominator is device pointers to CSR structures.

Unless we plan on either defining a graph API and requiring hornet and gunrock (and any future collaborators we work with) to comply to that API - and significantly rewrite all of the current graph analytics, I think we're limited to a data structure that is like this.

My thought we be to use GraphCOO, GraphCSR and GraphCSC as basic lower level data structures and then writing a higher level abstraction based upon them to provide iterators for new analytics.

seunghwak: Yes, you're right.

So if we view this PR as a temporary solution to maintain cugraph functional when libcudf dumps gdf_column, I'm fine with any legitimate solution. Better not invest too much on a temporary solution.

In a long run, having nvgraph, hornet, gunrock, and actual cugraph all under cugraph is not a scalable solution. It will inevitably become a maintainability nightmare as cuGraph gets bigger & bigger. But I guess this is a much bigger work (requiring the Great libcugraph port) something way beyond this PR.

ChuckHastings:
rerun tests

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+ *                                   Indices must be in the range [0, V-1].
+ * @Param[in]  number_of_vertices    The number of vertices in the graph
+ * @Param[in]  number_of_edges       The number of edges in the graph
+ *
+ * @Returns    GraphCOO object
+ *
+ * @throws     cugraph::logic_error when an error occurs.
+ */
+template <typename VT = int, typename WT = float>
+GraphCOO<VT,WT> from_edge_list(cudf::column_view const &source_indices,
```
ChuckHastings: Based on how the python bindings are working, I'm not sure that we need these conversions.

The python bindings seem to be going directly (in python) to the C++ pointers, so this code could be deleted and handled from python.

ChuckHastings: (the whole file)

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+namespace detail {
+
+void check_edge_list_columns(cudf::column_view const &src_indices,
+                             cudf::column_view const &dst_indices) {
+  CUGRAPH_EXPECTS( src_indices.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( dst_indices.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( src_indices.type() == dst_indices.type() , "Unsupported data type" );
+  CUGRAPH_EXPECTS( src_indices.type().id() == cudf::experimental::type_to_id<int32_t>() , "Unsupported data type" );
+  CUGRAPH_EXPECTS( src_indices.size() == dst_indices.size() , "Source and Destination must have same number of rows" );
+  CUGRAPH_EXPECTS( src_indices.size() > 0 , "No edges");
```
seunghwak: Is this really an error? Can't users pass a graph with only vertices or even an empty graph?

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+void check_adj_list_columns(cudf::column_view const &indices,
+                            cudf::column_view const &offsets) {
+  CUGRAPH_EXPECTS( offsets.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( indices.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( offsets.type() == indices.type() , "Unsupported data type" );
```
seunghwak: This error message is misleading. And should we really enforce this?

offsets's  type is basically determined by the number of edges.

Indices's type is determined by the number of vertices.

And in many cases, number of edges >> number of vertices, so we may use int32_t for indices but int64_t for offsets.

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+ *             function does not allocate memory.
+ *
+ * @Param[in]  source_indices        This column_view of size E (number of edges) contains the index of the source for each edge.
+ *                                   Indices must be in the range [0, V-1].
+ * @Param[in]  destination_indices   This column_view of size E (number of edges) contains the index of the destination for each edge.
+ *                                   Indices must be in the range [0, V-1].
+ * @Param[in]  number_of_vertices    The number of vertices in the graph
+ * @Param[in]  number_of_edges       The number of edges in the graph
+ *
+ * @Returns    GraphCOO object
```
seunghwak: Take this with a grain of salt, but IIRC, @returns require return type before explanation.

e.g.
@returns GraphCOO<VT,WT> GraphCOO object

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+                                         cudf::column_view const &edge_data,
+                                         size_t number_of_vertices,
+                                         size_t number_of_edges) {
+  detail::check_adj_list_columns(offsets, indices);
+  CUGRAPH_EXPECTS( size_t{edge_data.size()} == number_of_edges , "Edge data must contain number_of_edges elements");
+  return GraphCSC<VT,WT>(offsets.data<VT>(), indices.data<VT>(), edge_data.data<WT>(), number_of_vertices, number_of_edges);
+}
+
+
+#if 0
```
seunghwak: Delete commented out code.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+namespace cugraph {
+namespace experimental {
+
+typedef enum prop_type{PROP_UNDEF, PROP_FALSE, PROP_TRUE} PropType;
```
seunghwak: Shouldn't we better use enum class?

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+  bool bipartite;
+  bool tree;
+  PropType has_negative_edges;
+  GraphProperties() : directed(false), weighted(false), multigraph(false), bipartite(false), tree(false), has_negative_edges(PROP_UNDEF){}
+};
+
+/**
+ * @Synopsis    Base class graphs, all but vertices and edges
+ *
+ * @tparam VT   Type of vertex (defaults to int)
```
seunghwak: No VT in GraphBase definition

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * @tparam WT   Type of weight (defaults to float)
+ */
+template <typename WT>
+class GraphBase {
+public:
+  WT const *edge_data;     ///< edge weight
+
+  GraphProperties          prop;
+
+  size_t                   number_of_vertices;
```
seunghwak: Better go with VT (or vertex_id or vertex_t or ...).

If we set VT to int32_t or int64_t, this can require comparing signed with unsigned, and generates a warning without an explicit type casting.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+  size_t                   number_of_edges;
```
seunghwak: And better go with ET (or edge_id or edge_t or ...).

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+
+typedef enum prop_type{PROP_UNDEF, PROP_FALSE, PROP_TRUE} PropType;
+
+struct GraphProperties {
+  bool directed;
+  bool weighted;
+  bool multigraph;
+  bool bipartite;
+  bool tree;
+  PropType has_negative_edges;
```
seunghwak: bool directed{false};
bool weighted{false};
bool multigraph{false};
bool tree{false};
PropType has_negative_edges{prop_type::PROP_UNDEF};

will be simpler & clearer than adding a constructor for simply value initialization.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * @Synopsis    A graph stored in COO (COOrdinate) format.
+ *
+ * @tparam VT   Type of vertex (defaults to int)
+ * @tparam WT   Type of weight (defaults to float)
+ */
+template <typename VT = int, typename WT = float>
+class GraphCOO: public GraphBase<WT> {
+public:
+  VT const *src_indices;   ///< rowInd
+  VT const *dst_indices;   ///< colInd
```
seunghwak: And it's safe to always initialize class member variables.

VT const *src_indices{nullptr};
VT const *dst_indices{nullptr};

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+    GraphBase<WT>(edge_data_, number_of_vertices_, number_of_edges_),
+    src_indices(src_indices_),
+    dst_indices(dst_indices_)
+  {}
+};
+
+/**
+ * @Synopsis    Base class for graph stored in CSR (Compressed Sparse Row) format.
+ *
+ * @tparam VT   Type of vertex (defaults to int)
```
seunghwak: I think we need ET as well (useful if # vertices < std::numeric_limits<int32_t>::max() but # edges > std::numeric_limits<int32_t>::max()). And specifying the default value, int32_t might be more specific than int (though these two are same in most systems).

seunghwak: And shouldn't we better go for a safer option (int64_t & double) for default parameters as long as we're providing an option to override default parameters to save memory?

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+ * @tparam WT   Type of weight (defaults to float)
+ */
+template <typename VT = int, typename WT = float>
+class GraphCSRBase: public GraphBase<WT> {
+public:
+  VT const *offsets;       ///< CSR offsets
+  VT const *indices;       ///< CSR indices
```
seunghwak: VT const *offsets{nullptr};
VT const* indices{nullptr};

cpp/include/graph.hpp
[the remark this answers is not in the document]
seunghwak: Having a transposed flag sounds like a good idea, or we can add a template parameter
```
template <bool transposed = false>
```
If this parameter will be accessed in performance critical parts of the code (I don't think so, but just in case).

cpp/include/graph.hpp
```diff
@@ -0,0 +1,213 @@
...
+  GraphCSRBase(VT const *offsets_, VT const *indices_, WT const *edge_data_,
+               size_t number_of_vertices_, size_t number_of_edges_):
+    GraphBase<WT>(edge_data_, number_of_vertices_, number_of_edges_),
+    offsets(offsets_),
+    indices(indices_)
+  {}
+};
+
+/**
+ * @Synopsis    A graph stored in CSR (Compressed Sparse Row) format.
```
seunghwak: What's the difference between GraphCSR and GraphCSRBase? Shouldn't we document those?

cpp/src/utilities/error_utils.h
```diff
@@ -152,6 +152,7 @@ inline void check_stream(cudaStream_t stream, const char* file,
  * GDF_REQUIRE should be considered deprecated.
  *
  *---------------------------------------------------------------------------**/
+#ifndef CUDA_TRY
```
seunghwak: Can't we just include

https://github.com/rapidsai/cudf/blob/branch-0.13/cpp/include/cudf/utilities/error.hpp

?

ChuckHastings: @afender ?

I think the goal was to minimize dependencies on cudf, although I'm not sure that's necessary in this case.  I had to add the #ifndef in a couple of places in case both were included.  My code cleanup work today might eliminate that issue.

cpp/include/algorithms.hpp
```diff
@@ -0,0 +1,61 @@
...
+ *                                   Setting too small a tolerance can lead to non-convergence due to numerical roundoff. Usually values between 0.01 and 0.00001 are acceptable.
+ * @Param[in] max_iter               (optional) The maximum number of iterations before an answer is returned. This can be used to limit the execution time and do an early exit before the solver reaches the convergence tolerance.
+ *                                   If this value is lower or equal to 0 cuGRAPH will use the default value, which is 500.
+  * @Param[in] has_guess             (optional) This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user does not have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
+ *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+ * @Param[out] *pagerank             The PageRank : pagerank[i] is the PageRank of vertex i. Memory remains provided and owned by the caller.
+ *
+ * @throws                           cugraph::logic_error with a custom message when an error occurs.
```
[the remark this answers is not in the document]
ChuckHastings: Most of this documentation already existed like this, I added the tparam.  I'll adjust to look like cuDF.

cpp/include/algorithms.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * @Param[in] tolerance              Set the tolerance the approximation, this parameter should be a small magnitude value.
+ *                                   The lower the tolerance the better the approximation. If this value is 0.0f, cuGRAPH will use the default value which is 1.0E-5.
+ *                                   Setting too small a tolerance can lead to non-convergence due to numerical roundoff. Usually values between 0.01 and 0.00001 are acceptable.
+ * @Param[in] max_iter               (optional) The maximum number of iterations before an answer is returned. This can be used to limit the execution time and do an early exit before the solver reaches the convergence tolerance.
+ *                                   If this value is lower or equal to 0 cuGRAPH will use the default value, which is 500.
+  * @Param[in] has_guess             (optional) This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user does not have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
```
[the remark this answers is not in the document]
ChuckHastings: Fixed

cpp/include/algorithms.hpp
```diff
@@ -0,0 +1,61 @@
...
+ */
+/* ----------------------------------------------------------------------------*/
+template <typename VT, typename WT>
+void pagerank(experimental::GraphCSC<VT,WT> const &graph,
+              WT* pagerank,
+              size_t personalization_subset_size=0, 
+              VT* personalization_subset=nullptr, 
+              WT* personalization_values=nullptr,
+              float alpha = 0.85,
+              float tolerance = 1e-5,
```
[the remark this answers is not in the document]
ChuckHastings: I find it unlikely that we would need a double or an int64_t here.  However you are correct... there's no harm in using the larger data type.  Made this change.

cpp/include/cudf_conversion.hpp
```diff
@@ -0,0 +1,373 @@
...
+namespace cugraph {
+namespace experimental {
+
+namespace detail {
+
+void check_edge_list_columns(cudf::column_view const &src_indices,
+                             cudf::column_view const &dst_indices) {
+  CUGRAPH_EXPECTS( src_indices.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( dst_indices.null_count() == 0 , "Input column has non-zero null count");
+  CUGRAPH_EXPECTS( src_indices.type() == dst_indices.type() , "Unsupported data type" );
```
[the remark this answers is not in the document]
ChuckHastings: Deleted this entire file of functions.  The python layer is already doing this extraction.

cpp/include/graph.hpp
ChuckHastings: I'll look into this.  This was just a copy/paste from the current file.

cpp/include/graph.hpp
ChuckHastings: Honestly, I'm not sure we should even provide a default.  I think the caller should explicitly identify what class they want.  I think I'll remove the defaults.

There's no legacy code using these graph objects, so there's nothing to gain by introducing this potential ambiguity.

seunghwak: > Honestly, I'm not sure we should even provide a default
=>
I 100% agree. Better be explicit in this regard.

cpp/include/graph.hpp
ChuckHastings: Renamed things to try and clarify.

cpp/include/graph.hpp
ChuckHastings: Done

ChuckHastings:
I think this is ready for a review.  I could keep picking at it, but this seems like (assuming we like the approach) that we could merge it and translate the other algorithms to use this approach as part of the other stories that Alex already defined.

Certainly, let me know if there are things I should address here.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,217 @@
...
+ * @brief       Base class for graph stored in CSR (Compressed Sparse Row) format or CSC (Compressed Sparse Column) format
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class GraphCompressedSparseBase: public GraphBase<VT,ET,WT> {
+public:
+  VT const *offsets{nullptr};       ///< CSR offsets
```
seunghwak: This should be ET, maximum offset is determined by the maximum number of edges.

cpp/include/graph.hpp
```diff
@@ -0,0 +1,217 @@
...
+   * @param  offsets               This array of size V+1 (V is number of vertices) contains the offset of adjacency lists of every vertex.
+   *                               Offsets must be in the range [0, E] (number of edges).
+   * @param  indices               This array of size E contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  edge_data             This array of size E (number of edges) contains the weight for each edge.  This
+   *                               array can be null in which case the graph is considered unweighted.
+   * @param  number_of_vertices    The number of vertices in the graph
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  GraphCompressedSparseBase(VT const *offsets_, VT const *indices_, WT const *edge_data_,
```
seunghwak: VT const *offsets_
=>
ET const * offsets_

cpp/include/graph.hpp
```diff
@@ -0,0 +1,217 @@
...
+   * @param  offsets               This array of size V+1 (V is number of vertices) contains the offset of adjacency lists of every vertex.
+   *                               Offsets must be in the range [0, E] (number of edges).
+   * @param  indices               This array of size E contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  edge_data             This array of size E (number of edges) contains the weight for each edge.  This
+   *                               array can be null in which case the graph is considered unweighted.
+   * @param  number_of_vertices    The number of vertices in the graph
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  GraphCSR(VT const *offsets_, VT const *indices_, WT const *edge_data_,
```
seunghwak: VT const *offsets_
=>
ET const* offsets_

cpp/include/graph.hpp
```diff
@@ -0,0 +1,217 @@
...
+   * @param  offsets               This array of size V+1 (V is number of vertices) contains the offset of adjacency lists of every vertex.
+   *                               Offsets must be in the range [0, E] (number of edges).
+   * @param  indices               This array of size E contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  edge_data             This array of size E (number of edges) contains the weight for each edge.  This array
+   *                               can be null in which case the graph is considered unweighted.
+   * @param  number_of_vertices    The number of vertices in the graph
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  GraphCSC(VT const *offsets_, VT const *indices_, WT const *edge_data_,
```
seunghwak: VT const *offsets_
=>
ET const* offsets_

cpp/src/link_analysis/pagerank.cu
```diff
@@ -1,286 +1,278 @@
...
+#include <graph.hpp>
+
+namespace cugraph { 
+namespace detail {
+
+#ifdef DEBUG
+  #define PR_VERBOSE
+#endif
+
+template <typename IndexType, typename ValueType>
```
seunghwak: I think we should replace IndexType with VT and ET to be future proof.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -1,286 +1,278 @@
...
+      tol = tolerance;
+  else
+      return -1;
+
+  if (alpha <= 0.0f || alpha >= 1.0f)
+      return -1;
+
+  cudaStream_t stream{nullptr};
+
+  ALLOC_TRY((void**)&b, sizeof(ValueType) * n, stream);
```
seunghwak: We should start replacing ALLOC_TRY with rmm::device_vector or rmm::device_buffer.

cpp/include/algorithms.hpp
```diff
@@ -0,0 +1,65 @@
...
+ *                                   Setting too small a tolerance can lead to non-convergence due to numerical roundoff. Usually values between 0.01 and 0.00001 are acceptable.
+ * @param[in] max_iter               (optional) The maximum number of iterations before an answer is returned. This can be used to limit the execution time and do an early exit before the solver reaches the convergence tolerance.
+ *                                   If this value is lower or equal to 0 cuGRAPH will use the default value, which is 500.
+ * @param[in] has_guess              (optional) This parameter is used to notify cuGRAPH if it should use a user-provided initial guess. False means the user does not have a guess, in this case cuGRAPH will use a uniform vector set to 1/V.
+ *                                   If the value is True, cuGRAPH will read the pagerank parameter and use this as an initial guess.
+ * @param[out] *pagerank             The PageRank : pagerank[i] is the PageRank of vertex i. Memory remains provided and owned by the caller.
+ *
+ */
+template <typename VT, typename ET, typename WT>
```
seunghwak: And this might be a bit out-of-scope for this PR,

but we should think more about consistently naming VT, ET, WT.

One concern about this naming is that we're using "value" for "weight" in multiple places. If we consistently use this, then we should use VT (vertex), ET, and VT (value); so we have two VTs.

So, if we decide to consistently use value, we'd better go for vertex_t, edge_t, and value_t.

If we decide to consistently use weight, we'd better avoid using value for weight.

And I don't think pagerank has WT. WT* pagerank or WT* personalization values should have a separate type from edge weights (or edge values).

cpp/include/graph.hpp
ChuckHastings: Fixed all of these.

cpp/src/link_analysis/pagerank.cu
ChuckHastings: I left these because the current implementation assumes both VT and ET are int.  Changing to support VT and ET being different sizes and supporting int64_t for either of these types would require more extensive changes, I kind of felt like that was appropriate for a different PR.

I felt like partially changing this would be deceptive and perhaps might make us think that this would work for the future data structure.

seunghwak: I see, then could you submit another issue report on this problem, so we'll not forget about this?

cpp/src/link_analysis/pagerank.cu
ChuckHastings: Made this change where it could be made.

cpp/src/link_analysis/pagerank.cu
ChuckHastings: Created https://github.com/rapidsai/cugraph/issues/729

cpp/include/algorithms.hpp
ChuckHastings: I see two issues here, that we should probably resolve...

1. Our general concept of what a "weighted graph" or "property graph" means, and
1. Consistent naming of these things within the software.

I started a google doc to discuss the first.  I'm inclined to continue that discussion and create a new issue to address any changes we decide to make - unless we can come to a consensus within the next couple of days.

The second issue is something we could grapple with.  I used VT, ET and WT out of Alex's PR that started trying to address this issue.  I tend to prefer less ambiguous names, vertex_t, edge_t and (value_t or weight_t) seem less ambiguous to me.

It might also be beneficial to define some typedefs within the Graph to expose those.  We could consider having the graph type passed into page rank be the templated type and have the parameters from page rank "discover" vertex_t and edge_t from the templated graph object.  That might lead to fewer compiler issues.

I started with what I thought would be the least invasive approach here.

seunghwak: Yup, we have multiple issues, and it is impractical to address all those issues in a single PR, just resolve issues you can locally resolve, and for much more involved issues, let's document those and discuss how to resolve those later. I'll just approve this PR, but let me know if you need more reviews.

BradReesWork:
rerun tests

PR #728, opened 2020-02-21
Temporary fix for dask issue

Iroy30:


Iroy30:
> Looks good to me.
> 
> I see this is the temporary fix :
> 
> * What is the trade-off?
> * What would be the long term fix?

Numba expects numba.cuda.cudadrv.driver.MemoryPointer for getting ipc_handle and searches specifically for an attribute called handle. However, cudf has recently switched to using Buffers. Currently the workaround is to ducktype the handle attribute(which is a pointer of type size_t) in the Buffers. This requires an additional step. The long term fix would either be numba updating their code to accept objects of types other than their MemoryPointer or cudf adding the handle attribute to their Buffers.

PR #731, opened 2020-02-27
[WIP][skip-ci] Add native betweenness centrality

xcadet:
This is a work in progress PR for the ongoing implementation of Betweenness Centrality.

### DIFF
--- a/cpp/src/centrality/betweenness_centrality.cuh
+++ b/cpp/src/centrality/betweenness_centrality.cuh
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Author: Xavier Cadet xcadet@nvidia.com
+#pragma once
+
+namespace cugraph {
+namespace detail {
+template <typename IndexType, typename DistType>
+class BC {
+   private:
+      DistType* betweenness;
+
+      cudaStream_t stream;
+      void setup();
+      void clean();
+
+   public:
+      virtual ~BC(void) { clean(); }
+      BC(cudaStream_t _stream = 0) : stream(_stream) { setup(); }
+
+      void configure();
+      void compute();
+};
+} // namespace detail
+
+} // namespace cugraph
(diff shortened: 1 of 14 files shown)


python/cugraph/traversal/sssp_wrapper.pyx
```diff
@@ -77,6 +77,10 @@ def sssp(input_graph, source):
     df['distance'] = cudf.Series(np.zeros(num_verts, dtype=data_type))
     df['predecessor'] = cudf.Series(np.zeros(num_verts, dtype=np.int32))
 
+    # TODO(xcadet) Should be generated only if user asks for it
```
afender: We will need a python/cugraph/centrality/bc_wrapper.pyx

cpp/src/traversal/sssp_kernels.cuh
```diff
@@ -462,7 +469,7 @@ __global__ void relax_edges(
               DistType* update_addr = &next_distances[dst_id];
               DistType old_val = distances[dst_id];
               DistType new_val = distances[src_id] + edge_weights[edge];
-              if (new_val < old_val) {
+              if (new_val <= old_val) {
```
pgera: This will break sssp for graphs with self loops and 0 weight cycles if you use this information to determine predecessors. Bellman Ford is supposed to work for such graphs and return the shortest simple path.

cpp/src/traversal/sssp_kernels.cuh
```diff
@@ -236,13 +238,18 @@ __global__ void populate_frontier_and_preds(
                     vec_frontier_candidate[iv] = dst_id;
                     ++naccepted_vertices;
                   }
-
-                  // Add src_id to predecessor in either case if needed
-                  if (predecessors) {
-                    predecessors[dst_id] = src_id;
+                }
```
pgera: The BC specific flows (here and elsewhere) should ideally be behind a flag or a template param. The additional atomics and data transfers between the different predecessor arrays will likely affect performance.

cpp/src/centrality/betweenness_centrality.cuh
```diff
@@ -0,0 +1,69 @@
...
+      VT const *sample_seeds;    //
+      VT number_of_sample_seeds; //
+
+      // --- Output ----
+      // betweenness is set/read by users - using Vectors
+      result_t *betweenness = nullptr;
+
+      // --- Data required to perform computation ----
+      WT *distances = nullptr;      // array<WT>(|V|) stores the distances gathered by the latest SSSP
+      VT *predecessors = nullptr;   // array<WT>(|V|) stores the predecessors of the latest SSSP
```
kaatish: Can all of the pointers in this class be changed to rmm::device_vectors?

cpp/src/traversal/sssp_kernels.cuh
xcadet: Thank you for pointing it out! I am currently looking into other approaches

cpp/src/traversal/sssp_kernels.cuh
xcadet: Indeed, I will try to avoid affecting the performance of the core SSSP  with the BC specific modifications

PR #733, opened 2020-02-28
[REVIEW] fix multi-colum renumbering 

BradReesWork:
Fixed an issue that iroy30 found where the index of the returned series causes the order to changed.  Added a reset_index call to the function

### DIFF
--- a/python/cugraph/structure/renumber.py
+++ b/python/cugraph/structure/renumber.py
@@ -104,7 +104,7 @@ def renumber_from_cudf(_df, source_cols_names, dest_cols_names):
     dst_ids : cudf.Series
         The new destination vertex IDs
     numbering_df : cudf.DataFrame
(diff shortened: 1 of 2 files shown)


python/cugraph/structure/renumber.py
```diff
@@ -167,4 +167,10 @@ def renumber_from_cudf(_df, source_cols_names, dest_cols_names):
     _dst_ids = _tmp_df_dst.merge(
         _tmp_df, on=_vals, how='left').drop(_vals).sort_values(by='index')
 
-    return _src_ids['id'], _dst_ids['id'], _tmp_df
+    _s_id = cudf.Series(_src_ids['id']).reset_index(drop=True)
```
Iroy30: We could change the line154 similar to this, i.e, _tmp_df = _tmp_df.drop_duplicates().reset_index(drop=True) since it is faster

PR #734, opened 2020-02-28
[REVIEW] edit viewing of edgelist for Graph()

### LINKED ISSUE #603: [FEA] Undirected graph type upgrade
Python now has `Graph` (undirected) and `DiGraph` (directed).

Internally, a `Graph` should store each undirected edge `(u,v)` as two directed edges `(u,v)` and `(v,u)`. This is needed for all cuda kernels and solvers to work. This is an implementation detail that should not be exposed to the user. Users are expected to provide undirected edges when creating an undirected `Graph`.

Currently, the initialization matches Nx, the only difference is in storage and viewing. Nx stores and displays Graph() as undirected while cugraph stores and displays as undirected symmetrized. We will keep storing undirected symmetrized internally as we do now and ‘unsymmetrize’ it for display when the user asks.

We need : 
- a feature that returns the number of undirected edges from our internal representation
- a feature that displays / returns the `Graph` as a set of undirected edges from our internal representation

Iroy30:
-> Fixed column access issues in unrenumbering after recent cudf changes. Added test to test this path for future detection of errors.
-> Closes #603. There were two unsymmetrizing ways considered. 
1) Return a partial view of the Dataframe with initial unsymmetrized input from user
2) Return the upper triangle of the matrix(As per @afender suggestion). 
Method 2) was implemented since it produced results more similar to networkX


Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
Rerun tests

Iroy30:
rerun tests

python/cugraph/tests/test_unrenumber.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
ChuckHastings: Should update to 2020, I think.

python/cugraph/link_prediction/jaccard_wrapper.pyx
```diff
@@ -114,8 +114,8 @@ def jaccard(input_graph, vertex_pair=None):
             if isinstance(input_graph.edgelist.renumber_map, cudf.DataFrame):
                 unrenumered_df_ = df.merge(input_graph.edgelist.renumber_map, left_on='source', right_on='id', how='left').drop(['id', 'source'])
```
ChuckHastings: Should we change variable to unrenumbered_df_?  Doesn't really matter, but the spelling error distracts me :-(

Iroy30: oops, typo, fixing it

python/cugraph/tests/test_unrenumber.py
Iroy30: updated

python/cugraph/link_prediction/jaccard_wrapper.pyx
Iroy30: fixed, thanks

Iroy30:
rerun tests

afender:
rerun tests

Iroy30:
rerun tests

Iroy30:
> I like the elegant python implementation.
> 
> Just need to :
> 
> * Add tests
> * Document the behavior of the view function for `Graph`

Looking more closely into networkx, I came across some discrepancies regarding whether they keep their upper triangle or the user input for Graph() edgelist :
1) This shows they keep upper triangle
>>> g=nx.Graph(); g.add_edges_from([(0,1),(1,2),(2,1),(2,0),(3,2)])
>>> g.edges()
EdgeView([(0, 1), (0, 2), (1, 2), (2, 3)])
2) This shows they don't
>>> g=nx.Graph(); g.add_edges_from([(2, 0), (0, 2), (0, 1), (1, 0),(2,1)])
>>> g.edges()
EdgeView([(2, 0), (2, 1), (0, 1)])

I believe this is because internally, networkx renumbers the input and stores the upper triangle of this renumbered input which might or might not correlate to the upper triangle of original input. However, in cugraph, we have our own renumbering which doesn't match networkx (since we have a sorted renumber_map order). Hence we cannot be exactly compatible with networkx in display of edgelist, even though we follow the same method of keeping the upper triangle of the edgelist matrix.

PR #735, opened 2020-02-28
[WIP] [skip-ci] Integration gunrock's betweenness centrality

ChuckHastings:
This PR integrates the betweenness centrality implementation from gunrock into cugraph.

For this release, the gunrock analytics assume that they are called with host-side arrays.  They are working to fix this, however in an effort to get the integration underway this PR will copy the graph (and any other pertinent information) from the device onto the host before calling gunrock.  Once gunrock has adapted to accepting device-side arrays we will update to eliminate the extra copy.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -164,6 +164,29 @@ if(BUILD_TESTS)
     endif(GTEST_FOUND)
 endif(BUILD_TESTS)
 
+###################################################################################################
+# - gunrock ------------------------------------------------------------------------------------------
+#   TODO:  TEMPORARY... Rick is working on a better solution
+
+find_path(GUNROCK_INCLUDE "gunrock"
+    HINTS
+    "$ENV{GUNROCK_ROOT}/include"
+    "$ENV{CONDA_PREFIX}/include/gunrock"
+    "$ENV{CONDA_PREFIX}/include")
+
+find_library(GUNROCK_LIBRARY "gunrock"
+    HINTS
+    "$ENV{GUNROCK_ROOT}/lib"
+    "$ENV{CONDA_PREFIX}/lib")
+
+message(STATUS "GUNROCK: GUNROCK_LIBRARY set to ${GUNROCK_LIBRARY}")
+message(STATUS "GUNROCK: GUNROCK_INCLUDE set to ${GUNROCK_INCLUDE}")
+
+add_library(gunrock STATIC IMPORTED ${GUNROCK_LIBRARY})
+if (GUNROCK_INCLUDE AND GUNROCK_LIBRARY)
+    set_target_properties(gunrock PROPERTIES IMPORTED_LOCATION ${GUNROCK_LIBRARY})
+endif (GUNROCK_INCLUDE AND GUNROCK_LIBRARY)
+
 
 ###################################################################################################
 # - NVStrings -------------------------------------------------------------------------------------
@@ -286,6 +309,7 @@ add_library(cugraph SHARED
     src/snmg/utils.cu
     src/components/connectivity.cu
     src/centrality/katz_centrality.cu
+    src/centrality/betweenness_centrality.cu
     src/snmg/degree/degree.cu
     src/snmg/COO2CSR/COO2CSR.cu
     src/nvgraph/arnoldi.cu
@@ -345,7 +369,7 @@ target_include_directories(cugraph
 # - link libraries --------------------------------------------------------------------------------
 
 target_link_libraries(cugraph PRIVATE
-    ${CUDF_LIBRARY} ${RMM_LIBRARY} ${NVSTRINGS_LIBRARY} cublas cusparse curand cusolver cudart cuda ${LIBCYPHERPARSER_LIBRARY})
+    ${CUDF_LIBRARY} ${RMM_LIBRARY} ${GUNROCK_LIBRARY} ${NVSTRINGS_LIBRARY} cublas cusparse curand cusolver cudart cuda ${LIBCYPHERPARSER_LIBRARY})
 if(OpenMP_CXX_FOUND)
 target_link_libraries(cugraph PRIVATE
 ###################################################################################################
--- a/cpp/include/algorithms.hpp
+++ b/cpp/include/algorithms.hpp
@@ -62,4 +62,28 @@ void pagerank(experimental::GraphCSC<VT,ET,WT> const &graph,
               int64_t max_iter = 500,
               bool has_guess = false);
 
+/**
+ * @brief     Compute betweenness centrality for a graph
+ *
+ * Betweenness centrality for a vertex is the sum of the fraction of
+ * all pairs shortest paths that pass through the vertex.
+ * 
+ * @throws                           cugraph::logic_error with a custom message when an error occurs.
+ *
+ * @tparam VT                        Type of vertex identifiers. Supported value : int (signed, 32-bit)
+ * @tparam ET                        Type of edge identifiers.  Supported value : int (signed, 32-bit)
+ * @tparam WT                        Type of edge weights. Supported values : float or double.   
+ * @tparam result_t                  Type of computed result.  Supported values :  float
+ *
+ * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a transposed adjacency list (CSR). Edge weights are not used for this algorithm.
+ * @param[out] result                Device array of centrality scores
+ *
+ *
+ * TODO:  Could add parameters for sampling and normalization, current implementation
+ *        computes exact answer and normalizes.
+ */
+template <typename VT, typename ET, typename WT, typename result_t>
+void betweenness_centrality(experimental::GraphCSR<VT,ET,WT> const &graph,
+                            result_t *result);
+
 } //namespace cugraph
--- a/cpp/src/centrality/betweenness_centrality.cu
+++ b/cpp/src/centrality/betweenness_centrality.cu
@@ -1,0 +1,82 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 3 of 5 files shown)


BradReesWork:
are you seeing any potential memory issues between Gunrock and RMM?

ChuckHastings:
> are you seeing any potential memory issues between Gunrock and RMM?

Gunrock doesn't use RMM.  We should probably include that as a topic for your discussion at GTC.  The current version of gunrock will cudaMalloc it's own copy of the graph and internal data.  I haven't looked for memory leaks, but at first glance it looks like they free the memory when they are finished.

So, when we're using RMM in a RAPIDS notebook and we call a gunrock function it might not have enough memory to operate (if RMM has grabbed all of the memory).

ChuckHastings:
rerun tests

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -0,0 +1,115 @@
+/*
```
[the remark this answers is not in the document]
ChuckHastings: We need a source file to compile the object into the library.  Since there are no kernels, I could make it a .cpp file.  This would compile slightly faster and matches the technique used in cuDF.

cpp/include/algorithms.hpp
```diff
@@ -62,4 +62,27 @@ void pagerank(experimental::GraphCSC<VT,ET,WT> const &graph,
...
+ *
+ * @tparam VT                        Type of vertex identifiers. Supported value : int (signed, 32-bit)
+ * @tparam ET                        Type of edge identifiers.  Supported value : int (signed, 32-bit)
+ * @tparam WT                        Type of edge weights. Supported values : float or double.   
+ * @tparam result_t                  Type of computed result.  Supported values :  float
+ *
+ * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a transposed adjacency list (CSR). Edge weights are not used for this algorithm.
+ * @param[in] normalized             If true, return normalized scores, if false return unnormalized scores.
+ * @param[out] result                Device array of centrality scores
+ *
```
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/include/algorithms.hpp
```diff
@@ -62,4 +62,27 @@ void pagerank(experimental::GraphCSC<VT,ET,WT> const &graph,
               int64_t max_iter = 500,
               bool has_guess = false);
 
+/**
+ * @brief     Compute betweenness centrality for a graph
+ *
+ * Betweenness centrality for a vertex is the sum of the fraction of
+ * all pairs shortest paths that pass through the vertex.
+ *
```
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/src/centrality/betweenness_centrality.cu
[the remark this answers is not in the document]
ChuckHastings: We don't do that too much in cugraph at this point.

cpp/CMakeLists.txt
```diff
@@ -249,6 +248,31 @@ if (RMM_INCLUDE AND RMM_LIBRARY)
...
+
+# https://cmake.org/cmake/help/v3.0/module/ExternalProject.html
+include(ExternalProject)
+
+set(CUGUNROCK_DIR ${CMAKE_CURRENT_BINARY_DIR}/cugunrock CACHE STRING
+  "Path to cugunrock repo")
+
+ExternalProject_Add(cugunrock
+  GIT_REPOSITORY    https://github.com/rapidsai/cugunrock.git
+  GIT_TAG           fea_full_bc      # provide a branch, a tag, or even a commit hash
```
afender: Question : Do we have to edit the GIT_TAG each time gunrock changes?

cpp/src/centrality/betweenness_centrality.cu
afender: Can we call it `betweeness_centrality_gunrock.cu` or anything that won't conflict with the native version in #731

cpp/include/algorithms.hpp
```diff
@@ -62,4 +62,29 @@ void pagerank(experimental::GraphCSC<VT,ET,WT> const &graph,
...
+ * @tparam WT                        Type of edge weights. Supported values : float or double.   
+ * @tparam result_t                  Type of computed result.  Supported values :  float
+ *
+ * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a transposed adjacency list (CSR). Edge weights are not used for this algorithm.
+ * @param[out] result                Device array of centrality scores
+ * @param[in] normalized             If true, return normalized scores, if false return unnormalized scores.
+ *
+ */
+template <typename VT, typename ET, typename WT, typename result_t>
+void betweenness_centrality(experimental::GraphCSR<VT,ET,WT> const &graph,
```
afender: Can we expose all parameters in the API so that we don't break the code using it (or add another BC entry) when we add support for the rest of Nx parameters?

I noticed they are already accounted for in the internal code but not in this API it seems.

https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.betweenness_centrality.html#networkx.algorithms.centrality.betweenness_centrality

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -0,0 +1,57 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from cugraph.centrality import betweenness_centrality_wrapper
+
+
+def betweenness_centrality(G, normalized=True):
```
afender: I think we should expose all parameters in the API now so that we don't have to go through API upgrade or duplication.

https://networkx.github.io/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.betweenness_centrality.html#networkx.algorithms.centrality.betweenness_centrality

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -0,0 +1,66 @@
...
+import numpy.ctypeslib as ctypeslib
+
+
+def betweenness_centrality(input_graph, normalized):
+    """
+    Call betweenness centrality
+    """
+
+    if not input_graph.adjlist:
+        input_graph.view_adj_list()
```
Iroy30: Are datatype conversions taking place? Cugraph allows users to be able to enter datatypes incompatible with algorithms, like int64 vertices and int weights without worrying what the algorithm demands internally. Would that input still work as expected?

cpp/CMakeLists.txt
rlratzel: We have the option to do that, or we can just specify `master` and pull in the latest cugunrock changes that get pushed to `master` every time we rebuild.  This is currently set to `fea_full_bc` until cugunrock `master` has the features of that branch (build a static library, accept an install target).

CHANGELOG.md
```diff
@@ -12,6 +13,7 @@
 - PR #720 Updated error messages
 - PR #722 Refactor graph to remove gdf_column
 - PR #723 Added notebook testing to gpuCI gpu build
+- PR #737 Integrate cugunrock as a cmake ExternalProject
```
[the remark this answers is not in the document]
afender: BTW I realized https://github.com/rapidsai/cugraph/pull/737 (@rlratzel) only adds gunrock to cmake. 
Why not moving all dependencies to cmake and getting rid of submodules? AFAIK that's what cuML (@dantegd) did and recommended.

If we move forward with this, 0.13 will have some dependencies managed by git submodules and others cmake.  
Having two-path to deal with cuGraph dependencies is definitely worst than having one or the other (regardless of which one it is).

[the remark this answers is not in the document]
ChuckHastings: @Iroy30 - 737 got merged into this, I believe that we are closing 737.  I suppose I should delete this entry?

@afender - I have changed the rest to use this technique as well.  Will be in the next push.

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
ChuckHastings: We don't do anything with weights yet (although I will probably add that based on Alex' comments.

Do we create graphs with int64 vertices?  I thought we used renumber (which generates int32 vertices) for the vertex ids that were stored in the python graph object.

CHANGELOG.md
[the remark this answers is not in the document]
afender: Excellent! Thanks, @ChuckHastings.
It seems `.gitmodules` can be deleted now.

cpp/src/centrality/betweenness_centrality.cu
ChuckHastings: I left it in the same file, but put the gunrock implementation into a gunrock namespace.

That should allow our native version to be distinguishable.

python/cugraph/centrality/betweenness_centrality.py
ChuckHastings: Added all of the betweenness parameters to the python layer and pass them down to the C++ layer - where they are currently ignored.

The python layer does raise an exception if the extra parameters are passed, which we can remove later when they are supported.

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
ChuckHastings: OK, I have added datatype conversions.  I missed the significance of those when I was looking at the other code, I thought it was just for making CSR out of COO.

cpp/src/centrality/betweenness_centrality.cu
afender: That works too

python/cugraph/centrality/betweenness_centrality.py
afender: perfect, thanks

PR #737, opened 2020-03-02
[WIP] [skip-ci] Integrate cugunrock as a cmake ExternalProject

rlratzel:
Integrate cugunrock as a cmake ExternalProject.

This was only tested to build without error, still need to run a test that exercises `cugunrock` code in `libgunrock.a`

This also requires the `cugunrock` `fea_full_bc` branch that adds an install step and the ability to build a static library.


rlratzel:
This is now included in PR #735

PR #742, opened 2020-03-07
[REVIEW] Remove gdf column from jaccard

### LINKED ISSUE #654: [BUG] Remove gdf_column in all Jaccard-based features
Next step for #538 

Apply #629 to Jaccard features and variants. It makes more sense to do them all at once as they are often grouped together in the code.

- jaccard
- overlap
- weighted versions
- two hops

ChuckHastings:
Remove the gdf_column from jaccard/overlap methods.  Use the new GraphCSR object instead.

Streamlined a few things while working through the code.  Specifically, I added the weight as an optional argument to the wrapper and eliminated the need for a separate wjaccard_wrapper and woverlap_wrapper.

Closes #654


ChuckHastings:
FYI - created #743 to address some discussion about the jaccard API.

cpp/include/algorithms.hpp
```diff
@@ -62,6 +63,106 @@ void pagerank(experimental::GraphCSC<VT,ET,WT> const &graph,
...
+ * @tparam ET              Type of edge identifiers. Supported value : int (signed, 32-bit)
+ * @tparam WT              Type of edge weights. Supported value : float or double.   
+ *
+ * @param[in] graph        The input graph object
+ * @param[in] weights      device pointer to input vertex weights for weighted Jaccard, may be NULL for
+ *                         unweighted Jaccard.
+ * @param[out] result      Device pointer to result values, memory needs to be pre-allocated by caller
+ */
+template <typename VT, typename ET, typename WT>
+void jaccard(experimental::GraphCSR<VT,ET,WT> const &graph,
```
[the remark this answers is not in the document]
ChuckHastings: Pushed an update that addresses this in the python layer.

cpp/src/link_prediction/jaccard.cu
```diff
@@ -20,60 +20,67 @@
...
+#include "graph.hpp"
 #include "rmm_utils.h"
 #include "utilities/error_utils.h"
 
 namespace cugraph { 
 namespace detail {
+
   // Volume of neighboors (*weight_s)
-  template<bool weighted, typename IdxType, typename ValType>
+  template<bool weighted, typename vertex_t, typename edge_t, typename weight_t>
```
seunghwak: Not a request to change, and just a comment.

Maybe a task for 0.14.

We're mixing VT, ET, WT and vertex_t, edge_t, and weight_t, and we'd better not do this.

We need to decide on template parameter names and should use those consistently.

cpp/src/link_prediction/jaccard.cu
```diff
@@ -20,60 +20,67 @@
...
-                  ValType *work) {
-    IdxType row, start, end, length;
-    ValType sum;
+  jaccard_row_sum(vertex_t n,
+                  edge_t const *csrPtr,
+                  vertex_t const *csrInd,
+                  weight_t const *v,
+                  weight_t *work) {
+
+    vertex_t row;
```
seunghwak: This is tedious, but we may better avoid leaving variables to have an arbitrary value.

vertex_t row{};

will help us to avoid some surprising bugs in the future.

And roughly speaking, initializing with {} is safer and more modern than initializing with () or =.

int i = std::numeric_limits<long>::max() generates a warning

but

int i{std::numeric_limits<long>::max()} fails to compile.

afender:
There are 2 conflicting files after merging #734

cpp/src/link_prediction/jaccard.cu
ChuckHastings: I introduced this to trigger conversation about the names.  Agree we should not mix things.  Hopefully we can address this at the graph alignment meeting.

cpp/src/link_prediction/jaccard.cu
ChuckHastings: Agree.  I've started using the {} initialization in areas where an initialization occurs.  Hadn't really thought about when a variable is left uninitialized.  

I'll try to keep this in mind going forward.

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

PR #747, opened 2020-03-12
[REVIEW] Fix view and datatypecast

### LINKED ISSUE #752: [BUG] int64 for src/dst fails number_of_vertices
When using int32 for src/dst everything is fine. But int64:
```
G.number_of_vertices()
RuntimeError: CUGRAPH failure at: /conda/conda-bld/libcugraph_1580841414814/work/cpp/src/structure/cugraph.cu:256: Unsupported data type
```

this is only to be reproduced on my real data, not on the karate example graph from cugraph github

Iroy30:
--> Added datatype casting in pagerank wrapper
--> Updated viewing of graphs to work with int64 in case of implicit conversions(COO to CSR or vice versa)
--> Added multi-column unrenumbering for two_hop_neighbors
--> Updated number_of_vertices() to work with int64

Closes #752


Iroy30:
rerun tests

Iroy30:
rerun tests

Iroy30:
rerun tests

PR #756, opened 2020-03-13
[REVIEW] Add force atlas 2

### LINKED ISSUE #700: [FEA] ForceAtlas2 Algorithm
add the ForceAtlas2 algorithm

hlinsen:
This PR implements the Force Atlas 2 layout.
The API is based on current python implementations of the algorithm. Two versions of the algorithm are available, implementing the exact repulsion and the Barnes Hut approximation.

closes #700


GPUtester:

afender:
@hlinsen let's target branch 0.14

trxcllnt:
Would it be possible to add a callback that python can provide, similar to [cuML's UMAP](https://github.com/rapidsai/cuml/blob/branch-0.14/python/cuml/manifold/umap.pyx#L233) implementation? This would enable some really neat possibilities for visualization.

Another cool feature would be optionally accepting the position/force buffers as arguments, so the `force_atlas` could be paused/resumed with the same state.

hlinsen:
> Would it be possible to add a callback that python can provide, similar to [cuML's UMAP](https://github.com/rapidsai/cuml/blob/branch-0.14/python/cuml/manifold/umap.pyx#L233) implementation? This would enable some really neat possibilities for visualization.

Sure we can add that, this is a good idea.

> Another cool feature would be optionally accepting the position/force buffers as arguments, so the `force_atlas` could be paused/resumed with the same state.

Yes positions can be passed to the function like on current CPU implementations.

trxcllnt:
> Yes positions can be passed to the function like on current CPU implementations.

Right, I meant exposing that at the Python/Cython layer as well.

trxcllnt:
@exactlyallan asked me to provide more info about our use-case. For visualization, we're likely going to call `force_atlas2` in a python loop, with `max_iter=1`. This is why it's important for us to be able to reuse the position/force buffers between calls. Something like this is what I'd expect we will do:

```python
do_layout = true
ui_button.on('click', lambda _: run_layout = not do_layout)

x_start = cudf.Series(...)
y_start = cudf.Series(...)
pos = cudf.DataFrame({'x': x_start, 'y': y_start}).to_gpu_matrix()

while do_layout:
  force_atlas2(max_iter=1, pos, x_start, y_start, ...)
  visualize_graph(pos)
  await asyncio.sleep(0)
```

hlinsen:
> @exactlyallan asked me to provide more info about our use-case. For visualization, we're likely going to call `force_atlas2` in a python loop, with `max_iter=1`. This is why it's important for us to be able to reuse the position/force buffers between calls. Something like this is what I'd expect we will do:
> 
> ```python
> do_layout = True
> x_start = cudf.Series(...)
> y_start = cudf.Series(...)
> pos = cudf.DataFrame({'x': x_start, 'y': y_start}).to_gpu_matrix()
> 
> def start_fa2():
>   while do_layout:
>     force_atlas2(max_iter=1, pos, x_start, y_start, ...)
>     visualize_graph(pos)
>     await asyncio.sleep(0)
> 
> def start_stop_handler():
>   do_layout = not do_layout
>   if do_layout:
>     start_fa2()
> 
> start_stop_button.on('click', start_stop_handler)
> 
> start_fa2()
> ```

Sure, that makes sense. At the moment the API is accepting a cudf just like you provided with positions to start with. I also added the callback from Umap so that a (n, 2) matrix of positions is printed at each call.

GPUtester:

python/cugraph/layout/force_atlas2.py
```diff
@@ -0,0 +1,115 @@
...
+            should be undirected where an undirected edge is represented by a
+            directed edge in both direction.
+
+        max_iter : integer
+            This controls the maximum number of levels/iterations of the Force Atlas
+            algorithm. When specified the algorithm will terminate after no more
+            than the specified number of iterations. No error occurs when the
+            algorithm terminates early in this manner.
+        pos_list: cudf.DataFrame
+            Data frame with initial positions containing two columns:
```
afender, line 52: Data frame with initial positions -> Data frame with initial **vertex** positions

Also, how do you know what 'x' and 'y' correspond to after the graph is renumbered? We can't assume that it implicitly matches internal data. Should we accept "vetrex_id","x","y"?

cpp/include/algorithms.hpp
```diff
@@ -186,6 +187,69 @@ void overlap_list(experimental::GraphCSRView<VT, ET, WT> const &graph,
...
+ * @param[in] prevent_overlapping               Prevent nodes from overlapping.
+ * @param[in] edge_weight_influence             How much influence you give to the edges weight. 0
+ * is “no influence” and 1 is “normal”.
+ * @param[in] jitter_tolerance                  How much swinging you allow. Above 1 discouraged.
+ * Lower gives less speed and more precision.
+ * @param[in] barnes_hut_optimize:              Whether to use the fast Barnes Hut or use the slower
+ * exact version.
+ * @param[in] barnes_hut_theta:                 Float between 0 and 1. Tradeoff for speed (1) vs
+ * accuracy (0) for Barnes Hut only.
+ * @params[in] scaling_ratio                    How much repulsion you want. More makes a more
```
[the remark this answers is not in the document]
hlinsen: It should be strictly positive, but this is really free for the user to play with. The original paper mentions it should be adjusted when using lin log mode though.

cpp/tests/test_utils.h
```diff
@@ -318,6 +318,78 @@ int mm_properties(FILE* f, int tg, MM_typecode* t, IndexType_* m, IndexType_* n,
...
+ *  @param tg Boolean indicating whether to convert matrix to general
+ *  format (from symmetric, Hermitian, or skew symmetric format).
+ *  @param nnz Number of non-zero matrix entries.
+ *  @param adj_matrix (Output) Real component of matrix
+ *  entries. Should have at least nnz entries. Ignored if null
+ *  pointer.
+ *  @return Zero if matrix was read successfully. Otherwise non-zero.
+ */
+template <typename IndexType_, typename ValueType_>
+int mm_to_matrix(FILE* f, IndexType_ nnz, std::vector<std::vector<ValueType_>>& adj_matrix)
```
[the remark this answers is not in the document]
hlinsen: This could be reconciled with mm_properties to build the adjacency matrix at the same time.

python/cugraph/layout/force_atlas2.py
```diff
@@ -0,0 +1,115 @@
...
+            as an edge list.
+            The adjacency list will be computed if not already present. The graph
+            should be undirected where an undirected edge is represented by a
+            directed edge in both direction.
+
+        max_iter : integer
+            This controls the maximum number of levels/iterations of the Force Atlas
+            algorithm. When specified the algorithm will terminate after no more
+            than the specified number of iterations. No error occurs when the
+            algorithm terminates early in this manner.
```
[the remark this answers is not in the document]
hlinsen: Yes, FA2 achieves short term quality really fast. About 100 iterations is sufficient for a decent layout for most graphs, above 1000 is discouraged.

python/cugraph/tests/test_force_atlas2.py
```diff
@@ -0,0 +1,90 @@
...
+                               barnes_hut_optimize=barnes_hut_optimize,
+                               barnes_hut_theta=barnes_hut_theta,
+                               scaling_ratio=scaling_ratio,
+                               strong_gravity_mode=strong_gravity_mode,
+                               gravity=gravity)
+    t2 = time.time() - t1
+    print('Cugraph Time : ' + str(t2))
+    return pos
+
+DATASETS = ['../datasets/karate.csv', '../datasets/polbooks.csv', '../datasets/dolphins.csv']
```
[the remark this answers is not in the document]
hlinsen: Netscience runs fast as well, added it to the list.

python/cugraph/layout/force_atlas2.py
hlinsen: Indeed we should add it.

exactlyallan:
@trxcllnt, @AjayThorve  Could you review to see if this fits with our future plans?

ci/gpu/build.sh
```diff
@@ -66,6 +66,7 @@ conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaul
       distributed>=2.12.0 \
       dask-cudf=${MINOR_VERSION} \
       dask-cuda=${MINOR_VERSION} \
+      scikit-learn>=0.21 \
```
raydouglass, line 69: Why is scikit-learn being added as a dependency?

hlinsen: Don't we need it here since I'm relying on scikit learn for my tests ?

raydouglass: Is the dependency only for tests?

hlinsen: Yes

BradReesWork:
rerun tests

cpp/src/layout/barnes_hut.h
```diff
@@ -0,0 +1,332 @@
...
+
+  // Scalars used to adapt global speed.
+  float speed                     = 1.f;
+  float speed_efficiency          = 1.f;
+  float outbound_att_compensation = 1.f;
+  float jt                        = 0.f;
+
+  // If outboundAttractionDistribution active, compensate.
+  if (outbound_attraction_distribution) {
+    int sum                   = thrust::reduce(d_massl.begin(), d_massl.begin() + n);
```
kaatish, line 170: Use rmm exec policy

cpp/src/layout/barnes_hut.h
```diff
@@ -0,0 +1,332 @@
...
+                const float jitter_tolerance                  = 1.0,
+                const float theta                             = 0.5,
+                const float scaling_ratio                     = 2.0,
+                bool strong_gravity_mode                      = false,
+                const float gravity                           = 1.0,
+                bool verbose                                  = false,
+                internals::GraphBasedDimRedCallback *callback = nullptr)
+{
+  const edge_t e   = graph.number_of_edges;
+  const vertex_t n = graph.number_of_vertices;
```
kaatish, line 56: Would all of these functions work fine if you pass a zero edge graph? Is there any sort of checking required to avoid zero block cuda kernel launches?

cpp/src/layout/exact_fa2.h
```diff
@@ -0,0 +1,188 @@
...
+  const vertex_t *col = graph.dst_indices;
+  const weight_t *v   = graph.edge_data;
+
+  float speed                     = 1.f;
+  float speed_efficiency          = 1.f;
+  float outbound_att_compensation = 1.f;
+  float jt                        = 0.f;
+
+  if (outbound_attraction_distribution) {
+    int sum                   = thrust::reduce(mass.begin(), mass.end());
```
kaatish, line 103: rmm exec policy should be used here.

cpp/src/layout/exact_fa2.h
```diff
@@ -0,0 +1,188 @@
...
+                        d_old_forces,
+                        d_old_forces + n,
+                        d_mass,
+                        d_swinging,
+                        d_traction,
+                        n);
+
+    // Compute global swinging and traction values.
+    const float s = thrust::reduce(swinging.begin(), swinging.end());
+    const float t = thrust::reduce(traction.begin(), traction.end());
```
kaatish, line 159: rmm exec policy

cpp/src/layout/barnes_hut.h
hlinsen: So edges are used only for the attraction kernel and indeed having 0 edge graph results in a division by zero when computing block dimensions. Added a check for that so we avoid launching the kernel at all.

cpp/include/internals.h
```diff
@@ -0,0 +1,49 @@
+/*
```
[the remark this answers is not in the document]
hlinsen: Actually should the rest of the headers be renamed to .cuh or .hpp ?

BradReesWork:
rerun tests

PR #760, opened 2020-03-17
[REVIEW] Weighted KTruss

### LINKED ISSUE #577: [FEA] Wrap K-Truss from cuHornet
Expose the KTruss analytic found in cuHornet via cuGraph

kaatish:
closes #577


afender:
Also, we will need to port https://github.com/rapidsai/cudf/pull/4534 to cugraph before CI build passes

afender:
#761

afender:
rerun tests

Iroy30:
> Looks good.
> 
> Should we document ktruss behavior when weights are provided?
> 
> I am not sure if a user will always want to use the graph's weights for ktruss when the graph is a weighted graph. A `use_weights` flag could address this but Nx does not provide this. Passing an `unweighted` view is a good alternative for now.

Would that mean ktruss is going to internally run without weights in the absence of a flag?

afender:
> > Looks good.
> > Should we document ktruss behavior when weights are provided?
> > I am not sure if a user will always want to use the graph's weights for ktruss when the graph is a weighted graph. A `use_weights` flag could address this but Nx does not provide this. Passing an `unweighted` view is a good alternative for now.
> 
> Would that mean ktruss is going to internally run without weights in the absence of a flag?

Yes. 

Currently, there are two cases : 
- if there are weights attached to the graph ktruss uses them
- if there are no weights ktruss does not use weights

 My point is a user could have a graph with weights attached to it but these weights may not be relevant for ktruss algo. What should the user do in this case?

Iroy30:
Currently cugraph doesn't expose a way to remove edge_attribute/weights or have a unweighted view of the whole Graph. Passing an unweighted graph would need to be explicitly done by the user's script. The way would be to either create a new graph without weights(edge_attr) or utilize the current graph G with weights to create a new graph G_new = cugraph.from_cudf_edgelist(G.edgelist.edgelist_df, 'src', 'dst' , create_using=_GraphType_) , and using G_new for ktruss call.

BradReesWork:
rerun tests

BradReesWork:
rerun tests

PR #774, opened 2020-03-20
[REVIEW] [BUG] Raise TypeError if a DiGraph is used with spectral*Clustering()

rlratzel:
Updated tests, added tests, and changed Cython code to ensure a `DiGraph` cannot be passed to `cugraph.spectralModularityMaximizationClustering()` or `cugraph.spectralBalancedCutClustering()`

The new code will result in the following traceback if a `DiGraph` is used for the above algos:
```
  File "cugraph/community/spectral_clustering_wrapper.pyx", line 50, in cugraph.community.spectral_clustering_wrapper.spectralBalancedCutClustering
TypeError: DiGraph objects are not supported
```

I also added a FIXME to note a possible infinite loop situation encountered during benchmark development. The above type check for a `DiGraph` would have prevented this, but would have also masked it as well.

_sorry for the noise in the diff - my editor automatically removes trailing whitespace and there apparently was a lot in `kmeans.cu`_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - PR #753 ECG Error
 - PR #758 Fix for graph comparison failure
 - PR #761 Added flag to not treat deprecation warnings as errors, for now
+- PR #774 Raise TypeError if a DiGraph is used with spectral*Clustering()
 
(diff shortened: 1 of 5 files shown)


BradReesWork:
If a DiGraph is passed in, why do we not just convert it to a Graph?

Issue #245, opened 2019-04-25
[BENCH] Benchmark Jaccard's performance issue after recent changes


afender:
When computing all weights ( including 2 hops) cuGraph is not faster than Nx. We need to identify the performance bottleneck and find out how to make Jaccard faster for this use case.

ChuckHastings:
Are we certain that this is true?  How did we measure this?

I recently discovered (by code inspection of networkx) that networkx's Jaccard implementation does lazy computation of the similarity.  So the time spent calling nx.jaccard_coefficient itself **only** includes the time required to construct a lazy iterator.  The real time spent computing nx.jaccard_coefficient would include populating a pandas frame with all of the results.  It seems like that should explode for even a small graph.

BradReesWork:
Dropping this issues.  The Jaccard code is slated to be complexly redone in 0.17 for MNMG at which time it should also be update to better address 2-hop comparisons and performance

PR #780, opened 2020-03-24
[REVIEW] Remove gdf_column from cuhornet features

### LINKED ISSUE #655: [BUG] Remove gdf_column in cuHornet based features
Next step for #538 

Apply #629 to all cuHornet based features (listed below). It makes more sense to do them all at once as they share a similar design.

- katz_centrality
- k_core
- core_number

ChuckHastings:
Modify the algorithms we use from cuhornet to remove use of the gdf_column.  This includes:

- Katz_centrality
- Core Number
- K-Core

Had to also update the way that degrees and two_hop_neighbors are calculated.

Closes #655


cpp/include/functions.h
```diff
@@ -182,22 +182,6 @@ void delete_edge_list(Graph* graph);
...
- *
- * @param[out] first                 out : An uninitialized gdf_column which will be initialized to contain the
- *                                         first entry of each result pair.
- * @param[out] second                out : An uninitialized gdf_column which will be initialized to contain the
- *                                         second entry of each result pair.
- *
- * @throws     cugraph::logic_error when an error occurs.
- */
-/* ----------------------------------------------------------------------------*/
-void get_two_hop_neighbors(Graph* graph, gdf_column* first, gdf_column* second);
```
ChuckHastings: NOTE:  I made get_two_hop_neighbors an algorithm.  Seemed more like an algorithm than a utility.

cpp/include/functions.h
```diff
@@ -219,22 +203,6 @@ void get_two_hop_neighbors(Graph* graph, gdf_column* first, gdf_column* second);
...
- *                                         1 : in-degree
- *                                         2 : out-degree
- *
- * @Param[out] *degree               out : gdf_column of size V (V is number of vertices) initialized to zeros.
- *                                         Contains the computed degree of every vertex.
- *
- * @throws     cugraph::logic_error when an error occurs.
- */
-/* ----------------------------------------------------------------------------*/
-void degree(Graph* graph, gdf_column *degree, int x);
```
ChuckHastings: NOTE: I made degree a function on the Graph object.  This seemed more like a function of the graph to me.  It also allowed me to easily customize the implementation based on whether the graph was CSR/CSC (where the computation is just a subtraction on the offsets array) or COO (where the computation is a histogram)

cpp/src/ktruss/ktruss.cu
```diff
@@ -82,9 +82,8 @@ void ktruss_subgraph_impl(experimental::GraphCOO<VT, ET, WT> const &graph,
 
   kt.copyGraph(out_src, out_dst);
 
-  experimental::GraphCOO<VT, ET, WT> subgraph(
```
ChuckHastings: Fixed constness in GraphCOO, so this is unnecessary

ChuckHastings:
rerun tests

cpp/src/structure/graph.cu
```diff
@@ -28,7 +61,61 @@ void GraphCompressedSparseBase<VT,ET,WT>::get_source_indices(VT *src_indices) co
...
+    degree_from_vertex_ids(GraphBase<VT,ET,WT>::number_of_edges, src_indices, degree, stream);
+  }
+
+  if (x != 2) {
+    degree_from_vertex_ids(GraphBase<VT,ET,WT>::number_of_edges, dst_indices, degree, stream);
+  }
+}
+
+template <typename VT, typename ET, typename WT>
+void GraphCompressedSparseBase<VT,ET,WT>::degree(ET *degree, int x) const {
```
kaatish: Might be overkill but would it make sense to make x an enum?

cpp/tests/centrality/katz_centrality_test.cu
```diff
@@ -72,56 +77,55 @@ typedef struct Katz_Usecase_t {
...
   Tests_Katz() {}
   static void SetupTestCase() {}
   static void TearDownTestCase() {}
   virtual void SetUp() {}
   virtual void TearDown() {}
 
   void run_current_test(const Katz_Usecase& param) {
-       Graph_ptr G{new cugraph::Graph, Graph_deleter};
-       gdf_column_ptr col_src, col_dest, col_katz_centrality;
+    FILE* fpin = fopen(param.matrix_file.c_str(),"r");
```
kaatish: It would be great if we moved to the C++ style of file reading.

ChuckHastings: I haven't focused on this while removing gdf_column, but I'd be happy with consistency between the tests.  We have many different ways that we use of loading data for the tests.

cpp/include/functions.h
BradReesWork: what about degrees (get in and out degree rather than just total count)?

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2019, NVIDIA CORPORATION.
+# Copyright (c) 2020, NVIDIA CORPORATION.
```
BradReesWork: why not 2019-2020 like the rest of the header blocks?

cpp/include/functions.h
ChuckHastings: degrees is implemented in python as two calls to this function, with x = 1 and x = 2.  That's how it was, I didn't think about changing it.  The only real performance impact would be that the vertex id column gets created an extra time in the degrees call.

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
ChuckHastings: This was a copy/paste error previously and I noticed it now.  The betweenness centrality code was new in 2020.  I thought the convention was to copyright new code in the year that it was created and to update the copyright when you change existing code.

cpp/src/structure/graph.cu
ChuckHastings: We do that in other cases, it would be a little clearer.  I've mostly been keeping the interfaces the same as I work through the gdf_column changes except for the gdf_column portions of the interface.

This might be a  reasonable place to make a further change to the C++ API.  Opinions?  @afender , @BradReesWork?

cpp/src/traversal/two_hop_neighbors.cu
```diff
@@ -29,154 +31,96 @@
...
+    block_bucket_offsets[num_blocks] = graph.number_of_edges;
 
     // Scatter the expanded edge lists into temp space
-    grid.x = min((IndexType) MAXBLOCKS, num_blocks);
-    scatter_expand_kernel<<<grid, block, 0, nullptr>>>(exsum_degree,
-                                                       indices,
-                                                       offsets,
-                                                       block_bucket_offsets,
-                                                       num_verts,
+    grid.x = min((ET) MAXBLOCKS, num_blocks);
```
afender: grid is a `dim3` which is afaik based on `uint3`. 
`ET` is expected to be 64 bit soon and `MAXBLOCKS` is a constant that does not require 64bits. It is probably safer to avoid involving 64 bit here when dealing with grid sizes.

cpp/src/traversal/two_hop_neighbors.cu
```diff
@@ -29,154 +31,96 @@
...
-    auto tuple_start = thrust::make_zip_iterator(thrust::make_tuple(first_pair, second_pair));
+    auto tuple_start = thrust::make_zip_iterator(thrust::make_tuple(d_first_pair, d_second_pair));
     auto tuple_end = tuple_start + output_size;
     thrust::sort(rmm::exec_policy(stream)->on(stream), tuple_start, tuple_end);
     tuple_end = thrust::copy_if(rmm::exec_policy(stream)->on(stream),
                                                             tuple_start,
                                                             tuple_end,
                                                             tuple_start,
-                                                            self_loop_flagger<IndexType>());
+                                                            self_loop_flagger<ET>());
```
afender: Should this be `VT` since we are looking at indices pairs and not offsets?
```suggestion
                                                            self_loop_flagger<VT>());
```

cpp/src/traversal/two_hop_neighbors.cuh
```diff
@@ -25,111 +25,106 @@
...
-			return (vec[high] <= val) ? high : low;
-
-		IndexType mid = low + (high - low) / 2;
-
-		if (vec[mid] > val)
-			high = mid - 1;
-		else
-			low = mid;
-	}
+                                     const IndexType val,
```
afender: typename consistency would be great. I found IndexType ambiguous now that we have VT and ET.

cpp/tests/centrality/katz_centrality_test.cu
afender: > It would be great if we moved to the C++ style of file reading.

Would you expect a big performance difference or is it primarily motivated by having a more consistent C++ style?

python/cugraph/centrality/katz_centrality_wrapper.pyx
```diff
@@ -36,38 +34,28 @@ def katz_centrality(input_graph, alpha=0.1, max_iter=100, tol=1.0e-5, nstart=Non
...
-        add_adj_list(g)
-        offsets, indices, values = graph_wrapper.get_adj_list(graph)
-        input_graph.adjlist = input_graph.AdjList(offsets, indices, values)
+    [offsets, indices] = graph_wrapper.datatype_cast([input_graph.adjlist.offsets, input_graph.adjlist.indices], [np.int32])
 
-    # we should add get_number_of_vertices() to Graph (and this should be
-    # used instead of g.adjList.offsets.size - 1)
-    num_verts = g.adjList.offsets.size - 1
+    num_verts = input_graph.number_of_vertices()
+    num_edges = len(indices)
```
afender: How about
```suggestion
    num_edges = input_graph.number_of_edges()
```

cpp/include/graph.hpp
```diff
@@ -88,7 +109,7 @@ class GraphCOO: public GraphBase<VT, ET, WT> {
    * @param  number_of_vertices    The number of vertices in the graph
    * @param  number_of_edges       The number of edges in the graph
    */
-  GraphCOO(VT const *src_indices_, VT const *dst_indices_, WT const *edge_data_,
+  GraphCOO(VT *src_indices_, VT *dst_indices_, WT *edge_data_,
```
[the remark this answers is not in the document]
ChuckHastings: In the case where we *create* a GraphCOO (or technically a GraphCSR/GraphCSC) we want to be able to modify the contents of the indices arrays.

When we pass the GraphCOO/GraphCSR/GraphCSC into a function from python we use the const qualifier to indicate that we won't change the Graph.

This *would* allow us to create a GraphCOO from a cudf column and change the contents of the cudf column.  But only if we had an algorithm that accepted as input a GraphCOO &.  All of the algorithms take as input a const GraphCOO&.

python/cugraph/centrality/katz_centrality_wrapper.pyx
ChuckHastings: input_graph.number_of_edges(), in the case of a Graph object (undirected Graph), returns len(indices)/2 (the number of unique unsymmetrized edges) based upon a python change that @Iroy30 put in during 0.13.  This num_edges variable needs to be the number of edges present in the COO/CSR data structure.

I originally had used the number_of_edges() function until Ishika's PR was merged.

cpp/src/traversal/two_hop_neighbors.cu
ChuckHastings: If ET becomes 64-bit at some point, then somewhere in the derivation of grid.x we need to reduce from 64-bit to 32-bit.  I had deferred that to the last moment (here).  I could make num_blocks a 32-bit integer.  I felt like deferring to this point was more future-proof - in case dim3 eventually changes to a different type we don't have to worry about changing types elsewhere.

cpp/src/traversal/two_hop_neighbors.cu
ChuckHastings: Fixed

cpp/src/traversal/two_hop_neighbors.cuh
ChuckHastings: I can change this.  I didn't tease through the entire implementation to understand which things where VT and which things were ET, I just let the compiler figure it out for me.  I'll work through the logic and update them.

cpp/tests/centrality/katz_centrality_test.cu
ChuckHastings: Note that we're using a set of functions in test_utils.h to manipulate the Matrix Market format.  This fopen call opens the file so we can pass the FILE * into the other test functions.  Changing to C++ style would require rewriting all of that code.

kaatish: > > It would be great if we moved to the C++ style of file reading.
> 
> Would you expect a big performance difference or is it primarily motivated by having a more consistent C++ style?

Just for consistency. Since we are trying to move (almost) everything in cugraph to RAII style it would be nice to take care of it in this cycle.

python/cugraph/centrality/katz_centrality_wrapper.pyx
afender: Makes sense. I forgot about this side effect.

cpp/include/functions.h
BradReesWork: it does need to be redone for better performance (https://github.com/rapidsai/cugraph/issues/245)

cpp/src/structure/graph.cu
BradReesWork: If you can quickly create it as an enum, that would be great.  Else you can add the /** TO D0 **/ tag and it will be added to our tech debt

cpp/tests/centrality/katz_centrality_test.cu
ChuckHastings: @kaatish I'd suggest writing up an issue for this so it can get prioritized.  It would potentially affect all of the C++ unit tests.

ChuckHastings:
I think I made all of the requested changes.

PR #782, opened 2020-03-25
[REVIEW] Use Cython's `new_build_ext` (if available)

trxcllnt:
Copies over the changes made in this PR: https://github.com/rapidsai/ucx-py/pull/462

Using `new_build_ext` allows parallelizing the `cythonize()` call from the command line via `-j` like so:
```
python setup.py -j<num_cores>
```


trxcllnt:
rerun tests

PR #796, opened 2020-03-31
[WIP] [skip-ci] new idea for returning graphs

ChuckHastings:
Per slack conversation... an idea for extending the graph structures for returning graphs

### DIFF
--- a/cpp/include/graph.hpp
+++ b/cpp/include/graph.hpp
@@ -214,5 +214,182 @@ public:
   {}
 };
 
+/**
+ * @brief       Base class for graphs constructed in the C++ API
+ *
+ * This class will own edge data, until the data is moved.
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphBase {
+public:
+  std::unique_ptr<WT>  edge_data{};     ///< edge weight
+  VT                   number_of_vertices{0};
+  ET                   number_of_edges{0};
+
+  ConstructedGraphBase() {}
+
+  ConstructedGraphBase(std::unique_ptr<WT> &&edge_data_, VT number_of_vertices_, ET number_of_edges_):
+    edge_data(edge_data_),
+    number_of_vertices(number_of_vertices_),
+    number_of_edges(number_of_edges_)
+  {}
+};
+
+/**
+ * @brief       A constructed graph stored in COO (COOrdinate) format.
+ *
+ * This class will src_indices and dst_indicies (until moved)
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCOO: public ConstructedGraphBase<VT, ET, WT> {
+public:
+  std::unique_ptr<VT> src_indices{};   ///< rowInd
+  std::unique_ptr<VT> dst_indices{};   ///< colInd
+
+  /**
+   * @brief      Default constructor
+   */
+  ConstructedGraphCOO(): ConstructedGraphBase<VT,ET,WT>() {}
+  
+  /**
+   * @brief      Take ownership of the provided graph arrays in COO format
+   *
+   * @param  source_indices        This array of size E (number of edges) contains the index of the source for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  destination_indices   This array of size E (number of edges) contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  edge_data             This array size E (number of edges) contains the weight for each edge.  This array can be null
+   *                               in which case the graph is considered unweighted.
+   * @param  number_of_vertices    The number of vertices in the graph
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  ConstructedGraphCOO(std::unique_ptr<VT> &&src_indices_,
+                      std::unique_ptr<VT> &&dst_indices_,
+                      std::unique_ptr<VT> &&edge_data_,
+                      VT number_of_vertices_,
+                      ET number_of_edges_):
+    ConstructedGraphBase<VT,ET,WT>(edge_data_, number_of_vertices_, number_of_edges_),
+    src_indices(src_indices_), dst_indices(dst_indices_)
+  {}
+};
+
+/**
+ * @brief       Base class for constructted graphs stored in CSR (Compressed Sparse Row) format or CSC (Compressed Sparse Column) format
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCompressedSparseBase: public ConstructedGraphBase<VT,ET,WT> {
+public:
+  std::unique_ptr<VT> offsets{};   ///< CSR offsets
+  std::unique_ptr<VT> indices{};   ///< CSR indices
+
+  /**
+   * @brief      Take ownership of the provided graph arrays in CSR/CSC format
+   *
+   * @param  offsets               This array of size V+1 (V is number of vertices) contains the offset of adjacency lists of every vertex.
+   *                               Offsets must be in the range [0, E] (number of edges).
+   * @param  indices               This array of size E contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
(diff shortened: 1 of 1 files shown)


cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+ * @brief       A constructed graph stored in COO (COOrdinate) format.
+ *
+ * This class will src_indices and dst_indicies (until moved)
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCOO: public ConstructedGraphBase<VT, ET, WT> {
```
ChuckHastings, line 252: This class would allow k_core and k_truss to return a subgraph in COO format.  We can generate unique_ptr<VT> within the algorithms which will clean up if the algorithm aborts early.  At the end we pass them into the constructor for ConstructedGraphCOO, copy them into the output and return.

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+
+/**
+ * @brief       Base class for constructted graphs stored in CSR (Compressed Sparse Row) format or CSC (Compressed Sparse Column) format
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCompressedSparseBase: public ConstructedGraphBase<VT,ET,WT> {
```
ChuckHastings, line 292: This and the ConstructedGraphCSR/ConstructedGraphCSC below are here for generating a CSR structure.  coo2csr could use this instead of its current structure, which would be safer because of the unique_ptr.

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+/**
+ * @brief       Base class for graphs constructed in the C++ API
+ *
+ * This class will own edge data, until the data is moved.
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
```
seunghwak, line 226: I think this is a good time to decide between VT, ET, WT and vertex_t, edge_t, weight_t (or value_t). May start a slack discussion and a quick vote?

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+class ConstructedGraphBase {
+public:
+  std::unique_ptr<WT>  edge_data{};     ///< edge weight
+  VT                   number_of_vertices{0};
+  ET                   number_of_edges{0};
+
+  ConstructedGraphBase() {}
+
+  ConstructedGraphBase(std::unique_ptr<WT> &&edge_data_, VT number_of_vertices_, ET number_of_edges_):
```
seunghwak, line 235: Similarly, we should start a discussion on naming conventions for function input parameters for constructors and class member variables.

Here, you're adding _ at the of constructor input parameters. In other places, we are adding _ at the end (or at the beginning) of class member variables. And in yet other places, we are doing nothing. We need to do something and start to enforce consistency.

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+ * This class will own edge data, until the data is moved.
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphBase {
+public:
+  std::unique_ptr<WT>  edge_data{};     ///< edge weight
```
seunghwak, line 229: Why unique_ptr instead of rmm::device_buffer?

First, if you are storing more than one weights, this should be std:::unique_ptr<WT[]>. And if this memory is allocated on GPU through cudaMalloc, custom deleter should appear in the declaration. AFAIK, std::make_unique (which is exception safe) does not work with a custom deleter. So, if you need to allocate using cudaMalloc and create a unique_ptr object and this leaves a hole in exception safety. And this escapes our rmm based memory management policy.

So, is there really a reason to go for a unique_ptr considering all these complexities?

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCOO: public ConstructedGraphBase<VT, ET, WT> {
+public:
+  std::unique_ptr<VT> src_indices{};   ///< rowInd
+  std::unique_ptr<VT> dst_indices{};   ///< colInd
```
seunghwak, line 255: Again, why unique ptr instead of rmm::device_buffer?

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+ *
+ * @tparam VT   Type of vertex id
+ * @tparam ET   Type of edge id
+ * @tparam WT   Type of weight
+ */
+template <typename VT, typename ET, typename WT>
+class ConstructedGraphCompressedSparseBase: public ConstructedGraphBase<VT,ET,WT> {
+public:
+  std::unique_ptr<VT> offsets{};   ///< CSR offsets
+  std::unique_ptr<VT> indices{};   ///< CSR indices
```
seunghwak, line 295: Again, why unique ptr instead of rmm::device_buffer?

cpp/include/graph.hpp
```diff
@@ -214,5 +214,182 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+   */
+  ConstructedGraphCSC(std::unique_ptr<ET> &&offsets_,
+                      std::unique_ptr<VT> &&indices_,
+                      std::unique_ptr<WT> &&edge_data_,
+                      VT number_of_vertices_,
+                      ET number_of_edges_):
+    ConstructedGraphCompressedSparseBase<VT,ET,WT>(offsets_, indices_, edge_data_, number_of_vertices_, number_of_edges_)
+  {}
+};
+
```
seunghwak, line 391: And why are we creating this new class? Can we just return existing graph class objects?

cpp/include/graph.hpp
ChuckHastings: Sure, I'm fine with having a discussion and making a decision.

cpp/include/graph.hpp
ChuckHastings: Agreed

cpp/include/graph.hpp
ChuckHastings: @kaatish has created a separate PR #799 that I think supersedes this idea.  He took this idea and has expanded it based upon a bunch of discussion on using rmm:device_buffer.  I should probably delete this PR, but was leaving it in place so I could look back at it while comparing to his.

cpp/include/graph.hpp
ChuckHastings: Probably worth following up in #799

cpp/include/graph.hpp
kaatish: Sure, I would prefer VT, etc. We will have to type these a lot and I would like to type less.

cpp/include/graph.hpp
kaatish: I would vote for variables ending with underscore to be private variables.

cpp/include/graph.hpp
seunghwak: > Sure, I would prefer VT, etc. We will have to type these a lot and I would like to type less.

The number one goal should be code readability. In large long running projects, people spend more time on reading code than writing, so I think "typing less" is not as important as code readability.

ChuckHastings: Agree about the goal of readability.

I try to use short variable names and template type names only in short blocks of code where you won't lose the context of what the variable means (e.g. short loops, index variables in a lambda, templates where the type is truly generic).

cpp/include/graph.hpp
seunghwak: I agree with Aatish, that sounds more standard than adding trailing _ to function input parameters.

ChuckHastings:
Superseded by #799

PR #798, opened 2020-04-02
[REVIEW] Edit type of returned graph

Iroy30:
--> Change return Graph type to input Graph type in kcore, ktruss and subgraph extraction
--> Fix number of edges in ktruss wrapper 
--> Add tests


python/cugraph/traversal/bfs.py
```diff
@@ -26,10 +26,6 @@ def bfs(G, start, directed=True):
         as an adjacency list.
     start : Integer
         The index of the graph vertex from which the traversal begins
-    directed : bool
```
afender, line 29: Only the doc was removed but the API still has the option

Iroy30: ah, looks like i forgot to revert that when I moved the bfs fix to a different PR.

python/cugraph/tests/test_k_truss_subgraph.py
```diff
@@ -47,22 +47,27 @@ def ktruss_ground_truth(graph_file):
     return df
 
 
-def cugraph_k_truss_subgraph(graph_file, k):
+def cugraph_k_truss_subgraph(graph_file, k, directed):
     cu_M = utils.read_csv_file(graph_file)
-    G = cugraph.DiGraph()
+    if directed:
```
afender, line 52: I was confused about this at first since the doc says `cugraph_k_truss_subgraph` gets a `cuGraph.Graph` and returns a `cuGraph.Graph`. Can we add a comment explaining why the `directed` parameter and `DiGraph` get involved here? I think this would apply to other algos as well.

python/cugraph/cores/ktruss_subgraph_wrapper.pyx
```diff
@@ -38,7 +38,7 @@ def ktruss_subgraph_double(input_graph, k, use_weights, subgraph_truss):
         input_graph.view_edge_list()
 
     num_verts = input_graph.number_of_vertices()
-    num_edges = input_graph.number_of_edges()
+    num_edges = len(input_graph.edgelist.edgelist_df)
```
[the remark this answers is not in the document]
Iroy30: After recent modifications to follow networkx more closely, the function number_of_edges() now returns number of undirected(unsymmetrized) edges for Graph() type. But the algorithm passes in symmetrized input data whose num_edges parameter should be number of symmetrized edges which we get from len(input_graph.edgelist.edgelist_df)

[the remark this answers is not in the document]
BradReesWork: I'll approve, but we should talk about the function

PR #799, opened 2020-04-02
[REVIEW] Refactored Graph Class with RAII

kaatish:
~In this proposal I am adding a typed wrapper around rmm device buffer which is movable and returns a pointer that can be used internally by algorithms.~

~The other addition is a returnable graph object that owns data abstracted by this typed wrapper.~

A returnable graph object that wraps rmm device_buffer objects. The pointers to these buffers are returned by the objects. They also have a release function so that the internal contents can be moved by the python layer.


cpp/include/graph.hpp
```diff
@@ -14,10 +14,39 @@
  * limitations under the License.
  */
 #pragma once
+#include <rmm/device_buffer.hpp>
 
 namespace cugraph {
 namespace experimental {
 
+template <typename T>
+class cugraph_vector {
```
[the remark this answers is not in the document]
kaatish: I want cugraph_vector to wrap a device_buffer because right now there is a clear way to move the buffer to a column at the python level. I am not sure how an rmm::device_vector can be moved to the column.

And the naming might give the wrong idea but my main aim is to wrap the device_buffer and not provide a full blown vector class and its features.

The wrapper in my opinion should be
- movable
- has the ability to extract a buffer from it
- does not bother the analytic programmer from dealing with rmm device_buffer. Just get pointer and length and be done with it.

cpp/include/graph.hpp
```diff
@@ -208,11 +237,201 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+   * @param  source_indices        This array of size E (number of edges) contains the index of the source for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  destination_indices   This array of size E (number of edges) contains the index of the destination for each edge.
+   *                               Indices must be in the range [0, V-1].
+   * @param  edge_data             This array size E (number of edges) contains the weight for each edge.  This array can be null
+   *                               in which case the graph is considered unweighted.
+   * @param  number_of_vertices    The number of vertices in the graph
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  GraphCOO(cugraph_vector<VT> &&src_indices,
```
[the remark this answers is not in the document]
kaatish: The whole point of introducing cugraph_vector is to wrap a device_buffer to be passed to the python side. Internally we can just used device_vector and work on it.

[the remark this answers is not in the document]
Iroy30: Isn't rmm::device_buffer a class with data pointer and size within?

ChuckHastings:
One more thought to ponder...   Do we even need a "return graph" type?

For k-core, ktruss and subgraph (the three examples I know of), we're going to return a COO.  A COO could be returned across the cython layer as 2 rmm::device_buffers and a number of edges.  Do we feel there is enough justification to create a more complex answer?

cpp/include/graph.hpp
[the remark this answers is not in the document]
seunghwak: Shouldn't something like this work?

```
class CSRGraph {
public:
  edge_t* get_offsets() const {
    return reinterpret_cast<edge_t const*>(offsets.data());
  }
private:
  rmm::device_bufffer offsets{};
}
```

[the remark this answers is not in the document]
kaatish: @seunghwak Yeah. That's what i'll go for in the next push to this PR.

kaatish:
> One more thought to ponder... Do we even need a "return graph" type?
> 
> For k-core, ktruss and subgraph (the three examples I know of), we're going to return a COO. A COO could be returned across the cython layer as 2 rmm::device_buffers and a number of edges. Do we feel there is enough justification to create a more complex data structure just to return 1 thing?

We also need to return a CSR to the python layer when calling get_adj_list()

ChuckHastings:
> > One more thought to ponder... Do we even need a "return graph" type?
> > For k-core, ktruss and subgraph (the three examples I know of), we're going to return a COO. A COO could be returned across the cython layer as 2 rmm::device_buffers and a number of edges. Do we feel there is enough justification to create a more complex data structure just to return 1 thing?
> 
> We also need to return a CSR to the python layer when calling get_adj_list()

Conversion and renumbering have to be addressed as well.  Coo2csr could return 2 rmm::device_buffers (offsets and indices) and a number of vertices.

So, to rephrase my question more completely...

Passing a graph to an algorithm is required for every graph algorithm.  Creating a graph data structure for passing that makes sense, since it simplifies the coding for every graph algorithm.  The graph classes we use for passing the graph to the algorithms can be a view of existing data.

Returning graph data structures occurs in the following situation:
 * coo2csr takes COO format as input and returns CSR as output
 * renumber takes a COO format as input and returns a COO format as output
 * a small number (currently k-core, truss and subgraph) of algorithms take a COO or CSR as input and generate a COO format as output

Is it better for us to create a companion set of "owning" graph classes, as this PR does, or should we simply return the raw building blocks and let the caller (currently the python layer) own the data?

Specifically:
 * coo2csr could take a GraphCOO as input and return an rmm::device_buffer for offsets, an rmm::device_buffer for indices and the number of vertices
 * renumber already takes and returns raw arrays - although perhaps it should take a GraphCOO as input and return an rmm::device_buffer for each of the renumbered arrays and the numbering map
 * algorithms that return a COO could return an rmm::device buffer for src and an rmm::device_buffer for dest

I'm just not sure that the extra cost of coding and maintaining a return graph class is justified for this small number of use cases.  The alternative (returning the rmm::device_buffer instances) is not *that* bad.

kaatish:
>  * coo2csr could take a GraphCOO as input and return an rmm::device_buffer for offsets, an rmm::device_buffer for indices and the number of vertices

What happens when the input graph is weighted? We would need to have two versions of coo2csr. One case to return three buffers for weighted and the other case for unweighted. That goes for any analytic that needs to deal with both weighted and unweighted (kcore, subgraph extraction, truss). I think this adds to the complexity, number of functions and maintenance.

We could pass a flag as well to specify that there is no weight and always pass three buffers. That could remove maintainability from the c++ side but you would just be moving all that logic to the python side. If CSR is being returned then the first device_buffer is of type ET, and if COO is being returned its of type VT. This is not obvious by the returned struct and the programmer will have to look at the documentation of the c++ function to be sure exactly what is being returned. This could lead to mistakes and bugs.

Not abstracting away this complexity also makes it difficult to increase the number of edge attributes or having strings as weights in the future. Abstractions give us more flexibility to adapt to changes.

Adding a returnable graph class has the same motivation as adding the normal graph class. After all, we could remove the graph class and directly pass the pointers to the source, destination, edge weights etc. It helps to reduce complexity by hiding it.

cpp/include/graph.hpp
```diff
@@ -240,11 +237,188 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+    offsets_(sizeof(ET)*(number_of_vertices + 1)),
+    indices_(sizeof(VT)*number_of_edges),
+    edge_data_(has_data? sizeof(WT)*number_of_edges : 0)
   {}
+
+  ET* offsets(void) { return static_cast<ET*>(offsets_.data()); }
+  VT* indices(void) { return static_cast<VT*>(indices_.data()); }
+  WT* edge_data(void) { return static_cast<WT*>(edge_data_.data()); }
+
+  struct contents {
```
Iroy30, line 344: I would assume the algorithms will be returning these contents conatining rmm device_buffers when they have to return a Graph. So in my current understanding an algorithm workflow could possibly look like this:

        Algorithm_xyz(py_Graph, py_subgraph):
            py_Graph.view_adj_list()     // view_adj_list() defined as a graph functionality [1]
            // Populating input Cpp Graph from py_graph adjlist data. Same as before... [2]
            num_verts = py_graph.number_of_vertices()
            num_edges = len(py_indices)
            cdef uintptr_t c_offsets = py_offsets.__cuda_array_interface__['data'][0]
            cdef uintptr_t c_indices = py_indices.__cuda_array_interface__['data'][0]
            cdef uintptr_t c_weights = <uintptr_t>NULL
            graph_float = GraphCSRView[int,int,float](<int*>c_offsets, <int*>c_indices, <float*>c_weights, num_verts, num_edges)
            // Algorithm returns a subgraph
            std::unique_ptr<GraphCOO> c_algorithm_xyz(graph_float)
            // Put subgraph data into py_subgraph [2]
            get_**contents**_from_GraphCOO     // similar to what cudf employs for their column 
            subgraph_offsets = DeviceBuffer.c_from_unique_ptr(contents.offsets)
            ....
            py_subgraph.adjlist.offsets = cudf.Series(Buffer(subgraph_offsets), dtype=np.int32)
            ....
            return py_subgraph

        view_adj_list(self):  // Graph function called in [1]
            Populate GraphCOOView from python Graph  // Similar to [2]
            call cpp function for adding adjlist that returns a GraphCSR
            Populate python Graph from GraphCSR // Same as [3]

kaatish: Yes, the classes have a function `contents release() noexcept;` which need to be called by the python layer. As you wrote in the example, this will get you access to the underlying buffers.

cpp/include/graph.hpp
```diff
@@ -45,23 +67,16 @@ enum class DegreeDirection {
  * @tparam WT   Type of weight
  */
 template <typename VT, typename ET, typename WT>
-class GraphBase {
+class GraphViewBase {
```
seunghwak, line 70: What's the purpose of this View class? (compare to creating a const reference to a non-view class?)

kaatish: I have just renamed GraphBase class to denote that it is a view. The class is already in place in cugraph.

cpp/include/graph.hpp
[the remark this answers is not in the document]
kaatish: I haven't had the time to address this PR but cugraph_vector is to be removed.

cpp/include/graph.hpp
seunghwak: I am not sure I am following this. We have both GraphCOO, GraphCSR, GraphCSC AND GraphCOOView, GraphCSRView, GraphCSCView.

e.g.
https://github.com/rapidsai/cugraph/pull/799/files#diff-ba59ce02957abb53c35f1a1929fc0374R189
AND
https://github.com/rapidsai/cugraph/pull/799/files#diff-ba59ce02957abb53c35f1a1929fc0374R362

What are the purposes of non-view and view classes?

ChuckHastings: The current model in the python code is that the python graph object owns a set of cudf columns that contain the COO representation of the graph, and possible a CSR and/or CSC version as well.

In the current code base, we have a GraphCOO, GraphCSR and GraphCSC that contain the device pointers to these arrays so that the C++ graph algorithms can access the data and execute the algorithms.

@kaatish is proposing renaming these classes to be a view of the device pointers (following the cudf design pattern, these three objects do not own the data, they simply have a view of this data to execute some algorithm).

For most algorithms this is sufficient.  However, there are a handful of algorithms that take a read-only graph as input (this view), but also return a graph as output.  This is the focus of what he's trying to grapple with here.

The non-view classes would be passed ownership of the data before being returned to the python layer.  The python layer would then claim ownership of the data and put it into a new python graph object.

kaatish: I did not understand the scope of your question before @seunghwak. So the main issue is that RMM_ALLOC and RMM_FREE will be deprecated.

In order to return a graph created at the C++, we currently allocate pointers and let the python layer take ownership. This won't be possible very soon due to changes in RMM ([issue](https://github.com/rapidsai/cugraph/issues/727)).

We tried to think of a few workarounds so that ALLOC_TRY usage would work transparently but it looks like it won't be possible. The only option according to me is to return a bunch of rmm::device_buffers in the graph class that python can take ownership of. That is why we need two graph classes. One to be a view, and the other to be an owning class that can enable moving of resources.

seunghwak: Thanks for the clarification, and this makes sense.

One concern is that how we will handle more complex graph structures (e.g.Hornet style dynamic graphs) if python owns the memory; but we can revisit this once this actually happens.

python/cugraph/cores/k_core_wrapper.pyx
```diff
@@ -28,83 +28,105 @@ from libc.float cimport FLT_MAX_EXP
...
-        
+    return df
+
 
+def k_core_double(input_graph, k, core_number):
+    c_src, c_dst, c_weights = graph_params(input_graph)
+    c_vertex, c_values = core_number_params(core_number)
+
+    num_verts = input_graph.number_of_vertices()
+    num_edges = input_graph.number_of_edges()
```
Iroy30, line 103: num_edges = len(input_graph.edgelist.edgelist_df)

python/cugraph/cores/k_core_wrapper.pyx
```diff
@@ -28,83 +28,105 @@ from libc.float cimport FLT_MAX_EXP
...
-    cdef GraphCOO[int,int,float] in_graph_float
-    cdef GraphCOO[int,int,float] out_graph_float
-    cdef GraphCOO[int,int,double] in_graph_double
-    cdef GraphCOO[int,int,double] out_graph_double
+def k_core_float(input_graph, k, core_number):
+    c_src, c_dst, c_weights = graph_params(input_graph)
+    c_vertex, c_values = core_number_params(core_number)
+
+    num_verts = input_graph.number_of_vertices()
+    num_edges = input_graph.number_of_edges()
```
Iroy30, line 74: num_edges = len(input_graph.edgelist.edgelist_df)

python/cugraph/cores/k_core_wrapper.pyx
```diff
@@ -28,83 +28,101 @@ from libc.float cimport FLT_MAX_EXP
...
+    wgt = DeviceBuffer.c_from_unique_ptr(move(contents.edge_data))
+    src = Buffer(src)
+    dst = Buffer(dst)
 
     df = cudf.DataFrame()
+    df['src'] = cudf.core.column.build_column(data=src, dtype="int32")
+    df['dst'] = cudf.core.column.build_column(data=dst, dtype="int32")
+    if weight_type(input_graph) == np.float32:
+        wgt = Buffer(wgt)
+        df['weight'] = cudf.core.column.build_column(data=wgt, dtype="float32")
```
Iroy30, line 91: Maybe df['src'] = cudf.Series(src, dtype="float32") could be more intuitive.

GPUtester:

cpp/src/converters/COOtoCSR.cuh
```diff
@@ -222,3 +225,121 @@ void ConvertCOOtoCSR_weighted(T const * sources, T const * destinations, W const
...
+  auto iter = thrust::make_reverse_iterator(offsets + number_of_vertices);
+  thrust::inclusive_scan(rmm::exec_policy(stream)->on(stream),
+      iter, iter + number_of_vertices + 1, iter, thrust::minimum<ET>());
+  return offsets_buffer;
+}
+
+} //namespace detail
+
+template <typename VT, typename ET, typename WT>
+std::unique_ptr<GraphCSR<VT ,ET ,WT>> coo_to_csr(GraphCOOView<VT, ET, WT> const &graph) {
```
ChuckHastings, line 321: So is the intention that this implementation would replace ConvertCOOtoCSR and ConvertCOOtoCSR_Weighted in all places in the code?

python/cugraph/cores/k_core_wrapper.pyx
```diff
@@ -28,83 +28,101 @@ from libc.float cimport FLT_MAX_EXP
...
 
+
+def k_core_double(input_graph, k, core_number):
+    c_src, c_dst, c_weights, num_verts, num_edges = graph_params(input_graph)
+    c_vertex, c_values = core_number_params(core_number)
+
+    cdef GraphCOOView[int,int,double] in_graph
+    in_graph = GraphCOOView[int,int,double](<int*>c_src, <int*>c_dst, <double*>c_weights, num_verts, num_edges)
+    cdef unique_ptr[GraphCOO[int,int,double]] out_graph = move(c_k_core[int,int,double](in_graph, k, <int*>c_vertex, <int*>c_values, len(core_number)))
+    cdef GraphCOOContents[int,int,double] contents = move(out_graph.get()[0].release())
```
ChuckHastings, line 103: This seems like it will be a common idiom for things that return a graph.

Would it make sense to add a function to graph_new_wrapper.pyx that takes: `cdef GraphCOO[*]` and returns a cudf data frame (or a triple of cudf.Series objects: src, dst, weight)?  That way k-core, subgraph extraction, subgraph matching (once gunrock gets it working), and any other functions that generate a graph can call it?

Might also wants one that takes GraphCSR[*] and returns cudf.Series for offsets, indices, weights.

I don't know a way to "template" them, so we might need to create all variations.

cpp/src/converters/COOtoCSR.cuh
kaatish: Yes, as far as I can tell these are the only two functions that actually allocate memory for csr information. That's why it made sense to me to tackle this here along with the graph objects.

cpp/include/graph.hpp
```diff
@@ -240,10 +245,291 @@ class GraphCSC: public GraphCompressedSparseBase<VT,ET,WT> {
...
+   * @param  number_of_edges       The number of edges in the graph
+   */
+  GraphCOO(VT number_of_vertices,
+           ET number_of_edges,
+           bool has_data = false):
+    number_of_vertices_(number_of_vertices),
+    number_of_edges_(number_of_edges),
+    src_indices_(sizeof(VT)*number_of_edges),
+    dst_indices_(sizeof(VT)*number_of_edges),
+    edge_data_(has_data? sizeof(WT)*number_of_edges : 0)
```
BradReesWork, line 314: Edge data will just be an empty array?

python/cugraph/cores/k_core_wrapper.pyx
BradReesWork: It does seem like the results of function that return a graph (GraphCSR or GraphCOO) that a user would then not be able to pass that graph into another algorithm

ChuckHastings: You can't pass the GraphCSR or GraphCOO into another algorithm.  You would first convert the resulting GraphCSR/GraphCOO into a python graph object (python object takes ownership of the pointers).  That's what this code is doing.  Technically the wrapper is creating a Dataframe with the GraphCOO in it.  The k_core.py takes this data frame and creates a python graph object from it.  The python graph object could be used to call another algorithm.

cpp/include/graph.hpp
kaatish: Yes, if the graph does not have any weights associated with it, then the edge_data buffer is effectively not allocated.

seunghwak: In https://github.com/rapidsai/cugraph/pull/799/files#diff-ba59ce02957abb53c35f1a1929fc0374R248,
trailing _ is used for constructor input parameters, and here, trailing _ is used for class member variables. This can be confusing and I think we'd better be consistent. And I think a more common approach across RAPIDS is to use preceding _ for class member variables (raft does this https://github.com/rapidsai/raft/pull/3/files#diff-245ee3425e4d2e84ecf153f459ec06efR154 and I think we'd better follow raft's convention).

kaatish: @seunghwak I am perfectly fine altering this. I was apprehensive keeping underscore because of this quote.

"Don't overlegislate naming, but do use a consistent naming convention: There are only two must-dos: a) never use "underhanded names," ones that begin with an underscore or that contain a double underscore;" (p2 , C++ Coding Standards, Herb Sutter and Andrei Alexandrescu)

seunghwak: That is interesting, and did they just say not use use a name "begin" with an underscore? (but OK with a name end with an underscore, or disapprove both?)

Google C++ coding style guide asks to use a trailing underscore for class member variables (https://google.github.io/styleguide/cppguide.html#General_Naming_Rules), and I am curious whether a trailing underscore is OK but a preceding one is not, or Goolge and Herb & Andrei have different opinions.

seunghwak: I found

https://www.drdobbs.com/c-coding-standards/184401863
"name private member variables likeThis_;"

and it seems like Herb & Andrei support using trailing _ but does not like preceding _. Do you know why? If there is a persuasive reason, we should better go for trailing _, but I can't think of any reason that trailing _ is better than preceding _.

kaatish: > I found
> 
> https://www.drdobbs.com/c-coding-standards/184401863
> "name private member variables likeThis_;"
> 
> and it seems like Herb & Andrei support using trailing _ but does not like preceding _. Do you know why? If there is a persuasive reason, we should better go for trailing _, but I can't think of any reason that trailing _ is better than preceding _.

[This](https://stackoverflow.com/a/20496955) could be a reason why.

seunghwak: Thanks for the explanation!!! Then, we should consistently use trailing _, I will try to enforce the same for raft as well.

kaatish: I would have liked to make the same changes to the view classes but I thought it is out of scope for the current PR. Would like to hear your and @ChuckHastings 's thoughts.

ChuckHastings: I think we need to settle on a few C++ coding standards (probably beyond just naming, perhaps we start with naming, white space vice tabs, indentation rules) and agree on them as a team.  Then I think we should come up with a strategy to update the code base to that agreed upon standard.  Perhaps something like:
* all new code will be developed to those standards
* code substantially changed in a PR will be adapted to these standards - but not code that is tangentially modified in a PR

Then we can get comfortable with our standards.  At some point, perhaps over a couple of releases, we could create issues to address updating older code to the new standards if it hasn't already been addressed in PRs based on the above bullets.

It might be best to coordinate this with the automatic code reformatting that has been proposed for a while but not yet implemented.

Toward this, I would consider the edits to the view classes tangential.  While I wouldn't oppose making the edits in this type of PR, I would not feel that they were necessary.

seunghwak: > Toward this, I would consider the edits to the view classes tangential. While I wouldn't oppose making the edits in this type of PR, I would not feel that they were necessary.

Cleary they are not critical, but I just wanted to move in a direction to make our codebase more consistent than the opposite. If it takes too much work and we are under pressure, we can leave those as technical debt, but if we have room, I think it is a good idea to address what we can address now rather than pushing everything as future work (even if this can cause some additional work in the future).

kaatish:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

rlratzel:
rerun tests

BradReesWork:
rerun tests

PR #802, opened 2020-04-03
[REVIEW] Remove gdf_column from db

jwyles:
This PR removes the use of gdf_column from the db code. It also re-enables find_matches tests which were previously failing in CI to see if they still fail now.


BradReesWork:
rerun tests

BradReesWork:
@jwyles Test_FindMatches.fifthTest is failing

Issue #778, opened 2020-03-20
[ENH] BFS - Test


xcadet:
**Describe the solution you'd like**
- [ ] C++ tests for BFS
- [ ] Tests should include predecessors validation for BFS in C++/ Python (BFS is tested in SSSP tests)

PR #764, opened 2020-03-17
[REVIEW] update sssp and bfs with new graph class

xcadet:
Remove the Graph and gdf_column from sssp and bfs methods. Use the new GraphCSR object instead.

PR #817, opened 2020-04-07
[REVIEW] Add Vertex Betweenness Centrality

xcadet:
Replace previous BC PR #731 due to changes from 0.14 and #764 updates on BFS / SSSP.

implementation parameter allows to switch between calling the default version or gunrock.
k parameter also accepts a list of vertices to use for an approximation.
endpoints and weight parameters are still not supported

Also includes code for BFS that could be used in #778


GPUtester:

xcadet:
rerun tests

python/cugraph/traversal/bfs.py
```diff
@@ -15,7 +15,7 @@
 from cugraph.structure.graph import Graph
 
 
-def bfs(G, start):
+def bfs(G, start, return_sp_counter=False):
```
xcadet, line 18: Should we split it into 2 functions?
`bfs(G, start)`: never return the sp_counter
`bfs_spc(G, start)`: that always return the sp_counter

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -75,16 +106,52 @@ def betweenness_centrality(G, k=None, normalized=True,
...
+    if endpoints is not False:
+        raise NotImplementedError("endpoints accumulation for betweenness "
+                                  "centrality not currently supported")
 
     if weight is not None:
-        raise Exception("weighted implementation of betweenness "
-                        "centrality not currently supported")
+        raise NotImplementedError("weighted implementation of betweenness "
+                                  "centrality not currently supported")
+    if result_dtype not in [np.float32, np.float64]:
```
xcadet, line 147: Is there a better way or different types that should be accepted?

BradReesWork:
rerun tests

cpp/src/traversal/bfs.cu
```diff
@@ -468,13 +497,22 @@ namespace detail {
     //In that case, distances is a working data
     if (directed && !computeDistances)
       ALLOC_FREE_TRY(distances, nullptr);
+
+    // In that  case, previous_visited_bmap has been allocated
+    if (sp_counters) {
+      ALLOC_FREE_TRY(previous_visited_bmap, nullptr);
```
afender, line 503: Aren't we getting of ALLOC_TRY / ALLOC_FREE_TRY?  
Need to reconcile with @kaatish on status/plan regarding RMM changes for BC/BFS feature

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
     G : cuGraph.Graph
         cuGraph graph descriptor with connectivity information. The graph can
         be either directed (DiGraph) or undirected (Graph)
```
afender, line 32: We should mention that weights stored in the graph are ignored

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -11,11 +11,15 @@
...
 from cugraph.centrality import betweenness_centrality_wrapper
 
 
+# NOTE: result_type=float could ne an intuitive way to indicate the result type
 def betweenness_centrality(G, k=None, normalized=True,
-                           weight=None, endpoints=False, seed=None):
+                           weight=None, endpoints=False, implementation=None,
+                           seed=None, result_dtype=np.float32):
     """
     Compute the betweenness centrality for all nodes of the graph G. cuGraph
```
afender, line 24: This could be misinterpreted. I'd recommend :
```suggestion
    Compute the betweenness centrality for all nodes of the graph G from a sample of 'k' sources. cuGraph 
```

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
     G : cuGraph.Graph
         cuGraph graph descriptor with connectivity information. The graph can
         be either directed (DiGraph) or undirected (Graph)
-    k : int, optional
-        Default is None.
+
+    k : int or list or None, optional, default=None
         If k is not None, use k node samples to estimate betweenness.  Higher
         values give better approximation
+        If k is a list, use the content of the list for estimation
```
afender, line 37: How is this list content used? We should specify that it contains **sources** to run sample traversals from.

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
...
+
     normalized : bool, optional
         Default is True.
         If true, the betweenness values are normalized by
```
afender, line 41: When one would want to set normalized to true/false?

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
...
         2/((n-1)(n-2)) for Graphs (undirected), and
         1 / ((n-1)(n-2)) for DiGraphs (directed graphs)
         where n is the number of nodes in G.
-    weight : cudf.Series
+
+    weight : cudf.Series, optional, default=None
         Specifies the weights to be used for each vertex.
```
afender, line 47: 1) Isn't it for each edge instead of vertices? 
2) I think this should be a table containing  ID - weight pairs because of renumbering the code can't assume implicit matching/ordering

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
...
-    endpoints : bool, optional
+
+    endpoints : bool, optional, default=False
         If true, include the endpoints in the shortest path counts
+
+    implementation : string, optional, default=None
+        if implementation is None or "default", uses native cugraph,
```
afender, line 53: If we need to keep both, we should advise on when to use one or the other and set a default (it seems the code does but the doc says something different).

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
...
+        if "gunrock" uses gunrock based bc
+
     seed : optional
-        k is specified and seed is not None, use seed to initialize the random
-        number generator
+        if k is specified and seed is not None, use seed to initialize the
```
afender, line 57: if k is specified -> what happens when k is a list with more than one element?

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -26,29 +30,44 @@ def betweenness_centrality(G, k=None, normalized=True,
...
+        random number generator
+
+    result_dtype : np.float32 or np.float64, optional, default=np.float32
```
afender, line 60: I understand bc scores can be very big. Should this be np.float64 by default?

python/cugraph/centrality/betweenness_centrality.py
afender: What if we just return doubles? bc value often get over FLT MAX, correct?

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -58,21 +73,57 @@ def betweenness_centrality(input_graph, normalized, endpoints, weight, k, vertic
...
-    c_betweenness_centrality[int,int,float,float](graph, <float*> c_betweenness, normalized, endpoints, <float*> c_weight, c_k, <int*>c_vertices)
-
-    graph.get_vertex_identifiers(<int*>c_identifier)
-
+    # NOTE: The current implementation only has <int, int, float, float> and
+    #       <int, int, double, double> as explicit template declaration
+    #       The current BFS requires the GraphCSR to be declared
+    #       as <int, int, float> or <int, int double> even if weights is null
+    if result_dtype == np.float32:
+        graph_float = GraphCSR[int, int, float](<int*> c_offsets, <int*> c_indices,
```
afender, line 90: watch out for #799

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -30,82 +32,448 @@
...
+                 df['betweenness_centrality'].to_array())}
+    return cu_bc, nx_bc
+
+
+# =============================================================================
+# Utils
+# =============================================================================
+def prepare_rmm(managed_memory, pool_allocator):
+    gc.collect()
     rmm.reinitialize(
```
afender, line 208: This won't fly with #849.
Removing some combinations will help to reduce the runtime of all the BC tests

python/cugraph/tests/test_bfs.py
```diff
@@ -13,100 +13,266 @@
...
+# =============================================================================
+# Parameters
+# =============================================================================
+RMM_MANAGED_MEMORY_OPTIONS = [False, True]
+RMM_POOL_ALLOCATOR_OPTIONS = [False, True]
 
-    G = cugraph.DiGraph()
-    G.from_cudf_edgelist(cu_M, source='0', destination='1',
-                         edge_attr='2')
+DIRECTED_GRAPH_OPTIONS = [True]
```
afender, line 41: Most of these graphs are actually undirected. I think it would make more sense to store the undirected versions and let cuGraph load a `Graph` instead of storing them as directed graphs and loading them in `DiGraph`. This is not related to this PR in particular but something we can address separately for all the test suite. @Iroy30, thoughts?

python/cugraph/traversal/bfs.py
afender: I am fine with the current, single-function version. People can still use it exactly like before if they want to. Just need to confirm that perf is not impacted in this case.

cpp/src/centrality/betweenness_centrality.cuh
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
```
afender, line 2: 2019-2020

python/cugraph/centrality/betweenness_centrality.py
xcadet: The Betweenness Centrality values themselves do not seem to get over FLT MAX and the GAP benchmark expects the result as:
> |V|-sized array of 32-bit floating point numbers
That is why I considered both float32 and float64 for the result type

cpp/src/centrality/betweenness_centrality.cuh
```diff
@@ -0,0 +1,75 @@
...
+      VT number_of_sources;            // Number of vertices in sources
+
+      // --- Output ----
+      // betweenness is set/read by users - using Vectors
+      result_t *betweenness = nullptr;
+
+      // --- Data required to perform computation ----
+      VT *distances = nullptr;         // array<VT>(|V|) stores the distances gathered by the latest SSSP
+      VT *predecessors = nullptr;      // array<WT>(|V|) stores the predecessors of the latest SSSP
+      double *sp_counters = nullptr;   // array<VT>(|V|) stores the shortest path counter for the latest SSSP
```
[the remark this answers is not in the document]
xcadet: The `sp_counters` is used to gather the number of shortest path in each vertices. It could be reused by other algorithms that rely on shortest path counting, I wanted to use `sp` for `shortest path`, the BC specific values are `betweenness` to store the result itself and `deltas` for the dependencies in the accumulation phase.

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,261 @@
...
+    thrust::fill(rmm::exec_policy(stream)->on(stream), betweenness,
+                betweenness + number_of_vertices, static_cast<result_t>(0));
+    cudaStreamSynchronize(stream);
+    if (sources) {
+      for (VT source_idx = 0; source_idx < number_of_sources; ++source_idx) {
+        VT source_vertex = sources[source_idx];
+        compute_single_source(source_vertex);
+      }
+    } else { // Otherwise process every vertices
+      // TODO: Maybe we could still use number of sources and set it to number_of_vertices?
```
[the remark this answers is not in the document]
xcadet: I thought about using TODO to keep some ideas and potential modifications that could help to improve some aspects, while FIXME is more about paths that are not implemented or that would require more attention / highlight some potential issues for later.
Should I change all the TODOs into FIXMEs instead?

python/cugraph/tests/test_betweenness_centrality.py
xcadet: Updated in bc82d30, `prepare_rmm()` is replaced by `prepare_test()` that only calls `gc.collect()` for now.
Tests are no longer parametrized with rmm parameters

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,261 @@
 
 #include <gunrock/gunrock.h>
```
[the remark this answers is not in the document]
xcadet: Using all the sources gives a precise value but when using the `k` and `vertices` parameters we can get an approximation of the centrality much faster, using only some of the vertices to obtain an approximation did not seem to be a possibility using the current public interface from gunrock and only exposes float datatype. Speed-wise the gunrock implementation may run faster when all the sources are used for computing the Betweenness Centrality.  However, the current implementation should cover the same options as gunrock's version, thus we could deprecate

python/cugraph/centrality/betweenness_centrality.py
xcadet: BC score themselves can reach high values (roughly  |V|^2) but its extremely unlikely (this value should be achieved only when one node is crossed by every single shortest path).

cpp/src/centrality/betweenness_centrality.cuh
xcadet: Updated in 50b27a1

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 31ea259

cpp/include/algorithms.hpp
```diff
@@ -394,9 +401,11 @@ void sssp(experimental::GraphCSR<VT,ET,WT> const &graph,
  *
  * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a CSR
  *
- * @param[out] distances            If set to a valid column, this is populated by distance of every vertex in the graph from the starting vertex
+ * @param[out] distances             If set to a valid column, this is populated by distance of every vertex in the graph from the starting vertex
```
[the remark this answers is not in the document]
xcadet: Updated in 50b27a1

cpp/include/algorithms.hpp
```diff
@@ -163,28 +163,34 @@ void overlap_list(experimental::GraphCSR<VT,ET,WT> const &graph,
...
  *
  * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity information as a CSR
  * @param[out] result                Device array of centrality scores
  * @param[in] normalized             If true, return normalized scores, if false return unnormalized scores.
  * @param[in] endpoints              If true, include endpoints of paths in score, if false do not
  * @param[in] weight                 If specified, device array of weights for each edge
- * @param[in] k                      If specified, number of vertex samples defined in the vertices array
- * @param[in] vertices               If specified, device array of sampled vertex ids to estimate betweenness centrality.
+ * @param[in] k                      If specified, number of vertex samples defined in the vertices array .
+ * @param[in] vertices               If specified, host array of vertex ids to estimate betweenness centrality.
```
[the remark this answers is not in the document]
xcadet: Updated in 50b27a1

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,261 @@
...
 
+#include "betweenness_centrality.cuh"
+
+#ifndef MAXBLOCKS
+ #define MAXBLOCKS 65535 // This value is also in traversal_common.cuh
```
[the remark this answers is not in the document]
xcadet: Updated in 50b27a1, values are obtained once, at initialization of the algorithm,  added a FIXME for future RAFT handle

python/cugraph/centrality/betweenness_centrality.py
xcadet: Updated in 50b27a1, also added information about the fact that it is not supported

python/cugraph/centrality/betweenness_centrality.py
xcadet: Updated in 50b27a1

python/cugraph/centrality/betweenness_centrality.py
xcadet: Added in 50b27a1, indicated that is should contain the vertices identifiers.

python/cugraph/centrality/betweenness_centrality.py
xcadet: Added in 50b27a1

python/cugraph/centrality/betweenness_centrality.py
xcadet: 1. Corrected in 50b27a1
2. How should the user define such table?
Having this as  a boolean option, True: use the weights from the Graph, False: Ignore the weights from the Graph would ensure that the weights are renumbered but the API but would give less flexibility.
If users can define the weights manually how can they get the proper ID - Value mapping when the graph is renumbered? How should we handle the case of Un-directed graphs, as the  edges would get 'duplicated'?

python/cugraph/centrality/betweenness_centrality.py
xcadet: Updated in 50b27a1

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,287 @@
...
+  // --- Bind betweenness output vector to internal ---
+  betweenness       = _betweenness;
+  normalized        = _normalized;
+  endpoints         = _endpoints;
+  sources           = _sources;
+  number_of_sources = _number_of_sources;
+  edge_weights_ptr  = _weights;
+
+  // --- Working data allocation ---
+  ALLOC_TRY(&distances, number_of_vertices * sizeof(VT), nullptr);
```
kaatish, line 61: You should add the following device_vectors as private members of class `BC` in `betweenness_centrality.cuh`
```
rmm::device_vector<VT> distance_vec;
rmm::device_vector<VT> predecessor_vec;
rmm::device_vector<double> counter_vec;
rmm::device_vector<double> delta_vec;
```
In this configure function you should do the folllowing :

```
distance_vec.resize(number_of_vertices);
predecessor_vec.resize(number_of_vertices);
counter_vec.resize(number_of_vertices);
delta_vec.resize(number_of_vertices);
distances = distance_vec.data().get();
predecessors = predecessor_vec.data().get();
sp_counters = counter_vec.data().get();
deltas = delta_vec.data().get();
```

This will elimiate `ALLOC_TRY` and `ALLOC_FREE_TRY` from your code with minimal changes. Making the device_vectors private members ensures that they are alive for the lifetime of the object of class BC and are only destroyed along with BC. So make sure that the pointers you get from the allocation (`obj.data().get()`) are only used internally in the class.

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,287 @@
...
+  CUDA_TRY(cudaDeviceGetAttribute(&max_grid_dim_1D, cudaDevAttrMaxGridDimX, device_id));
+  CUDA_TRY(cudaDeviceGetAttribute(&max_block_dim_1D, cudaDevAttrMaxBlockDimX, device_id));
+
+  // --- Confirm that configuration went through ---
+  configured = true;
+}
+template <typename VT, typename ET, typename WT, typename result_t>
+void BC<VT, ET, WT, result_t>::clean()
+{
+  ALLOC_FREE_TRY(distances, nullptr);
```
kaatish, line 77: Once you remove `ALLOC_TRY` and get these pointers from the associated vectors, these `ALLOC_FREE_TRY` calls can be removed.

cpp/src/centrality/betweenness_centrality.cu
kaatish: Every memory allocation that is being done to temporarily hold a value to be used internally by the code should be done via device_vectors or device_buffers this way.

cpp/src/centrality/betweenness_centrality.cu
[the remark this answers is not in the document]
afender: @BradReesWork what do you think, should we keep both or replace?

cpp/src/centrality/betweenness_centrality.cu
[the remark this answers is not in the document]
afender: that's reasonable

python/cugraph/centrality/betweenness_centrality.py
afender: > Having this as a boolean option, True: use the weights from the Graph, False: Ignore the weights from the Graph would ensure that the weights are renumbered but the API but would give less flexibility.

That's reasonable I think although different than Nx.

> If users can define the weights manually how can they get the proper ID - Value mapping when the graph is renumbered? How should we handle the case of Un-directed graphs, as the edges would get 'duplicated'?

They provide their own IDs and we renumber before passing. See Weighted Jaccard for instance

python/cugraph/centrality/betweenness_centrality.py
afender: Ok, so fp32 is always good enough in practice?

python/cugraph/traversal/bfs.py
xcadet: Based on the value of return_sp_counter the C++ is provided with a valid or a null pointer. Based on the value of the pointer shortest path counting specific actions are enable(valid) or disabled(nullptr). Performance should not be affected leaving the return_sp_counter as a parameter of the python function.

xcadet:
rerun tests

cpp/src/centrality/betweenness_centrality.cuh
```diff
@@ -0,0 +1,88 @@
...
+  // --- These information are extracted on setup ---
+  VT number_of_vertices;  // Number of vertices in the graph
+  VT number_of_edges;     // Number of edges in the graph
+  ET const *offsets_ptr;  // Pointer to the offsets
+  VT const *indices_ptr;  // Pointers to the indices
+
+  // --- Information from configuration ---
+  bool configured = false;  // Flag to ensure configuration was called
+  bool normalized = false;  // If True normalize the betweenness
+  // TODO: For weighted version
```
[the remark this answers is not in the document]
xcadet: Updated in 4bdf822.

cpp/include/algorithms.hpp
```diff
@@ -448,12 +458,15 @@ void sssp(experimental::GraphCSRView<VT, ET, WT> const &graph,
  * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity
  * information as a CSR
  *
- * @param[out] distances            If set to a valid column, this is populated by distance of every
- * vertex in the graph from the starting vertex
+ * @param[out] distances             If set to a valid poiner, this is populated by distance of
```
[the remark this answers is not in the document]
xcadet: Updated in 4bdf822.

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in 4bdf822, removed the `clean` function from the BC class.

cpp/src/centrality/betweenness_centrality.cu
xcadet: Changed the BC class accordingly in 4bdf822.

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -26,10 +26,286 @@
...
+    accumulation_kernel<VT, ET, WT, result_t><<<grid, block, 0, stream>>>(betweenness,
+                                                                          number_of_vertices,
+                                                                          graph.indices,
+                                                                          graph.offsets,
+                                                                          distances,
+                                                                          sp_counters,
+                                                                          deltas,
+                                                                          source,
+                                                                          depth);
+    cudaDeviceSynchronize();
```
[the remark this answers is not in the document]
xcadet: Removed the synchronizations and reran tests in 4bdf822.

python/cugraph/centrality/betweenness_centrality.py
xcadet: Moved the default value to float64 in 4bdf822 as it is less restrictive and kept the option to return float32

python/cugraph/centrality/betweenness_centrality.py
xcadet: Modified in 4bdf822.

python/cugraph/centrality/betweenness_centrality.py
xcadet: Moved the default value to float64 in 4bdf822 as it is less restrictive and kept the option to return float32, it also avoids some casts between the output result type and `sp_counter` and `deltas`  as they are already in double

cpp/src/traversal/bfs.cu
xcadet: Updated in d5b9a32, BC now also uses rmm::device_vector, the clean function still exists as there is cub related allocation

xcadet:
rerun tests

seunghwak:
> rerun tests

Unrelated to this feature,

but I think

https://github.com/rapidsai/cugraph/blob/f6242e77bf32ed12c78ddb3f9a06321b2fd11806/cpp/include/graph.hpp#L33
should be
```
 bool multigraph{false};
```
and could you piggy-back this one line change to this PR?

python/cugraph/tests/test_bfs.py
afender: https://github.com/rapidsai/cugraph/issues/872

cpp/src/centrality/betweenness_centrality.cu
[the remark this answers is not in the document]
afender: > want to avoid all device synchronization calls. @afender thoughts?

Exactly, these are expensive and can be avoided most of the time.

xcadet:
rerun tests

python/cugraph/centrality/betweenness_centrality.py
xcadet: Udpated the documentation for later, using an cudf.DataFrame so that the users can provide their own IDs.

cpp/src/centrality/betweenness_centrality.cu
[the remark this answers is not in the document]
BradReesWork: Our version should be the default, but having the Gunrock option is fine.  If/when the guncrock code breaks, then we can deprecate that feature

PR #818, opened 2020-04-08
[REVIEW] Initial version of new "benchmarks" folder

rlratzel:
I'm leaving the current benchmarks in `python/utils` in place until everyone has migrated over and we're happy with the new version.

The new version here replaces `run_benchmarks.sh` and `run_benchmarks.py` in `python/utils` and is about a quarter of the size, and much more readable and maintainable IMO.

The new benchmarks use `pytest` with the `pytest-benchmark` plugin to take advantage of the popular and efficient patterns used for writing and running tests, including fixtures to reuse common steps for setup and teardown, markers to easily group benchmarks into categories ("fast", "slow", "ETL", etc.), the pytest discovery mechanisms allowing a dev to just add a new benchmark and have it automatically included in the suite, reporting plugins, etc.

`pytest-benchmark` provides several benchmarking conveniences as well, such as cache warmup options, run averaging, history saving and comparisons to past runs, plotting, and more [here](https://pytest-benchmark.readthedocs.io/en/latest/index.html) .

NOTE: This cusignal PR adds the same benchmarking approach: https://github.com/rapidsai/cusignal/pull/43

cc @mnicely @dillon-cullinan @mike-wendt 

## Still to do in this PR
- [x] Extend `benchmark` fixture to add GPU metrics
- [x] Add additional fixtures for ETL steps so they aren't included in the metrics (pre-compute adjacency lists, renumber)
- [x] Use `Graph` and `DiGraph` appropriately for different datasets
- [x] re-enable commented out benchmarks once some of the above steps are done

## Requests for comments
* I intentionally did not add subdirs under `benchmarks` at this time, but we can eventually add subdirs for C++ benches (if we ever add those) or other types of organization.  Subdirs for categories like "e2e", "micro", etc. could be added, but those might be better handled by simply separating those by module and using pytest markers IMO
* We'll eventually need to mark specific benchmarks for use on a per-CI (continuous benchmarking) basis vs. longer-running E2E nightly, since we cannot run the full suite as-is for a PR.
  * Once the above is done, do we want to have only "fast" benches run as part of CI (as was done in https://github.com/rapidsai/cusignal/pull/43)?
* @BradReesWork : I'm also working on the GAP benchmark using this framework. Should that benchmark be included in this directory?


GPUtester:

benchmarks/bench_algos.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Returns a cuDF DataFrame containing the columns read in from
+    csvFileName. Optional delim string defaults to ' ' (space) for CSV reading.
+    """
+    cols = ["src", "dst", "val"]
+    dtypes = OrderedDict([
+            ("src", "int32"),
+            ("dst", "int32"),
+            ("val", "float32"),
+            ])
```
BradReesWork, line 36: you can just define a list of types. ["int32",int32","float32"]

benchmarks/bench_algos.py
```diff
@@ -0,0 +1,169 @@
...
+    gpubenchmark = pytest_benchmark.plugin.benchmark
+    def setFixtureParamNames(*args, **kwargs):
+        pass
+
+###############################################################################
+# Utilities
+#
+# FIXME: move utilities to a more reusable location/module
+def getEdgelistFromCsv(csvFileName, delim=' '):
+    """
```
BradReesWork, line 27: there is a similar function under python/cugraph/tests/utils.py

benchmarks/bench_algos.py
```diff
@@ -0,0 +1,169 @@
...
+    symmetrized can be set to True to perform those operation on construction.
+    """
+    if createDiGraph:
+        G = cugraph.DiGraph()
+    else:
+        G = cugraph.Graph(symmetrized=symmetrized)
+    G.from_cudf_edgelist(edgelistGdf, source="src",
+                         destination="dst", edge_attr="val",
+                         renumber=renumber)
+    return G
```
BradReesWork, line 64: not sure the utility of this function.  Within you if block, just do a cugraph.from_cudf_edgelist( and set the create_using flag

rlratzel:
rerun tests

benchmarks/bench_algos.py
```diff
@@ -0,0 +1,207 @@
...
+    gpubenchmark = pytest_benchmark.plugin.benchmark
+
+    def setFixtureParamNames(*args, **kwargs):
+        pass
+
+
+# FIXME: write and use mechanism described here for specifying datasets:
+#        https://docs.rapids.ai/maintainers/datasets
+# FIXME: rlr: soc-twitter-2010.csv crashes with OOM error on my HP-Z8!
+UNDIRECTED_DATASETS = [
```
afender, line 26: How the csv files were generated? The mtx2csv script creates directed edges in both directions for undirected graphs (which is really a DiGraph). We should make sure that undirected data is stored there (ie. if we have `u v` there's no `v u` explicitly stored there). 
This will matter for reading time and symmetrize time.

benchmarks/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Prerequisites
+
+* `pytest` and the `pytest-benchmark` plugin
+* cugraph built and installed (or `cugraph` sources and built C++ extensions available on `PYTHONPATH`)
+* *< show dependencies needed for GPU metrics (should be just pynvml) >*
+* *< show optional dependencies to install for plotting >*
+
+## Usage
+
```
afender, line 20: We should document how to quickly run:
- a specific set of graphs
- a specific set of algo

These reqs will keep coming.

benchmarks/bench_algos.py
```diff
@@ -0,0 +1,207 @@
+import pytest
+
+import cugraph
+from cugraph.tests import utils
+
+import pytest_benchmark
+# FIXME: Remove this when rapids_pytest_benchmark.gpubenchmark is available
+# everywhere
```
afender, line 8: What would be a good path to optionally enable managed memory?

benchmarks/README.md
rlratzel: good idea. I need to update the README for this anyway so I'll resolve this in the next commit to this PR.

benchmarks/bench_algos.py
rlratzel: I'm currently using datasets from the `/datasets/afender` mount on the lab machines, and using CSV files that were already in place. I can look at the undirected CSVs to make sure both directions are present.

benchmarks/bench_algos.py
rlratzel: `pytest-benchmark` lets benchmark authors add their own command-line options, which I'm not doing here.  I can add one for managed memory and have a fixture look for it accordingly though. Do you think that's a good way to go?

rlratzel:
> 

> LGTM. Would be nice to precompute adjist/transposedadjlists before running algorithms.

Agreed.  I have the fixtures in place to do that but just haven't used them yet.  I'll look into doing that though possibly in the next commit to this PR.

benchmarks/bench_algos.py
rlratzel: I added new markers for fine-grained control of how RMM reinit happens, and a new command-line option (`--no-rmm-reinit`) to disable it completely.

PR #820, opened 2020-04-09
[REVIEW] OPG infra and all-gather smoke test

afender:
- get nccl from conda (with the option of using a local one)
- cmake and setup.py find/include/link nccl. 
- find/include/link MPI
- All gather test

Use `-DBUILD_MPI=ON` to build and link OPG features.

This stack is compatible with the OPG CSR Pagernak proto
In a few weeks, we may want to replace MPI by UCX (after we can access the comms in RAFT). This current stack allows adding and running what we already have in the meantime.

Screen Shot 2020-04-08 at 4 15 41 PM


ci/gpu/build.sh
```diff
@@ -66,6 +66,7 @@ conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaul
       distributed>=2.12.0 \
       dask-cudf=${MINOR_VERSION} \
       dask-cuda=${MINOR_VERSION} \
+      nccl>=2.5 \
```
afender, line 69: @rlratzel anything else we should do in `ci/` to enable OPG testing?

cpp/CMakeLists.txt
```diff
@@ -104,6 +104,13 @@ set(CMAKE_EXE_LINKER_FLAGS "-Wl,--disable-new-dtags")
 option(BUILD_TESTS "Configure CMake to build tests"
        ON)
 
+option(BUILD_MPI "Build with MPI" OFF)
+if (BUILD_MPI)
+    find_package(MPI REQUIRED)
```
afender, line 109: MPI must be installed if this flag is `ON`

ci/gpu/build.sh
rlratzel: I'm not very familiar with what an OPG build needs, but does MPI need to be enabled too?  If so, will you need to set the `BUILD_MPI` cmake flag?  If that's the case, you may also want to update the `build.sh` in the root of cugraph so users (and scripts like this one) can build from source correctly.

cpp/tests/test_utils.h
```diff
@@ -62,6 +62,23 @@ extern "C" {
 }
 #endif
 
+#define NCCLCHECK(cmd) do {                                   \
+    ncclResult_t nccl_status = cmd;                           \
+    if (nccl_status!= ncclSuccess) {                          \
+      printf("NCCL failure %s:%d '%s'\n",                     \
```
rlratzel, line 68: Should this fail the test?  It looks like it's just going to print a message to stdout.

cpp/tests/test_utils.h
```diff
@@ -62,6 +62,23 @@ extern "C" {
...
+          __FILE__,__LINE__,ncclGetErrorString(nccl_status)); \
+    }                                                 \
+  } while(0)
```
rlratzel, line 71: Can you explain (preferably in a code comment for others) why the 1-iteration do-while loop is needed? Is this to force a new scope?

cpp/tests/test_utils.h
```diff
@@ -62,6 +62,23 @@ extern "C" {
...
+
+#define MPICHECK(cmd) do {                          \
```
rlratzel, line 73: Minor suggestion about the macro name: `NCCLCHECK` appears to just print messages, but `MPICHECK` calls `exit()`. Maybe this should be called `ENSURE_MPI_SUCCESS` or something more self-documenting, or if you switch it to use a gtest assertion: `ASSERT_MPI_SUCCESS`?

ci/gpu/build.sh
afender: MPI can be enabled as long as the machine have :
- MPI installed
-  multiple GPUs 
- call OPG gtests while spanning multiple processes

Considering these requirements this cannot be set as a default path for users in `build.sh` as of know.

My question was related to CI machines and capabilities to support this OPG testing from DevOps perspective though.

rlratzel: > My question was related to CI machines and capabilities to support this OPG testing from DevOps perspective though.

@raydouglass do you know the answer to this question?

cpp/tests/test_utils.h
```diff
@@ -62,6 +62,23 @@ extern "C" {
...
+  int e = cmd;                                      \
+  if( e != MPI_SUCCESS ) {                          \
+    printf("Failed: MPI error %s:%d '%d'\n",        \
+        __FILE__,__LINE__, e);   \
+    exit(EXIT_FAILURE);                             \
```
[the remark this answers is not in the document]
afender: > Will this cause the entire gtest binary to exit out, possibly before it gets to run any other tests?

Yes

> Maybe this should be a gtest assertion instead, so any test using this macro will just fail with a proper failure message instead of exiting the entire run (assuming that's what will happen).

MPI errors associated with MPI_COMM_WORLD are fatal by default so this macro shouldn't have much impact here. 
Notice that this is just a specific test macro and that the stand-alone test is just checking that the infra can access MPI/NCCL. 
As we add OPG code inside cugraph (ie. as a feature) we should have an OPG error handling. This should be as part of the RAFT communicator abstraction classes. Eventually, we will test OPG cuGraph features using the proper communicator and proper error handling.

cpp/tests/test_utils.h
afender: yes

cpp/tests/test_utils.h
afender: like the printf in the comment above, these are artifacts from debugging the prototype, I need to update it for gtest.

afender: I think the original source for these macros was NCCL doc https://docs.nvidia.com/deeplearning/sdk/nccl-developer-guide/docs/examples.html

cpp/tests/test_utils.h
afender: Reconciled naming and behavior 0dbc162

ci/gpu/build.sh
raydouglass: Multi GPU testing is not available in gpuCI and I don't think MPI is installed. So sounds like you should not set OPG testing as default.

PR #823, opened 2020-04-13
[REVIEW] Remove gdf column from nvgraph

### LINKED ISSUE #572: [BUG] Subgraph Extraction - needs to return renumbering mapping
**Describe the bug**
When running cugraph.subgraph(G, V) the returned list of vertices are renumbered.  While that is needed to reduce the size of the extracted matrix, a mapping of old vertex IDs to renumbered needs to be returned.   

**What should be returned**
  subG, df
    - the initialized subgraph
    - a dataframe with vertex ID mapping

### LINKED ISSUE #653: [BUG] Remove gdf_column in nvgraph based features
Next step for #538 

Apply #629 to all nvgraph based features (listed below). It makes more sense to do them all at once as they are often grouped together in the code

Features involved : 
- balancedCutClustering_nvgraph 
- spectralModularityMaximization_nvgraph 
- analyzeClustering_modularity_nvgraph 
- analyzeClustering_edge_cut_nvgraph
- analyzeClustering_ratio_cut_nvgraph
- extract_subgraph_vertex_nvgraph

The API header `nvgraph_gdf.h` should get rid of cuDF and should be renamed `nvgraph_imported.h`.

ChuckHastings:
This PR refactors the nvgraph code.

closes #653
closes #572

The primary motivation is to remove the gdf_column from nvgraph, but this PR will also address eliminating parts of nvgraph code that we no longer use and migrating nvgraph code toward using RAII constructs for memory allocation.

This PR needs to be adjusted based on #799.  If PR 799 is merged first, ECG.cu and extract_subgraph_by_vertex.cu need to be updated to reflect the final approach to handling returning graphs.  If this PR is merged first then PR 799 should be updated after merging this to make appropriate changes to those two files.

Changed the unit test for #572 and updated the code to return the vertices with the original numbering.


GPUtester:

cpp/src/community/extract_subgraph_by_vertex.cu
```diff
@@ -0,0 +1,167 @@
...
+
+#include <graph.hpp>
+#include <algorithms.hpp>
+
+#include <rmm/thrust_rmm_allocator.h>
+
+#include <utilities/error_utils.h>
+#include <utilities/cuda_utils.cuh>
+
+// FIXME: Update with new graph return object
```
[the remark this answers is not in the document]
afender: @kaatish do you have an ETA on when you think #499 will be ready?

cpp/CMakeLists.txt
```diff
@@ -342,11 +342,13 @@ add_library(cugraph SHARED
     src/traversal/sssp.cu
     src/link_prediction/jaccard.cu
     src/link_prediction/overlap.cu
-    src/converters/nvgraph.cu
     src/converters/renumber.cu
     src/converters/COOtoCSR.cu
-    src/community/nvgraph_gdf.cu
+    src/community/nvgraph_clustering.cu
```
afender, line 347: `spectral_clustering` would work better than `nvgraph_clustering` I think

cpp/CMakeLists.txt
```diff
@@ -342,11 +342,13 @@ add_library(cugraph SHARED
...
+    src/community/louvain.cu
     src/community/ECG.cu
+    src/community/triangles_counting.cu
```
afender, line 350: I noticed some files were moved out of `nvgraph` directory but others (most of the spectral clustering backend and some helpers) are still there. What are the selection criteria? Should we rename or remove the directory?

cpp/src/nvgraph/include/nvgraph_cusparse.hxx
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019, NVIDIA CORPORATION.
+ * Copyright (c) 2019-2020, NVIDIA CORPORATION.
```
afender, line 2: We have 3 cuSparse wrappers... This needs to be reconciled with the wrapper in `cpp/src/utils` and the one in RAFT. It doesn't have to be in this PR though, I'll create a separate issue.

cpp/src/nvgraph/matrix.cu
```diff
@@ -216,22 +216,6 @@ namespace nvgraph {
...
-      nnz(G.get_num_edges()),
-      descrA(_descrA), 
-      csrValA(G.get_raw_values()),
-      csrRowPtrA(G.get_raw_row_offsets()),
-      csrColIndA(G.get_raw_column_indices()) {
-    Cusparse::set_pointer_mode_host();
-  }
-
   /// Destructor for CSR matrix class
   template <typename IndexType_, typename ValueType_>
```
afender, line 220: Afaik this is really just the matrix class used for the spectral clustering feature. We could add a `spectral_` prefix to the file name for now and try to see if it can go away in the long run.

python/cugraph/community/ecg_wrapper.pyx
```diff
@@ -34,39 +31,52 @@ def ecg(input_graph, min_weight=.05, ensemble_size=16):
...
+    [offsets, indices] = graph_new_wrapper.datatype_cast([input_graph.adjlist.offsets,
                                                       input_graph.adjlist.indices], [np.int32, np.int64])
-    [weights] = graph_wrapper.datatype_cast([input_graph.adjlist.weights], [np.float32, np.float64])
-
-    cdef uintptr_t graph = graph_wrapper.allocate_cpp_graph()
-    cdef Graph * g = <Graph*> graph
+    [weights] = graph_new_wrapper.datatype_cast([input_graph.adjlist.weights], [np.float32, np.float64])
 
-    graph_wrapper.add_adj_list(graph, offsets, indices, weights)
+    print("offsets = ", offsets.values)
```
afender, line 44: Looks like debug prints

python/cugraph/community/subgraph_extraction_wrapper.pyx
```diff
@@ -32,52 +29,95 @@ import numpy as np
...
-    if input_graph.renumbered is True:
+    if not input_graph.edgelist:
+        input_graph.view_edge_list()
+
+    [src, dst] = graph_new_wrapper.datatype_cast([input_graph.edgelist.edgelist_df['src'], input_graph.edgelist.edgelist_df['dst']], [np.int32])
+
+    if input_graph.edgelist.weights:
+        [weights] = graph_new_wrapper.datatype_cast([input_graph.edgelist.edgelist_df['weights']], [np.float32, np.float64])
+        if weights.dtype == np.float64:
+            use_float = False
```
afender, line 48: If weights happen to be integers or fp16 in the future, this will break. The following change would make things future proof

```
if weights.dtype == np.float64:
  ...
elif weights.dtype == np.float32:
  ...
else
    throw error "unsupported weight type"
```

python/cugraph/community/subgraph_extraction_wrapper.pyx
```diff
@@ -32,52 +29,95 @@ import numpy as np
...
+                                        dtype=np.int32)
+        df['src'] = cudf.Series(tmp)
+
+        tmp = rmm.device_array_from_ptr(<uintptr_t>out_graph_float.dst_indices,
+                                        nelem=out_graph_float.number_of_edges,
+                                        dtype=np.int32)
+
+        df['dst'] = cudf.Series(tmp)
+        if weights is not None:
+            tmp = rmm.device_array_from_ptr(<uintptr_t>out_graph_float.edge_data,
```
afender, line 92: I remember there were discussions on addressing https://github.com/rapidsai/cugraph/issues/572 along with other subgraph extraction improvements.
I can't find it in the PR. Am I missing something (if so this PR should close 572) or will this be addressed separately later on?

python/cugraph/community/triangle_count_wrapper.pyx
```diff
@@ -31,25 +30,24 @@ def triangles(input_graph):
...
+    [offsets, indices] = graph_new_wrapper.datatype_cast([input_graph.adjlist.offsets,
+                                                          input_graph.adjlist.indices], [np.int32])
+
+    num_verts = input_graph.number_of_vertices()
+    num_edges = len(indices)
+
+    cdef uintptr_t c_offsets = offsets.__cuda_array_interface__['data'][0]
+    cdef uintptr_t c_indices = indices.__cuda_array_interface__['data'][0]
+
+    cdef GraphCSR[int,int,float] graph
```
afender, line 48: Random general question: 
we don't need weights (or weights types) here. Is there a way we could just declare a `GraphCSR[int,int]` so that we avoid dealing with the dummy `float` type in unweighted analytics wrappers? We could leverage default template parameters at the C++ level so that the graph class implementation could just default to a null float*, but not sure if cython would support that. Thoughts?

cpp/CMakeLists.txt
ChuckHastings: Good point.

cpp/CMakeLists.txt
ChuckHastings: I made several attempts at doing this and settled on this one.  Here's my criteria:

1. The primary entry points were created in the src/community directory (where nvgraph_gdf.cu was)
1. The top level of implement was moved into the src/community directory co-resident with the primary entry point
1. All library-like things and headers were left under nvgraph.

In one of my attempts, I actually started refactoring a bunch of the other components more completely, but I realized two things.  First, it would be a great deal of programming effort.  Second, the new API that we are developing as part of OPG (and the GraphBLAS work) would and up replacing much of this.  What I settled on as a reasonable compromise was to completely refactor things that were in src/community (or that I moved to src/community) and do selective editing of the nvgraph core functionality to make it work with the new graph structure.

We can certainly go further if we like, although I'd be inclined to do that in a separate PR rather than make this even bigger.

cpp/src/nvgraph/matrix.cu
ChuckHastings: I can rename this.  I actually looked into getting rid of it entirely, but that was going to be more work.  I imagine we'll want to rework these spectral techniques to use GraphBLAS once it's ready.

python/cugraph/community/ecg_wrapper.pyx
ChuckHastings: Missed a few :-(

python/cugraph/community/subgraph_extraction_wrapper.pyx
ChuckHastings: The datatype_cast (on line 46) should force the result to be np.float32 if it is not np.float32 or np.float64.  If the weight is integer or fp16, it will create a new cudf.Series object that is np.float32.  I'm not sure what would happen if the type is not convertible to np.float32, I assume that the cudf call that tries to convert the type to np.float32 will throw.

So I think this change would be superfluous.

python/cugraph/community/subgraph_extraction_wrapper.pyx
ChuckHastings: I am going to construct a test for #572 and test it against this.  If it doesn't work I was planning on adding a fix to this PR.  Once I have that test in place I'll add the appropriate annotation.

python/cugraph/community/triangle_count_wrapper.pyx
ChuckHastings: I thought about that when I was constructing the objects.  I don't know cython well enough to know if it would support defaulted parameters.  It seemed an innocuous amount of extra typing to specify float.

The underlying C++ classes allow the pointer to be null and just ignore it.  There's no real penalty other than having to specify a type of something that's unused.

cpp/src/community/extract_subgraph_by_vertex.cu
```diff
@@ -0,0 +1,167 @@
...
+        new_weight_v.resize(count);
+        d_new_weight = new_weight_v.data().get();
+      }
+#endif
+      vertex_t *d_new_src{nullptr};
+      vertex_t *d_new_dst{nullptr};
+      weight_t *d_new_weight{nullptr};
+
+      ALLOC_TRY(&d_new_src, count * sizeof(vertex_t), nullptr);
+      ALLOC_TRY(&d_new_dst, count * sizeof(vertex_t), nullptr);
```
seunghwak, line 95: We need to remove ALLOC_TRY as well, ALLOC_TRY calls RMM_ALLOC.

cpp/tests/community/ecg_test.cu
```diff
@@ -31,42 +30,105 @@ TEST(ecg, success)
...
+
+  cugraph::experimental::GraphCSR<int,int,float> graph_csr(offsets_v.data().get(),
+                                                           indices_v.data().get(),
+                                                           weights_v.data().get(),
+                                                           num_verts,
+                                                           num_edges);
+
+  ASSERT_NO_THROW((cugraph::nvgraph::ecg<int32_t, int32_t, float>(graph_csr, .05, 16, result_v.data().get())));
+
+  cudaMemcpy ((void*) &(cluster_id[0]), result_v.data().get(), sizeof(int)*num_verts, cudaMemcpyDeviceToHost);
```
seunghwak, line 51: cluster_id[0] = result_v[0]; should work. thrust::device_vector and thrust::device_ptr (note that thrust device_ptr is different from a raw pointer) properly handles data transfer between host and device.

cpp/src/community/extract_subgraph_by_vertex.cu
ChuckHastings: My comment above was intended to address this.  This is temporary code until #799 is merged.  At that point the input will be a GraphCOOView and the output will be a GraphCOO, so these will have been allocated as rmm::device_buffer objects.

cpp/tests/community/ecg_test.cu
ChuckHastings: That didn't work, but I changed it to std::copy in the latest commit.

python/cugraph/community/subgraph_extraction_wrapper.pyx
ChuckHastings: Added a fix to the PR down a few lines (after the comment: `renumber vertices to match original input`) TO ADDRESS #572.  Also updated the unit test to remove the [now] unnecessary lookup of ids in the vertex array.

cpp/tests/community/ecg_test.cu
seunghwak: What do you mean by this didn't work.

thrust::device_vector<int> d_ints(100, 0.0);
thrust::host_vector<int> h_ints = d_ints;

Something like the above is what I have used extensively. Could you show the code that didn't work?

ChuckHastings: Ah... so I need to change cluster_id from being std::vector to being thrust::host_vector.  Missed that part of your comment.

cpp/src/community/extract_subgraph_by_vertex.cu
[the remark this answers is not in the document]
BradReesWork: #499 was merged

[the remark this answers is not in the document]
ChuckHastings: I think @afender meant #799

BradReesWork:
rerun tests

ChuckHastings:
Merged in the latest from the branch, including all of the clang format changes and the graph object update from #799 

This is a very large PR, it deletes large amounts of unused code from nvgraph.  More could be done, but I think this is a good stopping point.  All nvgraph functions now use the RAII constructs, and no longer use the gdf_column objects.

I disabled the ECG python tests.  ECG itself works as before.  However, I believe the unit test fails sometimes because the Louvain code is not generating a correct result.  Once Louvain is working we should be able to enable that test again.

BradReesWork:
rerun tests

Issue #699, opened 2020-02-06
[FEA] Hungarian algorithm for bipartite graph


BradReesWork:
https://en.wikipedia.org/wiki/Hungarian_algorithm

CPU implementations for reference and comparison

CPU implementations:
https://github.com/hrldcpr/hungarian
https://github.com/RoboJackets/hungarian
https://github.com/mcximing/hungarian-algorithm-cpp
https://github.com/maandree/hungarian-algorithm-n3
 

- [ ] C/CUDA Code
- [ ] C/C++ API
- [ ] Python API
- [ ] Bipartite Graph Class
- [ ] Test
- [ ] Documentation

PR #819, opened 2020-04-08
[WIP] [skip-ci] Implement hungarian algorithm

ChuckHastings:
Implement the Hungarian algorithm for graph matching of bipartite graphs.

Closes #699

PR #826, opened 2020-04-15
[REVIEW] Bipartite python graph

Iroy30:
Bipartite functionality for python Graph class.

Related to Issue #699


Iroy30:
The following is the networkx pipeline for bipartite graphs:
>>> B = nx.Graph()
>>> B.add_nodes_from([1, 2, 3, 4], bipartite=0)
>>> B.add_nodes_from(['a', 'b', 'c'], bipartite=1)
>>> B.add_edges_from([(1, 'a'), (1, 'b'), (2, 'b'), (2, 'c'), (3, 'c'), (4, 'a')])
>>> bottom_nodes, top_nodes = bipartite.sets(B)
>>> bipartite.bipartite_algo(B, bottom_nodes) 

The same can be accomplished via this PR as:
>>> G = cugraph.Graph()
>>> G.add_nodes_from([1, 2, 3, 4], bipartite=0)
>>> G.add_nodes_from([0, 5, 6, 7], bipartite=1)
>>> G.from_cudf_edgelist(........)
>>> bottom_nodes, top_nodes = G.sets()
>>> cugraph.bipartite_algo(G, bottom_nodes) #Run the Algorithm, like in #819 

Things to Note:
-> sets() currently relies on the bipartite arg in add_nodes_from. It can be eventually replaced by coloring algorithm
-> networkx has all bipartite functions in a separate module accessed by importing "from networkx.algorithms import bipartite" and then calling bipartite.algo(Graph, nodes) or bipartite.sets(Graph). We could follow something similar in cugraph if needed.

ChuckHastings:
It would be nice to have a test that exercises this code.  The code "looks" right to me, but it's hard to be sure without a test.

BradReesWork:
Now sure we need the full Nx syntax in the initial version
- B.add_nodes_from([1, 2, 3, 4], bipartite=0)
- B.add_nodes_from(['a', 'b', 'c'], bipartite=1)
- B.add_edges_from([(1, 'a'), (1, 'b'), (2, 'b'), (2, 'c'), (3, 'c'), (4, 'a')])

For the initial version we need to push back some control to the user for creating a proper bipartite graph.  We can expand with defining the various vertex types in future releases

Iroy30:
> Now sure we need the full Nx syntax in the initial version
> 
> * B.add_nodes_from([1, 2, 3, 4], bipartite=0)
> * B.add_nodes_from(['a', 'b', 'c'], bipartite=1)
> * B.add_edges_from([(1, 'a'), (1, 'b'), (2, 'b'), (2, 'c'), (3, 'c'), (4, 'a')])
> 
> For the initial version we need to push back some control to the user for creating a proper bipartite graph. We can expand with defining the various vertex types in future releases

Following networkx and PR #819, we need to pass to bipartite algorithms the graph itself and a set of nodes in one "partition". i.e. bipartite_algorithm(Graph, set_of_nodes). The Graph is a normal Graph or DiGraph. The set_of_nodes could be entered by the user either directly when they are running the algorithm OR can be initially saved within graph itself using G.add_nodes_from([1, 2, 3, 4], bipartite=0), then extracting this information when required and passing it as the set_of_nodes.   Currently, its entirely in the user's control to provide a bipartite graph and its set of nodes, there is no internal check for the validity( which can be added in future releases)

Iroy30:
> It would be nice to have a test that exercises this code. The code "looks" right to me, but it's hard to be sure without a test.

I have added a small test as well as docs. The test exercises the add_nodes_from and sets function. Since the user is responsible for entering a bipartite graph for now, add_nodes_from allows them to store a set of nodes in one partition and sets() retrieves it.

python/cugraph/tests/test_graph.py
```diff
@@ -793,6 +793,48 @@ def test_has_node(managed, pool, graph_file):
...
+    # added the partition in add_nodes_from()
+    assert G.is_bipartite()
+
+    # Call sets() to get the bipartite set of nodes.
+    set1, set2 = G.sets()
+
+    # assert if the input set1_exp is same as returned bipartite set1
+    assert set1.equals(set1_exp)
+    # assert if set2 is the remaining set of nodes not in set1_exp
+    assert set2.equals(set2_exp)
```
ChuckHastings, line 835: This tests the mechanics of the bipartite class.  I'm comfortable with the API based on this test code.

However, it might be worth noting that neither of the data sets in DATASETS2 (karate and dolphin) are actually bipartite.  This test, while mechanically correct, might be misleading.  Should we find/create a bipartite test data set?

Perhaps we should do the following:
 1. make a note in this test case that we're testing the mechanics of the API and that these data sets are not actually bipartite
 1. create an issue to add some bipartite test data to the test graphs and update this test case once that is available

python/cugraph/structure/graph.py
```diff
@@ -72,7 +72,8 @@ def __init__(self, m_graph=None, edge_attr=None, symmetrized=False,
         """
         self.symmetrized = symmetrized
         self.renumbered = False
-        self.bipartite = bipartite
+        self.bipartite = False
+        self._nodes = {}
```
BradReesWork, line 76: shouldn't nodes be an array of sets?  You need to know which nodes are in sets A and set B.   We should also allow for N-partite

python/cugraph/tests/test_graph.py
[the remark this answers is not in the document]
BradReesWork: This test does not create a bipartite graph since all the edges are within the same node set

python/cugraph/structure/graph.py
ChuckHastings: @Iroy30 can answer more definitively.  I believe that nodes is a map (python dictionary) with the key being the node type and the value being a list of nodes that are of that type.

python/cugraph/structure/graph.py
```diff
@@ -103,6 +104,51 @@ def clear(self):
...
+        bipartite and the entered nodes are saved as the set of nodes in one
+        partition.
+        If bipartite argument is not passed, the nodes are saved as information
+        and Graph is not considered to be bipartite.
+        """
+        set_names = [i for i in self._nodes.keys() if i != 'nodes']
+        if len(set_names) < 2:
+            if bipartite != 'nodes':
+                self.bipartite = True
+            self._nodes[bipartite] = cudf.Series(nodes)
```
ChuckHastings, line 120: Should we allow for appending additional nodes to one of the bipartite sets?  This code would do the following, if I interpret it correctly:
 * If you call it with bipartite = 'A' twice, the second call would overwrite the first call
 * If you call it with bipartite = 'A', then call it with bipartite = 'B', then call it with bipartite = 'A' it would fail because it would think we're adding a third type.

Also, with Brad's comment... might want to be able to support multi-partite, so perhaps checking the length of set_names is unnecessary.  is_bipartite would be true if there are exactly 2 keys in the _nodes map.

python/cugraph/structure/graph.py
```diff
@@ -103,6 +104,51 @@ def clear(self):
         self.adjlist = None
         self.transposedadjlist = None
 
+    def add_nodes_from(self, nodes, bipartite='nodes'):
```
ChuckHastings, line 107: Would it be better to use:  `bipartite=None` as the final parameter.  This introduces a limitation that a user can't name one of their type sets "none".

python/cugraph/tests/test_graph.py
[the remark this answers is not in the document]
ChuckHastings: Line 815 above arbitrarily segregates the vertices into two different node sets.  I think if you were to print out set1 and set2 they each contain half of the vertices.

afender:
@BradReesWork @Iroy30 should this be 0.15?

GPUtester:

python/cugraph/structure/graph.py
Iroy30: Changed to bipartite = None. If this argument is not provided, the nodes are simply assumed to be all the nodes present in the graph.

python/cugraph/tests/test_graph.py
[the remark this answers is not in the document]
Iroy30: Updated the comments to mention that the current testing is to only test the mechanics of the API including initialization, storage and retrieval of a bipartite graph. The datasets are not bipartite by themselves and the test or API does not verify/validate whether the input is indeed bipartite.

Iroy30: Made a note in the comments

python/cugraph/structure/graph.py
Iroy30: Updated the API to allow for multipartite graphs. The nodes are a dictionary mapping partition name to the set of nodes in that partition

python/cugraph/structure/graph.py
Iroy30: Good catch. Updated to fix this discrepancy. Now the code always follows "If you call it with bipartite = 'A' twice, the second call would overwrite the first call"

python/cugraph/structure/graph.py
```diff
@@ -116,6 +118,81 @@ def clear(self):
         self.adjlist = None
         self.transposedadjlist = None
 
+    def add_nodes_from(self, nodes, bipartite=None, multipartite=None):
```
afender, line 121: Are we adding a notebook separately? 

How about documentation on bipartite graphs, when to use them instead of regular cugraph graphs, and what is supported? Something along the lines of https://github.com/rapidsai/cugraph/blob/branch-0.15/docs/source/dask-cugraph.rst?

python/cugraph/structure/graph.py
BradReesWork: I am okay starting with a map, but also think that it will need to change as we flesh out use cases

the most likely case is that the data will start in a cudf dataframe where there is a column for vertex ID and a column for vertex type.  Creating a python dictionary from that data sounds very inefficient

python/cugraph/structure/graph.py
BradReesWork: agree on docs and sample notebook

python/cugraph/structure/graph.py
```diff
@@ -116,6 +118,81 @@ def clear(self):
...
+            The nodes of the graph to be stored. If bipartite and multipartite
+            arguments are not passed, the nodes are considered to be a list of
+            all the nodes present in the Graph.
+        bipartite : str
+            Sets the Graph as bipartite. The nodes are stored as a set of nodes
+            of the partition named as bipartite argument.
+        multipartite : str
+            Sets the Graph as multipartite. The nodes are stored as a set of
+            nodes of the partition named as multipartite argument.
+        """
```
BradReesWork, line 137: I am not sure that bipartite and multipartite need to be separate.  The difference is only partite count == 2 versus a different number

Looking at NetworkX:  B.add_nodes_from([1, 2, 3, 4], bipartite=0)

The add node lines specified which partite layer the nodes related to.  So you can keep track of the number of layer added and set bipartite vs multipartite accordingly.

python/cugraph/structure/graph.py
BradReesWork: If we look at some work we did with customers.  each partite layer had its own numbering scheme, so there can be collisions

python/cugraph/structure/graph.py
Iroy30: We can add the docs within the code as now. As for notebooks i believe we should first add some bipartite datasets to cugraph to run on some algorithms to demonstrate its actual usage

python/cugraph/structure/graph.py
Iroy30: We can. However it might also be confusing to a user to mention the partition with the bipartite argument but enter a npartite graph

python/cugraph/structure/graph.py
Iroy30: If I understand you correctly, on the Dataframe input having a column of vertex IDs and a column of partitionID/name, a cudf groupby could be used to store the map. However, I am not completely clear about the:

> If we look at some work we did with customers. each partite layer had its own numbering scheme, so there can be collisions

Maybe an example usecase would be more helpful in understanding better.

BradReesWork: vertices in set A = [0,1,2,3,4], vertices in set B = [0,1,2,3, 4]  where A != B.  So the vertex with ID 0 in A is not the same as the vertex in B with ID 0

Iroy30: How would the user input of the cudf DataFrame containing the edgelist look like for a Graph() structure in this case.

PR #832, opened 2020-04-25
[REVIEW] Removed uses of RMM ALLOC functions from db

jwyles:
this is related to #727


GPUtester:

afender:
#727

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
   offsets_size = _offsets_size;
-  indirection = _indirection;
+  indirection = std::move(_indirection);
   indirection_size = _indirection_size;
 }
 
 template<typename idx_t>
 idx_t* db_column_index<idx_t>::getOffsets() {
-  return offsets;
+  return (idx_t*) offsets.data();
```
afender, line 178: is this safe? Can always cast offsets to idx_t or should we have some check?

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
-
 template<typename idx_t>
 db_column_index<idx_t>::db_column_index() {
-  offsets = nullptr;
   offsets_size = 0;
-  indirection = nullptr;
   indirection_size = 0;
 }
 
 template<typename idx_t>
```
afender, line 129: Most of the codebase is decoupling `idx` type from `offset` type as the latter is expected to get bigger than 2B sooner. Would that be relevant for DB?

cpp/src/db/db_object.cu
```diff
@@ -216,19 +194,22 @@ idx_t db_column_index<idx_t>::getIndirectionSize() {
...
-std::string db_column_index<idx_t>::toString(){
+std::string db_column_index<idx_t>::toString() {
   std::stringstream ss;
   ss << "db_column_index:\n";
   ss << "Offsets: ";
-  idx_t* hostOffsets = (idx_t*)malloc(sizeof(idx_t) * offsets_size);
-  cudaMemcpy(hostOffsets, offsets, sizeof(idx_t) * offsets_size, cudaMemcpyDefault);
+  idx_t* hostOffsets = (idx_t*) malloc(sizeof(idx_t) * offsets_size);
```
afender, line 201: As per today's discussion, we are trying to get rid of `malloc`

cpp/src/db/db_object.cu
```diff
@@ -345,21 +317,15 @@ db_table<idx_t>::db_table() {
...
-      ALLOC_FREE_TRY(columns[i], nullptr);
-      columns[i] = nullptr;
-    }
-  }
 }
 
 template<typename idx_t>
 void db_table<idx_t>::addColumn(std::string name) {
   if (columns.size() > size_t { 0 } && column_size > 0)
     throw new std::invalid_argument("Can't add a column to a non-empty table");
```
seunghwak, line 325: We should use CUGRAPH_EXPECTS and CUGRAPH_FAIL for consistency.

cpp/src/db/db_object.cu
```diff
@@ -304,9 +277,8 @@ void db_result<idx_t>::allocateColumns(idx_t size) {
   if (dataValid)
     throw new std::invalid_argument("Already allocated columns");
```
seunghwak, line 278: We should use CUGRAPH_EXPECTS and CUGRAPH_FAIL for consistency.

cpp/src/db/db_object.cu
```diff
@@ -378,38 +344,34 @@ void db_table<idx_t>::rebuildIndices() {
...
+    cudaMemcpy(tempColumn.data(), columns[i].data(), sizeof(idx_t) * size, cudaMemcpyDefault);
 
     // Construct an array of ascending integers
-    idx_t* indirection;
-    ALLOC_TRY(&indirection, sizeof(idx_t) * size, nullptr);
-    thrust::sequence(rmm::exec_policy(nullptr)->on(nullptr), indirection, indirection + size);
+    rmm::device_buffer indirection(sizeof(idx_t) * size);
+    thrust::sequence(rmm::exec_policy(nullptr)->on(nullptr),
+                     (idx_t*) indirection.data(),
+                     (idx_t*) indirection.data() + size);
```
seunghwak, line 354: C style casting is used extensively in this PR. Better use C++ style casting (e.g. reinterpret_cast here as this code reinterprets void* to idx_t*). And if you don't need to interact with .cpp or .hpp files, rmm::device_vector is generally a safer/cleaner option than rmm::device_buffer (please ignore this comment if you need to interact with .cpp and .hpp files).

cpp/src/db/db_object.cu
jwyles: Yes, this is safe. The offsets buffer is always treated as idx_t type.

cpp/src/db/db_object.cu
jwyles: DB isn't set up to distinguish between offset type and index type, and I don't really think that it needs to be.

cpp/src/db/db_object.cu
jwyles: The malloc's used here are inside of a debugging-only toString() method. They aren't in code that will get exposed to end users. Brad has said it's okay to leave it as is.

cpp/src/db/db_object.cu
jwyles: Changed to use CUGRAPH_EXPECTS rather than just throwing an exception.

cpp/src/db/db_object.cu
jwyles: Changed to use CUGRAPH_EXPECTS rather than just throwing an exception.

cpp/src/db/db_object.cu
jwyles: I have replaced the C-style casting with C++-style casting. Sticking with rmm::device_buffer as it's cleaner to use (no multiple steps to get a raw pointer out).

cpp/src/db/db_object.cu
seunghwak: My biggest concern is, this will appear in search, and it is very difficult to figure out which malloc is benign and which is not. This will make finding problematic mallocs very difficult as we need to inspect every search appearance of malloc case by case. If this is not too much additional work, it is great if you can avoid this.

seunghwak: thrust::host_vector<idx_t> host_offsets = offsets;
for (idx_t i = 0; i < offsets_size; ++i) {
  ss << host_offsets[i] << " ";
}
works and simpler.

And if you don't care about performance (if this is debugging only),

offsets[i] in the host side works correctly (offsets here is rmm::device_vector which is a wrapper for thrust::device_vector, right?).

cpp/src/db/db_object.cu
seunghwak: OK, but in general, if you are not using old style functions expecting raw pointers, thrust::device_vector.data() returns device_ptr, device_ptr can be safely dereferenced in both host and device, and this often leads to safer & cleaner implementation in many cases. Not a huge deal, but just a recommendation.

cpp/src/db/db_object.cu
seunghwak: I think something like the following will be cleaner.
```
  ss << "Offsets:";
  for (auto it = offsets.cbegin(); it != offsets.cend(); ++it) {
    ss << " " << *it;
  }
  ss << "\n";
  
  ss << "Indirection:";
  for (auto it = indirection.cbegin(); it != indirection.cend(); ++it) {
    ss << " " << *it;
  }
  ss << "\n";
```

cpp/src/db/db_object.cuh
```diff
@@ -20,172 +20,178 @@
...
+
+/**
+ * Class which encapsulates a CSR-style index on a column
+ */
+template<typename idx_t>
+class db_column_index {
+  rmm::device_buffer offsets;
+  rmm::device_buffer indirection;
+  idx_t offsets_size;
+  idx_t indirection_size;
```
seunghwak, line 75:
```suggestion
  idx_t offsets_size{0};
  idx_t indirection_size{0};
```

seunghwak: And if you initialize here, no need to define your own default constructor, destructor, copy & move constructor & operator.

Just mark them to use default one (e.g. db_column_index() = default;) will work and will be much easier to understand.

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
   indirection_size = 0;
 }
 
 template<typename idx_t>
-db_column_index<idx_t>::db_column_index(idx_t* _offsets,
+db_column_index<idx_t>::db_column_index(rmm::device_buffer&& _offsets,
                                         idx_t _offsets_size,
-                                        idx_t* _indirection,
+                                        rmm::device_buffer&& _indirection,
                                         idx_t _indirection_size) {
```
seunghwak, line 133: I recommend using trailing _ for class member variables and avoid using either preceding/trailing _ for function input parameters. There is a good chance that this will go to our coding guidelines.

And using preceding _ in variable names is highly discouraged (see https://stackoverflow.com/questions/3136594/naming-convention-underscore-in-c-and-c-sharp-variables/20496955#20496955).

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
-  }
-}
-
 template<typename idx_t>
 db_column_index<idx_t>::db_column_index() {
-  offsets = nullptr;
   offsets_size = 0;
-  indirection = nullptr;
   indirection_size = 0;
 }
```
seunghwak, line 127: db_column_index() = default; should be sufficient if you initialize offset_size and indirection_size on definition.

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
+  offsets = std::move(other.offsets);
   offsets_size = other.offsets_size;
-  indirection = other.indirection;
+  indirection = std::move(other.indirection);
   indirection_size = other.indirection_size;
-  other.offsets = nullptr;
   other.offsets_size = 0;
-  other.indirection = nullptr;
   other.indirection_size = 0;
 }
```
seunghwak, line 148: db_column_index(db_column_index<idx_t>&& other) = default; should be sufficient.

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
   offsets_size = other.offsets_size;
-  indirection = other.indirection;
+  indirection = std::move(other.indirection);
   indirection_size = other.indirection_size;
-  other.offsets = nullptr;
   other.offsets_size = 0;
-  other.indirection = nullptr;
   other.indirection_size = 0;
   return *this;
 }
```
seunghwak, line 163: Like the above comments, default ones work and no need to add unnecessary code here.

cpp/src/db/db_object.cu
```diff
@@ -120,84 +120,62 @@ bool db_pattern<idx_t>::isAllConstants() {
...
+                                       rmm::device_buffer&& _indirection,
                                        idx_t _indirection_size) {
-  deleteData();
-  offsets = _offsets;
+  offsets = std::move(_offsets);
   offsets_size = _offsets_size;
-  indirection = _indirection;
+  indirection = std::move(_indirection);
   indirection_size = _indirection_size;
 }
```
seunghwak, line 174: Isn't this redundant to the move assignment operator?

cpp/src/db/db_object.cu
seunghwak, line 178:
```suggestion
  return reinterpret_cast<idx_t*>(offsets.data());
```

cpp/src/db/db_object.cu
```diff
@@ -207,7 +185,7 @@ idx_t db_column_index<idx_t>::getOffsetsSize() {
 
 template<typename idx_t>
 idx_t* db_column_index<idx_t>::getIndirection() {
-  return indirection;
+  return (idx_t*) indirection.data();
```
seunghwak, line 188:
```suggestion
  return reinterpret_cast<idx_t*>(indirection.data());
```

cpp/src/db/db_object.cu
```diff
@@ -265,14 +246,6 @@ db_result<idx_t>& db_result<idx_t>::operator =(db_result<idx_t>&& other) {
 
 template<typename idx_t>
 db_result<idx_t>::~db_result() {
-  deleteData();
-}
```
seunghwak, line 269: If you initialize dataValid and columnSize on definition, no need to have your custom default constructor, destructor, copy & move constructor & operators.

cpp/src/db/db_object.cu
```diff
@@ -282,31 +255,28 @@ idx_t db_result<idx_t>::getSize() {
...
-  if (dataValid)
-    throw new std::invalid_argument("Already allocated columns");
+  CUGRAPH_EXPECTS(!dataValid, "Already allocated columns");
+
   for (size_t i = 0; i < names.size(); i++) {
-    idx_t* colPtr = nullptr;
-    ALLOC_TRY(&colPtr, sizeof(idx_t) * size, nullptr);
-    columns.push_back(colPtr);
+    rmm::device_buffer col(sizeof(idx_t) * size);
+    columns.push_back(std::move(col));
```
seunghwak, line 279: columns.push_back(rmm::device_buffer(sizeof(idx_t) * size));

This makes it clear that you're passing an rvalue without using std::move. This also prevents users from accidentally using moved-out variable after this statement.

cpp/src/db/db_object.cuh
```diff
@@ -20,172 +20,178 @@
...
+  bool isAllConstants();
+};
+
+/**
+ * Class which encapsulates a CSR-style index on a column
+ */
+template<typename idx_t>
+class db_column_index {
+  rmm::device_buffer offsets;
+  rmm::device_buffer indirection;
```
seunghwak, line 73: And this is .cuh file so this requires nvcc anyways (if not, this shouldn't be .cuh, this should be .hpp). And if this file is not to interface with the plain C++ code, use rmm::device_vector instead of rmm::device_buffer. This will make this code simpler/safer in multiple places.

cpp/src/db/db_object.cu
```diff
@@ -216,19 +194,22 @@ idx_t db_column_index<idx_t>::getIndirectionSize() {
...
+  idx_t* hostOffsets = (idx_t*) malloc(sizeof(idx_t) * offsets_size);
+  cudaMemcpy(hostOffsets, offsets.data(), sizeof(idx_t) * offsets_size, cudaMemcpyDefault);
   for (idx_t i = 0; i < offsets_size; i++) {
     ss << hostOffsets[i] << " ";
   }
   free(hostOffsets);
   ss << "\nIndirection: ";
-  idx_t* hostIndirection = (idx_t*)malloc(sizeof(idx_t)  * indirection_size);
-  cudaMemcpy(hostIndirection, indirection, sizeof(idx_t) * indirection_size, cudaMemcpyDefault);
+  idx_t* hostIndirection = (idx_t*) malloc(sizeof(idx_t) * indirection_size);
```
seunghwak, line 208: Better avoid using malloc.

cpp/src/db/db_object.cu
```diff
@@ -322,7 +292,7 @@ std::string db_result<idx_t>::toString() {
   std::vector<idx_t*> hostColumns;
   for (size_t i = 0; i < columns.size(); i++) {
     idx_t* hostColumn = (idx_t*) malloc(sizeof(idx_t) * columnSize);
```
seunghwak, line 294: Better avoid using malloc.

cpp/src/db/db_object.cu
```diff
@@ -464,7 +425,7 @@ std::string db_table<idx_t>::toString() {
...
```
seunghwak, line 427: better avoid using malloc.

cpp/tests/db/find_matches_test.cu
```diff
@@ -137,12 +137,11 @@ TEST_F(Test_FindMatches, thirdTest) {
   int32_t* frontier_ptr;
   cudaMalloc(&frontier_ptr, sizeof(int32_t));
```
seunghwak, line 138: Better avoid using cudaMalloc. And if this is absolutely necessary, guard this with CUDA_TRY.

cpp/tests/db/find_matches_test.cu
```diff
@@ -137,12 +137,11 @@ TEST_F(Test_FindMatches, thirdTest) {
...
-  cugraph::detail::gdf_col_set_defaults(frontier);
-  gdf_column_view(frontier, frontier_ptr, nullptr, 1, GDF_INT32);
 
-  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches(q, table, frontier, 0);
+  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches<int32_t>(q, table, frontier_ptr, 1, 0);
 
+
+  cudaFree(frontier_ptr);
   ASSERT_EQ(result.getSize(), 1);
   int32_t* resultA = new int32_t[result.getSize()];
```
seunghwak, line 146: Better avoid using new;

cpp/tests/db/find_matches_test.cu
```diff
@@ -71,7 +71,7 @@ TEST_F(Test_FindMatches, firstTest){
   p.addEntry(p1);
   p.addEntry(p2);
   p.addEntry(p3);
-  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches(p, table, nullptr, 1);
+  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches<int32_t>(p, table, nullptr, 0, 1);
   ASSERT_EQ(result.getSize(), 1);
   int32_t* resultA = new int32_t[result.getSize()];
   int32_t* resultB = new int32_t[result.getSize()];
```
seunghwak, line 77: Better avoid using new.

cpp/src/db/db_object.cu
```diff
@@ -418,33 +380,32 @@ void db_table<idx_t>::flush_input() {
...
-    newColumns.push_back(newCol);
+    rmm::device_buffer newCol(sizeof(idx_t) * newSize);
+    newColumns.push_back(std::move(newCol));
   }
   for (size_t i = 0; i < columns.size(); i++) {
     if (currentSize > 0)
-      cudaMemcpy(newColumns[i], columns[i], sizeof(idx_t) * currentSize, cudaMemcpyDefault);
-    cudaMemcpy(newColumns[i] + currentSize,
-               tempColumns[i],
+      cudaMemcpy(newColumns[i].data(),
```
seunghwak, line 400: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -418,33 +380,32 @@ void db_table<idx_t>::flush_input() {
...
+                 columns[i].data(),
+                 sizeof(idx_t) * currentSize,
+                 cudaMemcpyDefault);
+    cudaMemcpy(reinterpret_cast<idx_t*>(newColumns[i].data()) + currentSize,
```
seunghwak, line 404: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -464,7 +425,7 @@ std::string db_table<idx_t>::toString() {
   std::vector<idx_t*> hostColumns;
   for (size_t i = 0; i < columns.size(); i++) {
     idx_t* hostColumn = (idx_t*) malloc(sizeof(idx_t) * columnSize);
-    cudaMemcpy(hostColumn, columns[i], sizeof(idx_t) * columnSize, cudaMemcpyDefault);
+    cudaMemcpy(hostColumn, columns[i].data(), sizeof(idx_t) * columnSize, cudaMemcpyDefault);
```
seunghwak, line 428: guard with CUDA_TRY

cpp/tests/db/find_matches_test.cu
```diff
@@ -137,12 +137,11 @@ TEST_F(Test_FindMatches, thirdTest) {
...
-  gdf_column_view(frontier, frontier_ptr, nullptr, 1, GDF_INT32);
 
-  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches(q, table, frontier, 0);
+  cugraph::db::db_result<int32_t> result = cugraph::db::findMatches<int32_t>(q, table, frontier_ptr, 1, 0);
 
+
+  cudaFree(frontier_ptr);
   ASSERT_EQ(result.getSize(), 1);
   int32_t* resultA = new int32_t[result.getSize()];
   cudaMemcpy(resultA, result.getData("a"), sizeof(int32_t) * result.getSize(), cudaMemcpyDefault);
```
seunghwak, line 147: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -216,19 +194,22 @@ idx_t db_column_index<idx_t>::getIndirectionSize() {
...
 template<typename idx_t>
-std::string db_column_index<idx_t>::toString(){
+std::string db_column_index<idx_t>::toString() {
   std::stringstream ss;
   ss << "db_column_index:\n";
   ss << "Offsets: ";
-  idx_t* hostOffsets = (idx_t*)malloc(sizeof(idx_t) * offsets_size);
-  cudaMemcpy(hostOffsets, offsets, sizeof(idx_t) * offsets_size, cudaMemcpyDefault);
+  idx_t* hostOffsets = (idx_t*) malloc(sizeof(idx_t) * offsets_size);
+  cudaMemcpy(hostOffsets, offsets.data(), sizeof(idx_t) * offsets_size, cudaMemcpyDefault);
```
seunghwak, line 202: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -216,19 +194,22 @@ idx_t db_column_index<idx_t>::getIndirectionSize() {
...
   }
   free(hostOffsets);
   ss << "\nIndirection: ";
-  idx_t* hostIndirection = (idx_t*)malloc(sizeof(idx_t)  * indirection_size);
-  cudaMemcpy(hostIndirection, indirection, sizeof(idx_t) * indirection_size, cudaMemcpyDefault);
+  idx_t* hostIndirection = (idx_t*) malloc(sizeof(idx_t) * indirection_size);
+  cudaMemcpy(hostIndirection,
+             indirection.data(),
+             sizeof(idx_t) * indirection_size,
+             cudaMemcpyDefault);
```
seunghwak, line 212: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -322,7 +292,7 @@ std::string db_result<idx_t>::toString() {
   std::vector<idx_t*> hostColumns;
   for (size_t i = 0; i < columns.size(); i++) {
     idx_t* hostColumn = (idx_t*) malloc(sizeof(idx_t) * columnSize);
-    cudaMemcpy(hostColumn, columns[i], sizeof(idx_t) * columnSize, cudaMemcpyDefault);
+    cudaMemcpy(hostColumn, columns[i].data(), sizeof(idx_t) * columnSize, cudaMemcpyDefault);
```
seunghwak, line 295: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -378,38 +341,37 @@ void db_table<idx_t>::rebuildIndices() {
   for (size_t i = 0; i < columns.size(); i++) {
     // Copy the column's data to a new array
     idx_t size = column_size;
-    idx_t* tempColumn;
-    ALLOC_TRY(&tempColumn, sizeof(idx_t) * size, nullptr);
-    cudaMemcpy(tempColumn, columns[i], sizeof(idx_t) * size, cudaMemcpyDefault);
+    rmm::device_buffer tempColumn(sizeof(idx_t) * size);
+    cudaMemcpy(tempColumn.data(), columns[i].data(), sizeof(idx_t) * size, cudaMemcpyDefault);
```
seunghwak, line 345: guard with CUDA_TRY

cpp/src/db/db_object.cu
```diff
@@ -378,38 +341,37 @@ void db_table<idx_t>::rebuildIndices() {
...
 
     // Compute offsets array based on sorted column
     idx_t maxId;
-    cudaMemcpy(&maxId, tempColumn + size - 1, sizeof(idx_t), cudaMemcpyDefault);
-    idx_t* offsets;
-    ALLOC_TRY(&offsets, (maxId + 2) * sizeof(idx_t), nullptr);
+    cudaMemcpy(&maxId,
+               reinterpret_cast<idx_t*>(tempColumn.data()) + size - 1,
+               sizeof(idx_t),
+               cudaMemcpyDefault);
```
seunghwak, line 364: guard with CUDA_TRY

Iroy30:
rerun tests

Iroy30:
rerun tests

cpp/src/db/db_object.cu
seunghwak:
```suggestion
    CUDA_TRY(cudaMemcpy(tempColumn.data(), columns[i].data(), sizeof(idx_t) * size, cudaMemcpyDefault));
```

cpp/src/db/db_operators.cu
```diff
@@ -16,405 +16,406 @@
...
+  rmm::device_buffer tempSpace(tempSpaceSize);
+  cub::DeviceSelect::Flagged(tempSpace.data(),
+                             tempSpaceSize,
+                             col_ptr,
+                             reinterpret_cast<int8_t*>(flags.data()),
+                             col_ptr,
+                             reinterpret_cast<idx_t*>(compactSize_d.data()),
+                             output_size);
+  idx_t compactSize_h;
+  cudaMemcpy(&compactSize_h, compactSize_d.data(), sizeof(idx_t), cudaMemcpyDefault);
```
seunghwak, line 381:
```suggestion
  CUDA_TRY(cudaMemcpy(&compactSize_h, compactSize_d.data(), sizeof(idx_t), cudaMemcpyDefault));
```

PR #834, opened 2020-04-27
[REVIEW] Update local gpuci build

efajardo-nv:
Added Docker 19 support

### DIFF
--- a/ci/local/build.sh
+++ b/ci/local/build.sh
@@ -22,7 +22,7 @@ where:
 if [[ -z "${CUDA_VISIBLE_DEVICES}" ]]; then
     NVIDIA_VISIBLE_DEVICES="all"
 else
-    NVIDIA_VISIBLE_DEVICES="device=${CUDA_VISIBLE_DEVICES}"
+    NVIDIA_VISIBLE_DEVICES=${CUDA_VISIBLE_DEVICES}
 fi
 
 while getopts ":hHr:i:s" option; do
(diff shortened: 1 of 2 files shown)


GPUtester:

ci/local/build.sh
```diff
@@ -22,7 +22,7 @@ where:
 if [[ -z "${CUDA_VISIBLE_DEVICES}" ]]; then
     NVIDIA_VISIBLE_DEVICES="all"
 else
-    NVIDIA_VISIBLE_DEVICES=${CUDA_VISIBLE_DEVICES}
+    NVIDIA_VISIBLE_DEVICES="device=${CUDA_VISIBLE_DEVICES}"
```
dillon-cullinan, line 25: I don't think this should be changed. Instead, move the "device=" string down to the GPU_OPTS variable. Otherwise this won't work for Docker < 19.

ci/local/build.sh
```diff
@@ -76,7 +76,7 @@ mkdir -p "${REPO_PATH}/${PYTHON_BUILD_DIR}"
 
 BUILD_SCRIPT="#!/bin/bash
 set -e
-WORKSPACE=${REPO_PATH_IN_CONTAINER}
+export WORKSPACE=${REPO_PATH_IN_CONTAINER}
```
dillon-cullinan, line 79: Was this not working before?

ci/local/build.sh
efajardo-nv: thanks

ci/local/build.sh
efajardo-nv: not working before adding export. invalid path errors during notebook testing because workspace wasn't set.

PR #838, opened 2020-04-28
[REVIEW] Pattern accelerator based implementation of PageRank, Katz Centrality, BFS, & SSSP

seunghwak:
OK, I will try to merge this and plan to address multi-GPU extensions & performance tuning in separate PRs.

This PR is already very large and also there are multiple works dependent on this, so I think this works better (and this code is not linked to any python user code yet, so there isn't much risk in premature merging).

This API aims to achieve

1. thrust-like API for graph algorithms
2. Abstract out implementation issues in different target systems (Single GPU, multi-GPU, ...) inside the pattern accelerator API, Graph, and Handle; Same analytics code will be used for different target systems.
3. Minimize redundancy in cuGraph codebase and better enforce consistency.


GPUtester:

afender:
This PR is for 0.15, correct? The target branch is still 0.14.

seunghwak:
> This PR is for 0.15, correct? The target branch is still 0.14.

Yes, 0.15, will fix sometime soon.

pgera:
Is this meant to replace the existing BFS/SSP implementations in the near future? I was trying to use BFS/SSSP for larger graphs for some of my research, and currently it is specialized only for ints. Even specializing it for unsigned breaks it. I haven't tried int64 or uint64, and ideally the row offset type and col idx type needn't be the same. I was about to dive in to debug it, but if it is going to be deprecated, I can use this PR instead for larger graphs.

seunghwak:
> Is this meant to replace the existing BFS/SSP implementations in the near future? I was trying to use BFS/SSSP for larger graphs for some of my research, and currently it is specialized only for ints. Even specializing it for unsigned breaks it. I haven't tried int64 or uint64, and ideally the row offset type and col idx type needn't be the same. I was about to dive in to debug it, but if it is going to be deprecated, I can use this PR instead for larger graphs.

This will eventually replace existing BFS/SSSP, but it may take several months. This is likely to be the first PR of three related PRs to make this happen. Once we have this PR merged, we will have low-performance just working correctly versions of BFS/SSSP for a single-GPU. Then, the follow-up PR will work on renumbering/data partitioning. The final PR will work on performance optimization/multi-GPU extension. This code will likely to be linked to the python layer only after this code gets optimized.

pgera:
Thanks. I was able to fix some of the type related issues in the existing BFS/SSSP code. I'll send patches after additional testing.

GPUtester:

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+          typename EdgeOp,
+          typename T,
+          typename VertexValueOutputIterator>
+void copy_v_transform_reduce_in_nbr(raft::handle_t const& handle,
+                                    GraphViewType const& graph_view,
+                                    AdjMatrixRowValueInputIterator adj_matrix_row_value_input_first,
+                                    AdjMatrixColValueInputIterator adj_matrix_col_value_input_first,
+                                    EdgeOp e_op,
+                                    T init,
+                                    VertexValueOutputIterator vertex_value_output_first)
```
ChuckHastings, line 171: Suggest moving the output iterator to be the fifth parameter (before `EdgeOp e_op`).  This would be more consistent with `thrust::transform` and `thrust::copy_if`.

cpp/include/patterns/transform_reduce_e.cuh
```diff
@@ -0,0 +1,220 @@
...
+ * column indices, respectively) and returns a transformed value to be reduced.
+ * @param init Initial value to be added to the transform-reduced input vertex properties.
+ * @return T Reduction of the @p edge_op outputs.
+ */
+template <typename GraphViewType,
+          typename AdjMatrixRowValueInputIterator,
+          typename AdjMatrixColValueInputIterator,
+          typename EdgeOp,
+          typename T>
+T transform_reduce_e(raft::handle_t const& handle,
```
ChuckHastings, line 156: This function failed to compile for me when T is a simple arithmetic type.  Something in the call stack expects a tuple.  Probably a subtle SFINAE bug on a function in the call stack.

The following failed to compile:
```
    weight_t total_edge_weight = experimental::transform_reduce_e(
      handle_,
      graph_,
      thrust::make_constant_iterator(0),
      thrust::make_constant_iterator(0),
      [] __device__(auto, auto, weight_t wt, auto, auto) { return wt; },
      weight_t{0});
```

cpp/include/patterns/transform_reduce_e.cuh
```diff
@@ -0,0 +1,220 @@
...
+ * @param graph_view Non-owning graph object.
+ * @param adj_matrix_row_value_input_first Iterator pointing to the adjacency matrix row input
+ * properties for the first (inclusive) row (assigned to this process in multi-GPU).
+ * `adj_matrix_row_value_input_last` (exclusive) is deduced as @p adj_matrix_row_value_input_first +
+ * @p graph_view.get_number_of_adj_matrix_local_rows().
+ * @param adj_matrix_col_value_input_first Iterator pointing to the adjacency matrix column input
+ * properties for the first (inclusive) column (assigned to this process in multi-GPU).
+ * `adj_matrix_col_value_output_last` (exclusive) is deduced as @p adj_matrix_col_value_output_first
+ * + @p graph_view.get_number_of_adj_matrix_local_cols().
+ * @param e_op Binary (or ternary) operator takes *(@p adj_matrix_row_value_input_first + i), *(@p
```
ChuckHastings, line 145: Documentation indicates this should be a binary or ternary operator.  However, it requires 4 or 5 parameters:
`(vertex_t src, vertex_t dst, weight_t w, auto src_val, auto dst_val)`
or (presumably)
`(vertex_t src, vertex_t dst, auto src_val, auto dst_val)`

ChuckHastings (changes requested):

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+ * @param graph_view Non-owning graph object.
+ * @param adj_matrix_row_value_input_first Iterator pointing to the adjacency matrix row input
+ * properties for the first (inclusive) row (assigned to this process in multi-GPU).
+ * `adj_matrix_row_value_input_last` (exclusive) is deduced as @p adj_matrix_row_value_input_first +
+ * @p graph_view.get_number_of_adj_matrix_local_rows().
+ * @param adj_matrix_col_value_input_first Iterator pointing to the adjacency matrix column input
+ * properties for the first (inclusive) column (assigned to this process in multi-GPU).
+ * `adj_matrix_col_value_output_last` (exclusive) is deduced as @p adj_matrix_col_value_output_first
+ * + @p graph_view.get_number_of_adj_matrix_local_cols().
+ * @param e_op Binary (or ternary) operator takes *(@p adj_matrix_row_value_input_first + i), *(@p
```
ChuckHastings, line 150: Seems to really need 4 or 5 parameters:  src, dest, optional weight, src_val, dst_val.

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+ * (exclusive) is deduced as @p vertex_value_output_first + @p
+ * graph_view.get_number_of_local_vertices().
+ */
+template <typename GraphViewType,
+          typename AdjMatrixRowValueInputIterator,
+          typename AdjMatrixColValueInputIterator,
+          typename EdgeOp,
+          typename T,
+          typename VertexValueOutputIterator>
+void copy_v_transform_reduce_in_nbr(raft::handle_t const& handle,
```
ChuckHastings, line 165: I'm getting an SFINAE error compiling this with a graph that is not transposed.  Haven't diagnosed why, but it complains that there is no instance of an overloaded function `cugraph::experimental::detail::accumulate_edge_op_result` that matches the argument list.

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+          typename T,
+          typename VertexValueOutputIterator>
+void copy_v_transform_reduce_out_nbr(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  AdjMatrixRowValueInputIterator adj_matrix_row_value_input_first,
+  AdjMatrixColValueInputIterator adj_matrix_col_value_input_first,
+  EdgeOp e_op,
+  T init,
+  VertexValueOutputIterator vertex_value_output_first)
```
ChuckHastings, line 255: Suggest moving the output iterator to be the fifth parameter (before `EdgeOp e_op`). This would be more consistent with `thrust::transform` and `thrust::copy_if`.

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+ * @param graph_view Non-owning graph object.
+ * @param adj_matrix_row_value_input_first Iterator pointing to the adjacency matrix row input
+ * properties for the first (inclusive) row (assigned to this process in multi-GPU).
+ * `adj_matrix_row_value_input_last` (exclusive) is deduced as @p adj_matrix_row_value_input_first +
+ * @p graph_view.get_number_of_adj_matrix_local_rows().
+ * @param adj_matrix_col_value_input_first Iterator pointing to the adjacency matrix column input
+ * properties for the first (inclusive) column (assigned to this process in multi-GPU).
+ * `adj_matrix_col_value_output_last` (exclusive) is deduced as @p adj_matrix_col_value_output_first
+ * + @p graph_view.get_number_of_adj_matrix_local_cols().
+ * @param e_op Binary (or ternary) operator takes *(@p adj_matrix_row_value_input_first + i), *(@p
```
ChuckHastings, line 233: Seems to really need 4 or 5 parameters: src, dest, optional weight, src_val, dst_val

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
```diff
@@ -0,0 +1,293 @@
...
+ * adj_matrix_col_value_input_first + j), (and optionally edge weight) (where i and j are row and
+ * column indices, respectively) and returns a value to be reduced.
+ * @param init Initial value to be added to the reduced @e_op return values for each vertex.
+ * @param vertex_value_output_first Iterator pointing to the vertex property variables for the first
```
ChuckHastings, line 237: I was confused by this while trying to use this function.  If I pass `output_v.begin()` this fails to compile, I need to pass `output_v.data().get()`.

It would be helpful to either be able to pass a thrust iterator (`thrust::copy_if` takes an iterator) or to document that you need to pass a raw pointer (or whatever the actual criteria is).

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
seunghwak: Yeah... initially, this was designed without considering hypergraph partitioning.

In that case, we can provide src & dst as part of adj_matrix_row_value_input_first & adj_matrix_col_value_input_first with thrust counting iterator.

But if hypergrpah partitioning is considered, doing this from the user space requires something more complex (that increase the cost to access src (row-major) or dst (col-major) requiring binary search).

So, I fixed this recently but forget to update the comment :-) Thank you for catching this.

See the code below for a use case.
https://github.com/rapidsai/cugraph/pull/838/files#diff-615b093890b957db6e03f2a3f49d096bR243
```
    copy_v_transform_reduce_in_nbr(
      handle,
      pull_graph_view,
      adj_matrix_row_pageranks.begin(),
      thrust::make_constant_iterator(0) /* dummy */,
      [alpha] __device__(vertex_t src, vertex_t dst, weight_t w, auto src_val, auto dst_val) {
        return src_val * w * alpha;
      },
      unvarying_part,
      pageranks);
```

cpp/include/patterns/copy_v_transform_reduce_nbr.cuh
seunghwak: So, this has two parts, copy_v and transform_reduce_in_nbr.

```raft::handle_t const& handle``` and ```GraphViewType const& graph_view``` are for both.

```AdjMatrixRowValueInputIterator adj_matrix_row_value_input_first```, ```AdjMatrixColValueInputIterator adj_matrix_col_value_input_first```, ```EdgeOp e_op```, and ```T init``` are for transform_reduce_in_nbr.

```VertexValueOutputIterator vertex_value_output_first``` is for copy_v (copying the output of transform_reduce_in_nbr for each vertex to the memory pointed by this output iterator.

So, moving ```vertex_value_output_first``` before ```e_op``` places a parameter for ```copy_v``` in the middle of parameters for ```transform_reduce_i_nbr```. This being said, do you still recommend moving?

Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:231]: If I compare this to thrust::copy, 

```AdjMatrixRowValueInputIterator adj_matrix_row_value_input_first```, ```AdjMatrixColValueInputIterator adj_matrix_col_value_input_first```, ```EdgeOp e_op```, and ```T init``` are ```InputIterator first``` and ```InputIterator last``` gets deduced from ```first``` and ```graph_view```.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 7 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:293]: Yes, the code is fixed when updating this PR to work with the new graph class, I will fix the comment.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 6 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:315]: See the reply to "copy_v_transform_reduce_in_nbr". Same here.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 8 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:297]: This sounds like a bug, this should work with both .begin() and .data().get(). I will take a look.

File: cpp/include/patterns/transform_reduce_e.cuh (hunk 3 above)
Comment (seunghwak) [reply cpp/include/patterns/transform_reduce_e.cuh:188]: Yes, this comment needs to be fixed. Thanks for catching this.

File: cpp/include/patterns/transform_reduce_e.cuh (hunk 2 above)
Comment (seunghwak) [reply cpp/include/patterns/transform_reduce_e.cuh:199]: Thanks, yes this should work for simple arithmetic types or a tuple of simple arithmetic types. Sounds like a bug, I will take a look.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 1 above)
Comment (ChuckHastings) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:231]: I think of this functionality as a transform rather than a copy, where the transform operation is a reduction over the neighbors of each vertex.  So my intuition was to think of this as a fancy thrust::transform rather than a copy.  The transform API specifies the input and output iterators first and the transformation details last.

Of course, your naming of the function suggests copy of a reduction, which is obviously how you were thinking of the function.

I think it's fine either way.  It just was not quite what I expected.  I suppose if you were to change to specifying them like a transform it might be better to name the function differently (with the transform first to emphasis that intended focus).

File: cpp/include/patterns/transform_reduce_e.cuh (hunk 2 above)
Comment (seunghwak) [reply cpp/include/patterns/transform_reduce_e.cuh:199]: This should be fixed by the new commit, and let me know if this still fails.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 1 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:230]: OK, make sense, and I will keep this as is for now.

File: cpp/include/patterns/transform_reduce_e.cuh (hunk 3 above)
Comment (seunghwak) [reply cpp/include/patterns/transform_reduce_e.cuh]: Fixed.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 4 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh]: Fixed.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 7 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh]: Fixed.

File: cpp/include/patterns/copy_v_transform_reduce_nbr.cuh (hunk 8 above)
Comment (seunghwak) [reply cpp/include/patterns/copy_v_transform_reduce_nbr.cuh:296]: This should be fixed by now. vertex_value_output_first can be either raw pointer, thrust::device_ptr, thrust::discard_iterator, or thrust::zip_iterator having a mix of raw pointer, thrust::device_ptr, and/or thrust::discard_iterator. Actual values pointed by either raw pointer or thrust::device_ptr should be arithmetic type.

**Review by ChuckHastings — APPROVED** [review]
I've been using these changes and they seem fine.

I assume more changes will be in a different PR (e.g. implementations for is_multi_gpu reductions)

**Review by afender — APPROVED** [review]
That's a lot of very useful code :tada: 

Questions: 

1. How fast is it compared to what we have?

2. What are the next steps and timeline to deprecate the existing implementations of BFS,SSSP Pagerank, and Katz to use the pattern accelerator instead?

**Comment by BradReesWork** [comment]
rerun tests

**Comment by BradReesWork** [comment]
rerun tests

**Comment by BradReesWork** [comment]
rerun tests

**Comment by pgera** [comment]
Just some early feedback on memory usage. I tested BFS with some small graphs, and it produces the correct output. I wasn't able to test twitter (5.6 GB in CSR) on a 12 GB GPU whereas the current implementation works fine for that. The graph is already in CSR here. So there is no additional memory usage beyond the CSR graph, distance array, and whatever BFS uses. That is, I'm calling BFS on 

```
cugraph::experimental::graph_view_t<VT, ET, float, false, false> G(            
      handle,                                                                   
      thrust::raw_pointer_cast(t_vlist.data()),                                 
      thrust::raw_pointer_cast(t_elist.data()),                                                                                              
      nullptr,                                                                  
      std::vector<VT> (),                                                       
      vertex_cnt,                                                               
      edge_cnt,                                                                 
      cugraph::experimental::graph_properties_t(),                              
      false,                                                                    
      false);

cugraph::experimental::bfs<VT, ET, float, false>(                           
      handle,                                                                   
      G,                                                                        
      thrust::raw_pointer_cast(t_distances.data()),                             
      static_cast<VT*>(nullptr),                                                
      static_cast<VT>(src),                                                     
      false,                                                                    
      std::numeric_limits<VT>::max(),                                           
      false);
```

BradReesWork:
rerun tests

seunghwak:
> Just some early feedback on memory usage. I tested BFS with some small graphs, and it produces the correct output. I wasn't able to test twitter (5.6 GB in CSR) on a 12 GB GPU whereas the current implementation works fine for that. The graph is already in CSR here. So there is no additional memory usage beyond the CSR graph, distance array, and whatever BFS uses. That is, I'm calling BFS on
> 
> ```
> cugraph::experimental::graph_view_t<VT, ET, float, false, false> G(            
>       handle,                                                                   
>       thrust::raw_pointer_cast(t_vlist.data()),                                 
>       thrust::raw_pointer_cast(t_elist.data()),                                                                                              
>       nullptr,                                                                  
>       std::vector<VT> (),                                                       
>       vertex_cnt,                                                               
>       edge_cnt,                                                                 
>       cugraph::experimental::graph_properties_t(),                              
>       false,                                                                    
>       false);
> 
> cugraph::experimental::bfs<VT, ET, float, false>(                           
>       handle,                                                                   
>       G,                                                                        
>       thrust::raw_pointer_cast(t_distances.data()),                             
>       static_cast<VT*>(nullptr),                                                
>       static_cast<VT>(src),                                                     
>       false,                                                                    
>       std::numeric_limits<VT>::max(),                                           
>       false);
> ```
> 
> The last two calls in the back trace are
> 
> ```
> #6  0x0000555555567878 in rmm::mr::cuda_memory_resource::do_allocate(unsigned long, CUstream_st*) ()
> #7  0x00007fffedd46940 in void cugraph::experimental::update_frontier_v_push_if_out_nbr<cugraph::experimental::graph_view_t<int, int, float, false, false, void>
> ```

Thanks for testing this, and BFS & SSSP in this PR is not optimized for performance & memory footprint (if you actually read the PR code, you may find several FIXMEs to reduce memory footprint). It will happen in future PRs, and I will make sure memory requirement is actually smaller than the previous one in the final version.

PR #840, opened 2020-04-28
[REVIEW] OPG degree

### LINKED ISSUE #810: [FEA] Degree
Same as #297 with one process per GPU and NCCL for device communications.

Degree needs to be computed from **indices**.

**Task List**
- [ ] Implementation 
- [ ] Tests : verify the result by comparing to the single GPU version
- [ ] Record performance
- [ ] API : this feature needs to be exposed in cugraph's OPG API. In addition, this needs to interact well with other OPG PageRank functions.

Dependency of #485

afender:
close #810 

- Added temporary comm class (to be replaced by RAFT). It is a lightweight stopgap that should be easy to replace.
- Added a path to set the communicator in GraphBase (might move to the Handle once we have it)
- OPG edge list host partitioner (1D) for testing
- C++ OPG Out degree implementation and test


afender:
Expected output
Screen Shot 2020-04-28 at 6 45 22 PM

cpp/include/graph.hpp
```diff
@@ -61,8 +61,11 @@ class GraphBase {
    */
   void get_vertex_identifiers(VT *identifiers) const;
 
+  void set_communicator(Comm& comm_) {comm = comm_;}
```
seunghwak, line 64: Should we set communicator for a graph object or pass a communicator object (or a handle) to a OPG function call?

And I can ask a similar question for stream as well.

If we run graph analytics, I guess there can be cases running multiple graph analytics with the same graph on multiple different streams; in this case, if we set stream for the graph object but passing both a graph object and another stream to a graph analytics function, this can be quite confusing.

cpp/src/comms/mpi/comms_mpi.cpp
```diff
@@ -0,0 +1,94 @@
...
+    ncclCommDestroy(_nccl_comm);
+
+  if (_finalize_mpi) {
+    MPI_Finalize();
+  }
+#endif
+}
+
+void Comm::barrier() {
+  cudaDeviceSynchronize();
```
seunghwak, line 89: Not a huge deal considering that this code is just to fill the gap, but this looks very expensive (and better guard with CUDA_TRY).

cpp/src/comms/mpi/comms_mpi.hpp
```diff
@@ -0,0 +1,256 @@
...
+    void allgather (size_t size, value_t* sendbuff, value_t* recvbuff) const;
+
+    template <typename value_t>
+    void allreduce (size_t size, value_t* sendbuff, value_t* recvbuff, ReduceOp reduce_op) const;
+
+};
+
+template <typename value_t>
+void Comm::allgather (size_t size, value_t* sendbuff, value_t* recvbuff) const {
+#if USE_NCCL
```
seunghwak, line 244: If we are not providing both NCCL based and CUDA-aware MPI based implementations, better just delete #if USE_NCCL everywhere.

cpp/src/structure/graph.cu
```diff
@@ -32,7 +34,9 @@ void degree_from_offsets(vertex_t number_of_vertices,
 }
 
 template <typename vertex_t, typename edge_t>
-void degree_from_vertex_ids(edge_t number_of_edges,
+void degree_from_vertex_ids(const cugraph::experimental::Comm& comm,
```
seunghwak, line 37: If we are passing comm here, why do we have https://github.com/rapidsai/cugraph/pull/840/files#diff-ba59ce02957abb53c35f1a1929fc0374R64

cpp/src/structure/graph.cu
```diff
@@ -32,7 +34,9 @@ void degree_from_offsets(vertex_t number_of_vertices,
...
+                            vertex_t number_of_vertices,
+                            edge_t number_of_edges,
                             vertex_t const *indices,
                             edge_t *degree,
```
seunghwak, line 41: And with OPG, shouldn't we better distribute degree? If every process has the degree values for the entire set of vertices, we cannot run this unless the entire set of vertices fits in a single GPU.

cpp/include/graph.hpp
afender: I think we should pass the handle everywhere and set the communicator and streams there. Then, the handle class will have all the tools to deal with those aspects and benefit to all other API at once. 
The handle PR is being reviewed in RAFT, so this should be available soon. In the meantime, setting it there looked like the shortest reasonable path to me.

cpp/src/comms/mpi/comms_mpi.cpp
afender: yea, mostly used this for debugging. I can rename this function and add a lighter barrier implementation.

cpp/src/comms/mpi/comms_mpi.hpp
afender: I'm not sure. That helps to activate/deactivate MPI dependent code. If/when we set OPG build ON by default, this will be leveraged to get a successful build for all users that don't have MPI installed.

cpp/src/structure/graph.cu
afender: `degree_from_vertex_ids` is not a member of `GraphBase` while `comm` is set in `GraphBase`.  Hence we need to pass it somehow from the Graph to this stand-alone function. Does this answer your question?

cpp/src/structure/graph.cu
afender: We should, as part of 2D partitioning. Just focussing on the shortest path for plugging in the 1D PageRank for now and identifying obstacles across the stack/product. There should be more incremental OPG work ( through RAFT and 2D) before all this reaches the "final" version.

cpp/include/graph.hpp
seunghwak: If this is just a temporary solution, then, I don't care, but in long term, I think there will be use cases like
```
// funcA and funcB can concurrently run.
graph.func_A(..., commA, streamA);
graph.func_B(..., commB, streamB);
```
We should be able to support this case, and graph.set_handle() may not be the best way to support this.

cpp/src/comms/mpi/comms_mpi.hpp
seunghwak: I don't think I am following this comment. Here, if USE_NCCL is set to 0, this function just becomes no-op. If we want to enable users to run this without MPI, removing #if USE_NCCL and #endif does no harm (I didn't mean to delete the code in between, if you interpret this as that, then, my bad, I just meant to delete the #if USE_NCCL lines and matching #endif lines, not the NCCL based implementations between those lines).

cpp/src/structure/graph.cu
seunghwak: I see, I thought degree_from_vertex_ids is a member function, and thanks for the explanation!

cpp/src/comms/mpi/comms_mpi.hpp
afender: > If we want to enable users to run this without MPI, removing #if USE_NCCL and #endif does no harm

`#include <mpi.h>` won't build if MPI is not installed

cpp/include/graph.hpp
afender: Agreed, but what you have in the 2D PoC could be a good fit too. We could have `px` and `py` communicators inside the Comm class in the Handle. The advantage is that would be a lower burden on the API and Single GPU path.

cpp/src/comms/mpi/comms_mpi.hpp
seunghwak: I see, in that case USE_NCCL might be a misnomer (better add USE_MPI).

https://github.com/rapidsai/cugraph/pull/840/files#diff-4a9c63a50daac99ed3fa518d6887cfd8R91
```
#if USE_NCCL
  MPI_Barrier(MPI_COMM_WORLD);
#endif 
```

The code above also looks problematic. Better not bog down too deep for a temporary code, but int his code, setting USE_NCCL to 1 will not allow users without MPI to build this code.

cpp/include/graph.hpp
seunghwak: Just don't forget that there might be a case to use multiple communicators with the same set of processes. I have used that in solving multiple PDEs with adaptive mesh refinement in parallel.

cpp/src/comms/mpi/comms_mpi.hpp
afender: Yea naming is misleading. Currently, MPI and NCCL are interdependent and the `USE_NCCL` macro is controlling it.

afender: Fixed naming 7ac19b1

cpp/src/comms/mpi/comms_mpi.cpp
```diff
@@ -0,0 +1,291 @@
...
+  MPI_TRY(MPI_Comm_size(MPI_COMM_WORLD, &mpi_world_size));
+  CUGRAPH_EXPECTS( (_p == mpi_world_size), 
+                   "Invalid input arguments: p should match the number of MPI processes.");
+
+  _mpi_comm = MPI_COMM_WORLD;
+
+  // CUDA
+
+  CUDA_TRY(cudaGetDeviceCount(&_device_count));
+  _device_id = _rank % _device_count; // FixMe : assumes each node has the same number of GPUs
```
BradReesWork, line 206: This seems like it could be an issues.  We should discuss configuration and how to handle
you also only set one device and not one per rank

cpp/src/comms/mpi/comms_mpi.cpp
```diff
@@ -0,0 +1,291 @@
...
+}
+
+#if ENABLE_OPG
+Comm::Comm(ncclComm_t comm, int size, int rank)
+  : _nccl_comm(comm), _p(size), _rank(rank) {
+
+  // CUDA
+  CUDA_TRY(cudaGetDeviceCount(&_device_count));
+  _device_id = _rank % _device_count; // FixMe : assumes each node has the same number of GPUs
+  CUDA_TRY(cudaSetDevice(_device_id)); // FixMe : check if this is needed or if python takes care of this
```
BradReesWork, line 238: Need to define what the FixMe tag looks like since you have various versions "Fix me" vs "FixMe"

Should be standard FIXME

cpp/src/structure/graph.cu
```diff
@@ -72,11 +78,13 @@ void GraphCOO<VT,ET,WT>::degree(ET *degree, DegreeDirection direction) const {
   cudaStream_t stream{nullptr};
 
   if (direction != DegreeDirection::IN) {
-    degree_from_vertex_ids(GraphBase<VT,ET,WT>::number_of_edges, src_indices, degree, stream);
+    if (GraphBase<VT,ET,WT>::comm.get_p()) // FixMe retrieve global source indexing for the allreduce work
+      CUGRAPH_FAIL("OPG degree not implemented for OUT degree");
+    degree_from_vertex_ids(GraphBase<VT,ET,WT>::comm, GraphBase<VT,ET,WT>::number_of_vertices, GraphBase<VT,ET,WT>::number_of_edges, src_indices, degree, stream);
```
BradReesWork, line 83: Is it GraphBase or new GraphBaseView?  PR #799 might need to be merged first

Iroy30:
It would be good to be able to run these changes on CI with mpi build ON

cpp/CMakeLists.txt
```diff
@@ -330,6 +330,7 @@ link_directories(
     "${CMAKE_CUDA_IMPLICIT_LINK_DIRECTORIES}")
 
 add_library(cugraph SHARED
+    src/comms/mpi/comms_mpi.cpp
```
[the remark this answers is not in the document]
afender: When `BUILD_MPI` is OFF, the communicator member functions become no-op. This allows running through the existing path without contaminating the whole code base with `#ifdef ENABLE OPG`outside of the communicator class.

cpp/src/structure/graph.cu
afender: We would need to reconcile either way.

cpp/src/comms/mpi/comms_mpi.cpp
afender: 5b45a5c5

cpp/include/comms_mpi.hpp
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright (c) 2019, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
afender: Updated copyrights and fixed the one `graph.cu` as well. f4c407c

cpp/src/comms/mpi/comms_mpi.cpp
afender: > This seems like it could be an issues. We should discuss configuration and how to handle

I think it is fine. We expect python to set the device and this code to be removed. This is just for C++ smoke tests for now so that I can validate the progress on the C++ part.

> you also only set one device and not one per rank

No, it does set one device per rank. In an OPG environment, this section of the code is traversed by all ranks, each one will retrieve and set its own GPU based on its rank.

PR #846, opened 2020-05-04
[WIP][skip-ci] Opg python changes

Iroy30:
Adds dask/python infrastructure for opg pagerank. Depends on PR #840  changes. Test within python/script.py can be run as pytest -s script.py for prining out communicator data (rank and size) within processes. CUDA_VISIBLE_DEVICES should be set according to the number of gpus.


GPUtester:

PR #852, opened 2020-05-06
[REVIEW] Fix BFS Notebook

BradReesWork:
this is a quick fix to the BFS notebook.  The notebook should be redone


BradReesWork:
rerun tests

PR #862, opened 2020-05-11
[REVIEW] Katz Centrality : Auto calculation of alpha parameter if set to none

### LINKED ISSUE #730: [ENH] Katz should auto-compute Alpha if not specified
The current default Alpha value seems to be too large and throws and error even on small files.  That value should be auto-computed if not specified

kaatish:
Fixes #730 
- Katz will automatically calculate alpha parameter if it is not provided by the user
- The code will no longer inform the user if the value of alpha is greater than the minimum alpha for convergence
- The code will no longer inform the user if the centrality calculation does not converge

### DIFF
--- a/python/cugraph/centrality/katz_centrality.py
+++ b/python/cugraph/centrality/katz_centrality.py
@@ -15,7 +15,7 @@ from cugraph.centrality import katz_centrality_wrapper
 
 
 def katz_centrality(G,
-                    alpha=0.1,
+                    alpha=None,
                     max_iter=100,
                     tol=1.0e-6,
                     nstart=None,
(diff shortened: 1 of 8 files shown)


GPUtester:

python/cugraph/centrality/katz_centrality.py
```diff
@@ -37,9 +37,9 @@ def katz_centrality(G,
         cuGraph graph descriptor with connectivity information. The graph can
         contain either directed (DiGraph) or undirected edges (Graph).
     alpha : float
-        Attenuation factor with a default value of 0.1.  If alpha is not less
-        than 1/(lambda_max) where lambda_max is the maximum degree
-        GDF_CUDA_ERROR is returned
+        Attenuation factor defaulted to None. If alpha is not specified then
+        it is internally calculated as 1/(lambda_max) where lambda_max is the
+        maximum degree
```
ChuckHastings, line 42: I thought that lambda_max is the absolute value of the largest eigenvalue of the graph.  I'm not sure that 1/max_degree is going to be small enough

FYI - there is a kernel in nvgraph to compute the maximum eigenvalue of the graph if you don't have something already in cuHornet.

kaatish: Largest eigenvalue of a graph is lesser than or equal to maximum degree of a graph.
`lm <= dm`
`(1/lm) >= (1/dm)`
You want your alpha to be lesser than or equal to `(1/lm)` for convergence. Therefore setting it to `(1/dm)` will gurantee convergence.

ChuckHastings: But lambda_max is usually defined as the largest eigenvalue of the matrix.  The documentation should reflect that we are setting alpha to (1 / maximum degree) - perhaps noting that the algorithm converges as long as alpha < (1 / lambda_max) and that (1 / max_degree) < (1 / lambda_max).

kaatish: Yes, I can clarify it that way.

PR #865, opened 2020-05-12
[REVIEW] Enable C/C++ docs via Doxygen

### LINKED ISSUE #847: [ENH] Publish C++ doc
The C++ documentation has mostly been for internal developers so far. We would like to expose it to users as people are showing interest for the C++ API . We should :
- Clean up and improve the doc (Doxygen format)
- Generate it through CI as we do for Python
- Publish on https://docs.rapids.ai/api

BradReesWork:
enable creation of C/C++ documents

closes #847


GPUtester:

BradReesWork:
rerun tests

docs/README.md
```diff
@@ -17,3 +17,16 @@ make html
 
 Outputs to `build/html/index.html`
 
+## View docs web page by opening HTML in browser:
+
+First navigate to `/build/html/` folder, i.e., `cd build/html` and then run the following command:
```
rlratzel, line 22: Do docs get built by default? It looks like the CMakeLists.txt defines a new target (`docs_cugraph`) which (I'm assuming?) needs to be explicitely added to a call to `make` by the user.  If that's the case, maybe the README should explain that (and possibly what new dependencies if any that need to be installed to build them)?

PR #870, opened 2020-05-13
[REVIEW] Fix louvain

### LINKED ISSUE #509: Refactor Louvain
- [x] Error Handling
- [x] C++ API
- [x] Style

### LINKED ISSUE #668: [FEA] Louvain : how delta modularity weights are passed to the aggregator?
@jwyles and I couldn't see modularity weights being used by looking at the code.
Modularity is still properly maximized so it is puzzling.
We should clarify how delta modularity weights are being used in Louvain's code. This can be done as part of #552.

ChuckHastings:
This PR refactors the implementation of Louvain.  The old implementation did not really implement Louvain.  It computed modularity (almost correctly) but then ignored the modularity computation when computing the clustering at each level of algorithm.

The new implementation is partially serial.  The portion of the algorithm that attempts to compute delta modularity for each vertex will operate on the vertices in serial.  All other aspects of the algorithm have been parallelized.  Serializing this portion of the algorithm allows us to compare results with networkx and other serial implementations to validate that the results are correct.

There is some parallelism in that step.  While we operate on the vertices in serial, we do a parallel computation (using only one warp) so that computing the delta modularity will get 1 warp worth of parallelism.

It can definitely be improved with more parallelism, but doing this and guaranteeing convergence is hard, it seemed like getting something correct and faster than serial CPU code would be a good first step.

Closes #509 
Closes #668 

With correct Louvain scores, this PR also fixes the ECG that was disabled in PR #823 and deletes some code now obsolete that was referenced by the old Louvain implementation.


GPUtester:

cpp/tests/community/ecg_test.cu
```diff
@@ -54,8 +46,7 @@ TEST(ecg, success)
   cugraph::experimental::GraphCSRView<int, int, float> graph_csr(
     offsets_v.data().get(), indices_v.data().get(), weights_v.data().get(), num_verts, num_edges);
 
-  ASSERT_NO_THROW(
-    (cugraph::nvgraph::ecg<int32_t, int32_t, float>(graph_csr, .05, 16, result_v.data().get())));
+  cugraph::nvgraph::ecg<int32_t, int32_t, float>(graph_csr, .05, 16, result_v.data().get());
```
afender, line 49: remove nvgraph namespase, ecg should not be in nvgraph

afender: (this may be added to another PR in 0.15)

ChuckHastings: Done

PR #871, opened 2020-05-14
[WIP] Updated PyTests

BradReesWork:
Updated the pytest so that all datasets are not in utils.py and shared across all tests.  This allows for adding new test datasets easier.

### DIFF
--- a/python/cugraph/tests/test_k_truss_subgraph.py
+++ b/python/cugraph/tests/test_k_truss_subgraph.py
@@ -11,6 +11,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import gc
+
 import pytest
 
 import cugraph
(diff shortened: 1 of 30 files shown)


GPUtester:

BradReesWork:
rerun tests

python/cugraph/tests/utils.py
```diff
@@ -15,6 +15,48 @@
 import pandas as pd
 
 
+DATASETS_1 = ['../datasets/netscience.csv']
```
afender, line 18: Can't figure out the naming 2A 2B 4K etc.
Should we use self-documented names?

BradReesWork: My hope was that there would have been only a common dataset, that was not the case.  I'll updated the names, the numbers started off meaning how many items in the array - but then there were odd cases.   I'll have it redo today

python/cugraph/tests/test_k_truss_subgraph.py
```diff
@@ -79,21 +77,13 @@ def compare_k_truss(graph_file, k, ground_truth_file, directed=True):
...
 
-DATASETS = [('../datasets/polbooks.csv',
-             '../datasets/ref/ktruss/polbooks.csv'),
-            ('../datasets/netscience.csv',
-             '../datasets/ref/ktruss/netscience.csv')]
-
-
-@pytest.mark.parametrize('graph_file, nx_ground_truth', DATASETS)
+@pytest.mark.parametrize('graph_file, nx_ground_truth', utils.DATASETS_4K)
 def test_ktruss_subgraph_DiGraph(graph_file, nx_ground_truth):
```
kaatish, line 81: This function has been removed in master because ktruss is not expected to run on directed graph.

BradReesWork: updated

BradReesWork:
branch got too far out of sync

PR #879, opened 2020-05-18
[REVIEW] Add docs build.sh [skip-ci]

dillon-cullinan:
Moving the docs build script used in our Jenkins job into the repository itself to streamline docs building pipeline and to push possible docs build changes to developer view for editing.

This script is written to run in a devel docker image for RAPIDS, which assumes the project is already built from source in PROJECT_WORKSPACE (this variable is defined inside the jenkins job as /rapids/$PROJECT).

The script also assumes that the docs repo is cloned and located in $DOCS_WORKSPACE. After building all the docs, it will move all html files into the proper api folders in this repo, then add the relevant folders for a git commit and git push to be executed inside the Jenkins job.

@BradReesWork Adding your review to make sure I'm doing the C++ docs correctly. Make sure lines `41-43` and `61` are correct for the build process and moving the html.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 - PR #867 Updates to support the latest flake8 version
 - PR #874 Update setup.py to use custom clean command
 - PR #878 Updated build script
+- PR #879 Add docs build script to repository
 
 ## Bug Fixes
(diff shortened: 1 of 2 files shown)


GPUtester:

PR #880, opened 2020-05-18
[REVIEW] Remove remaining gdf_column references

### LINKED ISSUE #538: [FEA] Get rid of gdf_column

ChuckHastings:
This PR removes all remaining references to gdf_column.

closes #538

Also deleted some obsolete code (C++ and python) and a few ALLOC issues that I encountered.


GPUtester:

ChuckHastings:
rerun tests

cpp/tests/Graph/Graph.cu
```diff
@@ -102,6 +102,7 @@ TEST(gdf_edge_list, success_no_weights)
 }
 */
 
+#if 0
```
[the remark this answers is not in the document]
afender: We should have some equivalent stand-alone tests for testing the new graph structure. This can be in another PR though. 
In the meantime, I think it is ok to delete dead code. We can still access it from Github history.

ChuckHastings:
> Should libcudf be removed from cmake as well or are we still depending on libcudf at the C++ level?
> 
> https://github.com/rapidsai/cugraph/blob/9330bf2c5190d6bd23bf6e7bd42c735a93974fb6/cpp/CMakeLists.txt#L177

Looks like it can be removed.  I'll do that in my next push.

ChuckHastings:
rerun tests

cpp/tests/Graph/Graph.cu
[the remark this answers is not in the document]
ChuckHastings: Removed this code.

cpp/src/converters/renumber.cuh
```diff
@@ -171,20 +105,18 @@ void renumber_vertices(size_t size,
...
+  rmm::device_vector<renumber::index_type> hash_bins_start_v(1 + hash_size,
+                                                             renumber::index_type{0});
+  rmm::device_vector<renumber::index_type> hash_bins_end_v(1 + hash_size);
 
-  ALLOC_TRY(&hash_data, 2 * size * sizeof(T_in), stream);
-  ALLOC_TRY(&hash_bins_start, (1 + hash_size) * sizeof(renumber::index_type), stream);
-  ALLOC_TRY(&hash_bins_end, (1 + hash_size) * sizeof(renumber::index_type), stream);
+  T_in *hash_data                       = hash_data_v.data().get();
+  renumber::index_type *hash_bins_start = hash_bins_start_v.data().get();
+  renumber::index_type *hash_bins_end   = hash_bins_end_v.data().get();
```
seunghwak, line 115: Just FYI, thrust has raw pointer cast I started to use this (i.e. thrust::raw_pointer_cast(hash_data_v.data()), I think this makes the intention (casting a thrust::device_ptr to a raw pointer) clearer.

ChuckHastings: I used that construct when I started the cudf++ work and Jake preferred this syntax... so I've been using this syntax instead.  It's certainly a construct we use a great deal, we should probably come to some agreement on what we prefer.

seunghwak: > I used that construct when I started the cudf++ work and Jake preferred this syntax... so I've been using this syntax instead. It's certainly a construct we use a great deal, we should probably come to some agreement on what we prefer.

Good to know! I just searched cudf codebase for raw_pointer_cast and data().get() and raw_pointer_cast appeared only few times except for the legacy parts and data().get() is dominantly used.

Better just follow cudf's convention, and I will switch back to data().get() :-)

PR #881, opened 2020-05-19
[REVIEW] Raft integration infra

Iroy30:
To test the infra, follow the cugraph build/install steps and run pytest -s cugraph/test/test_raft.py


GPUtester:

python/setup.py
```diff
@@ -58,6 +59,12 @@
     conda_include_dir = conda_prefix + '/include'
     conda_lib_dir = conda_prefix + '/lib'
 
+# Optional location of C++ build folder that can be configured by the user
```
dantegd, line 62: I would add some printing to inform developers how things are being compiled, you can use the `get_environment_option` of setuputils, that'll print a cmake style message of what options are being used at build time.

cpp/CMakeLists.txt
```diff
@@ -334,6 +367,7 @@ add_library(cugraph SHARED
 #        wait until after cugunrock is constructed.
 #
 add_dependencies(cugraph cugunrock)
+add_dependencies(cugraph raft)
 
 if (BUILD_MPI)
```
afender, line 372: `if (BUILD_MPI)` should not impact RAFT.

Recall that this is false on CI so CI will ignore this RAFT integration if we leave this as is. We should : 
- Disable MPI code (or keep it encapsulated)
- Enable NCCL and RAFT by default and test through CI.

python/setup.py
Iroy30: The get_environment_option in the consequent line is indeed helpful for the user at build time. Also, added the update you mentioned for including the raft include directory.

PR #882, opened 2020-05-19
[REVIEW] Add Force Atlas 2 to benchmarks

hlinsen:


GPUtester:

hlinsen:
rerun tests

benchmarks/bench_algos.py
```diff
@@ -177,6 +177,11 @@ def bench_bfs(gpubenchmark, anyGraphWithAdjListComputed):
     gpubenchmark(cugraph.bfs, anyGraphWithAdjListComputed, 0)
 
 
+def bench_force_atlas2(gpubenchmark, anyGraphWithAdjListComputed):
+    gpubenchmark(cugraph.force_atlas2, anyGraphWithAdjListComputed,
+                 max_iter=50)
```
afender, line 182: just curious why `max_iter=50` and not rely on the default 500 (or set default to 50)?

hlinsen: The benchmark takes a lot of time with 500 iterations. I wasn't sure if we care to have an algorithm that takes much more time to benchmark, so I wanted to keep it a minimum consistent with the other iterative algorithms.

PR #887, opened 2020-05-20
[REVIEW] Update test datasets

BradReesWork:
moved all datasets into utils and updated all test to use the common arrays from utils.  This will make it easier to add more test sets and to ensure consistent testing


GPUtester:

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -39,11 +39,11 @@
 DEFAULT_EPSILON = 0.0001
 IMPLEMENTATION_OPTIONS = ['default', 'gunrock']
 
-TINY_DATASETS = ['../datasets/karate.csv']
+# TINY_DATASETS = ['../datasets/karate.csv']
```
afender, line 42: delete?

python/cugraph/tests/utils.py
```diff
@@ -14,6 +14,51 @@
 import cudf
 import pandas as pd
 
+#
+# Datasets are numbered based on the number of elements in the array
+#
+DATASETS_1 = ['../datasets/netscience.csv']
```
afender, line 20: We could also remove datasets that don't bring much value? Not sure if dolphins/polbooks are really testing something different than karate.

Lots of duplicates in the lists also. I wonder if we could just group them by directed/undirected sets.

BradReesWork: You hit on the point/scope of this refactoring.  The first step was to just centralize the data.  The next is to make better tests.

BradReesWork:
Pushing this to 0.15

python/cugraph/tests/test_betweenness_centrality.py
BradReesWork: There are some test that do not pass on the larger dataset, or get really slow.  Addressing those is step 2

python/cugraph/tests/test_bfs.py
```diff
@@ -250,7 +242,7 @@ def test_bfs(graph_file, directed, seed):
                 seed=seed)
 
 
-@pytest.mark.parametrize('graph_file', DATASETS)
+@pytest.mark.parametrize('graph_file', utils.DATASETS)
```
Iroy30, line 245: Should this be DATASETS_5

python/cugraph/tests/test_filter_unreachable.py
```diff
@@ -36,7 +35,7 @@
 SOURCES = [1]
 
 
-@pytest.mark.parametrize('graph_file', ['../datasets/netscience.csv'])
+@pytest.mark.parametrize('graph_file', utils.DATASETS)
```
Iroy30, line 38: Should this be DATASETS_1 as in test_sssp.py

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -37,7 +37,6 @@
 # =============================================================================
 DIRECTED_GRAPH_OPTIONS = [False, True]
 DEFAULT_EPSILON = 0.0001
-IMPLEMENTATION_OPTIONS = ['default', 'gunrock']
 
 TINY_DATASETS = ['../datasets/karate.csv']
```
Iroy30, line 41: can we get these datasets from utils as well?

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -258,64 +248,56 @@ def prepare_test():
 # =============================================================================
 @pytest.mark.parametrize('graph_file', TINY_DATASETS)
```
Iroy30, line 249: get datasets from utils

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -258,64 +248,56 @@ def prepare_test():
...
     """Test Unnormalized Betweenness Centrality"""
     prepare_test()
     cu_bc, nx_bc = calc_betweenness_centrality(graph_file, directed=directed,
                                                normalized=True,
-                                               implementation=implementation,
                                                result_dtype=result_dtype)
     compare_scores(cu_bc, nx_bc)
 
 
 @pytest.mark.parametrize('graph_file', SMALL_DATASETS)
```
Iroy30, line 291: get datasets from utils

python/cugraph/tests/test_utils.py
```diff
@@ -17,12 +17,6 @@
 import cugraph
 from cugraph.tests import utils
 
-DATASETS = ['../datasets/dolphins.csv',
```
Iroy30, line 20: get the karate dataset from util

python/cugraph/tests/test_filter_unreachable.py
BradReesWork: To match what was there yes, but to move every test to using a common set then no.  
I had left some using the smaller number of test since there seemed to be issues

python/cugraph/tests/test_bfs.py
BradReesWork: To match what was there yes, but to move every test to using a common set then no.

python/cugraph/tests/test_betweenness_centrality.py
BradReesWork: This was a case where the test defined a different dataset than others.  The next step is to start addressing issues what the common test set does not work across all test cases

python/cugraph/tests/test_betweenness_centrality.py
BradReesWork: I did want to creat yet another data test set.  This is slated for the next round of clean up and refactoring

PR #889, opened 2020-05-20
[REVIEW] Added missing conftest.py file to benchmarks

rlratzel:
I forgot to add this to the last commit: this file adds the `--no-rmm-reinit` feature, which is useful for faster benchmarking runs when RMM reinit isn't needed.


GPUtester:

PR #891, opened 2020-05-21
[REVIEW] A few gdf_column stragglers

ChuckHastings:
Deleted code (wasn't being compiled, but would certainly lead to confusion) that still referenced gdf_column.

Missed a legacy cudf reference in a .pxd file.

Deleted a few obsolete headers.


GPUtester:

PR #893, opened 2020-05-21
[REVIEW] Add external_repositories dir and raft symlink to .gitignore

trxcllnt:
These files seem to be created by setup.py but shouldn't be tracked by git.


GPUtester:

PR #896, opened 2020-05-21
[REVIEW] dask infra updates

Iroy30:
Small fixes for when https://github.com/rapidsai/raft/pull/7 merged

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -68,6 +68,7 @@ conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaul
       dask-cuda=${MINOR_VERSION} \
       scikit-learn>=0.21 \
       nccl>=2.5 \
+      ucx-py=${MINOR_VERSION} \
       libcypher-parser \
       ipython=7.3* \
       jupyterlab
--- a/conda/environments/cugraph_dev_cuda10.0.yml
+++ b/conda/environments/cugraph_dev_cuda10.0.yml
@@ -14,10 +14,10 @@ dependencies:
 - dask-cuda=0.14*
 - dask-cudf=0.14*
 - nccl>=2.5
+- ucx-py=0.14*
 - scipy
 - networkx
 - python-louvain
-- nccl
 - cudatoolkit=10.0
 - cmake>=3.12
 - python>=3.6,<3.8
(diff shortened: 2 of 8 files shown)


conda/environments/cugraph_dev_cuda10.0.yml
```diff
@@ -14,10 +14,10 @@ dependencies:
 - dask-cuda=0.14*
 - dask-cudf=0.14*
 - nccl>=2.5
+- ucx-py=0.14*
```
raydouglass, line 17: Is this an actual change in dependencies for cugraph? If so, the conda recipes need to be updated as well.

afender: Yes, and probably `ci/gpu/build.sh` around line 70 too?

Iroy30: Thanks, I have added those

afender:
rerun tests

conda/recipes/cugraph/meta.yaml
```diff
@@ -28,11 +28,13 @@ requirements:
     - cython>=0.29,<0.30
     - libcugraph={{ version }}
     - cudf={{ minor_version }}
+    - ucx-py {{ minor_version }}
   run:
     - python x.x
     - libcugraph={{ version }}
     - cudf={{ minor_version }}
     - nccl>=2.5
```
raydouglass, line 36: You removed nccl from the environment files. Should it also be removed here?

ci/gpu/build.sh
```diff
@@ -68,6 +68,7 @@ conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaul
       dask-cuda=${MINOR_VERSION} \
       scikit-learn>=0.21 \
       nccl>=2.5 \
```
raydouglass, line 70: You removed nccl from the environment files. Should it also be removed here?

Iroy30: nccl is still needed in the environment. I removed nccl as it was redundant. We already have a nccl>=2.5 dependency

conda/environments/cugraph_dev_cuda10.1.yml
```diff
@@ -13,10 +13,10 @@ dependencies:
 - dask-cuda=0.14*
 - dask-cudf=0.14*
 - nccl>=2.5
+- ucx-py=0.14*
 - scipy
 - networkx
 - python-louvain
-- nccl
```
Iroy30, line 19: removed nccl as there is already a dependency

Issue #877, opened 2020-05-15
[ENH] Remove ALLOC_TRY / ALLOC_FREE_TRY from BFS and SSSP


xcadet:
**Describe the solution you'd like**
BFS and SSSP currently use ALLOC_TRY and ALLOC_FREE_TRY and `rmm::device_vector` / `thrust::device_vector` could be used instead in the BFS and SSSP classes.

BradReesWork:
closed by #897

Issue #459, opened 2019-08-14
[ENH] Enforce exception safety for memory allocations.


seunghwak:
**Describe the bug**
If we allocate memory through malloc or RMM_ALLOC, this memory block doesn't get automatically freed if exception occurs between memory allocation and memory deallocation. We need to replace all the direct calls to RMM_ALLOC and malloc with memory allocations with RAII objects.

In the planned code restructuring phase, we should use std::vector, rmm:device_vector and rmm::device_buffer (https://github.com/rapidsai/rmm/pull/99, PR submitted, haven't merged, yet) where-ever possible (and at least smart-pointers which automatically release memory on stack unwinding).

afender:
> We need to replace all the direct calls to RMM_ALLOC and malloc with memory allocations with RAII objects. [...] we should use std::vector, rmm:device_vector and rmm::device_buffer

cuML is also moving and generalizing its device buffer in RAFT. Seems like we could use that too and rely on it for interacting with RMM properly. What do you think?

seunghwak:
It depends on what we see in RAFT and RAPIDS policies.

The current approach(https://github.com/rapidsai/cuml/blob/branch-0.11/wiki/cpp/DEVELOPER_GUIDE.md#device-and-host-memory-allocations) is not exception safe (still requires explicit deallocate calls).

If cugraph is also expected to be based on RAFT, then we should try our best to adopt RAFT (in a way we are currently trying our best to adopt cudf & rmm) as much as possible. If not, I think using cudf & rmm containers (resource handles) is good enough.

seunghwak:
I just found there is MLCommon::device_buffer and if we are encouraged use raft::device_buffer (which might be a wrapper for rmm::device_buffer), then, we should, but I am a bit hesitant about adding one more abstraction layer unless there is a clear benefit (may be stream management?).

seunghwak:
RMM_ALLOC is one source of memory leak on exceptions, but there are more (e.g. malloc, new, and cudaMalloc, and ALLOC_TRY).

And we're using these extensively all over the codebase. Fixing all these will touch possibly over a hundred of files and can lead to many PR merge conflicts.

I think it is dangerous to fix all these in this release.

BradReesWork:
closed by #897

PR #897, opened 2020-05-22
[REVIEW] Remove RMM_ALLOC calls

kaatish:
Addresses #459 #877 #727


GPUtester:

cpp/src/components/rmmAllocatorAdapter.hpp
```diff
@@ -1,96 +0,0 @@
-/*
```
afender, line 1: How about ` cuml_allocator.hpp`?

```

./cpp/src/components/cuml_allocator.hpp:104:/** Default cudaMalloc/cudaFree based device allocator */
./cpp/src/components/cuml_allocator.hpp:110:    CUDA_CHECK(cudaMalloc(&ptr, n));
./cpp/src/components/cuml_allocator.hpp:125:/** Default cudaMallocHost/cudaFreeHost based host allocator */
./cpp/src/components/cuml_allocator.hpp:131:    CUDA_CHECK(cudaMallocHost(&ptr, n));

```

cpp/src/utilities/cusparse_helper.h
```diff
@@ -15,6 +15,7 @@
  */
 #pragma once
 #include <cusparse.h>
+#include <rmm/device_buffer.hpp>
 #include "rmm_utils.h"
```
afender, line 19: Should rmm_utils.h be removed? 
Seems like `./cpp/include/rmm_utils.h:37:#define ALLOC_TRY(ptr, sz, stream)` should go away

afender: If we do so we can get rid of this too: https://github.com/kaatish/cugraph/blob/a0699aa3f8863df00ca30cf7e39da52d0c92ea52/cpp/src/utilities/error_utils.h#L113

cpp/src/components/rmmAllocatorAdapter.hpp
kaatish: I left this alone because those lines belong to the default allocator. Perhaps that whole file can be removed since MLCommon::deviceAllocator is not being used anywhere?

cpp/src/utilities/cusparse_helper.h
kaatish: rmm_utils.h is used everywhere just to include error_utils.h, rmm.h and thrust_rmm_allocator.h. We could remove this file, but it would mean adding all those three everywhere else. What would you suggest?

cpp/src/components/rmmAllocatorAdapter.hpp
afender: ok, let's give it a try

cpp/src/utilities/cusparse_helper.h
afender: Reducing the number of `#include`  may not be a strong enough motivation to keep  `rmm_util.hpp`. If we start going down that path the risk is that we keep adding header files to encapsulate other `#include`. What do you think?

kaatish: Sounds good. Removed `rmm_utils.h`

cpp/src/components/rmmAllocatorAdapter.hpp
kaatish: done

cpp/tests/community/balanced_edge_test.cpp
```diff
@@ -78,7 +78,7 @@ TEST(balanced_edge, success)
 int main(int argc, char** argv)
 {
   testing::InitGoogleTest(&argc, argv);
-  auto resource = std::make_unique<rmm::mr::cuda_memory_resource>();
```
kaatish, line 81: @afender @ChuckHastings I have changed cuda memory resource to cnmem given that its my understanding that we would prefer to test for the pool allocator. Any inputs on this? Should all tests do this?

afender: Yes rmmInitialize(nullptr) is fine. We stopped testing all pool/managed configs for all algorithms.

afender: This is going to be refactored in 0.15 so better not spend too much time on this.

PR #907, opened 2020-05-27
[REVIEW] Fix BFS for directed graphs

xcadet:
The current version of BFS may skip vertices in the last iteration when running on directed graphs.
This fixes the behavior, and avoid a potential Kernel Configuration issue by ensuring that the kernel is started with a positive value.


GPUtester:

PR #913, opened 2020-05-28
[REVIEW] Use Numba to allocate the array directly

jakirkham:
Since Numba has added support for external memory managers (EMM) and cuDF configures Numba to use RMM on import, we can just use Numba directly here. This should eliminate the last usage of `device_array` in cuGraph :)

xref: https://github.com/rapidsai/rmm/issues/180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## New Features
 
 ## Improvements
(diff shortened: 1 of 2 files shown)


afender:
Let's rerun tests once #916 and #917 are merged.

jakirkham:
rerun tests

PR #916, opened 2020-05-28
[REVIEW] Fix CI error on Force Atlas 2 test

hlinsen:
Test happen to break on trustworthiness assertion in pytest.
since the algorithm is not deterministic it is possible that the trustworthiness computed is not the same everytime. I adjusted the threshold values so that we are more flexible against randomness.


GPUtester:

PR #917, opened 2020-05-28
[REVIEW] Remove gunrock from Betweenness Centrality

xcadet:
The current version of Betweenness Centrality proposed two implementation, 'default' and 'gunrock', as the 'default' implementation cover more cases the 'gunrock' path can be removed. Furthermore some tests over the 'gunrock' path seemed to break under various configuration.
This removes the 'gunrock' path and the possibility to choose between implementation.
The implementation parameter is removed, and only the 'default' path persists.


GPUtester:

PR #919, opened 2020-05-29
[REVIEW] Pagerank fix

afender:
double-free bug found by @rlratzel 
`pagerank` and `tmp` thrust vectors are both scope managed now, but raw pointers to their device memory are std::swapped inside the solver. By the time the scope ends for `pagerank` its device pointer may be pointing to the data that was freed by `tmp`.
Just doing a cpy for now as a fix since this may soon be replaced by the pattern accelerator.
A thrust::swap would probably be ideal if the vectors were passed to this section of the code instead of  raw pointers.


GPUtester:

conda/recipes/libcugraph/meta.yaml
```diff
@@ -4,18 +4,17 @@
...
 
 source:
-  path: ../../..
+  git_url: ../../..
 
 build:
-  number: {{ git_revision_count }}
-  string: cuda{{ cuda_version }}_{{ git_revision_count }}
+  number: {{ GIT_DESCRIBE_NUMBER }}
+  string: cuda{{ cuda_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
raydouglass, line 17: Something is wrong with how you branched or you're targeting the wrong branch. These are v0.15 changes.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -77,7 +77,10 @@ bool pagerankIteration(IndexType n,
     return true;
   } else {
     if (iter < max_iter) {
-      std::swap(pr, tmp);
+      // A thrust::swap would probably be more efficent if the vectors were passed everywhere
+      // instead of pointers. std::swap is unsafe though. Just copying for now, as this may soon be
+      // replaced by the pattern accelerator.
+      copy(n, pr, tmp);
```
seunghwak, line 83: and just FYI, what you could do instead is something like
```
pagerank(value_t* pr, ...) {
  rmm::device_uvector<value_t> tmp{};

  value_t* p_new = pr;
  value_t* p_old = tmp.data().get();
  ...
  // at the end of iteration
  if (iter % 2 == 0) {
    p_new = tmp.data().get();
    p_old = pr;
  }
  else {
    p_new = pr;
    p_old = tmp.data().get();
  } 
}
```

afender:
#923 made it to 0.15 as well

PR #920, opened 2020-06-01
[REVIEW] Graph edges, test and storage updates

Iroy30:
This PR updates the following:
--> Bfs test to test on unsymmetric directed input. 
--> Add a directed_edges flag to number_of_edges() so as to get directed edges in algorithm wrappers
--> avoid recalculation/storage of adjlist(transposedadjlist) if transposedadjlist(adjlist) is already present in case of Graph() input. When the user initializes a Graph(), the adj list is same as transposed adj list.

### DIFF
--- a/python/cugraph/tests/test_bfs.py
+++ b/python/cugraph/tests/test_bfs.py
@@ -61,12 +61,15 @@ def prepare_test():
 def build_graphs(graph_file, directed=True):
     # cugraph
     cu_M = utils.read_csv_file(graph_file)
+    # Get unsymmetrized/directed edges
+    cu_M = cu_M[cu_M['0'] <= cu_M['1']].reset_index(drop=True)
     G = cugraph.DiGraph() if directed else cugraph.Graph()
     G.from_cudf_edgelist(cu_M, source='0', destination='1')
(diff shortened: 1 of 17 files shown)


python/cugraph/tests/test_bfs.py
```diff
@@ -61,12 +61,15 @@ def prepare_test():
...
     cu_M = utils.read_csv_file(graph_file)
+    # Get unsymmetrized/directed edges
+    cu_M = cu_M[cu_M['0'] <= cu_M['1']].reset_index(drop=True)
     G = cugraph.DiGraph() if directed else cugraph.Graph()
     G.from_cudf_edgelist(cu_M, source='0', destination='1')
     G.view_adj_list()  # Enforce CSR generation before computation
 
     # networkx
     M = utils.read_csv_for_nx(graph_file)
+    M = M[M['0'] <= M['1']]
```
rlratzel, line 72: Does this Pandas DF also need `.reset_index(drop=True)` ?

Iroy30: Indices in networkx case don't cause an issue as they are stored as dictionary in their Graphs internally. Cugraph stores the cudf dataframes as is and works on it, so indices should ideally be reset

PR #930, opened 2020-06-04
[REVIEW] Remove nvcc/thrust dependency in test utility header file

### LINKED ISSUE #921: [ENH] We have two test utility header files (test_utils.h & test_utils.hpp) and they cannot be included from cpp files.
We have two test utility header files (test_utils.h and test_utils.hpp) and it's hard to justify having two header files with the only difference in extension.

These files use thrust calls requiring nvcc to compile. This prevents including these files from cpp files (compiled with normal C++ compilers). We should support people calling cuGraph from C++ files so our test code mimicking end user use cases should test using cuGraph in C++ files.

We should merge the two header files to one. Thrust should not be the part of the public test_utils API. If thrust call is needed, they should be moved to test_utils.cu file and should not be visible to end users including this header file.

seunghwak:
fix #921 

This PR includes

* remove thrust dependency in test_utils.h to allow including this file from .cpp files.
* remove dead code
* rename tests/test_utils.h to tests/utilities/test_utilities.hpp following cuDF convention.
* bug fix in generate_graph_csr_from_mm
* pagerank_test.cu to pagerank_test.cpp (ideally, all our test cases should compile with standard C++ compiler to test our public API is callable from .cpp files).


BradReesWork:
rerun tests

cpp/tests/centrality/betweenness_centrality_test.cu
```diff
@@ -208,7 +212,7 @@ typedef struct BC_Usecase_t {
   {
     // assume relative paths are relative to RAPIDS_DATASET_ROOT_DIR
     // FIXME: Use platform independent stuff from c++14/17 on compiler update
-    const std::string &rapidsDatasetRootDir = get_rapids_dataset_root_dir();
+    const std::string &rapidsDatasetRootDir = cugraph::test::get_rapids_dataset_root_dir();
```
rlratzel, line 215: I'm assuming we're moving this to RAFT someday too? (I may already have a FIXME for that somewhere)

cpp/tests/utilities/test_utilities.hpp
```diff
@@ -15,39 +15,22 @@
...
-#include <iostream>
-#include <limits>
-#include <map>
-#include <sstream>
-#include <string>
-#include <utility>
-extern "C" {
-#include "mmio.h"
-}
+#include "utilities/test_utilities.hpp"
```
rlratzel, line 18: Is this file `#include`'ing itself?

cpp/tests/centrality/betweenness_centrality_test.cu
seunghwak: I assume so.

cpp/tests/utilities/test_utilities.hpp
seunghwak: Oops, nice catch!!! I will fix this.

seunghwak:
rerun tests

seunghwak:
rerun tests

seunghwak:
rerun tests

PR #937, opened 2020-06-05
[REVIEW] add wrapper for gunrock HITS algorithm

ChuckHastings:
Integration of the gunrock HITS algorithm.

Note, gunrock currently assumes that input and output comes from host memory.  https://github.com/gunrock/gunrock/issues/759 is a request for gunrock to modify this.  In the meantime, this wrapper will pull the CSR from device memory to the host to call gunrock, and will take the host results and copy them into the device memory.

### DIFF
--- a/cpp/src/link_analysis/gunrock_hits.cu
+++ b/cpp/src/link_analysis/gunrock_hits.cu
@@ -1,0 +1,101 @@
+/*
+ * Copyright (c) 2020, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * ---------------------------------------------------------------------------*
+ * @brief wrapper calling gunrock's HITS analytic
+ * --------------------------------------------------------------------------*/
+
+#include <algorithms.hpp>
+#include <graph.hpp>
+
+#include <utilities/error_utils.h>
+
+#include <rmm/thrust_rmm_allocator.h>
+#include <thrust/for_each.h>
+
+#include <gunrock/gunrock.h>
+
+namespace cugraph {
+
+namespace gunrock {
+
+template <typename vertex_t, typename edge_t, typename weight_t>
+void hits(cugraph::experimental::GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+          int max_iter,
+          weight_t tolerance,
+          weight_t const *starting_value,
+          bool normalized,
+          weight_t *hubs,
+          weight_t *authorities)
+{
+  //
+  //  NOTE:  gunrock doesn't support tolerance parameter
+  //         gunrock doesn't support passing a starting value
+  //         gunrock doesn't support the normalized parameter
+  //
+  //  FIXME: gunrock uses a 2-norm, while networkx uses a 1-norm.
+  //         They will add a parameter to allow us to specify
+  //         which norm to use.
+  //
+  std::vector<edge_t> local_offsets(graph.number_of_vertices + 1);
+  std::vector<vertex_t> local_indices(graph.number_of_edges);
+  std::vector<weight_t> local_hubs(graph.number_of_vertices);
+  std::vector<weight_t> local_authorities(graph.number_of_vertices);
+
+  //    Ideally:
+  //
+  //::hits(graph.number_of_vertices, graph.number_of_edges, graph.offsets, graph.indices,
+  //       max_iter, hubs, authorities, DEVICE);
+  //
+  //    For now, the following:
+
+  CUDA_TRY(cudaMemcpy(local_offsets.data(),
+                      graph.offsets,
+                      (graph.number_of_vertices + 1) * sizeof(edge_t),
+                      cudaMemcpyDeviceToHost));
+  CUDA_TRY(cudaMemcpy(local_indices.data(),
+                      graph.indices,
+                      graph.number_of_edges * sizeof(vertex_t),
+                      cudaMemcpyDeviceToHost));
+
+  ::hits(graph.number_of_vertices,
+         graph.number_of_edges,
+         local_offsets.data(),
+         local_indices.data(),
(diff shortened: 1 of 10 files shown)


ChuckHastings:
rerun tests

BradReesWork:
rerun tests

cpp/src/link_analysis/gunrock_hits.cu
```diff
@@ -0,0 +1,101 @@
+/*
```
BradReesWork, line 1: Why can't this be header only?   Also, if not header,  since there are no cuda kernels here, why not .cpp

python/cugraph/link_analysis/hits.pxd
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
BradReesWork, line 1: date needs to be updated to 2020.  The date represents year when the file was created and if two years, the  modified year

python/cugraph/link_analysis/hits_wrapper.pyx
```diff
@@ -0,0 +1,73 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
BradReesWork, line 1: date

cpp/src/link_analysis/gunrock_hits.cu
```diff
@@ -0,0 +1,101 @@
...
+{
+  //
+  //  NOTE:  gunrock doesn't support tolerance parameter
+  //         gunrock doesn't support passing a starting value
+  //         gunrock doesn't support the normalized parameter
+  //
+  //  FIXME: gunrock uses a 2-norm, while networkx uses a 1-norm.
+  //         They will add a parameter to allow us to specify
+  //         which norm to use.
+  //
```
BradReesWork, line 53: no error checking?   I'm fine with error checking be a FIXME so that users can start testing out the value of HITS sooner, but eventually, it should validate that a graph was passed in

python/cugraph/tests/test_hits.py
```diff
@@ -0,0 +1,137 @@
+# Copyright (c) 2019, NVIDIA CORPORATION.
```
BradReesWork, line 1: date

python/cugraph/tests/test_hits.py
```diff
@@ -0,0 +1,137 @@
...
+    # so here we do this explicitly
+    print('Format conversion ... ')
+
+    # Directed NetworkX graph
+    Gnx = nx.from_pandas_edgelist(M, source='0', target='1',
+                                  create_using=nx.DiGraph())
+
+    # Networkx Hits Call
+    print('Solving... ')
+    t1 = time.time()
```
BradReesWork, line 74: in order to accurately compare run times, "from_cudf_edgelist" and "from_pandas_edgelist"  need to be in the timed blocks.
cuGraph does late binding, so creating the CSR is not done until the HITS call, while Nx will spend a lot of time creating the graph when data is loaded.

python/cugraph/tests/test_hits.py
```diff
@@ -0,0 +1,137 @@
...
+
+    # same parameters as in NVGRAPH
+    pr = nx.hits(Gnx, max_iter, tol, normalized=True)
+    t2 = time.time() - t1
+
+    print('Networkx Time : ' + str(t2))
+
+    return pr
+
+
```
BradReesWork, line 84: I'm trying to get all tests to use a common set of data files.   Please use   utils.DATASETS

cpp/src/link_analysis/gunrock_hits.cu
ChuckHastings: Can't be header only.  We need to instantiate the template in order for it to be called from the python library.  I should be able to make it cpp.

python/cugraph/link_analysis/hits.pxd
ChuckHastings: Fixed

python/cugraph/tests/test_hits.py
ChuckHastings: Fixed

python/cugraph/tests/test_hits.py
ChuckHastings: Fixed.

This was literally copied from test_pagerank.py, so this change should be applied there as well.

cpp/src/link_analysis/gunrock_hits.cu
ChuckHastings: Added a couple of error checks.

We should talk about, generally, what level of error checking we want.  When the C++ code was converted to use GraphCOOView/GraphCSRView/GraphCSCView, most of the old error checks for graph became irrelevant.

The old checks verified that graph pointer was valid, and that the graph had the correct type (COO, CSR or CSC) as expected by the algorithm.  There is no graph pointer, it is a reference.  The reference is to the expected type.  So the compiler automatically does those checks (the code won't compile if you don't pass a GraphCSRView to the call).

python/cugraph/tests/test_hits.py
ChuckHastings: Two of the data sets in the test_utils.DATASETS list fail this test.

1. polbooks.csv raises an exception (power iteration doesn't converge).  Since gunrock hits doesn't attempt to converge, just runs a fixed number of iterations... it generates an answer.
1. netscience creates a different answer.  Would need to investigate why the answers are different.  It's not subtle, some of the answers in netscience are null, some of the vertices have scores that are large where the gunrock score is zero.

Not sure how we can accommodate issue 1 when all tests use the same data.  It's certainly possible that a data set might not converge in a power iteration algorithm.

I will need to investigate further why the netscience data set creates a totally different answer.

cpp/src/link_analysis/gunrock_hits.cu
ChuckHastings: Changed it to a cpp

python/cugraph/tests/test_hits.py
BradReesWork: thanks.  I'll update the test timings

python/cugraph/link_analysis/hits.py
```diff
@@ -0,0 +1,81 @@
...
+    Examples
+    --------
+    >>> gdf = cudf.read_csv('datasets/karate.csv', delimiter=' ',
+    >>>                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> G = cugraph.Graph()
+    >>> G.from_cudf_edgelist(gdf, source='0', destination='1')
+    >>> hits = cugraph.hits(G, max_iter = 50)
+    """
+
+    df = hits_wrapper.hits(G, max_iter, tol)
```
Iroy30, line 79: Maybe pass nstart to wrapper to raise the error to the user

BradReesWork:
rerun tests

PR #940, opened 2020-06-09
[REVIEW] MG (Batch) Betweenness Centrality

xcadet:
This work adds Multi-GPU Batch Betweenness Centrality for both Vertices and Edges:
- Data Replication is operated once for the graph using dask and RAFT.
- Task Distribution is operated at the Cython level

This version allows to use the same Graph for both Single GPU algorithms and algorithms supporting Multi GPUs Batch 
processing.
Graphs can be loaded the same way they are with `cudf` for single GPU, and can leverage Multi GPUs Batch Processing by using `G.enable_mg_batch()` where `G` is a `Graph` or `DiGraph`. By enabling `mg_batch` edge list and / or adjacency lists are replicated across device using `dask` and `RAFT`'s communicator. When enabled and when adding  representations to the graph, replications of the representations are operated.  Replication of the transposed adjacency list is not yet implemented.

This also adds functions for replciation of `cudf.DataFrame` and `cudf.Series`  over multiple GPUs, and a [notebook](https://github.com/xcadet/cugraph/blob/fea_batch_opg/notebooks/demo/batch_betweenness.ipynb) to show how to use Multi GPUs Batch Processing for Betweenness Centrality

A boolean parameter is added at the C++ level to BFS to disable MG BFS path to be taken when called from MG Batch BC.


GPUtester:

xcadet:
rerun tests

cpp/include/graph.hpp
```diff
@@ -661,4 +663,178 @@ template <typename ET>
...
+  void distribute_indices()
+  {
+    size_t indices_size = GT::graph.number_of_edges;
+    GT::handle_->get_comms().bcast(GT::graph.indices, indices_size, 0, GT::handle_->get_stream());
+  }
+
+  rmm::device_vector<ET> d_offsets_;
+  rmm::device_vector<VT> d_indices_;
+};
+}  // namespace opg
```
xcadet, line 839: Should this be moved somewhere else?

kaatish:
Are the current CI failures related to [this](https://github.com/rapidsai/cudf/pull/5340#pullrequestreview-448965075)?

cpp/include/graph.hpp
```diff
@@ -661,4 +663,179 @@ template <typename ET>
...
+ * @brief              A Replicatable Graph representation to replicate
+ *                     a graph using the handle
+ *
+ * @tparam VT              Type of vertex id
+ * @tparam ET              Type of edge id
+ * @tparam WT              Type of weight
+ * @tparam GraphTypeView   Type of Graph View
+ */
+template <typename VT, typename ET, typename WT, typename GraphTypeView>
+class ReplicatableGraph {
```
kaatish, line 677: Would it be possible/easier to have the replication done on the python side? That way we won't need to add another class. Is there anything that the c++ layer makes it better to do this as opposed to the python layer?

xcadet: > Are the current CI failures related to this?

Yes, the current CI failures should be fixed after merging #1008.

xcadet:
> Are the current CI failures related to this?

Yes, the current CI failures should be fixed after merging #1008.

cpp/include/graph.hpp
```diff
@@ -661,4 +663,179 @@ template <typename ET>
...
+   *
+   * @param[in] handle                Library handle (RAFT). If a communicator is expected
+   * to be initialized already
+   * @param[in]  graph_to_replicate  Pointer to the graph, should be null if rank != 0
+   * != 0
+   */
+  ReplicatableGraph(const raft::handle_t &handle, GraphTypeView const *graph_to_replicate)
+  {
+    handle_   = &handle;
+    rank_     = handle_->get_comms().get_rank();
```
kaatish, line 691: Perhaps there should be a check to see if comms are set before calling get_comms()

cpp/include/algorithms.hpp
```diff
@@ -275,39 +276,43 @@ void force_atlas2(GraphCOOView<VT, ET, WT> &graph,
...
+ * @param[in] endpoints                     If true, include endpoints of paths in score, if false
+ * do not
+ * @param[in] weight                        If specified, device array of weights for each edge
+ * @param[in] k                             If specified, number of vertex samples defined in the
+ * vertices array.
+ * @param[in] vertices                      If specified, host array of vertex ids to estimate
+ * betweenness these vertices will serve as sources for the traversal
+ * algorihtm to obtain shortest path counters.
+ * @param[in] total_number_of_source_used   If specified use this number to normalize results
+ * when using subsampling, it allows accumulation of results across multiple calls,
```
seunghwak, line 303:
```suggestion
 * when using subsampling, it allows accumulation of results across multiple calls.
```

cpp/include/algorithms.hpp
```diff
@@ -320,37 +325,40 @@ void betweenness_centrality(raft::handle_t const &handle,
...
+ * @param[in] normalized                    If true, return normalized scores, if false return
+ * unnormalized scores.
+ * @param[in] weight                        If specified, device array of weights for each edge
+ * @param[in] k                             If specified, number of vertex samples defined in the
+ * vertices array.
+ * @param[in] vertices                      If specified, host array of vertex ids to estimate
+ * betweenness these vertices will serve as sources for the traversal
+ * algorihtm to obtain shortest path counters.
+ * @param[in] total_number_of_source_used   If specified use this number to normalize results
+ * when using subsampling, it allows accumulation of results across multiple calls,
```
seunghwak, line 350:
```suggestion
 * when using subsampling, it allows accumulation of results across multiple calls.
```

cpp/include/algorithms.hpp
```diff
@@ -587,6 +595,8 @@ void sssp(GraphCSRView<VT, ET, WT> const &graph,
  *
  * @param[in] directed               Treat the input graph as directed
  *
+ * @param[in] mg_batch               If set to true use SG BFS path when comms are initialized.
+ *
  * @throws     cugraph::logic_error when an error occurs.
```
seunghwak, line 600: This repeats 

https://github.com/rapidsai/cugraph/pull/940/files#diff-5fe5e7f6e43a9f3405cdb72bb11fa4caR572

cpp/include/algorithms.hpp
```diff
@@ -587,6 +595,8 @@ void sssp(GraphCSRView<VT, ET, WT> const &graph,
  *
  * @param[in] directed               Treat the input graph as directed
  *
+ * @param[in] mg_batch               If set to true use SG BFS path when comms are initialized.
```
seunghwak, line 598: This may go outside the scope of this PR, but I think we may better look for graph (rather than handle) to decide between SG & MG. If the graph is for single-GPU (or is not partitioned), we go for single GPU and go for multi-GPU otherwise.

With the current approach, each algorithm needs to consider three cases (SG, MG batch, MG). If the graph type dictates execution mode, each algorithm needs to consider only SG & MG (The caller needs to path SG graph for MG Batch).

cpp/include/graph.hpp
seunghwak: Ideally, we should be able to efficiently replicate DataFrames using DASK. Not sure about the reality (and if I am not mistaken, you complained about performance issue using DASK), but if the performance difference is not dramatic, relying on DASK leads to more readable and maintainable code.

cpp/include/graph.hpp
```diff
@@ -661,4 +663,179 @@ template <typename ET>
...
+  {
+    initialize_info_storage();
+    if (rank_ == 0) { fill_info_storage(); }
+    handle_->get_comms().bcast(d_buffer_.data().get(), d_buffer_.size(), 0, handle_->get_stream());
+    if (rank_ != 0) { read_info_storage(); }
+  }
+
+  virtual void initialize_info_storage()
+  {
+    size_t info_size = get_required_size_for_info();
```
seunghwak, line 713: What's the purpose of the code above instead of
```
size_t info_size = sizeof(GraphTypeView);
```
?

Are you considering that the required size can actually deviate from sizeof(GraphTypeView) in the future?

I generally advise against creating 2-3 lines functions that are used only once. This goes against readability as one needs to jump between multiple locations (if that short routine is widely re-used, it is a different story, e.g. something like std::max, std::max(a,b) is more readable than a > b ? a : b).

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -300,6 +312,11 @@ void BC<VT, ET, WT, result_t>::add_reached_endpoints_to_source_betweenness(VT so
 template <typename VT, typename ET, typename WT, typename result_t>
 void BC<VT, ET, WT, result_t>::add_vertices_dependencies_to_betweenness()
 {
+  thrust::host_vector<result_t> h_betweenness(number_of_vertices_, 0);
+  result_t *ptr = h_betweenness.data();
+  CUDA_TRY(
+    cudaMemcpy(ptr, betweenness_, number_of_vertices_ * sizeof(result_t), cudaMemcpyDeviceToHost));
```
seunghwak, line 318: Better use cudaMemcpyAsync with ```stream_``` to be stream-ready

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -369,12 +379,12 @@ template <typename VT, typename ET, typename WT, typename result_t>
 void BC<VT, ET, WT, result_t>::rescale_vertices_betweenness_centrality(result_t &rescale_factor,
```
seunghwak, line 379: Better return std::tuple<result_t, bool> instead to be more functional.

cpp/src/traversal/bfs.cu
```diff
@@ -487,7 +488,7 @@ void bfs(raft::handle_t const &handle,
   static_assert(std::is_floating_point<WT>::value,
                 "Unsupported edge weight type. Use floating point types");  // actually, this is
                                                                             // unnecessary for BFS
-  if (handle.comms_initialized()) {
+  if (handle.comms_initialized() && !mg_batch) {
     CUGRAPH_EXPECTS(sp_counters == nullptr,
                     "BFS Traversal shortest path is not supported in OPG path");
```
seunghwak, line 493: We are asked to avoid using OPG naming.

python/cugraph/dask/opg_pagerank/pagerank.py
```diff
@@ -116,17 +116,17 @@ def pagerank(input_graph,
...
+                    Comms.get_session_id(),
+                    wf[1],
+                    data.local_data,
+                    alpha,
+                    max_iter,
+                    tol,
+                    personalization,
+                    nstart,
+                    workers=[wf[0]]))
+                   for idx, wf in enumerate(data.worker_to_parts.items())])
```
seunghwak, line 129: python/cugraph/dask/opg_pagerank/pagerank.py

We're asked to avoid using "opg".

python/cugraph/tests/dask/opg_betweenness_centrality_testing.py
```diff
@@ -0,0 +1,68 @@
+import pytest
```
seunghwak, line 1: Avoid "opg" in file names.

python/cugraph/tests/dask/opg_context.py
```diff
@@ -0,0 +1,81 @@
...
+import cugraph.comms as Comms
+
+# Maximal number of verifications of the number of workers
+DEFAULT_MAX_ATTEMPT = 100
+
+# Time betweenness each attempt in seconds
+DEFAULT_WAIT_TIME = 0.5
+
+
+class OPGContext:
```
seunghwak, line 14: Avoid "OPG".

cpp/src/centrality/betweenness_centrality.cu
seunghwak: Look for https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/cudart_utils.h#L131

python/cugraph/tests/dask/opg_context.py
xcadet: Replaced `OPG` by `MG` in eab569b.

python/cugraph/tests/dask/opg_betweenness_centrality_testing.py
xcadet: Replaced in eab569b

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in eab569b

cpp/include/graph.hpp
xcadet: The function used to be longer but the final version simply used `sizeof(GraphTypeView)`, the function has been removed in eab569b

cpp/include/algorithms.hpp
xcadet: Updated in eab569b

cpp/include/algorithms.hpp
xcadet: Updated in eab569b

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in eab569b to use `cudaMemcpyAsync`

cpp/include/graph.hpp
xcadet: Added a check in eab569b

cpp/include/graph.hpp
xcadet: The current version leverage the code currently present for MG - Distributed at the Python level, it would require some further refining that could be operated in future work.

cpp/src/traversal/bfs.cu
xcadet: Should this be done in this PR for the `bfs` code or should this be update in an unifying PR?

python/cugraph/dask/opg_pagerank/pagerank.py
xcadet: Should this be done in this PR for the `pagerank` code or should this be updated in an unifying PR?

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,85 +27,325 @@ from libcpp cimport bool
...
+        graph_double.get_vertex_identifiers(<int*> c_identifier)
+    elif result_dtype == np.float32:
         graph_float.get_vertex_identifiers(<int*> c_identifier)
-    elif result_dtype == np.float64:
+    else:
+        raise ValueError("result_dtype can only be np.float64 or np.float32")
+
+    return result_df
+
+def run_regular_work(handle, normalized, endpoints, weights,
```
afender, line 176: regular_work is not an ideal name.

cpp/include/algorithms.hpp
xcadet: Updated in feb6ff4

python/cugraph/dask/opg_pagerank/pagerank.py
afender: Anyone feel free to self assign https://github.com/rapidsai/cugraph/issues/996 this needs to complete this week

cpp/include/graph.hpp
afender: Un-resolving.

I agree with Aatish and Seunghwa on avoiding adding these C++ graph classes in graph.hpp. 
We were able to reuse the existing Di/Graph classes for MG distributed graphs, we should find a way for Batch as well. 

For replication in C++, the motivation was performance because the cost of the initial version in python was prohibitive. Let's continue the discussion on slack to find the best tradeoff between performance, UX, and specific code.

python/cugraph/dask/opg_pagerank/pagerank.py
afender: I updated all the code base in my PPR PR. Make sure to sync with it and avoid adding any opg occurrence.

cpp/include/graph.hpp
xcadet: The replication is now done at the Python side, the C++ class has been removed.

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -37,8 +38,9 @@ void betweenness_centrality_impl(raft::handle_t const &handle,
                                  bool normalize,
                                  bool endpoints,
                                  WT const *weight,
-                                 VT const number_of_sources,
-                                 VT const *sources)
+                                 VT number_of_sources,
+                                 VT const *sources,
+                                 VT total_number_of_sources)
```
afender, line 43: total_number_of_sources : can't we retrieve this if everyone bcast its local sources? That would make the API nicer I think if we don't add too many batch-specific parameters

python/cugraph/structure/replication.pyx
```diff
@@ -0,0 +1,150 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.
+# Licensed under the Apache License, Version 2.0 (the "License");
```
afender, line 2: Should this be under `dask`?

python/cugraph/structure/utils.pxd
```diff
@@ -19,9 +19,16 @@
...
 
 cdef extern from "functions.hpp" namespace "cugraph":
 
     cdef unique_ptr[GraphCSR[VT,ET,WT]] coo_to_csr[VT,ET,WT](
             const GraphCOOView[VT,ET,WT] &graph) except +
 
+    cdef void comms_bcast[value_t](
+            const handle_t &handle,
+            value_t *dst,
+            size_t size) except +
```
afender, line 34: probably need an extra line break here

xcadet:
rerun tests

python/cugraph/centrality/betweenness_centrality.pxd
```diff
@@ -19,18 +19,22 @@
 from cugraph.structure.graph_new cimport *
 from libcpp cimport bool
 
+cdef extern from "raft/handle.hpp" namespace "raft":
```
Iroy30, line 22: Dont need to expose handle, it should already be cimported from graph_new.pxd

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -225,13 +226,18 @@ def edge_betweenness_centrality(
...
+        G, normalized, weight, vertices, result_dtype
     )
 
     if G.renumbered:
         df = G.unrenumber(df, "src")
         df = G.unrenumber(df, "dst")
 
+    if type(G) is cugraph.Graph:
+        lower_triangle = df['src'] >= df['dst']
+        df[["src", "dst"]][lower_triangle] = df[["dst", "src"]][lower_triangle]
```
Iroy30, line 238: Does this return the lower triangle? or do we want this to be df = df[["dst", "src"]][lower_triangle]

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,82 +27,227 @@ import cudf
 import numpy as np
 import numpy.ctypeslib as ctypeslib
 
+import dask_cudf
+import dask_cuda
+import cugraph.raft
```
Iroy30, line 32: The effort is to encapsulate raft in cugraph Comms. A raft import here can be avoided

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,82 +27,227 @@ import cudf
...
 
-def get_output_df(input_graph, result_dtype):
-    number_of_vertices = input_graph.number_of_vertices()
+
+import cugraph.comms.comms as Comms
+from cugraph.dask.common.mg_utils import get_client
+from cugraph.raft.dask.common.comms import worker_state
```
Iroy30, line 37: these can be accessed as https://github.com/rapidsai/cugraph/blob/38b032efa630ae01e29b8c73b81c34bfa3f5a905/python/cugraph/dask/link_analysis/pagerank.py#L24 instead of raft import

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,82 +27,227 @@ import cudf
...
+    return result_df
+
+def run_mg_work(input_data, normalized, endpoints,
+                weights, sources,
+                result_dtype, session_id):
+    result = None
+    # 1. Get session information
+    session_state = worker_state(session_id)
+    number_of_workers = session_state["nworkers"]
+    worker_idx = session_state["wid"]
```
Iroy30, line 177: these can be accessed as

cugraph/python/cugraph/dask/link_analysis/pagerank.py

Line 24 in 38b032e

 wid = Comms.get_worker_id(sID) 
instead of raft import

Iroy30, line 177: can be accessed as https://github.com/rapidsai/cugraph/blob/38b032efa630ae01e29b8c73b81c34bfa3f5a905/python/cugraph/dask/link_analysis/pagerank.py#L24

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,82 +27,227 @@ import cudf
...
+
+    # 2. Get handle
+    handle = session_state['handle']
```
Iroy30, line 180: https://github.com/rapidsai/cugraph/blob/38b032efa630ae01e29b8c73b81c34bfa3f5a905/python/cugraph/dask/link_analysis/pagerank.py#L25

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
```diff
@@ -26,82 +27,227 @@ import cudf
...
+def betweenness_centrality(input_graph, normalized, endpoints, weights,
+                           vertices, result_dtype):
+    """
+    Call betweenness centrality
+    """
+    df = None
+    comms = Comms.get_comms()
+    if not input_graph.adjlist:
+        input_graph.view_adj_list()
+    if comms and input_graph.mg_batch_enabled == True:
```
Iroy30, line 239: We can utilize https://github.com/rapidsai/cugraph/blob/38b032efa630ae01e29b8c73b81c34bfa3f5a905/python/cugraph/comms/comms.py#L44

python/cugraph/centrality/betweenness_centrality.py
xcadet: This swaps `dst` and `src` values when `src` >= `dst` so that we have the same number of elements in the DataFrame but we ensure that all edges are respecting `src` < `dst` before merging them together, we do not  want to get rid of any values as we will sum scores among similar edges (i.e (src, dst, score): (0, 1, 0.5) and (1, 0, 0.25) should be viewed as (0, 1, 0.5) and (0, 1, 0.25) so that we can merge the scores and obtain (0, 1, 0.75))

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 09cbb8c.

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 09cbb8c

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 09cbb8c, and added `Comms.get_n_workers(sID)`

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 09cbb8c

python/cugraph/centrality/betweenness_centrality_wrapper.pyx
xcadet: Updated in 09cbb8c

python/cugraph/centrality/betweenness_centrality.pxd
xcadet: Removed in 09cbb8c

python/cugraph/structure/utils.pxd
xcadet: Updated in 09cbb8c

xcadet: Updated in 09cbb8c

cpp/include/functions.hpp
```diff
@@ -77,4 +77,23 @@ std::unique_ptr<rmm::device_buffer> renumber_vertices(
   ET *map_size,
   rmm::mr::device_memory_resource *mr = rmm::mr::get_default_resource());
 
+/**
+ * @brief    Broadcast using handle communicator
+ *
+ * Use handle's communicator to operate broadcasting.
```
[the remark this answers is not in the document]
xcadet: Added `#include <raft/handle.hpp>` in 09cbb8c

python/cugraph/structure/replication.pyx
xcadet: Moved it to `python/cugraph/dask/structure/replication.pyx`

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in 89c95c8, C++ no longer expects `total_number_of_sources`

cpp/include/algorithms.hpp
```diff
@@ -272,35 +273,37 @@ void force_atlas2(GraphCOOView<VT, ET, WT> &graph,
...
+ * @param[in] k                             If specified, number of vertex samples defined in the
+ * vertices array.
+ * @param[in] vertices                      If specified, host array of vertex ids to estimate
+ * betweenness these vertices will serve as sources for the traversal
+ * algorihtm to obtain shortest path counters.
+ * @param[in] total_number_of_source_used   If specified use this number to normalize results
+ * when using subsampling, it allows accumulation of results across multiple calls.
  *
  */
 template <typename VT, typename ET, typename WT, typename result_t>
```
seunghwak, line 305: So now we agreed to use vertex_t, edge_t, and weight_t, it will be nice if you can replace VT, ET, and WT in the code you have modified in this PR.

cpp/include/functions.hpp
```diff
@@ -77,4 +78,23 @@ std::unique_ptr<rmm::device_buffer> renumber_vertices(
...
+ * @param[out] value          Point to the data
+ * @param[in]  count          Number of elements to broadcast
+ *
+ */
+
+template <typename value_t>
+void comms_bcast(const raft::handle_t &handle, value_t *value, size_t count)
+{
+  handle.get_comms().bcast(value, count, 0, handle.get_stream());
+}
```
seunghwak, line 99: I think we'd better avoid adding an additional layer of indirection without clear benefit.

```comms_bcast(...)``` is not really that more readable than ```handle.get_comms().bcast(...)``` And I can immediately figure out what handle.get_comms().bcast() is but if I see ```comms_bcast``` I should jump to this code to see what this actually does.

File: cpp/src/centrality/betweenness_centrality.cu [hunk 29]
```diff
@@ -72,6 +76,21 @@ void edge_betweenness_centrality_impl(raft::handle_t const &handle,
...
+  // NOTE: As of 07/2020 NetworkX does not apply rescaling based on number
+  // of sources
+  // bc.rescale_by_total_sources_used(total_number_of_sources);
+}
+template <typename VT>
+VT get_total_number_of_sources(raft::handle_t const &handle, VT local_number_of_sources)
+{
+  VT total_number_of_sources_used = local_number_of_sources;
+  if (handle.comms_initialized()) {
+    rmm::device_vector<VT> d_number_of_sources(1, local_number_of_sources);
```
seunghwak, line 88: rmm has device_scalar for this purpose (a scalar variable on device).

Look for https://github.com/rapidsai/rmm/blob/branch-0.15/include/rmm/device_scalar.hpp

You need to verify the code below, but I think you can do something like

```
rmm::device_scalar<vertex_t> d_number_of_sources(local_number_of_sources, handle.get_stream());
handle.get_comms().allreduce(d_number_of_souruces.data(), d_number_of_sources.data(), d_number_of_sources.size(), raft::comms::op_t::SUM, handle.get_stream());
totoal_number_of_sources = d_number_of_souruces.value(handle.get_stream());
```

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -404,11 +451,34 @@ void betweenness_centrality(raft::handle_t const &handle,
...
+                                        graph,
+                                        betweenness.data().get(),
+                                        normalize,
+                                        endpoints,
+                                        weight,
+                                        k,
+                                        vertices,
+                                        total_number_of_sources_used);
+    handle.get_comms().reduce(
+      betweenness.data().get(), result, betweenness.size(), raft::comms::op_t::SUM, 0, 0);
```
seunghwak, line 467: handle.get_stream() instead of 0.

cpp/src/centrality/betweenness_centrality.cu
```diff
@@ -435,17 +505,34 @@ void edge_betweenness_centrality(raft::handle_t const &handle,
...
+    detail::edge_betweenness_centrality_impl(handle,
+                                             graph,
+                                             betweenness.data().get(),
+                                             normalize,
+                                             weight,
+                                             k,
+                                             vertices,
+                                             total_number_of_sources_used);
+    handle.get_comms().reduce(
+      betweenness.data().get(), result, betweenness.size(), raft::comms::op_t::SUM, 0, 0);
```
seunghwak, line 520: handle.get_stream() instead of 0.

cpp/src/centrality/betweenness_centrality.cuh
```diff
@@ -110,8 +111,6 @@ class BC {
     nullptr;                  // array<VT>(|V|) stores the shortest path counter for the latest SSSP
   double *deltas_ = nullptr;  // array<result_t>(|V|) stores the dependencies for the latest SSSP
 
-  // FIXME: This should be replaced using RAFT handle
-  int device_id_        = 0;
   int max_grid_dim_1D_  = 0;
   int max_block_dim_1D_ = 0;
   cudaStream_t stream_;
```
seunghwak, line 116: So this stream_ is set to handle_.get_stream(), isn't it? Then, why store both handle_ and stream_? This can be confusing (need additional effort to figure out stream_ here is actually handle_.get_stream())

cpp/include/algorithms.hpp
xcadet: Updated in 38a0e27

cpp/src/centrality/betweenness_centrality.cuh
xcadet: Updated in e6c5010

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in e6c5010

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in e6c5010

cpp/src/centrality/betweenness_centrality.cu
xcadet: Updated in e7c4c4a

cpp/include/functions.hpp
seunghwak: Just add a FIXME comment saying this is a temporary solution and this function will be removed. Ideally, this should be performed using Dask scatter (if we can close the performance gap). If not, we need to provide more consistent solution to call RAFT communication collectives from python/cython.

xcadet: Added a FIXME in 40e7349

BradReesWork:
rerun tests

xcadet:
rerun tests

BradReesWork:
rerun tests

PR #941, opened 2020-06-11
[REVIEW] regression fix

Iroy30:
Fixed renumber_map accesses that were causing regression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - PR #903 Add short commit hash to conda package
 - PR #920 modify bfs test, update graph number_of_edges, update storage of transposedAdjList in Graph
 - PR #934 Update conda dev environment.yml dependencies to 0.15
+- PR #941 Regression python/cudf fix
 
 ## Bug Fixes
 - PR #936 Update Force Atlas 2 doc and wrapper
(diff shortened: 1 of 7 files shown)


GPUtester:

rlratzel:
> Did we experimentally confirm that this was the cause of the regression?

We did a before-and-after of the PR when it was WIP and saw an improvement. @Iroy30 are you able to generate a comparison report on the latest commit here?  If so, maybe we should attach it to the PR for verification and future reference?

rlratzel:
@Iroy30 I'm still looking for others, but are there any places in `link_prediction/overlap_wrapper.pyx` that need these updates?

rlratzel:
> What do you think is the best to compare against

Hmmm, this might be a lot to ask, but the _best_ comparison report would have our 0.13 numbers (fast), numbers right before your PR (regression), and the numbers with your PR in place (to compare how the PR changes get us closer to 0.13 performance).

I suppose next best would just be numbers before and after your PR, so you can at least show that there is a delta.

Iroy30:
> @Iroy30 I'm still looking for others, but are there any places in `link_prediction/overlap_wrapper.pyx` that need these updates?

I did indeed miss jaccard and overlap since they were not calling unremumber function. Fixed it. Thanks!

rlratzel:
FWIW: here's the comparison between this PR at commit 5365d6b (labeled as "NOW") and cugraph nightly build at commit ff15493:

[cugraph_WIP_PR941.txt](https://github.com/rapidsai/cugraph/files/4766954/cugraph_WIP_PR941.txt)

I'll create another comparison using the latest commit in the PR later.

PR #944, opened 2020-06-12
[REVIEW] OPG pagerank (Py)

Iroy30:


GPUtester:

cpp/include/graph.hpp
```diff
@@ -60,19 +60,32 @@ class GraphViewBase {
...
+  VT local_offset;
+
   /**
    * @brief      Fill the identifiers array with the vertex identifiers.
    *
    * @param[out]    identifier      Pointer to device memory to store the vertex
    * identifiers
    */
   void get_vertex_identifiers(VT *identifiers) const;
+  void set_local_data(VT number_of_local_vertices_, ET number_of_local_edges_, VT local_offset_)
```
afender, line 74: can we pass and store arrays with this info for all ranks?

`void set_local_data(VT* number_of_local_vertices_, ET* number_of_local_edges_, VT* local_offset_)
`
so that we can access `number_of_local_vertices_[rank_i] ` from everywhere.

afender: We need this for communications like allgatherv

Iroy30: Yes. Working on that

Iroy30: Updated

cpp/src/link_analysis/pagerank.cu
```diff
@@ -382,3 +382,39 @@ template void pagerank<int, int, double>(experimental::GraphCSCView<int, int, do
                                          bool has_guess);
 
 }  // namespace cugraph
+
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+namespace cugraph {
+
+template <typename VT, typename ET, typename WT>
+void mg_pagerank_temp(raft::handle_t &handle,
```
afender, line 391: Let's make this match the C++ API in PR #947 so that it is compatible from the start.

Iroy30:
rerun tests

cpp/src/link_analysis/pagerank.cu
Iroy30: updated

python/cugraph/opg/link_analysis/mg_pagerank_wrapper.pyx
```diff
@@ -0,0 +1,79 @@
...
+
+from cugraph.structure.utils_wrapper import *
+from cugraph.opg.link_analysis cimport mg_pagerank as c_pagerank
+import cudf
+from cugraph.structure.graph_new cimport *
+import cugraph.structure.graph_new_wrapper as graph_new_wrapper
+from libc.stdint cimport uintptr_t
+from cython.operator cimport dereference as deref
+
+def mg_pagerank(input_df, local_data, handle, alpha=0.85, max_iter=100, tol=1.0e-5):
```
afender, line 25: We don't support `tol` currently, we should throw an error if users set it (or add it later to the API).

python/cugraph/dask/pagerank/pagerank.py
```diff
@@ -162,7 +162,6 @@ def pagerank(edge_list, alpha=0.85, max_iter=30):
     >>>                                    dtype=['int32', 'int32'])
     >>> pr = dcg.pagerank(ddf_edge_list, alpha=0.85, max_iter=50)
     """
-
```
afender, line 165: Is there some dead code from SNMG or duplicate from RAFT in this file?

python/cugraph/opg/link_analysis/mg_pagerank_wrapper.pyx
afender, line 25: https://github.com/rapidsai/cugraph/blob/aeee8aae1f72fdb823c235ae57f927dd4be4cb47/python/cugraph/snmg/link_analysis/mg_pagerank.py#L15

Should we get rid of this? It could be confusing since they are called the same.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -350,6 +351,13 @@ void pagerank(experimental::GraphCSCView<VT, ET, WT> const &graph,
 {
   CUGRAPH_EXPECTS(pagerank != nullptr, "Invalid API parameter: Pagerank array should be of size V");
 
+  if (handle.comms_initialized()) {
```
kaatish, line 354: Is this a debugging artifact?
@afender do you need the pagerank code to be a no-op when it is in a multi gpu environment?
And if so, do you need some sort of a print in that case?

python/cugraph/tests/dask/opg_pagerank_testing.py
```diff
@@ -0,0 +1,47 @@
...
+
+import cugraph.dask as dcg
+from dask.distributed import Client
+import gc
+import cugraph
+import dask_cudf
+
+# Move to conftest
+from dask_cuda import LocalCUDACluster
+# cluster = LocalCUDACluster(protocol="tcp", scheduler_port=0)
```
kaatish, line 22: Should this comment be removed from here?

python/cugraph/dask/opg_pagerank/pagerank.py
```diff
@@ -0,0 +1,49 @@
...
+
+from cugraph.raft.dask.common.comms import Comms
+from cugraph.dask.common.input_utils import DistributedDataHandler
+from dask.distributed import wait, default_client
+from cugraph.raft.dask.common.comms import worker_state
+from cugraph.opg.link_analysis import mg_pagerank_wrapper as mg_pagerank
+
+
+def common_func(sID, data, local_data):
+    print("Dataframe: ", data)
```
kaatish, line 24: Are all of these debugging artifacts? Is it required to print all of this?

python/cugraph/dask/pagerank/pagerank.py
Iroy30: This is entirely snmg code. It simply connects to cugraph/python/cugraph/snmg/link_analysis/mg_pagerank.py which throws an exception saying it is disabled. We could update it to throw exception at python/cugraph/dask/pagerank/pagerank.py or just remove it completely for 0.15 release once we have OPG pagerank.

python/cugraph/opg/link_analysis/mg_pagerank_wrapper.pyx
Iroy30: https://github.com/rapidsai/cugraph/pull/944#discussion_r441028764, It is to let the user know that snmg is disabled. We could remove it once OPG is in place, or, move the exception throwing at python/cugraph/dask/pagerank/pagerank.py

Iroy30: I think we could add it to the final PR after #947 is merged, along with opg pagerank tests.

cpp/src/link_analysis/pagerank.cu
Iroy30: This is a place holder for updates in #947 . The prints are going to be replaced by opg code

afender: That's ok for this PR, I have the upgrade in #947 that will replace this function.

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: This is a temporary test (which is not run by CI) This is to check the validity of the PR if someone wants to test the pipeline. This is going to be replaced by a pagerank test after #947 is merged and e2e runs successfully.

python/cugraph/dask/opg_pagerank/pagerank.py
Iroy30: This going to be updated/removed in the final PR. Currently these prints exist for checking input validity as we can't verify results pagerank without Opg cuda code.

PR #945, opened 2020-06-12
[REVIEW] [ENH] Simplified benchmark --no-rmm-reinit option, updated default options

rlratzel:
Simplified --no-rmm-reinit to use existing markers instead of modifying fixture params directly (this also means all runs always have the same number of params, which helps reporting). Also updated ini file to only show common columns in report by default and to remove -x from default options to work better with nightly runs.


GPUtester:

rlratzel:
rerun tests

rlratzel:
rerun tests

Issue #485, opened 2019-09-11
[FEA]MG Pagerank


afender:
The goal is to start from the current implementation (one process running on multiple GPUs) and transform it to run with one process per GPU. 

The current version was designed to take advantage of DGX architectures. The new solution should work on cloud-based environments and non-nvlink systems. 

We should bring OPG to cuGraph though generalizing and sharing the comm libs with cuML. That involves moving a generic version of cuML's handle and device allocator to RAFT and cuML's communicator to a separate comm lib. The comm part can't be in RAFT as it is designed to allow the user to swap out the implementation at runtime (so it is not headers only). We will then add these new dependencies in cuGraph and build OPG analytics on top of it. 

**C++/CUDA** 
We need to move from shared memory, multithreaded code to multi-process code at the CPP/CUDA level. 
- [ ] COO2CSR
- [ ] Degree
- [ ] Pagerank setup
- [ ] Power iteratinon
- [ ] CAPI
- [ ] Tests

**Dask/Python**
Need to move from the IPC based solution to the recent solution deployed in cuML.
- [ ] python API backend upgrade
- [ ] Tests

**Update Doc**

**Verify Notebooks**

Issue #815, opened 2020-04-06
[FEA] C++ cuGraph library handle


afender:
There are many reasons why we may want a handle :
- communicator
- Keep track of library wide events (status)
- Counters (devices, streams, calls, … )
- Stream(s)
- Allocator
- Cuda library handles (expensive to create)

We managed to postpone this so far.
It becomes pretty much required in the OPG paradigm though. We need to pass the communicator so that we have a constant view of the multi-GPU / multi-process context between calls

afender:
the handle is now in RAFT : https://github.com/rapidsai/raft/pull/3

afender:
The graph class has been updated and multi-GPU functions accept the handle.
We now need a consistent solution so that the handle content can be accessed from every feature.

Some of the challenges are:
- avoid undefined behavior when the graph has a different handle than the algo
- prevent recreating handle(s) as this can be costly

This task unblock stream support #957

PR #947, opened 2020-06-12
[REVIEW] OPG pagerank (CUDA)

### LINKED ISSUE #816: [FEA] Productize OPG CSR pagerank
Start from the C++ 1D proto that accepts CSR matrices (#812), degrees (#810), and communicator (#814 which should really be passed through #815 ).
This proto is is fairly compatible with the cugraph opg infra as defined in #811.
The code should be cleaned up and tested on real graphs before we can say it is production-ready.
 
This is the core solver for #485

afender:
1D OPG pagerank checkpoint for the CUDA part
- accepts a regular CSC graph locally with partitioning offsets from #944 
- gets comm info from RAFT handle
- uses cusparse spmv

There are still several things to do for OPG PageRank : 
- More tests and analysis
- Get the result back to dask cudf 
- Library-wise handle strategy #815  
- Upgrade to raft cusparse wrapper once upgrades for cuda 11 are completed

Close #816, progress on #485


GPUtester:

cpp/tests/pagerank/pagerank_test.cpp
```diff
@@ -129,6 +130,7 @@ class Tests_Pagerank : public ::testing::TestWithParam<Pagerank_Usecase> {
     ASSERT_EQ(fclose(fpin), 0);
 
     //  Pagerank runs on CSC, so feed COOtoCSR the row/col backwards.
+    raft::handle_t handle;
```
Iroy30, line 133: Do we need this

afender: Yea, PageRank asks for a `raft::handle_t const &handle`. Can't pass null there.

Iroy30: then we should probably change line 147 and 154 to handle instead of G.handle[0]

afender:
rerun tests

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+  v_loc          = G.local_vertices[comm.get_rank()];
+  e_loc          = G.local_edges[comm.get_rank()];
+  part_off       = G.local_offsets;
+  local_vertices = G.local_vertices;
+  off            = G.offsets;
+  ind            = G.indices;
+  sm_count       = handle_.get_device_properties().multiProcessorCount;
+
+  is_setup = false;
+  bookmark.resize(v_glob);
```
ChuckHastings, line 52: If you made these part of the initializer list then you could create bookmark and val to the correct size directly rather than initializing them and resizing them.

cpp/src/link_analysis/pagerank_1D.cuh
```diff
@@ -0,0 +1,105 @@
...
+#include "utilities/error_utils.h"
+#include "utilities/spmv_1D.cuh"
+
+namespace cugraph {
+namespace opg {
+
+template <typename VT, typename ET, typename WT>
+class Pagerank {
+ private:
+  size_t v_glob;                     // global number of vertices
```
ChuckHastings, line 34: Is there a reason to prefer size_t here?  With the object templated, I believe the following would work and be a little cleaner:
```
  VT v_glob;
  VT v_loc;
  ET e_loc;
```

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+#include "pagerank_1D.cuh"
+#include "utilities/graph_utils.cuh"
+
+namespace cugraph {
+namespace opg {
+
+#define CUDA_MAX_KERNEL_THREADS 256
+#define CUDA_MAX_BLOCKS 65535
+
+template <typename VT, typename ET, typename WT>
```
ChuckHastings, line 29: I would leave off typename ET, since nothing passed is of that type.  No point compiling multiple variations of an identical function.

cpp/include/algorithms.hpp
```diff
@@ -59,18 +62,18 @@ namespace cugraph {
  * @param[in] tolerance              Set the tolerance the approximation, this parameter should be a
```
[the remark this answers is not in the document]
seunghwak: OPG PageRank should support tolerance, why not?

cpp/src/link_analysis/pagerank.cu
```diff
@@ -350,23 +351,29 @@ void pagerank(raft::handle_t const &handle,
...
-                                           personalization_values,
-                                           alpha,
-                                           tolerance,
-                                           max_iter,
-                                           has_guess);
+    CUGRAPH_EXPECTS(personalization_subset == nullptr,
+                    "Invalid API parameter: Multi-GPU Pagerank does not support Personalized "
+                    "variant, please use the single GPU version for this feature");
+    CUGRAPH_EXPECTS(
+      tolerance > 1e-6 && tolerance < 1e-4,
```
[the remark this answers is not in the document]
seunghwak: +1, if tolerance is 0, it will never triggered.

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+#include <graph.hpp>
+#include "pagerank_1D.cuh"
+#include "utilities/graph_utils.cuh"
+
+namespace cugraph {
+namespace opg {
+
+#define CUDA_MAX_KERNEL_THREADS 256
+#define CUDA_MAX_BLOCKS 65535
+
```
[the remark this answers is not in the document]
seunghwak: This should be queried from RAFT handle.

RAFT currently has two different places for querying device properties (https://github.com/rapidsai/raft/issues/24), and this inconsistency needs to be resolved, but we'd better not make RAPIDS codebase even more inconsistent.

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+template <typename VT, typename ET, typename WT>
+__global__ void __launch_bounds__(CUDA_MAX_KERNEL_THREADS)
+  transition_kernel(const size_t e, const VT *ind, const VT *degree, WT *val)
+{
+  for (auto i = threadIdx.x + blockIdx.x * blockDim.x; i < e; i += gridDim.x * blockDim.x)
+    val[i] = 1.0 / degree[ind[i]];
+}
```
seunghwak, line 35: What happens if degree[ind[i]] == 0? Doesn't matter as this will not be used?

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+
+template <typename VT, typename ET, typename WT>
+Pagerank<VT, ET, WT>::Pagerank(const raft::handle_t &handle_,
+                               experimental::GraphCSCView<VT, ET, WT> const &G)
+  : comm(handle_.get_comms())
```
seunghwak, line 40: Eventually we need to adopt RAFT's clang-tidy.

https://github.com/rapidsai/raft/blob/3038c773636540941b8bf9e30c487f20bddc3cdc/cpp/.clang-tidy#L185

Class private member variables should have a suffix "_"

Better adopt this convention in new code.

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+  off            = G.offsets;
+  ind            = G.indices;
+  sm_count       = handle_.get_device_properties().multiProcessorCount;
+
+  is_setup = false;
+  bookmark.resize(v_glob);
+  val.resize(e_loc);
+
+  // intialize cusparse. This can take some time.
+  cugraph::detail::Cusparse::get_handle();
```
seunghwak, line 56: Can't we use RAFT cuSparse wrapper instead? (https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/sparse/cusparse_wrappers.h)

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+template <typename VT, typename ET, typename WT>
+Pagerank<VT, ET, WT>::~Pagerank()
+{
+}
+
+template <typename VT, typename ET, typename WT>
+void Pagerank<VT, ET, WT>::transition_vals(const VT *degree)
+{
+  int threads = min(static_cast<VT>(e_loc), 256);
+  int blocks  = min(static_cast<VT>(32 * sm_count), CUDA_MAX_BLOCKS);
```
seunghwak, line 68: What are 256 and 32 here? If 32 is the size of a warp, we'd better use https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/cudart_utils.h#L120 instead of directly writing 32.

seunghwak: And what's the point of static_cast<VT> here as blocks is int?

And is "min" here std::min? Better use std::min than creating our own.

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+  transition_kernel<VT, ET, WT><<<blocks, threads>>>(e_loc, ind, degree, val.data().get());
+  CUDA_CHECK_LAST();
```
seunghwak, line 70: Aren't we using stream?

And better use https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/cudart_utils.h#L102

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+}
+
+template <typename VT, typename ET, typename WT>
+void Pagerank<VT, ET, WT>::flag_leafs(const VT *degree)
+{
+  int threads = min(static_cast<VT>(v_glob), 256);
+  int blocks  = min(static_cast<VT>(32 * sm_count), CUDA_MAX_BLOCKS);
+  cugraph::detail::flag_leafs_kernel<VT, WT>
+    <<<blocks, threads>>>(v_glob, degree, bookmark.data().get());
```
seunghwak, line 79: Need to be stream ready.

cpp/src/link_analysis/pagerank_1D.cuh
seunghwak:
```
VT v_glob{};
VT v_loc{};
ET e_loc{};
```
will be even better.

cpp/src/utilities/spmv_1D.cu
```diff
@@ -0,0 +1,73 @@
...
+                               WT *val_,
+                               WT *x)
+  : comm(comm_),
+    local_vertices(local_vertices_),
+    part_off(part_off_),
+    off(off_),
+    ind(ind_),
+    val(val_)
+{
+  stream = nullptr;
```
seunghwak, line 36: Shouldn't we take stream from raft handle?

cpp/src/utilities/spmv_1D.cu
```diff
@@ -0,0 +1,73 @@
...
+  i      = comm.get_rank();
+  p      = comm.get_size();
+  v_glob = part_off[p - 1] + local_vertices[p - 1];
+  v_loc  = local_vertices[i];
+  VT tmp;
+  cudaMemcpy(&tmp, &off[v_loc], sizeof(VT), cudaMemcpyDeviceToHost);
```
seunghwak, line 42: Guard this with CUDA_TRY

cpp/src/utilities/spmv_1D.cu
```diff
@@ -0,0 +1,73 @@
...
+  CUDA_CHECK_LAST();
```
seunghwak, line 43: No need for this if we guard the above with CUDA_TRY

cpp/src/utilities/spmv_1D.cuh
```diff
@@ -0,0 +1,67 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <rmm/thrust_rmm_allocator.h>
+#include <raft/handle.hpp>
+#include "utilities/cusparse_helper.h"
```
seunghwak, line 20: Better use RAFT cuSparse wrapper.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -350,23 +351,29 @@ void pagerank(raft::handle_t const &handle,
...
-                                           pagerank,
-                                           personalization_subset_size,
-                                           personalization_subset,
-                                           personalization_values,
-                                           alpha,
-                                           tolerance,
-                                           max_iter,
-                                           has_guess);
+    CUGRAPH_EXPECTS(personalization_subset == nullptr,
+                    "Invalid API parameter: Multi-GPU Pagerank does not support Personalized "
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/src/link_analysis/pagerank_1D.cu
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+
+template <typename VT, typename ET, typename WT>
+Pagerank<VT, ET, WT>::~Pagerank()
+{
+}
+
+template <typename VT, typename ET, typename WT>
+void Pagerank<VT, ET, WT>::transition_vals(const VT *degree)
+{
+  int threads = min(static_cast<VT>(e_loc), 256);
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -0,0 +1,137 @@
...
+  int blocks  = min(static_cast<VT>(32 * sm_count), CUDA_MAX_BLOCKS);
+  transition_kernel<VT, ET, WT><<<blocks, threads>>>(e_loc, ind, degree, val.data().get());
+  CUDA_CHECK_LAST();
+}
+
+template <typename VT, typename ET, typename WT>
+void Pagerank<VT, ET, WT>::flag_leafs(const VT *degree)
+{
+  int threads = min(static_cast<VT>(v_glob), 256);
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/include/algorithms.hpp
```diff
@@ -59,18 +62,18 @@ namespace cugraph {
...
  *                                   Setting too small a tolerance can lead to non-convergence due
  to numerical roundoff. Usually values between 0.01 and 0.00001 are acceptable.
  * @param[in] max_iter               (optional) The maximum number of iterations before an answer is
  returned. This can be used to limit the execution time and do an early exit before the solver
  reaches the convergence tolerance.
- *                                   If this value is lower or equal to 0 cuGRAPH will use the
+ *                                   If this value is lower or equal to 0 cuGraph will use the
  default value, which is 500.
- * @param[in] has_guess              (optional) This parameter is used to notify cuGRAPH if it
+ * @param[in] has_guess              (optional) This parameter is used to notify cuGraph if it
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
afender: > This is an oddly specific way of checking tolerance.

Indeed, that's because the default value for a single GPU is 1e-5. This the value we are trying to identify here. There are a few other ways to do that but none is really concise.

> Specifying a tolerance of 0.01 or 0.00000001 will not trigger the error.
I think I am missing something. 0.01 > 1e-4 and  0.00000001 < 1e-6 they should both trigger the error.

> Would tolerance == WT{0} be sufficient?
I don't think so because it is not expected to be 0.

cpp/src/link_analysis/pagerank.cu
```diff
@@ -350,23 +351,29 @@ void pagerank(raft::handle_t const &handle,
...
-  return detail::pagerank_impl<VT, ET, WT>(graph,
-                                           pagerank,
-                                           personalization_subset_size,
-                                           personalization_subset,
-                                           personalization_values,
-                                           alpha,
-                                           tolerance,
-                                           max_iter,
-                                           has_guess);
+    CUGRAPH_EXPECTS(personalization_subset == nullptr,
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
afender: > OPG PageRank should support tolerance, why not?

Yes, planning on adding it in the next Pagerank PR or PPR

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
seunghwak: So are we supporting tolerance but only allowing the value between [1e-6, 1e-4]? This sounds like an odd requirement? Why do we need this?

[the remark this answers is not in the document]
afender: The default value for a single GPU is 1e-5

cpp/src/link_analysis/pagerank.cu
```diff
@@ -350,23 +351,29 @@ void pagerank(raft::handle_t const &handle,
...
+                    "Invalid API parameter: Multi-GPU Pagerank does not support Personalized "
+                    "variant, please use the single GPU version for this feature");
+    CUGRAPH_EXPECTS(
+      tolerance > 1e-6 && tolerance < 1e-4,
+      "Invalid API parameter: Multi-GPU Pagerank does not support tolerance, please use "
+      "the number of iteration as stopping criteria");
+    CUGRAPH_EXPECTS(has_guess == false,
+                    "Invalid API parameter: Multi-GPU Pagerank does not guess, please use the "
+                    "single GPU version for this feature");
+    return cugraph::opg::pagerank<VT, ET, WT>(handle, graph, pagerank, alpha, max_iter);
```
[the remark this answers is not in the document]
afender: 5f7dd4c

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
seunghwak: I get the default value, but we don't allow users to override default value?

cpp/src/link_analysis/pagerank_1D.cu
kaatish: You should probably use bookmark flag set in flag_leaf to avoid calculation when degree[ind] is 0.

cpp/src/utilities/spmv_1D.cuh
afender: Yup, that's what the code comment on the next line explains.

```
// FIXME : use RAFT cusparse wrapper
// The wrapper is about to be upgraded in RAFT
// better wait for it to be stable before building on top of it
// #include <raft/sparse/cusparse_wrappers.h>
```

cpp/src/link_analysis/pagerank_1D.cu
seunghwak: Isn't that require additional memory for bookmark? We'd better avoid that. Here the performance will be likely to be limited by getting degree[ind[i]] (random access) and having a predicated instruction (to check degree is 0) will have only very minor impact on performance (or can be faster by avoiding additional memory access to get bookmark). Better avoid unnecessarily increasing the memory footprint.

cpp/src/utilities/spmv_1D.cu
afender: It is just a place holder currently. But you are correct it should come from the raft handle if cuGraph was really supporting streams. We can definitely look at stream support and priorities at the library level in another issue.

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
afender: Users are allowed to override tolerance in SG path but not MG.
Any suggestions on how to better detect that tolerance was not overridden by the user in the MG path?
This is likely just for a few weeks as we will likely support tolerance soon.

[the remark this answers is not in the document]
ChuckHastings: I'm confused.  The error message is indicates that OPG pagerank does not support tolerance.  This check, as written, will throw an exception if a tolerance of 1e-5 is passed in.

If you pass in a tolerance of 0.01 then it will check:
* 0.01 > 1e-6: TRUE
* 0.01 < 1e-4: FALSE
* TRUE && FALSE == FALSE  ... thus the exception will not be thrown.  Right?

cpp/src/utilities/spmv_1D.cuh
afender: 2d430fc

cpp/src/utilities/spmv_1D.cu
afender: 2d430fc

cpp/src/utilities/spmv_1D.cu
afender: 2d430fc

cpp/include/algorithms.hpp
```diff
@@ -38,7 +39,9 @@ namespace cugraph {
  32-bit)
  * @tparam WT                        Type of edge weights. Supported value : float or double.
  *
- * @param[in] graph                  cuGRAPH graph descriptor, should contain the connectivity
+ * @param[in] handle                 cuGraph library handle. If a communicator is set in the handle,
```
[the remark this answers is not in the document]
afender: 2d430fc

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 ## New Features
 - PR #937 Add wrapper for gunrock HITS algorithm
 - PR #939 Updated Notebooks to include new features and benchmarks
-- PR #944 opg pagerank dask
+- PR #944 opg pagerank (dask)
+- PR #947 OPG pagerank (CUDA)
```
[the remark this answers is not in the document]
afender: 2d430fc

cpp/src/link_analysis/pagerank_1D.cu
afender: 2d430fc

cpp/src/link_analysis/pagerank_1D.cuh
afender: 2d430fc

cpp/src/link_analysis/pagerank_1D.cu
afender: 2d430fc

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
seunghwak: > Users are allowed to override tolerance in SG path but not MG.

Why? To avoid reduction? We should support this for both SG and MG.

cpp/src/utilities/spmv_1D.cuh
afender: Just waiting on RAFT cuSparse wrapper to be stable, there are likely two other PR coming before we close #485.

cpp/src/link_analysis/pagerank_1D.cu
afender: https://github.com/rapidsai/cugraph/pull/947#discussion_r442907236

cpp/src/link_analysis/pagerank_1D.cu
afender: https://github.com/rapidsai/cugraph/pull/947#discussion_r442922727

cpp/src/link_analysis/pagerank_1D.cu
afender: https://github.com/rapidsai/cugraph/pull/947#discussion_r442922727

cpp/src/link_analysis/pagerank_1D.cu
afender: >  If 32 is the size of a warp

It is not. Just an empirical value that gave good perf in the past. I might change this again after analyzing perf.

> And what's the point of static_cast here as blocks is int?

Legacy of size_t to VT, fixed

> Better use std::min than creating our own.

Switched to std::min

cpp/src/link_analysis/pagerank_1D.cu
afender: Do you know of a tool that does this on the code base like for clang?

cpp/src/link_analysis/pagerank_1D.cu
afender: > What happens if degree[ind[i]] == 0? 

This can't happen mathematically.

cpp/src/link_analysis/pagerank_1D.cu
seunghwak: clang-tidy does this. This isn't set-up for cuGraph but for RAFT, python3 cpp/scripts/run-clang-tidy.py -cdb cpp/biuld/compile_commands.json does this.

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
afender: > > Users are allowed to override tolerance in SG path but not MG.

> Why? To avoid reduction? We should support this for both SG and MG.

https://github.com/rapidsai/cugraph/pull/947#discussion_r442914612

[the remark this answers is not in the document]
afender: So let me add some context : 
- Same API entry point for SG and OPG
- SG supports tolerance but OPG doesn't currently (but should soon)
- SG uses 1e-5 as default tolerance.

Temporarily, the idea was to throw an error in the MG path when tolerance was not 1e-5 (ie. when we can assume it was set by the user) but I'm open to other suggestions.

Notice that this only impacts C++ API users and will go away before a few weeks.

cpp/src/link_analysis/pagerank_1D.cu
afender: ed2f6fe

cpp/src/link_analysis/pagerank_1D.cu
afender: Degree contains `IN` degree. So all `degree[ind[i]]` where incremented by definition.

cpp/src/link_analysis/pagerank_1D.cu
afender: Let's proceed as we did for clang support then and port everything at once. 
Once this is done and checks and tools are enabled, new PRs will be requested to comply with tidy.

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
ChuckHastings: If it's going away that soon, why even have the check?  But if it's going away soon, I'm fine either way.

cpp/src/link_analysis/pagerank_1D.cu
afender: https://github.com/rapidsai/cugraph/issues/958

cpp/src/utilities/spmv_1D.cu
afender: https://github.com/rapidsai/cugraph/issues/957

cpp/src/link_analysis/pagerank_1D.cu
seunghwak: Could you add a comment here? This is non-trivial to figure out, so comment will be helpful if others need to maintain this code.

cpp/src/link_analysis/pagerank.cu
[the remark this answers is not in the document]
afender: Ok, removed

cpp/src/link_analysis/pagerank_1D.cu
afender: Added a comment

cpp/src/utilities/spmv_1D.cu
seunghwak: Yeah... but should at least be consistent in handling stream.

stream is defined as the member variable of PageRank class below
https://github.com/rapidsai/cugraph/pull/947/files#diff-1ca4da61cab77812b7b10fe0adac80f9R41

and here stream is defined here as a local variable.

Some kernels specify stream and some kernels do not specify stream.

Fixing this really requires reading and analyzing the whole code again to properly use stream from RAFT handle. If we are consistent, just searching for stream, and replacing stream with handle.get_stream() will fix the problem.

If we are not ready to take RAFT handle as input parameter, should we at least be consistent in exposing stream?

PR #950, opened 2020-06-16
[WIP][skip-ci] OpenCypher bindings

jwyles:
This adds python bindings wrapping the db_object class. It depends on commits in PR 861.


GPUtester:

python/setup.py
```diff
@@ -100,6 +100,7 @@ def run(self):
               sources=CYTHON_FILES,
               include_dirs=[conda_include_dir,
                             '../cpp/include',
+                            '../cpp/src',
```
afender, line 103: Better encapsulate the public C++ API in a public header and move it to `../cpp/include`.
Files under `../cpp/src` are internals and don't get deployed.

PR #952, opened 2020-06-17
[REVIEW] [ENH] Updated get_test_data.sh to also (optionally) download and install datasets for benchmark runs

rlratzel:
Updated `get_test_data.sh` to also (optionally) download and install datasets for benchmark runs, and updated the corresponding README.md to describe it.

Also did some minor refactoring on `get_test_data.sh` to make it easier to add new datasets like this in the future.

### DIFF
--- a/datasets/get_test_data.sh
+++ b/datasets/get_test_data.sh
@@ -1,14 +1,6 @@
 #!/bin/bash
 set -e
 set -o pipefail
-NUMARGS=$#
-ARGS=$*
-
-# FIXME: consider using getopts for option parsing
-# Arg parsing function
-function hasArg {
-    (( ${NUMARGS} != 0 )) && (echo " ${ARGS} " | grep -q " $1 ")
-}
 
 # Update this to add/remove/change a dataset, using the following format:
 #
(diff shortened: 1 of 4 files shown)


GPUtester:

datasets/get_test_data.sh
```diff
@@ -44,19 +39,41 @@ https://s3.us-east-2.amazonaws.com/rapidsai-data/cugraph/benchmark/hibench/hiben
 benchmark
 "
 
+BENCHMARK_DATASET_DATA="
+# ~90s download - these are used for benchmarks runs (code in <cugraph root>/benchmarks)
+https://rapidsai-data.s3.us-east-2.amazonaws.com/cugraph/benchmark/benchmark_csv_data.tgz
```
afender, line 44: Do we keep the list of what's in this archive? 
We will soon add benchmark datasets, not sure we will rebuild the whole archive for this (we could just add new links). It would be great to keep track of what's in there.

afender: We could think of something similar to this section for python testing datasets https://github.com/rapidsai/cugraph/tree/branch-0.15/datasets#python

rlratzel: @afender I agree since I think there's probably some additional cleanup we could do to all the datasets to reorg them so they make better sense, have less overlap, etc. For instance, right now there's a folder created for testing datasets called "benchmarks", which is confusing.
I'll take a pass at updating the doc you linked as a start.

rlratzel:
rerun tests

PR #953, opened 2020-06-18
[REVIEW] fix setting RAFT_DIR from the RAFT_PATH env var

trxcllnt:
Fixes setting the RAFT_PATH env var for source builds.

Without this fix the path `-I../../../../ENV{RAFT_PATH}/cpp/include` is included in the C++ build args instead of the correct path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - PR #941 Regression python/cudf fix
 - PR #945 Simplified benchmark --no-rmm-reinit option, updated default options
 - PR #946 Install meta packages for dependencies
+- PR #953 fix setting RAFT_DIR from the RAFT_PATH env var
 
 ## Bug Fixes
 - PR #936 Update Force Atlas 2 doc and wrapper
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -263,7 +263,7 @@ endif(NOT NCCL_PATH)
 if(DEFINED ENV{RAFT_PATH})
   message(STATUS "RAFT_PATH environment variable detected.")
   message(STATUS "RAFT_DIR set to $ENV{RAFT_PATH}")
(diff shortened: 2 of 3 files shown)


GPUtester:

trxcllnt:
rerun tests

trxcllnt:
rerun tests

trxcllnt:
@afender when source-building `raft` and setting `RAFT_PATH`, I'm getting this error when I try to import cugraph in python:
```
ModuleNotFoundError: No module named 'cugraph.raft'
```

I notice here it's trying to symlink `../$RAFT_PATH/python/raft` to `cugraph/python/cugraph/raft`, but that doesn't seem right when `$RAFT_PATH` is provided as an environment variable:
https://github.com/rapidsai/cugraph/blob/e630ffb768af8af95d189ba5775ce6dad38476a2/python/setuputils.py#L99-L103

Mind if I push a fix for this issue also?

afender:
> @afender when source-building `raft` and setting `RAFT_PATH`, I'm getting this error when I try to import cugraph in python:
> 
> ```
> ModuleNotFoundError: No module named 'cugraph.raft'
> ```
> 
> I notice here it's trying to symlink `../$RAFT_PATH/python/raft` to `cugraph/python/cugraph/raft`, but that doesn't seem right when `$RAFT_PATH` is provided as an environment variable:
> https://github.com/rapidsai/cugraph/blob/e630ffb768af8af95d189ba5775ce6dad38476a2/python/setuputils.py#L99-L103
> 

Indeed, thanks for catching that!
+@Iroy30, looking at https://github.com/rapidsai/cuml/blob/branch-0.15/python/setuputils.py, it seems we got a few things out of sync. 

> Mind if I push a fix for this issue also?

That would be great if that's not too much trouble. Otherwise, we will address it as part of the next RAFT related PR. Thanks!!

afender:
Thank you, @trxcllnt !

Issue #955, opened 2020-06-18
[ENH] Large graph support in BFS/SSSP


pgera:
**Describe the solution you'd like**
Currently, BFS and SSSP are specialized for int types for number of vertices and edges. This places a limitation on the graph size. Ideally, the types for |V| and |E| should be flexible, but as a first step, adding uniform support for other types would get us most of the way. The main types of interest are unsigned (32 bits) and long (signed 64 bits). uint64_t can also be added, but doesn't buy us much. Several large graphs like com-friendster, uk-2007-05, etc. can be implemented with unsigned since the number of edges and vertices is less than max_uint32, but greater than max_int.

**Current Status**
I have a WIP implementation that adds these features. Other than some straightforward changes, the main change is that current BFS/SSSP kernels use -1 as an invalid marker. This does not work for unsigned types (since -1 is max unsigned). I have changed all such instances to use a bitmask for validity rather than using -1 as an invalid value.

**What works**
Top-Down BFS for unsigned, int, and long. (SSSP should be similar).

**What doesn't work**
Bottom-Up BFS for large graphs, either unsigned or long.  That is, it produces wrong results. This may be a separate issue in the bottom up implementation that is unrelated to signed v/s unsigned since it also fails for long (signed 64). The bottom up code is quite complicated, and I haven't had a chance to debug it.

**Preliminary Performance Metrics**

BFS Top Down with unsigned type on GV100
com-friendster (65.6M nodes, 3.61B edges): 300 ms
uk-2007-05 (105.2M nodes, 3.74B edges): 87 ms

I'll create a PR with my changes soon.

BradReesWork:
Please create a PR and we will do a code review.  

We are aware of the limitations of using int32 and do have on-going work to address it, especially as we work towards multi-gpu support and trillion edge scale graphs.  

Please see PR:  https://github.com/rapidsai/cugraph/pull/838

pgera:
The bottom-up issue is likely separate. I'll create a separate issue for that. I am seeing some issues even with the current branch.

BradReesWork:
the new MNMG BFS work should address this

afender:
Related to #902 and #1177. More optimizations coming for MG BFS and larger inputs in 0.17.

pgera:
Hi cugraph folks! I published code for work on GPU based compressed graph traversals that I did some time ago at https://github.com/pgera/efg . Feel free to use/adapt it any way. There's more documentation in my thesis and a paper is in the pipeline.

PR #959, opened 2020-06-19
[Review] Add support for uint32_t and int64_t types for BFS (cpp side)

pgera:
Progress towards https://github.com/rapidsai/cugraph/issues/955


GPUtester:

seunghwak:
Need to add a test case for unsigned and long types in https://github.com/rapidsai/cugraph/blob/branch-0.15/cpp/tests/traversal/bfs_test.cu

pgera:
Yes, I'll add that. I have been testing the kernels independently with other large graphs that I have.

pgera:
Btw, I also cleaned up some code in sm_utils. There were a few unused shfl wrappers, and they all had the old non-syn variants for old cuda archs. I'm assuming that cuda arch older than 300 is anyway not supported by cugraph. Those can be reverted/added back if they are needed.

seunghwak:
> Btw, I also cleaned up some code in sm_utils. There were a few unused shfl wrappers, and they all had the old non-sync variants for old cuda archs. I'm assuming that cuda arch older than 300 is anyway not supported by cugraph. Those can be reverted/added back if they are needed.

Thanks, we support Pascal or later, so it's safe to remove the code.

pgera:
I realized that the bitmask logic is overengineered. It is correct, but not necessary. When I implemented it, I implemented it with the assumption that max_unsigned can be a valid vid. However, that cannot happen since num_vertices is also represented by unsigned. So the largest valid vid has to be max - 1, which means that max can be used as the invalid marker. So the original implementation just needs to be changed from -1 to ~(0) to support unsigned. I've marked the PR as WIP. I'll push one more change to simplify the logic.

pgera:
This is ready now. I'll do SSSP in a separate PR.

BradReesWork:
okay to test

BradReesWork:
rerun tests

GPUtester:

BradReesWork:
@pgera The issue with this PR is that it does not address the full end-to-end analytic workflow.  From the Python layer, data will still be specified as int32.  Likewise, data from any of the graph classes will also be int32.

pgera:
Yes, I do not use the python layer at all. I made the changes to use cugraph's performance numbers as a comparison point for large graphs in some ongoing work. The python side can be a separate PR, which I can do a bit later once I get some free time. Nothing changes for external users with this PR. Perhaps the PR's title can be renamed to emphasize that this is only the cpp portion.

pgera:
Additionally, supporting multiple types at the python layer would require going through all the algorithms to ensure that they all work for multiple types. So I think that fixing all the underlying cpp bits for all the algorithms would anyway be necessary before exposing anything globally.

BradReesWork:
ok to test

cpp/src/traversal/bfs.cu
```diff
@@ -474,10 +476,14 @@ void bfs(experimental::GraphCSRView<VT, ET, WT> const &graph,
...
-  CUGRAPH_EXPECTS((typeid(WT) == typeid(float)) || (typeid(WT) == typeid(double)),
-                  "Unsupported weight data type, please use float or double");
+  static_assert(std::is_same<VT, unsigned>::value || std::is_same<VT, int>::value ||
+                  std::is_same<VT, long>::value,
+                "Unsupported vertex id data type, please use unsigned, int, or long");
+
+  static_assert(std::is_same<VT, ET>::value, "VT and ET should be the same time");
+
+  static_assert(std::is_same<WT, float>::value || std::is_same<WT, double>::value,
+                "Unsupported weight data type, please use float or double");
```
seunghwak, line 486: These are more dependent on explicit instantiation rather than fundamental requirements of the algorithm and these will be detected in link time anyway.

Better be

```
static_assert(std::is_integral<VT>::value, "...");
static_assert(std::is_integral<ET>::value, "...");
static_assert(std::is_floating_point<WT>::value, "...");  // actually, this is unnecessary for BFS
```

cpp/src/traversal/bfs.cu
```diff
@@ -494,17 +500,48 @@ void bfs(experimental::GraphCSRView<VT, ET, WT> const &graph,
   bfs.traverse(start_vertex);
 }
 
+template void bfs<unsigned, unsigned, float>(
```
seunghwak, line 503: add
```
// explicit instantiation
```

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -92,7 +92,7 @@ __global__ void fill_unvisited_queue_kernel(int *visited_bmap,
     // saving the common offset
     if (threadIdx.x == (FILL_UNVISITED_QUEUE_DIMX - 1)) {
       IndexType total               = unvisited_thread_offset + n_unvisited_in_int;
-      unvisited_common_block_offset = atomicAdd(unvisited_cnt, total);
+      unvisited_common_block_offset = traversal::atomicAdd(unvisited_cnt, total);
```
seunghwak, line 95: Ideally, traversal::atomicAdd is not specific to traversal and better go to RAFT (e.g. https://github.com/rapidsai/raft/blob/branch-0.15/cpp/include/raft/cudart_utils.h)

If it is too much for this PR, better add a comment like
```
// FIXME: this should better go to RAFT
```

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -299,8 +299,9 @@ __global__ void main_bottomup_kernel(const IndexType *unvisited,
     // by different in in visited_bmap)
     IndexType visited_bmap_index[1];  // this is an array of size 1 because CUB
                                       // needs one
-    visited_bmap_index[0]      = -1;
-    IndexType unvisited_vertex = -1;
+
+    visited_bmap_index[0]      = ~IndexType(0);
+    IndexType unvisited_vertex = ~IndexType(0);
```
seunghwak, line 304: See

https://github.com/rapidsai/cugraph/pull/838/files#diff-ba59ce02957abb53c35f1a1929fc0374R682

This is not merged yet (and you can just copy and paste this part and to graph.hpp), but I think this is a better way to handle this.

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -353,7 +354,7 @@ __global__ void main_bottomup_kernel(const IndexType *unvisited,
 
       // If we haven't found a parent and there's more edge to check
       if (!found && degree > MAIN_BOTTOMUP_MAX_EDGES) {
-        left_unvisited_off = atomicAdd(left_unvisited_cnt, (IndexType)1);
+        left_unvisited_off = traversal::atomicAdd(left_unvisited_cnt, (IndexType)1);
```
seunghwak, line 357: static_cast<IndexType>(1) instead of old C type cast.

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -555,7 +556,7 @@ __global__ void bottom_up_large_degree_kernel(IndexType *left_unvisited,
     // is know with inactive threads
     for (IndexType i_edge = first_i_edge + logical_lane_id; i_edge < end_i_edge;
          i_edge += BOTTOM_UP_LOGICAL_WARP_SIZE) {
-      IndexType valid_parent = -1;
+      IndexType valid_parent = ~IndexType(0);
```
seunghwak, line 559: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -564,7 +565,8 @@ __global__ void bottom_up_large_degree_kernel(IndexType *left_unvisited,
         if (lvl_u == (lvl - 1)) { valid_parent = u; }
       }
 
-      unsigned int warp_valid_p_ballot = cugraph::detail::utils::ballot((valid_parent != -1));
+      unsigned int warp_valid_p_ballot =
+        cugraph::detail::utils::ballot(valid_parent != ~IndexType(0));
```
seunghwak, line 569: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -703,16 +705,21 @@ __global__ void topdown_expand_kernel(
...
-  IndexType block_offset = (blockDim.x * blockIdx.x) * max_items_per_thread;
-  IndexType n_items_per_thread_left =
-    (totaldegree - block_offset + TOP_DOWN_EXPAND_DIMX - 1) / TOP_DOWN_EXPAND_DIMX;
+  IndexType block_offset            = (blockDim.x * blockIdx.x) * max_items_per_thread;
+  IndexType n_items_per_thread_left = 0;
+
+  if (totaldegree > block_offset) {
+    n_items_per_thread_left =
+      (totaldegree - block_offset + TOP_DOWN_EXPAND_DIMX - 1) / TOP_DOWN_EXPAND_DIMX;
+  }
```
seunghwak, line 714:
```
IndexType n_items_per_thread_left = total_degree > block_offset ? (totaldegree - block_offset + TOP_DOWN_EXPAND_DIMX - 1) / TOP_DOWN_EXPAND_DIMX : 0;
```

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -703,16 +705,21 @@ __global__ void topdown_expand_kernel(
...
 
   n_items_per_thread_left = min(max_items_per_thread, n_items_per_thread_left);
 
   for (; (n_items_per_thread_left > 0) && (block_offset < totaldegree);
 
        block_offset += MAX_ITEMS_PER_THREAD_PER_OFFSETS_LOAD * blockDim.x,
-       n_items_per_thread_left -= MAX_ITEMS_PER_THREAD_PER_OFFSETS_LOAD) {
+       n_items_per_thread_left -=
+       min(n_items_per_thread_left, (IndexType)MAX_ITEMS_PER_THREAD_PER_OFFSETS_LOAD)) {
```
seunghwak, line 722: Better use std::min instead of custom min.

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -703,16 +705,21 @@ __global__ void topdown_expand_kernel(
...
     // In this loop, we will process batch_set_size batches
     IndexType nitems_per_thread =
       min(n_items_per_thread_left, (IndexType)MAX_ITEMS_PER_THREAD_PER_OFFSETS_LOAD);
```
seunghwak, line 725: Better use std::min instead of custom min.

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -834,8 +841,8 @@ __global__ void topdown_expand_kernel(
             vec_u[iv]                                    = frontier[k];  // origin of this edge
             vec_frontier_degrees_exclusive_sum_index[iv] = frontier_degrees_exclusive_sum[k];
           } else {
-            vec_u[iv]                                    = -1;
-            vec_frontier_degrees_exclusive_sum_index[iv] = -1;
+            vec_u[iv]                                    = ~IndexType(0);
+            vec_frontier_degrees_exclusive_sum_index[iv] = ~IndexType(0);
```
seunghwak, line 845: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -844,7 +851,7 @@ __global__ void topdown_expand_kernel(
         for (int iv = 0; iv < TOP_DOWN_BATCH_SIZE; ++iv) {
           IndexType u = vec_u[iv];
           // row_ptr for this vertex origin u
-          vec_row_ptr_u[iv] = (u != -1) ? row_ptr[u] : -1;
+          vec_row_ptr_u[iv] = (u != ~IndexType(0)) ? row_ptr[u] : ~IndexType(0);
```
seunghwak, line 854: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -856,12 +863,18 @@ __global__ void topdown_expand_kernel(
...
-
-          // Destination of this edge
-          vec_dest_v[iv] = (row_ptr_u != -1) ? col_ind[edge] : -1;
+          // Need this check so that we don't use invalid values of edge to index
+          if (row_ptr_u != ~IndexType(0)) {
+            IndexType edge = row_ptr_u + gid - vec_frontier_degrees_exclusive_sum_index[iv];
+
+            if (edge_mask && !edge_mask[edge]) {
+              // Disabling edge
+              row_ptr_u = ~IndexType(0);
```
seunghwak, line 872: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -874,7 +887,7 @@ __global__ void topdown_expand_kernel(
         for (int iv = 0; iv < TOP_DOWN_BATCH_SIZE; ++iv) {
           IndexType v = vec_dest_v[iv];
           vec_v_visited_bmap[iv] =
-            (v != -1) ? previous_bmap[v / INT_SIZE] : (~0);  // will look visited
+            (v != ~IndexType(0)) ? previous_bmap[v / INT_SIZE] : (~int(0));  // will look visited
```
seunghwak, line 890: invalid_vertex_id<IndexType>::value is much more readable than ~IndexType(0)

cpp/src/traversal/bfs_kernels.cuh
```diff
@@ -889,7 +902,7 @@ __global__ void topdown_expand_kernel(
 
           int is_visited = vec_v_visited_bmap[iv] & m;
 
-          if (is_visited) vec_frontier_candidate[iv] = -1;
+          if (is_visited) vec_frontier_candidate[iv] = ~IndexType(0);
```
seunghwak, line 905: Search for ~IndexType(0) and replace with invalid_vertex_id<IndexType>::value

cpp/src/traversal/traversal_common.cuh
```diff
@@ -204,6 +218,19 @@ binsearch_maxle(const IndexType* vec, const IndexType val, IndexType low, IndexT
...
+template <typename T>
+__device__ static __forceinline__ T atomicAdd(T* addr, T val)
+{
+  return ::atomicAdd(addr, val);
+}
+
+template <>
+__device__ __forceinline__ long atomicAdd<long>(long* addr, long val)
+{
+  return ::atomicAdd(reinterpret_cast<unsigned long long*>(addr),
```
seunghwak, line 230: Better use int64_t instead of long and better add
static_cast(sizeof(unsigned long long) == sizeof(int64_t), "");

cpp/src/traversal/bfs.cu
pgera: My concern was that is_integral is too wide (it includes all widths including 8 and 16), and unless the kernel is really tested to work for all widths, it will likely break if you instantiate it with anything other than what is tested. There are implicit assumptions in places for sizes to be at least 32 bits. One example is the dimx params for block/grid etc. So this assert was to signal the current state of affairs. Same for std::is_same<VT, ET>::value.

cpp/src/traversal/bfs_kernels.cuh
pgera: Will require --expt-relaxed-constexpr. Is that OK ? I can change all the mins in the file for both host and device code to std::min.

cpp/src/converters/COOtoCSR.cu
```diff
@@ -19,13 +19,36 @@
 
 namespace cugraph {
 
+// unsigned + float
+template std::unique_ptr<experimental::GraphCSR<unsigned, unsigned, float>>
+coo_to_csr<unsigned, unsigned, float>(
+  experimental::GraphCOOView<unsigned, unsigned, float> const &graph,
+  rmm::mr::device_memory_resource *);
```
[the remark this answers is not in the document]
pgera: Done.

cpp/src/converters/COOtoCSR.cu
```diff
@@ -19,13 +19,36 @@
 
 namespace cugraph {
```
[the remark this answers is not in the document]
pgera: Done.

cpp/src/traversal/bfs.cu
```diff
@@ -461,6 +461,8 @@ void BFS<IndexType>::clean()
   // the vectors have a destructor that takes care of cleaning
 }
```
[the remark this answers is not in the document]
pgera: Done.

cpp/src/traversal/bfs.cu
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Deferring the move to RAFT for now since RAFT is a separate repo. Added the comment.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Changed all mins to std::min and added --expt-relaxed-constexpr.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/traversal_common.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done.

cpp/src/traversal/bfs_kernels.cuh
pgera: Done. Added the invalid_idx type to graph.hpp, and invalid_vertex_id and invalid_edge_id derive from it. Btw, technically, there shouldn't be a reason to use -1 for int. std::numeric_limits<T>::max() should work for all types. The only place it will affect things is possibly at the API level where currently -1 is returned as invalid (for eg., for predecessors). In the future, it could be changed to be std::numeric_limits<T>::max() to make it uniform.

cpp/src/traversal/bfs.cu
seunghwak: OK, I see... If you expect this code to break if sizeof(ET) > sizeof(VT), then, better avoid relaxing the check.

cpp/src/traversal/bfs_kernels.cuh
seunghwak: Oh... Sorry, I didn't see this is a kernel code :-(

pgera: Ok. So leave device mins as is ?

cpp/src/traversal/bfs.cu
pgera: I'll update the check to assert is_integral, size at least 4 bytes, and VT and ET are the same type. I think the kernels need one additional pass to make them really type agnostic and safe. Need to go through all the arithmetic line by line to check for breakage. There may also be orthogonal overflow issues. One way to weed them out would be to use graphs of sizes at the limit of the type. For eg., things like ceil division (a + K -1)/K will start breaking at that point.

cpp/src/traversal/bfs_kernels.cuh
seunghwak: Yeah... Better leave as the original code. If our codebase currently does not use --expt-relaxed-constexpr, better leave as is.

pgera: Reverted.

cpp/tests/traversal/bfs_test.cu
```diff
@@ -166,7 +171,7 @@ class Tests_BFS : public ::testing::TestWithParam<BFS_Usecase> {
       // that the predecessor obtained with the GPU implementation is one of the
       // predecessors obtained during the C++ BFS traversal
       VT pred = cugraph_pred[i];  // It could be equal to -1 if the node is never reached
-      if (pred == -1) {
+      if (pred == ~VT(0)) {
```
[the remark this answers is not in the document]
pgera: Done.

BradReesWork:
rerun tests

BradReesWork:
rerun tests

kaatish:
@pgera Do you have any more changes in mind for this PR?

pgera:
I'm not sure why some of the CI tests are failing. Was there some known CI issue ? I just merged the upstream branch into mine. So if there were any fixes in the last few days, they should be pulled in now. @kaatish, this PR is done. There should be no impact to users at the moment. The next steps in this direction would be to decouple VT and ET and python bindings, which can be done in follow ups. The bottom up issue (#960) still needs to be fixed, but that is independent.

PR #962, opened 2020-06-22
[REVIEW] dask pagerank updates

### LINKED ISSUE #821: [FEA] OPG at dask-cudf and python level
cuGraph needs a multi-GPU python API that accepts and returns distributed dask dataframes in an OPG paradigm. It should then pass the graph to OPG functions of the C++ layer.

Most components should come from cuML, ideally through RAFT.

- [x] dask opg helper functions
- [x] nccl bindings
- [x] all gather test
- [x] document the steps to run opg features from python

Iroy30:
Closes #821


GPUtester:

python/cugraph/dask/opg_pagerank/pagerank.py
```diff
@@ -20,30 +20,54 @@
...
+                                   max_iter,
+                                   tol,
+                                   personalization,
+                                   nstart)
 
 
-def pagerank(input_graph):
-    print("INSIDE DASK PAGERANK")
+
+def pagerank(input_graph,
```
afender, line 36: We should probably have doc here and update https://github.com/rapidsai/cugraph/blob/branch-0.15/docs/source/dask-cugraph.rst.tmp

python/cugraph/opg/link_analysis/mg_pagerank_wrapper.pyx
```diff
@@ -39,15 +39,14 @@ def mg_pagerank(input_df, local_data, handle, alpha=0.85, max_iter=100, tol=1.0e
     local_offset = dst.min() # Find using local data offset array
```
afender, line 39: Perhaps size info should be set in the python graph directly only once and just accessed afterward.

python/cugraph/dask/opg_pagerank/pagerank.py
```diff
@@ -20,30 +20,54 @@
...
+    personalization = None
+    nstart = None
+
     client = default_client()
     _ddf = input_graph.edgelist.edgelist_df
     ddf = _ddf.sort_values(by='dst', ignore_index=True)
     data = DistributedDataHandler.create(data=ddf)
     comms = Comms(comms_p2p=False)
     comms.init(data.workers)
     local_data = data.calculate_local_data(comms)
```
afender, line 56: Let's review options for decoupling `calculate_local_data` from PageRank. From a user perspective, there is no way to separate ETL time than actual Pagerank time otherwise.

python/cugraph/tests/dask/opg_pagerank_testing.py
```diff
@@ -38,10 +35,28 @@ def test_dask_pagerank():
...
+                       delimiter=' ',
+                       names=['src', 'dst', 'value'],
+                       dtype=['int32', 'int32', 'float32'])
 
-    dcg.pagerank(g)
+    g = cugraph.DiGraph()
+    g.from_cudf_edgelist(df, 'src', 'dst')
+
+    dg = cugraph.DiGraph()
+    dg.from_dask_cudf_edgelist(ddf)
```
afender, line 47: 1. How do we support datasets stored in multiple files?
2. If the file is on an NFS this will be limited by network speed. I wonder if each node first gets a local copy before parsing with the GPU.

python/cugraph/tests/dask/opg_pagerank_testing.py
```diff
@@ -38,10 +35,28 @@ def test_dask_pagerank():
...
+
+    expected_pr = cugraph.pagerank(g)
+    result_pr = dcg.pagerank(dg)
```
afender, line 50: `dcg` -> `dask_cugraph`

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: dask read_csv takes care of reading from multiple files, just like snmg version.

afender:
`    cluster = LocalCUDACluster(protocol="tcp", scheduler_port=0)
`

Do we really need to always explicitly specify `tcp` and scheduler port?
There's also an option called `enable_tcp_over_ucx`  https://dask-cuda.readthedocs.io/en/latest/api.html#cluster

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: we import cugraph.dask as dcg and run pagerank. This is to follow cuml's import structure--> from cuml.dask import TheModel

python/cugraph/dask/opg_pagerank/pagerank.py
afender: For instance, we could have the option to call `addTransposedAdjList` before Pagerank. In this case, `calculate_local_data` would happen at that time, as well as coo2csr and renumbering potentially. Then Pagerank would only do this if CSC is not present already.

Just need to look at compatibility if we `addAdjList` or run OPG BFS afterward.

python/cugraph/tests/dask/opg_pagerank_testing.py
afender: dask cuDF features are imported from `dask_cudf` though.

python/cugraph/dask/opg_pagerank/pagerank.py
Iroy30: updated

python/cugraph/opg/link_analysis/mg_pagerank_wrapper.pyx
Iroy30: updated

python/cugraph/structure/graph.py
```diff
@@ -471,13 +503,21 @@ def number_of_vertices(self):
 
         """
         if self.node_count is None:
-            if self.adjlist is not None:
+            if self.distributed:
+                if self.edgelist is not None:
```
afender, line 507: If an adj list is present I assume we could just get it from there too.

python/cugraph/structure/graph.py
```diff
@@ -493,6 +533,11 @@ def number_of_edges(self, directed_edges=False):
         Get the number of edges in the graph.
 
         """
+        if self.distributed:
+            if self.edgelist is not None:
+                return len(self.edgelist.edgelist_df)
+            else:
```
afender, line 539: If an adj list is present I assume we could just get it from there too.

python/cugraph/structure/graph.py
```diff
@@ -212,7 +215,26 @@ def add_edge_list(self, source, destination, value=None):
...
+        if type(self) is Graph:
+            raise Exception('Undirected distributed graph not supported')
+        if isinstance(input_ddf, dask_cudf.DataFrame):
+            self.distributed = True
+            self.local_data = None
+            self.edgelist = self.EdgeList(input_ddf)
+        else:
+            raise Exception('input should be a dask_cudf dataFrame')
+
+    def compute_local_data(self, by):
```
afender, line 229: I like the concise design. Missing test and doc though.

python/cugraph/dask/common/part_utils.py
```diff
@@ -149,3 +151,82 @@ def _extract_partitions(dask_obj, client=None):
...
+                             cumsum[i].iloc[index] +
+                             move_count)
+        move_count = cumsum[i].iloc[-1] - cumsum[i].iloc[index]
+    new_divisions.append(new_divisions[i+1] +
+                         cumsum[-1].iloc[-1] +
+                         move_count - 1)
+    return ddf.repartition(divisions=tuple(new_divisions))
+
+
+def load_balance_func(ddf_, by, client=None):
```
afender, line 194: doc

python/cugraph/dask/common/part_utils.py
```diff
@@ -149,3 +151,82 @@ def _extract_partitions(dask_obj, client=None):
...
+def set_global_index(df, cumsum):
+    df.index = df.index + cumsum
+    return df
+
+
+def get_cumsum(df, by):
+    return df[by].value_counts(sort=False).cumsum()
+
+
+def repartition(ddf, cumsum):
```
afender, line 174: doc

python/cugraph/dask/opg_pagerank/pagerank.py
```diff
@@ -13,37 +13,120 @@
...
+
+    """
+    Find the PageRank values for each vertex in a graph using multiple GPUs.
+    cuGraph computes an approximation of the Pagerank using the power method.
+    The input graph must contain edge list as  dask-cudf dataframe with
+    one partition per GPU.
+
+    Parameters
+    ----------
+    graph : cugraph.Graph
```
afender, line 53: we support DiGraph too

Iroy30: i think on the contrary, we should probably mention that we support only DiGraph. since we dont auto-symmetrize the distributed dataframe

python/cugraph/structure/graph.py
Iroy30: added both

python/cugraph/dask/common/part_utils.py
Iroy30: updated

python/cugraph/dask/common/part_utils.py
Iroy30: updated

python/cugraph/dask/opg_pagerank/pagerank.py
Iroy30: updated to mention that we currently accept only directed graphs for opg algorithms

python/cugraph/tests/dask/opg_pagerank_testing.py
```diff
@@ -38,10 +34,33 @@ def test_dask_pagerank():
...
+
     g = cugraph.DiGraph()
-    g.from_dask_cudf_edgelist(ddf)
+    g.from_cudf_edgelist(df, 'src', 'dst')
+
+    dg = cugraph.DiGraph()
+    dg.from_dask_cudf_edgelist(ddf)
+
+    # Pre compute local data
+    # dg.compute_local_data(by='dst')
```
kaatish, line 49: Is this comment meant to be removed?

Iroy30: This is only to be used for benchmarking so we pre-compute local data. Otherwise it is not required to be run prior to calling the algorithm

python/cugraph/structure/graph.py
Iroy30: Currently we don't allow initialization of distributed graphs using adjlist. So for a distributed graph, if it is not empty,  would mean an edgelist is always present.

python/cugraph/structure/graph.py
Iroy30: Currently we don't allow initialization of distributed graphs using adjlist. So for a distributed graph, if it is not empty,  would mean an edgelist is always present.

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: To be discussed in the near future.

Issue #925, opened 2020-06-02
[ENH] recover renumbering from from_cudf_edgelist


lmeyerov:
**Describe the solution you'd like**

`from_cudf_edgelist` needs a way to recover the mapping when `renumber=True`, such as part of the return or as an attribute of the graph. Or, if already doable, the docs for `from_cudf_edgelist` and `cugraph` need to clarify.

**Additional context**

`add_edge_list`, which returns the mapping, throws a near-term deprecation warning in favor of `from_cudf_edgelist`, yet without a way to recover mappings, it is not a usable replacement when `renumber=True`. (`from_cudf_edgelist` does not appear to preserve edge order, so cannot infer, either.)

This is in `0.13`, so may already be too late for to `0.14`...

afender:
Hi, thanks for your input!

> from_cudf_edgelist needs a way to recover the mapping when renumber=True

Typically, after 0.13, users do not need to be aware of renumbering or maps. cuGraph takes care of all the mapping so that features accept and return results with users' identifiers. Our vision is that renumbered representations are internal matters that should not need to be explicitly exposed by default.  If you think you found a bug, please let us know what workflow is causing issues or what analytics caused ID mismatches. 

> if already doable [...]

As an attribute of the internal structure, advanced users can still access it : `graph.edgelist.renumber_map`

> without a way to recover mappings, it is not a usable replacement when renumber=True

One may call the separate renumbering function prior to calling `from_cudf_edgelist` while setting auto renumbering to `False` there. It seems to me that this pipeline is equivalent to pre-0.12 usage.

lmeyerov:
Thanks, +1 for separating user-level IDs from internal for the typical case :)  

Will keep an eye open if we hit this again  with 0.14+ APIs!

Issue #812, opened 2020-04-06
[FEA] coo2csr


afender:
**Describe the solution you'd like**

1.  **Input** 
- Dask cudf edge list.
Note : Since Pagerank runs on the transposed, source and destination arrays will be swapped.

2. **Output** 
- CSR matrices : Each GPU has its local, rectangular, CSR matrix. The rows are between 0 and number_of_local_vertices and the columns are between 0 and number_of_global_vertices. 
- Start identifiers : of the global ID of the first vertex of each partition
- End identifiers : of the global ID of the last vertex of each partition

Implementation
- 32-bit is needed but the code should be templated to allow 64bit offsets in the future
- Ideally, all the distributed calls are done through dask cudf and single-GPU coo2csr is called locally. 

**Task List**

- [ ] Implementation 
- [ ] Tests : verify the result by comparing to the single GPU version
- [ ] Record performance
- [ ] API : this feature needs to be exposed in cugraph's OPG API. In addition, this needs to interact well with other OPG PageRank functions.

**Dependency of #485**

afender:
The pipeline allows just leveraging the local coo2csr in #944

Issue #809, opened 2020-04-06
[FEA] Consolidation


afender:
Some workflows are using dask cuDF on multi-GPUs for large scale ETL. By the time the edge list is extracted, it may fit in 1 GPU and benefit from cuGraph's signe-GPU path.

cuGraph should accept a distributed cudf dataframe in an OPG context and consolidate it internally into a single GPU (or as few GPUs as possible).

This work allows building and validating part of the python OPG stack without dealing with the distributed call at C++ level.

PR #963, opened 2020-06-22
[REVIEW] Renumbering refactor, add multi GPU support

### LINKED ISSUE #678: [ENH] Accept users notation for auto multi-column renumbering/unrenumbering
**Describe the bug**
The cugraph API supports auto multi-column renumbering with PR #666 . All cugraph algorithms use the renumbered notation for running algorithms and un-renumber the vertices in the result back to user's notation for viewing. The bindings need to be further updated to accept additional arguments(vertices in the case of multi column renumbering map) in the API as user's notation while calling algorithms and internally renumber them, for example, nstart in pagerank, source in sssp, etc. 

The following needs to be resolved:
1) Currently there are two renumbering methods implemented and in use. Identify if multi-column renumbering #636 can handle all renumbering cases and proceed with one(or both) renumbering method accordingly for further changes.
2) Auto renumber multi column vertices passed by the user in algorithms. 
3) Update (un)renumbering degree calculation and two hop neighbors in graph.py
4) Test and verify functionality of all changes

### LINKED ISSUE #779: [BUG] Inconsistency of where we unrenumber
**Describe the bug**
In some cases we implementing un-renumbering within the Cython xxxx_wrapper.pyx code.  In other cases we implement un-renumbering within the python code xxxx.py.  For example, PageRank will renumber nstart (guess at the PageRank vector) and personalization and will unrenumber the output within the pagerank_wrapper.pyx code and the pagerank.py file is a simple pass thru.  On the other hand, get_two_hop_neighbors does the renumbering in the the python code and the graph_wrapper returns a raw result.

This should be consistent, as it is confusing and requires looking through multiple files to understand what is happening.

My inclination would be to move all renumbering and unrenumbering up into the python code, and let the cython code be slightly simpler.  But I'm not entirely sure of all of the trade-offs.

### LINKED ISSUE #813: [FEA] Renumbering
1.  **Input** 
- Dask cudf edge list where IDs are any integer type or strings

2. **Output** 
- Dask cudf edge list where IDs are continuous and starting at 0. 
- 32 bit by default (64 bit may be needed latter)

3. **Test**
- verify the result by comparing to the single GPU version
- test it combined with #812 in #485 pipeline

Misc.
- [ ] Record performance
- [ ] API : this feature needs to be exposed in cugraph's OPG API. In addition, this needs to interact well with other OPG PageRank functions.

### LINKED ISSUE #922: [ENH] doc whether renumber preserves edge order
Docs clarify whether renumbering preserves edge orders. 

(Ideally they do, because this simplifies then plugging in different edge weights without manually swizzling.)

**Additional context**
Encountered in various `renumber` calls in 0.13, e.g.: 
https://docs.rapids.ai/api/cugraph/stable/api.html?highlight=renumber#cugraph.structure.renumber.renumber_from_cudf

### LINKED ISSUE #972: [BUG] degree function crashes when multi-column renumbering is performed.
If a graph is build from multiple coulmn source/destination, then degree and degrees crash

datafile='.karate-data.csv'
gdf = cudf.read_csv(datafile, delimiter='\t', names=['srca', 'dsta'], dtype=['int32', 'int32'] )

gdf['srcb'] = gdf['srca']
gdf['dstb'] = gdf['dsta']

G = cugraph.Graph()
G.from_cudf_edgelist(gdf, source=['srca','srcb'], destination=['dsta','dstb'] )

G.degree()

ChuckHastings:
This PR will refactor the renumbering implementation to include:
 * A new high level design using a python NumberMap object which will encapsulate numbering functionality
 * A python/cudf/dask_cudf based implementation that provides a functional single GPU and MG implementation
 * All python algorithms converted to use the new implementation

Goal is:
 * Closes #813 
 * Closes #922
 * Closes #779
 * Closes #972 
 * Closes #678

This will NOT address a custom C++ implementation.  That will be added later as an optimization, if desired.  However, this implementation is intended to make it easier to address a custom C++ implementation. 

Dependent on #1008


GPUtester:

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,263 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import cugraph
+
+class NumberMap:
```
afender, line 16: Where does this object live in:
- single GPU pipeline?
- multi GPU pipeline?

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,263 @@
...
+    def __init__(self):
+        self.implementation = None
+
+    def compute_vals(column_names):
+        """
+        Helper function to compute internal column names based on external column names
+        """
+        return [str(i) for i in range(len(column_names))]
+
+    def from_dataframe(self, df, col_names):
```
afender, line 69: This PR can't close #813 without integration into 0.15 OPG pipeline (#944, #947, #964) and renumbering tests passing.

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,263 @@
...
+            This list of 1 or more strings contain the names
+            of the columns that uniquely identify an external
+            vertex identifier
+        """
+        if self.implementation is not None:
+            raise Exception('NumberMap is already populated')
+
+        if type(df) is cudf.DataFrame:
+            self.implementation = NumberMap.SingleGPU(df, col_names)
+        elif type(df) is dask_cudf.DataFrame:
```
afender, line 89: How does this integrate into _Consolidation_ workflows #809 ?

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,263 @@
...
+            numbered by the NumberMap class.
+        src_col_names: list of strings
+            This list of 1 or more strings contain the names
+            of the columns that uniquely identify an external
+            vertex identifier for source vertices
+        dst_col_names: list of strings
+            This list of 1 or more strings contain the names
+            of the columns that uniquely identify an external
+            vertex identifier for destination vertices
+        """
```
afender, line 114: Need to document multi-GPU return behavior 
- Are internal IDs randomly spread or are they grouped together so that all edges for the same source (or perhaps destination) are on the same GPU?
- How does this connect to the next step in the OPG pipeline (coo2csr #812 )? Are they natively compatible or do we need an intermediate step?

python/cugraph/structure/number_map.py
ChuckHastings: My goal was that the same object would be used in both pipelines.

python/cugraph/structure/number_map.py
ChuckHastings: Hadn't seen that particular thought process.  I would see consolidation as a separate step that could be done either before or after renumbering.  If consolidation reduces to single GPU then use the Single GPU renumbering.  If consolidation reduces to a cluster of GPUs then the output of consolidation would still be a dask_cudf data frame and it wouldn't really matter whether it had been consolidated or not.

python/cugraph/structure/number_map.py
ChuckHastings: Regarding the ids, the approach I've been using is as follows:
 * External vertex ids are hashed and that hash value is used to partition vertex ids across the cluster.  Each node of the cluster then is responsible for all vertices that hash to it.
 * Each local node of the cluster now has all of the vertex information, perform a local renumbering to get 0..(n-1) numbering of all vertices that hashed to that node
 * "Somehow" assign global ids.  Current prototype does assignment by computing the number of vertices on each partition and doing a prefix sum to create a base vertex identifier for each node.  The consequence is that we end up with an overall 0..(N-1) mapping, each node getting a contiguous subrange of ids that it created.

Regarding the OPG pipeline, this NumberMap does not entirely implement renumbering, it implements the core capability and will be hidden inside the graph object (although externally callable for those who wish to access the renumbering independently).  The renumbering implementation itself will end up using this to create a COO that should be directly usable as input to #812.

python/cugraph/structure/number_map.py
afender: Yes but where in each pipeline

ChuckHastings: On a single GPU, this object would be internal to the Graph object.  It would be automatically generated when from_cudf_edgelist is called, it would be used by all analytics when renumbering/unrenumbering capabilities are required.

If a user decided to access the renumbering features themselves (currently supported in single GPU) the output of the renumbering function would be this object instead of the current Dataframe that gets returned, so that the user can directly call the to_vertex_id and from_vertex_id methods as they need to.

I haven't seen a complete description of the multi GPU pipeline.  My assumption is that it will be similar to the single GPU pipeline.  We would have some overarching python graph object that would create and maintain this object.

python/cugraph/structure/number_map.py
afender: That's fair. Trying to see how would this look from the user's perspective when the dask dataframe and multiGPU and pipeline transfers to a single GPU.

To take the most out of consolidation, we would probably want to renumber first as renumbered IDs are expected to be smaller (and we will leverage multi GPU for renumbering). In this case, the consolidation feature should : 
- accept an un-renumbered distributed dask data frame
- use the multi-GPU renumber path from this PR
- consolidate and return a single GPU Di/Graph.

Just need to make sure that renumbering maps and related objects are compatible with this workflow.

python/cugraph/structure/number_map.py
afender: > On a single GPU, this object would be internal to the Graph object. It would be automatically generated when from_cudf_edgelist is called, it would be used by all analytics when renumbering/unrenumbering capabilities are required.

Sounds good.

> If a user decided to access the renumbering features themselves (currently supported in single GPU) the output of the renumbering function would be this object instead of the current Dataframe that gets returned, so that the user can directly call the to_vertex_id and from_vertex_id methods as they need to.

Ok. We would need to document this object and add an example in the user doc.

> I haven't seen a complete description of the multi GPU pipeline. 

There are GitHub Issues and PRs for OPG PageRank and BFS. Let me know if you have any specific questions. 

> My assumption is that it will be similar to the single GPU pipeline. We would have some overarching python graph object that would create and maintain this object.

To some extent. As of now, there is not one global multi GPU graph object. A distributed dask cudf edge list is accepted and single GPU Di/Graphs are created locally for each rank.

python/cugraph/structure/number_map.py
afender: Sounds good, I think adding doc about this would be helpful. 
Perhaps with a note as well on getting these properties for CSC as the OPG Pagerank pipeline is the first one?

SG does (1) renumber (2) swap src and dst (3) coo2csr. (1) being at edge list loading time and (2,3) at analytics time. 
Now looking at MG, are we relying on the assumption that (2) gets done before (1) in order to get coo2csr successfully building the local CSC matrices?

python/cugraph/structure/number_map.py
ChuckHastings: The NumberMap object would support this flow.

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,716 @@
...
+
+import cudf
+import dask_cudf
+import numpy as np
+import bisect
+
+
+class NumberMap:
+    class SingleGPU:
+        def __init__(self, df, src_col_names, dst_col_names):
```
seunghwak, line 22: Can we add an additional parameter to sort IDs based on in-degree or out-degree?

If we want to insure that higher degree vertices have a smaller ID after renumbering (within a segment if we partition vertices to multiple segments especially for multi-GPU case), I guess we can do something like ```df[src_col_names].groupby(src_col_names).size()``` and sort the resulting DataFrame by size (which is degree).

Things are actually more complicated as we need to consider df[dst_col_names] as well... but I think there is a way to handle this... If we can do this with relatively small performance overhead, this can be used to reduce memory footprint for our Graph class and this will make the pattern accelerator optimization simpler (in code complexity) and faster with smaller memory footprint.

If we intend to generate a CSR-ish graph (to implement the push model, store the graph adjacency matrix in row-major order), we need to sort based on out-degree, and if we intend to generate a CSC-ish graph (to implement the pull model, store the graph adjacency matrix in column-major order), we need to sort based on in-degree.

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -250,8 +268,9 @@ def _initialize_vertices_from_identifiers_list(G, identifiers):
     # FIXME: There might be a cleaner way to obtain the inverse mapping
```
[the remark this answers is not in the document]
ChuckHastings: Done

python/cugraph/centrality/katz_centrality.py
```diff
@@ -93,7 +90,17 @@ def katz_centrality(G,
     >>> kc = cugraph.katz_centrality(G)
     """
 
-    df = katz_centrality_wrapper.katz_centrality(G, alpha, max_iter,
-                                                 tol, nstart, normalized)
+    if nstart is not None:
+        if G.renumbered is True:
+            nstart["vertex"] = G.edgelist.renumber_map.to_vertex_id(
```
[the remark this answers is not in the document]
ChuckHastings: That's where it was, so I left it there.  It seems it would be better up a level.

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,716 @@
...
+                .groupby(src_col_names)
+                .count()
+                .reset_index()
+                .rename(
+                    columns=dict(zip(src_col_names, self.col_names)),
+                    copy=False,
+                )
+            )
+
+            if dst_col_names is not None:
```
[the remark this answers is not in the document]
ChuckHastings: Editing error.  Corrected in next push.

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,716 @@
...
+            if external_column_names is None:
+                return tmp_df
+            else:
+                return tmp_df.rename(
+                    columns=dict(zip(self.col_names, external_column_names)),
+                    copy=False,
+                )
+
+    class MultiGPU:
+        def extract_vertices(
```
seunghwak, line 115: I am still digging into the details and thinking about potential implementation issues,

but if we wish to generate a CSR-ish graph, similar to the single-GPU case, if we can generate a list of (vid, local-out-degree) pairs here, and hash vertex id pairs to compute the target partition (I think you're already doing this), and collect (vid, local-out-degree) pairs in the target partition. In the target partition, we can reduce to compute global-out-degree, and sort before renumbering. If we can do this with low performance overhead (local sorting within a partition might be the most expensive part, but this is cheaper than global sorting), this will make remaining steps much easier (and partitioning becomes really straightforward as well).

seunghwak: So at the end of renumbering, a NumberMap object will hold partitioning of [0, V) (assuming V vertices) to P processes, and within each partition (approximately sizeof V/P assuming good hash function), vertices are either sorted by out-degree (if we wish to generate a CSR-ish graph) or in-degree (if we wish to generate a CSC-ish graph).

I think the additional overhead for maintaining this sorted order within a partition can be pretty moderate unless E/V is really small; shuffling edges based on partitioning to build a CSR-ish graph representation (something we cannot avoid) will be more expensive than this...

ChuckHastings: Question is, do we need to address these performance optimizations in this PR, or can they be added as we advance a bit further down the multi-GPU path?

This PR makes a significant API change to the Single GPU implementation and affects every current algorithm.  The longer we keep this open the more changes get made to the algorithms that have to be retrofitted to the new API.  My inclination would be to defer these optimizations to later in the MG development.

seunghwak: If you want to merge this PR early, you can do this in a later PR, I think we can actually do things in two more PRs, one additional PR doing this optimization without actually writing high-performance C++ code (so we can concurrently working on following steps relying on this), and the last PR trying to cut the graph preparation time as much as possible.

python/cugraph/structure/graph.py
```diff
@@ -969,24 +947,138 @@ def neighbors(self, n):
...
+        >>> pr = number_map.unrenumber(pr, 'vertex')
+        >>>
+        """
+        return self.renumber_map.unrenumber(df, column_name, preserve_order)
+
+    def lookup_vertex_id(self, df, column_name=None):
+        """
+        Given a DataFrame containing external vertex ids in the identified
+        columns, or a Series containing external vertex ids, return a
+        Series with the vertex ids.
```
seunghwak, line 1019: with the "internal" vertex ids?

I think "vertex_id" in look_vertex_id, add_vertex_id, to_vertex_id, from_vertex_id ... can be confusing as both external & internal vertex_IDs are both vertex IDs. Can I get little more explanation about this naming or can we bring better names?

python/cugraph/centrality/katz_centrality.py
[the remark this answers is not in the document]
ChuckHastings: Moved it up.

python/cugraph/structure/number_map.py
```diff
@@ -0,0 +1,614 @@
...
+        """
+        Helper function to compute internal column names based on external
+        column names
+        """
+        return [str(i) for i in range(len(column_names))]
+
+    def from_dataframe(self, df, src_col_names, dst_col_names=None):
+        """
+        Populate the numbering map with vertices from the specified
+        columns of the provided data frame.
```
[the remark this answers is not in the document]
ChuckHastings: Updated documentation

python/cugraph/structure/graph.py
ChuckHastings: Each of these refer to internal vertex ids.  I could rename these functions to:
* lookup_internal_vertex_id
* add_internal_vertex_id
* to_internal_vertex_id
* from_internal_vertex_id
If that would introduce more clarity.

Unfortunately, one of my complaints about our python formatting standards, longer function and variable names make for more silly indenting due to the 80 character constraint on line length.

seunghwak: If the goal is to make the name shorter, we may try something like "lookup_int_vid", but I guess this will raise consistency issue with the rest of cuGraph. If I need to pick between confusing names vs ugly formatting, I will go for ugly formatting.

ChuckHastings: I'd rather get rid of the 80 character line limit.  Thought we got rid of punched cards at the beginning of my career :-)

I also prefer the longer names.  I was mostly whining.

afender: The character line limit is helpful for reviews on github and browsing code in split-window mode.
Now 80 characters look really limited to me. I have no objection increasing it after we double-check what was the initial motivation for this number. Pretty sure 120 would still be ok for github reviews. In comparison, C++ limit is 200.
Also agree that naming > formating

python/cugraph/tests/test_renumber.py
```diff
@@ -15,201 +15,327 @@
...
+    cluster = LocalCUDACluster()
+    client = Client(cluster)
+    client
+
+
+# Test all combinations of default/managed and pooled/non-pooled allocation
+@pytest.mark.parametrize("graph_file", utils.DATASETS)
+def test_opg_renumber(graph_file, client_setup):
+    gc.collect()
+
```
afender, line 284: Don't we need `cugraph.comms` to be initialized for muliple GPUs / nodes test?

python/cugraph/tests/test_renumber.py
```diff
@@ -186,33 +218,135 @@ def test_renumber_files_col(graph_file):
...
 
-    src, dst, numbering = cugraph.renumber_from_cudf(
-        gdf, ['src', 'src_old'], ['dst', 'dst_old'])
 
-    for i in range(len(gdf)):
-        assert sources[i] == (numbering['0'].iloc[src[i]] - translate)
-        assert destinations[i] == (numbering['0'].iloc[dst[i]] - translate)
+# Test all combinations of default/managed and pooled/non-pooled allocation
+@pytest.mark.parametrize("graph_file", utils.DATASETS)
+def test_opg_renumber2(graph_file, client_setup):
```
afender, line 322: Can we add an MG pipeline test involving cuGraph `Graph` and `from_dask_cudf_edgelist`?

python/cugraph/structure/graph.py
```diff
@@ -184,56 +195,60 @@ def from_cudf_edgelist(self, input_df, source='source',
...
         if value is not None:
-            input_df['weights'] = value
-            self.from_cudf_edgelist(input_df, edge_attr='weights')
+            input_df["weights"] = value
+            self.from_cudf_edgelist(input_df, edge_attr="weights")
         else:
             self.from_cudf_edgelist(input_df)
 
-    def from_dask_cudf_edgelist(self, input_ddf):
+    def from_dask_cudf_edgelist(self, input_ddf, renumber=False):
```
afender, line 251: `renumber=False` should be True to be consistent with the single GPU `from_cudf_edgelist`.
Notice that later degree and pagerank MG tests use both in the same test without specifying this option which would result in different input. I think these tests pass just by luck because the input is already renumbered.

python/cugraph/structure/number_map.py
afender: @ChuckHastings I lost track if you added doc for accessing the renumbered data. 
This is motivated by users requests, see for instance #925. 
Feel free to resolve and comment with the commit id.

python/cugraph/structure/number_map.py
afender: Added more details on this in my latest review. Resolving this one.

python/cugraph/structure/number_map.py
afender: Should we have a test?

python/cugraph/structure/number_map.py
afender: Can't see the doc for the multi-GPU return behavior, let me know if you added it elsewhere

python/cugraph/layout/force_atlas2.py
```diff
@@ -15,21 +15,23 @@
...
-                 edge_weight_influence=1.0,
-                 jitter_tolerance=1.0,
-                 barnes_hut_optimize=True,
-                 barnes_hut_theta=0.5,
-                 scaling_ratio=2.0,
-                 strong_gravity_mode=False,
-                 gravity=1.0,
-                 verbose=False,
-                 callback=None):
+def force_atlas2(
```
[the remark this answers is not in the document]
ChuckHastings: Switching to the new renumbering caused formatting issues in almost every .py file.  It was suggested that I use black to reformat the code.  I used it somewhat selectively, but it still affected many of the files.

Not sure what other RAPIDS devs are doing, perhaps we should use black on all of the files with a consistent configuration for it?

python/cugraph/tests/test_renumber.py
ChuckHastings: I'm thinking I should move the MG renumber tests into the dask directory with the other MG unit tests.

python/cugraph/tests/test_renumber.py
ChuckHastings: I have one that's not checked in, I can add it to the PR.

python/cugraph/structure/graph.py
ChuckHastings: Agreed.  I set it as False so that I would - by default - not change the existing MG unit tests.

python/cugraph/tests/test_renumber.py
ChuckHastings: The renumbering software doesn't actually do anything with cugraph.Comms, it is implemented entirely in cudf.

python/cugraph/structure/number_map.py
ChuckHastings: Not sure what you're asking for.  The typical workflow is outlined in the renumber notebooks.  It works like this:

* Instantiate a NumberMap object
* Call from_dataframe to populate the number map
* Call to_internal_vertex_id, add_internal_vertex_id, from_internal_vertex_id as desired to convert between internal and external vertex ids.

Alternatively, you can call NumberMap.renumber on a DataFrame and it will return a fully populated NumberMap and a renumbered DataFrame.  Then you can call NumberMap.unrenumber on a DataFrame and it will convert internal vertex ids back into external vertex ids.

If you call it with a cudf.DataFrame, it uses Single GPU logic to create and translate everything.  If you call it with a dask_cudf.DataFrame it uses MG logic and generated dask_cudf.DataFrame objects everywhere.  The NumberMap object keeps track internally of everything necessary to do this.

ChuckHastings: I further wrapped this in the Graph renumber and unrenumber functions which are automatically used if you specify renumber=True on the from_cudf_edgelist or from_dask_cudf_edgelist calls.

python/cugraph/layout/force_atlas2.py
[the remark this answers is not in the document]
afender: Yes, let's find out and use the same tool+config consistently: https://github.com/rapidsai/cugraph/issues/1016

python/cugraph/structure/number_map.py
afender: > Not sure what you're asking for.

Documentation that describes the data distribution behavior of MG renumbering output in general. Users of the renumbering feature (and MG developer) will wonder if each node gets a contiguous range or is it scattered for instance.

python/cugraph/structure/graph.py
afender: :+1: 

One more thing I realized is that the doc regarding renumbering is outdated for 
- `from_cudf_edgelist`
- `from_dask_cudf_edgelist`

ChuckHastings:
> @ChuckHastings if the input data is int32 in the source and destination, what is the dtype of renumbered data, is it same as input or always int64?
> Combining mg renumbering with mg pagerank makes dask_cudf's sorting go haywire, specifically during searchsorting https://github.com/rapidsai/cudf/blob/049f93c4387907553d614e69202cc8e0d2ddc793/python/dask_cudf/dask_cudf/sorting.py#L25 because of int32/int64 discrepancy arising from dask's lazy compute which could be the cause of the whole dataset being partitioned into a single gpu/patition.

Found the disconnect in the calls (I was using int32 in some places and int64 in another place).  Latest push fixes this.  The output type of renumber is specified as an optional parameter to the NumberMap constructor, we default to int32.

python/cugraph/structure/graph.py
ChuckHastings: Updated this to True now that everything is working correctly.

I copied the renumber parameter documentation over to the from_dask_cudf_edgelist in the latest push.

PR #964, opened 2020-06-22
[REVIEW] Opg bfs

kaatish:


GPUtester:

cpp/src/traversal/opg/worker_kernels.cuh
```diff
@@ -0,0 +1,128 @@
...
+
+template <typename VT, typename ET, typename WT, typename Operator>
+void large_vertex_worker(
+    cugraph::experimental::GraphCSRView<VT, ET, WT>& graph,
+    DegreeBucket<VT, ET>& bucket,
+    Operator op,
+    cudaStream_t stream) {
+  GraphCSRDeviceView<VT, ET, WT> graph_device_view(graph);
+  int block_size = 32;
+  int block_count = 1024;
```
afender, line 60: Please document the logic behind block sizes and counts.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+  size_t tid = threadIdx.x + (blockIdx.x * blockDim.x);
+  if (tid < word_count) {
+    for (int i = 0; i < frontier_count; ++i) {
+      word = word | frontiers[i][tid];
+    }
+    frontiers[my_rank][tid] = word;
+  }
+}
+
+template <typename VT, typename ET, typename WT>
```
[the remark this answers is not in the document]
kaatish: The primary reason for having bfs communicator right now is because we do not have bitwise or all reduce in raft. This class allows me to abstract out the resource allocation and calls for this specific use case.

For experimentation stage with custom nccl, I plan to switch the internals of this class so that I can easily test the performance difference without a lot of code changes.

Once a particular implementation is selected, I intend to remove this class.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+  {
+    for (int i = 0; i < total_gpu_count_; ++i) {
+      if (i != my_gpu_rank_) {
+        data_[i].resize(word_count_);
+        ptr_[i] = data_[i].data().get();
+      }
+    }
+  }
+
+  void allreduce(rmm::device_vector<unsigned>& frontier) {
```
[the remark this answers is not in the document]
kaatish: Ok, I could try to get it in there. Hopefully they have a use case for it.

cpp/src/traversal/opg/graph_csr_device_view.cuh
```diff
@@ -0,0 +1,63 @@
...
+#define GRAPH_CSR_DEVICE_VIEW_CUH
+
+#include <graph.hpp>
+
+namespace detail {
+
+namespace opg {
+
+template <typename VT, typename ET, typename WT>
+class GraphCSRDeviceView {
```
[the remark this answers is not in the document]
kaatish: This was probably an oversight. I think I will remove this class and make the kernels work with CSR View class.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+    cudaStream_t stream = handle_.get_stream();
+    ptr_[my_gpu_rank_] = frontier.data().get();
+    for (int i = 0; i < total_gpu_count_; ++i) {
```
afender, line 87: This is a RAFT allgatherv : https://github.com/rapidsai/raft/blob/3038c773636540941b8bf9e30c487f20bddc3cdc/cpp/include/raft/comms/std_comms.hpp#L377

An alternative would be to try and reduce as you go in order to avoid storing all frontiers locally.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <raft/handle.hpp>
+#include <graph.hpp>
+
+
+namespace detail {
+
+namespace opg {
```
seunghwak, line 22: Are we plan to use the opg namesapce?

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <raft/handle.hpp>
+#include <graph.hpp>
+
+
```
seunghwak, line 19: namespace cugraph here?

cpp/src/traversal/opg/bfs_comms.cuh
seunghwak: And the functions in the opg namespace are not really opg specific.

divup below resembles

https://github.com/rapidsai/cudf/blob/7f91a17a967b0e0b502d56d37818aa0461d38c67/cpp/include/cudf/detail/utilities/integer_utils.hpp#L77

from cuDF.

To maximize code duplication/code reuse, I don't think bfs_comms.cuh is the right file to define these functions. Ideally, this should go to RAFT.

cpp/src/traversal/opg/bfs_comms.cuh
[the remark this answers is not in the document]
seunghwak: In that case, better add a comment saying this code is for temporary testing and will not be merged. Then, I will worry less about code bifurcation. We'd better avoid code duplication/bifurcation as this makes our codebase very difficult to maintain and understand.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  void allreduce(rmm::device_vector<unsigned>& frontier) {
+    cudaStream_t stream = handle_.get_stream();
+    ptr_[my_gpu_rank_] = frontier.data().get();
+    for (int i = 0; i < total_gpu_count_; ++i) {
+      handle_.get_comms().bcast(ptr_[i], frontier.size(), i, stream);
+    }
+    ptr_device_ = ptr_;
+    reduce_bitwise_or<<<divup(frontier.size(), 512), 512>>>(
```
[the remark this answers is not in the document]
seunghwak: Ideally we should bring something like

https://github.com/rapidsai/cudf/blob/branch-0.15/cpp/include/cudf/detail/utilities/cuda.cuh#L40

to RAFT, and embed check for max_num_threads & max_grid_size inside the class.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+        ptr_[i] = data_[i].data().get();
+      }
+    }
+  }
+
+  void allreduce(rmm::device_vector<unsigned>& frontier) {
+    cudaStream_t stream = handle_.get_stream();
+    ptr_[my_gpu_rank_] = frontier.data().get();
+    for (int i = 0; i < total_gpu_count_; ++i) {
+      handle_.get_comms().bcast(ptr_[i], frontier.size(), i, stream);
```
seunghwak, line 88: What's the expected value of froniter.size()? Is it V / 8?

And the aggregate memory footprint for this code is total_gpu_count * V / 8?

If this is correct, memory footprint will become very large with a large number of GPUs. This code may not work beyond a small number of GPUs.

cpp/src/traversal/opg/bfs_comms.cuh
```diff
@@ -0,0 +1,103 @@
...
+void reduce_bitwise_or(
+    unsigned** frontiers,
+    size_t frontier_count,
+    int my_rank,
+    size_t word_count) {
+
+  unsigned word = 0;
+  size_t tid = threadIdx.x + (blockIdx.x * blockDim.x);
+  if (tid < word_count) {
+    for (int i = 0; i < frontier_count; ++i) {
```
seunghwak, line 48: If I am not mistaken, frontier_count here is the # of GPUs. So, the amount of computing here actually increases with an increasing number of GPUs. This will kill scalability unless you are expecting this code to work only for a small number of GPUs.

cpp/src/traversal/opg/frontier_expansion.cu
```diff
@@ -0,0 +1,152 @@
...
+    cugraph::experimental::GraphCSRView<VT, ET, WT>& graph,
+    VT *predecessors,
+    const VT start_vertex) {
+
+  //We need to keep track if a vertex is visited or its status
+  //This needs to be done for all the vertices in the global graph
+  size_t word_count = number_of_words(graph.number_of_vertices);
+  rmm::device_vector<unsigned> input_frontier(word_count);
+  rmm::device_vector<unsigned> output_frontier(word_count);
+  rmm::device_vector<unsigned> visited(word_count);
```
seunghwak, line 83: This will not scale beyond a small number of GPUs. Memory footprint stays same with the increasing number of GPUs. This cannot process a graph unless number_of_vertices bits * 3 fits inside a single GPU.

cpp/src/traversal/opg/frontier_expansion.cu
```diff
@@ -0,0 +1,152 @@
...
+        input_frontier.begin(), input_frontier.end(),
+        visited.begin(),
+        visited.begin(),
+        bitwise_or());
+
+    //2. Clear out output frontier
+    thrust::fill(
+        output_frontier.begin(),
+        output_frontier.end(),
+        static_cast<unsigned>(0));
```
seunghwak, line 117: What will be the work complexity of 1 and 2 for V vertices and P GPUs? Is it O(V) (instead of O(V/P))?

cpp/src/traversal/opg/frontier_expansion.cu
```diff
@@ -0,0 +1,152 @@
...
+      *flag_ = 1;
+    }
+  }
+};
+
+template <typename VT, typename ET, typename WT>
+void bfs(raft::handle_t const &handle,
+    cugraph::experimental::GraphCSRView<VT, ET, WT>& graph,
+    VT *predecessors,
+    const VT start_vertex) {
```
seunghwak, line 76: How does this partition work/data across GPUs? What are the expectations on the input graph?

cpp/src/traversal/opg/bfs_comms.cuh
kaatish: The frontier size V/32. As I have mentioned earlier, this is a stand in till allreduce with bitwise or takes shape. The intention is to find out the performance and memory requirement and see if getting bitwise or reduction operator in nccl is necessary or not.

cpp/src/traversal/opg/frontier_expansion.cu
kaatish: The word count is (number of vertices / 32) since every unsigned word takes care of 32 vertices. That amounts to approx 10 percent memory overhead.

cpp/src/traversal/opg/frontier_expansion.cu
kaatish: For now the graph is 1d partitioned

cpp/src/traversal/opg/frontier_expansion.cu
kaatish: Since the partitioning is 1d, every GPU needs to keep track of the global frontier which is V/32 in size. The complexity therefore is O(V)

cpp/src/traversal/opg/bfs_comms.cuh
kaatish: This is a standin code for nccl bcast with bitwise or. Having said that, the computation required for an allreduce operation will increase with an increasing number of GPUs.

cpp/src/traversal/opg/frontier_expansion.cu
seunghwak: OK, so V/32 words(V/8 bytes), and if I am not mistaken, this does not decrease with P (graph.number_of_vertices is the total number of vertices, right?). This is 10% memory overhead (with a single GPU?) but if other memory allocations go down with more GPUs, this will eventually become the dominating factor.

seunghwak: Is this PR targeting a small number of GPUs? (2 or 4?)

kaatish: >  this does not decrease with P (graph.number_of_vertices is the total number of vertices, right?).

No it doesn't.

>  if other memory allocations go down with more GPUs
I am not allocating more frontiers in this section if more GPUs are present. Each of these vectors only needs to exist once per GPU. Input and output frontier for the local frontier calculation and visited for the global accumulation of visited vertices.

> Is this PR targeting a small number of GPUs? (2 or 4?)
This PR is not targeting small number of GPUs. Ideally around 32/64.

cpp/src/traversal/opg/bfs_comms.cuh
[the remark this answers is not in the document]
afender: Eventually, anyone who needs to exchange/reduce bitmaps would need that.

cpp/src/traversal/opg/frontier_expansion.cu
seunghwak: I see. Please let me know if you are currently just experimenting and I jumped to code review too soon. I'd better not complain too much about the code that will not be merged :-)

cpp/src/traversal/opg/load_balance.cuh
```diff
@@ -0,0 +1,78 @@
...
+    dist_.setup(graph_.offsets, active_bitmap, vertex_begin_, vertex_end_);
+    auto distribution = dist_.run(reorganized_vertices_, stream);
+
+    DegreeBucket<VT, ET> large_bucket = distribution.degreeRange(16);
+    large_vertex_worker(graph_, large_bucket, op, stream);
+
+    DegreeBucket<VT, ET> medium_bucket = distribution.degreeRange(12, 16);
+    medium_vertex_worker(graph_, medium_bucket, op, stream);
+
+    DegreeBucket<VT, ET> small_bucket_0 = distribution.degreeRange(10,12);
```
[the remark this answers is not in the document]
kaatish: The choice to map kernels to the degree ranges has more to do with the performance of the kernel rather than characteristics of the graph. We can do some more work to check out what kernel is better suited to a degree later on.

cpp/tests/traversal/load_balance.cu
```diff
@@ -0,0 +1,134 @@
...
+#include <raft/handle.hpp>
+#include <utilities/error.hpp>
+#include <rmm/mr/device/cuda_memory_resource.hpp>
+
+#include "traversal/opg/load_balance.cuh"
+
+// ============================================================================
+// Test Suite
+// ============================================================================
+typedef struct BFS_Usecase_t {
```
xcadet, line 29: This should probably renamed to `LB_Usecase_t`

cpp/src/traversal/bfs.cu
```diff
@@ -486,21 +487,24 @@ void bfs(raft::handle_t const &handle,
...
-  int beta  = 18;
-  // FIXME: Use VT and ET in the BFS detail
-  cugraph::detail::BFS<VT> bfs(
-    number_of_vertices, number_of_edges, offsets_ptr, indices_ptr, directed, alpha, beta);
-  bfs.configure(distances, predecessors, sp_counters, nullptr);
-  bfs.traverse(start_vertex);
+  if (handle.comms_initialized()) {
+    CUGRAPH_EXPECTS(sp_counters != nullptr, "BFS Traversal shortest path is not supported in OPG path");
+    opg::bfs<VT, ET, WT>(handle, graph, distances, predecessors, start_vertex);
+  } else {
```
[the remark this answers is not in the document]
afender: Fair point, we do need a better way to separate MG distributed from MG batch other than checking the handle communicator. There are 2 things we should do :
- If we add a batch parameter at the C++ API level, this is something we could leverage. 
- When a distributed graph is constructed we fill some specific structures. We should also set an "is_distributed" status in the graph that analytics can check this.

Other suggestions?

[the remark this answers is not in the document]
kaatish: @xcadet Essentially, when you are calling MG BC, you want to call SF BFS right?

Is the graph distributed or just replicated across all ranks. Or in other words, is the graph's parameters 
```
  VT *local_vertices;
  ET *local_edges;
  VT *local_offsets;
```
set to a non nullptr?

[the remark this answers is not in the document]
seunghwak: > When a distributed graph is constructed we fill some specific structures. We should also set an "is_distributed" status in the graph that analytics can check this.

I think we should go for

```
template <bool multi_gpu = false, ...>
class Graph {
  static constexpr bool is_multi_gpu = multi_gpu;
  ...
};
```

The above approach enables compile-time resolution of multi-GPU vs single-GPU.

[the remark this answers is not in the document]
xcadet: > Essentially, when you are calling MG BC, you want to call SG BFS right?

Yes I would need to call SG BFS from MG BC
> Is the graph distributed or just replicated across all ranks.

The graph would be replicated across all ranks.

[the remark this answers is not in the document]
afender: @seunghwak I like the template parameter approach better as well.
Let's address this in a separate issue https://github.com/rapidsai/cugraph/issues/995

BradReesWork:
rerun tests

PR #968, opened 2020-06-24
[REVIEW] cuGraph Benchmark Integration [skip-ci]

dillon-cullinan:
- [x] Write benchmark build script
- [x] Test internally

The `benchmark/build.sh` script will be used to run and store the benchmark results to move into our S3 bucket. This script is intended to be used in both **nightly** and **PR** benchmark runs by CI.


GPUtester:

ci/benchmark/build.sh
```diff
@@ -0,0 +1,123 @@
...
+# TODO: Delete build section when artifacts are available
+
+logger "Check environment..."
+env
+
+logger "Check GPU usage..."
+nvidia-smi
+
+logger "Activate conda env..."
+source activate gdf
```
ajschmidt8, line 64: the `gdf` environment has a symlink to keep it operational, but it might be better to use the new `rapids` conda environment.

src: https://github.com/rapidsai/gpuci-build-environment/blob/branch-0.15/rapidsai/base-runtime.Dockerfile#L59

PR #974, opened 2020-06-26
[REVIEW] fix_column_rename

Iroy30:

PR #976, opened 2020-06-26
[REVIEW] Fix error in Graph.edges(), update cuDF rename() calls

trxcllnt:
* Fixes places that are breaking from changes in cuDF's `DataFrame.rename()` method
* Ensures columns are named `src, dst` in `Graph.view_edge_list()` when graph was created via `renumber_from_cudf`. Presently they're named `0_x` and `0_y`, causing `Graph.edges()` to throw an error

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,6 @@ cpp/doxygen/html
 # Raft symlink
 python/cugraph/raft
 python/_external_repositories/
+
+# created by Dask tests
+python/dask-worker-space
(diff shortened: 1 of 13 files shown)


GPUtester:

trxcllnt:
I see these changes are also in https://github.com/rapidsai/cugraph/pull/974. The main difference is I'm also passing `copy=False` to `df.rename()` since it's safe to avoid copying in cuGraph tests/internals.

afender:
rerun tests

trxcllnt:
I'm not sure what the `codecov/patch` CI check is verifying?

PR #977, opened 2020-06-26
[REVIEW] Update force_atlas2 to call on_train_end after iterating

trxcllnt:
* Updates `force_atlas2` BH approximation to call `callback->on_train_end()` after iterating
* Updates the test to verify each of the callback's methods are invoked the expected number of times


GPUtester:

trxcllnt:
@afender seems like the same erroneous codecov patch failure here?

PR #978, opened 2020-06-26
[WIP] Add cugraph.hypergraph implementation to convert DataFrames into Graphs

trxcllnt:
This PR adds a method to transform an arbitrary table into a hypergraph by constructing nodes for each row and unique value in a set of columns, and edges between the row nodes and unique value nodes. The implementation is inspired by the [`hypergraph()` method](https://github.com/graphistry/pygraphistry/blob/3bd9052266aff6c1d1e9c5aa9d57d76a0be8eb9b/graphistry/hyper.py#L196) in pygraphistry. Opening as a draft WIP PR until I can add docstrings/comments.

It's possible for the resulting nodes table to contain more rows than `graph.nodes()`, because constructing the graph from an edgelist implicitly prunes orphans.
* Is this acceptable?
* If not, is there a way to represent orphan nodes in cugraph?
* Alternatively, should we prune the orphans from the nodes table?

### DIFF
--- a/python/cugraph/__init__.py
+++ b/python/cugraph/__init__.py
@@ -28,6 +28,7 @@ from cugraph.structure import (
     Graph,
     DiGraph,
     from_cudf_edgelist,
+    hypergraph,
     renumber,
     symmetrize,
     symmetrize_df,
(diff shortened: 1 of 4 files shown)


GPUtester:

trxcllnt:
Closed in favor of #979 because I forgot to select the "Draft PR" button.

PR #979, opened 2020-06-26
[REVIEW] Add cugraph.hypergraph implementation to convert DataFrames into Graphs

trxcllnt:
This PR adds a method to transform an arbitrary table into a hypergraph by constructing nodes for each row and unique value in a set of columns, and edges between the row nodes and unique value nodes. The implementation is inspired by the [`hypergraph()` method](https://github.com/graphistry/pygraphistry/blob/3bd9052266aff6c1d1e9c5aa9d57d76a0be8eb9b/graphistry/hyper.py#L196) in pygraphistry. Opening as a draft WIP PR until I can add docstrings/comments.

It's possible for the resulting nodes table to contain more rows than `graph.nodes()`, because constructing the graph from an edgelist implicitly prunes orphans.
* Is this acceptable?
* If not, is there a way to represent orphan nodes in cugraph?
* Alternatively, should we prune the orphans from the nodes table?


GPUtester:

trxcllnt:
Note, this is blocked by https://github.com/rapidsai/cugraph/pull/976

lmeyerov:
Cool!

1. From discussion - Singleton nodes may be a common missing case (e.g., event with null entities)
- Is this test missing from PyGraphistry? I can add
- For handling in cugraph: A hacky workaround is adding self-edges and then filtering out after. Guessing this is / will be a better way?

2. Shaping: The `direct=True` option to override shaping (`'EDGES': {'from_col1': ['to_col1', 'to_col2', ...]}`) is useful in the context of `direct=False` as well and not yet supported in PyGraphistry, so not in the existing test cases. It would be interpretable in the `direct=False` case by allowing the event node to be a valid value for `from_col1`.

lmeyerov:
From conv w/ Paul: sounds like 1+2 are relevant, and already good on 3+4

trxcllnt:
rerun tests

ChuckHastings:
Is there some documentation as to what you're trying to accomplish in constructing a hypergraph in this manner?  I'm having to make a bunch of assumptions as I read the python documentation and I'd be more comfortable if they were spelled out.  Perhaps just a more complete simple example.  Perhaps adding some description of what the unit tests are attempting to do.

trxcllnt:
@ChuckHastings The implementation was written to match the [description](https://pygraphistry.readthedocs.io/en/latest/graphistry.html#graphistry.pygraphistry.PyGraphistry.hypergraph) and [implementation](https://github.com/graphistry/pygraphistry/blob/3bd9052266aff6c1d1e9c5aa9d57d76a0be8eb9b/graphistry/hyper.py#L196) in pygraphistry.

The transformation finds relationships between rows by linking rows that share unique values in each column. This is a quick and way to make a graph from any structured high-dimensional dataset, for example with cyber-security event logs. @BartleyR and co. currently use the one in pygraphistry, but all the pystring and pandas ops means it's prohibitively slow in many scenarios. For example, I have a table that takes ~22m20s with pygraphistry, but ~8s with this implementation.

#### How it works
To create the node list, we create nodes representing each column's unique values (or "entities"); if the same value appears in multiple columns, each occurrence is considered a distinct entity, and multiple nodes are created<sup>1</sup>. After that, we create hypernodes representing each row (or "event") of the input table.

The unique values for each column are then related via a second-degree connection through their shared hypernodes (i.e. the originating rows). The set of incoming edges for each hypernode is a set of hyperedges.

Consider this example:

<details><summary>Click to expand example</summary><pre>
def print_cugraph_hyper(xs):
    print('nodes:\n' + str(xs['graph'].nodes()) + '\n')
    print('edges:\n' + str(xs['graph'].edges()) + '\n')
    print('node_attributes:\n' + str(xs['nodes']) + '\n')
    print('edge_attributes:\n' + str(xs['edges']) + '\n')
    print('')<br>
print_cugraph_hyper(cugraph.hypergraph(cudf.DataFrame({
    "col_1": ["a", "a", "a"],
    "col_2": ["a", "b", "b"],
    "col_3": ["a", "b", "c"],
}), EVENTID='row_id'))
</pre></details>

The first line prints the following node list:
<details><summary>Click to expand example output</summary><pre>
# note: cuGraph's column `0` renamed to `node`<br>
        node  id
0   col_1::a   0
1   col_2::a   1
2   col_2::b   2
3   col_3::a   3
4   col_3::b   4
5   col_3::c   5
6  row_id::0   6
7  row_id::1   7
8  row_id::2   8
</pre></details>

From the output we see nodes were created for each column's unique values, plus three nodes for each of the rows.

The second line prints the following edge list:
<details><summary>Click to expand example output</summary><pre>
        src        dst
0  col_1::a  row_id::0
1  col_1::a  row_id::1
2  col_1::a  row_id::2
3  col_2::a  row_id::0
4  col_2::b  row_id::1
5  col_2::b  row_id::2
6  col_3::a  row_id::0
7  col_3::b  row_id::1
8  col_3::c  row_id::2
</pre></details>

From the output we see the nodes for each unique column value are linked to the originating row. We can also see which values are related to each other, via their second-degree connection through a shared row node<sup>2</sup>. For example:
```
(col_2::b) ---> (row_id::1) <--- (col_1::a)
```

#### Additional property tables:
The extra tables returned are extra property tables for each graph component:

<details><summary>Click to expand example output</summary><pre>
node_attributes:
     node_id node_type category col_1 col_2 col_3     row_id
0   col_1::a     col_1    col_1     a  None  None       None
1   col_2::a     col_2    col_2  None     a  None       None
2   col_2::b     col_2    col_2  None     b  None       None
3   col_3::a     col_3    col_3  None  None     a       None
4   col_3::b     col_3    col_3  None  None     b       None
5   col_3::c     col_3    col_3  None  None     c       None
6  row_id::0    row_id    event     a     a     a  row_id::0
7  row_id::1    row_id    event     a     b     b  row_id::1
8  row_id::2    row_id    event     a     b     c  row_id::2<br>
edge_attributes:
      row_id edge_type attrib_id col_1 col_2 col_3
0  row_id::0     col_1  col_1::a     a     a     a
1  row_id::1     col_1  col_1::a     a     b     b
2  row_id::2     col_1  col_1::a     a     b     c
3  row_id::0     col_2  col_2::a     a     a     a
4  row_id::1     col_2  col_2::b     a     b     b
5  row_id::2     col_2  col_2::b     a     b     c
6  row_id::0     col_3  col_3::a     a     a     a
7  row_id::1     col_3  col_3::b     a     b     b
8  row_id::2     col_3  col_3::c     a     b     c
</pre></details>

In the indirectly-connected scenario (the default), the node lists above are unions of two inner lists -- the unique values for each column ("entities"), and the hypernodes representing each row ("events"). These two tables are returned in the result, in case the user wants to perform further processing on the graph component properties.

These kinds of graphs have a few distinct visual characteristics in Graphistry, this kind of "octopus-star" layout:
<details><summary>Click here to expand a visual representation</summary>
</details>

#### Footnotes

1. This behavior can be controlled by the "categories" parameter, a dictionary mapping `column_name` to `category_name`. If the same value appears in two columns, and those columns are mapped to the same category, those occurrences will collapse to a single node instead of two. For example, with a table of source and destination IP addresses:
<details><summary>Click to expand example</summary><pre>
print_cugraph_hyper(cugraph.hypergraph(
    cudf.DataFrame({
        "src_ip": ["0.0.0.0", "1.1.1.1"],
        "dst_ip": ["0.0.0.0", "2.2.2.2"],
    }),
    EVENTID='row_id',
    categories={'src_ip': 'ip', 'dst_ip': 'ip'}
))
</pre></details>
<details><summary>Click to expand example output</summary><pre>
nodes:
          node  id
0  ip::0.0.0.0   0
1  ip::1.1.1.1   1
2  ip::2.2.2.2   2
3    row_id::0   3
4    row_id::1   4<br>
edges:
           src        dst
0  ip::0.0.0.0  row_id::0
1  ip::1.1.1.1  row_id::1
2  ip::2.2.2.2  row_id::1
</pre></details>

2. Passing `direct=True` will cause nodes with second-degree connections via their shared row to be connected directly, and the row nodes are removed from the result. For example, the directly-connected output of the IP-addresses hypergraph in 1. is this:
<details><summary>Click to expand example</summary><pre>
print_cugraph_hyper(cugraph.hypergraph(
    cudf.DataFrame({
        "src_ip": ["0.0.0.0", "1.1.1.1"],
        "dst_ip": ["0.0.0.0", "2.2.2.2"],
    }),
    EVENTID='row_id',
    categories={'src_ip': 'ip', 'dst_ip': 'ip'},
    direct=True
))
</pre></details>
<details><summary>Click to expand example output</summary><pre>
nodes:
             0  id
0  ip::0.0.0.0   0
1  ip::1.1.1.1   1
2  ip::2.2.2.2   2<br>
edges:
           src          dst
0  ip::0.0.0.0  ip::0.0.0.0
1  ip::1.1.1.1  ip::2.2.2.2
</pre></details>

ChuckHastings:
@trxcllnt Awesome description.  Can I suggest that you take that documentation and put it into a sample jupyter notebook that uses this feature as part of this PR?  That would help any users that aren't currently using this feature understand the context, why they might want to use it, and how to go about using it to analyze their data.

I now understand enough of the context to review this PR.

lmeyerov:
@ChuckHastings this is all reimplementing https://github.com/graphistry/pygraphistry

@trxcllnt an idea here is in the nb doing something like df -> cugraph hg -> g.plot() (new Graphistry Hub means it's free)

lmeyerov:
RE:notebook, had some fun:

```
import cudf, cugraph, graphistry

#free account @ https://www.graphistry.com/get-started
graphistry.register(api=3, protocol='https', server='hub.graphistry.com', username='leotest', password='accountaccount')

n = 100
e = cudf.DataFrame({
    's': range(n), 
    'd':  [(x+1) % n for x in range(n)], 
    'w': cudf.Series([x/100 for x in range(n)], dtype='float32')
})
G = cugraph.from_cudf_edgelist(e, source='s', destination='d', edge_attr='w')
graphistry\
    .edges(G.view_edge_list())\
    .bind(source='src', destination='dst', edge_weight='weights')\
    .settings(url_params={'edgeInfluence': 3})\
    .plot()
```

Replacing `e` with the result of something like:

```
e = cugraph.hypergraph(
  cudf.DataFrame({
    'events': ['a', 'b', 'c'],
    'user': ['x', 'y', 'z',
    'ip': [0, 0, 1]
    'email_domain': [1, 2, 3]
  })['edges']
...

plot()

```

python/cugraph/structure/hypergraph.py
```diff
@@ -0,0 +1,503 @@
...
+    values. Conversely, if an event has unique metadata, the unique metadata
+    will turn into nodes that only have connections to the event node.
+
+    For best results, set ``EVENTID`` to a row's unique ID, ``SKIP`` to all
+    non-categorical columns (or ``columns`` to all categorical columns),
+    and ``categories`` to group columns with the same kinds of values.
+
+    Parameters
+    ----------
+    values : cudf.DataFrame
```
afender, line 89: Just curious, how hard would it be to also accept a distributed dask cudf DataFrame?

trxcllnt: I haven't tried it, but theoretically we should be able to support it with some testing.

trxcllnt:
> @Iroy30 Is there a reason we keep the dataframe around (columns a and b) at the end or can these be dropped as they are not edgeattrs ?

This can be controlled by the `drop_edge_attrs=True` option, which has the added bonus of reducing the number of columns to be concatenated for the edge attributes.

> Also is there a way which would allow us to combine two columns as one node? For example if we have a dataframe with columns as "src_ip", "src_port", "dst_ip", "dst_port", is it possible to create a hypergraph with "src_ip" and "src_port" combination to be one unique node?

Not in the current iteration -- you'd need to combine the `src_ip` and `src_port` columns into a single `source_addr` column before calling `cugraph.hypergraph()`.

However, this might not lead to as useful a graph depending on the dataset, since each a node is constructed for each unique value. For example if you had a portscan of a single `src_ip` on 50k ports, you'd get 50k independent nodes whose only edges are to their `row_id` node. If they're kept separate, you'd have 50k port nodes all with edges to the single `src_ip` hypernode.

PR #980, opened 2020-06-28
[REVIEW] Replace nvgraph Spectral Clustering (SC) functionality with RAFT SC

aschaffer:
This is part of refactoring Spectral Clustering (SC) functionality inside RAFT and removing nvgraph's SC.

Dependent on RAFT PR:  https://github.com/rapidsai/raft/pull/12


GPUtester:

cpp/src/community/spectral_clustering.cu
```diff
@@ -21,24 +21,20 @@
...
-#include <graph.hpp>
 
-#include <nvgraph/include/sm_utils.h>
 #include <rmm/thrust_rmm_allocator.h>
 #include <thrust/transform.h>
 #include <ctime>
-#include <nvgraph/include/nvgraph_error.hxx>
-#include <utilities/error.hpp>
 
-#include <nvgraph/include/modularity_maximization.hxx>
```
ChuckHastings, line 33: I would suggest removing:
 * nvgraph/*.cu
 * nvgraph/*.cpp
Removing the corresponding entries from CMakeLists.txt.  If we're picking this all up from raft now, we shouldn't need these files.

I would suggest removing anything in nvgraph/include/* that is not referenced once you remove the above files.

That will make it clear what remaining nvgraph references we have internally.

cpp/src/community/spectral_clustering.cu
```diff
@@ -21,24 +21,20 @@
...
-#include <nvgraph/include/nvgraph_cublas.hxx>
-#include <nvgraph/include/nvgraph_cusparse.hxx>
-#include <nvgraph/include/partition.hxx>
+#include <graph.hpp>
```
afender, line 29: Which `graph.hpp` is this? I noticed it was added in https://github.com/rapidsai/raft/pull/12 but it is still in cuGraph as well.
If we plan to get it from RAFT we need to figure out how to deploy it since it is a public API header of cuGraph.

aschaffer: This is one from `cugraph`. I am working on a solution to remove the one in RAFT, as the SC routines should not, after more careful thought, depend on `GraphCSRView<>` (they only use the CSR data from `GraphCSRView`, anyway). 

If, at a later date we find it worth to expose the `graph.hpp` to RAFT as well, we may do it separately. But SC shouldn't need it. I think RAFT should stay graph-agnostic as much as possible.

afender: Agreed. I also like the solution you are proposing because it allows us to keep `graph.hpp`. This file had over 10 commits in the past month, all driven by cuGraph, so it is best if it stays with us for now.

aschaffer: Done. (Already fixed in https://github.com/rapidsai/raft/pull/12 and in this PR; `cuml` side to follow soon)

cpp/src/community/spectral_clustering.cu
aschaffer: Done. `nvgraph` dependencies were removed (specifically, form: spectral_clustering, triangle_counting, extract_by_vertex, and louvain).

And `nvgraph` dir was moved under `src/legacy/`.

cpp/src/community/triangles_counting.cu
```diff
@@ -16,17 +16,19 @@
...
-#include <nvgraph/include/nvgraph_error.hxx>
+#include <utilities/sm_utils.h>
+#include <utilities/error.hpp>
 
 #include <rmm/thrust_rmm_allocator.h>
 #include <rmm/device_buffer.hpp>
 
 #include <thrust/iterator/counting_iterator.h>
 
+#include <raft/utils/sm_utils.hpp>
```
seunghwak, line 31: We're including utilities/sm_utils.h in https://github.com/rapidsai/cugraph/pull/980/files#diff-a8b84d43fbea60927dbb836c2f31b577R23

and raft/utils/sm_utils.hpp here.

What are the differences between the two? Do we need both? Can't we consolidate the two to one?

aschaffer: The `raft` one is much more comprehensive. The `cugraph` one is much shorter and provides slightly different utils.

PR #985, opened 2020-07-02
[REVIEW] Add raft handle to BFS, BC and edge BC

kaatish:
related to #815


GPUtester:

cpp/include/algorithms.hpp
```diff
@@ -298,7 +300,8 @@ void force_atlas2(experimental::GraphCOOView<VT, ET, WT> &graph,
  *
  */
 template <typename VT, typename ET, typename WT, typename result_t>
-void betweenness_centrality(experimental::GraphCSRView<VT, ET, WT> const &graph,
+void betweenness_centrality(raft::handle_t const &handle,
```
seunghwak, line 303: I am not sure handle can be const.

cpp/src/traversal/bfs.cu
```diff
@@ -520,15 +526,17 @@ template void bfs<uint32_t, uint32_t, double>(
   bool directed);
 
 // Explicit Instantiation
-template void bfs<int, int, float>(experimental::GraphCSRView<int, int, float> const &graph,
+template void bfs<int, int, float>(raft::handle_t const &handle,
```
seunghwak, line 529: better use int32_t instead of int. int is not always 4 byte in every system.

cpp/tests/centrality/edge_betweenness_centrality_test.cu
```diff
@@ -228,6 +229,7 @@ typedef struct EdgeBC_Usecase_t {
 } EdgeBC_Usecase;
 
 class Tests_EdgeBC : public ::testing::TestWithParam<EdgeBC_Usecase> {
+  raft::handle_t handle;
```
seunghwak, line 232: handle_ as this is a private member variable.

cpp/src/link_analysis/pagerank_1D.cu
```diff
@@ -51,7 +51,7 @@ Pagerank<VT, ET, WT>::Pagerank(const raft::handle_t &handle_,
   is_setup = false;
 
   // intialize cusparse. This can take some time.
-  cugraph::detail::Cusparse::get_handle();
+  handle_.get_cusparse_handle();
```
kaatish, line 54: This should work fine correct? We want cusparse handle creation to be relegated to the raft handle?

cpp/include/algorithms.hpp
kaatish: It is supposed to be const. That's how it is in page rank right now. In raft repo you can see handle being passed as const [here](https://github.com/rapidsai/raft/blob/3038c773636540941b8bf9e30c487f20bddc3cdc/cpp/include/raft/comms/test.hpp#L33).

seunghwak: Umm... yeah... maybe it can be const under the current situation, but I don't think it is future ready. If a handle holds a resource that can be modified, const does not work unless you mark those variables as _mutable_ (and I think we'd better avoid doing that as much as possible).

But I may not complain about this too much as changing this from const to non-const is pretty easy, so we can defer this till this actually becomes problematic.

kaatish: We can move the conversation to a new issue concerning constness of handle if you would like. In my opinion, if we have a non static raft handle then we must pass raft handle by const reference or by value. The reason being that you do not want the caller to do able to assign a raft handle from within the function. This could also be done by deleting copy constructor and equality operator but I think that will be more cumbersome. Not too sure about that.

> const does not work unless you mark those variables as mutable (and I think we'd better avoid doing that as much as possible).

Actually this is currently the case in raft handle. Every resource in the raft handle (and I suspect cuml handle) is mutable and every function to create/set the resources within are const functions.

Maybe you would like to create an issue on raft repo to discuss about this?

BradReesWork:
rerun tests

BradReesWork:
rerun tests

PR #989, opened 2020-07-09
[FIX] update docker image [skip-ci]

sean-frye:
PR fixes local docker image used by build script and programmatically inserts minor version num into `DOCKER_IMAGE` string

### DIFF
--- a/ci/local/build.sh
+++ b/ci/local/build.sh
@@ -3,7 +3,7 @@
 GIT_DESCRIBE_TAG=`git describe --tags`
 MINOR_VERSION=`echo $GIT_DESCRIBE_TAG | grep -o -E '([0-9]+\.[0-9]+)'`
 
-DOCKER_IMAGE="gpuci/rapidsai:${MINOR_VERSION}-cuda10.1-devel-ubuntu16.04-py3.6"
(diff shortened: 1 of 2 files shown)


GPUtester:

ci/local/build.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
-DOCKER_IMAGE="gpuci/rapidsai:cuda10.0-ubuntu16.04-gcc5-py3.6"
+GIT_DESCRIBE_TAG=`git describe --tags`
+MINOR_VERSION=`echo $GIT_DESCRIBE_TAG | grep -o -E '([0-9]+\.[0-9]+)'`
+
+DOCKER_IMAGE="gpuci/rapidsai:${MINOR_VERSION}-cuda10.1-devel-ubuntu16.04-py3.6"
```
ajschmidt8, line 6: looks good Sean. In the PR below, we'll be removing Python 3.6 support though. So it's probably a good idea to change this image to end with `py3.7` instead of `py3.6`.

https://github.com/rapidsai/gpuci-build-environment/pull/104

sean-frye: Thanks for catching that, AJ! Change has been added to this PR. Will make the change to the other PRs as well.

PR #993, opened 2020-07-13
[REVIEW] Persistent handle

### LINKED ISSUE #1001: [ENH] Create the handle only once
This is moivated by performance analysis results: 
- Handle (empty) : < 1ms
- Handle (cuda libs set): 700ms to 1119ms
- Comms: 20ms (2GPUs) 400ms (4GPUs), 67000ms (32GPUs)
- Pagerank iteration on twitter (1GPU) : 150ms

Hence, the handle is costlier than cugraph analytics. We should create it once, outside of other cugraph features. This is a tradeoff. 

We also considered 
- a handle per graph : discarded because of conflicting handles when passing/returning multiple graphs
- recreating the handle on-demand from the environment info : discarded because it is missing the purpose of the handle and cost
- a pool of handles: discarded because lack of immediate benefit

Iroy30:
This PR integrates persistent Handle to avoid reinitialization of nccl. The changes are as follows:
-> When running opg(multi-gpu) algorithm, the user is expected to perform Comms initialization right after cluster and client initialization. This step initializes nccl and the ranks remain the same for all the workers throughout the session.
-> For Single GPU algorithms, Comms initialization is not required. Internally, the default Handle can be obtained using get_default_handle() and passed to C++ apis. This default handle would not have nccl initialized. 

There is no change in the python API apart from the additional step of Comms initialization for running opg. The next step would be to access the handle and pass it to all C++ function APIs including Single GPU functionalities (in which case the default_handle would be accessed instead of raft Comms handle)

Close #1001


GPUtester:

python/cugraph/tests/dask/opg_pagerank_testing.py
```diff
@@ -24,9 +25,9 @@ def test_dask_pagerank():
     gc.collect()
     cluster = LocalCUDACluster()
     client = Client(cluster)
+    Comms.initialize()
```
afender, line 28: Should this be `reinitialize` like RMM? We should document this API and this should be optional when using default parameters.

Some doc on behavior :
`To configure options to be used in cuGraph before loading, simply do the above before you import cugraph. You can re-initialize the cugraph communicator with different settings at run time by calling reinitialize().`

python/cugraph/tests/dask/opg_degree_testing.py
```diff
@@ -12,8 +12,9 @@
 def test_dask_opg_degree():
 
     gc.collect()
-    cluster = LocalCUDACluster(protocol="tcp", scheduler_port=0)
+    cluster = LocalCUDACluster()
     client = Client(cluster)
+    Comms.initialize()
```
afender, line 17: Is `Comms` the best name we can think of here? 
The handle stores CUDA libs context and device properties in addition to the communicator. It is also quite specific to cuGraph. I think the name should capture all this so be a bit more self-documented.

Iroy30: It is a cugraph encapsulation of raft Comms. It does contain the handle which in turn stores much more but the thought behind naming it Comms is the fact that it is only accessed and initialized when a communicator is needed for multi-gpu. Else we internally use default_handle in single gpu functionalities and this import is not required. However, I do see the point you put forth. We could call it environment initialization instead of communicator but that could be confusing.

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: Depends on whether we want to reinitialize. rmm allows reinitialization. On the contrary, currently the Comms here allows initialization just once and uses that throughout. Any reinitialization at a later point might cause some assumptions to be False(rank initialization and partition order of workers) and precomputed distributed data with new comms initialization might misbehave.

Iroy30:
rerun tests

Iroy30:
rerun tests

python/cugraph/tests/dask/opg_pagerank_testing.py
afender: Can Comms.initialize() be embedded elsewhere or do we need to explicitly call it?

python/cugraph/tests/dask/opg_degree_testing.py
afender: I see. I think it is ok then, we can just carefully select the  `import cugraph.comms.comms as ...` in notebooks for instance.

python/cugraph/tests/dask/opg_pagerank_testing.py
Iroy30: We could embed it within but it can still be initialized only once. The point would be the user to be aware of the communicator, because an explicit destroy is needed too

PR #1007, opened 2020-07-17
[REVIEW] Add tolerance support to MG Pagerank and fix kernel launch

### LINKED ISSUE #999: [ENH] Add tolerance support to MG Pagerank
Currently, the MG solver stops after a fixed number of iterations, we should support a tolerance parameter with the same behavior as for single GPU.

afender:
- Closes #999 
- Added tolerance in MG PR
- Removed error throwing when not reaching tolerance in both MG and SG. The solver now exits when the first one between max_iter and tolerance is reached.
- Fixed cuda launch bounds (so that we can accept this from the device properties/handle and not limit it with the old #define)
- The MG pagerank backend returns the number of iterations to the cpp API driver

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@
 - PR #991 Update conda upload versions for new supported CUDA/Python
 - PR #988 Add clang and clang tools to the conda env
 - PR #997 Update setup.cfg to run pytests under cugraph tests directory only
+- PR #1007 Add tolerance support to MG Pagerank and fix
 
 ## Bug Fixes
 - PR #936 Update Force Atlas 2 doc and wrapper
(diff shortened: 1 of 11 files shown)


GPUtester:

GPUtester:

Iroy30:
CI fails will be fixed with #1008

afender:
> LGTM. Maybe we need to set the same defaults for tolerance and max_iter on both python and C++ level to avoid confusion.
> Also, there's the float/double discrepancy for tolerance and alpha, and int32/64 for max_iter
> These can be fixed in a separate PR as well.

@Iroy30 c79712b added
- 64b types for it/tol as this is what we wrapped.
- python default parameters for C++

PR #1009, opened 2020-07-17
[REVIEW] Add tooltip requirements to benchmarks/build.sh [skip-ci]

ajschmidt8:
This PR adds the `requirements` field to the `benchmark-asv-metadata` flag of `pytest` in order to enable tooltips to appear on ASV graphs that contain dependency versions.

### DIFF
--- a/ci/benchmark/build.sh
+++ b/ci/benchmark/build.sh
@@ -65,26 +65,13 @@ source activate rapids
 
 
 # Enter dependencies to be shown in ASV tooltips.
-# These dependencies will also be installed by conda.
-# A dependency can exist in both arrays without causing issues.
 CUGRAPH_DEPS=(cudf rmm)
 LIBCUGRAPH_DEPS=(cudf rmm)
 
-# Concatenate dependency arrays, convert to JSON array,
-# and remove duplicates.
-X=("${CUGRAPH_DEPS[@]}" "${LIBCUGRAPH_DEPS[@]}")
-DEPS=$(printf '%s\n' "${X[@]}" | jq -R . | jq -s 'unique')
-
-# Create install args for conda (i.e. "cudf=0.15" "rmm=0.15")
-CONDA_INSTALL=
-for DEP in $(echo "${DEPS}" | jq -r '.[]'); do
-  CONDA_INSTALL+="${DEP}=${MINOR_VERSION}"
-  CONDA_INSTALL+=" "
-done
-
 logger "conda install required packages"
 conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaults \
-      ${CONDA_INSTALL} \
+      "cudf=${MINOR_VERSION}" \
+      "rmm=${MINOR_VERSION}" \
       "cudatoolkit=$CUDA_REL" \
       "dask-cudf=${MINOR_VERSION}" \
       "dask-cuda=${MINOR_VERSION}" \
@@ -125,6 +112,12 @@ if (( ${ERRORCODE} != 0 )); then
     exit ${ERRORCODE}
 fi
 
+
+# Concatenate dependency arrays, convert to JSON array,
+# and remove duplicates.
+X=("${CUGRAPH_DEPS[@]}" "${LIBCUGRAPH_DEPS[@]}")
+DEPS=$(printf '%s\n' "${X[@]}" | jq -R . | jq -s 'unique')
+
 # Build object with k/v pairs of "dependency:version"
 DEP_VER_DICT=$(jq -n '{}')
 for DEP in $(echo "${DEPS}" | jq -r '.[]'); do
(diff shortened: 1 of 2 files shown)


GPUtester:

GPUtester:

ci/benchmark/build.sh
```diff
@@ -63,10 +63,28 @@ nvidia-smi
...
+# and remove duplicates.
+X=("${CUGRAPH_DEPS[@]}" "${LIBCUGRAPH_DEPS[@]}")
+DEPS=$(printf '%s\n' "${X[@]}" | jq -R . | jq -s 'unique')
+
+# Create install args for conda (i.e. "cudf=0.15" "rmm=0.15")
+CONDA_INSTALL=
+for DEP in $(echo "${DEPS}" | jq -r '.[]'); do
+  CONDA_INSTALL+="${DEP}=${MINOR_VERSION}"
+  CONDA_INSTALL+=" "
+done
```
dillon-cullinan, line 83: I don't think we need this block.

ci/benchmark/build.sh
```diff
@@ -63,10 +63,28 @@ nvidia-smi
...
+
 logger "conda install required packages"
 conda install -c nvidia -c rapidsai -c rapidsai-nightly -c conda-forge -c defaults \
-      "cudf=${MINOR_VERSION}" \
-      "rmm=${MINOR_VERSION}" \
+      ${CONDA_INSTALL} \
```
dillon-cullinan, line 87: I would revert these lines back. It is important that the conda install lines are easily readable/changeable for the devs.

ci/benchmark/build.sh
```diff
@@ -107,17 +112,59 @@ if (( ${ERRORCODE} != 0 )); then
...
-    --benchmark-asv-output-dir=${ASVRESULTS_DIR}
+    --benchmark-asv-output-dir="${ASVRESULTS_DIR}" \
+    --benchmark-asv-metadata="${BECHMARK_META}"
+
+
 
 EXITCODE=$?
 
+# libcugraph reqs
+# REQS=$(getReqs "${LIBCUGRAPH_DEPS[@]}")
```
rlratzel, line 167: Is this comment needed for some reason or can it be removed?

ci/benchmark/build.sh
```diff
@@ -107,17 +112,59 @@ if (( ${ERRORCODE} != 0 )); then
...
 logger "Running Benchmarks..."
 cd $BENCHMARKS_DIR
 set +e
 time pytest -v -m "small and managedmem_on and poolallocator_on" \
     --benchmark-gpu-device=0 \
     --benchmark-gpu-max-rounds=3 \
-    --benchmark-asv-metadata="machineName=${NODE_NAME}, commitBranch=branch-${MINOR_VERSION}" \
-    --benchmark-asv-output-dir=${ASVRESULTS_DIR}
+    --benchmark-asv-output-dir="${ASVRESULTS_DIR}" \
+    --benchmark-asv-metadata="${BECHMARK_META}"
```
rlratzel, line 160: nice

ci/benchmark/build.sh
```diff
@@ -63,6 +63,11 @@ nvidia-smi
 logger "Activate conda env..."
 source activate rapids
 
+
+# Enter dependencies to be shown in ASV tooltips.
+CUGRAPH_DEPS=(cudf rmm)
+LIBCUGRAPH_DEPS=(cudf rmm)
```
rlratzel, line 69: You could drop the `LIBCUGRAPH_DEPS` here since we don't have benchmarks for that, and just collapse all deps into `CUGRAPH_DEPS`. Since the deps are the same for now, you'd end up just deleting `LIBCUGRAPH_DEPS`.

I'm assuming you have both as more of a placeholder for eventual C++ benches?  If so, I usually like that idea but cugraph may never have those (other repos may though), and if you choose to remove `LIBCUGRAPH_DEPS` you can probably remove some of the other code (that removes duplicates, etc.) too.  It's up to you but I'm leaning more towards removing and saving this code somewhere for use on repos that have multiple benchmark types.

ci/benchmark/build.sh
```diff
@@ -107,17 +112,59 @@ if (( ${ERRORCODE} != 0 )); then
...
+
+
 
 EXITCODE=$?
 
+# libcugraph reqs
+# REQS=$(getReqs "${LIBCUGRAPH_DEPS[@]}")
+
 set -e
 JOBEXITCODE=0
```
rlratzel, line 170: I know you didn't add this but I'm wondering if we need to eventually let jobs return non-zero when there's a legit error (eg. conda couldn't install or something)

ci/benchmark/build.sh
rlratzel: oh, I think I see, I guess this is also a placeholder?  I usually like that but cugraph will (probably) never have C++ benches.  If you choose to leave it, Can you add a comment to the future maintainer that explains it should be uncommented and used when libcugraph benchmarks are added?

ci/benchmark/build.sh
```diff
@@ -107,17 +112,59 @@ if (( ${ERRORCODE} != 0 )); then
...
+    VER=$(echo "${DEP_VER_DICT}" | jq -r --arg DEP "${DEP}" '.[$DEP]')
+    REQS=$(echo "${REQS}" | jq -c --arg DEP "${DEP}" --arg VER "${VER}" '. + { ($DEP): $VER }')
+  done
+
+  echo "${REQS}"
+}
+
+REQS=$(getReqs "${CUGRAPH_DEPS[@]}")
+
+BECHMARK_META=$(jq -n \
```
dillon-cullinan, line 142:
```suggestion
BENCHMARK_META=$(jq -n \
```

ci/benchmark/build.sh
dillon-cullinan, line 160:
```suggestion
    --benchmark-asv-metadata="${BENCHMARK_META}"
```

ci/benchmark/build.sh
ajschmidt8: good catch.

ci/benchmark/build.sh
ajschmidt8: i'm gonna leave this one up to @dillon-cullinan. i don't think that needs to be addressed in this PR.

ci/benchmark/build.sh
ajschmidt8: i'll add a comment about this

ci/benchmark/build.sh
ajschmidt8: i'm going to leave this for now. there's some code that appears later on in the script that depends on `LIBCUGRAPH_DEPS` and I don't want to remove that also since it will serve as a good template for copying this script to other repos that do have python and C++ benchmarks.

PR #1020, opened 2020-07-23
[ENH] Updated Louvain to honor max_level, ECG now calls Louvain for 1 level, then full run

### LINKED ISSUE #915: [ENH] ECG implementation not quite right
The old implementation of ECG (based on the old implementation of Louvain) was not generating good clustering results consistently.  As a result, it was difficult to validate that ECG was generating the correct results when I modified it to remove gdf_column and RMM_ALLOC references.

The ECG implementation needs to stop at 1 level of the Louvain dendrogram.  Based on the description from the ECG paper, the algorithm should run until it would generate super vertices and then it should return the answer.

Suggest the following:
1. Rename the Louvain max_iter parameter to be max_level.  Don't it expose it to the python layer
1. Modify Louvain to use the max_level and return at the appropriate time
1. The final Louvain call in ECG should do a full run, not the partial run
1. Test that ECG is generating good results

rlratzel:
Changed Louvain param `max_iter` to `max_level` and updated code to use that param in determining when to return, ECG now calls Louvain for 1 level, then full run, as described in https://arxiv.org/pdf/1809.05578.pdf and issue #915 

Also applied changes to clean up inconsistent type names (eg. `VT` -> `vertex_t`) and removed old comments.

closes #915


GPUtester:

GPUtester:

PR #1022, opened 2020-07-24
[REVIEW] Fix support for using a cudf.DataFrame with a MG graph

ChuckHastings:
Fix bug found in testing MG pagerank.  Mixing cudf.DataFrame and dask_cudf.DataFrame wasn't working properly in NumberMap.

Also, cudf dropped support for tolist, which breaks one of our unit tests.  Changed to use the suggested method in the error message.


ChuckHastings:
rerun tests

PR #1025, opened 2020-07-29
[REVIEW] Explicitly skip raft test folder for pytest 6.0.0 [skip-ci]

dantegd:
Similar to https://github.com/rapidsai/cuml/pull/2619, there is a very high chance CI in cuGraph will run into the same issue as cuML now with pytest 6.0.0. 

The PR also changes `py.test` (which is the old school invocation) to the current `pytest` command.


GPUtester:

dantegd:
@afender I would leave both just to keep CI unblocked, then we can look whether the conftest/skip flag are both needed in a follow up, what do you think?

dantegd:
rerun tests

dantegd:
there seems to be an (unrelated) test failing in cuda 10.1 CI: 
```
FAILED cugraph/tests/test_balanced_cut.py::test_edge_cut_clustering[2-../datasets/dolphins.csv]
```
is that a known issue?

dantegd:
rerun tests

PR #1034, opened 2020-08-05
[REVIEW] Expose resolution (gamma) parameter in Louvain

### LINKED ISSUE #674: [FEA] Expose gamma (resolution) parameter for Louvain.
This is to expose the gamma parameter (in the formula for modularity score) as a parameter the user can specify.

Since currently Louvain assumes a value of 1 for gamma some of the code will need to be rewritten in order to use a different value of gamma.

Modularity and the Louvain implementation are known to suffer from a resolution issues. To address that, iGraph has an option to set a resolution limit. That same feature should be added to our version.

https://scanpy.readthedocs.io/en/stable/api/scanpy.tl.louvain.html

Modularity is defined as
image

This modification is to allow the resolution parameter, y, to be specified in the range of 0.0001 to 1.0

ChuckHastings:
Add a resolution parameter (gamma in the mathematical formulation) to the Louvain score.

Closes #674 

Includes an update to the delta_Q computation


GPUtester:

ChuckHastings:
rerun tests

cpp/include/algorithms.hpp
```diff
@@ -605,24 +605,25 @@ void bfs(raft::handle_t const &handle,
...
-             WT *final_modularity,
+template <typename vertex_t, typename edge_t, typename weight_t>
+void louvain(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+             weight_t *final_modularity,
              int *num_level,
-             VT *louvain_parts,
-             int max_iter = 100);
+             vertex_t *louvain_parts,
+             int max_iter        = 100,
+             weight_t resolution = weight_t{1});
```
rlratzel, line 626: I'm assuming this default means we don't need to update the calls to Louvain in ECG, and that the default value is fine for ECG?

ChuckHastings: The default of 1 makes Louvain behave just as it did before - which should be fine for ECG.

PR #1036, opened 2020-08-06
[BUG] Fixed benchmarks for new renumbering API, updated comments, added quick test-only benchmark run to CI

### LINKED ISSUE #1035: [BUG] benchmarks need to be updated for new renumbering API
The benchmarks needed to be updated to match the updates to the renumbering API.  In addition to simply updating them, they should also be run in a fast "test-only" mode by CI so future breakages don't go unnoticed.

rlratzel:
* Fixed benchmarks for new renumbering API
* Updated docstrings for clarity and added minor comments, also changed var names in one case to match docstring.
* Added quick test-only benchmark run to CI so breakages won't go unnoticed.

closes #1035


GPUtester:

PR #1057, opened 2020-08-10
[REVIEW] Updated raft dependency to pull fixes on cusparse selection in CUDA 11.

aschaffer:
This addresses the problem of incorrect selection of deprecated cusparse primitives due to usage of `nvcc`-injected macros (`__CUDACC_VER_MAJOR/MINOR__`). 

Those would have required changing the *.cpp extensions to *.cu. Instead, `CUDART_VERSION` is now used which doesn't require `nvcc` involvement. Hence, no need for touching file extensions on so many source files.


GPUtester:

BradReesWork:
rerun tests

PR #1063, opened 2020-08-11
[REVIEW] Leiden

### LINKED ISSUE #552: [FEA] Leiden Algorithm for Community Detection
I am currently trying to apply graph based community detection for one of the case studies in my organization.

I have used Louvain's Algorithm available in CUGRAPH. However it spits out lot of bigger communities ( multiple communities merged) 

Leiden algorithm seems to be a better fit as it was tested in CPU. Due to the high volume of data, we want to run this in GPU. Can we have a Leiden implementation as well in CUGRAPH ?

Let me know if it is feasible.

Leiden Algorithm for Community Detection Literature Link:
https://www.nature.com/articles/s41598-019-41695-z

jwyles:
This PR includes the implementation of the Leiden algorithm. It's based on Louvain and re-uses most of the code from Louvain (after a little formatting).

There is a known issue with Leiden and the dolphins data set. It should produce a clustering which has higher modularity than Louvain, but it does not it is significantly lower instead.

closes #552


GPUtester:

cpp/include/algorithms.hpp
```diff
@@ -625,6 +625,34 @@ void louvain(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
...
+ * @param[in]  max_iter              (optional) maximum number of iterations to run (default 100)
+ */
+template <typename vertex_t, typename edge_t, typename weight_t>
+void leiden(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+            weight_t *final_modularity,
+            int *num_level,
+            vertex_t *leiden_parts,
+            int max_iter        = 100,
+            weight_t resolution = weight_t{1});
+
```
BradReesWork, line 655: great job with documentation but:
* the @param argument names should match what is in the function definition 
* You are missing an entry for resolution

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+  rmm::device_vector<vertex_t> &constraint,
+  cudaStream_t stream)
+{
+  rmm::device_vector<vertex_t> next_cluster(cluster);
+  rmm::device_vector<weight_t> delta_Q(graph.number_of_edges);
+  rmm::device_vector<vertex_t> cluster_hash(graph.number_of_edges);
+  rmm::device_vector<weight_t> old_cluster_sum(graph.number_of_vertices);
+
+  vertex_t *d_cluster_hash         = cluster_hash.data().get();
+  vertex_t *d_cluster              = cluster.data().get();
```
BradReesWork, line 53: you define d_cluster but I cannot see where it is used.  there is a call to cluster.data().get()

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+    weight_t new_Q = update_clustering_by_delta_modularity(total_edge_weight,
+                                                           resolution,
+                                                           current_graph,
+                                                           src_indices_v,
+                                                           vertex_weights_v,
+                                                           cluster_weights_v,
+                                                           cluster_v,
+                                                           stream);
+
+    std::cout << "Level " << *num_level << " initial partitioning modularity: " << new_Q << "\n";
```
BradReesWork, line 229: this should be wrapped in a debug define so that we are not generating a lot of extra messages in production

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+                                                              resolution,
+                                                              current_graph,
+                                                              src_indices_v,
+                                                              vertex_weights_v,
+                                                              cluster_weights_v,
+                                                              cluster_v,
+                                                              constraint,
+                                                              stream);
+
+    std::cout << "Level " << *num_level << " constrained partitioning modularity: " << new_Q
```
BradReesWork, line 246: wrap print in debug block

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+  vertex_t *d_cluster_hash         = cluster_hash.data().get();
+  vertex_t *d_cluster              = cluster.data().get();
+  weight_t const *d_vertex_weights = vertex_weights.data().get();
+  weight_t *d_cluster_weights      = cluster_weights.data().get();
+  weight_t *d_delta_Q              = delta_Q.data().get();
+  vertex_t *d_constraint           = constraint.data().get();
+  vertex_t const *d_src_indices    = src_indices.data().get();
+  vertex_t const *d_dst_indices    = graph.indices;
+
+  weight_t new_Q = modularity<vertex_t, edge_t, weight_t>(
```
aschaffer, line 61: Explicit template arguments should _not_ be necessary. The compiler should be able to deduce the `vertex_t, edge_t, weight_t` arguments from the `graph` argument. Explicit template arguments defeat the purpose of static polymorphism (via templates).

aschaffer: Example below should compile and link just fine:
```
#include <iostream>

using cudaStream_t = int;

template<typename vertex_t, typename edge_t, typename weight_t>
struct GraphCSRView
{
};

weight_t modularity(weight_t total_edge_weight,
                    weight_t resolution,
                    GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
                    vertex_t const *d_cluster,
                    cudaStream_t stream = 0)
{
    std::cout<<"modularity...\n";
    return resolution;
}

void foo(GraphCSRView<vertex_t, edge_t, weight_t> const &graph)
{
    vertex_t* d_cluster{nullptr};
    modularity(1.0, 2.0, graph, d_cluster); // compiler deduces the template args from the `graph` argument
}

int main(void)
{
    GraphCSRView<int, int, double> graph;
    foo(graph);

    return 0;
}
```

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+  rmm::device_vector<int32_t> const &,
+  rmm::device_vector<double> const &,
+  rmm::device_vector<double> &,
+  rmm::device_vector<int32_t> &,
+  rmm::device_vector<int32_t> &,
+  cudaStream_t);
+
+template <typename vertex_t, typename edge_t, typename weight_t>
+void leiden(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+            weight_t *final_modularity,
```
aschaffer, line 142: We should avoid proliferating use of raw pointers. We should use smart pointers (like `std::uniqie_ptr`) or, for device pointers we should use wrappers like `device_vector<>`, etc. 

Fixing this everywhere in our code is beyond the scope of this PR, obviously, but perhaps we could progressively address this issue, each of us in their own PRs.

cpp/src/community/leiden_kernels.cu
```diff
@@ -0,0 +1,302 @@
...
+            int *num_level,
```
aschaffer, line 143: `num_level` should be passed by reference, rather than a pointer to a non-array memory location. 

Perhaps even more readable would be to pass mutables and results like `num_level` as (part of a tuple of) return values.

cpp/src/community/leiden_kernels.cu
aschaffer, line 142: `final_modularity` should be passed by reference, rather than a pointer to a non-array memory location. 

Perhaps even more readable would be to pass mutables and results like `final_modularity` as (part of a tuple of) return values.

cpp/src/community/leiden_kernels.cu
jwyles: Removed the print statement.

cpp/src/community/leiden_kernels.cu
jwyles: Removed the print statement

cpp/src/community/leiden_kernels.cu
jwyles: It (along with three others) wasn't being used and has been removed.

cpp/include/algorithms.hpp
jwyles: Updated.

cpp/src/community/leiden_kernels.cu
jwyles: Updated to not have explicit template arguments.

cpp/src/community/leiden_kernels.cu
jwyles: I changed the pointers for final_modularity and num_level to be passed by reference. I am leaving the output partitioning as a raw pointer, since that is what we are doing to avoid dependencies on libcudf.

cpp/src/community/leiden_kernels.cu
jwyles: Changed.

cpp/src/community/leiden_kernels.cu
jwyles: Changed.

BradReesWork:
rerun tests

python/cugraph/community/leiden.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+    Compute the modularity optimizing partition of the input graph using the
+    Louvain heuristic
+
+    Parameters
+    ----------
+    input_graph : cugraph.Graph
+        cuGraph graph descriptor of type Graph
+
+        The adjacency list will be computed if not already present. The graph
```
afender, line 28: "where an undirected edge is represented by a directed edge in both direction." : this is now taken care of in the `Graph` class. User doesn't need to worry or know about this

python/cugraph/community/leiden.py
```diff
@@ -0,0 +1,78 @@
...
+        should be undirected where an undirected edge is represented by a
+        directed edge in both direction.
+
+    max_iter : integer
+        This controls the maximum number of levels/iterations of the Louvain
```
afender, line 33: Louvain -> Leiden

python/cugraph/community/leiden.py
```diff
@@ -0,0 +1,78 @@
...
+        algorithm. When specified the algorithm will terminate after no more
+        than the specified number of iterations. No error occurs when the
+        algorithm terminates early in this manner.
+
+    resolution: float/double, optional
+        Called gamma in the modularity formula, this changes the size
```
afender, line 39: I think we can improve the doc of this parameter by answering:
- What formula/paper are we referring to? 
- What are the acceptable values? 
- How does this impact the algo?

python/cugraph/community/leiden.py
```diff
@@ -0,0 +1,78 @@
...
+        of the communities.  Defaults to 1.
+
+    Returns
+    -------
+    parts : cudf.DataFrame
+        GPU data frame of size V containing two columns the vertex id and the
+        partition id it is assigned to.
```
afender, line 46: partition, cluster, or communities? We should use consistent terminology in public API.

PR #1081, opened 2020-08-14
[WIP] MNMG Renumbering - sort partitions by degree

### LINKED ISSUE #1048: [ENH] MNMG Renumbering including sort optimization
in support of 2D partitioning.
Support |V| > 2 billion

ChuckHastings:
This PR modifies the renumbering logic to sort vertices by degree.  The consequence of this will be that the lowest vertex id assigned by each partition will be the highest degree vertex assigned to that partition.

closes #1048

Note, this PR applies the logic to both SNMG and MNMG renumbering.

### DIFF
--- a/python/cugraph/community/ecg.py
+++ b/python/cugraph/community/ecg.py
@@ -66,6 +66,6 @@ def ecg(input_graph, min_weight=0.05, ensemble_size=16):
     parts = ecg_wrapper.ecg(input_graph, min_weight, ensemble_size)
 
     if input_graph.renumbered:
-        parts = input_graph.unrenumber(parts, "vertex")
+        parts = input_graph.unrenumber(parts, "vertex", preserve_order=True)
 
     return parts
--- a/python/cugraph/structure/graph.py
+++ b/python/cugraph/structure/graph.py
@@ -16,6 +16,7 @@ from cugraph.structure.symmetrize import symmetrize
 from cugraph.structure.number_map import NumberMap
 from cugraph.dask.common.input_utils import get_local_data
 import cudf
+import numpy as np
 import dask_cudf
 import warnings
 
@@ -87,6 +88,7 @@ class Graph:
         """
         self.symmetrized = symmetrized
         self.renumbered = False
+        self.renumber_map = None
         self.bipartite = False
         self.multipartite = False
         self._nodes = {}
@@ -207,6 +209,7 @@ class Graph:
         destination="destination",
         edge_attr=None,
         renumber=True,
+        store_row_major=True,
     ):
         """
         Initialize a graph from the edge list. It is an error to call this
@@ -249,6 +252,13 @@ class Graph:
         renumber : bool
             If source and destination indices are not in range 0 to V where V
             is number of vertices, renumber argument should be True.
+        store_row_major : bool
+            Identify how the graph adjacency will be used.  Graphs stored in
+            row major order will be optimized for organization by source
+            vertex.  Graphs stored in column major order will be optimized
+            for ogranization by destination vertex.
+            If True, store in row major order, if False, store in column
+            major order.
 
         Examples
         --------
@@ -277,14 +287,18 @@ class Graph:
             raise Exception('input should be a cudf.DataFrame or \
                               a dask_cudf dataFrame')
 
+        self.store_row_major=store_row_major
+        
         renumber_map = None
         if renumber:
             elist, renumber_map = NumberMap.renumber(
-                elist, source, destination
+                elist, source, destination,
+                store_row_major=store_row_major
             )
             source = 'src'
             destination = 'dst'
             self.renumbered = True
+            self.renumber_map = renumber_map
         else:
             if type(source) is list and type(destination) is list:
                 raise Exception('set renumber to True for multi column ids')
@@ -314,7 +328,6 @@ class Graph:
         self.edgelist = Graph.EdgeList(
             source_col, dest_col, value_col
         )
-        self.renumber_map = renumber_map
 
     def add_edge_list(self, source, destination, value=None):
         warnings.warn(
@@ -332,7 +345,8 @@ class Graph:
 
     def from_dask_cudf_edgelist(self, input_ddf, source='source',
                                 destination='destination',
-                                edge_attr=None, renumber=True):
+                                edge_attr=None, renumber=True,
+                                store_row_major=True):
         """
         Initializes the distributed graph from the dask_cudf.DataFrame
         edgelist. Undirected Graphs are not currently supported.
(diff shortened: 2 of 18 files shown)


GPUtester:

aschaffer:
LGTM.

python/cugraph/structure/graph.py
```diff
@@ -1113,10 +1130,14 @@ def has_edge(self, u, v):
...
+                tmp, 'id', 'src', preserve_order=True
+            )
 
-            u = tmp[0]
-            v = tmp[1]
+            u = tmp['id'][0]
+            v = tmp['id'][1]
 
         df = self.edgelist.edgelist_df
         if self.distributed:
```
seunghwak, line 1143: Clearly outside the scope of this PR, but we need to inspect the performance of this approach. Ideally, we can figure out which GPU can possibly hold (u, v), and in the target GPU, we just need to check ```u```'s adjacency list (if store_transposed = False). The code below sounds much more expensive than this (and does this assume availability of COO?, are we storing COO even after graph creation? This more than halves the maximum graph size we can handle)

python/cugraph/structure/graph.py
```diff
@@ -292,6 +293,7 @@ def from_cudf_edgelist(
         destination="destination",
         edge_attr=None,
         renumber=True,
+        store_transposed=False,
```
afender, line 296: The naming and effect of this flag can be confusing because we have been automatically storing both transposed and non-transposed (as needed) in SG path since cugraph's first release.
Perhaps `runmber_by=src/dst` would be clearer.

python/cugraph/structure/graph.py
```diff
@@ -445,6 +456,10 @@ def from_dask_cudf_edgelist(self, input_ddf, source='source',
         renumber : bool
             If source and destination indices are not in range 0 to V where V
             is number of vertices, renumber argument should be True.
+        store_transposed : bool
```
afender, line 459: Do we really need to re-pass flag here in addition to the graph constructor? 
It seems this could lead to undefined behaviour if we pass Ture in the former and False in the later and vice versa.

python/cugraph/structure/number_map.py
```diff
@@ -268,33 +303,41 @@ def compute(self):
...
                 )
 
+                val_types.pop('count')
+
                 numbering_map = rehashed_with_partition_id.map_partitions(
-                    lambda df: df.groupby(self.col_names).min().reset_index()
+                    lambda df: df.groupby(
+                        self.col_names + ["hash", "partition"]
+                    ).sum()
+                    .sort_values('count', ascending=False)
```
afender, line 319: Do we have perf data associated with this change at this time? It would be helpful to add nightly MG performance regression tests. This would help to keep an eye on this so we (devs) don't have to track that manually.

afender: Opened an issue there https://github.com/rapidsai/cugraph/issues/1085

python/cugraph/structure/graph.py
ChuckHastings: @seunghwak and I have waffled on naming this parameter.  We could do this instead.  Might be a bit clearer.

python/cugraph/structure/graph.py
ChuckHastings: It isn't in the graph constructor, it is in the `from_cudf_edgelist` and `from_dask_cudf_edgelist` calls.  If we were to call one and then the other that would raise an exception.

python/cugraph/structure/number_map.py
ChuckHastings: We don't have any performance results at this time.

python/cugraph/structure/graph.py
seunghwak: Before jumping to the naming, we need to make our strategy clear on storing both transposed and non-transposed (and actually edge list as well). This will largely cut the maximum graph size we can handle. I think we need to decide between "cuGraph becoming less magical" vs "largely reducing the maximum graph size cuGraph can handle".

seunghwak: And if we want to enable users to fit larger graphs, we need to explicitly ask users to pick the format. ```renumber_by=src/dst``` is not explicit in this regard.

python/cugraph/community/ecg.py
```diff
@@ -66,6 +66,6 @@ def ecg(input_graph, min_weight=0.05, ensemble_size=16):
     parts = ecg_wrapper.ecg(input_graph, min_weight, ensemble_size)
 
     if input_graph.renumbered:
-        parts = input_graph.unrenumber(parts, "vertex")
+        parts = input_graph.unrenumber(parts, "vertex", preserve_order=True)
```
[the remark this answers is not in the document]
ChuckHastings: I will check.  I added preserve_order initially to make a bunch of tests pass.  Then I fixed the spectral_clustering implementation, I probably forgot to remove this one.

python/cugraph/structure/number_map.py
seunghwak: Eventually, we need to implement fully optimized (most likely C++) version of this logic. Our python based approach can be a quick first step, but this doesn't look performance optimal.

python/cugraph/community/ecg.py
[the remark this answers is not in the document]
ChuckHastings: Just pushed an update to remove this.  It is working with the spectral clustering API change I had previously made.

python/cugraph/structure/number_map.py
seunghwak: https://github.com/rapidsai/cugraph/issues/1048#issuecomment-672907741

and it will be great we can check this code can handle the common crawl data (possibly in a separate issue/PR) before closing #1048 .

python/cugraph/community/ecg.py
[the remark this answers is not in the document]
seunghwak: There are few more places I saw ```preserve_order=True```. Make sure we're removing all the unnecessary ones.

python/cugraph/structure/graph.py
ChuckHastings: We will need a new version of `has_edge`.  Do we want to create an issue for that specifically, or will we naturally do that as we test the python graph object in MNMG mode?  Not sure if we want to create a large number of little issues as we see them or trust that when we try to use it we'll notice :-)

seunghwak: Yeah... just ignore this comment. It's clearly outside the scope, but just wanted to raise this issue to make sure more people aware of the problem.

ChuckHastings:
> And just to double check,
> 
> So, assume V vertices and P = P_row * P_column GPUs, this sorts vertices based on global degree within a partition having (roughly) V / P vertices, correct?

Yes

python/cugraph/structure/graph.py
```diff
@@ -334,6 +336,10 @@ def from_cudf_edgelist(
...
+            Identify how the graph adjacency will be used.
+            If True, the graph will be organized by destination.
+            If False, the graph will be organized by source
```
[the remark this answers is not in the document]
ChuckHastings: Created #1095 to address the overall documentation issue.

Will add some better text to this parameter definition.

PR #1098, opened 2020-08-19
[REVIEW] A new graph class to support 2D graph partitioning

seunghwak:
This PR is now ready for review.

This PR has NOT been fully tested (in particular, the multi-GPU part), but I am trying to merge this first and test more with the pattern accelerator PR as

This PR is not related to any user features yet,
several other works (e.g. @Iroy30 needs this to work on python C++ interface) dependent on this PR,
and we don't currently have a mechanism for C++ only multi-GPU testing (multi-GPU testing requires python binding).


GPUtester:

seunghwak:
@Iroy30 @ChuckHastings

So, to create a graph object from python (cython)

Cython should be able to call the graph constructor below

https://github.com/rapidsai/cugraph/pull/1098/files#diff-38e6450cfddac68f769d0eead79fee69R123
```
  graph_t(raft::handle_t const &handle,
          std::vector<edgelist_t<vertex_t, edge_t, weight_t>> const &edge_lists,
          partition_t<vertex_t> const &partition,
          vertex_t number_of_vertices,
          edge_t number_of_edges,
          bool is_symmetric,
          bool is_multigraph,
          bool is_weighted,
          bool sorted_by_global_degree_within_vertex_partition,
```

and to get the graph data back to python (cython)

we need to take memory out from member variables defined in

https://github.com/rapidsai/cugraph/pull/1098/files#diff-38e6450cfddac68f769d0eead79fee69R185
```
  std::vector<rmm::device_uvector<edge_t>> adj_matrix_partition_offsets_{};
  std::vector<rmm::device_uvector<vertex_t>> adj_matrix_partition_indices_{};
  std::vector<rmm::device_uvector<weight_t>> adj_matrix_partition_weights_{};
  std::vector<vertex_t>
    vertex_partition_segment_offsets_{};  // segment offsets within the vertex partition based on
                                          // vertex degree, relevant only if
                                          // sorted_by_global_degree_within_vertex_partition is true
```

Do you see any issues with the current interface?

GPUtester:

GPUtester:

cpp/include/experimental/graph.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+template <typename vertex_t, typename edge_t, typename weight_t>
+struct edgelist_t {
+  vertex_t const *p_src_vertices{nullptr};
+  vertex_t const *p_dst_vertices{nullptr};
+  weight_t const *p_edge_weights{nullptr};
+  edge_t number_of_edges{0};
+};
+
+template <typename vertex_t,
```
ChuckHastings, line 40: Not sure the best place to put it, but it would be useful to document the difference between graph_t and graph_view_t somewhere in the definition of the classes.

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+
+  CUGRAPH_EXPECTS((partition.hypergraph_partitioned &&
+                   (edgelists.size() == static_cast<size_t>(comm_p_row_size))) ||
+                    (!(partition.hypergraph_partitioned) && (edgelists.size() == 1)),
+                  "Invalid API parameter: errneous edgelists.size().");
+
+  CUGRAPH_EXPECTS(partition.vertex_partition_offsets.size() == static_cast<size_t>(comm_p_size),
+                  "Invalid API parameter: erroneous partition.vertex_partition_offsets.size().");
+
+  // optinoal expensive checks (part 1/3)
```
ChuckHastings, line 250: `optinoal` should be `optional`

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+
+  if (do_expensive_check) {
+    edge_t number_of_local_edges_sum{};
+    for (size_t i = 0; i < edgelists.size(); ++i) {
+      auto major_first =
+        partition.hypergraph_partitioned
```
ChuckHastings, line 256: Consider making this (and the other ternary operators below) a member function of the partition object.  Might make the code a bit cleaner.

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+    if (this->is_weighted()) { adj_matrix_partition_weights_.push_back(std::move(weights)); }
+  }
+
+  // update degree-based segment offsets (to be used for graph analytics kernel optimization)
+
+  if (sorted_by_global_degree_within_vertex_partition) {
+    auto degrees = detail::compute_major_degree(
+      *(this->get_handle_ptr()), adj_matrix_partition_offsets_, partition_);
+
+    // optinoal expensive checks (part 2/3)
```
ChuckHastings, line 328: `optinoal` should be `optional`

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+                         default_stream);
+
+    vertex_partition_segment_offsets_.resize(comm_p_row_size * (segment_offsets.size()));
+    raft::update_host(vertex_partition_segment_offsets_.data(),
+                      aggregate_segment_offsets.data(),
+                      aggregate_segment_offsets.size(),
+                      default_stream);
+  }
+
+  // optinoal expensive checks (part 3/3)
```
ChuckHastings, line 376: `optinoal` should be `optional`

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+  auto num_elements = thrust::distance(degree_first, degree_last);
+  auto num_segments = thrust::distance(threshold_first, threshold_last) + 1;
+
+  std::vector<vertex_t> h_segment_offsets(num_segments + 1);
+  h_segment_offsets[0]     = 0;
+  h_segment_offsets.back() = num_elements;
+
+  rmm::device_uvector<vertex_t> d_segment_offsets(num_segments - 1, handle.get_stream());
+
+  thrust::upper_bound(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
```
aschaffer, line 182: CAVEAT: thrust `lower_bound()` / `upper_bound()` / `binary_search()`: https://github.com/thrust/thrust/pull/1104

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+          bool is_weighted,
+          bool sorted_by_global_degree_within_vertex_partition,
+          bool do_expensive_check)
+  : detail::graph_base_t<vertex_t, edge_t, weight_t>(
+      handle, number_of_vertices, number_of_edges, is_symmetric, is_multigraph, is_weighted),
+    partition_(partition)
+{
+  // cheap error checks
+
+  auto &comm_p         = this->get_handle_ptr()->get_comms();
```
aschaffer, line 221: Similar comment as above: please consider `const`-ifying whatever can be `const` in the series of `auto`'s below. If nothing can be `const`, please ignore.

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+  auto comm_p_size     = comm_p.get_size();
+  auto &comm_p_row     = this->get_handle_ptr()->get_subcomm(comm_p_row_key);
+  auto comm_p_row_rank = comm_p_row.get_rank();
+  auto comm_p_row_size = comm_p_row.get_size();
+  auto &comm_p_col     = this->get_handle_ptr()->get_subcomm(comm_p_col_key);
+  auto comm_p_col_rank = comm_p_col.get_rank();
+  auto comm_p_col_size = comm_p_col.get_size();
+  auto default_stream  = this->get_handle_ptr()->get_stream();
+
+  for (size_t i = 0; i < edgelists.size(); ++i) {
```
aschaffer, line 231: Perhaps replace the `for` loop with a `find_if()` on the `edgelist` container?

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+
+  // update degree-based segment offsets (to be used for graph analytics kernel optimization)
+
+  if (sorted_by_global_degree_within_vertex_partition) {
+    auto degrees = detail::compute_major_degree(
+      *(this->get_handle_ptr()), adj_matrix_partition_offsets_, partition_);
+
+    // optinoal expensive checks (part 2/3)
+
+    if (do_expensive_check) {
```
aschaffer, line 330: Perhaps rather than always check flags like this at runtime, a template policy `typename validity_checker_t = pass_through_check_t` could be added. It encapsulates all such (expensive) checks; and it defaults to a "pass through" policy that does nothing (i.e., all the corresponding member function checkers are "no-op"'s).

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+    raft::update_device(
+      d_thresholds.data(), h_thresholds.data(), h_thresholds.size(), default_stream);
+
+    rmm::device_uvector<vertex_t> segment_offsets(detail::num_segments_per_vertex_partition + 1,
+                                                  default_stream);
+    segment_offsets.set_element_async(0, 0, default_stream);
+    segment_offsets.set_element_async(
+      detail::num_segments_per_vertex_partition, degrees.size(), default_stream);
+
+    thrust::upper_bound(rmm::exec_policy(default_stream)->on(default_stream),
```
aschaffer, line 355: See CAVEAT: thrust lower_bound() / upper_bound() / binary_search(): thrust/thrust#1104 (although vectorized versions are okay, I think)

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,543 @@
...
+                        segment_offsets.begin() + 1);
+
+    segment_offsets_.resize(segment_offsets.size());
+    raft::update_host(
+      segment_offsets_.data(), segment_offsets.data(), segment_offsets.size(), default_stream);
+  }
+
+  // optinoal expensive checks (part 3/3)
+
+  if (do_expensive_check) {
```
aschaffer, line 505: See comment above on (expensive) validity checks.

cpp/src/experimental/graph_view.cu
```diff
@@ -0,0 +1,316 @@
...
+                     (vertex_partition_segment_offsets.size() == 0)),
+                  "Invalid API parameter: vertex_partition_segment_offsets.size() does not match "
+                  "with sorted_by_global_degree_within_vertex_partition.");
+
+  CUGRAPH_EXPECTS(partition.vertex_partition_offsets.size() == static_cast<size_t>(comm_p_size),
+                  "Invalid API parameter: erroneous partition.vertex_partition_offsets.size().");
+
+  // optinoal expensive checks
+
+  if (do_expensive_check) {
```
aschaffer, line 113: See comment above on (expensive) validity checks.

cpp/src/experimental/graph_view.cu
```diff
@@ -0,0 +1,316 @@
...
+    auto default_stream = this->get_handle_ptr()->get_stream();
+
+    auto comm_p_row_rank = this->get_handle_ptr()->get_subcomm(comm_p_row_key).get_rank();
+    auto comm_p_col_rank = this->get_handle_ptr()->get_subcomm(comm_p_col_key).get_rank();
+
+    edge_t number_of_local_edges_sum{};
+    for (size_t i = 0; i < adj_matrix_partition_offsets.size(); ++i) {
```
aschaffer, line 120: Consider caching the `size()`. For a variety of (STL-like) containers, `size()` can be an expensive call.

Even better, if you _can_ `for_each() / transform()` with iterators on `adj_matrix_partition_offsets` that would lead to more readable code (e.g., by refactoring the checkers into separate functors that can be sequentially called on each iterator instance)

cpp/src/experimental/graph_view.cu
```diff
@@ -0,0 +1,316 @@
...
+
+  CUGRAPH_EXPECTS(
+    (sorted_by_degree &&
+     (segment_offsets.size() == (detail::num_segments_per_vertex_partition + 1))) ||
+      (!sorted_by_degree && (segment_offsets.size() == 0)),
+    "Invalid API parameter: segment_offsets.size() does not match with sorted_by_degree.");
+
+  // optinoal expensive checks
+
+  if (do_expensive_check) {
```
aschaffer, line 245: See comment above on (expensive) validity checks.

cpp/include/experimental/graph.hpp
```diff
@@ -0,0 +1,198 @@
...
+#include <string>
+#include <type_traits>
+#include <vector>
+
+namespace cugraph {
+namespace experimental {
+
+template <typename vertex_t, typename edge_t, typename weight_t>
+struct edgelist_t {
+  vertex_t const *p_src_vertices{nullptr};
```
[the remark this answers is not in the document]
seunghwak: This is .hpp file to be included by .cpp file (compiled with g++). thrust_device_ptr requires nvcc.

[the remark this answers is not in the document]
aschaffer: Sure, please ignore them.

cpp/include/experimental/graph.hpp
seunghwak: Added

// graph_t is an owning graph class (note that graph_view_t is a non-owning graph class)

and 

https://github.com/rapidsai/cugraph/pull/1098/files#diff-75b979f478be47f71bd6933f474074cbR128

as well

cpp/include/experimental/graph_view.hpp
```diff
@@ -0,0 +1,274 @@
...
+  vertex_t get_number_of_local_vertices() const
+  {
+    auto comm_p_rank = this->get_handle_ptr()->get_comms().get_rank();
+    return partition_.vertex_partition_offsets[comm_p_rank + 1] -
+           partition_.vertex_partition_offsets[comm_p_rank];
+  }
+
+  size_t get_number_of_adj_matrix_partitions() { return adj_matrix_partition_offsets_.size(); }
+
+  // Better avoid direct invocation in application code.
```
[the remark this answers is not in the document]
seunghwak: Yeah... this makes sense, but the problem is the legacy code. They currently directly accesses offsets, indices, and weights arrays.

I will update comments to better explain this and how this will change in the future.

Updated the comments to reflect this.

// FIXME: this function is not part of the public stable API.This function is mainly for pattern
  // accelerator implementation. This function is currently public to support the legacy
  // implementations directly accessing CSR/CSC data, but this function will eventually become
  // private or even disappear if we switch to CSR + DCSR (or CSC + DCSC).

cpp/include/experimental/graph_view.hpp
```diff
@@ -0,0 +1,274 @@
...
+std::string const comm_p_row_key = "comm_p_row";
+std::string const comm_p_col_key = "comm_p_key";
+
+/**
+ * @brief store vertex partitioning map
+ *
+ * We need to partition 1D vertex arrays (storing per vertex values) and the 2D graph adjacency
+ * matrix (or transposed 2D graph adjacency matrix) of G. An 1D vertex array of size V is divided to
+ * P linear partitions; each partition has the size close to V / P. The 2D matrix is partitioned to
+ * 1) P or 2) P * P_col rectangular partitions.
```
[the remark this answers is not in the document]
seunghwak: Done.

cpp/include/experimental/graph_view.hpp
```diff
@@ -0,0 +1,274 @@
...
+ *
+ * 1) is the default. One GPU will be responsible for 1 rectangular partition. The matrix will be
+ * horizontally partitioned first to P_row slabs (by combining P_col vertex partitions). Each slab
+ * will be further vertically partitioned to P_col rectangles (by combining P_row vertex
+ * partitions). Each rectangular partition will have the size close to V / P_row by V / P_col.
+ *
+ * 2) is to support hyper-graph partitioning. We may apply hyper-graph partitioning to divide V
```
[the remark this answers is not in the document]
seunghwak: Done.

cpp/include/experimental/graph_view.hpp
```diff
@@ -0,0 +1,274 @@
...
+    auto comm_p_rank = this->get_handle_ptr()->get_comms().get_rank();
+    return partition_.vertex_partition_offsets[comm_p_rank + 1] -
+           partition_.vertex_partition_offsets[comm_p_rank];
+  }
+
+  size_t get_number_of_adj_matrix_partitions() { return adj_matrix_partition_offsets_.size(); }
+
+  // Better avoid direct invocation in application code.
+  // This is mainly for pattern accelerator implementation.
+  // This function may disappear in the future if we swtich to CSR + DCSR (or CSC + DCSC)
```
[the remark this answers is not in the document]
seunghwak: Thanks for finding this and done.

cpp/src/experimental/graph.cu
seunghwak: Thanks for finding this and done.

cpp/src/experimental/graph.cu
seunghwak: Thanks for finding this and done.

cpp/src/experimental/graph.cu
seunghwak: Thanks for finding this and done.

cpp/include/experimental/detail/graph_utils.cuh
```diff
@@ -0,0 +1,129 @@
...
+// compute the numbers of nonzeros in rows (of the graph adjacency matrix, if store_transposed =
+// false) or columns (of the graph adjacency matrix, if store_transposed = true)
+template <typename vertex_t, typename edge_t>
+rmm::device_uvector<edge_t> compute_major_degree(
+  raft::handle_t const &handle,
+  std::vector<edge_t const *> const &adj_matrix_partition_offsets,
+  partition_t<vertex_t> const &partition)
+{
+  auto &comm_p_row     = handle.get_subcomm(comm_p_row_key);
+  auto comm_p_row_rank = comm_p_row.get_rank();
```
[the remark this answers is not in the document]
seunghwak: Done.

cpp/include/experimental/detail/graph_utils.cuh
```diff
@@ -0,0 +1,129 @@
...
+// false) or columns (of the graph adjacency matrix, if store_transposed = true)
+template <typename vertex_t, typename edge_t>
+rmm::device_uvector<edge_t> compute_major_degree(
+  raft::handle_t const &handle,
+  std::vector<rmm::device_uvector<edge_t>> const &adj_matrix_partition_offsets,
+  partition_t<vertex_t> const &partition)
+{
+  // we can avoid creating this temporary with "if constexpr" supported from C++17
+  std::vector<edge_t const *> tmp_offsets(adj_matrix_partition_offsets.size(), nullptr);
+  for (size_t i = 0; i < adj_matrix_partition_offsets.size(); ++i) {
```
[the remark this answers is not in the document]
seunghwak: Done.

cpp/src/experimental/graph.cu
seunghwak: This runtime check is very cheap (I mean runtime check of the flag, not the actual checks inside the if statement); so I think using the template policy here can be an unnecessary complication.

if (do_expensive_check) is executed only few times per graph object creation; this will add just some nanoseconds to the execution time. Graph object creation can take up to some seconds for large graphs. Some nanoseconds will be just noise.

aschaffer: This isn't really about performance (although, calling these "expensive" suggest it could be). 

It's more about clean code. Unless the conditions under which such checks can change at runtime (in which case template policies cannot help...much) then using the policies would make the code cleaner. By "cleaner" I mean more functional-like and less if-then-else-y. Branching code makes it harder to read and maintain. 

Arguably, if different / more conditions pop-up in the future, it's easier for the maintainer of the code (likely other than the original creator) to just define corresponding policy functors and instantiate the class with them.

Which is easier to read / maintain, below:
```
//(1):
//
void foo(void* p, size_t sz) {
  if( sz == 32 )
    f1(*(static_cast<int32_t*>(p)));
  else if( sz == 64 )
    f2(*(static_cast<int64_t*>(p)));
  else
    f3(*(static_cast<long long*>(p)));
}
```
Or:

```
//(2):
//
template<typename T>
struct callable_t {
   void operator()(T t) {
     f3(t);
   }
};
template<> struct callable_t<int32_t> {
   void operator()(int32_t t) {
       f1(t);
    }
};
template<> struct callable_t<int64_t> {
   void operator()(int64_t t) {
       f2(t);
    }
};

template<typename T>
void foo(void* p)
{
   callable_t<T> f{};
   f(*(static_cast<T*>(p)));
}
```
I would argue (2) is easier to read and maintain (provided `T` can be dispatched at compile-time). What if you have to branch in (1) on `sz == 32` on the signed vs. unsigned case? You'd need to add branching logic inside the first branch, etc. While, with (2) you just supply a `uint32_t` specialization.

That being said, I leave this decision to you. If you feel strongly about the initial implementation then feel free to resolve this conversation.

ChuckHastings: I think @seunghwak is trying to provide a runtime decision about whether to do these tests, not a compile time decision.  Looking at what's in the "expensive" tests, it's things like validating that the graph is, in fact, symmetric.  This (especially in MNMG mode) would be a very expensive test.  If the user wishes to assert/assume that the graph is in fact symmetric (they created it), they would want the option of being able to skip that validation test.

cpp/src/experimental/graph.cu
seunghwak: OK, now partition_t is a class and hypergraph_partitioned & vertex_partition_offsets are private.

cpp/src/experimental/graph.cu
seunghwak: Done.

cpp/src/experimental/graph.cu
seunghwak: Thanks but it seems like this does not apply to this case.

cpp/src/experimental/graph.cu
seunghwak: Thanks but it seems like this does not apply to this case.

cpp/src/experimental/graph.cu
seunghwak: replaced the for loop with std::any_of.

cpp/src/experimental/graph.cu
seunghwak: Chuck's comment is pretty much to the point. Even for python users, this should be false by default, but if users need to debug, turning this expensive check flag on is the first thing they can try; this decision happens in runtime.

cpp/src/experimental/graph_view.cu
seunghwak: adj_matrix_partition_offsets is std::vector; so size() is cheap. std::vector has three variables: pointer to the memory, size & capacity. So size() is just a getter. size() can be expensive for std::list kinda data structures but it's not a case here.

I tried to refactor this to two separate std::any_of calls and one transform_reduce call, but the problem is that I need to execute
```
raft::update_host(&number_of_local_edges,
                        adj_matrix_partition_offsets[i] + (major_last - major_first),
                        1,
                        default_stream);
```
2x more. This adds performance overhead and repeating this code 2x is also not desirable.

I can use thrust::make_zip_iterator (standard C++ does not support zip_iterator and I need to use boost instead, and I want to avoid this) and call thrust function with thrust::host execution policy, but I sort of want to avoid this; I guess it is common in RAPIDS to use std::algorithm for accessing host data and thrust::algorithm for accessing device data, and I did not want to go against this pattern. Let me know what you think.

seunghwak:
@aschaffer @ChuckHastings I either implemented the requested changes or replied to the request, let me know if you have any additional requests or comments.

cpp/tests/CMakeLists.txt
```diff
@@ -268,6 +268,15 @@ if (BUILD_MPI)
     ConfigureTest(NCCL_DEGREE_TEST "${NCCL_DEGREE_TEST_SRC}" "")
 endif(BUILD_MPI)
 
+###################################################################################################
```
afender, line 271: Do you mind deleting the dead code above in the `if (BUILD_MPI)` while editing this CMake file, please?

cpp/tests/experimental/graph_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+                        Tests_Graph,
+                        ::testing::Values(Graph_Usecase("test/datasets/karate.mtx", false),
+                                          Graph_Usecase("test/datasets/karate.mtx", true),
+                                          Graph_Usecase("test/datasets/web-Google.mtx", false),
+                                          Graph_Usecase("test/datasets/web-Google.mtx", true),
+                                          Graph_Usecase("test/datasets/ljournal-2008.mtx", false),
+                                          Graph_Usecase("test/datasets/ljournal-2008.mtx", true),
+                                          Graph_Usecase("test/datasets/webbase-1M.mtx", false),
+                                          Graph_Usecase("test/datasets/webbase-1M.mtx", true)));
+
```
afender, line 245: We also probably want to test :
- `multi_GPU = true/false` paths 
- `multi_GPU = True` but there's only one GPU (this is what will run on CI with the current setup)
This could be tested at the python level also

We also need to think about `multi_GPU = True` and Comm is set to use a subset of GPUs.

seunghwak: For true multi-GPU C++ tests, AFAIK, we don't have an infrastructure for this.

https://github.com/rapidsai/raft/pull/63 is recently merged to RAFT, and AFAIK, @rlratzel is working on C++ infrastructure for multi-GPU testing based on the above PR.

I can take a look for the multi-GPU with only one GPU case.

I don't think "We also need to think about multi_GPU = True and Comm is set to use a subset of GPUs." really matters. This code just works with the GPUs in the comm in the passed handle. No difference whether all the GPUs in the system is used or only a subset of the GPUs are used.

cpp/src/experimental/graph.cu
```diff
@@ -0,0 +1,552 @@
...
+    if (is_symmetric) {}
+    // FIXME: check for duplicate edges may better be implemented after deciding whether to sort
+    // neighbor list or not.
+    if (!is_multigraph) {}
+  }
+}
+
+// explicit instantiation
+
+template class graph_t<uint32_t, uint32_t, float, true, true>;
```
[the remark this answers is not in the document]
seunghwak: Related to 1, supporting integer needs more discussion. It may not matter much for SSSP but for PageRank, how should we handle division (a/b) if a is not an integer multiple of b? Should we really consider both integer & floating-point weights, or if users passes integer weights, should we just convert those to floating-point numbers?

2. Yeah... I guess we should use int32_t till those backward compatibility issues are resolved (IIRC, cuSparse is the main problem). In long run, better switch to uint32_t as this can cover nearly 2x more vertices.

cpp/src/experimental/graph_view.cu
```diff
@@ -0,0 +1,313 @@
...
+               std::vector<edge_t const*> const& adj_matrix_partition_offsets,
+               std::vector<vertex_t const*> const& adj_matrix_partition_indices,
+               std::vector<weight_t const*> const& adj_matrix_partition_weights,
+               std::vector<vertex_t> const& vertex_partition_segment_offsets,
+               partition_t<vertex_t> const& partition,
+               vertex_t number_of_vertices,
+               edge_t number_of_edges,
+               bool is_symmetric,
+               bool is_multigraph,
+               bool is_weighted,
```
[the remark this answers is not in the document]
seunghwak: Related to 1, just keeping the parameters used by the algorithms, but can definitely add more as needed.

Related to 2, currently, graph_properties_t is in the detail space and used just for graph internals, but I can make this part of the public API, if in the future, we expect passing 10+ graph properties through the public interface.

cpp/include/experimental/graph.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> view()
+  {
+    std::vector<edge_t const *> offsets(adj_matrix_partition_offsets_.size(), nullptr);
+    std::vector<vertex_t const *> indices(adj_matrix_partition_indices_.size(), nullptr);
+    std::vector<weight_t const *> weights(adj_matrix_partition_weights_.size(), nullptr);
+    for (size_t i = 0; i < offsets.size(); ++i) {
+      offsets[i] = adj_matrix_partition_offsets_[i].data();
+      indices[i] = adj_matrix_partition_indices_[i].data();
+      if (this->is_weighted()) { weights[i] = adj_matrix_partition_weights_[i].data(); }
```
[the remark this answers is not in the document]
seunghwak: If weights are passed but is_weighted is false, then, exception is raised (this is part of input argument check).

The question is whether we want to handle is_weighted same to other graph properties (e.g. is_symmetric, is_multigraph) or want to handle this differently and instead of keeping this value, we can compute this when asked (thankfully, checking weights is nullptr is cheap, but checking a graph has multi-edges or symmetric is not).

What do you think?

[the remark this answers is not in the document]
afender: I think the graph class responsibility is just to store what's given and infer if the graph is weighted or not. Later, algos may or may not try to use weights and exceptions should be triggered at that time.

cpp/src/experimental/graph.cu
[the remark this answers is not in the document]
afender: At first, if users pass integer weights we may just convert those to floating-point numbers. This does not really fly for SSSP and BC though as the largest `int` that can be stored properly in a `float` is about 16M. This not that big, especially when accumulating weights on a graph with billions of edges.

For 2), cusparse is one issue. More importantly, no cugraph structures and kernels have been updated or tested for unsigned.

cpp/src/experimental/graph_view.cu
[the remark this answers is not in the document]
afender: The graph C++ view will likely be called in pretty much each python binding and changing its signature as we add properties will trigger lots of edits. Having properties encapsulated may reduce maintenance in the long run. And, at run time, Python could just create it once and pass it as needed.

cpp/tests/CMakeLists.txt
afender: thanks!

cpp/tests/experimental/graph_test.cpp
afender: > I can take a look for the multi-GPU with only one GPU case.

CI will currently run in this context if you just set multi GPU to true

> This code just works with the GPUs in the comm in the passed handle. 

Agreed from a  theoretical perspective. In practice, we got surprises in the past. This is can be added later though.

cpp/src/experimental/graph.cu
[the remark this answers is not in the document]
seunghwak: What do you mean by "stored properly"?

"IEEE 754 32-bit base-2 floating-point variable has a maximum value of (2 − 2^−23) × 2^127 ≈ 3.4028235 × 10^38"
https://en.wikipedia.org/wiki/Single-precision_floating-point_format
3.4028235 × 10^38 is way larger than 2^32 - 1.

floating point numbers cannot accurately store integers (except for few values like 0 and 1) and the distance between two closest numbers increases as numbers get large, but I guess this does not matter unless a graph has a really large diameter.

cpp/tests/experimental/graph_test.cpp
seunghwak: Thought about adding MG tests, but I think this needs more discussion.

Are we going to have two separate gtests executable files or have one?

I assume SG version will be run as gtests/EXECUTABLE_NAME and MG version should be run as mpiexec -n #_gpus gtests/EXECUTABLE_NAME.

Mixing SG & MG tests in one executable may still work (but mpiexec -n #_gpus will run the same SG tests in every GPU) but can be wasteful. I think we need to decide our policy on running SG/MG C++ tests first before jumping to this (and are these C++ tests part of our CI/CD and benchmarking?)

cpp/include/experimental/graph.hpp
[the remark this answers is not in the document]
seunghwak: Done.

cpp/src/experimental/graph.cu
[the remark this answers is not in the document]
seunghwak: Addressed 2, supporting integer wights needs more discussion.

cpp/src/experimental/graph_view.cu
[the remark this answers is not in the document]
seunghwak: Done.

afender (approved):

BradReesWork:
rerun tests

Issue #1065, opened 2020-08-12
[FEA] sub-communicator initialization for 2D partitioning support


seunghwak:
**Is your feature request related to a problem? Please describe.**
In 2D partitioning, we align P workers (i.e. GPUs) to P_row * P_column and also partition the graph adjacency matrix in 2D.

Common communication patterns are collectives among the workers in the same row or column. To support this, we need to add sub-communicators for the workers in the same row and the same column.

https://github.com/rapidsai/raft/pull/18 and https://github.com/rapidsai/raft/pull/44 added sub-communicator support to RAFT comms.

cuGraph currently initializes only the global communicator. We need to initialize sub-communicaors as well.

https://github.com/rapidsai/raft/blob/f93dad05574b84d32ebbbd25681d2f9bcd7c0a14/cpp/include/raft/comms/comms.hpp#L95
```comm_split``` (similar to MPI_Comm_split https://www.mpich.org/static/docs/latest/www3/MPI_Comm_split.html) splits the global communicator to sub-communicators.

https://github.com/rapidsai/raft/blob/f93dad05574b84d32ebbbd25681d2f9bcd7c0a14/cpp/include/raft/handle.hpp#L163
https://github.com/rapidsai/raft/blob/f93dad05574b84d32ebbbd25681d2f9bcd7c0a14/cpp/include/raft/handle.hpp#L167
```set_subcomm``` and ```get_subcomm``` can be used add sub-communicators to the handle and retrieve the added sub-communicator when necessary.

- [x] Decide P_row & P_column for the given P. We may set P_row and P_column as close as possible to the square root of P by default with a user option to override the default value.
- [x] Call comm_split to create sub-communicators (two sub-communicators per worker, one for collectives among the workers in the same row, and the other for collectives among the workers in the same column)
- [x] Decide the string keys for the row sub-communicator and column sub-communicator.
- [x] Using the string keys, register the sub-communicators to the RAFT handle (call set_subcomm)

aschaffer:
> "Call comm_split to create sub-communicators (two sub-communicators per worker, one for collectives among the workers in the same row, and the other for collectives among the workers in the same column)"

2 sub-comms / worker means a total of: 2*P (= 2*P_row*P_col) sub-comms;

On the other hand,
> "Decide the string keys for the row sub-communicator and column sub-communicator."
Suggests one sub-comm per row and one per column; i.e., a total of (P_row + P_col) sub-comms.

Could we clarify what is intended here?

[...]

PR #1124, opened 2020-08-26
[REVIEW] Sub-communicator initialization for 2D partitioning support

aschaffer:
This tracks work on https://github.com/rapidsai/cugraph/issues/1065


cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,160 @@
...
+  void init_comms(void)
+  {
+    std::vector<pair_comms_t> empty_row(p_col_size_, std::make_pair(nullptr, nullptr));
+    comms_set_.assign(p_row_size_, empty_row);
+
+    raft::comms::comms_t const& communicator = handle_.get_comms();
+
+    for (size_type row = 0; row < p_row_size_; ++row)
+      for (size_type col = 0; col < p_col_size_; ++col) {
+        key_name_policy_t key{row, col};
```
seunghwak, line 117: I am still reading this PR, but it seems like there is some misunderstanding about sub-communicators for 2D partitioning. Each client should have only two sub-communicators (and two keys).

One for row-wise communication, and another for column-wise communication.

Say, P = 4 (GPU0, GPU1, GPU2, and GPU3), P_row = 2, and P_col = 2.

i.e.

GPU0 GPU1
GPU2 GPU3

For row-wise communication, there should be one group having GPU0 & GPU1 (say row-group0) and another group having GPU2 and GPU3 (say row-group1).

For column-wise communication, there should be one group having GPU0 & GPU2 (say col-group0) and another group having GPU 1 & GPU3 (say col-group1).

So each GPU0, needs to run comm_split only twice to get row-group0 and col-group0.
In case of GPU 3, need to run comm_split twice to get row-group1 and col-group1.

No need to loop over p_row_size * p_col_size.

seunghwak: See

https://github.com/rapidsai/cugraph/pull/1098/files#diff-75b979f478be47f71bd6933f474074cbR31
```
// FIXME: these should better be defined somewhere else.
std::string const comm_p_row_key = "comm_p_row";
std::string const comm_p_col_key = "comm_p_key";
```

and

https://github.com/rapidsai/cugraph/pull/1098/files#diff-51f3eab861b5285752c8047117a29f24R42
```
 auto &comm_p_row     = handle.get_subcomm(comm_p_row_key);
...
   comm_p_row.reduce(local_degrees.data(),
                      i == comm_p_col_rank ? degrees.data() : static_cast<edge_t *>(nullptr),
                      degrees.size(),
                      raft::comms::op_t::SUM,
                      comm_p_col_rank,
                      handle.get_stream());
```

For the ```handle.get_subcomm(comm_p_row_key);``` to work, ```set_subcomm``` should be called somewhere before this, and the names for the row-wise and column-wise communication keys should be defined (I temporarily defined the key names in graph_view.hpp but I think there will be a better place).

And this should be called when initializing our default_handle in python initialization. @Iroy30 may have a suggestion about the best place to call a function that initializes sub-communicators (two per GPU) and adds those to the default handle.

seunghwak: And row-group0 and row-group1 can use the same key name and col-group0 & col-group1 can use the same key. On GPU0, only row-group0 and col-group0 are visible, so no need to worry about row-group1.

seunghwak: A reading material:

https://mpitutorial.com/tutorials/introduction-to-groups-and-communicators

NCCL Comm split (and RAFT communicator split) is pretty similar to MPI Comm split.

aschaffer: Row/column key-naming aside, I think I've fixed the `subcomm` creation from a loop (one dedicated worker approach) to a row-and-column (hence two subcomms) per-worker approach.

As for the key-naming mechanism, is not clear to me how that should be fixed, yet. Taking this offline.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+struct key_naming_t {
+  key_naming_t(int row_indx,
+               int col_indx,
+               std::string const& col_suffix = std::string("_col"),
+               std::string const& row_suffix = std::string("_row"),
+               std::string const& prefix     = std::string("partition"))
+    : col_suffix_(col_suffix),
+      row_suffix_(row_suffix),
+      prefix_(prefix),
+      name_(prefix_ + "_" + to_string(row_indx) + "_" + to_string(col_indx))
```
seunghwak, line 108: I think this is overly complex.

If I am not mistaken, every GPU will separate names (keys) for row-wise communication & column-wise communication, and really no need for this.

We just need one key for row sub-communiator (something as simple as "comm_p_row") and one key for column sub-communicator.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+ protected:
+  virtual void partition2d(size_type p_size)
+  {
+    auto sqr = static_cast<size_type>(std::sqrt(p_size));
+
+    // find divisor of p_size
+    // nearest to sqr;
+    //
+    p_row_size_ = nearest_divisor(sqr, p_size);
+    p_col_size_ = p_size / p_row_size_;
```
seunghwak, line 71: @Iroy30 I remember you asked about finding p_row & p_col if only p is provided. Have you implemented this in python? Then, this code can be redundant.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+    assert(p_row_size_ > 1 && p_col_size_ > 1);
+  }
+
+  virtual void init_communicator(void)
+  {
+    // TODO: init's handle's communicator (singleton?)
+  }
+
+ private:
+  raft::handle_t& handle_;
```
seunghwak, line 82: Once handle_ is initialized, we can retrieve p_row_size_ from handle_.get_subcomms(row_key).get_size(); I think p_row_size_ & p_col_size_ below are redundant, and not sure we need this class.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+ public:
+  subcomm_factory_t(raft::handle_t& handle, size_type p_row_index, size_type p_col_index)
+    : handle_(handle), row_index_(p_row_index), col_index_(p_col_index)
+  {
+    init_row_col_comms();
+  }
+  virtual ~subcomm_factory_t(void) {}
+
+ protected:
+  virtual void init_row_col_comms(void)
```
seunghwak, line 143: I think what we need is just this function. Currently we're calling only set_comms() on handle. We need to call this function as well. Talk with @iroy30 about the best place to call this function.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+  {
+    name_policy_t key{row_index_, col_index_};
+    raft::comms::comms_t const& communicator = handle_.get_comms();
+
+    int const rank = communicator.get_rank();
+    int row_color  = rank / row_index_;
+    int col_color  = rank % row_index_;
+
+    auto row_comm = std::make_shared<raft::comms::comms_t>(
+      communicator.comm_split(row_color, static_cast<int>(key_2d_t::ROW)));
```
seunghwak, line 153: Yeah... I think you get confused between this key and the key to retrieve the sub-communicator.

The key here is to set the rank within a sub-communicator.

comm_split(rank / row_size, rank % row_size); works for row_comm and comm_split(rank % row_size, rank / row_size) works for col_comm.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+    handle_.set_subcomm(key.row_name(), row_comm);
```
seunghwak, line 154: We can use the same key here for every GPU; so we need only two keys: one for row_comm and one for col_comm.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+    auto col_comm = std::make_shared<raft::comms::comms_t>(
+      communicator.comm_split(col_color, static_cast<int>(key_2d_t::COL)));
+    handle_.set_subcomm(key.col_name(), col_comm);
```
seunghwak, line 158: We can use the same key here for every GPU; so we need only two keys: one for row_comm and one for col_comm.

cpp/include/partition_manager.hpp
aschaffer: In light of recent changes, we might _not_ need this class anymore. I kept it around because a previews comment said there are some constraints between the calls on communicators in a handle. 

As usually, I think such constraints / pre-conditions should be enforced (RAII-ed?) in code _somewhere_. Well, such a _somewhere_ can be this class.

cpp/include/partition_manager.hpp
aschaffer: Yes, this is the central piece of functionality. I will discuss this further with @Iroy30. Thanks.

cpp/include/partition_manager.hpp
aschaffer: Done.

cpp/include/partition_manager.hpp
aschaffer: Done.

cpp/include/partition_manager.hpp
aschaffer: Done.

cpp/include/partition_manager.hpp
aschaffer: Done.

cpp/include/partition_manager.hpp
aschaffer: @Iroy30, so where should this function called from? Thanks in advance.

cpp/include/partition_manager.hpp
aschaffer: @Iroy30, can you please comment on the above? I'd like to understand the action item for me, here. Thanks in advance.

Iroy30: https://github.com/rapidsai/cugraph/pull/1133/files#r488173099 I did implement it in python to be able to shuffle. If it needs to be in c++, I could call that func instead

cpp/include/partition_manager.hpp
Iroy30: @seunghwak When does this subcomm creation need to happen? I was thinking the cython wrapper right where set_comms is called , since it needs to be called by each process and we would just pass the handle and row_size to the subcomm_factory_t call their before initializing the 2D graph

cpp/include/partition_manager.hpp
seunghwak: I think it's better to compute this in the python layer as our current software architecture places more control at the python layer, and the C++ layer's main responsibility is in acceleration. This computation is cheap, so no need for acceleration.

cpp/include/partition_manager.hpp
seunghwak: @Iroy30 Yes, this needs to be called when we call set_comms. We set-up the global communicator first, then right after that, we set-up the sub-communicators. We put more control functions in the python layer, and if we call this function after setting up-the global communicator, I think most code outside this initi_row_col_comms function becomes irrelevant.

cpp/include/partition_manager.hpp
Iroy30: @aschaffer If this information is computed in python, how/where do you want me to pass it to you.

Iroy30: I believe we don't need the partition_manager_t class then and the subcomm_factory_t call could be called with python computed row_size in cython itself if my understanding correct

seunghwak: I think init_row_col_comms should take p_row_size and p_col_size as input parameters.

aschaffer: @Iroy30, yes, I agree that `partition_manager_t` may not be needed anymore. 

Its raison-d'etre was to fulfill the 1st requirement in https://github.com/rapidsai/cugraph/issues/1065: "Decide P_row & P_column for the given P. We may set P_row and P_column as close as possible to the square root of P by default with a user option to override the default value."

If that is covered in Python code, then I'll just remove this class and keep only ` subcomm_factory_t`. 

The other question that still needs an answer, then, is _where_ to call ` subcomm_factory_t{}` constructor in C++ code?

aschaffer: @seunghwak , why take _both_ `p_row_size` *and* `p_col_size`? Shouldn't `p_row_size` be enough? (as it is right now, except it's captured by the constructor). The `rank` info + the `p_row_size` info should be enough to allocate sub-comms, as it is right now. Correct?

seunghwak: @aschaffer I think you're right if p_col_size is not used (or can be computed from p_row_size & p) :-)

aschaffer: OK, removed `partition_manager_t` class, as discussed. Resolving all conversations related to that class.

cpp/include/partition_manager.hpp
aschaffer: Removed partition_manager_t class as requested. Resolving.

cpp/include/partition_manager.hpp
aschaffer: So, what is the action item for me, here? Where do I initialize the `subcomm_factory_t ` object?

If this is not yet clear (or the ideal location is "under construction") perhaps defer this to a follow-up PR and merge this as is?

Iroy30: Yes this just needs to be called in cython before calling the mnmg algorithms with 2D partitioning. The call should be included/updated in a separate mnmg pipeline PR. This PR can be merged as is.

Iroy30 (approved):

cpp/include/partition_manager.hpp
aschaffer: OK, resolving this. If this can be merged, can I get enough approvals to have it mergeable? (I'll handle conflicts shortly). Thanks in advance.

cpp/include/partition_manager.hpp
seunghwak: I think there was miscommunication,

No need for row_indx & col_indx to be added to name_.

We just need one name for row_communicator for the entire set of GPUs (e.g. comm_p_row) and one name for col_communicator for the entire set of GPUs.
```
  std::string col_name(void) const { return prefix_ + col_suffix_; 
  std::string row_name(void) const { return prefix_ + row_suffix_; }
```
should be sufficient.

seunghwak: and should we really allow users to change prefix_ and (row or col)_suffix_?

To retrieve the col_key, users need to do auto col_key = key_naming_t().get_col_name(). But if handle is initialized with non-default prefix_ or (row or col)_suffix_, this will not work.

aschaffer: > I think there was miscommunication,
> 
> No need for row_indx & col_indx to be added to name_.

There are two different constructors. One that takes `row_indx`, `col_indx_`; and one that does not. The semantics is that you don't _have to_ supply `row_indx`, `col_indx_`, but you can if you want (if for some reason in the future, that turns out to be useful; maybe for debug, I don't know). 
But, okay, I'll remove the constructor taking `row_indx`, `col_indx_`.

aschaffer: > and should we really allow users to change prefix_ and (row or col)_suffix_?
> 
> To retrieve the col_key, users need to do auto col_key = key_naming_t().get_col_name()

I assume you'd prefer here for prefix_ and (row or col)_suffix_ to be constants and the constructor to be a default one.

aschaffer: Done.

cpp/include/partition_manager.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+  std::string col_name(void) const { return name_ + col_suffix_; }
+
+  std::string row_name(void) const { return name_ + row_suffix_; }
+
+ private:
+  std::string const row_suffix_;
+  std::string const col_suffix_;
+  std::string const prefix_;
+  std::string name_;
```
seunghwak, line 60: It seems like now name_ & prefix_ are redundant. Besides this, this looks good to me.

seunghwak (approved):

cpp/include/partition_manager.hpp
aschaffer: Fixed.

PR #1131, opened 2020-09-01
[REVIEW] Fix style checker

raydouglass:
Do not merge


GPUtester:

ci/checks/style.sh
```diff
@@ -11,6 +11,7 @@
 # captured for returning a final status code. This allows all style check to
 # take place to provide a more comprehensive list of style violations.
 set -o pipefail
+set +e
```
rlratzel, line 14: If this has to be here then I'd add a comment since it looks redundant otherwise, and I'm afraid someone might remove it.

PR #1132, opened 2020-09-03
[REVIEW] upgrade thrust to latest commit

rongou:
This is to make it consistent with cuDF: https://github.com/rapidsai/cudf/blob/branch-0.16/thirdparty/CMakeLists.txt


GPUtester:

GPUtester:

afender (approved):
Thanks @rongou 

@ChuckHastings and @rlratzel : notice that this PR is switching to a commit hash instead of the version number for thrust and cub.

ChuckHastings:
> Thanks @rongou
> 
> @ChuckHastings and @rlratzel : notice that this PR is switching to a commit hash instead of the version number for thrust and cub.

We already did this (commit hash) with raft.  It provides much finer control (although it does require more regular maintenance... a fair trade IMO)

ChuckHastings (approved):

BradReesWork:
okay to test

BradReesWork:
rerun tests

rongou:
As @harrism suggested, now using the CUB submodule inside Thrust, instead of fetching CUB separately. Please take another look.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ## Improvements
 - PR 1081 MNMG Renumbering - sort partitions by degree
 - PR 1115 Replace deprecated rmm::mr::get_default_resource with rmm::mr::get_current_device_resource
+- PR 1132 Upgrade CUB/Thrust to laest commits
```
ChuckHastings, line 8: Didn't notice last time... do you want to fix the typo here 'laest' should be 'latest'

ChuckHastings (approved):
This looks fine also.

CHANGELOG.md
rongou: Fixed.

harrism:
> @ChuckHastings and @rlratzel : notice that this PR is switching to a commit hash instead of the version number for thrust and cub.

FWIW I think we should aim to switch back to a release tag ASAP and re-add GIT_SHALLOW so we don't fetch Thrust and CUB's entire git history.

rongou:
The latest CMake doc actually recommends commit hash over branch or tag (https://cmake.org/cmake/help/latest/module/ExternalProject.html):
>In general, however, specifying a commit hash should be preferred for a number of reasons:
> * If the local clone already has the commit corresponding to the hash, no git fetch needs to be performed to check for changes each time CMake is re-run. This can result in a significant speed up if many external projects are being used.
> * Using a specific git hash ensures that the main project’s own history is fully traceable to a specific point in the external project’s evolution. If a branch or tag name is used instead, then checking out a specific commit of the main project doesn’t necessarily pin the whole build to a specific point in the life of the external project. The lack of such deterministic behavior makes the main project lose traceability and repeatability.

For thrust, the difference between full clone (46MB) and shallow clone (37MB) is not too bad.

rongou:
The CI doesn't seem to be running. Does it need to be kicked?

raydouglass:
rerun tests

dillon-cullinan:
rerun tests

raydouglass:
ok to test

rongou:
rerun tests

BradReesWork:
okay to test

BradReesWork:
rerun tests

PR #1133, opened 2020-09-03
[REVIEW] python 2D shuffling

Iroy30:


GPUtester:

python/cugraph/structure/shuffle.py
```diff
@@ -0,0 +1,46 @@
...
+    from dask.distributed import default_client
+    client = default_client()
+    return len(client.scheduler_info()['workers'])
+
+
+def get_2D_div(ngpus):
+    pcols = int(math.sqrt(ngpus))
+    while ngpus % pcols != 0:
+        pcols = pcols - 1
+    return int(ngpus/pcols), pcols
```
Iroy30, line 15: Python code to find prows and pcols

python/cugraph/structure/shuffle.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _set_partitions_pre(df, num_verts, prows, pcols):
+    rows_per_div = math.ceil(num_verts/prows)
+    cols_per_div = math.ceil(num_verts/pcols)
+    partitions = df['src'].floordiv(rows_per_div) * pcols + df['dst'].floordiv(cols_per_div)
+    return partitions
+
+
+def shuffle(dg):
```
ChuckHastings, line 25: Suggest adding prows and pcols as parameters to the shuffle function, defaulting to None.

One of the easiest tuning things we want experienced users to be able to do is to configure prows and pcols, and it seems like this would be an easy way to do it.

If prows and pcols are specified then we should use them, otherwise we should compute them.  If a user specifies one then the other should be computable (ngpus / 'whichever one is specified').  If user specifies neither then we can compute as the current code does.

python/cugraph/structure/shuffle.py
```diff
@@ -0,0 +1,46 @@
...
+    ddf = dg.edgelist.edgelist_df
+    ngpus = get_n_workers()
+    prows, pcols = get_2D_div(ngpus)
```
ChuckHastings, line 28: Should probably add a check that prows * pcols = ngpus.  The implementation of get_2D_div guarantees that, but if the user specifies a value to tune things that isn't guaranteed.

python/cugraph/structure/shuffle.py
```diff
@@ -0,0 +1,46 @@
...
+
+def get_n_workers():
+    from dask.distributed import default_client
+    client = default_client()
+    return len(client.scheduler_info()['workers'])
+
+
+def get_2D_div(ngpus):
+    pcols = int(math.sqrt(ngpus))
+    while ngpus % pcols != 0:
```
ChuckHastings, line 13: Too lazy today to sit down and do math, but we should be able to directly compute this without a while loop.

ChuckHastings (changes requested):
We don't have MG CI resources, but could we have a unit test that developers could execute locally if they have > 1 GPU?

aschaffer (approved):

python/cugraph/structure/shuffle.py
```diff
@@ -0,0 +1,46 @@
...
+        pcols = pcols - 1
+    return int(ngpus/pcols), pcols
+
+
+def _set_partitions_pre(df, num_verts, prows, pcols):
+    rows_per_div = math.ceil(num_verts/prows)
+    cols_per_div = math.ceil(num_verts/pcols)
+    partitions = df['src'].floordiv(rows_per_div) * pcols + df['dst'].floordiv(cols_per_div)
```
seunghwak, line 21: Can we do this? If I am not mistaken, we apply a hash function to edge source & destination vertices to determine the partition each edge belongs to. So, each partition will have roughly same number of rows or columns but they will not be same (and can't be  determined in this way). Am I missing something?

afender:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

python/cugraph/structure/shuffle.py
afender, line 25: A few documentation lines would be a good addition

afender (approved):

python/cugraph/structure/shuffle.py
Iroy30: added

python/cugraph/structure/shuffle.py
Iroy30: updated

python/cugraph/structure/shuffle.py
Iroy30: added

Iroy30: added

PR #1135, opened 2020-09-03
[REVIEW] SG Updates to Louvain et. al.

### LINKED ISSUE #1092: [ENH] Restructure clustering software
The Louvain, Leiden and ECG code all share components.  The current configuration of the C++ code is awkward.  We should restructure the software to move common pieces into separate files so that the files aren't calling things defined in other algorithms.

This could also improve some of the templating awkwardness.

ChuckHastings:
Updates to SG version of Louvain and the related algorithms Leiden and ECG.  These changes are being made to simplify the MNMG implementation path but actually also improve the current SG implementation.

1) Moved Louvain implementation into a class to clean up a bunch of stuff and simplify function signatures
2) Added a renumber optimization when shrinking the graph
3) Modified Louvain to return a pair instead of passing in pointers to int and float/double for returning computed values
4) Added passing the handle so the function signature is ready for MNMG

Closes #1092


GPUtester:

ChuckHastings:
rerun tests

cpp/src/community/leiden.cu
```diff
@@ -0,0 +1,58 @@
...
+                                int max_level,
+                                weight_t resolution,
+                                cudaStream_t stream)
+{
+  CUGRAPH_EXPECTS(graph.edge_data != nullptr, "API error, leiden expects a weighted graph");
+  CUGRAPH_EXPECTS(leiden_parts != nullptr, "API error, leiden_parts is null");
+
+  Leiden<GraphCSRView<vertex_t, edge_t, weight_t>> runner(handle, graph, stream);
+
+  return runner.compute(leiden_parts, max_level, resolution);
```
aschaffer, line 35: `Leiden` class should perhaps be a functor, with overloaded `operator()(leiden_parts, max_level, resolution)`. Rather than calling `runner.compute(leiden_parts, max_level, resolution)` the call would just be `runner(leiden_parts, max_level, resolution)`.

Passing functors is a natural expectation for STL/Thrust algorithms.

cpp/src/community/leiden.cuh
```diff
@@ -0,0 +1,163 @@
...
+                     next_cluster_v.end(),
+                     this->cluster_v_.begin());
+      }
+    }
+
+    this->timer_stop(this->stream_);
+    return cur_Q;
+  }
+
+  std::pair<int, weight_t> compute(vertex_t *d_cluster_vec, int max_level, weight_t resolution)
```
aschaffer, line 100: Per above comment, please consider overloading `operator()(...)`, hence making this class a functor.

cpp/src/community/leiden.cuh
```diff
@@ -0,0 +1,163 @@
...
+    //
+    GraphCSRView<vertex_t, edge_t, weight_t> current_graph(this->offsets_v_.data().get(),
+                                                           this->indices_v_.data().get(),
+                                                           this->weights_v_.data().get(),
+                                                           this->number_of_vertices_,
+                                                           this->number_of_edges_);
+
+    current_graph.get_source_indices(this->src_indices_v_.data().get());
+
+    while (num_level < max_level) {
```
aschaffer, line 133: Please consider using a `for_each()` or `transform()` (or, `transform_if()`, `std::any_of()`, `std::all_of()`, `std::none_of()`) , due to the possibility of early break) on the sequence `{0,...,max_level-1}` rather than a while loop. 

This would lead to: 

1. better reading and maintenance;
2. flexibility on experimenting or using different execution policies;
3. better preparation for the future: C++17 is around the corner and STL algorithms can be (CPU) parallelized simply by passing different execution policies (`seq`, `par`, `par_unseq`);

cpp/src/community/louvain.cu
```diff
@@ -0,0 +1,57 @@
...
+                                 GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+                                 vertex_t *louvain_parts,
+                                 int max_level,
+                                 weight_t resolution,
+                                 cudaStream_t stream)
+{
+  CUGRAPH_EXPECTS(graph.edge_data != nullptr, "API error, louvain expects a weighted graph");
+  CUGRAPH_EXPECTS(louvain_parts != nullptr, "API error, louvain_parts is null");
+
+  Louvain<GraphCSRView<vertex_t, edge_t, weight_t>> runner(handle, graph, stream);
```
aschaffer, line 33: See comment on using a functor.

cpp/src/community/louvain.cuh
```diff
@@ -0,0 +1,639 @@
...
+                                                           (d_deg[community] * d_deg[community]) /
+                                                           (total_edge_weight * total_edge_weight));
+      },
+      weight_t{0.0},
+      thrust::plus<weight_t>());
+
+    return Q;
+  }
+
+  virtual std::pair<int, weight_t> compute(vertex_t *d_cluster_vec,
```
aschaffer, line 121: See comment on using a functor (hence, overloading `operator()(...)`)

cpp/src/community/louvain.cuh
```diff
@@ -0,0 +1,639 @@
...
+    //
+    GraphCSRView<vertex_t, edge_t, weight_t> current_graph(offsets_v_.data().get(),
+                                                           indices_v_.data().get(),
+                                                           weights_v_.data().get(),
+                                                           number_of_vertices_,
+                                                           number_of_edges_);
+
+    current_graph.get_source_indices(src_indices_v_.data().get());
+
+    while (num_level < max_level) {
```
aschaffer, line 151: See comment on using a `for_each()`, `transform()`, (or, `transform_if()`, `std::any_of()`, `std::all_of()`, `std::none_of()`) , due to the possibility of early break)` on the sequence `{0,...,max_level-1}` rather than a while loop.

cpp/src/community/louvain.cuh
```diff
@@ -0,0 +1,639 @@
...
+    weight_t new_Q = modularity(total_edge_weight, resolution, graph, cluster_v_.data().get());
+
+    weight_t cur_Q = new_Q - 1;
+
+    // To avoid the potential of having two vertices swap clusters
+    // we will only allow vertices to move up (true) or down (false)
+    // during each iteration of the loop
+    bool up_down = true;
+
+    while (new_Q > (cur_Q + 0.0001)) {
```
aschaffer, line 253: Please consider using a `for_each()` or `transform()` on the sequence `{0,...,max_level-1}` rather than a while loop.

cpp/src/community/louvain.cuh
```diff
@@ -0,0 +1,639 @@
...
+    //
+    thrust::for_each(
+      rmm::exec_policy(stream_)->on(stream_),
+      thrust::make_counting_iterator<edge_t>(0),
+      thrust::make_counting_iterator<edge_t>(graph.number_of_vertices),
+      [d_offsets, d_indices, d_weights, d_vertex_weights, d_cluster_weights] __device__(
+        vertex_t src) {
+        weight_t sum{0.0};
+
+        for (edge_t i = d_offsets[src]; i < d_offsets[src + 1]; ++i) { sum += d_weights[i]; }
```
aschaffer, line 215: Please consider using `thrust::reduce(thrust::seq,...)` with `__host__ __device__ make_counting_iterator()` instead of a for loop (and still avoid dynamic parallelism).

aschaffer (changes requested):

cpp/src/community/leiden.cuh
ChuckHastings: I'm not sure how I would restructure this loop efficiently in that manner, given the early break.

This loop must be executed sequentially.  The shrink_graph function (the last step in the loop) creates a new graph for the next iteration.

cpp/src/community/louvain.cuh
ChuckHastings: This loop runs an indeterminate number of times (nothing like a max_level).

cpp/src/community/leiden.cu
ChuckHastings: Done in next commit

cpp/src/community/leiden.cuh
ChuckHastings: Done in next commit

cpp/src/community/louvain.cu
ChuckHastings: Done in next commit

cpp/src/community/louvain.cuh
ChuckHastings: Done in next commit

cpp/src/community/louvain.cuh
ChuckHastings: Done in next commit

cpp/src/community/leiden.cuh
aschaffer: You could use `std::find_if` (or `thrust::find_if()` with `counting_iterator` on the `__host__`). `foo()` and `bar()` are equivalent in the example below:

```
template<typename fun_t,
         typename cond_t>
void foo(size_t max_iter, fun_t callable, cond_t condition)
{
    size_t num{0};
    while( num < max_iter) {
        callable();

        if( condition(num) ) break;
        ++num;
    }
    std::cout << "reached..." << num << '\n';
}

template<typename fun_t,
         typename cond_t>
void bar(size_t max_iter, fun_t callable, cond_t condition)
{
    std::vector<size_t> counter(max_iter);
    std::iota(std::begin(counter), std::end(counter), 0l);

    auto pos = std::find_if(std::begin(counter), std::end(counter),
                            [callable, condition](auto num) {
                                callable();
                                return condition(num);
                            });
    size_t num = std::distance(counter.begin(), pos);
    std::cout << "reached..." << num << '\n';
}
```
However, I admit this might be artificial and less readable. Your call here.

cpp/src/community/louvain.cuh
aschaffer: Please ignore then.

cpp/src/community/leiden.cuh
ChuckHastings: OK.  I see now.

I don't see a clarity advantage in the `std::find_if` approach for this loop (and the others structured like this in Louvain).  I could see if we might eventually want to use one of the non-sequential execution policies.

I did endeavor to make the entire contents of these while loops fit on a single screen in my editor, which I have found helps in keeping this type of construct clear.

aschaffer: No worries. Feel free to resolve.

BradReesWork (approved):

afender (approved):

cpp/include/algorithms.hpp
```diff
@@ -662,24 +664,26 @@ void louvain(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
...
+ *                                     2) modularity of the returned clustering
  */
 template <typename vertex_t, typename edge_t, typename weight_t>
-void leiden(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
-            weight_t &final_modularity,
-            int &num_level,
-            vertex_t *leiden_parts,
-            int max_iter        = 100,
-            weight_t resolution = weight_t{1});
+std::pair<int, weight_t> leiden(raft::handle_t const &handle,
```
seunghwak, line 682: I think we'd better use size_t for number of levels and maximum iterations (this will have negligible impact in memory footprint and performance while more future ready).

std::pair<size_t, weight_t> leiden(... , size_t max_iter = 100, ...);

cpp/src/community/leiden.cu
```diff
@@ -0,0 +1,58 @@
...
+                                GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+                                vertex_t *leiden_parts,
+                                int max_level,
+                                weight_t resolution)
+{
+  cudaStream_t stream{0};
+
+  return detail::leiden(handle, graph, leiden_parts, max_level, resolution, stream);
+}
+
```
seunghwak, line 51: add // Explicit template instantiations.

to be consistent with the rest of cuGraph

cpp/src/community/louvain.cu
```diff
@@ -0,0 +1,57 @@
...
+}  // namespace detail
+
+template <typename vertex_t, typename edge_t, typename weight_t>
+std::pair<int, weight_t> louvain(raft::handle_t const &handle,
+                                 GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+                                 vertex_t *louvain_parts,
+                                 int max_level,
+                                 weight_t resolution)
+{
+  cudaStream_t stream{0};
```
seunghwak, line 47: Is this necessary?

cpp/include/algorithms.hpp
ChuckHastings: I could make that change.  These parameters represent the number of iterations of the outer loop of Louvain, each loop shrinks the graph by a significant amount (don't have a hard heuristic on that).  I suspect we'd need a graph with something on the order of 2^256 edges in order to exceed 2^32 Louvain levels... but that's intuition not based on any measurements.

seunghwak: I don't think this will ever lead to overflow here, either.

My preference on size_t over int is that 1) size_t tells more about the nature of the variable than int and 2) consistency (it may not lead to overflow here but can happen elsewhere).

I generally try to avoid using (unsigned) short, (unsigned) int, (unsigned) long unless this is used by the external library I am interfacing with (e.g. MPI uses int, so I use int for number of processes...).

But this is just my personal preference, and just a recommendation.

cpp/include/algorithms.hpp
```diff
@@ -622,9 +622,8 @@ void bfs(raft::handle_t const &handle,
  *                                   Supported value : int (signed, 32-bit)
  * @tparam weight_t                  Type of edge weights. Supported values : float or double.
  *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
  * @param[in]  graph                 input graph object (CSR)
- * @param[out] final_modularity      modularity of the returned clustering
- * @param[out] num_level             number of levels of the returned clustering
  * @param[out] clustering            Pointer to device array where the clustering should be stored
```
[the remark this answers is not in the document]
ChuckHastings: Agree the names should be consistent.  I'm slightly more inclined to change louvain_parts to clustering, but could be talked into either one (or something different).

Regarding the allocation question, this is analogous to the pagerank result.  The current paradigm we use in the interface is to allocate memory when we know the structure of the return a priori in the python/cython layer and pass down the pointer.  Currently we only allocate memory in the C++ and return it if the exact size/shape of the memory required is unknown until during the execution of the algorithm.

Are you suggesting that we change the overall paradigm?  We could certainly have all of the algorithms (Louvain, Pagerank, etc) allocate their result array in C++ and return it.

I moved the final_modularity and num_level (which are a single value of type weight_t and int, respectively) into the pair to return to eliminate the pass by reference/address discrepancy between Louvain and Leiden.  But if we think it's better for all of the algorithms to allocate their return device vectors and return them I can blaze the trail in Louvain.

cpp/include/algorithms.hpp
```diff
@@ -633,14 +632,17 @@ void bfs(raft::handle_t const &handle,
...
+ *
  */
 template <typename vertex_t, typename edge_t, typename weight_t>
-void louvain(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
-             weight_t *final_modularity,
-             int *num_level,
-             vertex_t *louvain_parts,
-             int max_iter        = 100,
-             weight_t resolution = weight_t{1});
+std::pair<int, weight_t> louvain(raft::handle_t const &handle,
```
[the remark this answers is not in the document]
ChuckHastings: (see comment below)

cpp/include/algorithms.hpp
```diff
@@ -662,24 +664,26 @@ void louvain(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
  *                                   Supported value : int (signed, 32-bit)
  * @tparam weight_t                  Type of edge weights. Supported values : float or double.
  *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
  * @param[in]  graph                 input graph object (CSR)
- * @param[out] final_modularity      modularity of the returned clustering
- * @param[out] num_level             number of levels of the returned clustering
  * @param[out] clustering            Pointer to device array where the clustering should be stored
```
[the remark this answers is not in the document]
ChuckHastings: (see comment above)

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
seunghwak: I'm fine with either clustering or louvain_parts but comments and the code should agree.

Yeah... I don't suggest changing the overall paradigm. Returning 2+ variables using std::pair (or std::tuple) while still returning some results through input pointers looks a bit unorthodox, but this may be the best at this point.

May discuss how to resolve this later.

cpp/src/community/louvain.cu
ChuckHastings: I will clean this up.  The handle provides a stream, so this is unnecessary (as is the stream being passed in to the Louvain/Leiden constructor.

cpp/src/community/triangles_counting.cu
```diff
@@ -826,8 +826,8 @@ void TrianglesCount<IndexType>::count()
   else if (mean_deg < DEG_THR2)
     tcount_wrp();
   else {
-    const int shMinBlkXSM = 6;
-    if (size_t{m_shared_mem_per_block * 8 / shMinBlkXSM} < (size_t)m_mat.N)
+    const int shMinBlkXSM{6};
+    if (int64_t{m_shared_mem_per_block * 8} < int64_t{m_mat.N * shMinBlkXSM})
```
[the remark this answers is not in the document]
ChuckHastings: I was tired of the warning about losing precision on the conversion of a signed value to an unsigned value and this hack made it go away :-)

I agree this is less than ideal.  I can remove it from this PR and we can create a separate issue to resolve this issue.  I think largely the issue is that the triangle count implementation seems to be pretty careless about types.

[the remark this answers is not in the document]
ChuckHastings: Created #1140 

Reverting this change.  I'll ignore the warning for a while longer...

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/include/algorithms.hpp
ChuckHastings: Done

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Done

aschaffer (approved):

cpp/src/community/leiden.cu
```diff
@@ -0,0 +1,58 @@
...
+template <typename vertex_t, typename edge_t, typename weight_t>
+std::pair<int, weight_t> leiden(raft::handle_t const &handle,
+                                GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
+                                vertex_t *leiden_parts,
+                                int max_level,
+                                weight_t resolution,
+                                cudaStream_t stream)
+{
+  CUGRAPH_EXPECTS(graph.edge_data != nullptr, "API error, leiden expects a weighted graph");
+  CUGRAPH_EXPECTS(leiden_parts != nullptr, "API error, leiden_parts is null");
```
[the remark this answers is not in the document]
ChuckHastings: Fixed the ecg/louvain/leiden contribution to this issue

cpp/src/community/leiden.cu
ChuckHastings: Done

cpp/src/community/leiden.cuh
```diff
@@ -0,0 +1,163 @@
...
+      }
+    }
+
+    this->timer_stop(this->stream_);
+    return cur_Q;
+  }
+
+  std::pair<int, weight_t> operator()(vertex_t *d_cluster_vec, int max_level, weight_t resolution)
+  {
+    int num_level{0};
```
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/include/algorithms.hpp
```diff
@@ -692,21 +696,26 @@ void leiden(GraphCSRView<vertex_t, edge_t, weight_t> const &graph,
...
  * @param[in]  min_weight            The minimum weight parameter
  * @param[in]  ensemble_size         The ensemble size parameter
  * @param[out] ecg_parts             A device pointer to array where the partitioning should be
  * written
  */
-template <typename VT, typename ET, typename WT>
-void ecg(GraphCSRView<VT, ET, WT> const &graph_csr, WT min_weight, VT ensemble_size, VT *ecg_parts);
+template <typename vertex_t, typename edge_t, typename weight_t>
+void ecg(raft::handle_t const &handle,
+         GraphCSRView<vertex_t, edge_t, weight_t> const &graph_csr,
```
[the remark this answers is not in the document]
ChuckHastings: Renamed to graph (for consistency).

If we can agree on a naming convention we can update these as we go forward.

seunghwak (approved):
Thanks!!!

python/cugraph/community/leiden_wrapper.pyx
```diff
@@ -69,25 +72,23 @@ def leiden(input_graph, max_iter, resolution):
                                                   <float*>c_weights, num_verts, num_edges)
 
         graph_float.get_vertex_identifiers(<int*>c_identifier)
-        c_leiden(graph_float,
-                  final_modularity_float,
-                  num_level,
-                  <int*> c_partition,
-                  <int> max_iter,
-                  <float> resolution)
+        num_level, final_modularity_float = c_leiden(handle_ptr.get()[0],
```
rlratzel, line 75: Nice, I didn't realize the pair could be unpacked like this. I was defining a cdef pair and accessing with `.first`, etc.  I'll need to change my WIP PR for consistency.

rlratzel (approved):
LGTM

BradReesWork:
rerun tests

Issue #1052, opened 2020-08-10
[ENH] MNMG Louvain


BradReesWork:
distributed Louvain that scales beyond 2 billion vertices

PR #1139, opened 2020-09-10
[REVIEW] MNMG Louvain Python updates, Cython cleanup

rlratzel:
This adds a Python test and modules under `cugraph/dask/community` to support the MNMG Louvain C++ updates.
These changes are needed not only for exposing MG Louvain to users, but also to facilitate C++ testing.

This is still a WIP - it builds, but that's about it for now.
At a minimum, still need to:
- [ ] Improve test
- [x] Figure out Dask calls needed and update `dask/commuity/louvain.py` accordingly
- [ ] Call new 2D graph class from `louvain_wrapper.pyx` (https://github.com/rapidsai/cugraph/pull/1098)

Issue: #1052 

This also includes some cleanup: renamed the `graph_new*` files to (hopefully) better reflect their contents. The new name is `graph_primtypes*` but this can easily be changed if there's a better name that should be used.


GPUtester:

afender (commented):
Can you elaborate on the motivation for renaming `graph` into `graph_primtypes` and why the latter is a better name?
Are we reserving  `cugraph.structure.graph` for something else?

rlratzel:
> Can you elaborate on the motivation for renaming `graph` into `graph_primtypes` and why the latter is a better name?
> Are we reserving `cugraph.structure.graph` for something else?

I initially wanted to use "graph", but I had some issues with the graph module that cython generated colliding with the current graph.py, but I admittedly didn't push too hard to get them merged after the import errors I initially saw after that attempt.

I just took a stab at a simple rename of graph_new at that point, as well as removing the redundant graph.pxd file.  "primtypes" seemed like a decent fit based on the contents being the lower-level more primitive basic types used for graph algos.

I'm open to a better name, or even trying harder to combine the contents of graph.py with what the former graph_new cython modules were defining.  I can even back this cleanup out if we want to revisit this later.

afender:
Typically we add the `_wrapper` suffix for `.pyx` files to avoid conflicts with `.py` files. Would it help here? I noticed there was a pyx file that didn't have the suffix. 
Ideally, we should keep the most concise name to be exposed and add prefixes/suffixes to those that don't get exposed.

rlratzel:
> Typically we add the `_wrapper` suffix for `.pyx` files to avoid conflicts with `.py` files. Would it help here? I noticed there was a pyx file that didn't have the suffix.

I tried using that pattern as well, but I had an issue combining the code in `graph_new.pyx` and `graph_new_wrapper.pyx` into `graph_wrapper.pyx` - cython crashed, and it seemed it didn't like having a `cdef` function that was used by one of the `def` functions in the same file (my theory). I didn't investigate much and I didn't spend time refactoring it either, and instead just decided to rename for clarity and remove the redundant `graph.pxd` file. Combining the `def`s and `cdef`s this way in a single file seems like it should be legal, but the only way I could get it to build was to keep them separate (graph_new was doing this as well). So we still have 3 files, but they have hopefully better names now: `graph_primtypes.pxd`, `graph_primtypes.pyx` (cdefs), and `graph_primtypes_wrapper.pxd` (defs).

I'd like to solve this better, but I'm afraid of taking too much time away from the MG work.  I'm fine backing this out if it's too distracting now, or if we want to do it differently later.

rlratzel:
rerun tests

ChuckHastings (approved):
I think this is a good start.  It would allow work to proceed in parallel while we explore things.

It would make sense for some of the things that are custom to Louvain in this PR to actually be more general purpose.  Hopefully as we get more of the cython pieces in place it will become obvious which things should migrate to more generic functions.

cpp/include/utilities/cython.hpp
```diff
@@ -0,0 +1,70 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <graph.hpp>
+// #include <experimental/graph_view.hpp>
```
aschaffer, line 19: Consider removing dead code. If it's "experimental" code (useful for debugging, etc.) please consider guarding it between an `#ifdef - #endif` macro.

cpp/include/utilities/cython.hpp
```diff
@@ -0,0 +1,70 @@
...
+// FIXME: use std::variant instead of a union if possible
+// FIXME: add both CSRView and graph_type_t objects for easier testing during
+//        the transition
+union graphUnion {
+  graphUnion() {}
+  GraphCSRView<int, int, float> GraphCSRViewFloat;
+  GraphCSRView<int, int, double> GraphCSRViewDouble;
+};
+
+enum weightTypeEnum { floatType = 0, doubleType = 1 };
```
aschaffer, line 34: Please consider using scoped enumeration 
```
enum class weightTypeEnum : int { floatType = 0, doubleType = 1};
```
More readable, provides better name hiding (scoping), and stronger typing.

cpp/src/utilities/cython.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+  } else {
+    graph_container.graph.GraphCSRViewDouble = GraphCSRView<int, int, double>(
+      offsets, indices, reinterpret_cast<double*>(weights), num_vertices, num_edges);
+    graph_container.graph.GraphCSRViewDouble.set_local_data(
+      local_vertices, local_edges, local_offsets);
+    graph_container.graph.GraphCSRViewDouble.set_handle(const_cast<raft::handle_t*>(&handle));
+  }
+
+  return std::move(graph_container);
```
aschaffer, line 59: Please return-by-value,
```
return graph_container;
```
 rather than convert to r-value reference. This cast (`std::move()`) prevents the compiler to perform the very powerful return-by-value optimization (RVO), which allows elimination of copying (_copy elision_).

RVO, or copy elision allows the compiler to allocate (under a few typical circumstance, like the one here; i.e., local variable returned without any conversion) the local returned variable directly into the caller's variable that gets assigned to the result of calling the function.

aschaffer (changes requested):

cpp/include/utilities/cython.hpp
rlratzel: done in latest commit

cpp/include/utilities/cython.hpp
rlratzel: done in latest commit

cpp/src/utilities/cython.cpp
rlratzel: done in latest commit

aschaffer (approved):

python/cugraph/dask/community/louvain.py
```diff
@@ -0,0 +1,91 @@
...
+    >>> parts, modularity_score = dcg.louvain(dg)
+    """
+    # FIXME: finish docstring: describe parameters, etc.
+
+    # FIXME: import here to prevent circular import: cugraph->louvain
+    # wrapper->cugraph/structure->cugraph/dask->dask/louvain->cugraph/structure
+    # from cugraph.structure.graph import Graph
+
+    # FIXME: dask methods to populate graphs from edgelists are only present on
+    # DiGraph classes. Disable the Graph check for now and assume inputs are
```
afender, line 59: Why are we exposing symmetric DiGraph and throw errors for `Graph` or actual `DiGraph`? It is a bit awkward from a UX perspective since cuGraph has the `Graph` class for undirected graphs.

afender (approved):
Good progress. There's still work to do on UX, testing, and documentation

BradReesWork:
rerun tests

rlratzel:
rerun tests

python/cugraph/tests/dask/test_mg_louvain.py
```diff
@@ -0,0 +1,91 @@
...
+def daskGraphFromDataset(request, client_connection):
+    """
+    Returns a new dask dataframe created from the dataset file param.
+    """
+    # Since parameterized fixtures do not assign param names to param values,
+    # manually call the helper to do so.
+    setFixtureParamNames(request, ["dataset"])
+    dataset = request.param
+
+    chunksize = dcg.get_chunksize(dataset)
```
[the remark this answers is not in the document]
BradReesWork: > That won't fly for the targeted MG scale but ok for small smoke tests.

CI is not for scale testing.  So not hitting the targeted MG scale in this test is fine

BradReesWork:
rerun tests

cpp/include/utilities/cython.hpp
```diff
@@ -15,53 +15,113 @@
...
-//        the transition
-union graphUnion {
-  graphUnion() {}
-  GraphCSRView<int, int, float> GraphCSRViewFloat;
-  GraphCSRView<int, int, double> GraphCSRViewDouble;
-};
+enum class numberTypeEnum : int { intType, floatType, doubleType };
 
-enum weightTypeEnum { floatType = 0, doubleType = 1 };
+// FIXME: The GraphCSRView* types are not in use! Those are left in place in
```
ChuckHastings, line 27: Could add this in a later PR if desired.

Do we want to add GraphCSCViewFloat, GraphCSCViewDouble, GraphCOOViewFloat and GraphCOOViewDouble for completeness?  These are legacy classes, but it would allow all of the cython code to be modified to the new technique for graph allocation regardless of whether it hs been converted to the new graph objects or not.

Perhaps we defer this until we decide it would be useful.

This *might* simplify the later migration of all algorithms to MNMG.

cpp/include/utilities/cython.hpp
```diff
@@ -15,53 +15,113 @@
...
-                                 int weightType,
-                                 int num_vertices,
-                                 int num_edges,
-                                 int* local_vertices,
-                                 int* local_edges,
-                                 int* local_offsets,
-                                 bool transposed,
-                                 bool multi_gpu);
+// FIXME: Should local_* values be void* as well?
+void create_graph_t(graph_container_t& graph_container,
```
ChuckHastings, line 105: Add documentation for this function.

Would be useful to understand what local_vertices, local_edges and local_offsets mean while looking at the header.

cpp/include/utilities/cython.hpp
```diff
@@ -15,53 +15,113 @@
...
+    experimental::graph_view_t<int, int, double, false, true>* graph_view_t_double_mg_ptr;
+    experimental::graph_view_t<int, int, float, true, false>* graph_view_t_float_transposed_ptr;
+    experimental::graph_view_t<int, int, double, true, false>* graph_view_t_double_transposed_ptr;
+    experimental::graph_view_t<int, int, float, true, true>* graph_view_t_float_mg_transposed_ptr;
+    experimental::graph_view_t<int, int, double, true, true>* graph_view_t_double_mg_transposed_ptr;
+  };
+
+  inline graph_container_t() : graph_ptr_union{nullptr}, graph_ptr_type{graphTypeEnum::null} {}
+
+  inline ~graph_container_t()
```
ChuckHastings, line 67: The destructor deletes the pointer within the union.  A reasonable thing to do.

Probably need to decide the desired life cycle behavior of this class.  If the intention is to create the graph container at the cython level and pass it by reference (probably const reference since I don't think you would want a caller to change anything) then we should delete the copy constructor, assignment operators, etc so that we won't have pointer issues.

cpp/include/utilities/cython.hpp
```diff
@@ -15,53 +15,113 @@
...
+                    int* local_edges,
+                    int* local_offsets,
+                    bool transposed,
+                    bool multi_gpu);
 
 // Wrapper for calling Louvain using a graph container
 template <typename weight_t>
 weight_t call_louvain(raft::handle_t const& handle,
-                      graph_container_t graph_container,
+                      graph_container_t& graph_container,
```
ChuckHastings, line 124: Should be a const reference passed to `call_louvain`.

cpp/include/utilities/cython.hpp
```diff
@@ -15,53 +15,113 @@
...
                       int* parts,
```
ChuckHastings, line 125: This needs to match vertex_t in the graph.  Do you want to make this a void * and cast it inside the function?

python/cugraph/tests/dask/test_mg_louvain.py
```diff
@@ -0,0 +1,91 @@
...
+    ddf = dask_cudf.read_csv(dataset, chunksize=chunksize,
+                             delimiter=' ',
+                             names=['src', 'dst', 'value'],
+                             dtype=['int32', 'int32', 'float32'])
+
+    dg = cugraph.DiGraph()
```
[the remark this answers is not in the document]
rlratzel: Agreed. I'll probably have to defer this to the next PR though.

python/cugraph/dask/community/louvain.py
```diff
@@ -0,0 +1,91 @@
...
+                                handle,
+                                max_level,
+                                resolution)
+
+
+def louvain(input_graph, max_iter=100, resolution=1.0, load_balance=True):
+    """
+    Compute the modularity optimizing partition of the input graph using the
+    Louvain method on multiple GPUs
+
```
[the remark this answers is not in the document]
rlratzel: Agreed, that was an oversight on my part.  I'll try to do that in the next commit to address Chuck's feedback, or may defer to the next PR.

python/cugraph/dask/community/louvain.py
rlratzel: Agreed, there's still some tech debt there that I may have to address in the next PR.

cpp/include/utilities/cython.hpp
rlratzel: sounds good, I'll start this and see how involved it becomes. If it's holding up the PR too much I may defer it to the next PR.

BradReesWork (approved):

cpp/include/utilities/cython.hpp
rlratzel: I added these and a FIXME to state how we might use them in the function when we need them.

cpp/include/utilities/cython.hpp
rlratzel: Agreed. I added a copy ctor and assignment operator and deleted them.  I also added a comment in the `graph_container_t` struct describing the typical usage/lifespan of the struct, but moving forward if we need to use this differently we can support and document that as well.

cpp/include/utilities/cython.hpp
rlratzel: done

cpp/include/utilities/cython.hpp
rlratzel: done

cpp/include/utilities/cython.hpp
rlratzel: done

PR #1148, opened 2020-09-17
[REVIEW] Replace deprecated RMM get_default_resource

harrism:
`get_default_resource()` was deprecated in RMM 0.15 and will be removed in rapidsai/rmm#560. This PR replaces calls to this function with `get_current_device_resource()`.


cpp/include/algorithms.hpp
```diff
@@ -423,7 +423,7 @@ template <typename VT, typename ET, typename WT>
 std::unique_ptr<GraphCOO<VT, ET, WT>> k_truss_subgraph(
   GraphCOOView<VT, ET, WT> const &graph,
   int k,
-  rmm::mr::device_memory_resource *mr = rmm::mr::get_default_resource());
+  rmm::mr::device_memory_resource *mr = rmm::mr::get_current_device_resource());
```
seunghwak, line 426: This is very weird as these are fixed in https://github.com/rapidsai/cugraph/pull/1115 and the current code is actually identical to the modified code (https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/include/algorithms.hpp#L426), but github is listing this as an update.

seunghwak: Can you merge with the latest cuGraph 0.16 main branch? I think most (possibly all) of these updates will no longer be relevant.

aschaffer: Yeah, I agree w/ @seunghwak: I don't see any in `cugraph` call to ` get_default_resource()` in any `*.*h*` or `*.c*`. I verified that on a branch that was merged last night.

aschaffer: So, it looks like the only dependency on `get_default_resource()` is via an old version of `raft` (GIT_TAG           099e2b874b05555a78bed1666fa2d22f784e56a7):
```
/home/aschaffer/Development/Projects/Graph/forked/fea_ext_mnmg_py_support/cugraph/cpp/bld_p_11_0/raft/src/raft/cpp/include/raft/mr/device/allocator.hpp: In member function ‘virtual void* raft::mr::device::default_allocator::allocate(std::size_t, cudaStream_t)’:
/home/aschaffer/Development/Projects/Graph/forked/fea_ext_mnmg_py_support/cugraph/cpp/bld_p_11_0/raft/src/raft/cpp/include/raft/mr/device/allocator.hpp:40:26: warning: ‘rmm::mr::device_memory_resource* rmm::mr::get_default_resource()’ is deprecated [-Wdeprecated-declarations]
     void* ptr = rmm::mr::get_default_resource()->allocate(n, stream);
```
The latest `raft` doesn't seem to have any dependency on `get_default_resource()`, anymore, either (a `find-grep` found nothing), so perhaps this is just a matter of upgrading the `raft` dependency to latest?

seunghwak: Yeah, I made a fix to RAFT in https://github.com/rapidsai/raft/pull/56 as well. We just need to update RAFT git tag in https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/CMakeLists.txt#L287

The latest RAFT tag is 516106e3b515b25c863776fcc51fb12df6c0a186

harrism: LOL. I'm sorry team. I obviously had an out of date branch-0.16 when I started. My mistake. I will close this PR. Thanks!

PR #1152, opened 2020-09-18
[REVIEW] Graph container cleanup, added arg for instantiating legacy types and switch statements to factory function

rlratzel:
Refactored the graph container to use `unique_ptr`s instead of raw pointers after conversation with @aschaffer , added additional legacy graph types to the factory function which will be needed for MG Pagerank.

### DIFF
--- a/cpp/src/utilities/cython.cpp
+++ b/cpp/src/utilities/cython.cpp
@@ -56,80 +56,86 @@ void populate_graph_container(graph_container_t& graph_container,
   if (weightType == numberTypeEnum::floatType) {
     switch (legacyType) {
       case legacyGraphTypeEnum::CSR: {
-        auto g = new GraphCSRView<int, int, float>(reinterpret_cast<int*>(offsets),
-                                                   reinterpret_cast<int*>(indices),
-                                                   reinterpret_cast<float*>(weights),
-                                                   num_vertices,
-                                                   num_edges);
         graph_container.graph_ptr_union.GraphCSRViewFloatPtr =
-          std::unique_ptr<GraphCSRView<int, int, float>>(g);
+          std::make_unique<GraphCSRView<int, int, float>>(reinterpret_cast<int*>(offsets),
+                                                          reinterpret_cast<int*>(indices),
+                                                          reinterpret_cast<float*>(weights),
+                                                          num_vertices,
+                                                          num_edges);
         graph_container.graph_ptr_type = graphTypeEnum::GraphCSRViewFloat;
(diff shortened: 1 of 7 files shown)


GPUtester:

cpp/src/utilities/cython.cpp
```diff
@@ -124,7 +179,8 @@ void populate_graph_container(graph_container_t& graph_container,
         graph_props,
         sorted_by_global_degree_within_vertex_partition,
         do_expensive_check);
-      graph_container.graph_ptr_union.graph_view_t_double_mg_ptr = g;
+      graph_container.graph_ptr_union.graph_view_t_double_mg_ptr =
+        std::unique_ptr<experimental::graph_view_t<int, int, double, false, true>>(g);
```
ChuckHastings, line 183: Have you considered using std::make_unique?  This would clean u p this block of code a little bit (don't need the temporary variable g).

cpp/src/utilities/cython.cpp
```diff
@@ -185,15 +243,15 @@ weight_t call_louvain(raft::handle_t const& handle,
...
+              *(graph_container.graph_ptr_union.GraphCSRViewFloatPtr.get()),
               reinterpret_cast<int*>(parts),
               max_level,
               static_cast<float>(resolution));
     final_modularity = results.second;
   } else {
     std::pair<size_t, double> results =
       louvain(handle,
-              *(graph_container.graph_ptr_union.GraphCSRViewDoublePtr),
+              *(graph_container.graph_ptr_union.GraphCSRViewDoublePtr.get()),
```
ChuckHastings, line 254: Do you need the .get() here?  Doesn't the * operator do the right thing?

ChuckHastings (commented):
Couple of questions...

cpp/src/utilities/cython.cpp
rlratzel: ah, thanks, I don't need `.get()`.

aschaffer (approved):

cpp/src/utilities/cython.cpp
rlratzel: done

ChuckHastings (approved):

afender (approved):

rlratzel:
rerun tests

BradReesWork:
rerun tests

rlratzel:
rerun tests

rlratzel:
rerun tests

PR #1157, opened 2020-09-23
[REVIEW] Louvain API update to use graph_container_t

ChuckHastings:
Update the Louvain API to support calling with either old or new graphs.

### DIFF
--- a/cpp/src/community/louvain.cu
+++ b/cpp/src/community/louvain.cu
@@ -39,11 +39,10 @@ std::pair<size_t, weight_t> louvain(raft::handle_t const &handle,
 template <typename vertex_t,
           typename edge_t,
           typename weight_t,
-          bool store_transposed,
           bool multi_gpu>
 std::pair<size_t, weight_t> louvain(
   raft::handle_t const &handle,
-  experimental::graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> const
+  experimental::graph_view_t<vertex_t, edge_t, weight_t, false, multi_gpu> const
     &graph_view,
   vertex_t *clustering,
   size_t max_level,
@@ -52,7 +51,7 @@ std::pair<size_t, weight_t> louvain(
   CUGRAPH_EXPECTS(clustering != nullptr, "Invalid input argument: clustering is null");
 
   experimental::Louvain<
-    experimental::graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>>
+    experimental::graph_view_t<vertex_t, edge_t, weight_t, false, multi_gpu>>
     runner(handle, graph_view);
   return runner(clustering, max_level, resolution);
 }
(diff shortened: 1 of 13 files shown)


GPUtester:

ChuckHastings:
rerun tests

cpp/src/community/louvain.cu
```diff
@@ -15,25 +15,62 @@
...
+  Louvain<GraphCSRView<vertex_t, edge_t, weight_t>> runner(handle, graph_view);
+  return runner(clustering, max_level, resolution);
+}
 
+template <typename vertex_t,
+          typename edge_t,
+          typename weight_t,
+          bool store_transposed,
+          bool multi_gpu>
+std::pair<size_t, weight_t> louvain(
```
seunghwak, line 44: Just a question, and why do we have both this one and the old one (https://github.com/rapidsai/cugraph/pull/1157/files#diff-d8514b5ca06cd2748169ce04e7ad85a5R25). Do you plan to eventually delete the old one?

seunghwak: And if louvain works with only transposed or un-transposed matrix (I guess this expects CSR as the old function takes GraphCSRView), you can skip store_transposed from the template parameter list, and just provide the boolean value in graph_view_t.
e.g.
```
experimental::graph_view_t<vertex_t, edge_t, weight_t, false, multi_gpu> const &graph_view,
```

cpp/src/community/louvain.cu
```diff
@@ -15,25 +15,62 @@
...
+
+  experimental::Louvain<
+    experimental::graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>>
+    runner(handle, graph_view);
   return runner(clustering, max_level, resolution);
 }
 
+}  // namespace detail
+
+template <typename graph_t>
```
seunghwak, line 62: I guess this graph_t can be either old GraphCSRView or new graph_view_t? Is this a temporary solution to easily switch between the old one and the new one, or something that will stay? Or is this a mechanism to use the old one for SG and the new one for MG? (till we switch to the new class for SG as well)

cpp/src/community/louvain.cu
```diff
@@ -42,5 +79,17 @@ template std::pair<size_t, double> louvain(raft::handle_t const &,
                                            int32_t *,
                                            size_t,
                                            double);
+template std::pair<size_t, float> louvain(
+  raft::handle_t const &,
+  experimental::graph_view_t<int32_t, int32_t, float, false, false> const &,
```
seunghwak, line 84: If instantiate for vertex_t,edge_t = int32_t, then we're still limited to work with fewer than 2B vertices/edges (and I found that I did this too in my PR838, so you may copied this :-(...), so I guess we need to fix this. For the new graph class, it's explicitly instantiated for (int32_t, init32_t), (int32_t, int64_t), and (int64_t, int64_t) combinations (https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/src/experimental/graph.cu#L510).

cpp/src/community/louvain.cu
ChuckHastings: The experimental version doesn't work yet (in this PR), this is an interim step to allow the old code and the new code to coexist.  Once the new version is complete and tested the old one will be deleted.

cpp/src/community/louvain.cu
ChuckHastings: I intend this as an interim step, although it might be a reasonable mechanism to address the templated variations of the new graph_view_t.  Hadn't really thought about that in too much depth, just needed a means for handling the old and new so far.

cpp/src/community/louvain.cu
ChuckHastings: This PR is intended to help bootstrap the python bindings integration work.  I agree that adding the int64_t variations will ultimately have to be done as well.  I'm sure I'll need it once I start testing MNMG Louvain on multiple GPUs, so my intention would be to add it to my implementation PR.

ChuckHastings: Upon further reflection (and since I had to make an update anyway) I have added the additional explicit template instantiations.

seunghwak (approved):

afender (approved):

BradReesWork:
rerun tests

PR #1162, opened 2020-09-25
[REVIEW] ENH updating Nx compatibility testing

BradReesWork:


GPUtester:

python/cugraph/tests/test_nx_convert.py
```diff
@@ -29,19 +29,66 @@
...
+        nx_df = nx_df.drop(columns=["weight"])
+    nx_df = nx_df.rename(columns={"source": "src", "target": "dst"})
+    nx_df = nx_df.astype('int32')
+    nx_df = nx_df.sort_values(by=["src", "dst"]).reset_index(drop=True)
+
+    assert cu_df.to_dict() == nx_df.to_dict()
+
+
+@pytest.mark.parametrize("graph_file", utils.DATASETS)
+def test_networkx_compatibility(graph_file):
```
rlratzel, line 52: optional: maybe rename to "test_networkx_graph_compatibility" since just "compatibility" is pretty broad.

rlratzel (approved):
LGTM. I just had one minor suggestion which is optional.

python/cugraph/tests/test_nx_convert.py
rlratzel: On second though, the test is scoped to the "nx_convert" test file, so "convert" isn't as broad of a term in that sense. Your call.

Iroy30 (approved):

PR #1163, opened 2020-09-28
[REVIEW] ENH Integrate 2D shuffling and latest Louvain API

### LINKED ISSUE #1050: [ENH] Dask to 2D partitioin pipeline
Validate and add missing components to go from DASK to 2D partitioned data

rlratzel:
This PR integrates the latest updates to both 2D shuffling and the Louvain C++ work to allow for a complete e2e call from the user API into the Louvain MNMG C++.

closes #1050


GPUtester:

cpp/src/utilities/cython.cpp
```diff
@@ -140,155 +265,51 @@ void populate_graph_container(graph_container_t& graph_container,
...
   // edges will be available, and when that happens, additional castings will be
   // needed for the 'parts' arg in particular. For now, it is hardcoded to int.
-  if (graph_container.graph_ptr_type == graphTypeEnum::GraphCSRViewFloat) {
+  if (graph_container.graph_ptr_type == graphTypeEnum::graph_t_float_mg) {
+    results = louvain(handle,
+                      graph_container.graph_ptr_union.graph_t_float_mg_ptr->view(),
+                      reinterpret_cast<int*>(parts),
+                      max_level,
+                      static_cast<float>(resolution));
+
```
ChuckHastings, line 294: We'll need some kind of graph_t or graph_view_t analog to get_vertex_identifiers...

cpp/include/utilities/cython.hpp
```diff
@@ -66,22 +66,22 @@ struct graph_container_t {
...
+    std::unique_ptr<experimental::graph_t<int, int, double, false, true>>
+      graph_t_double_mg_ptr;
+    std::unique_ptr<experimental::graph_t<int, int, float, true, false>>
+      graph_t_float_transposed_ptr;
+    std::unique_ptr<experimental::graph_t<int, int, double, true, false>>
+      graph_t_double_transposed_ptr;
+    std::unique_ptr<experimental::graph_t<int, int, float, true, true>>
+      graph_t_float_mg_transposed_ptr;
+    std::unique_ptr<experimental::graph_t<int, int, double, true, true>>
+      graph_t_double_mg_transposed_ptr;
```
seunghwak, line 84: So, aren't we adding 64 bit integer support in this release? With only int, we still can't support graphs with more than 2B vertices.

And can't we use int32_t instead of int here? I am not sure about the restrictions but I found this (https://github.com/cython/cython/blob/master/Cython/Includes/libc/stdint.pxd) in googling so I guess it might be possible?

We may need to support (int32_t, int32_t), (int32_t, int64_t), and (int64_t, int64_t) for (vertex_t, edge_t) pairs; and eventually replace int##_t to uint##_t once we retire the code that relies on vertex_t & edge_t being signed integers.

cpp/include/utilities/cython.hpp
```diff
@@ -162,26 +151,40 @@ struct graph_container_t {
...
+                              void* dst_vertices,
                               void* weights,
-                              numberTypeEnum offsetType,
-                              numberTypeEnum indexType,
+                              void* vertex_partition_offsets,
+                              numberTypeEnum vertexType,
+                              numberTypeEnum edgeType,
                               numberTypeEnum weightType,
                               int num_vertices,
                               int num_edges,
```
seunghwak, line 163: This code will break with more than 2B (2^31-1) vertices and/or edges.

cpp/include/utilities/cython.hpp
```diff
@@ -162,26 +151,40 @@ struct graph_container_t {
...
+                                     void* indices,
+                                     void* weights,
+                                     numberTypeEnum offsetType,
+                                     numberTypeEnum indexType,
+                                     numberTypeEnum weightType,
+                                     int num_vertices,
+                                     int num_edges,
+                                     int* local_vertices,
+                                     int* local_edges,
+                                     int* local_offsets);
```
seunghwak, line 181: This code will break with more than 2B (2^31-1) vertices and/or edges.

cpp/src/utilities/cython.cpp
seunghwak: You can retrieve vertex ranges with get_local_vertex_first() & get_local_vertex_last().

Then, can't you just use thrust::make_counting_iterator()? This will save both memory & computing than explicitly creating an array and filling the array with thrust::sequence.

cpp/src/utilities/cython.cpp
```diff
@@ -140,155 +259,51 @@ void populate_graph_container(graph_container_t& graph_container,
...
+  } else if (graph_container.graph_ptr_type == graphTypeEnum::graph_t_double_mg) {
+    results = louvain(handle,
+                      graph_container.graph_ptr_union.graph_t_double_mg_ptr->view(),
+                      reinterpret_cast<int*>(parts),
+                      max_level,
+                      static_cast<double>(resolution));
+  } else if (graph_container.graph_ptr_type == graphTypeEnum::GraphCSRViewFloat) {
+     //     if (graph_container.graph_ptr_type == graphTypeEnum::GraphCSRViewFloat) {
+    graph_container.graph_ptr_union.GraphCSCViewFloatPtr->get_vertex_identifiers(
+      static_cast<int32_t*>(identifiers));
```
seunghwak, line 298: What happens if vertex_t is int64_t? Is this path irrelevant or it's not supported?

cpp/src/utilities/cython.cpp
```diff
@@ -140,155 +259,51 @@ void populate_graph_container(graph_container_t& graph_container,
...
     results = louvain(handle,
                       *(graph_container.graph_ptr_union.GraphCSRViewFloatPtr),
                       reinterpret_cast<int*>(parts),
                       max_level,
                       static_cast<float>(resolution));
-  } else {
+  } else if (graph_container.graph_ptr_type == graphTypeEnum::GraphCSRViewDouble) {
+    graph_container.graph_ptr_union.GraphCSCViewDoublePtr->get_vertex_identifiers(
+      static_cast<int32_t*>(identifiers));
```
seunghwak, line 306: What happens if vertex_t is int64_t? Is this path irrelevant or it's not supported?

python/cugraph/community/louvain_wrapper.pyx
```diff
@@ -79,28 +79,27 @@ def louvain(input_graph, max_level, resolution):
...
-    graph_container.get_vertex_identifiers(<void*>c_identifier)
+    populate_graph_container_legacy(graph_container,
+                                    <legacyGraphTypeEnum>(<int>(legacyGraphTypeEnum.CSR)),
+                                    handle_[0],
+                                    <void*>c_offsets, <void*>c_indices, <void*>c_weights,
+                                    <numberTypeEnum>(<int>(numberTypeEnum.intType)),
+                                    <numberTypeEnum>(<int>(numberTypeEnum.intType)),
+                                    <numberTypeEnum>(<int>(weightTypeMap[weights.dtype])),
+                                    num_verts, num_edges,
+                                    <int*>c_local_verts, <int*>c_local_edges, <int*>c_local_offsets)
```
seunghwak, line 90: What happens if vertex_t is int64_t? Is this path irrelevant or it's not supported?

cpp/src/utilities/cython.cpp
ChuckHastings: The python code is constructing a data frame with a column called 'vertex' that contains these identifiers that is returned to the caller.

I think something like:

`thrust::copy(..., thrust::make_counting_iterator(get_local_vertex_first()), thrust::make_counting_iterator(get_local_vertex_last(), identifiers)`

is probably the right implementation.

cpp/include/utilities/cython.hpp
rlratzel: yes, I still need to do a pass and address all the 32/64 int sizes issues you bring up once I'm successfully calling Louvain with a `graph_view_t`.

cpp/include/experimental/graph_view.hpp
```diff
@@ -140,27 +140,29 @@ class partition_t {
     auto major_first =
       hypergraph_partitioned_
         ? vertex_partition_offsets_[row_comm_size_ * partition_idx + row_comm_rank_]
-        : vertex_partition_offsets_[row_comm_rank_ * col_comm_size_];
+        : vertex_partition_offsets_[col_comm_rank_ * row_comm_size_];
     auto major_last =
       hypergraph_partitioned_
         ? vertex_partition_offsets_[row_comm_size_ * partition_idx + row_comm_rank_ + 1]
-        : vertex_partition_offsets_[(row_comm_rank_ + 1) * col_comm_size_];
+        : vertex_partition_offsets_[(col_comm_rank_ * row_comm_size_)+1];
```
seunghwak, line 147: This should be (col_comm_rank_ + 1) * row_comm_size

cpp/include/experimental/graph_view.hpp
```diff
@@ -140,27 +140,29 @@ class partition_t {
...
-             : vertex_partition_offsets_[row_comm_rank_ * col_comm_size_];
+             : vertex_partition_offsets_[col_comm_rank_ * row_comm_size_];
   }
 
   vertex_t get_matrix_partition_major_last(size_t partition_idx) const
   {
     return hypergraph_partitioned_
              ? vertex_partition_offsets_[row_comm_size_ * partition_idx + row_comm_rank_ + 1]
-             : vertex_partition_offsets_[(row_comm_rank_ + 1) * col_comm_size_];
+             : vertex_partition_offsets_[(col_comm_rank_ * row_comm_size_)+1];
```
seunghwak, line 165: This should be (col_comm_rank_ + 1) * row_comm_size

seunghwak: and for the minor_range, minor_first, minor_last

```
  std::tuple<vertex_t, vertex_t> get_matrix_partition_minor_range() const
  {
    auto minor_first = get_matrix_partition_minor_first();
    auto minor_last  = get_matrix_partition_minor_last();

    return std::make_tuple(minor_first, minor_last);
  }

  vertex_t get_matrix_partition_minor_first() const
  {
    return hypergraph_partitioned_ ? vertex_partition_offsets_[col_comm_rank_ * row_comm_size_]
                                   : vertex_partition_offsets_[row_comm_rank_ * col_comm_size_];
  }

  vertex_t get_matrix_partition_minor_last() const
  {
    return hypergraph_partitioned_
             ? vertex_partition_offsets_[(col_comm_rank_ + 1) * row_comm_size_]
             : vertex_partition_offsets_[(row_comm_rank_ + 1) * col_comm_size_];
  }
```

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,159 @@ namespace cython {
...
   CUGRAPH_EXPECTS(graph_container.graph_ptr_type == graphTypeEnum::null,
                   "populate_graph_container() can only be called on an empty container.");
 
+  bool do_expensive_check{false};
+  bool hypergraph_partitioned{false};
+
+  raft::comms::comms_t const& communicator = handle.get_comms();
+  int const rank = communicator.get_rank();
+  int partition_row_rank = rank / partition_row_size;
+  int partition_col_rank  = rank % partition_row_size;
```
seunghwak, line 57:
```
      auto& row_comm = handle.get_subcomm(cugraph::partition_2d::key_naming_t().row_name());
      auto const row_comm_rank = row_comm.get_rank();
      auto const row_comm_size = row_comm.get_size();
      auto& col_comm = handle.get_subcomm(cugraph::partition_2d::key_naming_t().col_name());
      auto const col_comm_rank = col_comm.get_rank();
      auto const col_comm_size = col_comm.get_size();
```

seunghwak: and I'd recommend using row_comm_size instead of partition_row_size for consistency. Inconsistency in naming here can be very confusing and can be a source of many bugs.

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,159 @@ namespace cython {
...
+
+  // Copy the contents of the vertex_partition_offsets (host array) to a vector
+  // as needed by the partition_t ctor.
+  int* vertex_partition_offsets_array = reinterpret_cast<int*>(vertex_partition_offsets);
+  std::vector<int> vertex_partition_offsets_vect;  // vertex_t
+
+  for (int32_t i=0; i<((partition_row_size * partition_col_size)+1); ++i) {
+     vertex_partition_offsets_vect.push_back(vertex_partition_offsets_array[i]);
+     std::cout<<"VPO"<<i<<": "<<vertex_partition_offsets_vect[i]<<std::endl;
+  }
```
seunghwak, line 71:
```
std::vector<int> vertex_partition_offsets_vect(
  vertex_partition_offsets_array,
  vertex_partition_offsets_array + (col_comm_size * row_comm_size)+1);
```

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,159 @@ namespace cython {
...
+  bool hypergraph_partitioned{false};
+
+  raft::comms::comms_t const& communicator = handle.get_comms();
+  int const rank = communicator.get_rank();
+  int partition_row_rank = rank / partition_row_size;
+  int partition_col_rank  = rank % partition_row_size;
+
+  // Setup the subcommunicators needed for this partition on the handle
+  partition_2d::subcomm_factory_t<partition_2d::key_naming_t, int>
+    subcomm_factory(handle, partition_row_size);
```
seunghwak, line 61: And shouldn't we better initialize subcomms right after initializing comms? Why are we delaying subcommunicator initialization to this point?

cpp/include/utilities/cython.hpp
rlratzel: done, will be passing as `size_t` in latest commit.

cpp/include/utilities/cython.hpp
rlratzel: done, will be passing as `size_t` in latest commit.

cpp/include/experimental/graph_view.hpp
rlratzel: done

cpp/include/experimental/graph_view.hpp
rlratzel: done

cpp/src/utilities/cython.cpp
rlratzel: I've added a FIXME for this in the next commit.

python/cugraph/dask/community/louvain_wrapper.pyx
```diff
@@ -98,25 +85,38 @@ def louvain(input_df, local_data, rank, handle, max_level, resolution):
...
+    print("CALLING PGC")
     populate_graph_container(graph_container,
-                             <legacyGraphTypeEnum>(<int>(legacyGraphTypeEnum.CSR)),
                              handle_[0],
-                             <void*>c_offsets, <void*>c_indices, <void*>c_weights,
+                             <void*>c_src_vertices, <void*>c_dst_vertices, <void*>c_edge_weights,
+                             <void*>c_vertex_partition_offsets,
                              <numberTypeEnum>(<int>(numberTypeEnum.intType)),
                              <numberTypeEnum>(<int>(numberTypeEnum.intType)),
                              <numberTypeEnum>(<int>(weightTypeMap[weights.dtype])),
```
Iroy30, line 95: Can weights be None?

cpp/src/utilities/cython.cpp
rlratzel: Using the handle to retrieve the ranks in the next commit, and changed var names to `*_comm_size` for consistency too.

cpp/src/utilities/cython.cpp
rlratzel: done in next commit. I thought there was a better way, thanks.

python/cugraph/dask/community/louvain_wrapper.pyx
rlratzel: No, that's a problem, but I fixed this in the next commit, thanks.

Iroy30 (approved):
lgtm

cpp/include/utilities/cython.hpp
rlratzel: @ChuckHastings has some refactorings to the cython factory C++ code which addresses the int64-based feedback, so I'll create another PR now to integrate that and defer these items until then.

cpp/src/utilities/cython.cpp
rlratzel: To be addressed in next PR (see comment above).

cpp/src/utilities/cython.cpp
rlratzel: To be addressed in next PR (see comment above).

python/cugraph/community/louvain_wrapper.pyx
rlratzel: To be addressed in next PR (see comment above).

cpp/src/utilities/cython.cpp
rlratzel: I'll defer this to the next PR for this code.

rlratzel:
rerun tests

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,161 @@ namespace cython {
...
+                              size_t num_global_edges,
+                              size_t row_comm_size,  // pcols
+                              size_t col_comm_size,  // prows
                               bool transposed,
                               bool multi_gpu)
 {
   CUGRAPH_EXPECTS(graph_container.graph_ptr_type == graphTypeEnum::null,
                   "populate_graph_container() can only be called on an empty container.");
 
+  bool do_expensive_check{true};
```
afender, line 51: Should this be `False` by default?

rlratzel: yes, thank you, that was left on for debugging. Fixed in latest commit.

ChuckHastings (approved):
Remaining things will be addressed in follow on PR.  This one is needed to be merged for easier integration with other work.

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,161 @@ namespace cython {
...
+                              void* vertex_partition_offsets,
+                              numberTypeEnum vertexType,
+                              numberTypeEnum edgeType,
                               numberTypeEnum weightType,
-                              int num_vertices,
-                              int num_edges,
-                              int* local_vertices,
-                              int* local_edges,
-                              int* local_offsets,
+                              int num_partition_edges,
```
seunghwak, line 40: I think this should be size_t as well.

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,161 @@ namespace cython {
...
+  // auto const row_comm_size = row_comm.get_size(); // pcols
+  auto& col_comm           = handle.get_subcomm(cugraph::partition_2d::key_naming_t().col_name());
+  auto const col_comm_rank = col_comm.get_rank();
+  // auto const col_comm_size = col_comm.get_size(); // prows
+
+  // Copy the contents of the vertex_partition_offsets (host array) to a vector
+  // as needed by the partition_t ctor.
+  int* vertex_partition_offsets_array = reinterpret_cast<int*>(vertex_partition_offsets);
+
+  // FIXME: this needs to be vertex_t, not int?
```
seunghwak, line 74: Yes, this needs to be vertex_t.

seunghwak (commented):
OK, I guess these are my last complaints :-)

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,161 @@ namespace cython {
...
+                                           row_comm_rank,
+                                           col_comm_rank);
+
+  experimental::graph_properties_t graph_props{.is_symmetric = false, .is_multigraph = false};
+
+  auto src_vertices_array = reinterpret_cast<int*>(src_vertices);
+  auto dst_vertices_array = reinterpret_cast<int*>(dst_vertices);
+
+  if (multi_gpu) {
+    bool sorted_by_global_degree_within_vertex_partition{false};
```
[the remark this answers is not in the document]
ChuckHastings: If renumbering was called then the data will be sorted within the vertex partition.

cpp/src/utilities/cython.cpp
```diff
@@ -27,28 +28,161 @@ namespace cython {
...
+        sorted_by_global_degree_within_vertex_partition,
+        do_expensive_check);
+
+      graph_container.graph_ptr_union.graph_t_double_mg_ptr =
+        std::unique_ptr<experimental::graph_t<int, int, double, false, true>>(g);
+      graph_container.graph_ptr_type = graphTypeEnum::graph_t_double_mg;
+    }
+
+  } else {
+    bool sorted_by_degree{false};
```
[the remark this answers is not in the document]
ChuckHastings: If renumbering was called then the data will be sorted by degree.

BradReesWork (approved):

cpp/src/utilities/cython.cpp
rlratzel: Created PR #1178

cpp/src/utilities/cython.cpp
rlratzel: The problem with using `size_t` here is that since `num_partition_edges` is passed to the `edgelist_t` ctor directly and is expecting it as an `edge_t`, it results in a "narrowing conversion" which doesn't compile, but I agree that `size_t` or even `edge_t` would be better.

I will address this in PR #1178

cpp/src/utilities/cython.cpp
rlratzel: To be addressed in PR #1178

cpp/src/utilities/cython.cpp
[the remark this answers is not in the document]
rlratzel: Thank you, I'll add logic to handle that in PR #1178 so this can get merged to unblock others.

cpp/src/utilities/cython.cpp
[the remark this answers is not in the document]
rlratzel: Thank you, I'll add logic to handle that in PR #1178 so this can get merged to unblock others.

seunghwak (approved):

cpp/src/utilities/cython.cpp
rlratzel: The `create_graph()` function in PR #1178 uses `vertex_t` for this.

cpp/include/utilities/cython.hpp
rlratzel: Confirmed support for int32 and int64 types is present, and I added a `FIXME` for updating to unsigned int types when vertex_t and edge_t code can support unsigned.

cpp/src/utilities/cython.cpp
rlratzel: @ChuckHastings is planning on addressing this in (I think) PR #1172

cpp/src/utilities/cython.cpp
rlratzel: @ChuckHastings is planning on addressing handling vertex identifiers in (I think) PR #1172

Looked closer at adding support for int64_t and talked with @ChuckHastings and realized we wouldn't support int64_t vertex types for legacy graph classes.

python/cugraph/community/louvain_wrapper.pyx
rlratzel: Looked closer at adding support for int64_t and talked with @ChuckHastings and realized we wouldn't support int64_t vertex types for legacy graph classes.

cpp/src/utilities/cython.cpp
rlratzel: Looked closer at adding support for int64_t and talked with @ChuckHastings and realized we wouldn't support int64_t vertex types for legacy graph classes which are only used for SG.

PR #1164, opened 2020-09-28
[REVIEW] MG symmetrize and conda env updates

### LINKED ISSUE #1159: [BUG] missing dependencies in conda recipe
**Describe the bug**
It appears that the dependencies on `dask-cuda` and `dask-cudf` are missing from the `cugraph` conda recipe: https://github.com/rapidsai/cugraph/blob/branch-0.16/conda/recipes/cugraph/meta.yaml#L31-L36

This doesn't work:
```
conda create -c rapidsai -c conda-forge -n cugraph_test cugraph
conda activate cugraph_test
python -c 'import cugraph'
```
Result:
```
  File "/home/sseibert/miniconda3/envs/cugraph_test/lib/python3.8/site-packages/cugraph/__init__.py", line 14, in <module>
    from cugraph.community import (
  File "/home/sseibert/miniconda3/envs/cugraph_test/lib/python3.8/site-packages/cugraph/community/__init__.py", line 14, in <module>
    from cugraph.community.louvain import louvain
  File "/home/sseibert/miniconda3/envs/cugraph_test/lib/python3.8/site-packages/cugraph/community/louvain.py", line 14, in <module>
    from cugraph.community import louvain_wrapper
  File "cugraph/community/louvain_wrapper.pyx", line 21, in init cugraph.community.louvain_wrapper
  File "/home/sseibert/miniconda3/envs/cugraph_test/lib/python3.8/site-packages/cugraph/structure/__init__.py", line 14, in <module>
    from cugraph.structure.graph import Graph, DiGraph
  File "/home/sseibert/miniconda3/envs/cugraph_test/lib/python3.8/site-packages/cugraph/structure/graph.py", line 14, in <module>
    from cugraph.structure import graph_new_wrapper
  File "cugraph/structure/graph_new_wrapper.pyx", line 28, in init cugraph.structure.graph_new_wrapper
ModuleNotFoundError: No module named 'dask_cudf'
```

Adding `dask-cudf` and `dask-cuda` explicitly in the `conda create` command fix the problem:
```
conda create -c rapidsai -c conda-forge -n cugraph_test cugraph dask-cudf dask-cuda
```

### LINKED ISSUE #1167: [ENH] MG symmetrize
https://github.com/rapidsai/cugraph/issues/310 for multi GPU. The goal is to accept a distributed undirected edge lists from the user perspective but pass a symmetric COO internally to the solver. From a UX perspective, this allows passing a `Graph` instead of a `DiGraph` containing edges in both directions. 

Implement at python level with dask cudf

afender:
closes #1159 
closes #1167

### DIFF
--- a/python/cugraph/structure/graph.py
+++ b/python/cugraph/structure/graph.py
@@ -347,6 +347,17 @@ class Graph:
         if self.edgelist is not None or self.adjlist is not None:
             raise Exception("Graph already has values")
 
+        s_col = source
+        d_col = destination
+        if not isinstance(s_col, list):
+            s_col = [s_col]
+        if not isinstance(d_col, list):
+            d_col = [d_col]
+        if not (set(s_col).issubset(set(input_df.columns)) and
+                set(d_col).issubset(set(input_df.columns))):
+            raise Exception('source column names and/or destination column \
+names not found in input. Recheck the source and destination parameters')
+
         # Consolidation
         if isinstance(input_df, cudf.DataFrame):
             if len(input_df[source]) > 2147483100:
@@ -445,6 +456,16 @@ class Graph:
         if type(self) is Graph:
             raise Exception('Undirected distributed graph not supported')
 
+        s_col = source
+        d_col = destination
+        if not isinstance(s_col, list):
+            s_col = [s_col]
+        if not isinstance(d_col, list):
+            d_col = [d_col]
+        if not (set(s_col).issubset(set(input_ddf.columns)) and
+                set(d_col).issubset(set(input_ddf.columns))):
+            raise Exception('source column names and/or destination column \
+names not found in input. Recheck the source and destination parameters')
         #
         # Keep all of the original parameters so we can lazily
         # evaluate this function
--- a/python/cugraph/structure/graph.py
+++ b/python/cugraph/structure/graph.py
@@ -133,8 +133,10 @@ class Graph:
         comms = Comms.get_comms()
 
         if client is None or comms is None:
-            msg = "MG Batch needs a Dask Client and the " \
+            msg = (
+                "MG Batch needs a Dask Client and the "
                 "Communicator needs to be initialized."
+            )
             raise Exception(msg)
 
         self.batch_enabled = True
@@ -159,9 +161,8 @@ class Graph:
         if client is None:
             return
         work_futures = replication.replicate_cudf_dataframe(
-            self.edgelist.edgelist_df,
-            client=client,
-            comms=comms)
+            self.edgelist.edgelist_df, client=client, comms=comms
+        )
 
         self.batch_edgelists = work_futures
 
@@ -175,22 +176,25 @@ class Graph:
 
         weights = None
         offsets_futures = replication.replicate_cudf_series(
-            self.adjlist.offsets,
-            client=client,
-            comms=comms)
+            self.adjlist.offsets, client=client, comms=comms
+        )
         indices_futures = replication.replicate_cudf_series(
-            self.adjlist.indices,
-            client=client,
-            comms=comms)
+            self.adjlist.indices, client=client, comms=comms
+        )
 
         if self.adjlist.weights is not None:
             weights = replication.replicate_cudf_series(self.adjlist.weights)
         else:
             weights = {worker: None for worker in offsets_futures}
 
-        merged_futures = {worker: [offsets_futures[worker],
-                                   indices_futures[worker], weights[worker]]
-                          for worker in offsets_futures}
+        merged_futures = {
+            worker: [
+                offsets_futures[worker],
+                indices_futures[worker],
+                weights[worker],
+            ]
+            for worker in offsets_futures
+        }
         self.batch_adjlists = merged_futures
 
     # FIXME: Not implemented yet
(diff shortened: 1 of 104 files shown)


GPUtester:

afender:
I need to merge https://github.com/rapidsai/cugraph/pull/1168 and skip tests before we retest this and merge.

ajschmidt8 (approved):

python/cugraph/structure/symmetrize.py
```diff
@@ -75,28 +73,84 @@ def symmetrize_df(df, src_name, dst_name):
...
+    Examples
+    --------
+    >>> M = cudf.read_csv('datasets/karate.csv', delimiter=' ',
+    >>>                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> sym_df = cugraph.symmetrize(M, '0', '1')
+    >>> G = cugraph.Graph()
+    >>> G.add_edge_list(sym_df['0]', sym_df['1'], sym_df['2'])
+    """
+    if weight_name:
+        ddf2 = df[[dst_name, src_name, weight_name]]
```
Iroy30, line 115: Maybe not of immediate concern, but does this support dst_name and src_name as list of columns?

Iroy30 (approved):

python/cugraph/structure/symmetrize.py
ChuckHastings: If we symmetrize before renumbering then we absolutely need to support a list of columns.  If we symmetrize after renumbering then we could get away with just a single column for each.

python/cugraph/structure/symmetrize.py
```diff
@@ -75,28 +72,84 @@ def symmetrize_df(df, src_name, dst_name):
...
+        ddf2.columns = [src_name, dst_name, weight_name]
+    else:
+        ddf2 = df[[dst_name, src_name]]
+        ddf2.columns = [src_name, dst_name]
+
+    ddf = df.append(ddf2).reset_index(drop=True)
+    result = (
+        ddf.groupby(by=[src_name, dst_name], as_index=False)
+        .min()
```
ChuckHastings, line 123: This is the point where we break a MultiGraph...

ChuckHastings (approved):
Implementation looks correct.  A few comments related to documentation and a flag that perhaps needs to be restored.

python/cugraph/structure/graph.py
```diff
@@ -402,7 +402,7 @@ def from_cudf_edgelist(
         else:
             value_col = None
 
-        if not self.symmetrized and not self.multi:
```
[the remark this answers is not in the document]
afender: Actually, it's me who got confused between` self.distributed `and` self.multi` :)
Adding the check back, symmetrize is not meant for multigraph.

python/cugraph/structure/symmetrize.py
afender: Right, cugraph doesn't support MG multigraph or SG multigraph in symmetrize atm.

python/cugraph/structure/symmetrize.py
afender: Are you planning on switching to symmetrizing before renumbering in the MG path? It is an option, but it seems to me that it is faster to symmetrize the COO matrix rather than a bunch of strings. That's what SG and MG design doc show. 

Let me know if this is needed for the first version or if this is something that can be added later as part of the C++ version.

python/cugraph/structure/graph.py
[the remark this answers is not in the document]
afender: b11829d

python/cugraph/structure/symmetrize.py
ChuckHastings: I would think symmetrizing after would be more efficient.  The documentation example (line 110 above) shows symmetrizing before calling before calling add_edge_list (as well as the documentation earlier).  That threw me off.

So the documentation in symmetrize.py should be updated (add_edge_list is no longer used).

Just found another problem, however.  Within `graph.py` the calls to symmetrize for SG code will work fine.  However, renumbering is now lazy.  That is, we don't actually renumber until right before executing a graph analytic - so we know whether to renumber based on whether we're executing against a CSR-style or CSC-style graph.

So if multi-gpu symmetrize is going to operate after renumbering, it would need to be part of that lazy evaluation.  The function `compute_renumber_edge_list` is called when the renumbering is triggered.

afender: Right, the doc was off. Fixed : 0ff8d63 and 8071c3e

For the lazing eval, do I understand correctly that you suggest adding symmetrize at the end of `compute_renumber_edge_list` ?  

Taking a step back, this is an interesting design discussion because when symmetrize is involved, the CSR vs CSC style becomes irrelevant. So, if this was the motivation for the lazy evaluation, this would only apply to `DiGraph` and could actually be waived for `Graph`. Since only a `Graph` would call symmetrize as part of the ETL automation and the release focus on undirected input (ie. `Graph`) I wonder if we really need to worry about it.

ChuckHastings: The motivation for lazy evaluation was to support a DiGraph.

It is true that with the implementation of symmetrize we could renumber and symmetrize eagerly (instead of lazily) on a Graph.  And our focus is on Graph vice DiGraph this release.  At the time that I shifted to lazy evaluation of renumbering we weren't sure we were going to implement symmetrize in this release.

It seems to me like extra work to remove the lazy evaluation from this release (since it already works) and then add it back in later to support DiGraphs.  It seems like the same amount of work this release, and no work later, to add the symmetrize call in the `compute_renumber_edge_list` so that it would work for either Graph or DiGraph.

PR #1165, opened 2020-09-28
[REVIEW] ENH added more Nx compatibility

BradReesWork:
Adding BSF, SSSP, and Subgraph Extraction
Fixing some test issues

### DIFF
--- a/notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
+++ b/notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
@@ -1,0 +1,220 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Benchmarking NetworkX compatibility\n",
+    "This notebook benchmark the use of a NetworkX Graph object as input into algorithms.  <p>\n",
+    "The intention of the feature is to be able to drop cuGraph into existing NetworkX code in spot where performance is not optimal.\n",
+    "\n",
+    "\n",
+    "### Betweenness Centrality\n",
+    "Both NetworkX and cuGraph allow for estimating the betweenness centrality score by using a subset of vertices rather than all the vertices.  WHile that does produce a less accurate answer, it dramatically improves performance when the sample is small.  For this test, the algorithms will use only 10% of the vertices to compute the estimate \n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "__Notebook Credits__\n",
+    "\n",
+    "* Original Authors: Bradley Rees\n",
+    "* Last Edit: 09/27/2020\n",
+    "\n",
+    "RAPIDS Versions: 0.16\n",
+    "\n",
+    "Test Hardware\n",
+    "```\n",
+    "    GV100 32G, CUDA 10,0\n",
+    "    Intel(R) Core(TM) CPU i7-7800X @ 3.50GHz\n",
+    "    32GB system memory\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import networkx as nx\n",
+    "import cugraph as cudanx\n",
+    "import time\n",
+    "import operator"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# starting number of Nodes\n",
+    "N = 100"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# average degree\n",
+    "M = 16"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def run_nx(G, k=None):\n",
+    "    t1 = time.time()\n",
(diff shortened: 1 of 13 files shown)


GPUtester:

python/cugraph/centrality/katz_centrality.py
```diff
@@ -46,6 +47,8 @@ def katz_centrality(
         (1/degree_max). Therefore, setting alpha to (1/degree_max) will
         guarantee that it will never exceed alpha_max thus in turn fulfilling
         the requirement for convergence.
+    beta : None
```
afender, line 50: Should we throw an error if `beta` is passed?

python/cugraph/traversal/bfs.py
```diff
@@ -72,3 +73,59 @@ def bfs(G, start, return_sp_counter=False):
...
+
+    Parameters
+    ----------
+    G : cugraph.graph or NetworkX.Graph
+        graph descriptor that contains connectivity information
+    source : Integer
+        The starting vertex index
+    reverse : boolean
+        If a directed graph, then process edges in a reverse direction
+        Currently not implemented
```
afender, line 92: Same here, we could throw "not implemented" errors when these parameters are passed. Otherwise, the risk is that people will fill issues saying the parameter doesn't work.

notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
```diff
@@ -0,0 +1,221 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import networkx as nx\n",
+    "import cugraph as cudanx\n",
```
afender, line 42: cudanx :) ?

notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
```diff
@@ -0,0 +1,221 @@
...
+    "    G = nx.barabasi_albert_graph(n, M)\n",
+    "    num_edges = G.number_of_edges()\n",
+    "    num_nodes = G.number_of_nodes()\n",
+    "    \n",
+    "    t1 = time.time()    \n",
+    "    nx_pr = nx.pagerank(G)\n",
+    "    time_nx = time.time() - t1\n",
+    "    \n",
+    "    t1 = time.time()    \n",
+    "    cp_pr = cugraph.pagerank(G)\n",
```
afender, line 171: The notebook comes with pre-existing speedup output for BC but not for Pagerank.

python/cugraph/traversal/bfs.py
```diff
@@ -72,3 +73,59 @@ def bfs(G, start, return_sp_counter=False):
...
+        df['sp_counter'][i] gives for the i'th vertex the number of shortest
+        path leading to it during traversal (Only if retrun_sp_counter is True)
+
+    Examples
+    --------
+    >>> M = cudf.read_csv('datasets/karate.csv', delimiter=' ',
+    >>>                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> G = cugraph.Graph()
+    >>> G.from_cudf_edgelist(M, source='0', destination='1')
+    >>> df = cugraph.bfs(G, 0)
```
afender, line 121: bfs -> bfs_edges

afender (commented):
Looks good, just added a few comments that can be addressed separately

python/cugraph/centrality/katz_centrality.py
BradReesWork: currently it is ignored and documented as being ignored.   This is how we treated the other Nx arguments, in BC for example, that we ignore.  We should define a policy for handling not used arguments.

python/cugraph/traversal/bfs.py
BradReesWork: we really need to fix the code so that those arguments are handled.  I welcome people submitting issues since it will motivate fixing and let us know that it is being used.

notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
BradReesWork: One suggestion was that we load cuGraph as "cuda.nx"  then the update would read as "cuda.nx.<algo> but python does not allow "dots" in "as" names.  The suggestion is to then do
      import cugraph as cnx
That makes for a single letter change to code

notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
BradReesWork: fixing notebook now

python/cugraph/traversal/bfs.py
BradReesWork: fixing

python/cugraph/centrality/katz_centrality.py
BradReesWork: added code to raise a  NotImplementedError error

python/cugraph/traversal/bfs.py
BradReesWork: fix

notebooks/cugraph_benchmarks/nx_cugraph_bc_benchmarking.ipynb
BradReesWork: updated in notebook to cnx

afender (approved):

ajschmidt8 (approved):

rlratzel (commented):
Sorry, it looks like I missed getting this feedback in time before merge. FWIW, here's some minor things I found.

PR #1175, opened 2020-09-30
[REVIEW] 2D cython/python infrastructure- PAGERANK

Iroy30:

PR #1177, opened 2020-09-30
[REVIEW] 2D cython/python infrastructure- BFS & SSSP

Iroy30:


GPUtester:

python/cugraph/tests/dask/test_mg_sssp.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Comms.destroy()
+    client.close()
+    cluster.close()
+
+
+@pytest.mark.skipif(
+    is_single_gpu(), reason="skipping MG testing on Single GPU system"
+)
+def test_dask_bfs(client_connection):
```
seunghwak, line 42: test_dask_bfs => test_dask_sssp

python/cugraph/dask/traversal/mg_bfs.pxd
```diff
@@ -18,11 +18,12 @@ from cugraph.structure.graph_primtypes cimport *
 from libcpp cimport bool
 
 
-cdef extern from "algorithms.hpp" namespace "cugraph":
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
 
-    cdef void bfs[VT,ET,WT](
+    cdef void call_bfs[VT,WT](
```
seunghwak, line 23: We'd better replace VT, ET, WT with vertex_t, edge_t, weight_t.

python/cugraph/dask/traversal/mg_sssp.pxd
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from cugraph.structure.graph_primtypes cimport *
+from libcpp cimport bool
+
+
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
+
+    cdef void call_sssp[VT,WT](
```
seunghwak, line 22: We'd better replace VT, ET, WT with vertex_t, edge_t, weight_t.

python/cugraph/traversal/bfs.pxd
```diff
@@ -20,11 +20,11 @@ from cugraph.structure.graph_primtypes cimport *
 from libcpp cimport bool
 
 
-cdef extern from "algorithms.hpp" namespace "cugraph":
-
-    cdef void bfs[VT,ET,WT](
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
+    cdef void call_bfs[VT,WT](
```
seunghwak, line 24: We'd better replace VT, ET, WT with vertex_t, edge_t, weight_t.

python/cugraph/traversal/sssp.pxd
```diff
@@ -18,10 +18,12 @@
 
 from cugraph.structure.graph_primtypes cimport *
 
-cdef extern from "algorithms.hpp" namespace "cugraph":
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
 
-    cdef void sssp[VT, ET, WT](
-        const GraphCSRView[VT, ET, WT] &graph,
+    cdef void call_sssp[VT, WT](
```
seunghwak, line 23: We'd better replace VT, ET, WT with vertex_t, edge_t, weight_t. Please search for this.

seunghwak (approved):
Except for renaming VT, ET, WT to vertex_t, edge_t, weight_t, this looks good to me.

And if you find the fix for the warnings I mentioned in the PageRank binding PR, don't forget to apply that to this PR as well.

Thank you!!!

python/cugraph/tests/dask/test_mg_sssp.py
```diff
@@ -0,0 +1,85 @@
...
+        dtype=["int32", "int32", "float32"],
+    )
+
+    g = cugraph.DiGraph()
+    g.from_cudf_edgelist(df, "src", "dst", renumber=True)
+
+    dg = cugraph.DiGraph()
+    dg.from_dask_cudf_edgelist(ddf, "src", "dst")
+
+    expected_dist = cugraph.sssp(g, 0)
```
seunghwak, line 69: and it seems like in expected_dist, unreachable vertices get distance of 2147483647 (INT_MAX?), but edge weight type is set to float, and based on

https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/src/traversal/traversal_common.cuh#L135
```
template <>
struct vec_t<float> {
  typedef float4 vec4;
  typedef float2 vec2;
  static constexpr float max = std::numeric_limits<float>::max();
};
```

and

https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/src/traversal/sssp.cu#L118
```
template <typename IndexType, typename DistType>
void SSSP<IndexType, DistType>::traverse(IndexType source_vertex)
{
  // Init distances to infinities
  traversal::fill_vec(distances, n, traversal::vec_t<DistType>::max, stream);
  traversal::fill_vec(next_distances, n, traversal::vec_t<DistType>::max, stream);
```

I think unreachable vertices should have the distance of std::numeric_limits<float>::max().

Any thoughts on why?

seunghwak: And for the result_dist, predecessors of unreachable vertices are printed as <NA>
```
RESULT
          distance  vertex predecessor
0     1.000000e+00       1           0
1     2.000000e+00     946           1
2     1.000000e+00    1084           0
0     3.402823e+38    1242        <NA>
1     3.402823e+38    1249        <NA>
...            ...     ...         ...
1453  3.402823e+38     108        <NA>
1454  3.402823e+38     724        <NA>
1455  3.402823e+38     111        <NA>
1456  3.402823e+38     991        <NA>
1457  3.402823e+38     112        <NA>
```
Any thoughts on why? If I am not mistaken, they are set to invalid_vertex_id (https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/include/experimental/graph.hpp#L176); that is -1 for vertex_t is int32_t.

python/cugraph/dask/traversal/mg_bfs.pxd
Iroy30: updated

python/cugraph/tests/dask/test_mg_sssp.py
Iroy30: updated

python/cugraph/dask/traversal/mg_sssp.pxd
Iroy30: updated

python/cugraph/traversal/sssp.pxd
Iroy30: updated

python/cugraph/traversal/bfs.pxd
Iroy30: updated

python/cugraph/tests/dask/test_mg_sssp.py
Iroy30: The reason for int_max is because for single gpu sssp if no weights are provided, bfs is called internally which has int32 as distance. I have updated sssp tests to add weights to graph which now gives float/double max for distances. 
Also updated predecessors to fill -1 when NA

BradReesWork:
rerun tests

python/cugraph/tests/dask/test_mg_sssp.py
ChuckHastings: Second problem looks similar to the issue that was fixed with pagerank values... note that the index restarts... perhaps the predecessor and distance aren't getting gathered from the correct GPU?

cpp/src/utilities/cython.cu
```diff
@@ -476,6 +476,185 @@ std::pair<size_t, weight_t> call_louvain(raft::handle_t const& handle,
...
+             reinterpret_cast<int32_t*>(personalization_subset),
+             reinterpret_cast<double*>(personalization_values),
+             alpha,
+             tolerance,
+             max_iter,
+             has_guess);
+    graph_container.graph_ptr_union.GraphCSCViewDoublePtr->get_vertex_identifiers(
+      reinterpret_cast<int32_t*>(identifiers));
+  } else if (graph_container.graph_type == graphTypeEnum::graph_t) {
+    if (graph_container.edgeType == numberTypeEnum::int32Type) {
```
ChuckHastings, line 520: This would be cleaner (for pagerank, BFS and SSSP) to use the technique used by louvain...  define a functor for each algorithm and then call `detail::call_function<true, void>(handle, graph_container, functor);`

ChuckHastings: Although we could defer this work to 0.17

python/cugraph/dask/traversal/mg_bfs_wrapper.pyx
```diff
@@ -21,7 +21,16 @@ from cugraph.structure.graph_primtypes cimport *
 import cugraph.structure.graph_primtypes_wrapper as graph_primtypes_wrapper
 from libc.stdint cimport uintptr_t
 
-def mg_bfs(input_df, local_data, rank, handle, start, result_len, return_distances=False):
+def mg_bfs(input_df,
+           num_global_verts,
+           num_global_edges,
+           partition_row_size,
+           partition_col_size,
```
seunghwak, line 28: partition_row_size & partition_col_size are no longer necessary, same for SSSP

afender (approved):
Looks good, but some conflicts here as well.

python/cugraph/dask/traversal/sssp.py
```diff
@@ -0,0 +1,126 @@
...
+    graph.
+    The input graph must contain edge list as  dask-cudf dataframe with
+    one partition per GPU.
+
+    Parameters
+    ----------
+    graph : cugraph.DiGraph
+        cuGraph graph descriptor, should contain the connectivity information
+        as dask cudf edge list dataframe(edge weights are not used for this
+        algorithm). Undirected Graph not currently supported.
```
afender, line 60: "Undirected Graph not currently supported."
Could we add that undirected graphs should be passed as Digraph with edges in both directions (apply to other MG APIs as well).

python/cugraph/dask/traversal/sssp.py
```diff
@@ -0,0 +1,126 @@
...
+    source : Integer
+        Specify source vertex
+
+    Returns
+    -------
+    df : cudf.DataFrame
```
afender, line 66: Should this be a dask data frame or are we assuming that 3 global V sets will fit in one GPU?

afender: Saw that in #1175 too

ChuckHastings (approved):

PR #1186, opened 2020-10-05
[REVEIW] BLD Installing raft headers under cugraph

afender:
In 0.16, we just include the necessary headers under the cugraph and cuml includes. 
In the 0.17 timeline, we can provide a means to install raft globally, but that will require shipping a new conda package and matching hashes.


GPUtester:

BradReesWork (approved):

rlratzel (approved):

PR #1196, opened 2020-10-07
[REVIEW] BUG Move subcomms init outside of individual algorithm functions

### LINKED ISSUE #1065: [FEA] sub-communicator initialization for 2D partitioning support

rlratzel:
This PR moves the subcomms init outside of the individual algorithm functions and up to the function a user calls to initialize comms. This allows the same subcomm configuration to be used across algorithms and also allows users to customize the call to optimize based on GPU configuration.

Without this, subcomms will be repeatedly created/initialized for each algo call, resulting in problems.

closes #1065

(thanks to @Iroy30 for commits https://github.com/rapidsai/cugraph/pull/1196/commits/71d4a0b652060d8bf365599028af3653d87b6838 and https://github.com/rapidsai/cugraph/pull/1196/commits/79f131c567ab9ccd8ecc601d60cb58c85b3aec9c)


GPUtester:

rlratzel:
closes #1065

python/cugraph/comms/comms.py
```diff
@@ -14,15 +14,65 @@
...
+    __subcomm = (prows, pcols, partition_type)
+
+
+def _subcomm_init(sID, partition_row_size):
+    handle = get_handle(sID)
+    c_init_subcomms(handle, partition_row_size)
 
 
 # Intialize Comms. If explicit Comms not provided as arg,
 # default Comms are initialized as per client information.
```
seunghwak, line 70: As this is a public interface, don't we need to update the documentation?

seunghwak (approved):
Besides the comment about documentation, this looks good to me. Thank you for the quick fix!

ChuckHastings (approved):
Much cleaner!

python/cugraph/comms/comms.py
seunghwak: And we eventually need to remove the C++ partition manager class. This class is basically a misnomer... C++ is not managing partitioning. The python layer is managing partitioning and the C++ layer is just performing a passive task based on instructions from the python layer.

rlratzel: Yes, thank you. I'll work on that now.

CHANGELOG.md
```diff
@@ -39,8 +39,9 @@
 - PR #1166 Fix misspelling of function calls in asserts causing debug build to fail
 - PR #1180 BLD Adopt RAFT model for cuhornet dependency
 - PR #1181 Fix notebook error handling in CI
-- PR #1186 BLD Installing raft headers under cugraph 
+- PR #1186 BLD Installing raft headers under cugraph
 - PR #1192 Fix benchmark notes and documentation issues in graph.py
+- PR #1196 Move subcomms init outside of individual algorithm functions
```
afender, line 44: Consider using new enhanced tags for automated reports:

```
ENH       Enhancement, new functionality
BUG       Bug Fix
DOC       Additions/updates to documentation
TST         Additions/updates to tests
BLD        Updates to the build process/scripts
PRF        Performance improvement
CLN        Code cleanup
```

cpp/include/utilities/cython.hpp
```diff
@@ -201,5 +199,8 @@ std::pair<size_t, weight_t> call_louvain(raft::handle_t const& handle,
                                          size_t max_level,
                                          weight_t resolution);
 
+// Helper for setting up subcommunicators
+void init_subcomms(raft::handle_t& handle, size_t row_comm_size);
```
afender, line 203: Public API header right? 
Users will wonder what `row_comm_size` is, how it should be set. Any specific default behavior should be documented as well.

python/cugraph/comms/comms.py
```diff
@@ -114,6 +174,10 @@ def get_worker_id(sID):
     return sessionstate['wid']
 
 
-def get_n_workers(sID):
-    sessionstate = worker_state(sID)
-    return sessionstate['nworkers']
+def get_n_workers(sID=None):
```
afender, line 177: I could sware I saw several ways in the python layer that already get the number of workers/GPUs:
- get_n_workers() (from read_utils)
- len(get_visible_devices())
- len(numba.cuda.gpus)

 We should document why this one is different/better or avoid multiplying code doing the same thing.

cpp/include/utilities/cython.hpp
rlratzel: Thanks, working on that now for the next commit.

python/cugraph/comms/comms.py
rlratzel: Agreed, I'll look into this for the next commit.

afender (approved):

rlratzel:
rerun tests

PR #1227, opened 2020-10-15
[REVIEW] Pin cmake policies to 3.17

kkraus14:
Planning to upgrade to cmake 3.18 in the near future and the CUDA_ARCHITECTURES change is a breaking change, so pinning to 3.17 policies to prevent the breakages.


GPUtester:

afender (approved):
I did try adding CUDA_ARCHITECTURES as part of the debugging efforts this weekend actually. It seems to go smoothly for cugraph once  `CUDA_ARCHITECTURES` is added. 
Also, I noticed this can be passed to older cmake versions concurrently with the legacy one (well, at least it built/ran locally).

BradReesWork:
rerun tests

PR #1237, opened 2020-10-20
[REVIEW] dask personalization, fix df query

Iroy30:


GPUtester:

python/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -141,21 +138,36 @@ def pagerank(input_graph,
...
+
+        result = [client.submit(call_pagerank,
+                                Comms.get_session_id(),
+                                wf[1],
+                                num_verts,
+                                num_edges,
+                                vertex_partition_offsets,
+                                alpha,
+                                max_iter,
+                                tol,
```
seunghwak, line 152: I think we need to discuss more about ```tol```.

The code below is what NetworkX does (compare ```err``` with ```N * tol```). This requires setting tol to a smaller value if N gets large (tol here basically means tolerance for a single PageRank value not the entire set of PageRank vector).

https://github.com/networkx/networkx/blob/master/networkx/algorithms/link_analysis/pagerank_alg.py#L155
```
err = sum([abs(x[n] - xlast[n]) for n in x])
        if err < N * tol:
            return x
```

The new PageRank algorithm adopts this NetworkX logic to determine convergence, so should we follow the NetworkX logic (and set tol to a smaller value with a larger N) or better remove ```N * ``` from the new PageRank code?

seunghwak (commented):
Please check that Personalization vector is normalized. The new MNMG PageRank code does not require this but it seems like the old SG PageRank code requires this. This leads to test failure with netscience.csv.

Iroy30:
rerun tests

Iroy30:
rerun tests

seunghwak:
Have you tested this with netscience.csv? The legacy SG personalized PageRank expects personalization values to be normalized and it seems like our test code does not normalize; this led to test failure even if (I believe) the new MNMG personalized PageRank returns correct values.

Iroy30:
> Have you tested this with netscience.csv? The legacy SG personalized PageRank expects personalization values to be normalized and it seems like our test code does not normalize; this led to test failure even if (I believe) the new MNMG personalized PageRank returns correct values.

The legacy SG personalized PageRank normalizes internally and doesn't require the user to send in normalized values. However, netscience.csv fails in the test. Need to look deeper into what is happening

seunghwak:
> > Have you tested this with netscience.csv? The legacy SG personalized PageRank expects personalization values to be normalized and it seems like our test code does not normalize; this led to test failure even if (I believe) the new MNMG personalized PageRank returns correct values.
> 
> The legacy SG personalized PageRank normalizes internally and doesn't require the user to send in normalized values. However, netscience.csv fails in the test. Need to look deeper into what is happening

When I tested, the sum of PageRank values decrease in every iteration with the legacy SG PageRank; it should stay at 1.0 if it's working properly.

afender (approved):
looks good
let's try and squeeze the notebook fix in this one

Iroy30:
> looks good
> let's try and squeeze the notebook fix in this one

p2p is defaulted to True now so the notebooks work without the explicit setting of p2p. However I edited the notebooks regardless so users are aware p2p is set as True

Iroy30:
> > > Have you tested this with netscience.csv? The legacy SG personalized PageRank expects personalization values to be normalized and it seems like our test code does not normalize; this led to test failure even if (I believe) the new MNMG personalized PageRank returns correct values.
> > 
> > 
> > The legacy SG personalized PageRank normalizes internally and doesn't require the user to send in normalized values. However, netscience.csv fails in the test. Need to look deeper into what is happening
> 
> When I tested, the sum of PageRank values decreases in every iteration with the legacy SG PageRank; it should stay at 1.0 if it's working properly.

There was an issue with netscience due to personalization nodes initialization in python. Fixed it.

seunghwak (approved):

PR #1242, opened 2020-10-23
[REVIEW] BUG Calling gunrock cmake using explicit -D options, re-enabling C++ tests

rlratzel:
* Changes to call gunrock cmake using explicit -D options for each arch, similar to a prior version of the CMakeLists.txt, since the update to expand a variable into the arch options was being expanded as a single quoted value with spaces that wasn't recognized, resulting in gunrock being built for only compute 60 support. This caused C++ test crashes.
* Re-enabled C++ tests.

This was tested by building cugraph with this change and running the Pagerank C++ test 1000 times without failures/crashes.  Prior to this, the Pagerank C++ test (and others) would crash after 1-50 test run iterations.

The issue and corresponding fix was verified with a standalone reproducer, which can be seen here: https://github.com/rlratzel/cugraph_build_problem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,10 @@
 ## Improvements
 
 ## Bug Fixes
+- PR #1242 Calling gunrock cmake using explicit -D options, re-enabling C++ tests
 
 
-# cuGraph 0.16.0 (Date TBD)
+# cuGraph 0.16.0 (21 Oct 2020)
 
 ## New Features
 - PR #1098 Add new graph classes to support 2D partitioning
--- a/ci/test.sh
+++ b/ci/test.sh
@@ -64,16 +64,12 @@ else
     cd $WORKSPACE/ci/artifacts/cugraph/cpu/conda_work/cpp/build
 fi
 
-# FIXME: temporarily disabling all C++ tests for 0.16 due to intermittent
-# failures from what appears to be an issue with Thrust (which does not appear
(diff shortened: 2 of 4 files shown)


GPUtester:

ajschmidt8 (approved):

ajschmidt8:
@rlratzel, nice job. I created an integration PR to revert these changes in that repo as well. Please let me know if that PR can be merged with this one.

https://github.com/rapidsai/integration/pull/163

seunghwak:
Wow, so if I am not mistaken, gunrock (built for an outdated architecture) affected the cugraph pagerank/SSSP behavior outside gunrock? Very surprising and great catch!

rlratzel:
> Wow, so if I am not mistaken, gunrock (built for an outdated architecture) affected the cugraph pagerank/SSSP behavior outside gunrock? Very surprising and great catch!

I admittedly don't fully understand why, but when the test binary loads a .so with gunrock code compiled for the old arch, the test binary will _occasionally_ crash in a Thrust call. I image it may not have to be gunrock code, but gunrock exposed the issue with the incorrect arch option.  I created a minimal reproducer here to verify the problem and the fix: https://github.com/rlratzel/cugraph_build_problem

I'd like to better understand why we see this particular crash even though the code being built for the old arch isn't even being called, but I figured we should move forward with this as-is if it allows us to re-enable C++ tests sooner.

BradReesWork (approved):

PR #1246, opened 2020-10-28
[REVIEW] Use latest Gunrock, update HITS implementation

### LINKED ISSUE #1154: [BUG] HITS Hub and Auth values are wrong
There appears to be no differences between the HUB and Auth scores - they should be different.  Ans the values do not match networkx

ChuckHastings:
Closes #1154 

This PR makes the following changes to our gunrock integration:
1. Update the CMake files to point directly to the gunrock repository instead of cugunrock.  This makes it easier to integrate their changes.
1. Latest repo adds features to allow the HITS function to accept data already on device memory and to accept a tolerance parameter and to select which normalization function to use (networkx uses the L1 norm gunrock was using the L2 norm, now we can specify which one that gunrock should use
1. With the updated HITS we can now compare results to networkx, so updated the python test to compare results to networkx.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ## Bug Fixes
 - PR #1242 Calling gunrock cmake using explicit -D options, re-enabling C++ tests
+- PR #1246 Use latest Gunrock, update HITS implementation
 
 
 # cuGraph 0.16.0 (21 Oct 2020)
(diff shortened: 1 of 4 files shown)


GPUtester:

BradReesWork (approved):

afender (approved):

PR #1250, opened 2020-10-30
[REVIEW] BUG Updated cuco commit hash to latest as of 2020-10-30

rlratzel:
Updated cuco commit hash to latest as of 2020-10-30.  This fixes build breakages as reported here: https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cugraph/job/branches/job/cugraph-cpu-branch-build/5381/console

Update: removing the `GIT_SHALLOW` param is the actual fix (thanks to @ChuckHastings), since its use is only supported when a branch or tag are specified (not a hash): https://cmake.org/cmake/help/latest/module/ExternalProject.html
I'll keep the change to update the hash though since it's just a README change and might help future debugging efforts by being of a more recent date (fewer commits to bisect)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ## Bug Fixes
 - PR #1242 Calling gunrock cmake using explicit -D options, re-enabling C++ tests
-- PR #1250 Updated cuco commit hash to latest as of 2020-10-30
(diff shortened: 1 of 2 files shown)


GPUtester:

BradReesWork (approved):

ChuckHastings (approved):
Based on experiments and documentation, removing GIT_SHALLOW is the necessary step.

Updating to the latest is a nice extra!

afender (approved):

PR #1251, opened 2020-10-30
[REVIEW] BUG Changed the MG context testing class to use updated parameters passed in from the individual tests

rlratzel:
Changed the MG context testing class to use updated parameters passed in from the individual tests. This allows tests to override the context with a different cluster instance and a p2p arg for the comms init call.

This PR will resolve issues in the nightly automated MG tests and allow us to see other potential failures.


GPUtester:

python/cugraph/tests/dask/mg_context.py
```diff
@@ -54,11 +54,12 @@ class MGContext:
         Number of devices to use, verification must be done prior to call
         to ensure that there are enough devices available.
```
ChuckHastings, line 55: Update documentation to reflect all parameters (new and the rmm_managed_memory that was missing before)

python/cugraph/tests/dask/mg_context.py
```diff
@@ -78,17 +79,18 @@ def _prepare_mg(self):
         self._prepare_comms()
 
     def _prepare_cluster(self):
-        self._cluster = CUDACluster(
-            n_workers=self._number_of_devices,
-            rmm_managed_memory=self._rmm_managed_memory
-        )
+        if self._cluster is not None:
```
ChuckHastings, line 82: Is there a reason to have _prepare_cluster be a separate thing?  Couldn't this just be done in the `__init__` method?

xcadet, line 82: Shouldn't it create a new CUDA Cluster only if the cluster is None?
```suggestion
        if self._cluster is None:
```

python/cugraph/tests/dask/mg_context.py
```diff
@@ -54,11 +54,12 @@ class MGContext:
         Number of devices to use, verification must be done prior to call
         to ensure that there are enough devices available.
     """
-    def __init__(self, number_of_devices=None, rmm_managed_memory=False):
+    def __init__(self, number_of_devices=None, rmm_managed_memory=False, cluster=None, p2p=False):
```
xcadet, line 57: If a cluster is provided, and ```MGContext``` is used as a context manager the ```__exit__``` section will close the cluster, which might not be the expected behavior (as we might want to reuse the cluster with a different client). Having an extra boolean ```close_on_exit = False``` could give more control over it?

python/cugraph/tests/dask/mg_context.py
rlratzel: Thanks, fixed in the latest commit.

python/cugraph/tests/dask/mg_context.py
rlratzel: Yes, thanks, I realized that just after I pushed this. Fixed in the latest commit.

rlratzel: I changed it to your suggestion in the latest commit.

ChuckHastings (approved):

python/cugraph/tests/dask/mg_context.py
rlratzel: I ended up removing the ability to override the entire cluster in the latest commit, since I realized it was already creating a `LocalCUDACluster`. I can add the ability to override the cluster completely (and apply your suggestion to add a flag to not close on exit) in the future if we end up needing that.  The `p2p` arg to the Comms init call to initialize the UCX endpoints was what we really needed for the current tests.

xcadet (approved):

afender (approved):

PR #1253, opened 2020-11-04
[REVIEW] BUG MG test fixes: updated additional comms.initialize() calls, fixed dask DataFrame comparisons

rlratzel:
MG test fixes: updated additional comms.initialize() calls, fixed errors resulting from calls that assume a DataFrame but were getting a dask DataFrame.

NOTE: There's currently some tech debt in these tests in that some use a MG context class, and others use a pytest fixture to do effectively the same thing. We'll need to revisit these tests and hopefully consolidate the setup code.


GPUtester:

BradReesWork (approved):

afender (commented):
Oops, I was hoping we could have sneaked in the ` p2p=true` in the notebook as well as per this morning discussion... Maybe in one of @Iroy30 PR then?

PR #1267, opened 2020-11-11
[REVIEW] Enhance compile time

aschaffer:
Look into alternative ways to instantiate templates to improve compilation time.


GPUtester:

cpp/CMakeLists.txt
```diff
@@ -58,39 +58,31 @@ set(GUNROCK_GENCODE_SM72 "OFF")
...
 set(GUNROCK_GENCODE_SM80 "OFF")
 
-# Check for aarch64 vs workstation architectures
-if(CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
-  message(STATUS "CMAKE Detected aarch64 CPU architecture, selecting appropriate gencodes")
-  # This is being build for Linux4Tegra or SBSA ARM64 CUDA
-  set(GPU_ARCHS "62") # Default minimum CUDA GenCode - not supported by gunrock
-  if(CUDA_VERSION_MAJOR GREATER_EQUAL 9)
-    set(GPU_ARCHS "${GPU_ARCHS};72")
-    set(GUNROCK_GENCODE_SM72 "ON")
```
rlratzel, line 68: Where are the `GUNROCK_GENCODE_SM*` variables being set to `"ON"` for the relevant architectures? The Gunrock build needs these to be built properly, otherwise Gunrock arch build options all default to `"OFF"`.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # script, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libcugraph cugraph docs -v -g -n --show_depr_warn -h --help"
+VALIDARGS="clean libcugraph cugraph docs -v -g -n --allgpuarch --show_depr_warn -h --help"
 HELP="$0 [<target> ...] [<flag> ...]
```
rlratzel, line 22: While you're updating this, can you also update the part that describes the default actions (ie. "default action (no args) is to...") since it's out of date and doesn't mention doc builds.

rlratzel (changes requested):

build.sh
aschaffer: Okay, what should that say? 

" default action (no args) is to build and install 'libcugraph' then 'cugraph' targets *_and then documentation_*"

That okay?

cpp/CMakeLists.txt
rlratzel: This was fixed in commit https://github.com/rapidsai/cugraph/pull/1267/commits/03573ecbd231395d47d26ac6c02cebe79f08227b

build.sh
rlratzel: looks better, thanks.

rlratzel (approved):

aschaffer:
Rerun tests.

ajschmidt8 (approved):

cpp/include/experimental/eidecl_graph.hpp
```diff
@@ -0,0 +1,85 @@
...
+extern template class graph_t<int32_t, int32_t, double, false, false, void>;
+extern template class graph_t<int32_t, int64_t, float, true, true, void>;
+extern template class graph_t<int32_t, int64_t, float, true, false, void>;
+extern template class graph_t<int32_t, int64_t, float, false, true, void>;
+extern template class graph_t<int32_t, int64_t, float, false, false, void>;
+extern template class graph_t<int32_t, int64_t, double, true, true, void>;
+extern template class graph_t<int32_t, int64_t, double, true, false, void>;
+extern template class graph_t<int32_t, int64_t, double, false, true, void>;
+extern template class graph_t<int32_t, int64_t, double, false, false, void>;
+extern template class graph_t<int64_t, int32_t, float, true, true, void>;
```
ChuckHastings, line 36: Same comment on <int64_t, int32_t, *> throughout this file.

cpp/include/eidecl_graph.hpp
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+namespace cugraph {
+extern template class GraphViewBase<int32_t, int32_t, float>;
+extern template class GraphViewBase<int32_t, int32_t, double>;
+extern template class GraphViewBase<int32_t, int64_t, float>;
+extern template class GraphViewBase<int32_t, int64_t, double>;
+extern template class GraphViewBase<int64_t, int32_t, float>;
```
ChuckHastings, line 23: The combination <int64_t, int32_t, *> is a combination we should avoid through the code in this PR.

I think it increases the number of combinations we need to build into the library with no real benefit.  The only way <int64_t, int32_t, *> would be useful is if the average degree of vertices in the graph < 2 and the number of edges is close to 2^31-1.  I've never encountered a graph with these characteristics that would actually be meaningful to analyze.

I think in the unlikely event of this case, requiring the user to go to <int64_t, int64_t, *> is not unreasonable.

Note this pattern recurs below in this file and in several other files.

cpp/include/eidir_graph.hpp
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+#pragma once
+
+namespace cugraph {
+template class GraphViewBase<int32_t, int32_t, float>;
+template class GraphViewBase<int32_t, int32_t, double>;
+template class GraphViewBase<int32_t, int64_t, float>;
+template class GraphViewBase<int32_t, int64_t, double>;
+template class GraphViewBase<int64_t, int32_t, float>;
```
ChuckHastings, line 23: Same comment on <int64_t, int32_t, *> throughout this file.

cpp/include/graph.hpp
```diff
@@ -475,6 +475,10 @@ class GraphCompressedSparseBase {
   bool has_data_p{false};
 
  public:
+  // previously missing, but invoked cnstr{
```
ChuckHastings, line 478: Do we intend to define constructors that match the default?  As I recall the cudf standard was to not define constructors if the default constructor was identical.

If we feel better defining it, would
`GraphCompressedSparseBase(void) = default;`
or 
`GraphCompressedSparseBase() = default;`
be better?

cpp/src/community/louvain.cu
```diff
@@ -15,7 +15,17 @@
  */
 
 #include <community/louvain.cuh>
+
+// Disable louvain(experimenta::graph_view_t,...)
+// versions for GPU architectures < 700
+//(this is because cuco/static_map.cuh would not
+// compile on those)
+//
```
ChuckHastings, line 23: Suggest adding a "FIXME" here reminding us to delete this after RAPIDS drops support for Pascal.

ChuckHastings (changes requested):
A few things I think we should consider excluding.  Otherwise I can't wait to have this merged.

cpp/include/eidecl_graph.hpp
aschaffer: But that's exactly what the code says: " _supress_ those instantiations by using `extern` on them". If those would have been in `eidir_*.hpp` then that would've been a problem.

cpp/include/experimental/eidecl_graph.hpp
aschaffer: In the ei*decl* files, it's fine. It means they are suppressed. If those would be present in the *eidir* files then that would trigger instantiations.

cpp/include/eidir_graph.hpp
aschaffer: Those don't exist in that header anymore. I fixed that a few days ago.

cpp/src/community/louvain.cu
aschaffer: Done.

cpp/include/graph.hpp
aschaffer: Done.

cpp/include/eidecl_graph.hpp
aschaffer: But, done (for the purpose of symmetry of eidir/eidecl...which really should not necessarily be symmetric; their structure is driven by different needs).

ChuckHastings (approved):

BradReesWork (approved):

PR #1271, opened 2020-11-13
[REVIEW] Add extra check to make SG Louvain deterministic

### LINKED ISSUE #1258: [FEA]  Louvain reproducibility
There is interest in making Louvain reproducible in some of our example notebooks in https://github.com/clara-parabricks/rapids-single-cell-examples. AFAICT, the popular (CPU-based) `Python` Louvain llibrary that is used in ScanPy is parallel but also accepts a `random_state` argument and is fully reproducible. 

It would be very useful if we were able to do this in cugraph's implementation as well.

ChuckHastings:
Closes #1258 

Original reduction to compute optimal delta modularity was dependent on how the work got partitioned to different GPU threads.  This change makes the result deterministic by picking the smaller neighbor vertex id in case of a tie.

Added check in unit test to verify the exact result since it's now deterministic.


GPUtester:

afender (approved):
lgtm

ChuckHastings:
rerun tests

BradReesWork (approved):

PR #1277, opened 2020-11-21
[REVIEW] FEA SciPy sparse matrix input support for WCC, SCC, SSSP, and BFS

rlratzel:
Added support for SciPy sparse matrix inputs to WCC, SCC, SSSP, and BFS.  This PR also adds APIs which allow cugraph to be used as a drop-in replacement for the corresponding SciPy sparse graph routines [here](https://docs.scipy.org/doc/scipy/reference/sparse.csgraph.html#module-scipy.sparse.csgraph). This compatibility extends to adding new optional parameters which are handled based on the input type (Graph obj or matrix) and the specific algo.

Additional refactorings were done to utilities to make the various input/output checking more reusable and compact, as well as a utility to clean up how optional dependencies are imported.


python/cugraph/utilities/utils.py
```diff
@@ -21,8 +21,25 @@
     from cupyx.scipy.sparse.coo import coo_matrix as cp_coo_matrix
```
rlratzel, line 21: I could use the new `import_optional()` utility defined below, but that would mean I'd either have to define it above the imports, or import all optional dependencies at the end of the file after the function is defined.

BradReesWork (approved):

rlratzel:
rerun tests

rlratzel:
rerun tests

rlratzel:
rerun tests

PR #1279, opened 2020-11-24
[REVIEW] add self loop check variable

### LINKED ISSUE #536: [FEA] self loop check
We should have a feature that identifies if a graph has self-loops. We could compute it once and store the information inside the graph structure.
Nx checks for that for features like k-core : https://github.com/rapidsai/cugraph/pull/520/files/8dc6015561d345f4deebeb91074c5172c638feac#r330361903

Iroy30:
close #536


afender (approved):
Looks good! Please add `close #536` in the descritpion.

PR #1280, opened 2020-11-24
[REVIEW] enable multigraph

Iroy30:


GPUtester:

python/cugraph/structure/symmetrize.py
```diff
@@ -54,26 +54,30 @@ def symmetrize_df(df, src_name, dst_name):
...
-            )
-        elif name == dst_name:
-            gdf[dst_name] = df[dst_name].append(
-                df[src_name], ignore_index=True
-            )
-        else:
-            gdf[name] = df[name].append(df[name], ignore_index=True)
-
-    return gdf.groupby(by=[src_name, dst_name], as_index=False).min()
+    if symmetrize:
```
BradReesWork, line 62: We should talk about, and document, what symmetrize means for Multigraphs.  
 Assume the there are edges: A-B, A-B, A-C, B-A.     
Is the B-A edge unique and should be  symmetrize, or is it already a reverse of one of the A-B edges.  
Currently there are two edges from A to B, and one from B to A.  Should symmetrize produce three edges between the pair or two?

python/cugraph/structure/symmetrize.py
```diff
@@ -16,7 +16,7 @@
 import dask_cudf
 
 
-def symmetrize_df(df, src_name, dst_name):
+def symmetrize_df(df, src_name, dst_name, multi=False, symmetrize=True):
```
BradReesWork, line 19: We should discuss rather we symmetrize on dataframe or if a graph class should be passed in

BradReesWork (changes requested):

afender (changes requested):
I think it is best to capture our multigraph discussions in the form of multigraph class documentation and add it as part of this PR from the start while it is still fresh in our mind.

python/cugraph/structure/symmetrize.py
Iroy30: As per our MultiGraph discussion we concluded that in cases like A-B, A-B, A-C, B-A, there should be no assumption made and every edge should be treated as unique.

BradReesWork (approved):

python/cugraph/structure/graph.py
```diff
@@ -113,24 +112,40 @@ def __init__(
         self.batch_transposed_adjlists = None
 
         if m_graph is not None:
-            if (type(self) is Graph and type(m_graph) is MultiGraph) or (
+            """if (type(self) is Graph and type(m_graph) is MultiGraph) or (
```
rlratzel, line 115: was this commented out section of code (using triple-quotes) meant to stay here, or should it be deleted now?

python/cugraph/structure/graph.py
```diff
@@ -409,24 +430,19 @@ def from_cudf_edgelist(
...
-            else:
-                source_col, dest_col = symmetrize(source_col, dest_col)
+        if value_col is not None:
+            source_col, dest_col, value_col = symmetrize(
+                source_col, dest_col, value_col, multi=self.multi,
+                symmetrize=not self.symmetrized)
+        else:
+            source_col, dest_col = symmetrize(
+                source_col, dest_col, multi=self.multi,
+                symmetrize=not self.symmetrized)
```
rlratzel, line 445: This might be out-of-scope for this PR, but I wonder if we can just change `symmeterize()` to always return a 3-tuple with a None for the value column if `value_col` passed in is None.  I think this will allow us to get rid of the `if-else` block here and just have a single call.

python/cugraph/structure/graph.py
```diff
@@ -1513,7 +1529,7 @@ def add_internal_vertex_id(
 class DiGraph(Graph):
     def __init__(self, m_graph=None, edge_attr=None):
```
rlratzel, line 1530: Maybe I'm missing something, but is the `edge_attr` arg here being used anymore?

python/cugraph/structure/symmetrize.py
rlratzel, line 19: Can you update the docstring with the new args?  I'm especially interested in the expected usage of the `symmetrize` arg, since at first glance it seems odd to call symmetrize_df with a flag telling it not to symmetrize.

python/cugraph/structure/symmetrize.py
```diff
@@ -129,7 +133,8 @@ def symmetrize_ddf(df, src_name, dst_name, weight_name=None):
     return result
 
 
-def symmetrize(source_col, dest_col, value_col=None):
+def symmetrize(source_col, dest_col, value_col=None, multi=False,
+               symmetrize=True):
```
rlratzel, line 137: Same with this one - it should have its docstring updated with the new args.

python/cugraph/tests/test_multigraph.py
```diff
@@ -0,0 +1,96 @@
...
+import networkx as nx
+from cugraph.tests import utils
+import pytest
+import gc
+import numpy as np
+
+
+@pytest.mark.parametrize("graph_file", utils.DATASETS)
+def test_multigraph(graph_file):
+    gc.collect()
```
rlratzel, line 11: I've started including a separate setup function that has this in it so it's not repeated in every test.

python/cugraph/tests/test_multigraph.py
```diff
@@ -0,0 +1,96 @@
...
+    cuM = utils.read_csv_file(graph_file)
```
rlratzel, line 12: We're working on some new test fixtures that will clean up much of the boilerplate (iterate over a set of input files, create a graph obj, create a Nx graph obj, etc.) that's used in many tests.  Would you mind adding a `FIXME` saying to remove most of this test setup and migrate to the new test fixtures when available?

rlratzel (changes requested):
Looks good, just a few minor requests.

python/cugraph/structure/graph.py
Iroy30: Removed

python/cugraph/structure/graph.py
Iroy30: Will fix this in a separate PR.

python/cugraph/structure/graph.py
Iroy30: Removed commented lines

python/cugraph/structure/symmetrize.py
Iroy30: Added doc

python/cugraph/structure/symmetrize.py
Iroy30: Added doc. This might be updated in a symmetrize PR to accept graphs

python/cugraph/tests/test_multigraph.py
Iroy30: updated

python/cugraph/tests/test_multigraph.py
Iroy30: Added a FIXME

python/cugraph/structure/symmetrize.py
Iroy30: To be added in a separate PR

rlratzel (approved):

afender (approved):

Issue #1282, opened 2020-11-27
[BUG] Betweenness Centrality crashes when there are edgge weights


BradReesWork:
The Betweenness Centrality algorithm crashes on weighted graphs.   This issues was reported by user geoHeil

```bc = cugraph.betweenness_centrality(G_undir)
bc
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-45-d74c605b30cb> in <module>
----> 1 bc = cugraph.betweenness_centrality(G_undir)
      2 bc
~/development/conda_envs/eco_3_graph/lib/python3.8/site-packages/cugraph/centrality/betweenness_centrality.py in betweenness_centrality(G, k, normalized, weight, endpoints, seed, result_dtype)
    123         raise TypeError("result type can only be np.float32 or np.float64")
    124
--> 125     df = betweenness_centrality_wrapper.betweenness_centrality(
    126         G, normalized, endpoints, weight, vertices, result_dtype
    127     )
cugraph/centrality/betweenness_centrality_wrapper.pyx in cugraph.centrality.betweenness_centrality_wrapper.betweenness_centrality()
cugraph/centrality/betweenness_centrality_wrapper.pyx in cugraph.centrality.betweenness_centrality_wrapper.sg_betweenness_centrality()
cugraph/centrality/betweenness_centrality_wrapper.pyx in cugraph.centrality.betweenness_centrality_wrapper.run_internal_work()
~/development/conda_envs/eco_3_graph/lib/python3.8/site-packages/cudf/core/series.py in __bool__(self)
    834         into a boolean.
    835         """
--> 836         raise TypeError("can't compute boolean for {!r}".format(type(self)))
    837
    838     def values_to_string(self, nrows=None):
TypeError: can't compute boolean for <class 'cudf.core.series.Series'>



- Create a graph with edge weights
- call betweenness_centrality

afender:
BC is BFS based and does not support weights as explained in CuGraph documentation.

Support for weight would require enabling BC counters in SSSP.  In the meantime, BC should not be crashing. Weights stored in the graph could be ignored and weights passed as parameter could throw a clean error.

hlinsen:
Hi @geoHeil, I haven't been able to reproduce the failure on our test datasets. Could you provide additional information about the failure? Some code or a dataset resulting in the crash would be useful. Thanks.

PR #1290, opened 2020-12-01
[REVIEW] BUG Update weights check in bc and graph prims wrappers

hlinsen:
This PR fixes the check for weights in betweenness_centrality and edge_betweenness_centrality wrappers.
Addresses issue https://github.com/rapidsai/cugraph/issues/1282.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@
 - PR #1270 Raise exception for p2p, disable bottom up approach for bfs
 - PR #1275 Force local artifact conda install
 - PR #1285 Move codecov upload to gpu build script
+- PR #1290 Update weights check in bc and graph prims wrappers
 
 
 # cuGraph 0.16.0 (21 Oct 2020)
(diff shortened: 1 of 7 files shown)


GPUtester:

afender (changes requested):
Thanks, @hlinsen. Could you elaborate on this fix in the context of my earlier comment: 
https://github.com/rapidsai/cugraph/issues/1282#issuecomment-734949024

In particular, I think we should throw an error when weights are passed since this is not supported.

BradReesWork (approved):

BradReesWork (changes requested):
there should be an else condition after the weight checks

hlinsen:
> Thanks, @hlinsen. Could you elaborate on this fix in the context of my earlier comment:
> [#1282 (comment)](https://github.com/rapidsai/cugraph/issues/1282#issuecomment-734949024)
> 
> In particular, I think we should throw an error when weights are passed since this is not supported.

The weights are currently throwing an error here: https://github.com/rapidsai/cugraph/blob/branch-0.17/python/cugraph/centrality/betweenness_centrality.py#L237. The ones inside the wrapper are used to construct the graph but are ignored during computation.

BradReesWork (approved):

afender (approved):
Thanks for adding the test and clarifying @hlinsen :+1:

hlinsen:
rerun tests

BradReesWork:
rerun tests

hlinsen:
rerun tests

hlinsen:
rerun tests

BradReesWork:
rerun tests

hlinsen:
rerun tests

PR #1308, opened 2020-12-10
[REVIEW] Updates to support nightly MG test automation

rlratzel:
This PR includes various updates to support nightly automated MG test runs, including:
* Adding a marker which nightly scripts use to run on all visible GPUs instead of a hardcoded number of GPUs, since the scripts rely on knowing the number of GPUs being used in the tests by setting the `CUDA_VISIBLE_DEVICES` env var.
  * In the nightly scripts, the marker is used like so: `pytest -m "not preset_gpu_count" ...`
* Added a `client.wait_for_workers()` call to various setups to both match the approach taken by the `MGContext` class, and to ensure workers are running. This seemed to increase reliability in the test runs. 
  * _side note: we should decide to use only the `MGContext` class or the `client_connection` pytest fixture in these tests, since they both aim to accomplish the same thing._

### DIFF
--- a/python/cugraph/tests/dask/test_mg_bfs.py
+++ b/python/cugraph/tests/dask/test_mg_bfs.py
@@ -27,6 +27,7 @@ from cugraph.dask.common.mg_utils import is_single_gpu
 def client_connection():
     cluster = LocalCUDACluster()
     client = Client(cluster)
+    client.wait_for_workers(None)  # number of devices None = all vsble devices
     Comms.initialize(p2p=True)
 
     yield client
--- a/python/cugraph/tests/dask/test_mg_bfs.py
+++ b/python/cugraph/tests/dask/test_mg_bfs.py
@@ -18,20 +18,15 @@ import cugraph
 import dask_cudf
 import cudf
 from cugraph.dask.common.mg_utils import (is_single_gpu,
-                                          setup_local_dask_cluster)
+                                          setup_local_dask_cluster,
+                                          teardown_local_dask_cluster)
 
 
 @pytest.fixture(scope="module")
 def client_connection():
-    # setup
-    (comms, client, cluster) = setup_local_dask_cluster(p2p=True)
-
+    (cluster, client) = setup_local_dask_cluster(p2p=True)
     yield client
-
-    # teardown
-    comms.destroy()
-    client.close()
-    cluster.close()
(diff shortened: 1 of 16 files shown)


python/cugraph/tests/dask/test_mg_bfs.py
```diff
@@ -27,6 +27,7 @@
 def client_connection():
     cluster = LocalCUDACluster()
     client = Client(cluster)
+    client.wait_for_workers(None)  # number of devices None = all vsble devices
```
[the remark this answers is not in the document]
rlratzel: Thank you, yes, I looked at the dask code and I think you're right.  I think I misinterpreted some example code elsewhere.  I'll fix this in the next commit to this PR.

rlratzel:
rerun tests

python/cugraph/tests/dask/test_mg_bfs.py
```diff
@@ -12,26 +12,24 @@
...
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def client_connection():
-    cluster = LocalCUDACluster()
-    client = Client(cluster)
-    Comms.initialize(p2p=True)
+    # setup
+    (comms, client, cluster) = setup_local_dask_cluster(p2p=True)
```
Iroy30, line 27: comms need not be returned. It can directly be accessed to call destroy

python/cugraph/tests/dask/test_mg_bfs.py
```diff
@@ -12,26 +12,24 @@
...
 
     yield client
 
-    Comms.destroy()
+    # teardown
```
Iroy30, line 31: Maybe a teardown function similar to setup above could be added

python/cugraph/tests/dask/test_mg_bfs.py
rlratzel: I like your suggestion of adding a separate teardown utility (below), so I'll do that instead.

python/cugraph/tests/dask/test_mg_bfs.py
rlratzel: Thanks, I'll add this in the next commit.

Iroy30 (approved):

PR #1312, opened 2020-12-11
Fix EXPERIMENTAL_LOUVAIN_TEST on Pascal

ChuckHastings:
`experimental::louvain` uses a `coco::static_map` which is not supported on Pascal and earlier GPUs.

The first attempt at implementing checks for Pascal attempted to exclusively use preprocessor directives to control building the executable and deciding whether to run it or not.  However, nvcc uses the preprocessor to create different files to be compiled for host, and each combination of devices being compiled.  Exclusively using preprocessor directives created a library that didn't have all of the host and device code necessary to run.

This PR adds a runtime check to raise an exception if running on Pascal, but it will still compile all of the necessary host and device code for other architectures.

### DIFF
--- a/cpp/src/experimental/louvain.cuh
+++ b/cpp/src/experimental/louvain.cuh
@@ -32,6 +32,17 @@
 #include <patterns/transform_reduce_e.cuh>
 #include <patterns/transform_reduce_v.cuh>
 
+// "FIXME": remove the guards below and references to CUCO_STATIC_MAP_DEFINED
+//
+// cuco/static_map.cuh depends on features not supported on or before Pascal.
+//
+// If we build for sm_60 or before, the inclusion of cuco/static_map.cuh wil
+// result in compilation errors.
+//
+// If we're Pascal or before we do nothing here and will suppress including
+// some code below.  If we are later than Pascal we define CUCO_STATIC_MAP_DEFINED
+// which will result in the full implementation being pulled in.
+//
 #if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
 #else
 #define CUCO_STATIC_MAP_DEFINED
--- a/cpp/src/experimental/louvain.cuh
+++ b/cpp/src/experimental/louvain.cuh
@@ -21,7 +21,6 @@
 
 #include <rmm/thrust_rmm_allocator.h>
 #include <compute_partition.cuh>
-#include <cuco/static_map.cuh>
 #include <experimental/shuffle.cuh>
 #include <utilities/comm_utils.cuh>
 #include <utilities/graph_utils.cuh>
@@ -32,6 +31,13 @@
 #include <patterns/copy_v_transform_reduce_in_out_nbr.cuh>
 #include <patterns/transform_reduce_e.cuh>
 #include <patterns/transform_reduce_v.cuh>
+
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
+#else
+#define CUCO_STATIC_MAP_DEFINED
+#include <cuco/static_map.cuh>
+#endif
+
 
 //#define TIMING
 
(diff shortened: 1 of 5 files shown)


cpp/src/experimental/louvain.cuh
```diff
@@ -33,6 +32,12 @@
 #include <patterns/transform_reduce_e.cuh>
 #include <patterns/transform_reduce_v.cuh>
 
+#if defined(__CUDA_ARCH__) && __CUDA_ARCH__ < 700
```
rlratzel, line 35: I'm wondering if this empty `if` block deserves a comment. Alternatively, can you instead do the conditional include in the `if` block by checking for `__CUDA_ARCH__` >= 700 ? I'm also wondering what the expected behavior should be if `__CUDA_ARCH__` is not defined and what that means - should we still include static_map (which is what will happen here)?   (FWIW, I believe if `__CUDA_ARCH__` is not defined the preprocessor treats it like 0 in the expression, based on [this](https://gcc.gnu.org/onlinedocs/gcc-3.0.1/cpp_4.html#SEC39))

note: I just noticed the other conditional preprocessor directives for detecting Pascal are using the same expression, so maybe you're just trying to use the same line everywhere even if you want the negative, hence the empty if block?  That makes sense, I have no strong preference, but I'm still wondering what it means to not have `__CUDA_ARCH__` defined and what the behavior should be.

cpp/tests/experimental/louvain_test.cu
```diff
@@ -77,8 +84,20 @@ class Tests_Louvain : public ::testing::TestWithParam<Louvain_Usecase> {
...
+    //
+    // Disable louvain(experimental::graph_view_t,...)
+    // versions for GPU architectures < 700
+    // (cuco/static_map.cuh depends on features not supported on or before Pascal)
+    //
+    cudaDeviceProp device_prop;
+    CUDA_CHECK(cudaGetDeviceProperties(&device_prop, 0));
+
+    if (device_prop.major < 7) {
+      EXPECT_THROW(louvain(graph_view), cugraph::logic_error);
```
rlratzel, line 97: Sorry if I missed it elsewhere, but are we filtering this test out for Pascal systems? (since we previously had it commented out of our cmake file)  If not, will this result in a CI failure on a Pascal system?

ChuckHastings: No.  The test runs on all systems.  The `device_prop.major < 7` test identifies if the system is running Pascal or earlier.  If so then the call will throw an exception so the `EXPECT_THROW` will make this test pass.  If the system is running Volta or later then we take the else block that follows and we run the test without the `EXPECT_THROW`.

cpp/src/experimental/louvain.cuh
ChuckHastings: I should add the FIXME comment block I use elsewhere and enhance it for the description here.

cpp/tests/experimental/louvain_test.cu
rlratzel: Oh I see, thanks for the explanation. I was initially thinking we'd need to just skip the Pascal run (by upgrading to googletest 1.10.0 and calling `GTEST_SKIP()`), but this is better because it tests both code paths (supported platforms run louvain, unsupported throw an exception) properly.

cpp/src/experimental/louvain.cuh
rlratzel: Sounds good, thanks.  I'll mark this as resolved pending the enhanced FIXME to unblock the PR.

rlratzel (approved):

ajschmidt8 (approved):

ChuckHastings:
rerun tests

BradReesWork (approved):

PR #1318, opened 2020-12-23
Create labeler.yml

jolorunyomi:
# Summary
This PR adds a GitHub action (PR Labeler) to auto label PRs based on a labeling rule and changes in PR.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -5,6 +5,14 @@
 python:
   - python/**
   - notebooks/**
+  - benchmarks/**
+  
+doc:
+  - docs/**
+  - **/*.md
+  
+datasets:
+  - datasets/**
   
 cuGraph:
   - cpp/**
@@ -13,12 +21,9 @@ CMake:
   - **/CMakeLists.txt
   - **/cmake/**
   
-Java:
-  - java/**
-  
 Ops:
   - .github/**
-  - /ci/**
+  - ci/**
   - conda/**
   - **/Dockerfile
   - **/.dockerignore
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -3,23 +3,23 @@
 # Labels culled from https://github.com/rapidsai/cugraph/labels
 
 python:
-  - python/
-  - notebooks/
+  - python/**
+  - notebooks/**
   
 cuGraph:
-  - cpp/
+  - cpp/**
 
 CMake:
   - **/CMakeLists.txt
-  - **/cmake/
+  - **/cmake/**
   
 Java:
-  - java/
+  - java/**
   
 Ops:
-  - .github/
-  - /ci/
-  - conda/
+  - .github/**
+  - /ci/**
+  - conda/**
   - **/Dockerfile
   - **/.dockerignore
-  - docker/
+  - docker/**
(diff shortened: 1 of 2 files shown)


ajschmidt8 (changes requested):
Please see the similar comment on the cuDF PR below about updating the file/path patterns.

https://github.com/rapidsai/cudf/pull/7044#discussion_r550248866

ajschmidt8 (approved):

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+  
+cuGraph:
+  - cpp/**
+
+CMake:
+  - **/CMakeLists.txt
+  - **/cmake/**
+  
+Java:
+  - java/**
```
afender, line 17: It should be fine, but just to confirm we do not have a `java` folder yet.

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/cudf/blob/main/.github/CODEOWNERS
+# Labels culled from https://github.com/rapidsai/cugraph/labels
+
+python:
+  - python/**
+  - notebooks/**
```
afender, line 7: @rlratzel should the `benchmark` directory be there?

afender (commented):
I noticed there's no labels for :
- `docs`
- `datasets`
- `md` files

This would be something to consider adding to make sure labels are always auto-created.

.github/labeler.yml
rlratzel: yes, the benchmark dir is python code we run nightly and as part of CI (in test-only mode) so we'd want that added as well.

.github/labeler.yml
```diff
@@ -0,0 +1,25 @@
...
+CMake:
+  - **/CMakeLists.txt
+  - **/cmake/**
+  
+Java:
+  - java/**
+  
+Ops:
+  - .github/**
+  - /ci/**
```
rlratzel, line 21: I'm assuming this is boilerplate we don't need to review, but the leading `/` here for the `ci` folder looked suspicious.

afender:
tangential topic: do we have doc owners for review (ie. owners for `md` and `rst` files)? I haven't seen that but we probably should as doc is expected to get upgraded a lot in the near future.

.github/labeler.yml
```diff
@@ -0,0 +1,30 @@
+# https://github.com/actions/labeler#common-examples
+# Adapted from https://github.com/rapidsai/cudf/blob/main/.github/CODEOWNERS
```
ajschmidt8, line 2:
```suggestion
# Adapted from https://github.com/rapidsai/cugraph/blob/main/.github/CODEOWNERS
```

should this be `cugraph` instead of `cudf`?

ajschmidt8 (changes requested):
one minor change requested. also, can you include a description on the PR?

rlratzel (approved):
Approving assuming AJ's question gets resolved and a description of the PR is added.

.github/labeler.yml
jolorunyomi: Cugraph doesn't have a code owners file. So it's actually adapted from the code owners in `cudf`.

ajschmidt8 (approved):

PR #1321, opened 2021-01-05
[REVIEW] FIX Set bash trap after PATH is updated [skip-ci]

dillon-cullinan:
Benchmark builds were failing after recent CI updates with the following error:

```
ci/benchmark/build.sh: line 25: gpuci_logger: command not found
```

This fix moves the `PATH` variable update above the trap setup, which should provide access to our conda installed gpuCI tools, including `gpuci_logger`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ## Improvements
 
 ## Bug Fixes
+- PR #1321 Fix benchmark script trap setup to come after the PATH variable update
 
 # cuGraph 0.17.0 (10 Dec 2020)
 ## New Features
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ## Improvements
 
 ## Bug Fixes
+- PR #1321 Fix benchmark script trap setup to come after the PATH variable update
```
ajschmidt8, line 9: changelog entries are no longer necessary. they're handled automatically

ajschmidt8 (approved):

PR #1333, opened 2021-01-14
Expose dense implementation of Hungarian algorithm

ChuckHastings:
Expose an API for the dense implementation of the Hungarian algorithm.

The implementation is actually done in dense form, the sparse version (previously exposed) converts to a dense matrix before calling.

This change exposes the dense version directly to python, allowing users with dense matrices to send their data directly to the algorithm without converting it to a sparse representation first.

Also include some documentation updates.
Also modified the python return value to include both the cost and the assignment.

### DIFF
--- a/python/cugraph/linear_assignment/lap.pxd
+++ b/python/cugraph/linear_assignment/lap.pxd
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2020-2021, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/python/cugraph/linear_assignment/lap.pxd
+++ b/python/cugraph/linear_assignment/lap.pxd
@@ -20,19 +20,19 @@ from cugraph.structure.graph_primtypes cimport *
 
 cdef extern from "algorithms.hpp" namespace "cugraph":
 
-    cdef WT hungarian[VT,ET,WT](
+    cdef weight_t hungarian[vertex_t,edge_t,weight_t](
         const handle_t &handle,
-        const GraphCOOView[VT,ET,WT] &graph,
-        VT num_workers,
-        const VT *workers,
-        VT *assignment) except +
+        const GraphCOOView[vertex_t,edge_t,weight_t] &graph,
+        vertex_t num_workers,
+        const vertex_t *workers,
+        vertex_t *assignment) except +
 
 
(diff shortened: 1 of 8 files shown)


BradReesWork (approved):

python/cugraph/linear_assignment/lap.py
```diff
@@ -69,9 +71,46 @@ def hungarian(G, workers):
...
+    Execute the Hungarian algorithm against a dense bipartite
+    graph representation.
+
+    The Hungarian algorithm identifies the lowest cost matching of vertices
+    such that all workers that can be assigned work are assigned exactly
+    on job.
+
+    Parameters
+    ----------
+    costs : cudf.Series
```
afender, line 93: Do we expect users to store dense matrices in cudf Series? Other alternatives would be cupy I suppose.
Ideally, we would need to figure out the dense class/type for cugraph before having too many people using this API. I think we should at least add a disclaimer that this API is not stable yet.

afender (changes requested):
Looks good, but I think we need more thoughts on how we expose and convert to/from dense format independently of this PR.

python/cugraph/linear_assignment/lap.py
ChuckHastings: I think really that we need to follow the cuML model.  They have a python class that wraps ALL things that can look like an array (device or host side) and facilitates using those across to the C++ side.

My hope was to defer that complexity to when @rlratzel is ready for that but to still get out something that could be used.

I'd be happy to add to the documentation that this is an unstable API

afender (approved):
Thanks!

python/cugraph/linear_assignment/lap.py
rlratzel: The cuML approach sounds good and I agree there's value in getting this out sooner so it can be used while a cuML-style abstract array object can be developed. I think the disclaimer is good here as well.

rlratzel (changes requested):
Minor change request for moving to the new naming convention for template args.

Also, should a test be added for `dense_hungarian()` or did I not see it?

python/cugraph/linear_assignment/lap.pxd
```diff
@@ -20,9 +20,19 @@ from cugraph.structure.graph_primtypes cimport *
 
 cdef extern from "algorithms.hpp" namespace "cugraph":
 
-    cdef void hungarian[VT,ET,WT](
+    cdef WT hungarian[VT,ET,WT](
```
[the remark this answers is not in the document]
ChuckHastings: Done

ChuckHastings:
Added a unit test

rlratzel (approved):

PR #1337, opened 2021-01-15
Using RAPIDS_DATASET_ROOT_DIR env var in place of absolute path to datasets in tests

jnke2016:
Using RAPIDS_DATASET_ROOT_DIR env var in place of absolute path to datasets in tests
https://docs.rapids.ai/maintainers/datasets/ 
Also added a fixture for building both NetworkX and Cugraph graphs


python/cugraph/tests/test_force_atlas2.py
```diff
@@ -120,7 +123,7 @@ def test_force_atlas2(graph_file, score, max_iter,
         iterations on a given graph.
     """
 
-    matrix_file = graph_file[:-4] + ".mtx"
+    matrix_file = PurePath(graph_file).with_suffix(".mtx")
```
rlratzel, line 126: `graph_file` should already be a `PurePath` object, so no need to construct a new one, in other words, you should be able to just do:

`matrix_file = graph_file.with_suffix(".mtx")`

python/cugraph/tests/utils.py
```diff
@@ -40,46 +40,61 @@
...
+        "karate.csv",
+        "dolphins.csv",
+        "polbooks.csv"]
 ]
 
+
 MATRIX_INPUT_TYPES = [
     pytest.param(
-        cp_coo_matrix, marks=pytest.mark.matrix_types, id="CuPy.coo_matrix"
+        cp_coo_matrix, marks=pytest.mark.cupy_types, id="CuPy.coo_matrix"
```
rlratzel, line 91: looks like this may have been updated accidentally somehow?

All `cupy_types` occurrences should be `matrix_types`

python/cugraph/tests/utils.py
```diff
@@ -40,46 +40,61 @@
...
+    PurePath(RAPIDS_DATASET_ROOT_DIR)/f for f in [
+        "dolphins.csv",
+        "netscience.csv",
+        "email-Eu-core.csv"]
 ]
 
+
 DATASETS_KTRUSS = [
-    ("../datasets/polbooks.csv", "../datasets/ref/ktruss/polbooks.csv")
+    (str(PurePath(RAPIDS_DATASET_ROOT_DIR)/"polbooks.csv"),
```
rlratzel, line 76: Rather than convert these to strings here, I'd do the string conversion right before you need to pass it to the function that requires it to be a string, just so things stay consistent as long as possible.  Since this is only used in `test_k_truss_subgraph.py`, and I think Nx is requiring the string, I'd just change the call to be `nx.read_edgelist(str(graph_file), ...)` in `test_k_truss_subgraph.py`

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -55,7 +55,7 @@
 # Comparison functions
 # =============================================================================
 def calc_betweenness_centrality(
-    graph_file,
+    Fixture_params_dts,
```
rlratzel, line 58: This seems like it could be a little confusing since pytest won't call this as a fixture like it would for a test. Maybe this could be named something like `graph_obj_tuple` or something that describes it's the tuple of graph objects passed in?

python/cugraph/tests/test_betweenness_centrality.py
```diff
@@ -299,80 +298,96 @@ def prepare_test():
...
+    (WEIGHTED_GRAPH_OPTIONS, "wgtd_gph_opts"))
+
+fixture_params_dts_urnbrd = utils.genFixtureParamsProduct(
+    (DATASETS_UNRENUMBERED, "grph"),
+    (DIRECTED, "dirctd"),
+    (WEIGHTED_GRAPH_OPTIONS, "wgtd_gph_opts"))
+
+
+@pytest.fixture(scope="module", params=fixture_params_dts_sml)
+def Fixture_params_dts_sml(request):
```
rlratzel, line 322: It might not be obvious that there's a variable for the fixture params and the fixture itself named almost the same.  I've noticed that other fixtures that come with pytest and elsewhere don't use a `Fixture_` prefix, so maybe change this to just `dts_small`, or maybe `dts_small_fixture` to differentiate from the variable of almost the same name?

Also, I found some of the abbreviations to not be as readable as just spelling out the words, like `datasets_small_fixture` or something similar.

rlratzel (changes requested):
Looks good, just some minor change requests.

python/cugraph/tests/test_force_atlas2.py
jnke2016: right, there is no need to cast graph_file since it is already a `PurePath`

python/cugraph/tests/utils.py
jnke2016: That was definitely a mistake when I was copying code back and forth from my machine to the docker container. I probably copied and old version

python/cugraph/tests/utils.py
jnke2016: you are right. since `nx.read_edgelist `doesn't support `PurePath` it is better to cast its input instead of altering the consistency of the `utils` file.

python/cugraph/tests/test_betweenness_centrality.py
jnke2016: I renamed it `graph_obj_tuple`

python/cugraph/tests/test_betweenness_centrality.py
jnke2016: I renamed what was passed in `params` `small_graph_fixture_params` to stay consistent with the naming from other tests like `test_bfs`.

rlratzel (approved):

rlratzel:
rerun tests

rlratzel:
closes #1345

Iroy30 (approved):

BradReesWork (approved):

PR #1355, opened 2021-01-22
Implement MNMG Renumber

aschaffer:
This exposes to cython / python newly added C++ functions `sort_and_shuffle_values()` and `renumber_edgelist()` in order to perform MNMG renumbering.


cpp/include/utilities/cython.hpp
```diff
@@ -246,6 +301,29 @@ void call_sssp(raft::handle_t const& handle,
                vertex_t* predecessors,
                const vertex_t source_vertex);
 
+// wrapper for shuffling:
+//
+template <typename vertex_t, typename edge_t, typename weight_t>
+major_minor_weights_t<vertex_t, weight_t> call_shuffle(
+  raft::handle_t const& handle,
+  vertex_t* edgelist_major_vertices /* [IN] */,  // make_zip_iterator does not accept const !
```
ChuckHastings, line 309: I have used make_zip_iterator with const.  Following function with simple thrust device vectors (adapted from the make_zip_iterator documentation) compiles and generates correct results for me.

```
void my_copy(int const *d_int_in,
              float const *d_float_in,
              int *d_int_out,
              float *d_float_out,
              int len) {
  thrust::copy(thrust::device,
               thrust::make_zip_iterator(thrust::make_tuple(d_int_in, d_float_in)),
               thrust::make_zip_iterator(thrust::make_tuple(d_int_in, d_float_in)) + len,
               thrust::make_zip_iterator(thrust::make_tuple(d_int_out, d_float_out)));
}
```

cpp/src/utilities/cython.cu
```diff
@@ -865,5 +959,24 @@ template void call_sssp(raft::handle_t const& handle,
                         int64_t* predecessors,
                         const int64_t source_vertex);
 
+// TODO: add the remaining relevant EIDIr's:
+//
+template major_minor_weights_t<int32_t, float> call_shuffle(raft::handle_t const& handle,
+                                                            int32_t* edgelist_major_vertices,
+                                                            int32_t* edgelist_minor_vertices,
+                                                            float* edegelist_weights,
```
ChuckHastings, line 967: misspelling here (immaterial, but it made me look)

python/cugraph/structure/graph_primtypes.pxd
```diff
@@ -142,6 +142,62 @@ cdef extern from "functions.hpp" namespace "cugraph":
...
+
+
+# 4. `sort_and_shuffle_values()` wrapper:
+#
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
+
+    cdef major_minor_weights_t[vertex_t, weight_t] call_shuffle[vertex_t, edge_t, weight_t](
+        const handle_t &handle,
+        vertex_t *edgelist_major_vertices,
+        vertex_t *edgelist_minot_vertices,
```
ChuckHastings, line 181: Spelling error here (minot should be minor)

python/cugraph/structure/graph_primtypes.pxd
```diff
@@ -142,6 +142,62 @@ cdef extern from "functions.hpp" namespace "cugraph":
...
+        weight_t* edegelist_weights,
```
ChuckHastings, line 182: `edege` should be `edge`

cpp/include/utilities/cython.hpp
```diff
@@ -246,6 +301,29 @@ void call_sssp(raft::handle_t const& handle,
...
+  vertex_t* edgelist_minor_vertices /* [IN] */,
+  weight_t* edegelist_weights,
```
ChuckHastings, line 311: spelling error here `edege` should be `edge`

python/cugraph/structure/graph_primtypes.pxd
```diff
@@ -142,6 +142,62 @@ cdef extern from "functions.hpp" namespace "cugraph":
...
+
+
+# 5. `renumber_edgelist()` wrapper
+#
+cdef extern from "utilities/cython.hpp" namespace "cugraph::cython":
+
+    cdef renum_quad_t[vertex_t, edge_t] call_renumber[vertex_t, edge_t](
+        const handle_t &handle,
+        vertex_t *edgelist_major_vertices,
+        vertex_t *edgelist_minot_vertices,
```
ChuckHastings, line 194: Spelling error here `minot` should be `minor`

cpp/src/utilities/cython.cu
```diff
@@ -686,6 +690,96 @@ void call_sssp(raft::handle_t const& handle,
   }
 }
 
+// wrapper for shuffling:
+//
+template <typename vertex_t, typename edge_t, typename weight_t>
+major_minor_weights_t<vertex_t, weight_t> call_shuffle(
+  raft::handle_t const& handle,
+  vertex_t* edgelist_major_vertices /* [IN] */,  // make_zip_iterator does not accept const !
```
ChuckHastings, line 698: See other comment about zip_iterator and const

ChuckHastings (changes requested):
Some questions/minor edit suggestions

cpp/include/experimental/graph.hpp
```diff
@@ -70,7 +70,11 @@ class graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enab
           bool sorted_by_global_degree_within_vertex_partition,
           bool do_expensive_check = false);
 
-  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> view()
+  // _needs_ to be `const`:
+  // because visitors act on `const` graph objects
+  // and need to extract the view():
+  //
```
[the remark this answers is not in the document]
seunghwak: I agree with @ChuckHastings. This should be const from the beginning, and this fixes the obvious mistake I made :-) Just add const and I think no comment is necessary.

cpp/include/utilities/cython.hpp
```diff
@@ -109,6 +110,60 @@ struct graph_container_t {
...
+//
+template <typename vertex_t, typename weight_t>
+struct major_minor_weights_t {
+  explicit major_minor_weights_t(raft::handle_t const& handle)
+    : shuffled_major_vertices_(0, handle.get_stream()),
+      shuffled_minor_vertices_(0, handle.get_stream()),
+      shuffled_weights_(0, handle.get_stream())
+  {
+  }
+  rmm::device_uvector<vertex_t>& get_major(void) { return shuffled_major_vertices_; }
```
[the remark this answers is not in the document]
aschaffer: No, because they're used by `std::tie()` to assign the fields directly.

cpp/include/utilities/cython.hpp
aschaffer: Well, I'm getting compiler errors with const pointers.

ChuckHastings: Perhaps there's something else that's the root cause of this.  It feels like we should tease that out, passing in const pointers here seems like the right thing to do.

cpp/include/utilities/cython.hpp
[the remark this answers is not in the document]
ChuckHastings: That feels awkward and non-intuitive.  That might be worth either reconsidering (although I'm not entirely sure there's a more natural idiom for what is required) or documenting explicitly so that it doesn't surprise someone maintaining the code later.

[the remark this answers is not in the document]
aschaffer: I'm not sure what exactly "feels awkward" for you. We have a non-const getter that allows direct assignment of a field, thereby avoiding copies. If that struct was an aggregate the fields would have been assigned directly. 

I don't know what "feels intuitive" either. These qualifications seem mighty subjective to me. Can you recommend a concrete course of action, here?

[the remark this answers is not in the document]
aschaffer: FYI, see https://github.com/aschaffer/cugraph/blob/47f7aff07aca087e6f6f6fedc7b4bd78c177aedb/cpp/src/experimental/coarsen_graph.cu#L337 .

It's the same thing here, except the fields are private, hence they need accessors by reference (the reasons that they're private: (1) better encapsulation; (2) this struct is exposed to cython, which theoretically sees the entire public interface, but these fields cannot be exposed to cython; hence they're private)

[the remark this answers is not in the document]
ChuckHastings: Clearly subjective...

I guess I'm accustomed to seeing a struct aggregate and directly assigning the fields when you're going to be assigning the fields in this way.  When I see accessors returning the fields then I expecting an RAII style object or something that more resembles an RAII style object.

Perhaps this is a common idiom that I just haven't seen much of.  I just worry if it's not a common idiom that it will be confusing to people just glancing at it.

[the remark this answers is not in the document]
seunghwak: std::tie(a, b) = some_func();

a & b needs to be lvalue references, but some_func() does not need to return a reference type. sort_and_shuffle_values in the above link return rvalue (lvalue does not work as copy constructor is deleted in rmm::device_uvector).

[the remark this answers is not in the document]
aschaffer: @seunghwak , if you have a sfunction, similar to `sort_and_shuffle()`:
```
template<typename T>
auto fv1(void)
{
  std::vector<T> v{6, 7, 8, 9};
  return std::move(v);
}
```
(or, returning `decltype(auto)`, which in general is more recommended), then the return type is `vector<T>`:
```
//... stuff...
  auto ret_f1 = fv1<double>();

  static_assert(std::is_same<decltype(ret_f1), std::vector<double> >::value,"Error: type mismatch...");
```
The above static assertion passes. 

And in cases like that `std::tie()` works as expected ("ties" the lvalue reference to the return). Example:
```
template <typename T, typename U>
std::tuple<std::vector<T>, U, T> foo(void)
{
  std::vector<T> v{1, 2, 3, 4, 5};
  U u{13};
  T t{17};
  return std::make_tuple(v, u, t);
}

template <typename T, typename U>
struct X {
  X(void) = default;

  std::vector<T>& get_v(void) { return v_; }
  U& get_u(void) { return u_; }
  T& get_t(void) { return t_; }

 private:
  std::vector<T> v_;
  U u_;
  T t_;
};

int main(void)
{
  X<int, double> x{};
  std::tie(x.get_v(), x.get_u(), x.get_t()) = foo<int, double>(); // assigns return of foo() to underlying X fields
}
```

So, I'm not sure what is the issue here? And what is the desired behavior?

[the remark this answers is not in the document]
aschaffer: If I change `foo()`'s signature to look more like `sort_and_shuffle()`, the outcome is the same:
```
template <typename T, typename U>
auto foo(void)
{
  std::vector<T> v{1, 2, 3, 4, 5};
  U u{13};
  T t{17};
  return std::make_tuple(std::move(v), u, t);
}
```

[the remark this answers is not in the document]
seunghwak: So, you're using .get_major() in the left side of the statement, something like

major_minor_weights_t obj{};
std::tie(obj.get_major(), obj.get_minor(), obj.get_weights()) = a_func_that_returns_a_tuple();

I thought you're trying to use this in the right side. I don't have a big complaint about this, but related to what @ChuckHastings said, we may consider few other options.

1. If we're using struct to just tie some variables, should the struct member variables be private? Directly exposing a non-const reference to a private member variable sounds non-idiomatic to me.

2. Another option is to add a setter function taking a tuple of majors, minors, weights as an input parameter if you really want to keep the struct member variables as private.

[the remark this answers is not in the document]
seunghwak: And do we have any guidelines on when to use struct vs class? I use struct to just simply tie variables (variables are public in this case) and class if I need to limit direct access to the member variables and truly need some class member functions.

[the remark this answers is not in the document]
aschaffer: `get_major()` etc. are returning refences. For all intent and purpose...they can be used as either getters or setters.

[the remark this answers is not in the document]
aschaffer: BTW, @seunghwak, what you're doing in coarsen_graph.cu:L337 ( https://github.com/aschaffer/cugraph/blob/47f7aff07aca087e6f6f6fedc7b4bd78c177aedb/cpp/src/experimental/coarsen_graph.cu#L337) is
```
    std::forward_as_tuple(
      std::tie(rx_edgelist_major_vertices, rx_edgelist_minor_vertices, rx_edgelist_weights),
      std::ignore) =
      sort_and_shuffle_values(...);
```
Which seems much less "intuitive" than what I do. I'm not sure why that is okay and my code above isn't.

In the end I'd like to resolve this conversation, which with all due respect doesn't seem to go anywhere. To that end, what is the desired change here? If none, can we resolve this?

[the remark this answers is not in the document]
seunghwak: So, the code above, it should ideally be std::tie(std::tie(), ...), but std::tie does not take rvalue reference, so std::forward_as_tuple replaces the outer std::tie (std::forward_as_tuple is nearly identical to std::tie in this context, but it takes a rvalue reference). People who are not familiar with std::forward_as_tuple may find the code above confusing, but it is just using std::tie in a nested way.

My biggest concern about the code above is, allowing users to directly modify the private member variable (by returning a non-const reference). Then, why define those variables as private from the first place?

If we are using "struct major_minor_weights_t" to just tie variables, then we may just do

```
struct major_minor_weights_t {
  rmm::device_uvector<vertex_t> shuffled_major_vertices;  <= public
  rmm::device_uvector<vertex_t> shuffled_minor_vertices;
  rmm::device_uvector<weight_t> shuffled_weights;
};

major_minor_weights_t ob{}j;
std::tie(obj.shuffled_major_vertices, obj.shuffled_minor_vertices, shuffled_weights) = func();
```

instead of defining the member variables as private but also provide getter functions for direct modification.

But this is just my suggestion, and I don't think this is a deal breaker, so I am not against marking this conversation as resolved.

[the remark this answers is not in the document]
aschaffer: Because, as already explained: 
(1) This gets exposed to Cython; Cython "sees" public interfaces; if you make those public we might have (if not right away, possibly in the future) issues with exposing a type (`device_uvector<>`) that Cython does _not_ understand (yet);
(2) data fields in classes should _always_ be private (it's in Scott Meyers idioms, it's in the code review guidelines, it's...good practice);

Also, as already noted you also use a getter to set something (a tuple element); again: 
```
tests/experimental/coarsen_graph_test.cpp:196:            std::get<0>(cur) = cugraph::experimental::invalid_vertex_id<vertex_t>::value;
```
If semantically this practice seems wrong to you...why did you use it yourself?

Resolving this.

python/cugraph/structure/graph_primtypes.pxd
aschaffer: Done.

python/cugraph/structure/graph_primtypes.pxd
aschaffer: Fixed.

cpp/include/utilities/cython.hpp
aschaffer: Fixed.

python/cugraph/structure/graph_primtypes.pxd
aschaffer: Fixed.

cpp/src/utilities/cython.cu
aschaffer: Fixed.

cpp/include/experimental/graph.hpp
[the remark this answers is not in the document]
aschaffer: Done.

cpp/include/experimental/graph.hpp
```diff
@@ -132,7 +136,11 @@ class graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enab
 
   vertex_t get_number_of_local_vertices() const { return this->get_number_of_vertices(); }
 
-  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> view()
+  // _needs_ to be `const`:
+  // because visitors act on `const` graph objects
+  // and need to extract the view():
+  //
```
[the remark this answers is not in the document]
aschaffer: Done.

cpp/include/utilities/cython.hpp
aschaffer: What seems to be happening is that `sort_and_shuffle()` assumes a zip iterator of non-const's. Which explains the compiler error. All other invocations of `sort_and_shuffle()` use zip iterators made of non const iterators / buffers.

Can we resolve this?

cpp/src/utilities/cython.cu
aschaffer: `sort_and_shuffle_values()` assumes a zip iterator of non-const's. Which explains the compiler error. All other invocations of sort_and_shuffle() use zip iterators made of non const iterators / buffers.

Can we resolve this?

cpp/include/utilities/cython.hpp
aschaffer: BTW, the explanation for the non-const pre-condition is `sort_and_shuffle_value()`'s (eventual) call, through `sort_and_count()`, to `thrust::sort()`, which is in-place, hence requiring non-const iterators.

cpp/src/utilities/cython.cu
aschaffer: See above explanation on non-const pre-condition.

cpp/include/utilities/cython.hpp
ChuckHastings: That makes more sense.  Can we either eliminate the comment or change the comment in this code (and the other place where it happens) to reflect that the values are not const because the arrays are modified in-place?

ChuckHastings: Actually, wouldn't this be [IN, OUT] because it's modified in-place?

aschaffer: Done.

cpp/src/utilities/cython.cu
aschaffer: Done.

ChuckHastings (approved):
Thanks for the updates.

aschaffer:
rerun tests.

python/cugraph/dask/structure/renumber_wrapper.pyx
```diff
@@ -88,33 +91,80 @@ def mg_renumber(input_df,
...
+    cdef unique_ptr[renum_quad_t[int, int]] ptr_renum_quad_32_32
+    cdef unique_ptr[renum_quad_t[int, long]] ptr_renum_quad_32_64
+    cdef unique_ptr[renum_quad_t[long, long]] ptr_renum_quad_64_64
+
+    cdef unique_ptr[vector[int]] uniq_partition_vector_32
+    cdef unique_ptr[vector[long]] uniq_partition_vector_64
+
+    cdef size_t rank_indx = <size_t>rank
+    
     if (vertex_t == np.dtype("int32")):
```
ChuckHastings, line 115: Feels like we should raise an exception if any of these 3 ifs fail (at least for now) rather than silently failing to do anything leaving the impression that it worked.

ChuckHastings (changes requested):
One minor concern

python/cugraph/dask/structure/renumber_wrapper.pyx
aschaffer: Yes. In fact I'm in the process of adding the missing instantiations in both `cython.cu` and `renumber_wrapper.pyx`. And I'll add an exception for the `else:` branch

aschaffer: @seunghwak , the above triggered an issue: `renumber_edgelist()` is explicitly instantiated _only_ for `<int32_t, int32_t, true | false>`. As a result, I cannot instantiate renumbering C++2Cy interface (`call_renumber()` in `cython.cu`) for anything else, either.

Was that the intended functionality for `renumber_edgelist()`, or an omission?

ChuckHastings: I added them in one of my other PRs.  I suspect this was an omission.

seunghwak: Yes, they need to be added.

seunghwak (approved):

python/cugraph/dask/structure/renumber_wrapper.pyx
aschaffer: Okay, for the sake of moving along, I'll just add the missing `renumber_edgelist()` instantiations. Going once, going twice,...

afender (approved):
Looks good considering integration and testing will be addressed in a future PR

python/cugraph/dask/structure/renumber_wrapper.pyx
aschaffer: Done: handled missing instantiation paths in both C++ and Cython. And added exception raising on un-handled paths in Cython.

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

Issue #1185, opened 2020-10-03
[FEA] Travelling Salesman Problem (TSP) and/or Hamiltonian path


jannisborn:
**Is your feature request related to a problem? Please describe.**
Is there any plan to integrate NP algorithms like finding Hamiltonian paths or solving the TSP?

**Describe the solution you'd like**
A function that you can call with an undirected, possibly weighted graph and a starting node that returns the solution to the TSP.

**Additional context**
`networkx` implementation of TSP is pretty slow, so this would be a great feature, I think.

BradReesWork:
@jannisborn TSP is on the roadmap for roughly the January 2021 release.  We have been doing a lot of research in that areas and are now ready to make it available.  

Hamiltonian paths is something that we have discussed but has not made it onto the roadmap for a specific release

jannisborn:
@BradReesWork that sounds fantastic!
for the TSP, I think it would be great if an option would be included to, unlike the original TSP, **_not_** return to the source node. This would be identical to finding the shortest Hamiltonian path for a given source node. 
Otherwise thanks a lot for the great work! Loooking forward to the 2021 release and please feel free to close if applicable.

BradReesWork:
@jannisborn breaking your feature request into two part - TSP will be in the current 0.18 release that we are workign on.   Hamilton paths a little later

BradReesWork:
closed by #1360

PR #1360, opened 2021-01-27
Add SG TSP

hlinsen:
This PR implements an approximated solution to the Traveling Salesperson Problem (TSP).
The algorithm is exposed under ```traversal``` through a Python API taking 2D pos as input and returning a route.
This PR relies on RAFT KNN: https://github.com/rapidsai/raft/pull/126
Solves: https://github.com/rapidsai/cugraph/issues/1185


cpp/CMakeLists.txt
```diff
@@ -294,6 +294,13 @@ else()
     set(NCCL_LIBRARIES ${NCCL_PATH}/lib/libnccl.so)
 endif(NOT NCCL_PATH)
 
+# - FAISS
+add_library(FAISS::FAISS SHARED IMPORTED)
+set_property(TARGET FAISS::FAISS PROPERTY
+	IMPORTED_LOCATION $ENV{CONDA_PREFIX}/lib/libfaiss.so)
```
rlratzel, line 300: I'm worried that we'll need to support users that aren't using conda as well.  Is there another option we can use to find Faiss that don't rely on a `CONDA_PREFIX`, or is this the best way?

cpp/tests/CMakeLists.txt
```diff
@@ -227,6 +227,15 @@ set(FA2_TEST_SRC
 
 ConfigureTest(FA2_TEST "${FA2_TEST_SRC}" "")
 
+###################################################################################################
+# - TSP  tests --------------------------------------------------------------------------
+
+set(TSP_TEST_SRC
+    "${CMAKE_SOURCE_DIR}/../thirdparty/mmio/mmio.c"
```
rlratzel, line 234: We're changing lines like these to use `CMAKE_CURRENT_SOURCE_DIR` instead (which means the relative path will be `"${CMAKE_CURRENT_SOURCE_DIR}/../../thirdparty/mmio/mmio.c"` to support using cugraph as a 3rd party package, as well as to be consistent with other RAPIDS repos.

datasets/get_test_data.sh
```diff
@@ -31,6 +31,12 @@ benchmark
 # ~1s download
 https://s3.us-east-2.amazonaws.com/rapidsai-data/cugraph/benchmark/hibench/hibench_1_small.tgz
 benchmark
+
+# ~0.6s download
```
rlratzel, line 35: Thanks for remembering to update this script!

python/cugraph/layout/force_atlas2_wrapper.pyx
```diff
@@ -94,8 +94,8 @@ def force_atlas2(input_graph,
     # We keep np.float32 as results for both cases
     pos = cuda.device_array(
             (num_verts, 2),
-            order="F",
-            dtype=np.float32)
+            order="C",
+            dtype=np.int32)
```
rlratzel, line 98: does this change the comment 4 lines above this?

python/cugraph/tests/test_traveling_salesman.py
```diff
@@ -0,0 +1,73 @@
...
+import pytest
+
+import cugraph
+from cugraph.tests import utils
+import rmm
+import cudf
+import time
+import numpy as np
+
+TSPLIB_REF = "../datasets/ref/tsp/tsplib.csv"
```
rlratzel, line 25: We've made a change to support decoupling tests from the source tree, which means we're not hardcoding relative paths anymore.  There's an example [here](https://github.com/rapidsai/cugraph/pull/1337/files#diff-d05ca511f68d1d1443c23b813f82c300d4ede3aeb6ef5b8edc02174f149b43bfR33) in PR #1337

python/cugraph/tests/utils.py
```diff
@@ -77,6 +77,12 @@
     PurePath(RAPIDS_DATASET_ROOT_DIR)/"ref/ktruss/polbooks.csv")
 ]
 
+DATASETS_TSPLIB = [
+    ["../datasets/tsplib/a280.tsp", 2579],
+    ["../datasets/tsplib/eil51.tsp", 426],
+    ["../datasets/tsplib/kroA100.tsp", 21282],
+    ["../datasets/tsplib/tsp225.tsp", 3916],
```
rlratzel, line 84: Same here, we're not hardcoding relative paths anymore (see example [here](https://github.com/rapidsai/cugraph/pull/1337/files#diff-d05ca511f68d1d1443c23b813f82c300d4ede3aeb6ef5b8edc02174f149b43bfR33))

python/cugraph/traversal/traveling_salesman.py
```diff
@@ -0,0 +1,74 @@
+# Copyright (c) 2020, NVIDIA CORPORATION.
```
rlratzel, line 1: This and all others in this PR will need a 2021 date.

rlratzel (changes requested):
I didn't review the TSP details as much, but did notice a few things that'll need to be addressed.

cpp/CMakeLists.txt
hlinsen: Yes actually I added an option to build faiss statically on the [RAFT PR](https://github.com/rapidsai/raft/pull/126). I'm planning to do the same for cugraph.
https://github.com/hlinsen/raft/blob/tsp-knn/cpp/CMakeLists.txt#L180-L190.

python/cugraph/layout/force_atlas2_wrapper.pyx
hlinsen: Will revert this change, forgot to clean it

ajschmidt8 (approved):

cpp/CMakeLists.txt
hlinsen: Updated this here: https://github.com/rapidsai/cugraph/pull/1360/commits/a3bb5ad769e3ca3342afd8f469e5574bbb951b2b

python/cugraph/traversal/traveling_salesman.py
hlinsen: Should all be updated now

python/cugraph/tests/utils.py
hlinsen: Fixed here: https://github.com/rapidsai/cugraph/pull/1360/commits/923bd9ef94bf6fb9dee55f344379a5485c68a830

python/cugraph/tests/test_traveling_salesman.py
hlinsen: Removed since the ref score is now in the dataset tuple: https://github.com/rapidsai/cugraph/pull/1360/commits/923bd9ef94bf6fb9dee55f344379a5485c68a830

cpp/tests/CMakeLists.txt
hlinsen: Changed it here: https://github.com/rapidsai/cugraph/pull/1360/commits/e099b5b9479ed33525710bd27f9952f7c4b1febd

BradReesWork (approved):

python/cugraph/tests/utils.py
```diff
@@ -204,7 +211,7 @@ def create_obj_from_csv(
         )
 
     else:
-        raise TypeError(f"unsupported type: {obj_type}")
+        raise TypeError("unsupported type: {obj_type}")
```
rlratzel, line 214: doesn't this still need to be an `f` string to properly expand `{obj_type}`?

BradReesWork:
rerun tests

cpp/CMakeLists.txt
```diff
@@ -294,6 +311,48 @@ else()
     set(NCCL_LIBRARIES ${NCCL_PATH}/lib/libnccl.so)
 endif(NOT NCCL_PATH)
 
+# - FAISS
+if(BUILD_STATIC_FAISS)
+  set(FAISS_DIR ${CMAKE_CURRENT_BINARY_DIR}/faiss CACHE STRING
+    "Path to FAISS source directory")
+  ExternalProject_Add(faiss
```
rlratzel, line 318: Sorry this wasn't mentioned earlier, but we're migrating to `FetchContent` instead of `ExternalProject` where possible, as see in [this PR](https://github.com/rapidsai/cugraph/pull/1367/files#diff-1bba462ab050e89360fd88110a689e85ee037749cea091a1848ab574381d3795R243).

If this is too involved for this PR, can you add a FIXME?

rlratzel (changes requested):
Looks good, just a few more items I noticed.

python/cugraph/tests/utils.py
hlinsen: I think you're right. I removed it because it was generating a flake8 error (locally)

cpp/include/algorithms.hpp
```diff
@@ -264,6 +264,43 @@ void force_atlas2(GraphCOOView<vertex_t, edge_t, weight_t> &graph,
...
+ * the better the solution will be approximated. The number of restarts depends on the problem
+ * size and should be kept low for instances above 2k cities.
+ * @param[in] beam_search 										Specify if the initial solution should use KNN
+ * for an approximation solution.
+ * @param[in] k 															Beam width to use in the search.
+ * @param[in] nstart 												  Start from a specific position.
+ * @param[in] verbose 												Logs configuration and iterative improvement.
+ *
+ */
+float traveling_salesman(raft::handle_t &handle,
```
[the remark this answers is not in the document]
hlinsen: Currently to get my outputs I return the minimum cost and the path is passed as a pointer.

cpp/src/traversal/tsp_solver.hpp
```diff
@@ -0,0 +1,351 @@
...
+__global__ void init(int *mylock, int *n_climbs, int *best_tour)
+{
+  *mylock    = 0;
+  *n_climbs  = 0;
+  *best_tour = INT_MAX;
+  best_soln  = nullptr;
+  best_route = nullptr;
+}
+
+__global__ __launch_bounds__(2048, 2) void two_opt_search(int *mylock,
```
aschaffer, line 45: This is a quite large hand-written kernel. Two issues:
(1.) Couldn't be this designed using Thrust/CUB primitives?
(2.) Is there a possibility to split this kernel into smaller `__device__` functions, so that reading and maintaining this code later would become possible / easier ?

cpp/src/traversal/tsp_utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+/* CPU side validation */
+#define cpudist(a, b)                                                                          \
```
aschaffer, line 20: Please consider replacing macro functions by actual inlined (possibly template) functions. Pre-processor code cannot be inspected, debugged, and compiler errors triggered by it are hard to understand.

cpp/src/traversal/tsp_utils.hpp
```diff
@@ -0,0 +1,59 @@
...
+  (sqrtf((pos[a] - pos[b]) * (pos[a] - pos[b]) + (pos[a + nodes_ + 1] - pos[b + nodes_ + 1]) * \
+                                                   (pos[a + nodes_ + 1] - pos[b + nodes_ + 1])))
+
+// Round all distances to the nearest integer, which is good enough for realistic distances in ft
+#define tilesize 128
+#define kswaps 4
+#define dist(a, b) \
```
aschaffer, line 27: Same comment as above: please consider using an inlined (template) function instead of macro.

aschaffer (changes requested):

cpp/tests/traversal/tsp_test.cu
```diff
@@ -0,0 +1,205 @@
...
+    std::string token;
+    std::istringstream tokenStream(s);
+    while (std::getline(tokenStream, token, delimiter)) {
+      if (token.size() == 0) continue;
+      tokens.push_back(token);
+    }
+    return tokens;
+  }
+
+  int load_tsp(const char* fname, Route* input)
```
afender, line 174: Why not use mtx or csv files?

hlinsen:
rerun tests

datasets/tsplib/a280.tsp
```diff
@@ -0,0 +1,287 @@
+NAME : a280
+COMMENT : drilling problem (Ludwig)
+TYPE : TSP
+DIMENSION: 280
+EDGE_WEIGHT_TYPE : EUC_2D
+NODE_COORD_SECTION
+1 288 149
```
afender, line 7: Seems like all these files are really graph/sparse matrices rather than general euclidian space positions. 
Are we supporting only positive integers within the same range? 
If yes this should be mentioned in the API doc. 
If not other test data should be added.

cpp/include/algorithms.hpp
```diff
@@ -192,6 +192,44 @@ void force_atlas2(GraphCOOView<vertex_t, edge_t, weight_t> &graph,
...
+ *
+ *        The current implementation does not support a weighted graph.
+ *
+ * @throws                                    cugraph::logic_error when an error occurs.
+ * @param[in] handle                          Library handle (RAFT). If a communicator is set in the
+ * handle, the multi GPU version will be selected.
+ * @param[in] vtx_ptr                         Device array containing the vertex identifiers used
+ * to initialize the route.
+ * @param[out] route                          Device array containing the returned route.
+ * @param[in] x_pos                           Device array containing starting x-axis positions.
```
afender, line 208: What's the argument ordering logic?

afender (changes requested):
Cool code. Just reviewed C++ part for now.

In addition to the comments, I'd like to see some performance data and basic profile (ie. what's the perf bottleneck). 

Kernels are nice but I would have prefered if used more of  Graph Prims, Thrust and RAFT in this PR. API/Doc needs polishing. I also added some comment regarding integration and testing, for instance I see an API that doesn't take graphs but graphs are passed in tests in a custom format that really looks like regular mtx/csv, so I think this should be reconciled.

hlinsen (commented):
> Cool code. Just reviewed C++ part for now.
> 
> In addition to the comments, I'd like to see some performance data and a basic profile (ie. what's the perf bottleneck).
> 
> Kernels are nice but I would have preferred if there was more of Graph Prims, Thrust, and RAFT in this PR. API/Doc needs polishing. I also added some comments regarding integration and testing, for instance, I see an API that doesn't take graphs but graphs are passed in tests in a custom format that really looks like regular mtx/csv, so I think this should be reconciled.

Thanks for the review. I'll work on adding timers. The current TSP implementation considers the graph is fully connected when looking for the best path. If we used a graph representation this means we would be O(V^2) in memory. We also consider than recomputing the distances is faster than storing them in shared memory and allows us to run on problems with larger size (more than 2k cities). One of the next step is indeed to reconcile with the rest of the API and take a graph as input but this would mean we have to refactor the kernel logic to look only at specific edges and have an efficient look up table (Cuco should have that) to get the weights. These optimizations are on the roadmap for next releases.

cpp/tests/traversal/tsp_test.cu
hlinsen: We want to compare against [TSPLIB](http://comopt.ifi.uni-heidelberg.de/software/TSPLIB95/) which solves TSP as a city problem, so the format corresponds to 2D positions as input.

cpp/src/traversal/tsp.cu
```diff
@@ -0,0 +1,223 @@
...
+
+  raft::copy(route_, route_sol, nodes_, stream_);
+
+  CUDA_TRY(cudaMemcpy(h_x_pos.data(), soln, sizeof(float) * (nodes_ + 1), cudaMemcpyDeviceToHost));
+  cudaDeviceSynchronize();
+  CUDA_TRY(cudaMemcpy(
+    h_y_pos.data(), soln + nodes_ + 1, sizeof(float) * (nodes_ + 1), cudaMemcpyDeviceToHost));
+  cudaDeviceSynchronize();
+
+  for (int i = 0; i < nodes_; ++i) {
```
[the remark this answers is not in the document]
hlinsen: Right now the problem size we target is up to 2k nodes

cpp/src/traversal/tsp_utils.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+namespace cugraph {
+namespace detail {
+
+__host__ __device__ inline float euclidean_dist(float *px, float *py, int a, int b)
+{
+  return sqrtf((px[a] - px[b]) * (px[a] - px[b]) + (py[a] - py[b]) * (py[a] - py[b]));
+}
+
+int best_thread_count(int nodes)
```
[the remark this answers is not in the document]
hlinsen: It's based on based on the number of cities in the input, their shared memory usage and the hardware. I will update what's relying on the hardware.

datasets/tsplib/a280.tsp
hlinsen: We're supporting float data, the tsp225 file contains purely floats. There are no constraints on the range between the cities

cpp/include/algorithms.hpp
hlinsen: I was following a priority ordering on the parameters for the Python API. The cpp layer follows the Python API except for the route that is a returned parameter. I will work on updating the return and use a float and rmm::uvector tuple instead

datasets/tsplib/a280.tsp
afender: Gotcha. Should we test negative coordinates?

cpp/tests/traversal/tsp_test.cu
afender: I see how it would be relevant in the context of a demo/benchmark. Not sure if that's a strong enough motivation in the context cugraph testing though. It seems to me it is going in the opposite direction than what @rlratzel is trying to do for making testing more generic.

cpp/include/algorithms.hpp
```diff
@@ -192,6 +192,44 @@ void force_atlas2(GraphCOOView<vertex_t, edge_t, weight_t> &graph,
...
+ * the better the solution will be approximated. The number of restarts depends on the problem
+ * size and should be kept low for instances above 2k cities.
+ * @param[in] beam_search                     Specify if the initial solution should use KNN
+ * for an approximation solution.
+ * @param[in] k                               Beam width to use in the search.
+ * @param[in] nstart                          Start from a specific position.
+ * @param[in] verbose                         Logs configuration and iterative improvement.
+ *
+ */
+float traveling_salesman(raft::handle_t &handle,
```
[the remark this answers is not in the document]
hlinsen: Created a [PR](https://github.com/rapidsai/raft/pull/143) to make the `brute_force_knn` handle const for 0.19

datasets/tsplib/a280.tsp
hlinsen: Added [file](https://github.com/rapidsai/cugraph/pull/1360/files#diff-bf9f6511a73f6a7380290414afb8f3a2de6da4f155d3f0fa914e4ed1bf4f6cd1) with negative weights

hlinsen:
rerun tests

python/cugraph/tests/test_traveling_salesperson.py
```diff
@@ -0,0 +1,71 @@
...
+                        )
+    gdf = gdf.dropna()
+    gdf['vertex'] = gdf['vertex'].str.strip()
+    gdf['vertex'] = gdf['vertex'].astype("int32")
+    return gdf
+
+
+@pytest.mark.parametrize("tsplib_file, ref_cost", utils.DATASETS_TSPLIB)
+def test_traveling_salesperson(tsplib_file, ref_cost):
+    gc.collect()
```
rlratzel, line 53: We've been slowly removing these in favor of a single setup function, as shown [here](https://github.com/rapidsai/cugraph/blob/466e29a587251703796e27acbb660e4234a69b77/python/cugraph/tests/test_sssp.py#L63).

python/cugraph/tests/test_traveling_salesperson.py
```diff
@@ -0,0 +1,71 @@
...
+    pos_list = load_tsp(tsplib_file)
+    # cugraph
+    t1 = time.time()
+    cu_route, cu_cost = cugraph.traveling_salesperson(pos_list, restarts=4096)
```
rlratzel, line 57: We've been slowly removing the timers and print statements in favor of the benchmark fixture.  To use that, you would add `gpubenchmark` to the list of args to the test function (as was done [here](https://github.com/rapidsai/cugraph/blob/466e29a587251703796e27acbb660e4234a69b77/python/cugraph/tests/test_sssp.py#L207)), then use the fixture to call the code to be timed:
```
cu_route, cu_cost = gpubenchmark(cugraph.traveling_salesperson, pos_list, restarts=4096)
```

python/cugraph/traversal/traveling_salesperson.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Returns
+    -------
+    route : cudf.Series
+        cudf.Series of size V containing the ordered list of vertices
+        than needs to be visited.
+    """
+
+    if not isinstance(pos_list, cudf.DataFrame):
+        raise Exception("Instance should be cudf.DataFrame")
```
rlratzel, line 59: Might be more correct to raise `TypeError`

python/cugraph/traversal/traveling_salesperson.py
```diff
@@ -0,0 +1,75 @@
...
+
+    null_check(pos_list['vertex'])
+    null_check(pos_list['x'])
+    null_check(pos_list['y'])
+
+    if not pos_list[pos_list['vertex'] == nstart].index:
+        raise Exception("nstart should be in vertex ids")
```
rlratzel, line 66: Maybe `ValueError` or `IndexError`?  (all built-in exception descriptions are [here](https://docs.python.org/3/library/exceptions.html))

conda/recipes/libcugraph/meta.yaml
```diff
@@ -33,6 +33,8 @@ requirements:
     - ucx-py {{ minor_version }}
     - ucx-proc=*=gpu
     - gtest
+    - faiss-proc=*=cuda
+    - libfaiss=1.6.3
```
rlratzel, line 37: Do these need to also be runtime dependencies (as done by cuML [here](https://github.com/rapidsai/cuml/blob/f1abbeb90522090154cc1de657667162fc2a8628/conda/recipes/libcuml/meta.yaml#L51))?

Also, we seem to be okay without it, but I wonder if we need to think about adding a `host` section as well to make our recipe "cross-capable" (as described [here](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#host))?  This would also be more consistent with cuML too.

rlratzel (changes requested):
Looks good, just a few minor change requests.

build.sh
```diff
@@ -19,7 +19,7 @@ ARGS=$*
 REPODIR=$(cd $(dirname $0); pwd)
 LIBCUGRAPH_BUILD_DIR=${LIBCUGRAPH_BUILD_DIR:=${REPODIR}/cpp/build}
 
-VALIDARGS="clean libcugraph cugraph docs -v -g -n --allgpuarch --show_depr_warn -h --help"
+VALIDARGS="clean libcugraph cugraph docs -v -g -n --allgpuarch --show_depr_warn -h --help --buildfaiss"
```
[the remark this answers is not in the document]
hlinsen: Thanks for noticing, it's fixed in this commit https://github.com/rapidsai/cugraph/pull/1360/commits/f85cb0d6d5a8f7bafef61891af3d03ec1f3d43c2

python/cugraph/tests/test_traveling_salesperson.py
hlinsen: Thanks, I updated the test accordingly: https://github.com/rapidsai/cugraph/pull/1360/commits/edf0ca511b583fe35cff84f05418b169e11c4d96

python/cugraph/traversal/traveling_salesperson.py
hlinsen: Updated here: https://github.com/rapidsai/cugraph/pull/1360/commits/c99fb38aee0636c84726e66735b4622c69babc0b

cpp/tests/traversal/tsp_test.cu
hlinsen: Added the fixme in this commit: https://github.com/rapidsai/cugraph/pull/1360/commits/f85cb0d6d5a8f7bafef61891af3d03ec1f3d43c2

BradReesWork:
rerun test

BradReesWork:
rerun tests

rlratzel (approved):

afender (approved):

aschaffer (approved):

BradReesWork:
@gpucibot merge

PR #1362, opened 2021-01-28
Use new coarsen_graph primitive in Louvain

ChuckHastings:
Modify experimental::louvain to use the new `coarsen_graph` primitive.  This replaces the original implementation of `shrink_graph`.


ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

cpp/src/experimental/louvain.cuh
```diff
@@ -1205,350 +1206,51 @@ class Louvain {
...
     local_num_cols_     = current_graph_view_.get_number_of_local_adj_matrix_partition_cols();
-    local_num_edges_    = new_src_v.size();
+    base_vertex_id_     = current_graph_view_.get_local_vertex_first();
+
+    raft::copy(&local_num_edges_,
+               current_graph_view_.offsets() +
+                 current_graph_view_.get_local_adj_matrix_partition_row_last(0) -
+                 current_graph_view_.get_local_adj_matrix_partition_row_first(0),
+               1,
+               stream_);
```
seunghwak, line 1232: Better avoid directly accessing offsets() here.

Currently we can do this with

```
local_num_edges_ = edge_t{0};
for (size_t i = 0; i < current_graph_view_.get_number_of_local_adj_matrix_partitions(); ++i) {
  local_num_edges += current_graph_view_.get_number_of_local_adj_matrix_partition_edges(i);
}
```

Ideally, I think we need to add get_number_of_local_edges() (which basically does what the code above does) to graph_view_t (I thought I did this but I cannot find...).

seunghwak (commented):
This looks good to me except for few minor complaints.

cpp/src/community/louvain.cu
```diff
@@ -87,8 +87,20 @@ std::pair<size_t, weight_t> louvain(
...
+                 thrust::make_counting_iterator<vertex_t>(graph_view.get_number_of_vertices()),
+                 vertex_ids_v.begin());
+
+    partition_at_level<vertex_t, multi_gpu>(handle,
+                                            runner.get_dendrogram(),
+                                            vertex_ids_v.data(),
+                                            clustering,
+                                            runner.get_dendrogram().num_levels());
 
     // FIXME: Consider returning the Dendrogram at some point
```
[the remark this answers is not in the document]
ChuckHastings: I could consider moving the FIXME, but I think it is still relevant.

The current python API (and every version since we integrated) expects the C++ code to flatten the result and return the final clustering.  The networkx python implementation of this returns a Dendrogram and the caller can choose at what level to flatten the Dendrogram later.  This seems better... but I didn't want to break the python API at this point.

cpp/src/experimental/louvain.cuh
ChuckHastings: I'm hoping that `local_num_edges_` will become obsolete once I modify `update_by_delta_modularity` to use the new primitives.  But that won't be until the next release.

I will make this change (occurs in at least one other place in `louvain.cuh`) and push an update.

cpp/src/community/louvain.cu
```diff
@@ -87,8 +87,20 @@ std::pair<size_t, weight_t> louvain(
...
-    // TODO: implement this...
-    // runner.get_dendrogram().partition_at_level(clustering, runner.get_dendrogram().num_levels());
+
+    rmm::device_uvector<vertex_t> vertex_ids_v(graph_view.get_number_of_vertices(),
+                                               handle.get_stream());
+
+    thrust::copy(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+                 thrust::make_counting_iterator<vertex_t>(0),
+                 thrust::make_counting_iterator<vertex_t>(graph_view.get_number_of_vertices()),
+                 vertex_ids_v.begin());
```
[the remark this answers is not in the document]
ChuckHastings: Of course.  Included in the next push.

cpp/src/experimental/louvain.cuh
```diff
@@ -418,11 +421,10 @@ class Louvain {
       base_src_vertex_id_ = graph_view.get_local_adj_matrix_partition_row_first(0);
       base_dst_vertex_id_ = graph_view.get_local_adj_matrix_partition_col_first(0);
 
-      raft::copy(&local_num_edges_,
-                 graph_view.offsets() + graph_view.get_local_adj_matrix_partition_row_last(0) -
-                   graph_view.get_local_adj_matrix_partition_row_first(0),
-                 1,
-                 stream_);
+      local_num_edges_ = edge_t{0};
+      for (size_t i = 0; i < graph_view.get_number_of_local_adj_matrix_partitions(); ++i) {
```
aschaffer, line 425: Consider replacing custom loop by algorithm (in this case, probably `std::accumulate()`).

cpp/src/experimental/louvain.cuh
```diff
@@ -1205,350 +1207,49 @@ class Louvain {
...
-    src_indices_v_.resize(new_src_v.size());
-
     local_num_vertices_ = current_graph_view_.get_number_of_local_vertices();
     local_num_rows_     = current_graph_view_.get_number_of_local_adj_matrix_partition_rows();
     local_num_cols_     = current_graph_view_.get_number_of_local_adj_matrix_partition_cols();
-    local_num_edges_    = new_src_v.size();
+    base_vertex_id_     = current_graph_view_.get_local_vertex_first();
+
+    local_num_edges_ = edge_t{0};
+    for (size_t i = 0; i < current_graph_view_.get_number_of_local_adj_matrix_partitions(); ++i) {
```
aschaffer, line 1229: Consider replacing custom loop by algorithm (in this case, probably `std::accumulate()`).

cpp/src/experimental/louvain.cuh
```diff
@@ -1205,350 +1207,49 @@ class Louvain {
...
-      dst_v.begin(),
-      dst_v.end(),
-      thrust::make_zip_iterator(thrust::make_tuple(src_v.begin(), weight_v.begin())));
-    thrust::stable_sort_by_key(
-      rmm::exec_policy(stream_)->on(stream_),
-      src_v.begin(),
-      src_v.end(),
-      thrust::make_zip_iterator(thrust::make_tuple(dst_v.begin(), weight_v.begin())));
+    rmm::device_uvector<vertex_t> numbering_indices(numbering_map.size(), stream_);
+    thrust::sequence(rmm::exec_policy(stream_)->on(stream_),
```
aschaffer, line 1239: Why creating O(n) sequences rather than using O(1) `counting_iterator`? (above and in a few other places)

aschaffer (changes requested):

cpp/src/experimental/louvain.cuh
ChuckHastings: `cugraph::experimental::relabel` expects `std::tuple<vertex_t const *, vertex_t const *>` as the second parameter (where this is used).  I don't know how to make a counting iterator that provides a `vertex_t const *`.

The original code (before converting to using `relabel`) did in place sorts so the sequence needed to be realized.  If there is a way to convert it to a counting iterator, I'd be happy to try it.

cpp/src/experimental/louvain.cuh
ChuckHastings: This is a temporary solution until @seunghwak exposes a `get_number_of_local_edges()` method on the graph view.

I cannot see a clean way - given the current `graph_view` API - to simplify this into an std algorithm.  The only good way I saw was to use `std::transform_reduce` which isn't available in C++ 14 (our current baseline requirement).  I could use `thrust::transform_reduce` and have it specify a host function, I suppose.  I'm inclined to let Seunghwa add the method (probably in version 0.19) and using it here rather than create a sophisticated temporary solution.

cpp/src/experimental/coarsen_graph.cu
```diff
@@ -569,5 +569,145 @@ coarsen_graph(raft::handle_t const &handle,
               int32_t const *labels,
               bool do_expensive_check);
 
+template std::tuple<std::unique_ptr<graph_t<int32_t, int64_t, float, true, true>>,
```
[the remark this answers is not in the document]
ChuckHastings: This should be done for all of the `graph_functions.hpp` functions.  The `renumber_edgelist` ones that you modified last month, these `coarsen_graph` ones, the `relabel` function and the `extract_induced_subgraphs`.  I'd suggest we do everything in that file at once.  Perhaps we should create an issue in 0.19 to address this.

cpp/src/experimental/louvain.cuh
aschaffer:
```
auto sz = graph_view.get_number_of_local_adj_matrix_partitions();
local_num_edges_ = thrust::reduce(thrust::host, 
                                                         thrust::make_counting_iterator(0),
                                                         thrust::make_counting_iterator(sz),
                                                         [graph_view] (auto indx1, auto indx2){
                                                            return graph_view.get_number_of_local_adj_matrix_partition_edges(indx1)+
graph_view.get_number_of_local_adj_matrix_partition_edges(indx2);
                                                         });
```
Wouldn't that work?

cpp/src/experimental/louvain.cuh
aschaffer: True, `thrust::sort()`, `thrust::sort_by_key()` expect a modifiable range (hence cannot be counting iterators).

cpp/src/experimental/louvain.cuh
ChuckHastings: I don't think that works.  The reduction function has to be able to be applied to the output of the reduction function, it can't do a transformation from indices to sums.  That's why transform_reduce exists.

I just pushed a new version using `thrust::transform_reduce`.  It's a bit ugly because the graph_view object is not trivially copyable.  Ultimately this will go away once the new method is provided.

cpp/tests/community/louvain_test.cu
```diff
@@ -62,20 +62,32 @@ TEST(louvain, success)
...
-  ASSERT_GE(min, 0);
-  ASSERT_GE(modularity, 0.402777 * 0.95);
-  ASSERT_EQ(result_v, result_h);
+  // "FIXME": remove this check once we drop support for Pascal
+  //
+  // Calling louvain on Pascal will throw an exception, we'll check that
+  // this is the behavior while we still support Pascal (device_prop.major < 7)
+  //
+  cudaDeviceProp device_prop;
+  CUDA_CHECK(cudaGetDeviceProperties(&device_prop, 0));
```
seunghwak, line 71: You can use 
```const cudaDeviceProp& get_device_properties() const```
https://github.com/rapidsai/raft/blob/branch-0.18/cpp/include/raft/handle.hpp#L181 instead.

File: cpp/tests/community/louvain_test.cu [hunk 9]
```diff
@@ -62,20 +62,32 @@ TEST(louvain, success)
...
+
+  if (device_prop.major < 7) {
```
seunghwak, line 73: Should this test file be .cu?

Our users are C++ and I think we should allow C++ users using g++ to use libcugraph, and to test that, I think our C++ files better be .cpp than .cu.

cpp/tests/community/louvain_test.cu
```diff
@@ -123,19 +135,31 @@ TEST(louvain_renumbered, success)
...
   raft::handle_t handle;
 
-  std::tie(num_level, modularity) = cugraph::louvain(handle, G, result_v.data().get());
+  // "FIXME": remove this check once we drop support for Pascal
+  //
+  // Calling louvain on Pascal will throw an exception, we'll check that
+  // this is the behavior while we still support Pascal (device_prop.major < 7)
+  //
+  cudaDeviceProp device_prop;
+  CUDA_CHECK(cudaGetDeviceProperties(&device_prop, 0));
```
seunghwak, line 144: Again, RAFT has a utility function to get cudaDeviceProp.

cpp/tests/community/louvain_test.cu
seunghwak: And better use rmm::device_uvector instead of rmm::device_vector. rmm::device_uvector compiles with g++. rmm:device_vector is a thrust::device_vector and requires nvcc.

cpp/tests/community/louvain_test.cu
```diff
@@ -62,20 +62,32 @@ TEST(louvain, success)
...
+
+  if (device_prop.major < 7) {
+    EXPECT_THROW(cugraph::louvain(handle, G, result_v.data().get()), cugraph::logic_error);
+  } else {
+    std::tie(num_level, modularity) = cugraph::louvain(handle, G, result_v.data().get());
+
+    cudaMemcpy((void*)&(cluster_id[0]),
+               result_v.data().get(),
+               sizeof(int) * num_verts,
+               cudaMemcpyDeviceToHost);
```
seunghwak, line 81: Better use raft::update_host followed by stream sync.

aschaffer (approved):

rlratzel (approved):
Mostly reviewed the updates for checking for Pascal and those look good but I agree with Seunghwa's suggestions for the C++ tests.

cpp/tests/community/louvain_test.cu
ChuckHastings: Updated in next push

cpp/tests/community/louvain_test.cu
ChuckHastings: Renamed all of the Louvain derived tests to `.cpp`.  Modified tests to use `rmm::device_uvector`.

cpp/tests/community/louvain_test.cu
ChuckHastings: Modified all Louvain derived tests to use `raft::update_host` followed by `cudaDeviceSynchronize`

cpp/tests/community/louvain_test.cu
ChuckHastings: Updated

seunghwak (approved):
Thanks!!!

rlratzel:
rerun tests

BradReesWork:
@gpucibot merge

Issue #1075, opened 2020-08-13
[ENH] Update CMake to use FetchContent


ChuckHastings:
Scrambling to get cuda11 working in 0.15 resulted in adding FetchContent to get CUB and THRUST from git instead of using the version in the cuda library.

FetchContent is an alternative to ExternalProject with the advantage that the content is downloaded at configuration time instead of at build time.  The FetchContent approach was borrowed from cudf.

Several internet articles suggest NOT mixing FetchContent with ExternalProject.  It would be helpful to evaluate the best approach and either migrate everything to FetchContent or migrate everything to ExternalProject.

As an alternative, cuCollections suggests in their README using https://github.com/TheLartians/CPM.cmake, which is a wrapper around FetchContent that might make things easier.

BradReesWork:
closed by #1367

PR #1367, opened 2021-02-01
Update CMakeLists.txt files for consistency with RAPIDS and to support cugraph as an external project and other tech debt removal

### LINKED ISSUE #1137: [ENH] Fix CMake warnings in external projects
This should be applied to cuGraph as well:  https://github.com/rapidsai/cudf/pull/6176

### LINKED ISSUE #1266: [ENH] Add gtest to conda package
Apparently cudf and cuml have shifted to installing the test conda package rather than bringing gtest in as a cmake dependency.

Adding gtest as a conda package to the cugraph_dev environment would make it easier to test raft code and cugraph code concurrently.

rlratzel:
This PR makes cuGraph's cmake files more consistent with other RAPIDS libs by matching the minimum required cmake version, adding `project()` statements to cugraph's thirdparty modules, and using `CMAKE_CURRENT_SOURCE_DIR` appropriately so paths are relative to the CMakeLists.txt file rather than the top-level cmake dir of the project (since that may not be the cugraph cpp dir in the case of cugraph being used as an external project by another application).

This also adds a `CUDA_ARCHITECTURES=OFF` setting to suppress the warning printed for each test target. This setting may be replaced/changed once the findcudatoolkit feature is used in a future cmake version.

This also removes the Arrow and GTest cmake files since Arrow is not a direct dependency and those files were not being used, and GTest is now a build requirement in the conda dev environment and does not need to be built from source (the conda dev env files have been updated accordingly).

This PR also addresses much of #1075 , but not completely since gunrock is still using `ExternalProject` due to (I think) updates that need to be made to their cmake files to support this.

This was tested by observing a successful clean build, however it was **not** tested by creating a separate cmake application to simulate cugraph being used as a 3rd party package.

Note: the changes in this PR were modeled after https://github.com/rapidsai/rmm/pull/541

closes #1137 
closes #1266


cpp/CMakeLists.txt
```diff
@@ -474,7 +474,7 @@ if(BUILD_TESTS)
         # link_directoires (we should switch to target_link_directories once 3.13 becomes the
```
ChuckHastings, line 474: Not your PR, but could you fix the typo and spell directories correctly here :-)

ChuckHastings (approved):
One little thing would be nice.

cpp/CMakeLists.txt
rlratzel: Thanks, fixed in next commit.

ChuckHastings (approved):
LGTM

ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

cpp/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 #=============================================================================
 
-cmake_minimum_required(VERSION 3.12..3.17 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.14...3.17 FATAL_ERROR)
```
BradReesWork, line 17: cmake version should be 3.18+

BradReesWork (changes requested):

cpp/CMakeLists.txt
rlratzel: Thanks, fixed in latest commit.

BradReesWork (approved):

PR #1375, opened 2021-02-02
Update cudf groupby calls to explicitly set the sort option

### LINKED ISSUE #1371: [ENH] Update cudf groupby calls to explicitly set the sort option
A potentially breaking cudf change was recently merged and cugraph should be updated to explicitly set the `sort` option so we don't rely on a default behavior that may be different than what we're expecting.

See https://github.com/rapidsai/cudf/pull/7180 for details about the change.

The request here is to simply add a `sort=True` option to all groupby calls that **do not** already specify a `sort` behavior.

In addition, @ChuckHastings mentioned:

> we may want to flag those to go back later and see if sort=True is necessary.  It would be faster without the sort.  Or we could preemptively try setting sort=False in each of those places to see if it will work.

jnke2016:
set `sort=True` for all calls `groupby` 
closes #1371

### DIFF
--- a/python/cugraph/bsp/traversal/bfs_bsp.py
+++ b/python/cugraph/bsp/traversal/bfs_bsp.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020, NVIDIA CORPORATION.
+# Copyright (c) 2019-2021, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/python/cugraph/bsp/traversal/bfs_bsp.py
+++ b/python/cugraph/bsp/traversal/bfs_bsp.py
@@ -127,7 +127,8 @@ def bfs_df_pregel(_df, start, src_col='src', dst_col='dst', copy_data=True):
         aggsOut = OrderedDict()
         aggsOut['predecessor'] = 'min'
         aggsOut['distance'] = 'min'
-        _a = one_hop.groupby(['vertex'], as_index=False, sort=True).agg(aggsOut)
+        _a = one_hop.groupby(
+            ['vertex'], as_index=False, sort=True).agg(aggsOut)
 
         answer = cudf.concat([answer, _a])
 
--- a/python/cugraph/bsp/traversal/bfs_bsp.py
+++ b/python/cugraph/bsp/traversal/bfs_bsp.py
@@ -127,7 +127,7 @@ def bfs_df_pregel(_df, start, src_col='src', dst_col='dst', copy_data=True):
         aggsOut = OrderedDict()
         aggsOut['predecessor'] = 'min'
         aggsOut['distance'] = 'min'
-        _a = one_hop.groupby(['vertex'], as_index=False).agg(aggsOut)
+        _a = one_hop.groupby(['vertex'], as_index=False, sort=True).agg(aggsOut)
 
         answer = cudf.concat([answer, _a])
 
--- a/python/cugraph/centrality/betweenness_centrality.py
+++ b/python/cugraph/centrality/betweenness_centrality.py
@@ -271,7 +271,7 @@ def edge_betweenness_centrality(
         # overwrite the df with the sum of the values for all alike src/dst
         # vertex pairs, resulting in half the edges of the original df from the
         # symmeterized graph.
-        df = df.groupby(by=["src", "dst"]).sum().reset_index()
+        df = df.groupby(by=["src", "dst"], sort=True).sum().reset_index()
 
     if isNx is True:
         return df_edge_score_to_dictionary(df, 'betweenness_centrality')
--- a/python/cugraph/structure/number_map.py
+++ b/python/cugraph/structure/number_map.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2020, NVIDIA CORPORATION.
+# Copyright (c) 2021, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/python/cugraph/structure/number_map.py
+++ b/python/cugraph/structure/number_map.py
@@ -96,7 +96,8 @@ class NumberMap:
 
         def compute(self):
             if not self.numbered:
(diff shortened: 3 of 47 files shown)


python/cugraph/bsp/traversal/bfs_bsp.py
```diff
@@ -127,7 +127,8 @@ def bfs_df_pregel(_df, start, src_col='src', dst_col='dst', copy_data=True):
         aggsOut = OrderedDict()
         aggsOut['predecessor'] = 'min'
         aggsOut['distance'] = 'min'
-        _a = one_hop.groupby(['vertex'], as_index=False).agg(aggsOut)
+        _a = one_hop.groupby(
```
BradReesWork, line 130: this doesn't really need to be sorted - but it is okay since we should consider dropping the BSP folder in 0.19

python/cugraph/centrality/betweenness_centrality.py
```diff
@@ -271,7 +271,7 @@ def edge_betweenness_centrality(
         # overwrite the df with the sum of the values for all alike src/dst
         # vertex pairs, resulting in half the edges of the original df from the
         # symmeterized graph.
-        df = df.groupby(by=["src", "dst"]).sum().reset_index()
+        df = df.groupby(by=["src", "dst"], sort=True).sum().reset_index()
```
ChuckHastings, line 274: I would be surprised if this requires `sort=True`.  If you specify `sort=False` do any of the BC python tests fail?

python/cugraph/structure/number_map.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, df, src_col_names, dst_col_names, id_type,
             tmp = (
                 df[src_col_names]
                 .assign(count=source_count)
-                .groupby(src_col_names)
+                .groupby(src_col_names, sort=True)
```
ChuckHastings, line 70: This shouldn't require `sort=True`

python/cugraph/structure/number_map.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, df, src_col_names, dst_col_names, id_type,
                 tmp_dst = (
                     df[dst_col_names]
                     .assign(count=dest_count)
-                    .groupby(dst_col_names)
+                    .groupby(dst_col_names, sort=True)
```
ChuckHastings, line 83: This shouldn't require `sort=True`

python/cugraph/structure/number_map.py
```diff
@@ -96,7 +96,8 @@ def __init__(self, df, src_col_names, dst_col_names, id_type,
 
         def compute(self):
             if not self.numbered:
-                tmp = self.df.groupby(self.col_names).sum().sort_values(
+                tmp = self.df.groupby(
+                    self.col_names, sort=True).sum().sort_values(
```
ChuckHastings, line 100: This shouldn't require `sort=True`.  The groupby is followed by a sort later on the line.

python/cugraph/structure/number_map.py
```diff
@@ -197,7 +198,7 @@ def extract_vertices(
             s = (
                 df[src_col_names]
                 .assign(count=source_count)
-                .groupby(src_col_names)
+                .groupby(src_col_names, sort=True)
```
ChuckHastings, line 201: This shouldn't require `sort=True`

python/cugraph/structure/number_map.py
```diff
@@ -211,7 +212,7 @@ def extract_vertices(
                 d = (
                     df[dst_col_names]
                     .assign(count=dest_count)
-                    .groupby(dst_col_names)
+                    .groupby(dst_col_names, sort=True)
```
ChuckHastings, line 215: This shouldn't require `sort=True`

python/cugraph/structure/number_map.py
```diff
@@ -314,7 +315,7 @@ def compute(self):
 
                 numbering_map = rehashed_with_partition_id.map_partitions(
                     lambda df: df.groupby(
-                        self.col_names + ["hash", "partition"]
+                        self.col_names + ["hash", "partition"], sort=True
```
ChuckHastings, line 318: This does not require `sort=True`.  The groupby is followed explicitly by a sort.

python/cugraph/structure/number_map.py
```diff
@@ -326,7 +327,7 @@ def compute(self):
                 #  Compute base address for each partition
                 #
                 counts = numbering_map.map_partitions(
-                    lambda df: df.groupby("partition").count()
+                    lambda df: df.groupby("partition", sort=True).count()
```
ChuckHastings, line 330: This shouldn't require a `sort=True`.  We're only after a count, order doesn't matter.

python/cugraph/structure/symmetrize.py
```diff
@@ -84,7 +84,8 @@ def symmetrize_df(df, src_name, dst_name, multi=False, symmetrize=True):
     if multi:
         return gdf
     else:
-        return gdf.groupby(by=[src_name, dst_name], as_index=False).min()
+        return gdf.groupby(
+            by=[src_name, dst_name], as_index=False, sort=True).min()
```
ChuckHastings, line 88: This shouldn't require `sort=True`.  Should check if the unit tests fail.

python/cugraph/structure/symmetrize.py
```diff
@@ -138,7 +139,7 @@ def symmetrize_ddf(df, src_name, dst_name, weight_name=None):
 
     ddf = df.append(ddf2).reset_index(drop=True)
     result = (
-        ddf.groupby(by=[src_name, dst_name], as_index=False)
+        ddf.groupby(by=[src_name, dst_name], as_index=False, sort=True)
```
ChuckHastings, line 142: This shouldn't require sort=True. Should check if the unit tests fail.

ChuckHastings (changes requested):
I suspect most of these do not require `sort=True` to be set.  Marked ones I specifically think can be skipped (although some I think might be able to be skipped).

I recommend setting `sort=False` and running the python unit tests.  If they all pass then we don't need `sort=True` on those calls.  If some of them fail you should be able to identify which of these might need the `sort=True`.

BradReesWork (changes requested):

BradReesWork:
after a review of the code by Joseph, it turns out that we are not dependent on needing a sort after the groupby call.  Since there is no need to update any code, this PR is being closed

PR #1386, opened 2021-02-06
Updated CI scripts to use a different error handling convention, updated LD_LIBRARY_PATH for project flash runs

rlratzel:
* Updated CI scripts to use a different error handling convention
* Updated LD_LIBRARY_PATH for project flash runs
* Added extra logging to report status after each test command
* Added comments
* Removed unused "top-20 slowest" report
* Minor updates for consistency

Tested locally by simulating various error conditions (removed .so files, inserted errors in NBs, killed processes, etc.) and checked exit codes.  Still need to verify in a project Flash env, but using CI for that.


rlratzel:
@raydouglass , @ajschmidt8 - I made some assumptions in the top-level `gpu/build.sh` about certain commands being allowed to silently fail (there's a comment block in the script), but let me know if those commands should result in a CI failure as well.  Also, I moved the gtest runs to *after* the project Flash commands, next to the other test commands, but let me know if they still need to be run prior to the project Flash commands.

ci/gpu/build.sh
```diff
@@ -117,18 +121,21 @@ else
...
-    echo "$(${WORKSPACE}/ci/getGTestTimes.sh testoutput.txt | head -20)"
-
+    gpuci_logger "Running cuGraph notebook test script..."
     ${WORKSPACE}/ci/gpu/test-notebooks.sh 2>&1 | tee nbtest.log
+    CMDRETCODE=$?; EXITCODE=$((EXITCODE | $CMDRETCODE))
+    gpuci_logger "Ran cuGraph notebook test script : return code was: $CMDRETCODE, gpu/build.sh exit code is now: $EXITCODE"
     python ${WORKSPACE}/ci/utils/nbtestlog2junitxml.py nbtest.log
 fi
 
 if [ -n "\${CODECOV_TOKEN}" ]; then
```
ajschmidt8, line 136:
```suggestion
if [ -n "${CODECOV_TOKEN}" ]; then
```

This isn't related to your changes, but can you remove this erroneous `\`?

ci/gpu/test-notebooks.sh
```diff
@@ -12,18 +12,27 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#RAPIDS_DIR=/rapids
+# Any failing command will set EXITCODE to non-zero
+set -e           # abort the script on error, this will change for running tests (see below)
+set -o pipefail  # piped commands propagate their error
+set -E           # ERR traps are inherited by subcommands
+trap "EXITCODE=1" ERR
```
ajschmidt8, line 19: nice, I like the comments for some of these less commonly used flags.

ci/gpu/build.sh
```diff
@@ -98,9 +98,13 @@ fi
 # TEST - Run GoogleTest and py.tests for libcugraph and cuGraph
 ################################################################################
 
-set +e -Eo pipefail
+# Switch to +e to allow failing commands to continue the script, which is needed
+# so all testing commands run regardless of pass/fail. This requires the desired
+# exit code to be managed manually by updating it only for commands that
+# represent test failures - other less critical commands can fail if necessary
+# (nbtestlog2junitxml, codecov, etc.) to keep CI unblocked.
```
ajschmidt8, line 105: Hm. I know we talked about removing the `trap` here, but now I'm thinking it might be better to keep it. I think we'd be setting ourselves up for some potential issues in the future if we're not trapping all errors when `set +e` is set. Using the `trap` would make this behave more closely to `set -e`, where all erroneous exit codes are accounted for (instead of just specific ones).

Can you re-add the `trap` logic here?

ci/gpu/build.sh
```diff
@@ -98,9 +98,13 @@ fi
 # TEST - Run GoogleTest and py.tests for libcugraph and cuGraph
 ################################################################################
 
-set +e -Eo pipefail
```
ajschmidt8, line 101: Not sure if this was intentional, but I don't see the `set -E` carried over in your changes.

ajschmidt8 (changes requested):

ci/gpu/build.sh
rlratzel: fixed in next commit

ci/gpu/build.sh
rlratzel: sure I'll add it back in the next commit.

ci/gpu/build.sh
rlratzel: added in next commit along with the trap.

ajschmidt8 (approved):

raydouglass:
rerun tests

raydouglass (approved):

rlratzel:
The failing CI checks should be resolved as soon as PR #1362 is merged.

ci/gpu/build.sh
```diff
@@ -117,18 +122,19 @@ else
...
-
+    gpuci_logger "Running cuGraph notebook test script..."
     ${WORKSPACE}/ci/gpu/test-notebooks.sh 2>&1 | tee nbtest.log
+    gpuci_logger "Ran cuGraph notebook test script : return code was: $?, gpu/build.sh exit code is now: $EXITCODE"
     python ${WORKSPACE}/ci/utils/nbtestlog2junitxml.py nbtest.log
 fi
 
-if [ -n "\${CODECOV_TOKEN}" ]; then
+if [ -n "${CODECOV_TOKEN}" ]; then
     codecov -t \$CODECOV_TOKEN
```
dillon-cullinan, line 136: L136 also needs the escape character removed

dillon-cullinan (approved):

ci/gpu/build.sh
rlratzel: Thanks, done in most recent commit.

ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

Issue #1208, opened 2020-10-13
[FEA] Add compute_out_degrees, compute_in_degrees, compute_out_weight_sums, and compute_in_weight_sums functions to the experimental graph_view_t class


seunghwak:
We have a utility function to quickly compute degrees from offsets if we are computing out-degrees (if we store un-transposed adjacency matrix) or in-degrees (if we store transposed adjacency matrix).

https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/include/experimental/detail/graph_utils.cuh#L119

If we need to sum weights or need to compute in-degrees (if we store un-transposed adjacency matrix) or out-degrees (if we store transposed adjacency matrix), we can use 

copy_v_transform_reduce_in_nbr or copy_v_transform_reduce_out_nbr

We're already doing this in individual algorithms (e.g. see the PageRank & Louvain code below), but for better reusability, we'd better provide this as graph_view_t class's member functions.

https://github.com/rapidsai/cugraph/blob/branch-0.16/cpp/src/experimental/pagerank.cu#L142
```
    tmp_vertex_out_weight_sums.resize(pull_graph_view.get_number_of_local_vertices(),
                                      handle.get_stream());
    // FIXME: better refactor this out (computing out-degree).
    copy_v_transform_reduce_out_nbr(
      handle,
      pull_graph_view,
      thrust::make_constant_iterator(0) /* dummy */,
      thrust::make_constant_iterator(0) /* dummy */,
      [alpha] __device__(vertex_t src, vertex_t dst, weight_t w, auto src_val, auto dst_val) {
        return w;
      },
      weight_t{0.0},
      tmp_vertex_out_weight_sums.data());
```

https://github.com/rapidsai/cugraph/pull/1172/files#diff-9191c39890810158440a0cabf04e6356c5252233458b994dcb563ab350912e4aR669
```
    experimental::copy_v_transform_reduce_out_nbr(
      handle_,
      current_graph_view_,
      thrust::make_constant_iterator(0),
      thrust::make_constant_iterator(0),
      [] __device__(auto src, auto, auto wt, auto, auto) { return wt; },
      weight_t{0},
      vertex_weights_v_.begin());
```

With the proposed updates, the code above can be rewritten as

```auto vertex_weights_v = current_graph_view_.compute_out_weight_sums();```

I think this will be a good task to familiarize oneself with the pattern accelerator (or graph primitives) API and multi-GPU programming & testing.

seunghwak:
[...]

PR #1394, opened 2021-02-09
Add new primitives: compute_in|out_degrees, compute_in|out_weight_sums to graph_view_t

seunghwak:
Close https://github.com/rapidsai/cugraph/issues/1208

- [x] add compute_in|out_degrees, compute_in|out_weight_sums
- [x] replace PageRank's custom code to compute out-weight-sums to use graph_view_t's compute_out_weight_sums
- [x] add SG C++ tests


seunghwak:
@ChuckHastings FYI, https://github.com/rapidsai/cugraph/blob/branch-0.18/cpp/src/experimental/louvain.cuh#L556 should eventually be replaced with graph_view_t's compute_out_weight_sums(). I am not touching this at this moment assuming you're actively modifying the Louvain code.

seunghwak:
rerun tests

seunghwak:
rerun tests

cpp/src/experimental/graph_view.cu
```diff
@@ -301,6 +379,154 @@ graph_view_t<vertex_t,
...
+          typename edge_t,
+          typename weight_t,
+          bool store_transposed,
+          bool multi_gpu>
+rmm::device_uvector<edge_t>
+graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enable_if_t<multi_gpu>>::
+  compute_out_degrees(raft::handle_t const& handle) const
+{
+  if (store_transposed) {
+    return compute_minor_degrees(handle, *this);
```
afender, line 430: I noticed we have started to use minor/major terminology but seeing this here used concurrently with src/dst and in/out terminology makes me wonder if that's really helping. We should try to stick to the smallest possible set of names that refers to the same thing. 
Reconciling this is beyond this PR, we should have a separate issue/PR for this.

afender (commented):
We need to create an issue to update the python bindings and delete the legacy version of degree functions.

cpp/src/experimental/graph_view.cu
seunghwak: So the origin of minor/major might be from https://github.com/rapidsai/cugraph/blob/branch-0.19/cpp/include/patterns/copy_v_transform_reduce_in_out_nbr.cuh and this terminology has been used since the start of graph primitives.

This was introduced mainly for source code reuse, noting that iterating over incoming edges when the graph adjacency matrix is stored as is (CSR) and iterating over outgoing edges when the graph adjacency matrix stored as transposed (CSC) can be implemented using the same code.

Yeah... we may open a several issue and we may need to go through more intense discussion on terminologies as these are now used all over the graph primitives. If there is a clear benefit to make a switch, yes we should make a switch, but I assume there should be a very strong justification.

seunghwak: And this might be a much bigger discussion, but we may later more intensively investigate the performance benefit of storing graph adjacency matrix as transposed.

This was mainly introduced to avoid atomics, but AFAIK, GPU's performance on atomics has improved drastically in the past some years and if the performance gain is not large enough, we may reconsider this issue. If the performance gain is marginal, dropping this will significantly simplify our codebase.

seunghwak: And this terminology issue will become irrelevant.

ChuckHastings (approved):

cpp/src/experimental/graph_view.cu
afender: Sounds good to me. Thanks for elaborating. Let's open an issue to keep track of this.

afender (approved):

BradReesWork:
@gpucibot merge

PR #1392, opened 2021-02-09
Rename sort_and_shuffle to groupby_gpuid_and_shuffle

seunghwak:
Rename to better reflect what this function should do than how it is currently implemented (which can change in the future for better performance).

PR #1398, opened 2021-02-10
[NO MERGE] Louvain using new graph primitives

ChuckHastings:
This is for @seunghwak to review and discuss.  Whatever we do in this PR will be migrated into a different branch for implementation.

This PR should not be merged.

### DIFF
--- a/cpp/tests/experimental/louvain_primitives.cu
+++ b/cpp/tests/experimental/louvain_primitives.cu
@@ -1,22 +1,11 @@
 #include <experimental/graph.hpp>
 
+#include <patterns/copy_v_transform_reduce_key_aggregated_out_nbr.cuh>
+#include <patterns/transform_reduce_by_adj_matrix_row_col_key_e.cuh>
+#include <utilities/dataframe_buffer.cuh>
 #include <utilities/shuffle_comm.cuh>
 
 #include <rmm/device_uvector.hpp>
-
-// TODO:
-//    The following 3 include files are not referenced in my file,
-//    however if they are not included prior to the two pattern
-//    includes below then the compilation fails.  I suspect these
-//    should be included somehow in the patterns includes below
-//
-#include <matrix_partition_device.cuh>
-#include <patterns/edge_op_utils.cuh>
-#include <utilities/host_scalar_comm.cuh>
-// END EXTRA INCLUDES
-
-#include <patterns/copy_v_transform_reduce_key_aggregated_out_nbr.cuh>
-#include <patterns/transform_reduce_by_adj_matrix_row_col_key_e.cuh>
 
 template <typename graph_t>
 void update_by_delta_modularity(raft::handle_t const& handle,
@@ -35,25 +24,29 @@ void update_by_delta_modularity(raft::handle_t const& handle,
 
   raft::update_device(cluster_v.data(), cluster_h.data(), num_verts, handle.get_stream());
 
-  auto tmp = cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
-    handle,
-    graph_view,
-    cluster_v.begin(),
-    cluster_v.begin(),
-    cluster_v.begin(),
-    [] __device__(auto r, auto c, auto w, auto rv, auto cv) {
-      printf("transform reduce (%d,%d,%g,%d,%d)\n", (int)r, (int)c, (float)w, (int)rv, (int)cv);
-      return w;
-    },
-    weight_t{0});
+  rmm::device_uvector<vertex_t> communities(0, handle.get_stream());
+  rmm::device_uvector<weight_t> community_in_weight_sums(0, handle.get_stream());
+
+  std::tie(communities, community_in_weight_sums) =
+    cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
+      handle,
+      graph_view,
+      cluster_v.begin(),
+      cluster_v.begin(),
+      cluster_v.begin(),
+      [] __device__(auto r, auto c, auto w, auto rv, auto cv) {
+        printf("transform reduce (%d,%d,%g,%d,%d)\n", (int)r, (int)c, (float)w, (int)rv, (int)cv);
+        return w;
+      },
+      weight_t{0});
 
   thrust::for_each(thrust::device,
                    thrust::make_counting_iterator(0),
                    thrust::make_counting_iterator(1),
-                   [b_key   = std::get<0>(tmp).begin(),
-                    e_key   = std::get<0>(tmp).end(),
-                    b_value = std::get<1>(tmp).begin(),
-                    e_value = std::get<1>(tmp).end()] __device__(auto) {
+                   [b_key   = communities.begin(),
+                    e_key   = communities.end(),
+                    b_value = community_in_weight_sums.begin(),
+                    e_value = community_in_weight_sums.end()] __device__(auto) {
                      char comma[2] = {0, 0};
                      printf("keys = ");
                      for (auto p = b_key; p != e_key; ++p) {
@@ -71,21 +64,25 @@ void update_by_delta_modularity(raft::handle_t const& handle,
                    });
 
   if (graph_t::is_multi_gpu) {
-    cugraph::experimental::sort_and_shuffle_kv_pairs(
-      handle.get_comms(),
-      std::get<0>(tmp).begin(),
-      std::get<0>(tmp).end(),
-      std::get<1>(tmp).begin(),
-      [] __device__(auto) { return 0; },
-      handle.get_stream());
+    std::tie(communities, community_in_weight_sums, std::ignore) =
+      cugraph::experimental::sort_and_shuffle_kv_pairs(
+        handle.get_comms(),
+        communities.begin(),
+        communities.end(),
+        community_in_weight_sums.begin(),
+        [] __device__(auto) { return 0; },
+        handle.get_stream());
   }
-
-  rmm::device_vector<thrust::tuple<vertex_t, weight_t>> output_v(num_verts);
 
   // NOTE: these are populated and shuffled before call to update_by_delta_modularity
   rmm::device_uvector<weight_t> old_cluster_sum_v(num_verts, handle.get_stream());
   rmm::device_uvector<weight_t> src_vertex_weight_v(num_verts, handle.get_stream());
   rmm::device_uvector<weight_t> src_cluster_weight_v(num_verts, handle.get_stream());
+
+  // rmm::device_vector<thrust::tuple<vertex_t, weight_t>> output_v(num_verts);
+  auto output_buffer =
+    cugraph::experimental::allocate_dataframe_buffer<thrust::tuple<vertex_t, weight_t>>(
+      num_verts, handle.get_stream());
 
   copy_v_transform_reduce_key_aggregated_out_nbr(
     handle,
@@ -93,9 +90,9 @@ void update_by_delta_modularity(raft::handle_t const& handle,
     thrust::make_zip_iterator(thrust::make_tuple(
       old_cluster_sum_v.begin(), src_vertex_weight_v.begin(), src_cluster_weight_v.begin())),
     cluster_v.begin(),
-    std::get<0>(tmp).begin(),
-    std::get<0>(tmp).end(),
-    std::get<1>(tmp).begin(),
+    communities.begin(),
+    communities.end(),
+    community_in_weight_sums.begin(),
     [total_edge_weight, resolution] __device__(
       auto src, auto neighbor_cluster, auto new_cluster_sum, auto src_info, auto a_new) {
       auto old_cluster_sum = thrust::get<0>(src_info);
@@ -113,19 +110,40 @@ void update_by_delta_modularity(raft::handle_t const& handle,
              old_cluster_sum,
              k_k,
              a_old,
-             a_new.load());
+             a_new);
 
       return thrust::make_tuple(neighbor_cluster, delta_modularity);
     },
     [] __device__(auto p1, auto p2) {
-      if (thrust::get<1>(p1) < thrust::get<1>(p2))
-        return p1;
-      else
-        return p2;
+      printf("reducing (%d, %g) with (%d, %g)\n",
+             thrust::get<0>(p1),
+             thrust::get<1>(p1),
+             thrust::get<0>(p2),
+             thrust::get<1>(p2));
+
+      return (thrust::get<1>(p1) < thrust::get<1>(p2)) ? p1 : p2;
     },
     thrust::make_tuple(vertex_t{-1}, weight_t{std::numeric_limits<weight_t>::max()}),
-    output_v.begin());
+    // output_v.begin());
+    cugraph::experimental::get_dataframe_buffer_begin<thrust::tuple<vertex_t, weight_t>>(
+      output_buffer));
 
+  thrust::for_each(
+    thrust::device,
+    thrust::make_counting_iterator(0),
+    thrust::make_counting_iterator(1),
+    [b_output =
+       cugraph::experimental::get_dataframe_buffer_begin<thrust::tuple<vertex_t, weight_t>>(
+         output_buffer),
+     num_verts] __device__(auto) {
+      char comma[2] = {0, 0};
+      printf("output = ");
+      for (auto p = b_output; p != (b_output + num_verts); ++p) {
+        printf("%s(%d, %g)", comma, thrust::get<0>(*p), thrust::get<1>(*p));
+        comma[0] = ',';
+      }
+      printf("\n");
+    });
   //
   //  output_v contains (for each vertex) the pair (neighbor_cluster, delta_modularity)
   //  for all neighboring clusters with the highest delta modularity
--- a/cpp/tests/experimental/louvain_primitives.cu
+++ b/cpp/tests/experimental/louvain_primitives.cu
@@ -1,0 +1,179 @@
+#include <experimental/graph.hpp>
+
+#include <utilities/shuffle_comm.cuh>
+
+#include <rmm/device_uvector.hpp>
+
+// TODO:
+//    The following 3 include files are not referenced in my file,
+//    however if they are not included prior to the two pattern
+//    includes below then the compilation fails.  I suspect these
+//    should be included somehow in the patterns includes below
+//
+#include <matrix_partition_device.cuh>
+#include <patterns/edge_op_utils.cuh>
+#include <utilities/host_scalar_comm.cuh>
+// END EXTRA INCLUDES
+
+#include <patterns/copy_v_transform_reduce_key_aggregated_out_nbr.cuh>
+#include <patterns/transform_reduce_by_adj_matrix_row_col_key_e.cuh>
+
+template <typename graph_t>
+void update_by_delta_modularity(raft::handle_t const& handle,
+                                graph_t const& graph_view,
+                                std::vector<typename graph_t::vertex_type> cluster_h,
+                                typename graph_t::weight_type total_edge_weight,
+                                typename graph_t::weight_type resolution)
+{
+  using vertex_t = typename graph_t::vertex_type;
+  using edge_t   = typename graph_t::edge_type;
+  using weight_t = typename graph_t::weight_type;
+
+  vertex_t num_verts = graph_view.get_number_of_vertices();
+
+  rmm::device_uvector<vertex_t> cluster_v(num_verts, handle.get_stream());
+
+  raft::update_device(cluster_v.data(), cluster_h.data(), num_verts, handle.get_stream());
+
+  auto tmp = cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
+    handle,
+    graph_view,
+    cluster_v.begin(),
+    cluster_v.begin(),
+    cluster_v.begin(),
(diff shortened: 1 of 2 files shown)


cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+    cugraph::experimental::sort_and_shuffle_kv_pairs(
+      handle.get_comms(),
+      std::get<0>(tmp).begin(),
+      std::get<0>(tmp).end(),
+      std::get<1>(tmp).begin(),
+      [] __device__(auto) { return 0; },
+      handle.get_stream());
+  }
+
+  rmm::device_vector<thrust::tuple<vertex_t, weight_t>> output_v(num_verts);
```
seunghwak, line 83: I am still working on fixing all the compile errors, but just to give you a heads up.

First pull updates from https://github.com/rapidsai/cugraph/pull/1399

graph primitives does not take an array of tuples but takes tuples of arrays instead.

So what you need to do is

1) #include <utilities/dataframe_buffer.cuh>
2) auto output_buffer = cugraph::experimental::allocate_dataframe_buffer<thrust::tuple<vertex_t, weight_t>>(num_verts, handle.get_stream()); instead of defining output_v here.
3) use cugraph::experimental::get_dataframe_buffer_begin<thrust::tuple<vertex_t, weight_t>>(output_buffer) instead of output_v.begin()

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+// TODO:
+//    The following 3 include files are not referenced in my file,
+//    however if they are not included prior to the two pattern
+//    includes below then the compilation fails.  I suspect these
+//    should be included somehow in the patterns includes below
+//
+#include <matrix_partition_device.cuh>
+#include <patterns/edge_op_utils.cuh>
+#include <utilities/host_scalar_comm.cuh>
+// END EXTRA INCLUDES
```
seunghwak, line 16: You can delete this after pull updates from https://github.com/rapidsai/cugraph/pull/1399

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+  using edge_t   = typename graph_t::edge_type;
+  using weight_t = typename graph_t::weight_type;
+
+  vertex_t num_verts = graph_view.get_number_of_vertices();
+
+  rmm::device_uvector<vertex_t> cluster_v(num_verts, handle.get_stream());
+
+  raft::update_device(cluster_v.data(), cluster_h.data(), num_verts, handle.get_stream());
+
+  auto tmp = cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
```
seunghwak, line 38:
```
-  auto tmp = cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
+  rmm::device_uvector<vertex_t> communities(0, handle.get_stream());
+  rmm::device_uvector<weight_t> community_in_weight_sums(0, handle.get_stream());
+  std::tie(communities, community_in_weight_sums) = cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+    },
+    weight_t{0});
+
+  thrust::for_each(thrust::device,
+                   thrust::make_counting_iterator(0),
+                   thrust::make_counting_iterator(1),
+                   [b_key   = std::get<0>(tmp).begin(),
+                    e_key   = std::get<0>(tmp).end(),
+                    b_value = std::get<1>(tmp).begin(),
+                    e_value = std::get<1>(tmp).end()] __device__(auto) {
```
seunghwak, line 56:
```
-                   [b_key   = std::get<0>(tmp).begin(),
-                    e_key   = std::get<0>(tmp).end(),
-                    b_value = std::get<1>(tmp).begin(),
-                    e_value = std::get<1>(tmp).end()] __device__(auto) {
+                   [b_key   = communities.begin(),
+                    e_key   = communities.end(),
+                    b_value = community_in_weight_sums.begin(),
+                    e_value = community_in_weight_sums.end()] __device__(auto) {
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+                     comma[0] = (char)0;
+                     for (auto p = b_value; p != e_value; ++p) {
+                       printf("%s%g", comma, *p);
+                       comma[0] = ',';
+                     }
+                     printf("\n");
+                   });
+
+  if (graph_t::is_multi_gpu) {
+    cugraph::experimental::sort_and_shuffle_kv_pairs(
```
seunghwak, line 74:
```
-    cugraph::experimental::sort_and_shuffle_kv_pairs(
+    rmm::device_uvector<vertex_t> tmp_communities(0, handle.get_stream());
+    rmm::device_uvector<weight_t> tmp_community_in_weight_sums(0, handle.get_stream());
+    std::tie(communities, community_in_weight_sums, std::ignore)
+    = cugraph::experimental::sort_and_shuffle_kv_pairs(
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+      handle.get_comms(),
+      std::get<0>(tmp).begin(),
+      std::get<0>(tmp).end(),
+      std::get<1>(tmp).begin(),
```
seunghwak, line 78:
```
-      std::get<0>(tmp).begin(),
-      std::get<0>(tmp).end(),
-      std::get<1>(tmp).begin(),
+      communities.begin(),
+      communities.end(),
+      community_in_weight_sums.begin(),
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+      [] __device__(auto) { return 0; },
+      handle.get_stream());
```
seunghwak, line 80:
```
+    communities = std::move(tmp_communities);
+    community_in_weight_sums = std::move(tmp_community_in_weight_sums);
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+
+  copy_v_transform_reduce_key_aggregated_out_nbr(
+    handle,
+    graph_view,
+    thrust::make_zip_iterator(thrust::make_tuple(
+      old_cluster_sum_v.begin(), src_vertex_weight_v.begin(), src_cluster_weight_v.begin())),
+    cluster_v.begin(),
+    std::get<0>(tmp).begin(),
+    std::get<0>(tmp).end(),
+    std::get<1>(tmp).begin(),
```
seunghwak, line 98:
```
-    std::get<0>(tmp).begin(),
-    std::get<0>(tmp).end(),
-    std::get<1>(tmp).begin(),
+    communities.begin(),
+    communities.end(),
+    community_in_weight_sums.begin(),
```

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+                                         (total_edge_weight * total_edge_weight));
+
+      printf("compute delta_modularity (%d, %d, %g, %g, %g, %g, %g)\n",
+             src,
+             neighbor_cluster,
+             new_cluster_sum,
+             old_cluster_sum,
+             k_k,
+             a_old,
+             a_new.load());
```
seunghwak, line 116: a_new.load() => a_new

cpp/tests/experimental/louvain_primitives.cu
seunghwak: Actually, you can skip this and avoid defining tmp_... by using std::tie(communities, community_in_weight_sums) = cugraph::experimental::sort_and_shuffle_kv_pairs(...);

Also note that sort_and_shuffle_kv_pairs will be renamed to groupby_gpuids_and_shuffle_kv_pairs in https://github.com/rapidsai/cugraph/pull/1392

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,179 @@
...
+
+      return thrust::make_tuple(neighbor_cluster, delta_modularity);
+    },
+    [] __device__(auto p1, auto p2) {
+      if (thrust::get<1>(p1) < thrust::get<1>(p2))
+        return p1;
+      else
+        return p2;
```
seunghwak, line 124:
```
return thrust::get<1>(p1) < thrust::get<1>(p2) ? p1 : p2;
```

Here ```p1```'s type is ```thrust::detail::tuple_of_iterator_references<int &, float &, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>```

and ```p2```'s type is ```thrust::tuple<int, float, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>```

so auto deduction of the return type based on ```return p1``` and ```return p2``` leads to conflict. Combining the two return statements to one fixes this.

If you pull the latest updates from PR 1399 and you apply this change as well, the code should at least compile.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: Changed this.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: Thanks, I deleted these and it compiled correctly

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: I knew something like this would work, thanks for providing the syntax I was too lazy to construct 👍 

I integrated this into the code.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: Done

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: Done

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+  vertex_t num_verts = graph_view.get_number_of_vertices();
+
+  rmm::device_uvector<vertex_t> cluster_v(num_verts, handle.get_stream());
+
+  raft::update_device(cluster_v.data(), cluster_h.data(), num_verts, handle.get_stream());
+
+  rmm::device_uvector<vertex_t> communities(0, handle.get_stream());
+  rmm::device_uvector<weight_t> community_in_weight_sums(0, handle.get_stream());
+
+  std::tie(communities, community_in_weight_sums) =
```
ChuckHastings, line 30: @seunghwak - Thanks.  This PR (when I use your bug fix PR with it) now compiles.

I added some additional print statements in the latest push so I can track the behavior and try and make this work.  I have added some annotations here based on what I see in these results.

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+    cugraph::experimental::transform_reduce_by_adj_matrix_col_key_e(
+      handle,
+      graph_view,
+      cluster_v.begin(),
+      cluster_v.begin(),
+      cluster_v.begin(),
+      [] __device__(auto r, auto c, auto w, auto rv, auto cv) {
+        printf("transform reduce (%d,%d,%g,%d,%d)\n", (int)r, (int)c, (float)w, (int)rv, (int)cv);
```
ChuckHastings, line 38: I see this transform reduce print statement 12 times, corresponding to the 12 edges in the graph.  This is as I expect.

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+        return w;
+      },
+      weight_t{0});
+
+  thrust::for_each(thrust::device,
```
ChuckHastings, line 43: This is a debug call to display the output from the `transform_reduce_by_adj_matrix_col_key_e` call above.

The output from this is:
```
keys = 1,1,3,3,0,1,0,1,0,3,0,3
values = 1,1,1,1,1,1,1,1,1,1,1,1
```

This demonstrates the transform (the keys reflect the current clustering).  However no reduction has been done at this point.  I would have expected that the output for vertex 0 (the only one affected in this particular example) would have two keys (1 and 3) and their corresponding values would have been added (so 2 and 2).

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+                     printf("values = ");
+                     comma[0] = (char)0;
+                     for (auto p = b_value; p != e_value; ++p) {
+                       printf("%s%g", comma, *p);
+                       comma[0] = ',';
+                     }
+                     printf("\n");
+                   });
+
+  if (graph_t::is_multi_gpu) {
```
ChuckHastings, line 66: This is skipped in my test, as I am single GPU.  However, isn't this going to have the same problem as my above example?  That is, if vertex u is distributed over multiple GPUs, and generates key/value pairs with the same key on different GPUs, this step will only migrate the key/value pairs to the correct GPU.  No reduction will take place here.

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+      auto src, auto neighbor_cluster, auto new_cluster_sum, auto src_info, auto a_new) {
+      auto old_cluster_sum = thrust::get<0>(src_info);
+      auto k_k             = thrust::get<1>(src_info);
+      auto a_old           = thrust::get<2>(src_info);
+
+      weight_t delta_modularity = 2 * (((new_cluster_sum - old_cluster_sum) / total_edge_weight) -
+                                       resolution * (a_new * k_k - a_old * k_k + k_k * k_k) /
+                                         (total_edge_weight * total_edge_weight));
+
+      printf("compute delta_modularity (%d, %d, %g, %g, %g, %g, %g)\n",
```
ChuckHastings, line 106: This print statement is never displayed in my output.  I would expect that this would be executed once for each reduced key/value pair in the communities/community_in_weight_sums.  In the current code, there are 12 key/value pairs (because of the reduction question I raised above), so I would expect this to appear.

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+             new_cluster_sum,
+             old_cluster_sum,
+             k_k,
+             a_old,
+             a_new);
+
+      return thrust::make_tuple(neighbor_cluster, delta_modularity);
+    },
+    [] __device__(auto p1, auto p2) {
+      printf("reducing (%d, %g) with (%d, %g)\n",
```
ChuckHastings, line 118: This print statement gets executed 7613 in the current code displaying junky output.

In this example, each vertex has 2 key/value pairs, so I would expect this to execute 5 times (once for each vertex).  Because of the reduction question above, vertex 0 could execute this multiple times, so perhaps 7 or 8 times.

cpp/tests/experimental/louvain_primitives.cu
seunghwak: This sounds like a bug, there should be no repetition in keys (the output size == # unique keys). Let me take a closer look.

cpp/tests/experimental/louvain_primitives.cu
seunghwak: The behavior above is clearly a bug. Values for the same keys are globally reduced. And a part of the final (key, value) pairs is returned in every GPU.

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+                   });
+
+  if (graph_t::is_multi_gpu) {
+    std::tie(communities, community_in_weight_sums, std::ignore) =
+      cugraph::experimental::sort_and_shuffle_kv_pairs(
+        handle.get_comms(),
+        communities.begin(),
+        communities.end(),
+        community_in_weight_sums.begin(),
+        [] __device__(auto) { return 0; },
```
seunghwak, line 73: Need to use ```detail::compute_gpu_id_from_vertex_t<vertex_t>{comm_size}``` instead of just returning 0.

cpp/tests/experimental/louvain_primitives.cu
seunghwak: The output should be keys: [0, 1, 3] values: [4, 4, 4].

and you can use https://github.com/rapidsai/raft/blob/branch-0.18/cpp/include/raft/cudart_utils.h#L252 instead of the custom for_each as a convenience utility for printing.

seunghwak: I skipped a necessary reduction step in the SG path... :-(

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: I haven't tested this.  I was speculating here.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: Isn't the objective here to do a reduce by key for each source vertex?  That's what I need.  I want the output to be:

vertex 0: { 1:2, 3:2 }
vertex 1: { 0:1, 1:1 }
vertex 2: { 0:1, 1:1 } 
vertex 3: { 0:1, 3:1 }
vertex 4: { 0:1, 3:1 }

(Using python dictionary representation of a map to describe the desired behavior).

cpp/tests/experimental/louvain_primitives.cu
```diff
@@ -0,0 +1,197 @@
...
+{
+  using vertex_t = int32_t;
+  using edge_t   = int32_t;
+  using weight_t = float;
+
+  weight_t resolution{1};
+
+  raft::handle_t handle{};
+
+  std::vector<vertex_t> src_h = {0, 0, 0, 0, 1, 1, 2, 2, 3, 4, 4, 4};
```
seunghwak, line 165: shouldn't ```std::vector<vertex_t> src_h = {0, 0, 0, 0, 1, 1, 2, 2, 3, 4, 4, 4};``` be ```std::vector<vertex_t> src_h = {0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4};```?

ChuckHastings: Yes

cpp/tests/experimental/louvain_primitives.cu
seunghwak: No, this doesn't.

This is added to compute something like Sigma_{in} in (https://en.wikipedia.org/wiki/Louvain_method); sum of all the weights of the links inside the community i. Community ID becomes a key and edge weights are values.

I think what you're looking for is ```copy_v_transform_reduce_key_aggregated_out_nbr```. This works per vertex, and for each vertex, first aggregate outgoing edges by key, and ```key_aggregated_edge_op``` is called with (src, key, sum_of_the_aggregated_edge_weights_for_the_key, src attributes, value for the key in the key, value pairs).

cpp/tests/experimental/louvain_primitives.cu
seunghwak: Yes, this should appear; pull the updates from PR1399.

You'll see

```
compute delta_modularity (0, 3, 2, 0, 1, 4.2039e-45, 4)
compute delta_modularity (1, 0, 1, 0, 1, 0, 4)
compute delta_modularity (2, 0, 1, 0, 1, 0, 4)
compute delta_modularity (3, 0, 1, 0, 1, 0, 4)
compute delta_modularity (3, 3, 1, 0, 1, 0, 4)
compute delta_modularity (4, 0, 1, 1.4013e-45, 1, 0, 4)
compute delta_modularity (4, 3, 1, 1.4013e-45, 1, 0, 4)
```

cpp/tests/experimental/louvain_primitives.cu
seunghwak: Follow the question I posted in the Slack thrust channel.

cpp/tests/experimental/louvain_primitives.cu
seunghwak: I got this ```output = (3, 0.263889),(1, 0.0972222),(1, 0.0972222),(3, 0.0972222),(3, 0.0972222)``` at the end and this sounds legitimate.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: I see.  I will evaluate this more closely.

cpp/tests/experimental/louvain_primitives.cu
ChuckHastings: With your bug fixes this seems to be working in a reasonable fashion now.  I did some experiments today to understand the exact behavior of the call.  I should be able to construct a complete implementation next week and try debugging that.

PR #1401, opened 2021-02-10
Add boost 1.0 license file.

seunghwak:
#1411 added code (to address #1329) that follows the BOOST 1.0 license and this PR adds the BOOST 1.0 license to cuGraph codebase.

PR #1411, opened 2021-02-16
Add R-mat generator

### LINKED ISSUE #1329: [FEA] Large scale RMAT graph generator
We need a distributed graph generator to facilitate MG testing and benchmarking. 
Data needs to be passed as dask cudf edge list.
It should be RMAT and comply with Graph500 Specs
Leveraging GPUs would be ideal.

seunghwak:
Close #1329 (with https://github.com/rapidsai/cugraph/pull/1401)


seunghwak:
Related to PR #1401

cpp/include/experimental/graph_generator.hpp
```diff
@@ -0,0 +1,81 @@
...
+#include <rmm/device_uvector.hpp>
+
+#include <cstdint>
+#include <tuple>
+
+namespace cugraph {
+namespace experimental {
+
+/**
+ * @brief generate an edge list for an R-mat graph.
```
afender, line 28: We should clarify in the description what's the behaviour regarding duplicates (multigraph)

cpp/include/experimental/graph_generator.hpp
```diff
@@ -0,0 +1,81 @@
...
+ * for additional details). a, b, c, d should be non-negative and a + b + c should be no larger
+ * than 1.0.
+ * @param b a, b, c, d (= 1.0 - (a + b + c)) in the R-mat graph generator (vist https://graph500.org
+ * for additional details). a, b, c, d should be non-negative and a + b + c should be no larger
+ * than 1.0.
+ * @param c a, b, c, d (= 1.0 - (a + b + c)) in the R-mat graph generator (vist https://graph500.org
+ * for additional details). a, b, c, d should be non-negative and a + b + c should be no larger
+ * than 1.0.
+ * @param seed Seed value for the random number generator.
+ * @param clip_and_flip Flag controlling whether to generate edges only in the lower triangular part
```
afender, line 58: The way I understand this is that when `clip_and_flip` is false this returns a directed graph. When it is true it can be seen as an undirected one but not the kind of format we use in cugraph. 

I think we would benefit from exposing an option that generates the undirected graph inputs that we expect in cugraph.

afender (commented):
Great! It is really concise and nicely done. Did you get some info on performance/scalability while you were developing it?

I think next we want to have bindings for this so @jnke2016 / @rlratzel can leverage this for the python test suite and potentially get rid of the mtx/csv files. We should open a separate issue for this that captures the requirements from the test suite perspective.

cpp/include/experimental/graph_generator.hpp
seunghwak: Yes, that is correct, and my plan is to symmetrize the output after this R-mat generator; so to keep the R-mat generator's behavior close to Graph 500. Any thoughts?

seunghwak:
> Great! It is really concise and nicely done. Did you get some info on performance/scalability while you were developing it?
> 

It took 3-4 seconds to fill my 32 GB GPU memory on GV100; the code may still have some room to performance tune... but my guess is that this is fast enough for most practical cases.

I haven't tested this on multi-GPUs, but basically every GPU independently generates edges, I assume this will scale very well (actual scaling issue will happen later when we generate a graph from edge list).

cpp/include/experimental/graph_generator.hpp
seunghwak: Done.

BradReesWork (approved):

cpp/include/experimental/graph_generator.hpp
afender: I am fine with doing symmetrize separately, I just recall that the symmetrize step was done at the python level with cudf (concat + group_by) before. Do we have that feature at C++ level now? If not, we would need to add it before we can use this for C++ testing for instance.

afender (approved):

cpp/include/experimental/graph_generator.hpp
seunghwak: I don't think we have the feature yet. I think symmetrize, transpose, and triangle counting are related problems and can be addressed together.

cpp/src/experimental/scramble.cuh
```diff
@@ -0,0 +1,88 @@
...
+template <typename vertex_t>
+__device__ std::enable_if_t<sizeof(vertex_t) == 8, vertex_t> scramble(vertex_t value, size_t lgN)
+{
+  assert(std::is_unsigned<vertex_t>::value || lgN < 64);
+  assert(value >= 0);
+
+  constexpr uint64_t scramble_value0{606610977102444280};    // randomly generated
+  constexpr uint64_t scramble_value1{11680327234415193037};  // randomly generated
+
+  auto v = static_cast<uint64_t>(value);
```
aschaffer, line 41: The common part of these 3 specialization could be factored out in a template function; e.g.,
```
template<typename precision_t, typename brev_wrapper_t>
__device__ decltype(auto) generate_v(precision_t const& value, 
 precision_t const& scramble0,
 precision_t const& scramble1, 
 precision_t const& mask0,
 precision_t const& mask1,
 size_t lgN,
 brev_wrapper_t brev)
{
  auto v = static_cast<precision_t>(value);
  v += scramble 0 + scramble 1;
  v *= (scramble0 | mask0);
  v = brev(v) >> (8*sizeof(precision_t) - lgN);
  v *= (scramble1| mask1);
  v = brev(v) >> (8*sizeof(precision_t) - lgN);
 return v;
}
``` 
Then in each specialization initialize the scrambles, masks, and the appropriate `brev` lambda; and pass all to `generate_v(...)`. Something like that. That way the underlying logic is factored out and only moving parts are kept separate.

aschaffer (changes requested):

cpp/src/experimental/scramble.cuh
aschaffer: And `brev` can be constructed in-place as a `__device__` generic lambda (if possible... I think there might be some issues with `nvcc` compiling generic device lambdas). But, if possible, something like:
```
auto brev = [] __device__ (auto v){
 return __brevll(v);
};
//...
auto v = generate_v(..., brev);
```
Obviously, different specializations of `scramble(...)` would instantiate different versions of `brev`, accordingly, etc.

seunghwak: OK, thanks for pointing this out, and I refactored the code to remove redundancy.

aschaffer (approved):

BradReesWork:
@gpucibot merge

Issue #1410, opened 2021-02-16
[BUG] traveling_salesperson route returned does not include all vertices


SchultzC:
**Describe the bug**
cugraph 0.18 traveling_salesperson returns a route and a cost.  The cost appears correct but the route does not show all vertex ids in the output.

**Steps/Code to reproduce bug**
```python
sample_points = cudf.DataFrame({"vertex":[101,102,103,104],"x":[1,2,3,4],"y":[1,2,3,4]})
print(sample_points)
route, cost = cugraph.traversal.traveling_salesperson(sample_points, restarts=10, beam_search=False, nstart=101, k=1, verbose=True)
print("Route: \n{}".format(route))
print("\nCost:\n{}".format(cost))
```

**Expected behavior**
All vertex ids should be included in the route output.

**Environment overview (please complete the following information)**
 - Environment location: Docker
 - Method of cuGraph install: Docker
   - If method of install is [Docker], provide `docker pull` & `docker run` commands used
docker pull rapidsai/rapidsai-core-nightly:0.18-cuda11.0-runtime-ubuntu18.04-py3.8
docker run --gpus all --rm -it -p 8888:8888 -p 8787:8787 -p 8786:8786 \
    rapidsai/rapidsai-core-nightly:0.18-cuda11.0-runtime-ubuntu18.04-py3.8

**Environment details**
<details><summary>Click here to see environment details</summary><pre>
     
     **git***
     commit 27430205b703b35d17bd17810212a981a32a790d (grafted, HEAD, tag: branch-0.18-latest, origin/branch-0.18, origin/HEAD, branch-0.18)
     Author: Hugo Linsenmaier <hlinsenmaier@gmail.com>
     Date:   Thu Feb 11 06:17:38 2021 -0800
     
     Add SG TSP (#1360)
     
     This PR implements an approximated solution to the Traveling Salesperson Problem (TSP).
     The algorithm is exposed under ```traversal``` through a Python API taking 2D pos as input and returning a route.
     This PR relies on RAFT KNN: https://github.com/rapidsai/raft/pull/126
     Solves: https://github.com/rapidsai/cugraph/issues/1185
     
     Authors:
     - Hugo Linsenmaier (@hlinsen)
     
     Approvers:
     - AJ Schmidt (@ajschmidt8)
     - Brad Rees (@BradReesWork)
     - Rick Ratzel (@rlratzel)
     - Alex Fender (@afender)
     - Andrei Schaffer (@aschaffer)
     
[...]

PR #1412, opened 2021-02-17
TSP fix route return

hlinsen:
The vertex list that was fed as input to TSP was of type `int64` which ended up being corrupted when passed down
to the cpp layer as `vtx_ptr`. I updated the wrapper to cast the vertices to `int32`. In addition, I fixed the handling of nstart in the wrapper which was assuming vertex ids were starting at 1.
Solves: https://github.com/rapidsai/cugraph/issues/1410


BradReesWork (approved):

ajschmidt8:

hlinsen:
rerun tests

BradReesWork:
@gpucibot merge

PR #1399, opened 2021-02-10
Adding new primitives: copy_v_transform_reduce_key_aggregated_out_nbr & transform_reduce_by_adj_matrix_row|col_key_e bug fixes

seunghwak:
bug fixes

PR #1423, opened 2021-02-23
Update Louvain to use new graph primitives and pattern accelerators

### LINKED ISSUE #1220: [ENH] MNMG Louvain code cleanup
PR #1172 created an MNMG Louvain implementation.

In order to achieve this implementation, some temporary APIs were exposed in the graph_view_t object.  These should be cleaned up.

ChuckHastings:
Implement the `update_by_delta_modularity` method using the new graph primitives and pattern accelerators.

This eliminates all of the custom MNMG implementation originally created for MNMG Louvain a few releases ago and replaces it with the new pattern accelerator and graph primitives that have been added in the last couple of releases.

This depends on the following PRs and should not be merged until after them:
* #1394 
* #1399 

closes #1220


cpp/src/experimental/louvain.cuh
```diff
@@ -671,549 +305,107 @@ class Louvain {
...
+      current_graph_view_,
+      thrust::make_zip_iterator(thrust::make_tuple(old_cluster_sum_v.begin(),
+                                                   d_src_vertex_weights_cache_,
+                                                   cluster_subtract_v.begin(),
+                                                   d_src_cluster_cache_)),
+
+      d_dst_cluster_cache_,
+      thrust::make_counting_iterator<vertex_t>(base_vertex_id_),
+      thrust::make_counting_iterator<vertex_t>(base_vertex_id_ + local_num_vertices_),
+      d_dst_cluster_weights_cache_,
```
seunghwak, line 334: ```copy_v_transform_reduce_key_aggregated_out_nbr``` requires ```map_key_first, map_key_last, map_value_first``` to be pre-shuffled (yes... this might be a bit confusing as relabel doesn't require pre-shuffling, so we may need to make this consistent, but this is the current requirement. ```transform_reduce_by_adj_matrix_row(or col)_key``` returns pre-shuffled key, value pairs, so this helps to avoid unnecessary shuffling.

You need to do something like the following, and pass ```keys.beign(), keys.end(), weights.begin()```

I will submit another PR having fixes in the graph primitives side. After these fixes, I am seeing a failure in thrust::sequence call (https://github.com/rapidsai/cugraph/blob/branch-0.19/cpp/src/experimental/louvain.cuh#L524).

```C++ exception with description: tabulate: failed to synchronize ... ```

```
    if (graph_view_t::is_multi_gpu) {
      rmm::device_uvector<vertex_t> keys(local_num_vertices_, handle_.get_stream());
      thrust::sequence(rmm::exec_policy(stream_)->on(stream_), keys.begin(), keys.end(), base_vertex_id_);
      rmm::device_uvector<weight_t> weights(local_num_vertices_, handle_.get_stream());
      thrust::copy(rmm::exec_policy(stream_)->on(stream_),d_dst_cluster_weights_cache_, d_dst_cluster_weights_cache_ + local_num_vertices_, weights.begin());

      auto const comm_size = handle_.get_comms().get_size();
      rmm::device_uvector<vertex_t> rx_keys(0, handle_.get_stream());
      rmm::device_uvector<weight_t> rx_weights(0, handle_.get_stream());
      auto pair_first = thrust::make_zip_iterator(thrust::make_tuple(keys.begin(), weights.begin()));
      std::forward_as_tuple(std::tie(rx_keys, rx_weights), std::ignore) = groupby_gpuid_and_shuffle_values(handle_.get_comms(), pair_first, pair_first + keys.size(), [key_func = cugraph::experimental::detail::compute_gpu_id_from_vertex_t<vertex_t>{comm_size}]__device__(auto val) { return key_func(thrust::get<0>(val)); }, handle_.get_stream());

      keys = std::move(rx_keys);
      weights = std::move(rx_weights);
    }
```

**Comment by seunghwak** [comment]
Did you merge with the latest cuGraph branch? I am seeing some updates I made in "Files Changed"

File: cpp/src/experimental/louvain.cuh (hunk 1 above)
Comment (ChuckHastings) [reply cpp/src/experimental/louvain.cuh]: Resolved.

**Comment by ChuckHastings** [comment]
> Did you merge with the latest cuGraph branch? I am seeing some updates I made in "Files Changed"

Just double checked.  I did merge them.

I had to make a couple of additional changes from my debugging.  There was a problem in `device_comm.cuh`, I suspect I'm the first code that actually compiles the `device_sendrecv_tuple_iterator_element_impl` code in this way.

I modified `generate_graph_from_edgelist` to allow us to skip renumbering.  1 level of Louvain is deterministic if the graphs are numbered the same.  This allows me to compare single GPU and multi GPU results to validate behavior of Louvain.  If renumbering is on there's no good way to compare results across runs.  The renumbering result is a function of the number of GPUs.

File: cpp/src/experimental/louvain.cuh [hunk 2]
```diff
@@ -15,28 +15,20 @@
...
-#include <compute_partition.cuh>
-#include <experimental/shuffle.cuh>
-#include <utilities/graph_utils.cuh>
-
-#include <raft/device_atomics.cuh>
-
 #include <experimental/graph_functions.hpp>
+
 #include <patterns/copy_to_adj_matrix_row_col.cuh>
 #include <patterns/copy_v_transform_reduce_in_out_nbr.cuh>
```
seunghwak, line 24: Do you still need this?

cpp/src/experimental/louvain.cuh
```diff
@@ -405,67 +60,35 @@ class Louvain {
...
-      vertex_weights_v_(graph_view.get_number_of_local_vertices()),
-      cluster_weights_v_(graph_view.get_number_of_local_vertices()),
-      number_of_vertices_(graph_view.get_number_of_local_vertices()),
+      cluster_keys_v_(graph_view.get_number_of_local_vertices(), handle.get_stream()),
+      cluster_weights_v_(graph_view.get_number_of_local_vertices(), handle.get_stream()),
+      vertex_weights_v_(graph_view.get_number_of_local_vertices(), handle.get_stream()),
+      src_vertex_weights_cache_v_(0, handle.get_stream()),
+      src_cluster_cache_v_(0, handle.get_stream()),
+      dst_cluster_cache_v_(0, handle.get_stream()),
       stream_(handle.get_stream())
```
seunghwak, line 69: What's the purpose of having both ```handle_``` and ```stream_``` as ```stream``` is just ```handle_.get_stream()```? This is like maintaining two copies of the same variable.

cpp/src/experimental/louvain.cuh
```diff
@@ -405,67 +60,35 @@ class Louvain {
       handle_(handle),
```
seunghwak, line 60: And are you copying a handle here?

cpp/src/experimental/louvain.cuh
```diff
@@ -405,67 +60,35 @@ class Louvain {
...
-      current_graph_view_.offsets(), local_num_rows_, src_indices_v_.data().get());
-
-    if (base_src_vertex_id_ > 0) {
-      thrust::transform(rmm::exec_policy(stream_)->on(stream_),
-                        src_indices_v_.begin(),
-                        src_indices_v_.end(),
-                        thrust::make_constant_iterator(base_src_vertex_id_),
-                        src_indices_v_.begin(),
-                        thrust::plus<vertex_t>());
+      rank_           = handle.get_comms().get_rank();
```
seunghwak, line 72: What's the purpose of creating a separate copy of rank_ here? This will add burden to remember that ```rank_``` and ```handle.get_comms().get_rank()``` are same; maintain those two variables in sync; and this will eat-up additional register/cache line/memory.

cpp/src/experimental/louvain.cuh
```diff
@@ -561,58 +188,94 @@ class Louvain {
...
+  template <typename T>
+  std::tuple<T *, T *> cache_vertex_properties(rmm::device_uvector<T> &input,
+                                               rmm::device_uvector<T> &src_cache_v,
+                                               rmm::device_uvector<T> &dst_cache_v)
+  {
+    auto src = cache_src_vertex_properties(input, src_cache_v);
+    auto dst = cache_dst_vertex_properties(input, dst_cache_v);
+
+    return std::make_tuple(src, dst);
+  }
```
seunghwak, line 264: Do we really need this function anymore? This may cut one line in the caller site but just add an additional layer of indirection to track.

cpp/src/experimental/louvain.cuh
```diff
@@ -629,591 +292,229 @@ class Louvain {
...
+        auto src_cluster      = thrust::get<3>(src_info);
+        auto a_old            = thrust::get<4>(src_info);
+
+        if (src_cluster == neighbor_cluster) new_cluster_sum -= cluster_subtract;
+
+        weight_t delta_modularity = 2 * (((new_cluster_sum - old_cluster_sum) / total_edge_weight) -
+                                         resolution * (a_new * k_k - a_old * k_k + k_k * k_k) /
+                                           (total_edge_weight * total_edge_weight));
+
+#if 0
```
seunghwak, line 449: Better delete temporary print code.

cpp/src/experimental/louvain.cuh
```diff
@@ -629,591 +292,229 @@ class Louvain {
...
+      next_cluster_v.begin(),
+      next_cluster_v.end(),
+      cugraph::experimental::get_dataframe_buffer_begin<thrust::tuple<vertex_t, weight_t>>(
+        output_buffer),
+      next_cluster_v.begin(),
+      [rank = rank_, up_down] __device__(vertex_t old_cluster, auto p) {
+        vertex_t new_cluster      = thrust::get<0>(p);
+        weight_t delta_modularity = thrust::get<1>(p);
+
+#if 0
```
seunghwak, line 490: Better delete temporary print code.

cpp/tests/community/louvain_mg_test.cpp
```diff
@@ -0,0 +1,189 @@
+/*
```
seunghwak, line 1: I recommend renaming this to mg_louvain_test.cpp as the executable name is MG_LOUVAIN_TEST.

cpp/tests/community/louvain_mg_test.cpp
```diff
@@ -0,0 +1,189 @@
...
+{
+  ASSERT_DOUBLE_EQ(modularity, sg_modularity);
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// Test param object. This defines the input and expected output for a test, and
+// will be instantiated as the parameter to the tests defined below using
+// INSTANTIATE_TEST_CASE_P()
+//
+struct Louvain_Testparams {
```
seunghwak, line 40: We're mostly using AnalyticsName_Usecase_t, what's the purpose of introducing a new naming convention?

seunghwak:
> I modified `generate_graph_from_edgelist` to allow us to skip renumbering. 1 level of Louvain is deterministic if the graphs are numbered the same. This allows me to compare single GPU and multi GPU results to validate behavior of Louvain. If renumbering is on there's no good way to compare results across runs. The renumbering result is a function of the number of GPUs.

You mean the outcome is dependent on vertex ordering?

ChuckHastings:
> > I modified `generate_graph_from_edgelist` to allow us to skip renumbering. 1 level of Louvain is deterministic if the graphs are numbered the same. This allows me to compare single GPU and multi GPU results to validate behavior of Louvain. If renumbering is on there's no good way to compare results across runs. The renumbering result is a function of the number of GPUs.
> 
> You mean the outcome is dependent on vertex ordering?

Yes.  The algorithm breaks ties by choosing the vertex with the smaller vertex id making it deterministic with consistent vertex ordering.

cpp/src/community/dendrogram.cuh
```diff
@@ -26,12 +26,14 @@ template <typename vertex_t>
 class Dendrogram {
  public:
   void add_level(vertex_t num_verts,
+                 vertex_t first_index,
```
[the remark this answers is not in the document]
ChuckHastings: Fixed

cpp/src/community/dendrogram.cuh
```diff
@@ -26,12 +26,14 @@ template <typename vertex_t>
...
                  cudaStream_t stream                 = 0,
                  rmm::mr::device_memory_resource *mr = rmm::mr::get_current_device_resource())
   {
     level_ptr_.push_back(
       std::make_unique<rmm::device_buffer>(num_verts * sizeof(vertex_t), stream, mr));
```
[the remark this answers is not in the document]
ChuckHastings: The code I copied this from used device_buffer...

Can we type erase and pass a device_uvector back to python?   This is currently C++ only, but I was intending that eventually the python interface would be able to fetch these values and store them in cudf columns.

cpp/src/experimental/louvain.cuh
ChuckHastings: The blank line?  Perhaps not.  It prevents clang-format from reordering includes across blocks, I felt like the patterns should be after the graph and functions definitions... but I haven't verified that it's required.

cpp/src/experimental/louvain.cuh
ChuckHastings: The code was adapted from a previous version before we had handle.  I can replace the stream_ references with handle_.get_stream()

cpp/src/experimental/louvain.cuh
ChuckHastings: handle_ is a const reference, this is just copying the reference.

cpp/src/experimental/louvain.cuh
seunghwak: No,  #include <patterns/copy_v_transform_reduce_in_out_nbr.cuh>

AFAIK, this was used to compute weight_sums which has been replaced by a graph_view_t class member function.

cpp/src/experimental/louvain.cuh
ChuckHastings: Deleted this.  In my original MNMG implementation this was called dozens of times.

cpp/src/experimental/louvain.cuh
ChuckHastings: Thanks.  Hadn't promoted this from draft yet, that was on my todo list.

cpp/src/experimental/louvain.cuh
seunghwak: I see.

cpp/src/experimental/louvain.cuh
ChuckHastings: Ah, yes.  You are correct.  I have deleted that.

ChuckHastings: Spoke too soon.  I actually still need it for a different purpose.

https://github.com/rapidsai/cugraph/blob/9e4421e743f7ddbbf64a1eefcfa46e3ef4846c53/cpp/src/experimental/louvain.cuh#L316

seunghwak:
> Yes. The algorithm breaks ties by choosing the vertex with the smaller vertex id making it deterministic with consistent vertex ordering.

This sounds against _**graph invariants**_ (https://en.wikipedia.org/wiki/Graph_property) we may try to preserve. Isn't there a better way to break a tie without violating _graph invariants_? (yeah... this sounds impossible; if there are two structurally equivalent communities to move in... no way to distinguish the two without relying on labeling...)

What's the purpose of providing determinism? Just for testing or there is a bigger reason?

cpp/src/community/dendrogram.cuh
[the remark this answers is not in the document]
seunghwak: @Iroy30 may be able to confirm, but I guess this is possible.

https://github.com/rapidsai/cugraph/blob/branch-0.19/cpp/include/utilities/cython.hpp#L21

rmm::device_uvector is widely used in cython.hpp.

ChuckHastings:
> > Yes. The algorithm breaks ties by choosing the vertex with the smaller vertex id making it deterministic with consistent vertex ordering.
> 
> This sounds against _**graph invariants**_ (https://en.wikipedia.org/wiki/Graph_property) we may try to preserve. Isn't there a better way to break a tie without violating _graph invariants_? (yeah... this sounds impossible; if there are two structurally equivalent communities to move in... no way to distinguish the two without relying on labeling...)
> 
> What's the purpose of providing determinism? Just for testing or there is a bigger reason?

Ease of testing.  Request from cuML for it to be deterministic.

Note that Louvain approximates optimizing modularity, exact solution is NP hard.  The serial implementation of Louvain randomly orders the vertices before each pass.  Consequently it randomly picks one of clusters when it encounters a tie.  The consequence of this is that every time you run Louvain you get a different approximation of the answer.

seunghwak:
> Ease of testing. Request from cuML for it to be deterministic.

Let me think more but I am little worried about the current testing approach as this bypasses all the optimizations relying on renumbering; and leaves those optimizations untested.

The current best approach I can think of is to 1) create a renumbered multi_gpu graph, 2) and using the renumber_map from here, renumber edge list to be used in creating a single_gpu graph, 3) and create a single_gpu graph with renumbering disabled.

The results will be deterministic and SG & MG results will coincide, and we can test optimization disabled path of SG implementation with optimization enabled path of MG implementation, so this will have a better coverage.

ChuckHastings:
> > Ease of testing. Request from cuML for it to be deterministic.
> 
> Let me think more but I am little worried about the current testing approach as this bypasses all the optimizations relying on renumbering; and leaves those optimizations untested.
> 
> The current best approach I can think of is to 1) create a renumbered multi_gpu graph, 2) and using the renumber_map from here, renumber edge list to be used in creating a single_gpu graph, 3) and create a single_gpu graph with renumbering disabled.
> 
> The results will be deterministic and SG & MG results will coincide, and we can test optimization disabled path of SG implementation with optimization enabled path of MG implementation, so this will have a better coverage.

Each renumbering pass in the SG and MG code would diverge.  The results would be deterministic within a dendrogram level.  But as soon as we coarsen the graph, the SG coarsen and the MG coarsen (with num GPUs > 1) will create different graphs and the results will diverge.

I had thought a bit about the fact that we're not testing the renumbering phase.  We can certainly create some tests to specifically test the graph coarsening logic so that we're sure that's tested independently of the Louvain testing.  But I can't come up with a good way to test that.

cpp/tests/community/louvain_mg_test.cpp
ChuckHastings: Changed Testparams to Usecase

seunghwak:
> > > Ease of testing. Request from cuML for it to be deterministic.
> > 
> > 
> > Let me think more but I am little worried about the current testing approach as this bypasses all the optimizations relying on renumbering; and leaves those optimizations untested.
> > The current best approach I can think of is to 1) create a renumbered multi_gpu graph, 2) and using the renumber_map from here, renumber edge list to be used in creating a single_gpu graph, 3) and create a single_gpu graph with renumbering disabled.
> > The results will be deterministic and SG & MG results will coincide, and we can test optimization disabled path of SG implementation with optimization enabled path of MG implementation, so this will have a better coverage.
> 
> Each renumbering pass in the SG and MG code would diverge. The results would be deterministic within a dendrogram level. But as soon as we coarsen the graph, the SG coarsen and the MG coarsen (with num GPUs > 1) will create different graphs and the results will diverge.
> 
> I had thought a bit about the fact that we're not testing the renumbering phase. We can certainly create some tests to specifically test the graph coarsening logic so that we're sure that's tested independently of the Louvain testing. But I can't come up with a good way to test that.

So, there are two issues.

1) Testing with max_levels = 1
2) Testing when there are more than one levels (so graph coarsening will be called).

In case of 2), neither the proposed approach or the current approach will allow directly comparing SG & MG results.
And we're currently testing only the case 1). In the case of 1), both

1-1) completely skipping renumbering
and
1-2) renumbering edges using the renumber maps from the MG path before creating an SG graph

will work and allow directly comparing SG & MG results.

Are these correct? Assuming this, I think 1-2) has a better coverage than 1-1).

cpp/src/experimental/louvain.cuh
ChuckHastings: Removed

cpp/tests/community/louvain_mg_test.cpp
ChuckHastings: Fixed

cpp/src/community/dendrogram.cuh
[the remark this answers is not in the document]
ChuckHastings: Looks like I can convert a device_uvector into a device_buffer when it's needed in python:

https://github.com/rapidsai/cugraph/blob/7256f329773afe93deecd43f3cea3eceefc7fed1/cpp/include/utilities/cython.hpp#L200

I'll pursue this change.

afender (commented):
Lots of great stuff, I posted some nonblocking comments on API and testing.

aschaffer (approved):

cpp/include/algorithms.hpp
```diff
@@ -14,10 +14,14 @@
  * limitations under the License.
  */
 #pragma once
+
+#include <dendrogram.hpp>
```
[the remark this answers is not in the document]
ChuckHastings: I was hoping to add bindings for a louvain that returns a Dendrogram, eventually.  I had to make the C++ change to support MNMG testing which needed the Dendrogram exposed.

cpp/tests/community/mg_louvain_test.cpp
```diff
@@ -0,0 +1,233 @@
...
+////////////////////////////////////////////////////////////////////////////////
+TEST_P(Louvain_MG_Testfixture, CheckInt32Int32Float)
+{
+  run_test<int32_t, int32_t, float>(GetParam());
+}
+
+INSTANTIATE_TEST_CASE_P(
+  simple_test,
+  Louvain_MG_Testfixture,
+  ::testing::Values(Louvain_Usecase("test/datasets/karate.mtx", true, 100, 1)
```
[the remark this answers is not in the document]
ChuckHastings: Comparing the output should work for more than Karate.  I didn't have enough time to verify other test cases and adding RMAT.  I was hoping to expand later.

cpp/src/community/louvain.cu
```diff
@@ -26,50 +26,28 @@ namespace cugraph {
...
 
 template <typename vertex_t, typename edge_t, typename weight_t>
-std::pair<size_t, weight_t> louvain(raft::handle_t const &handle,
-                                    GraphCSRView<vertex_t, edge_t, weight_t> const &graph_view,
-                                    vertex_t *clustering,
-                                    size_t max_level,
-                                    weight_t resolution)
+std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> louvain(
+  raft::handle_t const &handle,
+  GraphCSRView<vertex_t, edge_t, weight_t> const &graph_view,
```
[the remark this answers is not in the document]
ChuckHastings: Yes.  The legacy version is still faster on SG than the new version.

cpp/include/algorithms.hpp
```diff
@@ -637,6 +641,66 @@ std::pair<size_t, typename graph_t::weight_type> louvain(
...
+ * @tparam     graph_t               Type of graph
+ *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
+ * @param[in]  graph                 input graph object
+ * @param[in]  dendrogram            input dendrogram object
+ * @param[out] clustering            Pointer to device array where the clustering should be stored
+ *
+ */
+template <typename graph_t>
+void flatten_dendrogram(raft::handle_t const &handle,
```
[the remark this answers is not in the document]
ChuckHastings: Current python APIs create columns and send the pointer to the data into the C++ when the data size is known a priori.  I was following that pattern.

cpp/include/algorithms.hpp
```diff
@@ -637,6 +641,66 @@ std::pair<size_t, typename graph_t::weight_type> louvain(
...
+ *
+ *    VD Blondel, J-L Guillaume, R Lambiotte and E Lefebvre: Fast unfolding of
+ *    community hierarchies in large networks, J Stat Mech P10008 (2008),
+ *    http://arxiv.org/abs/0803.0476
+ *
+ * @throws     cugraph::logic_error when an error occurs.
+ *
+ * @tparam     graph_t               Type of graph
+ *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
```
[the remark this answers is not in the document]
ChuckHastings: Removed extraneous text from copy/paste.

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
afender: Fair point, this made me realize that all the feature I did in the past feature had unknown return size :)

cpp/tests/community/mg_louvain_test.cpp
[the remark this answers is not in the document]
afender: Sounds good, yeah this can go in the next PR along with python stuff etc.

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
afender: I like returning a dendrogram too (probably optionally). 

Did you take a look at SLHC btw? It should return a dendrogram (might be in RAFT already). Better be consistent as we consider adding SLHC to cuGraph also.

cpp/src/community/louvain.cu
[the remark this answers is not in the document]
seunghwak: So, are we using the legacy graph for this? or are we using the legacy version for SG and the new version for MG?

cpp/tests/utilities/generate_graph_from_edgelist.cu
```diff
@@ -30,39 +30,47 @@ namespace test {
...
-  auto& comm               = handle.get_comms();
-  auto const comm_size     = comm.get_size();
-  auto const comm_rank     = comm.get_rank();
-  auto& row_comm           = handle.get_subcomm(cugraph::partition_2d::key_naming_t().row_name());
-  auto const row_comm_size = row_comm.get_size();
-  auto& col_comm           = handle.get_subcomm(cugraph::partition_2d::key_naming_t().col_name());
-  auto const col_comm_size = col_comm.get_size();
-
-  vertex_t number_of_vertices = static_cast<vertex_t>(vertices.size());
+template <typename vertex_t>
```
seunghwak, line 33: Do you still need the changes in this file after the test updates?

cpp/src/community/louvain.cu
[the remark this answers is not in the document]
ChuckHastings: The C++ supports:
* Legacy graph for SG
* new graph for SG
* new graph for MG

The python is currently still using legacy graph for SG and new graph for MG.  I'm doing some performance analysis now of the new graph for SG vice the legacy graph for SG.  It might be necessary to get some optimizations in place in order for the new graph to be fast enough.  My last run today (haven't pushed the latest mods yet, but will shortly) had `web-Google.mtx` running in 14s on the legacy implementation and 56s on the new implementation.  `ljournal-2008.mtx` runs in 250s on legacy and crashes (OOM) on the new implementation.

cpp/tests/utilities/generate_graph_from_edgelist.cu
ChuckHastings: Not necessarily, I will verify this.  It does successfully allow supporting of generating a multi-gpu graph without renumbering - in case we find another need for that.

That can be useful for debugging algorithms (knowing the numbering a priori and that it remains consistent across different shape MG tests).

cpp/src/community/louvain.cu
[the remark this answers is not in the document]
seunghwak: Oops, ljournal is not large at all and if you see OOM with ljournal, I guess something really shouldn't happen is happening in this case; something we should take a look (maybe not in this PR).

cpp/tests/utilities/generate_graph_from_edgelist.cu
seunghwak: I changed this file significantly in PR https://github.com/rapidsai/cugraph/pull/1443, and this is not necessary for testing, we may better undo this change for now.

cpp/src/community/louvain.cu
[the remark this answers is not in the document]
seunghwak: I may dig deeper to both performance & memory footprint when I tune primitives for Louvain. The execution time and memory consumption sound very bad.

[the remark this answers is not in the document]
seunghwak: We should be able to process 1B-ish edge graph using one GPU. IIRC, ljournal is much smaller than that.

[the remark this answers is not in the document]
ChuckHastings: Agreed, that should probably be outside of this PR.  But that's why I've left the legacy implementation in place and the default for python.  When you start tuning these we can evaluate when to delete the legacy implementation.

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Corey and I talked about the dendrogram and potentially moving it to raft.  We thought waiting until the Louvain work was done in cugraph would be prudent.

cpp/tests/utilities/generate_graph_from_edgelist.cu
ChuckHastings: I will revert tomorrow.

ChuckHastings: Done

seunghwak (approved):
Thanks!

cpp/tests/community/mg_louvain_helper.hpp
```diff
@@ -0,0 +1,53 @@
+/*
```
rlratzel, line 1: Are `mg_louvain_helper.[cu|hpp]` really specific to louvain? I'm just wondering if they're generally useful, should they be renamed and placed in the `cugraphtestutil` lib?

cpp/tests/community/mg_louvain_test.cpp
```diff
@@ -0,0 +1,233 @@
...
+// will be instantiated as the parameter to the tests defined below using
+// INSTANTIATE_TEST_CASE_P()
+//
+struct Louvain_Usecase {
+  std::string graph_file_full_path{};
+  bool weighted{false};
+  size_t max_level;
+  double resolution;
+
+  // FIXME:  We really should have a Graph_Testparms_Base class or something
```
rlratzel, line 51: This is a good `FIXME`, I agree.

rlratzel (approved):
One minor doc question and a couple of comments, but nothing to hold up my approval now.

cpp/tests/community/mg_louvain_helper.hpp
ChuckHastings: They are currently specific to louvain.  I think they might be more generally useful.  But I wanted to have at least one more unit test that used it before moving into a more general location.

cpp/include/algorithms.hpp
```diff
@@ -637,6 +641,66 @@ std::pair<size_t, typename graph_t::weight_type> louvain(
...
+ * a graph.  This function will flatten the hierarchical clustering into
+ * a label for each vertex representing the final cluster/partition to
+ * which it is assigned
+ *
+ * @throws     cugraph::logic_error when an error occurs.
+ *
+ * @tparam     graph_t               Type of graph
+ *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
+ * @param[in]  graph                 input graph object
```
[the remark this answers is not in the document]
ChuckHastings: It's a bit more pervasive.  They are all graph views.  In fact all of our algorithms are using graph views and calling them graphs.

I'm testing a change that changes `louvain` calls to consistently call it a graph_view.  I believe we should do this with all algorithms (I'll try and make this change as I rework things to use the new graph primitives).

cpp/tests/community/mg_louvain_helper.hpp
seunghwak: And FYI,

renumber_ext_vertices
https://github.com/rapidsai/cugraph/pull/1443/files#diff-4106792213a4762c10634a44ac8c92f194994ec0b595dba4eccaa822e8353978R215

unrenumber_local_int_vertices
https://github.com/rapidsai/cugraph/pull/1443/files#diff-4106792213a4762c10634a44ac8c92f194994ec0b595dba4eccaa822e8353978R245

and unrenumber_local_int_vertices
https://github.com/rapidsai/cugraph/pull/1443/files#diff-4106792213a4762c10634a44ac8c92f194994ec0b595dba4eccaa822e8353978R279

will be added once PR 1443 gets merged.

There will be some overlaps and some functions here may become obsolete.

afender (approved):

BradReesWork:
@gpucibot merged

BradReesWork:
@gpucibot merge

PR #1426, opened 2021-02-24
See if splitting louvain up makes CI run faster

ChuckHastings:


ChuckHastings:
No notable improvement.

PR #1427, opened 2021-02-25
[REVIEW] Update and Test Renumber bindings

Iroy30:


cpp/include/utilities/cython.hpp
```diff
@@ -191,7 +191,7 @@ struct major_minor_weights_t {
 template <typename vertex_t, typename edge_t>
 struct renum_quad_t {
   explicit renum_quad_t(raft::handle_t const& handle)
-    : dv_(0, handle.get_stream()), part_(std::vector<vertex_t>(), false, 0, 0, 0, 0)
+    : dv_(0, handle.get_stream()), part_(std::vector<vertex_t>{0}, false, 0, 0, 0, 0)
```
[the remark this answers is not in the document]
seunghwak: And

```
  vertex_t nv_;
  edge_t ne_;
```
(private member variables of renum_quad_t) should better be 
```
  vertex_t nv_{}; or vertex_t nv_{0};
  edge_t ne_{}; or edge_t ne_{0};
```

Better not leave class member variables in uninitialized states.

python/cugraph/structure/new_number_map.py
```diff
@@ -0,0 +1,275 @@
...
+                self.df.merge(tmp_df, on=self.col_names, how="right")
+                .sort_values(index_name)
+                .drop(columns=[index_name])
+                .reset_index()["id"]
+            )
+
+        def from_internal_vertex_id(
+            self, df, internal_column_name, external_column_names
+        ):
+            tmp_df = self.df.merge(
```
seunghwak, line 66: Have you checked the performance of this?

From internal to external vertex ID, it will simply be external_vertex_ID = renumber_map[internal_vertex_ID] in SG. Should be very fast. I guess this merge can be much slower.

python/cugraph/structure/new_number_map.py
```diff
@@ -0,0 +1,275 @@
...
+            return self.ddf.merge(
+                ddf,
+                right_on=col_names,
+                left_on=self.col_names,
+                how="right",
+            )["global_id"]
+
+        def from_internal_vertex_id(
+            self, df, internal_column_name, external_column_names
+        ):
```
seunghwak, line 103: I think this can be much faster. External vertex id = local_map[internal vertex id - base vertex ID for this GPU];

python/script.py
```diff
@@ -0,0 +1,11 @@
+import pytest
```
seunghwak, line 1: I assume this file is for temporary testing and will be deleted.

Iroy30: Yes

python/cugraph/structure/new_number_map.py
Iroy30: This takes into account multi column unrenumbering

seunghwak: I see, and isn't this unnecessarily expensive in single column cases? We may take a cheaper approach if the original vertex IDs are single-column?

cpp/include/utilities/cython.hpp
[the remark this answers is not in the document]
Iroy30: updated

python/cugraph/structure/new_number_map.py
Iroy30: Will be addressed as a part of optimization in subsequent PRs

python/cugraph/structure/new_number_map.py
Iroy30: This can be addressed as part of optimization in subsequent PR

seunghwak (approved):
Thanks and don't forget to create an issue about optimizing the un-renumbering step.

aschaffer (approved):

BradReesWork:
@gpucibot merge

Issue #1407, opened 2021-02-12
[FIX] Remove duplication of RAFT python bindings


afender:
Raft cython/python features are implemented and deployed from RAFT (see https://github.com/rapidsai/raft/blob/branch-0.18/BUILD.md#pythoncython-integration)

cuGraph should not add or duplicate bindings as seen for the handle: https://github.com/rapidsai/cugraph/blob/27430205b703b35d17bd17810212a981a32a790d/python/cugraph/structure/graph_primtypes.pxd#L28

This defeats the purpose of RAFT and creates room for conflicts. Notice that for the handle, Raft provides a wrapper for 7 member function while cuGraph added only one. 

Task list:
- [ ] Get rid of the  duplicate handle binding
- [ ] Update files to import from raft : ` from cugraph.raft.common.handle cimport *`
- [ ] Check for other duplicate and update
- [ ] Move generic features like `move` outside of the graph_prims.px* after checking this is not already in raft

PR #1435, opened 2021-03-03
Streams infra + support in egonet 

### LINKED ISSUE #957: [ENH] Streams support in cugraph
RAFT handle #815 comes with streams in it. 
We should consider supporting these streams in cuGraph.

afender:
- Stream synchronization behavior: switched to per-thread default stream instead of the legacy default stream
- Update raft tag
- EgoNet upgrade to use `uvector` instead of `device_vector`
- EgoNet upgrade to execute on a different stream for each seed
- Perf analysis timers/app for EgoNet

Concurrency is limited by the number of available blocks on the device. Thrust-based codes may request a lot of blocks without a way to control this. In practice, smaller graphs leverage concurrency better than larger ones where tasks may end up waiting for available resources.

We may wait on #1407 before reconciling and merging this
Close #957


cpp/src/community/egonet.cu
```diff
@@ -63,56 +68,106 @@ extract(
 {
   auto v           = csr_view.get_number_of_vertices();
   auto e           = csr_view.get_number_of_edges();
-  auto stream      = handle.get_stream();
+  auto user_stream = handle.get_stream_view();
```
seunghwak, line 71: Sorry for nitpicking, but isn't this a misnomer? Better be user_stream_view or user_sv or any other name indicating this is a view object?

cpp/src/community/egonet.cu
```diff
@@ -63,56 +68,106 @@ extract(
...
   float avg_degree = e / v;
   rmm::device_vector<size_t> neighbors_offsets(n_subgraphs + 1);
   rmm::device_vector<vertex_t> neighbors;
 
-  // It is the right thing to accept device memory for source_vertex
-  // FIXME consider adding a device API to BFS (ie. accept source on the device)
   std::vector<vertex_t> h_source_vertex(n_subgraphs);
-  raft::update_host(&h_source_vertex[0], source_vertex, n_subgraphs, stream);
+  std::vector<size_t> neighbors_offsets_h(n_subgraphs + 1);
```
seunghwak, line 77: Better not mix h_variable_name (h_source_vertex) & variabe_name_h (neighbor_offsets_h) at least within a same function body.

cpp/src/community/egonet.cu
```diff
@@ -63,56 +68,106 @@ extract(
...
+
+  raft::update_host(&h_source_vertex[0], source_vertex, n_subgraphs, user_stream.value());
```
seunghwak, line 79: I think you need to call ```user_stream.synchronize()```.

There is no stream synchronization between here and bfs call where h_source_vertex[i] is used; so no guarantee that h_source_vertex[i] will have a valid value.

cpp/tests/community/egonet_test.cu
```diff
@@ -118,7 +126,7 @@ class Tests_InducedEgo : public ::testing::TestWithParam<InducedEgo_Usecase> {
     ASSERT_TRUE(h_cugraph_ego_edge_offsets[configuration.ego_sources.size()] ==
                 d_ego_edgelist_src.size());
     for (size_t i = 0; i < configuration.ego_sources.size(); i++)
-      ASSERT_TRUE(h_cugraph_ego_edge_offsets[i] < h_cugraph_ego_edge_offsets[i + 1]);
+      ASSERT_TRUE(h_cugraph_ego_edge_offsets[i] <= h_cugraph_ego_edge_offsets[i + 1]);
     auto n_vertices = graph_view.get_number_of_vertices();
     for (size_t i = 0; i < d_ego_edgelist_src.size(); i++) {
       ASSERT_TRUE(h_cugraph_ego_edgelist_src[i] >= 0);
```
seunghwak, line 132: Better use https://github.com/rapidsai/cugraph/blob/branch-0.19/cpp/tests/utilities/test_utilities.hpp#L134 as this will generate a warning if vertex_t is unsigned.

seunghwak (commented):
LGTM except for few small complaints.

cpp/src/community/egonet.cu
afender: Right, I think it was passing because the user stream was the blocking default stream but I think you're correct that it is unsafe in other cases.

BradReesWork:
rerun tests

cpp/tests/community/egonet_test.cu
afender: f81a2b4

cpp/src/community/egonet.cu
afender: f81a2b4

cpp/src/community/egonet.cu
afender: f81a2b4

cpp/src/community/egonet.cu
afender: f81a2b4

seunghwak (approved):

aschaffer (approved):

BradReesWork:
@gpucibot merge

PR #1441, opened 2021-03-07
Add additional datasets to improve coverage

### LINKED ISSUE #1214: Test with isolated vertices and self loops
While we have a disconnected dataset, most tests go through renumbering which removes singletons.
In practice, it is quite possible that some analytics still fail when such inputs are passed.

This could also be used for batch processing in some cases

We need to test and report the status of isolated vertices and self-loops support in cuGraph.

Related to #1206

jnke2016:
add datasets to test self-loops, string vertex IDs and isolated vertices
Update README
closes #1214


jnke2016:
One of the file (out of 6) I uploaded is above the 5242880 limit (the file is 5.3 MB). Should I just delete that file or include the file and update the download  script ?

BradReesWork (changes requested):
the large files need to be removed.   you can add them to the S3 store

datasets/README.md
```diff
@@ -16,6 +16,35 @@ This directory contains small public datasets in `mtx` and `csv` format used by
 
 **netscience** : The graph netscience contains a coauthorship network of scientists working on network theory and experiment, as compiled by M. Newman in May 2006.
 
+
+
+## Tests coverage
```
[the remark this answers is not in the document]
jnke2016: In this PR, I updated  `datasets/README.md` to reference and describe the new datasets. But after discussing with Rick, I instead modified the existing datasets (karate, dolphins) since they are much smaller, therefore easier to test. I will be uploading the new datasets in this PR

BradReesWork (changes requested):
large datasets need to be removed

BradReesWork (approved):

datasets/README.md
```diff
@@ -1,67 +1,125 @@
...
+
+## C++
+Cugraph's C++ analytics tests need larger datasets (>5GB uncompressed) and reference results (>125MB uncompressed). They can be downloaded by running the provided script from the `datasets` directory.
+```
+cd <repo>/datasets
+./get_test_data.sh
+```
+You may run this script from elsewhere and store C++ test input to another location.
+
+Before running the tests, you should let cuGraph know where to find the test input by using:
```
rlratzel, line 84: This is now used for both C++ and Python, so it should probably be moved outside of the C++ section.  You might just mention here that the C++ tests *require* it though.

datasets/README.md
```diff
@@ -1,67 +1,125 @@
...
+
+**karate_s_loop** : The graph "karate_s_loop" is a modified version of the  "karate" graph where self-loops were added
+
+**dolphins_s_loop** : The graph "dolphins_s_loop" is a modified version of the  "dolphin" graph where self-loops were added
+
+**karate_mod** : The graph "karate_mod" is a modified version of the  "karate" graph where vertices and edges were added
+
+**karate_str** : The graph "karate_str" contains the network of friendships between the 34 members of a karate club at a US university, as described by Wayne Zachary in 1977. The integer vertices were replaced by strings
+
+Larger datasets containing self-loops can be downloaded by running the provided script from `datasets` directory using the `--self_loops`
```
rlratzel, line 46: Are these larger datasets currently used by any tests or benchmarks?  If not, maybe you should mention that in order to take advantage of these additional datasets, the current tests or benchmarks will have to be edited to specify the new files.

datasets/README.md
```diff
@@ -1,67 +1,125 @@
...
+
+**dolphin** : The graph dolphins contains an undirected social network of frequent associations between 62 dolphins in a community living off Doubtful Sound, New Zealand, as compiled by Lusseau et al. (2003).
+
+**netscience** : The graph netscience contains a coauthorship network of scientists working on network theory and experiment, as compiled by M. Newman in May 2006.
+
+
+
+## Tests coverage 
+
+The datasets below was added to check for coverage where some of them include self-loops, string vertex IDs, isolated vertices and multi-edges
```
rlratzel, line 23: Minor suggestion for improving readability a bit: "The datasets below was added to check for coverage where some of them include self-loops, string vertex IDs, isolated vertices and multi-edges" -> "The datasets below were added to provide input that contains self-loops, string vertex IDs, isolated vertices, and multiple edges."

datasets/README.md
```diff
@@ -1,67 +1,125 @@
...
+
+**karate** : The graph "karate" contains the network of friendships between the 34 members of a karate club at a US university, as described by Wayne Zachary in 1977.
+
+**dolphin** : The graph dolphins contains an undirected social network of frequent associations between 62 dolphins in a community living off Doubtful Sound, New Zealand, as compiled by Lusseau et al. (2003).
+
+**netscience** : The graph netscience contains a coauthorship network of scientists working on network theory and experiment, as compiled by M. Newman in May 2006.
+
+
+
+## Tests coverage
```
rlratzel, line 21: I'm assuming this should be a subtopic of the "Python" section above?  If so, it should use '###'

Maybe also name it "Modified datasets" or something more descriptive?

rlratzel (changes requested):
Looks good, I just have a few minor requests.

datasets/README.md
rlratzel: Also, should this be a separate subsection?
```
### Additional Datasets
```
(or something descriptive like that)

rlratzel (approved):

BradReesWork:
@gpucibot merge

Issue #1442, opened 2021-03-08
[ENH] Graph primitives performance improvement


seunghwak:
**Describe the solution you'd like**
Performance optimize graph primitives.

In short term, we need to

- [ ] Update graph partitioning scheme to allow simpler/faster communication patterns and control trade-offs between memory footprint and parallelism
- [ ] Address thread-divergence issue when vertices have widely varying numbers of neighbors (e.g. Power-law graphs)
- [ ] Specialize update_frontier_v_push_if_out_nbr for pure function reduction operations (this will improve performance/memory footprint of BFS & SSSP)

PR #1447, opened 2021-03-11
Improve graph primitives performance on graphs with widely varying vertex degrees

seunghwak:
Partially addresses Issue #1442

Update graph primitives used by PageRank, Katz Centrality, BFS, and SSSP to launch 3 different kernels based on vertex degrees to address thread divergence issue. In addition, cut memory footprint of the VertexFrontier class used by BFS & SSSP.

The following highlights performance improvement with this optimization.

R-mat 2^25 vertices 2^25 * 32 edges
PageRank: 7.66, 7.42, 8.83, 8.83 seconds (the first two unweighted, the last two weighted, first & third without personalization)=> 1.07, 1.08, 1.36, 1.39 seconds
Katz: 1.08, 1.94 seconds (unweighted, weighted)=> 0.243, 0.275
BFS: 1.32 seconds=> 0.251
R-mat 2^25 vertices 2^25 * 16 edges
SSSP: 1.89 seconds (memory allocation fails with the edge factor of 32)=> 0.317

And now SSSP also works with 2^25 vertices 2^25 * 32 edges with the memory footprint improvement and it took 0.514 sec.

Still needs additional optimizations to reach the target performance

1. add BFS & SSSP specific optimizations (the current implementation assumes general reduction operations while BFS can pick any source vertex if a vertex is discovered by multiple source vertices and SSSP picks the one with the minimum edge weight, these pure function reduction operations allow additional optimizations).
2. Launch 3 different kernels in multiple streams to recover parallelism when the frontier size is relatively small (currently three kernels are queued in a single stream, and this leads to up to 3x decrease in parallelism)


afender (commented):
:tada:  lot of code that I didn't review yet, but do you have speedup numbers? It would be nice to attach that to the PR or the issue for the record.

seunghwak:
> 🎉 lot of code that I didn't review yet, but do you have speedup numbers? It would be nice to attach that to the PR or the issue for the record.

Yes, we do.

I have performance numbers before tuning,
R-mat 2^25 vertices 2^25 * 32 edges
PageRank: 7.66, 7.42, 8.83, 8.83 seconds (the first two unweighted, the last two weighted, first & third without personalization)
Katz: 1.08, 1.94 seconds (unweighted, weighted)
BFS: 1.32 seconds
R-mat 2^25 vertices 2^25 * 16 edges
SSSP: 1.89 seconds (memory allocation fails with the edge factor of 32)

and after this first step of performance tuning, I am getting 4x-8x speedup (close to 8x for PageRank & Katz, 4x-ish for SSSP, 5x-ish for BFS). And now we can run SSSP with the edge factor of 32. Now graph creation limits the maximum graph we can analyze for the C++ part for all four algorithms.

Still have multiple places to cut memory footprint and execution time and multiple untuned parameters ... but I think I will jump to those later when we are ready to run even larger graphs on larger systems.

seunghwak:
And due to https://github.com/NVIDIA/cuCollections/issues/71

To test with 2^25 vertices and 2^25 * 32 edges, you need to typecast blockDim.x to uint64_t in

https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/static_map_kernels.cuh#L199
https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/static_map_kernels.cuh#L217
https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/static_map_kernels.cuh#L260
https://github.com/NVIDIA/cuCollections/blob/dev/include/cuco/detail/static_map_kernels.cuh#L282

before compiling cuGraph, I am planning to fix the cuCollection bugs as the first 0.20 task.

cpp/include/patterns/copy_v_transform_reduce_in_out_nbr.cuh
```diff
@@ -42,23 +42,7 @@ namespace experimental {
...
-  lhs = plus_edge_op_result(lhs, rhs);
-}
-
-template <bool update_major, typename T>
-__device__ std::enable_if_t<!update_major, void> accumulate_edge_op_result(T& lhs, T const& rhs)
-{
-  atomic_add(&lhs, rhs);
-}
-#endif
+int32_t constexpr copy_v_transform_reduce_nbr_for_all_block_size = 512;
```
afender, line 45: Should this vary based on the average degree and graph size or do we think 512 is always optimal?

cpp/include/patterns/vertex_frontier.cuh
```diff
@@ -37,129 +36,74 @@
...
+      rmm::device_scalar<vertex_t> vertex(v, handle_ptr_->get_stream());
+      insert(vertex.data(), vertex_t{1});
+    } else {
+      elements_.resize(1, handle_ptr_->get_stream());
+      raft::update_device(elements_.data(), &v, size_t{1}, handle_ptr_->get_stream());
+    }
   }
 
-  size_t size() const { return size_; }
+  void insert(vertex_t const* sorted_unique_vertices, vertex_t num_sorted_unique_vertices)
```
afender, line 58: `sorted_unique_vertices` is assumed to be device memory right? Consider mentioning it or having a check for this.
It could be confusing because the other `insert` function takes host memory.

cpp/tests/experimental/katz_centrality_test.cpp
```diff
@@ -39,7 +40,7 @@
 // do the perf measurements
 // enabled by command line parameter s'--perf'
 //
-static int PERF = 0;
+static int PERF = 1;
```
afender, line 43: should this be 0 or 1 by default? I think we have the command line parameter `--perf` to control this.

afender (approved):

cpp/tests/experimental/katz_centrality_test.cpp
seunghwak: Yes, correct, this should be 0 by default, I will make a fix.

cpp/include/patterns/vertex_frontier.cuh
seunghwak: insert(vertex_t v) takes a scalar and otherwise explicitly mentioned, all pointers are device pointers. I will add a comment to additionally mark this requires a device memory pointer.

Eventually, I think we should adopt device_span and host_span from cuDF to avoid this kind of confusion.

cpp/include/patterns/copy_v_transform_reduce_in_out_nbr.cuh
seunghwak: I don't think it will be "always" optimal but at least reasonable for large graphs. If we wish to concurrently run multiple small graphs, a smaller block_size may possibly be beneficial. We may leave this as a future exhaustive tuning issue, but I don't think we're ready to jump to that level of exhaustive tuning at this moment.

cpp/include/patterns/copy_v_transform_reduce_in_out_nbr.cuh
```diff
@@ -416,60 +360,136 @@ void copy_v_transform_reduce_nbr(raft::handle_t const& handle,
...
-          <<<update_grid.num_blocks, update_grid.block_size, 0, handle.get_stream()>>>(
-            matrix_partition,
-            graph_view.get_vertex_partition_first(comm_root_rank),
-            graph_view.get_vertex_partition_last(comm_root_rank),
-            adj_matrix_row_value_input_first,
-            adj_matrix_col_value_input_first,
-            vertex_value_output_first,
-            e_op,
-            major_init);
+        if (GraphViewType::is_multi_gpu) {
```
ChuckHastings, line 407: A minor thought for readability.

Many of these 'if (GraphViewType::is_multi_gpu)` code blocks are identical except for where to put the result.

Would this read easier if the if statements assigned the buffer and there was only one call to `detail::for_all_major_for_all_nbr_mid_degree<update_major>`?

e.g.:
```
auto tmp_output_buffer = (GraphViewType::is_multi_gpu) ?
    (update_major ? major_buffer_first + segment_offests[1] : minor_buffer_first) :
    (vertex_value_output_first + (update_major ? segment_offsets[1] : vertex_t{0}));

detail::for_all_major_for_all_nbr_mid_degree<update_major>
  <<<update_grid.num_blocks, update_grid.block_size, 0, handle.get_stream()>>>(
    matrix_partition,
    matrix_partition.get_major_first() + segment_offsets[1],
    matrix_partition.get_major_first() + segment_offsets[2],
    adj_matrix_row_value_input_first + row_value_input_offset,
    adj_matrix_col_value_input_first + col_value_input_offset,
    tmp_output_buffer,
    e_op,
    major_init);
```

ChuckHastings (approved):

cpp/include/patterns/copy_v_transform_reduce_in_out_nbr.cuh
seunghwak: Thanks for the suggestion, but this requires ```major_buffer_first, minor_buffer_first``` and ```vertex_value_output_first``` to have the same type (e.g. vertex_value_output_first may include something like discard_iterator or permutation iterator); if not this code will not compile.

C++17 supports ```constexpr if``` (we might be switching to C++17 sometime soon) but AFAIK this doesn't work for ternary expressions (e.g. ```auto tmp_output_buffer = constexpr (GraphViewType::is_multi_gpu) ? buffer_a : buffer_b``` will not compile even with C++17).

Let me know if you're aware of the workaround for this problem and or any other concerns.

seunghwak: If we switch to C++17, it seems like we might be able to do this with the functor in between.

https://stackoverflow.com/questions/41011900/equivalent-ternary-operator-for-constexpr-if

```
auto get_output_buffer = [](auto buffer0, auto buffer1) {
  if constexpr (GraphViewType::is_multi_gpu) {
   return buffer0;
  }
  else {
    return buffer1;
  }
}
```

I will consider this update once we finally switch to C++17.

seunghwak: I added FIXME comments for this.

ChuckHastings: I think the code would be easier to understand and maintain with this change (when we can make it).  I would consider it a rather minor improvement however, I wouldn't prioritize it particularly high.

BradReesWork:
@gpucibot merge

BradReesWork (approved):

PR #1448, opened 2021-03-11
Updating docs

BradReesWork:
* added Hungarian to api.rst
* fixed missing examples
* fixed syntax

### DIFF
--- a/python/cugraph/dask/centrality/katz_centrality.py
+++ b/python/cugraph/dask/centrality/katz_centrality.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2019-2020, NVIDIA CORPORATION.
+# Copyright (c) 2019-2021, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/python/cugraph/dask/centrality/katz_centrality.py
+++ b/python/cugraph/dask/centrality/katz_centrality.py
@@ -115,7 +115,7 @@ def katz_centrality(input_graph,
     Examples
     --------
     >>> import cugraph.dask as dcg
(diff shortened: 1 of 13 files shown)


BradReesWork:
rerun tests

BradReesWork:
rerun tests

python/cugraph/dask/centrality/katz_centrality.py
```diff
@@ -115,7 +115,8 @@ def katz_centrality(input_graph,
     Examples
     --------
     >>> import cugraph.dask as dcg
-    >>> Comms.initialize(p2p=True)
+    >>> ... Init a DASK Cluster
+    >>    see https://docs.rapids.ai/api/cugraph/stable/dask-cugraph.html
```
rlratzel, line 119: minor: since these are typically copy-and-pasteable snippets, you could use a comment char here:
```
>>> import cugraph.dask as dcg
>>> # ... Init a DASK Cluster
>>> #   see https://docs.rapids.ai/api/cugraph/stable/dask-cugraph.html
>>> chunksize = dcg.get_chunksize(input_data_path)
```

rlratzel (approved):
LGTM. I had just one minor suggestion which need not hold up my approval.

BradReesWork:
rerun tests

python/cugraph/dask/centrality/katz_centrality.py
BradReesWork: true.  but it is not executable code.  There used to be a line initializing Comm, but that is not fully what is needed for DASK.  I pulled that out so that there is one place to maintain what setting up dask looks like.  If someone cut&pastes the above block, I would prefer it break on a non-executable line

afender (approved):
:+1:

BradReesWork:
@gpucibot merge

PR #1468, opened 2021-03-17
Removed unused dependencies from libcugraph recipe, moved non-test script code from test script to gpu build script

rlratzel:
* Removed unused dependencies from the `libcugraph` recipe.  This is motivated by the CuPy project to integrate `libcugraph` as the graph analytics backend with minimal extra dependencies ( https://github.com/cupy/cupy/issues/4219, https://github.com/cupy/cupy/issues/2431, https://github.com/cupy/cupy/pull/4054 )
* Moved non-test script code from test script to gpu build script. The `FIXME` addressed for this was added after discussing with @raydouglass earlier, and will allow any Project Flash failures to fail the build immediately instead of attempting to then run tests.
* Removed unused cudf lib reference from test cmake file.

Tested by doing a successful local `conda build` of the recipe.

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -90,8 +90,25 @@ conda list --show-channel-urls
 ################################################################################
 
 if [[ -z "$PROJECT_FLASH" || "$PROJECT_FLASH" == "0" ]]; then
-  gpuci_logger "Build from source"
-  $WORKSPACE/build.sh -v clean libcugraph cugraph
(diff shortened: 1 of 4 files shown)


BradReesWork (approved):

ChuckHastings (approved):

ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

PR #1469, opened 2021-03-17
MS BFS python APIs + EgoNet updates

afender:
There are various things in this PR.

Multi-Seed (MS) BFS: 
- API tentative 
- Saving research on memory feasability helper function (not in production)
- Saving research on running the current BFS concurrently with streams and threads for analysis perf comparison (not in production)

EgoNet:
- Multithreading in EgoNet which deserializes execution and comes with mild performance improvements on large sizes
- Some cleanup

### DIFF
--- a/python/cugraph/traversal/ms_bfs.py
+++ b/python/cugraph/traversal/ms_bfs.py
@@ -13,14 +13,16 @@
 
 import numpy as np
 import cudf
-from cugraph.structure.graph import Graph, DiGraph
-from cugraph.utilities.utils import get_device_memory_info
+
+# from cugraph.structure.graph import Graph, DiGraph
+# from cugraph.utilities.utils import get_device_memory_info
 import warnings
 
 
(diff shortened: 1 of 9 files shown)


ChuckHastings (approved):

python/cugraph/traversal/ms_bfs.py
```diff
@@ -0,0 +1,276 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
Iroy30, line 1: Do we want a test for this?

Iroy30: I see that it is not been implemented but exposed, is that right?
In that case, we could either raise error or not expose it to user

afender: Right, it is just the API proposal. I can raise an error.

Iroy30 (approved):

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1475, opened 2021-03-21
[REVIEW] Fix for bug in SCC on self-loops

### LINKED ISSUE #1471: [BUG] SCC results for graph with loops seems incorrect and does not match Nx nor SciPy
I recently observed behavior from `cugraph.strongly_connected_components` when handling directed graphs that have loops that appears to be a bug. The same input given to the SCC algos in SciPy and NetworkX both return identical results which match expectations (the input graph is trivially small and easy to predict results with, and is included in the test script attached below).

Other information:
* This was initially discovered through using our C++ API from `libcugraph` directly, so that should eliminate the root cause of this issue as something in the python layer.
* The handling of loops are assumed to be related, but only because the attached script uses them in the comparison and that may not be the root cause.
* The attachments describe the issue best, but to summarize:
For this input:
```
[[1., 1., 0., 0., 0.,],
 [1., 1., 0., 0., 0.,],
 [1., 0., 1., 0., 0.,],
 [0., 1., 0., 1., 0.,],
 [0., 0., 0., 0., 1.,],
]
```
this output is observed:
```
NUM LABELS: sp: 4, cu: 2, nx: 4
GROUPS: sp: [{0, 1}, {2}, {3}, {4}], cu: [{0, 1, 2, 3}, {4}], nx: [{0, 1}, {2}, {3}, {4}]
```
where `sp` is SciPy, `cu` is cuGraph, and `nx` is NetworkX. Each library's strongly connected components API was called using a directed graph created from the array above.

[connected_components_example_py.txt](https://github.com/rapidsai/cugraph/files/6172244/connected_components_example_py.txt)

[connected_components_example_output.txt](https://github.com/rapidsai/cugraph/files/6167453/connected_components_example_output.txt)

aschaffer:
This provides fixes for strongly connected components on graphs with self-loops: https://github.com/rapidsai/cugraph/issues/1471.

closes #1471


BradReesWork (approved):

BradReesWork:
rerun tests

cpp/src/components/connectivity.cu
```diff
@@ -78,7 +78,7 @@ std::enable_if_t<std::is_signed<VT>::value> connected_components_impl(
                                                    stream);
   } else {
     SCC_Data<ByteT, VT> sccd(nrows, graph.offsets, graph.indices);
-    sccd.run_scc(labels);
+    auto num_iters = sccd.run_scc(labels);
```
rlratzel, line 81: Is this an unused variable, and if so should it be removed? I don't see it anywhere else in the diff.

cpp/tests/components/scc_test.cu
```diff
@@ -208,4 +218,209 @@ INSTANTIATE_TEST_CASE_P(
...
+
+  cudaDeviceProp prop;
+  int device = 0;
+  cudaGetDeviceProperties(&prop, device);
+
+  ASSERT_TRUE(n2 < prop.totalGlobalMem);
+
+  // Allocate memory on host
+  std::vector<IndexT> cooRowInd{0, 0, 1, 1, 2, 2, 3, 3, 4};
+  std::vector<IndexT> cooColInd{0, 1, 0, 1, 0, 2, 1, 3, 4};
```
rlratzel, line 289: Thanks for adding the additional tests. I think a FIXME could be added to change these new tests to be parameterized, with input parameters being `nrows`, the COO `col` and `row` indices, and the expected labels and number of components. That would reduce a lot of redundant code and make adding new graphs to test easier.

rlratzel (approved):
I have two minor comments that need not hold up approval if they don't get addressed.

cpp/src/components/connectivity.cu
aschaffer: It's useful for debugging and harmless. If you build debug and put a breakpoint there, you immediately get the `num_iters`, as opposed to having to re-build to provide it. It's not for immediate consumption, but for debugging needs. Does not trigger a warning, so I'd leave it there.

cpp/tests/components/scc_test.cu
aschaffer: I'm not sure what is the desired course of action here. A `FIXME` saying what? Remember this code path is to be retired (soon). So there's a way bigger FIXME looming here. This would add confusion, I think.

BradReesWork:
@gpucibot merge

PR #1484, opened 2021-03-24
[REVIEW] Add indirection and replace algorithms with new renumbering

Iroy30:


python/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -32,6 +32,10 @@ def call_pagerank(sID,
                   personalization,
                   nstart):
     wid = Comms.get_worker_id(sID)
+    print("RANK: ", wid)
```
ChuckHastings, line 35: Do we want these print statements left here?

ChuckHastings (approved):

python/cugraph/structure/number_map.py
```diff
@@ -10,100 +11,45 @@
...
 import dask_cudf
 import numpy as np
-import bisect
+import cudf
+
+
+def call_renumber(sID,
+                  data,
+                  num_edges,
+                  is_mnmg,
```
seunghwak, line 28: In case of single node multi-gpu, shouldn't this be false? If this should be true in single-node multi-gpu, isn't this a misnomer?

seunghwak (commented):
What change should I make to call the C++ renumber code in both Pascal & non-Pascal architectures? (I am planning to update the C++ renumber code to use a binary search based approach for Pascal).

And once renumber_map is returned, currently, unrenumbering step is purely python and differences in partitioning mechanisms shouldn't affect the unrenumbering process, correct?

python/cugraph/dask/link_analysis/pagerank.py
Iroy30: No. Will remove them , thanks for catching

afender (approved):

seunghwak:
> What change should I make to call the C++ renumber code in both Pascal & non-Pascal architectures? (I am planning to update the C++ renumber code to use a binary search based approach for Pascal).
> 
> And once renumber_map is returned, currently, unrenumbering step is purely python and differences in partitioning mechanisms shouldn't affect the unrenumbering process, correct?

Please disregard my question if we are not supporting Pascal only in the single-GPU mode.

seunghwak (approved):

BradReesWork:
@gpucibot merge

PR #1485, opened 2021-03-26
ENH Update conda recipes pinning of repo dependencies

mike-wendt:
Ensure all conda packages created in this repo that depend on other packages are all version pinned to the same build number. This way it prevents a conda solve from picking mismatched versions of `cugraph` and `libcugraph` that can break this repo during builds and testing.


BradReesWork (approved):

raydouglass (approved):

BradReesWork:
@gpucibot merge

PR #1498, opened 2021-04-01
Fix MNMG test failures and skip tests that are not supported on Pascal

jnke2016:
* Made various fixes to `test_mg_betweenness_centrality.py` to address recent updates that were breaking these tests.
* @afender updated egonet.cu to address an occasional test failure related to mem usage in certain environments.
* @rlratzel updated python and notebook tests that contain code not supported on Pascal to be skipped when running on that arch. Also added a script to detect Pascal and used it to skip **all** C++ tests. _Note: a better way would be to edit each C++ test to use GTEST_SKIP to conditionally skip on Pascal, but since there are now so many to skip, and we get coverage from Python, this was the faster solution without sacrificing much/any coverage for a single platform._


rlratzel:
rerun tests

rlratzel:
rerun tests

afender (approved):
PR looks fine to me but somehow one test failure showed up in EgoNet (out of memory on Karate dataset). I can't repro locally at this time so I'm experimenting in this PR.

afender:
Seems like my fix addressed for the egonet faliure in CI :tada:

rlratzel (approved):

ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

PR #1503, opened 2021-04-01
Reduce the size of the cugraph libraries

robertmaynard:
By explicitly telling nvcc's fatbin pass to always compress device code
we can ensure that our binaries are the smallest possible size.

See https://github.com/rapidsai/cudf/pull/7583 for additional context.


rlratzel (approved):
Thanks!

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1516, opened 2021-04-07
Random Walks - Python Bindings

### LINKED ISSUE #1488: [FEA] RandomWalk - Python Bindings
create the Python bindings around the RandomWalk C code

jnke2016:
Python bindings for random walks
closes #1488
check the rendering after the PR is merged to make sure everything render as expected


BradReesWork (approved):

BradReesWork (changes requested):
Please add a tests

python/cugraph/sampling/random_walks.py
```diff
@@ -0,0 +1,93 @@
...
+        Weights in the graph are ignored.
+        Use weight parameter if weights need to be considered
+        (currently not supported)
+
+    start_vertices : int or list or cudf.Series
+        A single node or a list or a cudf.Series of nodes from which to run
+        the random walks
+
+    max_depth : int
+        The depth of the random walks
```
afender, line 41:
```suggestion
        The maximum depth of the random walks
```

python/cugraph/sampling/random_walks.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Returns
+    -------
+    random_walks_edge_lists : cudf.DataFrame
+        GPU data frame containing all random walks sources identifiers,
+        destination identifiers, edge weights
+
+    seeds_offsets: cudf.Series
+        Series containing the starting offset in the returned edge list
+        for each seed.
```
afender, line 52:
```suggestion
        for each vertex in start_vertices.
```

(seed is undefined here)

afender (changes requested):
Please add  Random Walk to:
- the main `__init__.py`
- the `.rst` file 
- the list of algorithms in the `readme`

aschaffer (approved):

afender (approved):
Thanks, @jnke2016 !

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,155 @@
...
+DATASETS = [pytest.param(d) for d in utils.DATASETS]
+DATASETS_SMALL = [pytest.param(d) for d in utils.DATASETS_SMALL]
+
+
+def calc_random_walks(
+    graph_file,
+    directed=False,
+    max_depth=None
+):
+    """
```
BradReesWork, line 37: your description text does not match the function API

BradReesWork (approved):

python/cugraph/tests/test_random_walks.py
jnke2016: Yes, I can either make changes to match this API's description or modify the description

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,154 @@
...
+import cugraph
+import random
+
+
+# =============================================================================
+# Parameters
+# =============================================================================
+DIRECTED_GRAPH_OPTIONS = [False, True]
+WEIGHTED_GRAPH_OPTIONS = [False, True]
+DATASETS = [pytest.param(d) for d in utils.DATASETS]
```
rlratzel, line 28: I've been adding `ids` to make it easier to see what dataset is being run in the event of a failure:
```suggestion
DATASETS = [pytest.param(d) for d in utils.DATASETS,
            ids=[f"dataset={d.as_posix()}" for d in utils.DATASETS]]
```

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,154 @@
...
+    assert invalid_edge == 0
+    assert invalid_seeds == 0
+    assert invalid_weight == 0
+
+# =============================================================================
+# Pytest Setup / Teardown - called for each test function
+# =============================================================================
+
+
+def prepare_test():
```
rlratzel, line 117: If you make this a setup function, pytest will automatically call it for you before each test, as done [here](https://github.com/rapidsai/cugraph/blob/79c3ba059d2cc4f2bfafece0e92671013a584175/python/cugraph/tests/test_bfs.py#L75).

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,154 @@
...
+@pytest.mark.parametrize("graph_file", utils.DATASETS_SMALL)
+@pytest.mark.parametrize("directed", DIRECTED_GRAPH_OPTIONS)
+@pytest.mark.parametrize("max_depth", [None])
+def test_random_walks_invalid_max_dept(
+    graph_file,
+    directed,
+    max_depth
+):
+    """Test calls random_walks an invalid type"""
+    prepare_test()
```
rlratzel, line 130: You can remove this line if you change the above function to a setup function.

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,154 @@
...
+        )
+
+
+@pytest.mark.parametrize("graph_file", utils.DATASETS_SMALL)
+@pytest.mark.parametrize("directed", DIRECTED_GRAPH_OPTIONS)
+def test_random_walks(
+    graph_file,
+    directed
+):
+    max_depth = random.randint(2, 10)
```
rlratzel, line 145: If the test fails, will the user know what randomly chosen `max_depth` was used to get the results? This may need to be printed somewhere too so devs can reproduce the error if necessary.

python/cugraph/tests/test_random_walks.py
```diff
@@ -0,0 +1,154 @@
...
+    invalid_edge = 0
+    invalid_seeds = 0
+    invalid_weight = 0
+    offsets_idx = 0
+    for i in range(len(df.index)):
+        src, dst, weight = df.iloc[i].to_array()
+        if i == offsets[offsets_idx]:
+            if df['src'].iloc[i] != seeds[offsets_idx]:
+                invalid_seeds += 1
+                print(
```
rlratzel, line 87: This can be a FIXME, but tests probably shouldn't rely on print statements to show failures (ie. they should use specific assertions).  I see that this allows you to check every path instead of stopping on the first failure, so we may need to rethink how the test works if we want both assertions instead of prints, and having it not stop on the first failure.

rlratzel (commented):
The cython code looks fine, but I just had some comments/suggestions about the API and py tests.

python/cugraph/sampling/random_walks.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    compute random walks for each nodes in 'start_vertices'
+
+    parameters
+    ----------
+    G : cuGraph.Graph or networkx.Graph
+        The graph can be either directed (DiGraph) or undirected (Graph).
+        Weights in the graph are ignored.
+        Use weight parameter if weights need to be considered
+        (currently not supported)
```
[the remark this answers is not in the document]
jnke2016: This should be removed. There is no weight parameter

python/cugraph/sampling/random_walks.py
```diff
@@ -0,0 +1,95 @@
...
+        df_ = cudf.DataFrame()
+        df_['vertex_set'] = vertex_set
+        df_ = G.unrenumber(df_, 'vertex_set', preserve_order=True)
+        vertex_set = cudf.Series(df_['vertex_set'])
+
+    edge_list = defaultdict(list)
+    next_path_idx = 0
+    offsets = [0]
+
+    df = cudf.DataFrame()
```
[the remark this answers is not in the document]
rlratzel: You may be able to just call this utility as is done [here](https://github.com/rapidsai/cugraph/blob/79c3ba059d2cc4f2bfafece0e92671013a584175/python/cugraph/community/egonet.py#L108) (you would have test with a Nx input to be sure though).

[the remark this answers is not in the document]
jnke2016: Ok. And Brad asked to remove networkX

python/cugraph/tests/test_random_walks.py
jnke2016: I borrowed the idea from test_BFS. I can fix and Fail the test when the first error appears. I liked this approach because the test does not crash at the first failure and I get to see the other mismatches to see a pattern when debugging.

rlratzel: I agree that can be nice when you want to see everything. A FIXME will let us revisit this later in a way that can give us both assertions and the ability to see other mismatches if you'd rather do that.

BradReesWork:
rerun tests

BradReesWork:
@gpucibot merge

PR #1520, opened 2021-04-07
[ENH] Undo cuco::static_map bug workarounds

seunghwak:
Undo workarounds due to cuco::static_map bugs (Issues 71, 72, 73, 74, 75 in https://github.com/NVIDIA/cuCollections/issues) as those get fixed and update the cuCollection git tag to pull the version after the bug fixes.

Issue #1467, opened 2021-03-16
[FEA] MNMG WCC


BradReesWork:
Develop a MNMG WCC analytic

PR #1529, opened 2021-04-12
[FEA] New C++ testing structure

### LINKED ISSUE #1534: [ENH] Google test is deprecating INSTANTIATE_TEST_CASE_P in favor of INSTANTIATE_TEST_SUITE_P
Our tests should be updated to reflect this.

### LINKED ISSUE #1536: [ENH] 'make test' not working properly for MG tests
If you compile cugraph with MG tests building as follows:
```
cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DBUILD_CUGRAPH_MG_TESTS=on
make -j
make test
```
The `make test` call will execute the executables `MG_*_TEST` directly instead of using `mpirun`.

ChuckHastings:
This PR reorganizes the tests for the new graph object and graph primitives to simplify testing and support an expansion of the number of test data sources.

The principal idea is to segregate the testing of the algorithm from how the graph is actually constructed.  This allows for the composition of tests by using different test graph generators or input sources with the same test graph algorithm implementation.

Closes #1534 
Closes #1536 
Supports #1467

Should be merged after PR #1520 which fixes python format issues


cpp/tests/components/wcc_test.cpp
```diff
@@ -0,0 +1,120 @@
...
+  run_current_test<int32_t, int32_t, float>(std::get<0>(param), std::get<1>(param));
+}
+TEST_P(Tests_WCC_LineGraph, WCC)
+{
+  auto param = GetParam();
+  run_current_test<int32_t, int32_t, float>(std::get<0>(param), std::get<1>(param));
+}
+
+// --gtest_filter=*simple_test*
+INSTANTIATE_TEST_CASE_P(
```
rlratzel, line 107: I believe `INSTANTIATE_TEST_CASE_P` is or will be deprecated in favor of the new name `INSTANTIATE_TEST_SUITE_P` ([doc update](https://github.com/google/googletest/pull/2193/files))

cpp/tests/experimental/katz_centrality_test.cpp
```diff
@@ -311,35 +260,48 @@ class Tests_KatzCentrality : public ::testing::TestWithParam<KatzCentrality_Usec
...
-  run_current_test<int32_t, int32_t, float, float>(GetParam());
+  auto param = GetParam();
+  run_current_test<int32_t, int32_t, float, float>(std::get<0>(param), std::get<1>(param));
 }
 
 INSTANTIATE_TEST_CASE_P(
   simple_test,
-  Tests_KatzCentrality,
+  Tests_KatzCentrality_File,
+  ::testing::Values(
```
rlratzel, line 285: For this and all others, can we use [`::testing::Combine()`](https://code.ornl.gov/ezx/random_ezx/-/blob/3f7be7147cf536b8c7fd9483eaaa8b6f4fbb1671/gtest/include/gtest/gtest-param-test.h#L1225) to auto-generate the input combinations?  I believe it may require you to handle a `std::tuple` instead of a pair though.

rlratzel (commented):
still reviewing, but here's two suggestions so far...

cpp/tests/components/wcc_test.cpp
ChuckHastings: OK.  Do we want to update that here (and then as we go: "Boy Scout Rule") or do we want to create a single PR that updates all of the test cases in one shot (might be easier to review)?

cpp/tests/experimental/katz_centrality_test.cpp
ChuckHastings: We could.  The question is do we want to test all combinations in the Cartesian product?  Some of the other examples (e.g. BFS and SSSP) require parameters that might not make sense.  For example, testing a graph with 100 vertices but specifying a starting vertex id of 1000 would fail.  But it's reasonable to test that using a starting vertex other than 0 does pass.

I'm inclined to explicitly enumerate them... although I could convert from pair to tuple (already using the `std::get` syntax) and construct an example to validate that the `::testing::Combine()` pattern would work.

cpp/tests/components/wcc_test.cpp
rlratzel: Thanks, I'm fine with either approach. BSR might be more successful since another wholesale change PR is easy to put on a TODO list and get forever deferred ;)

cpp/tests/experimental/katz_centrality_test.cpp
rlratzel: That's a good point about not always wanting every combo.  Is this something that can be used for tests where all combos are needed, but not for others?  (I was specifically noticing that the inputs for Katz looked like every combination).  I don't have a strong preference but was just thinking it could clean up some tests, your call.

I suppose if you at least refactored it to a pair as you suggested, then other tests could possibly use `Combine()` if it makes sense for them (and possibly be closer to other patterns gtest uses).

cpp/tests/components/wcc_test.cpp
```diff
@@ -0,0 +1,120 @@
...
+    cugraph::experimental::graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>,
+    rmm::device_uvector<vertex_t>>
+  construct_graph(raft::handle_t const& handle, bool test_weighted, bool renumber = true) const
+  {
+    // TODO: Write an implementation here
+    CUGRAPH_FAIL("Not implemented");
+  }
+
+ private:
+  size_t num_vertices_;
```
seunghwak, line 42: ```size_t num_vertices_{0};```

Better avoid uninitialized variables (unless there is a clear performance reason, e.g. rmm:device_uvector)

cpp/tests/components/wcc_test.cpp
```diff
@@ -0,0 +1,120 @@
...
+    // TODO: validate result
+  }
+};
+
+using cugraph::test::File_Usecase;
+using cugraph::test::Rmat_Usecase;
+
+using Tests_WCC_File = Tests_WCC<File_Usecase>;
+using Tests_WCC_Rmat = Tests_WCC<Rmat_Usecase>;
+using Tests_WCC_LineGraph = Tests_WCC<LineGraph_Usecase>;
```
seunghwak, line 88: What are the purposes of all these using statements?

cpp/tests/experimental/bfs_test.cpp
```diff
@@ -312,24 +265,45 @@ class Tests_BFS : public ::testing::TestWithParam<BFS_Usecase> {
   }
 };
 
+using cugraph::test::File_Usecase;
+using cugraph::test::Rmat_Usecase;
+
+using Tests_BFS_File = Tests_BFS<File_Usecase>;
+using Tests_BFS_Rmat = Tests_BFS<Rmat_Usecase>;
```
seunghwak, line 272: Do we really need all these using statements?

cpp/tests/experimental/pagerank_test.cpp
```diff
@@ -434,53 +376,84 @@ class Tests_PageRank : public ::testing::TestWithParam<PageRank_Usecase> {
   }
 };
 
+using cugraph::test::File_Usecase;
+using cugraph::test::Rmat_Usecase;
+
+using Tests_PageRank_File = Tests_PageRank<File_Usecase>;
+using Tests_PageRank_Rmat = Tests_PageRank<Rmat_Usecase>;
```
seunghwak, line 383: Do we need all these using statements?

cpp/tests/utilities/test_utilities.hpp
```diff
@@ -167,21 +168,120 @@ generate_graph_from_rmat_params(raft::handle_t const& handle,
...
+ private:
+  size_t scale_;
+  size_t edge_factor_;
+  double a_;
+  double b_;
+  double c_;
+  uint64_t seed_;
+  bool undirected_;
+  bool scramble_vertex_ids_;
+  bool multi_gpu_usecase_;
```
seunghwak, line 284: Better make it a habit to add ```{}``` at the end of variable declarations.

seunghwak (commented):
I like the high-level idea and posted some minor complaints about the code.

cpp/tests/experimental/katz_centrality_test.cpp
ChuckHastings: I'll push an update that uses this where appropriate.

cpp/tests/experimental/bfs_test.cpp
ChuckHastings: The first ones are for convenience, we could:
* Leave them out and always fully qualify references to the cugraph::test namespace
* Since these are all tests, we could just specify `using cugraph::test` and get anything in the test namespace

I'm fine with either.

The second pair is necessary for google test.  Google test uses macros to construct the test code below (`TEST_P`).  We can't specify a `Tests_BFS<File_Usecase>` as the first parameter to the macro.  I tried a number of variations to work around this, but it wants to construct a preprocessor token and can't include '_', '>' or ')' (not sure what else) in doing that.  The using statement redefines `Tests_BFS<File_Usecase>` or `Tests_BFS<cugraph::test::File_Usecase>` if we prefer into a token that google test can process through its macro expansion.

rlratzel (approved):

cpp/tests/components/wcc_test.cpp
ChuckHastings: Getting ready to just fix all of them in this PR.  Will push later today.

rlratzel (approved):
Looks good to me. I noticed one small formatting issue, but I'm re-approving now anyway.

cpp/tests/CMakeLists.txt
```diff
@@ -165,6 +166,117 @@ function(ConfigureTest CMAKE_TEST_NAME CMAKE_TEST_SRC)
...
 
+function(ConfigureTestMG CMAKE_TEST_NAME CMAKE_TEST_SRC)
+    add_executable(${CMAKE_TEST_NAME}
+        ${CMAKE_TEST_SRC})
+
+    target_include_directories(${CMAKE_TEST_NAME}
+        PRIVATE
+        "${CUB_INCLUDE_DIR}"
+        "${THRUST_INCLUDE_DIR}"
+	"${CUCO_INCLUDE_DIR}"
```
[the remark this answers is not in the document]
ChuckHastings: Interestingly... this was copied and pasted from an earlier section which has the same issue.

There is a TAB in front of CUCO_INCLUDE_DIR, the other lines used spaces.

I'll push a change that uses spaces for everything.

afender (approved):
I guess the next step is to create graphs with multiple components. 

We can leverage the multi edge list generator for this but I wonder if it would be exposed in the API like the rmat generators or as a C++ test function like the linear graph. Thoughts?

ChuckHastings:
> I guess the next step is to create graphs with multiple components.
> 
> We can leverage the multi edge list generator for this but I wonder if it would be exposed in the API like the rmat generators or as a C++ test function like the linear graph. Thoughts?

I am working in a separate branch on this.  Will create a Draft PR after this one is merged that shows the API I am proposing.

BradReesWork:
@gpucibot merge

PR #1531, opened 2021-04-13
[ENH] Added Random Walks COO convertor and profiling

aschaffer:
This PR is used to track enhancements to Random Walks functionality:
1. Paths2COO convertor: converts coalesced vertex/weight paths to COO format + offsets (including C++ API for Cython);
2. RW profiling;
3. Moving functionality / tests out of `experimental` sub-dirs;


cpp/src/utilities/high_res_timer.hpp
```diff
@@ -85,6 +85,17 @@ class HighResTimer {
     timers.clear();
   }
 
+  double average_runtime(std::string const &label)
```
rlratzel, line 88: This is a bit inconsistent with the naming convention used here (verbs). Perhaps it should be `get_average_runtime()`. I would also move it above the "display" methods for consistency as well.

cpp/src/utilities/high_res_timer.hpp
```diff
@@ -85,6 +85,17 @@ class HighResTimer {
...
+  {
+    auto it = timers.find(label);
+    if (it != timers.end()) {
+      return (static_cast<double>(it->second.second) / (1000000.0 * it->second.first));
+    } else {
+      std::cerr << "ERROR: timing label: " << label << "not found.";
```
rlratzel, line 94: "ERROR" -> "WARNING"

cpp/tests/sampling/random_walks_profiling.cu
```diff
@@ -0,0 +1,147 @@
...
+ public:
+  Tests_RandomWalksProfiling() {}
+  static void SetupTestCase() {}
+  static void TearDownTestCase() {}
+
+  virtual void SetUp() {}
+  virtual void TearDown() {}
+
+  template <typename vertex_t, typename edge_t, typename weight_t>
+  void run_current_test(RandomWalks_Usecase const& configuration)
```
rlratzel, line 82: I don't think we should be using gtest this way.  I'd rather see a simple executable that does this kind of benchmarking, or add to the existing python benchmarks. The latter adds more value since they'll show up in reports and can help catch regressions.

cpp/tests/sampling/rw_low_level_test.cu
```diff
@@ -782,3 +782,121 @@ TEST_F(RandomWalksPrimsTest, SimpleGraphRandomWalk)
 
   ASSERT_TRUE(test_all_paths);
 }
+
+TEST_F(RandomWalksPrimsTest, SingleRandomWalk)
```
rlratzel, line 786: Same feedback as last review: a proper gtest fixture should be used to drastically reduce the size of these tests and make them easier to maintain (especially since you're using `TEST_F`).

cpp/include/utilities/cython.hpp
```diff
@@ -207,6 +207,26 @@ struct random_walk_ret_t {
   std::unique_ptr<rmm::device_buffer> d_sizes_;
 };
 
+// aggregate for random_walks() COO return type
+// to be exposed to cython:
+//
+struct random_walk_coo_t {
```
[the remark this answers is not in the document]
aschaffer: Yes, maybe reusing `cy_multi_edgelists_t` might work.

Two different structures are used here (`random_walk_coo_t`, `random_walk_ret_t`) because the RW user may want both representations (COO and coalesced paths, respectively). We don't know yet, which will gain popularity. My suspicion is that the last would. Or, maybe both would, in different contexts. One was signed off for C++ (coalesced), the other for Python (COO). But COO done in Python is expensive. Prohibitive for large graphs.

cpp/src/experimental/random_walks.cuh
```diff
@@ -840,6 +847,156 @@ random_walks_impl(raft::handle_t const& handle,
   CUGRAPH_FAIL("Not implemented yet.");
 }
 
+// provides conversion to (coalesced) path to COO format:
```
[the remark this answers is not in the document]
aschaffer: Yes, I have plans to remove that from `experimental`, too.

cpp/tests/sampling/random_walks_profiling.cu
aschaffer: I don't understand what is the course of action, here. Can you point to an example of how should this be tested? Do we have what you want in our code already? If so, please point me there. If not, please provide a code snippet example.

This is done in many places in the code, including recent `egonet`. I'm not sure why it was okay there but not here. At which point I'd like to point out this is not a PR for revamping the way we test. That is a major code re-org beyond the scope not only of this PR, but also of this release, as I'm told.

If a simple fix is possible for this particular profiling, I'd like to see an example, please. Thank you.

cpp/tests/sampling/rw_low_level_test.cu
aschaffer: Same answer as before: please describe the course of action desired, here. `TEST_F` doesn't compile without `RandomWalksPrimsTest` "fixture" (if that's what you're referring to).

cpp/src/experimental/random_walks.cuh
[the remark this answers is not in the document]
aschaffer: Done.

cpp/src/utilities/high_res_timer.hpp
aschaffer: Why warning? If the user asks for a timing result that doesn't exist then something is erroneous; i.e., not something that can be ignored, like a warning is.

The debate might be whether to `throw` rather than dump an error message to `std::cerr`. But that would require surrounding this call by a `try`-block in tests. Which might not be a bad idea, actually.

cpp/src/utilities/high_res_timer.hpp
aschaffer: Done.

cpp/src/utilities/high_res_timer.hpp
aschaffer: For all intends and purposes, not finding a label is a failed test. Hence, I've added exception handling for this as a mechanism to communicate back to the tester that it should consider the test failed on label not found.

cpp/include/algorithms.hpp
```diff
@@ -1280,5 +1280,29 @@ random_walks(raft::handle_t const &handle,
...
+ * @param num_paths number of paths.
+ * @param d_coalesced_v coalesced vertex buffer.
+ * @param d_sizes paths size buffer.
+ * @return tuple of (src_vertex_vector, dst_Vertex_vector, path_offsets), where
+ * path_offsets are the offsets where the COO set of each path starts.
+ */
+template <typename vertex_t, typename index_t>
+std::
+  tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>, rmm::device_uvector<index_t>>
+  convert_paths_to_coo(raft::handle_t const &handle,
```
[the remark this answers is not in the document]
aschaffer: I didn't really like this defined in `algorithms.hpp`, either. But `cython.*` files don't include `functions.hpp` (not directly, at least). And if they do not, then including `functions.hpp` in `cython.*` would add yet another dependency in already large `cython.*`, with increased compile times, etc. 

Also, `functions.hpp` seems legacy code on its way to being retired after code re-org. Perhaps, you meant `graph_functions.hpp`?

cpp/include/utilities/cython.hpp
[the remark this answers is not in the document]
rlratzel: My preference would be to collaborate with @afender to see if `cy_multi_edgelist_t` can use the more generic `num_offsets_` member instead, but that's up to the both of you. It seems worth it to be able to propagate the use of a common type IMO.

I see the potential benefit of having both RW return types (COO and coalesced paths), but my preference would be to have `call_random_walks` return the one we intend to use in python now, and remove the need to expose the converter to cython completely. If we want to switch, then we change `call_random_walks` to return the other type. Even with both types and the converter exposed, a code change is still required for the python user to use a different return type, so it doesn't really save us anything to have both.  Again, that's my preference so feel free to resolve this if we want to keep this as-is.

cpp/tests/sampling/random_walks_profiling.cu
rlratzel: Sorry if I wasn't clear before. The issue here is that you're writing a benchmark but using gtest, which is not what it's intended for.  You're not actually testing anything (ie. there's no assertions).  I'm also concerned that this will propagate an anti-pattern (ie. writing gtests that aren't actually tests), and we'll have a large refactoring ahead of us. (BTW, the egonet tests are actual tests that have assertions).

For benchmarking, I'd consider either doing it in python (as done [here](https://github.com/rapidsai/cugraph/tree/branch-0.20/benchmarks)), or using gbench, or even a separate executable for C++. For profiling, we should also look into using [nvtx](https://docs.nvidia.com/gameworks/content/gameworkslibrary/nvtx/nvidia_tools_extension_library_nvtx.htm) which is what cudf and cuml are starting to use so we're consistent with RAPIDS.  Some of these suggestions may be considered out-of-scope for this PR though, so my recommendation would be to simply make this a (non-gtest) standalone executable.  Since you're not using many gtest features it shouldn't be a big change, and I can work with you offline if that doesn't make sense.  Another option would be to simply add the CUDA profiling and timer calls to your other gtests, so that way running a test will actually test something and it will be consistent with what egonet, etc. currently do.

I'm looking forward to more benchmarks and also profiling, especially ones that exercise our C++ library directly, and I think we can use this as an opportunity to start that practice.

aschaffer: It was specifically requested to add C++ profiling. I can certainly add testing predicates to qualify this as a test. 

Also, I don't mind introducing `gbench` in this PR, provided is not to time consuming and it doesn't need tons of boilerplate coding. This is meant as a _quick_ enhancement PR (emphasis on "quick") as we have more pressing priorities than `gbench`, right now. But, I will probably need some help with `gbench` as I have no prior experience with it. If it's something simple I can add that to this PR. If not, this would probably go into our re-org goals in following release(s).

cpp/tests/sampling/rw_low_level_test.cu
rlratzel: The request was to refactor the tests so they take advantage of a test fixture for reusing the repeated setup and teardown code.  I'm not proposing getting rid of `TEST_F` or the fixture name it requires.  [Here's](https://github.com/rapidsai/rmm/blob/branch-0.20/tests/device_buffer_tests.cu) a nice example of a test that uses a fixture (it's also using `TYPED_TEST` instead of `TEST_F`, but it's the same concept). It has just enough inlined in each test so you don't have to reverse engineer much to figure out what's going on, but the small amount of repeated setup which otherwise distracts from the test is handled by the fixture. Note also that almost every cuDF test uses a BaseFixture which handles the common RMM resource setup.

If you disagree that you need to refactor the repeated code into a fixture, then I'd recommend at least not using `TEST_F`, since it's misleading and unnecessary.  [Here's](https://github.com/rapidsai/rmm/blob/branch-0.20/tests/mr/device/cuda_async_mr_tests.cpp) an example of a test that is not using a fixture.

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
rlratzel: `graph_functions.hpp` seems better than `algorithms.hpp`, I agree.  I hate to add it to `experimental` though, but maybe that's still better than `algorithms.hpp`?  Your call, I was just pointing out that if you had a better idea on where it could live, maybe now is a good time to move it before it gets settled in :)

cpp/src/utilities/high_res_timer.hpp
rlratzel: Thanks.  FYI, I should have also mentioned this before, but the other issue with printing `ERROR` is that some scripts and CI systems look for that string in the output and may interpret it in a way we don't intend them to. As a best practice I usually throw/raise to communicate that I cannot reliably continue, and I use warnings for all other unclear situations. This looked like a warning since you were allowing the test to continue, but I like throwing better since like you said, it's an erroneous condition.

cpp/tests/sampling/rw_low_level_test.cu
aschaffer: Done.

cpp/tests/sampling/random_walks_profiling.cu
aschaffer: I am considering making it a separate executable. Do we have an example in our code you could point me to? If not what is the `cmake` machinery involved? Which `cmake` must be changed and how?

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
aschaffer: Hmm. `graph_functions.hpp` is dedicated to _generic_ graph functionality / helpers (@seunghwak , please correct me if I'm wrong). This function however is a dedicated helper (COO conversion) for a dedicated piece of functionality (RW). Also, `graph_functions.hpp` has definitions, not just declarations. I doubt that's the right place for it. I think `algorithms.hpp` is still better than it.

How about if I just _declare_ the function inside `cython.hpp` ? (we're only talking declaration here, anyway; the definition has always been in `random_walks.cuh`).

cpp/include/utilities/cython.hpp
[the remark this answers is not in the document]
aschaffer: I'd like to point out that `std::tuple<...>` was designed especially for this purpose: a mechanism to aggregate entities in an unnamed but flexible way. 

Unfortunately, `tuple` cannot be exposed to Cython (otherwise we wouldn't need many of these wrappers). The reason being, Cython doesn't currently support variadic templates. 

For that reason alone, a proliferation of these wrappers is inevitable. _Until_ I get to refactor all these and use the type-erasure proposed in the cascaded dispatch design (the so called _dispatch layer_; hopefully soon, but not in the next release). The gain obtained from reusing a different existent wrapper structure is overshadowed by added confusion; i.e., loss in readability and maintainability, IMHO.

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
aschaffer: Actually, probably better, I'll just put that declaration in its own header: `src/sampling/rw_functions.hpp` and include it in `cython.cu`.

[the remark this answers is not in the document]
hlinsen: I recently added this function https://github.com/rapidsai/cugraph/blob/branch-0.20/cpp/src/utilities/path_retrieval.cu
We might want to put accelerated utility functions like this in a single place.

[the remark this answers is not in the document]
aschaffer: @hlinsen , okay I see an `include/utilities/path_retrieval.hpp` with declarations only. I'll try including that in `cython.cu`. Thanks.

cpp/tests/sampling/random_walks_profiling.cu
rlratzel: I think that's the best approach for this as well.  I think that'll give us more flexibility too since we won't be constrained by gtest's testing assumptions.  I can help with this but I unfortunately don't have an example to point you to.  I'll chat with you offline and then you or we can update the PR (I don't think it'll be a big change).

cpp/include/algorithms.hpp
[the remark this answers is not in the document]
aschaffer: Done.

rlratzel (approved):

cpp/tests/sampling/rw_low_level_test.cu
```diff
@@ -782,3 +782,121 @@ TEST_F(RandomWalksPrimsTest, SimpleGraphRandomWalk)
 
   ASSERT_TRUE(test_all_paths);
 }
+
+TEST(RandomWalksSpecialCase, SingleRandomWalk)
```
ChuckHastings, line 786: Is there a reason this can't be `rw_low_level_test.cpp`?  We've been trying to migrate to .cpp files where possible since they compile faster.

ChuckHastings (approved):
Once comment... we can address it later.

cpp/tests/sampling/rw_low_level_test.cu
aschaffer: Yes. `check_col_indices()` calls `thrust`.

ChuckHastings: Might be worth considering migrating that to a utility file at some point.

Note we should probably use `thrust::count_if` in that function due to this
https://github.com/NVIDIA/thrust/issues/1016

BradReesWork:
@gpucibot merge

Issue #1528, opened 2021-04-12
[ENH] C++ 17


afender:
Update config files to support C++17

Please add new cuda verisons to cugraph_dev conda files as part of this (thanks, @Salonijain27 for catching that)

PR #1551, opened 2021-04-22
Update graph primitives to support vertex-tagging in frontier expansion

seunghwak:
To get early feedback and to create a baseline weakly-connected-component PR dependent on this PR.
This PR requires C++17 and will not compile till https://github.com/rapidsai/cugraph/issues/1528 gets resolved.

Tagging vertex IDs with root IDs is necessary in 

multi-source BFS, k-hop neighbors, and the newly designed weakly-connected-component algorithm based on multi-root collaborative frontier expansion.

To fully support multi-source BFS & k-hop neighbors, we also need to store/query properties for (vertex, tag) pairs, and this update will be added in a future PR.


ChuckHastings (commented):
LGTM so far.

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1556, opened 2021-04-22
Removed GIT_SHALLOW for cuHornet FetchContent since a specific commit must now be used

rlratzel:
Removed `GIT_SHALLOW` for cuHornet `FetchContent` since a specific commit must now be used.


rlratzel:
This was already fixed in https://github.com/rapidsai/cugraph/pull/1535, so this will be closed in favor of merging `branch-0.18` into `branch-0.19` to maintain proper history.  cc @raydouglass

Issue #1566, opened 2021-04-28
[BUG] Random Walk crashes on unweighted graph


BradReesWork:
**Describe the bug**

The graph must be weighted or Random Walk crashes

```
#  Import the modules
import cugraph
import cudf

datafile='./data/karate-data.csv'
gdf = cudf.read_csv(datafile, delimiter='\t', names=['src', 'dst'], dtype=['int32', 'int32'] )

G = cugraph.Graph()
G.from_cudf_edgelist(gdf, source='src', destination='dst')

seeds = [17,19]

rw, so = cugraph.random_walks(G, seeds, 4)

```

ChuckHastings:
Closed by #1567

PR #1567, opened 2021-04-28
[BUG] Fix Random Walks output format and handle missing weights.

aschaffer:
This PR tracks work on missing weights, https://github.com/rapidsai/cugraph/issues/1566, by providing default `weight_t{1}` weights.

And addresses output format changes, as follows:

1. offsets for vertex paths starting indices, instead of sizes;
2. set of pairs (offset, size) for edge (weight) paths;  Example: for an edge path with offsets 0,3,3,5,... meaning 1st path has 3 edges, 2nd path has 0 edges (!), 3rd has 2 edges, etc.; the return is: (0,3), (3,0), (3,2), (5,...), ...;
3. The remaining output format stays the same (i.e., coalesced vertex sets, and coalesced weight sets);


cpp/src/sampling/random_walks.cuh
```diff
@@ -321,6 +286,28 @@ struct col_indx_extract_t<graph_t, index_t, std::enable_if_t<graph_t::is_multi_g
       [] __device__(auto crt_out_deg) { return crt_out_deg > 0; });
   }
 
+  // singleton for handling the case of missing
+  // weights, in which case a device buffer filled with
+  // `weight_t{1}` should be used as scratch-pad;
+  // this is static data to avoid device memory fragmentation
+  // by (potentially needlesly) re-allocating this scratch-pad;
+  //
+  static weight_t const* get_weights_scratch_pad(index_t num_edges, raft::handle_t const& handle)
```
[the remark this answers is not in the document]
aschaffer: I agree. Unfortunately there's no common base type between `thrust::constant_iterator<T>` and `T*` (that I know of). So, what type should `values_` be? (currently it's `weight_t const*`). I have to think about other ways to avoid that allocation.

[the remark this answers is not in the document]
ChuckHastings: I wonder, now that we're in C++ 17, if you can use `std::optional` to address this?  Pass the weight as an `std::optional<T const *>` and check in the code whether to use it or not?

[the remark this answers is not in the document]
aschaffer: I deref `values_` in device. Although indirectly, via a capture `[...,values = values_]`. Maybe a `std::variant<weight_t const*, thrust::constant_iterator<weight_t>>` might work, but not in device...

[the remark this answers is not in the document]
aschaffer: Certainly that allocation must go.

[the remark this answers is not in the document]
ChuckHastings: How important is it to avoid testing for `values == nullptr` in the device code?  Not sure of your motivation for that concern.

You could make a real functor instead of a device lambda and use SFINAE to let the compiler optimize that away if you're concerned about that.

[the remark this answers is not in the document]
aschaffer: No, I cannot use SFINAE on runtime values. `values == nullptr` is something decided at runtime, not compile-time.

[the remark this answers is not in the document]
ChuckHastings: Yes, but you could dispatch to the right functor (use an if to pick the right implementation).

I know cudf has a number of examples of a templated device functor with a true/false in the template and anything within the `if (template_variable)` block is not compiled version where it's false and is in the version where it's true.  So you use the same code, compile it twice, and pick the right version at runtime.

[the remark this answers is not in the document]
aschaffer: I'm trying to avoid the `if`. Not at any cost, though. If the workaround turns out to be overkill or hard to read, I'll just go with the `if` branching in device.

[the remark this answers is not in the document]
ChuckHastings: The `if` outside the device function (to pick the right functor) is much lower cost.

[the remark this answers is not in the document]
aschaffer: Is it? It's done in CPU. While inside the device lambda kernel it's done in GPU (many times...but in parallel).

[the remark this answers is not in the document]
ChuckHastings: I believe it is if the graph is large enough.  It will get executed within each GPU thread multiple times, it would only get executed on the CPU once.  I suppose we could time it if we think the behavior is different.  But on a graph with a billion edges it feels like the if inside the device code would get executed thousands of times per thread which should be slower than executing it once on the CPU.

Anyway... however you think works best, I think either solution is better than allocating all of the extra memory.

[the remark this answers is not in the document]
aschaffer: Done.

cpp/src/sampling/random_walks.cuh
```diff
@@ -386,11 +354,15 @@ struct random_walker_t {
   random_walker_t(raft::handle_t const& handle,
                   graph_t const& graph,
                   index_t num_paths,
-                  index_t max_depth)
+                  index_t max_depth,
+                  vertex_t v_padding_val = 0,
+                  weight_t w_padding_val = 0)
```
seunghwak, line 359: Should we better pad with invalid_vertex_id<vertex_t>::value() and std::numeric_limits<weight_t>::max()?

seunghwak: and 0 here can be interpreted in two different ways: vertex ID 0 and an indicator saying v_padding_val is not provided.

I think it is better to just fix the padding value or use std::optional to allow users to optionally provide padding values.

cpp/src/sampling/random_walks.cuh
```diff
@@ -690,11 +662,31 @@ struct random_walker_t {
...
+                 weight_padding_value_);
+  }
+
  private:
   raft::handle_t const& handle_;
   index_t num_paths_;
   index_t max_depth_;
   device_vec_t<edge_t> d_cached_out_degs_;
+  vertex_t const vertex_padding_value_;
+  weight_t const weight_padding_value_;
```
seunghwak, line 689: Better add trailing {} to class member variables (unless the default constructor is deleted).

cpp/src/sampling/random_walks.cuh
```diff
@@ -1003,18 +1031,27 @@ struct coo_convertor_t {
  * @brief returns random walks (RW) from starting sources, where each path is of given maximum
  * length. Uniform distribution is assumed for the random engine.
  *
- * @tparam graph_t Type of graph (view).
+ * @tparam graph_t Type of graph/view (typically, graph_view_t).
```
seunghwak, line 1034: What is "/"?

cpp/src/sampling/random_walks.cuh
```diff
@@ -1003,18 +1031,27 @@ struct coo_convertor_t {
...
  * @param max_depth maximum length of RWs.
+ * @param use_padding (optional) specifies if return uses padded format (true), or coalesced
+ * (compressed) format; when padding is used the output is a matrix of vertex paths and a matrix of
+ * edges paths (weights); in this case the matrices are stored in row major order; the vertex path
+ * matrix is padded with `num_vertices` values and the weight matrix is padded with `0` values;
  * @return std::tuple<device_vec_t<vertex_t>, device_vec_t<weight_t>,
- * device_vec_t<index_t>> Triplet of coalesced RW paths, with corresponding edge weights for
- * each, and coresponding path sizes. This is meant to minimize the number of DF's to be passed to
- * the Python layer.
+ * device_vec_t<index_t>> Triplet of either padded or coalesced RW paths; in the coalesced case
```
seunghwak, line 1047: What is device_vec_t? I guess this alias is removed.

cpp/src/sampling/random_walks.cuh
```diff
@@ -1076,5 +1114,47 @@ std::
...
+                         ptr_d_sizes + num_paths,
+                         d_vertex_offsets.begin());
+
+  thrust::transform(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+                    ptr_d_sizes,
+                    ptr_d_sizes + num_paths,
+                    d_weight_sizes.begin(),
+                    [] __device__(auto vertex_path_sz) { return vertex_path_sz - 1; });
+
+  CUDA_TRY(cudaStreamSynchronize(handle.get_stream()));
```
seunghwak, line 1148: Just FYI, get_stream() will be eventually removed and we need to transition to handle.get_stream_view().synchronize(); so better start to use this in at least new code.

cpp/src/sampling/random_walks.cuh
```diff
@@ -1076,5 +1114,47 @@ std::
...
+
+  thrust::exclusive_scan(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
```
seunghwak, line 1150: Just FYI, similarly, we'll eventually transition to ```rmm::exec_policy(handle.get_stream_view())``` instead of lengthier ```rmm::exec_policy(handle.get_stream())->on(handle.get_stream())```. Can be a 0.21 task.

cpp/src/sampling/random_walks.cuh
aschaffer: A few notes: 
1. there's no `std::optional` on device.
2. there's no `std::optional` in Cython;
3. by now there have been extended discussions about the "golden numbers" for padding.

I don't mind changing padding (again), but at some point we need to freeze the requirements and implement them. In general, design decisions like that should be made _before_ implementation, not after. 

For the record, `num_vertices` is currently used as padding value for vertices (not `0`; default value of `0` says: "if padding is used, then use `num_vertices` as padding for vertexes; if non-zero, and padding is used, then the supplied `v_padding_val` is used). 

For the weight padding, `weight_t{0}` was specifically requested.

cpp/src/sampling/random_walks.cuh
aschaffer: Why? Those are `const` members. No chance their initialization could be missed in the cnstr. initalization list. That would trigger a compiler error.

In general, it might be helpful to differentiate between _personal_ code-style (which may differ from reviewer to reviewer, hence they should not / cannot be enforced) and generic code-style, or actual issues, missuses, mistakes (which is obvious that they need correcting).

cpp/src/sampling/random_walks.cuh
aschaffer: It means: "graph" or "graph_view" => graph/graph_view => (by contraction) "graph/view".

cpp/src/sampling/random_walks.cuh
aschaffer: `device_vec_t` is an alias that has been there already. I have explained the motivation of its existence in the very 1st PR that had that code.

cpp/src/sampling/random_walks.cuh
aschaffer: Fixed.

cpp/src/sampling/random_walks.cuh
aschaffer: I see its use in one file only:  `cpp/tests/utilities/thrust_wrapper.cu`. If / when larger scope adoption happened in our code I would change it.

cpp/src/sampling/random_walks.cuh
aschaffer: In fact, with brace default initializer, '{}` on the `const` members things are worse. Because, in that case, the constructor initialization list is allowed to _miss_ the initialization of `const` members. Which is _not_ an option here. The constructor must decide what value to use for those `const`'s, because the logic is not simply: "default to 0".

cpp/src/sampling/random_walks.cuh
aschaffer: Oh, my bad. I see, that's in the public API, I'll fix that.

cpp/src/sampling/random_walks.cuh
seunghwak: I see.

I am not sure ```there's no std::optional on device.``` matters but ```there's no std::optional in Cython;``` surely matters.

Just a suggestion, and if there has been an extended discussion and some contexts I am not aware of, no complaint.

cpp/src/sampling/random_walks.cuh
seunghwak: Oh, I didn't see const here. I was commenting about the private member variables in general.

seunghwak (approved):

cpp/src/sampling/random_walks.cuh
aschaffer: Done.

BradReesWork:
@gpucibot merge

PR #1571, opened 2021-05-03
[REVIEW] add multi-column support in algorithms - part 2

Iroy30:
Adds multicolumn support for:
jaccard
wjaccard
overlap
woverlap
pagerank
spectral clustering
forceatlas


BradReesWork (approved):

BradReesWork:
rerun tests

python/cugraph/layout/force_atlas2.py
```diff
@@ -109,13 +108,14 @@ def on_train_end(self, positions):
     """
 
     if pos_list is not None:
-        null_check(pos_list["vertex"])
-        null_check(pos_list["x"])
-        null_check(pos_list["y"])
         if input_graph.renumbered is True:
+            if len(input_graph.renumber_map.implementation.col_names) > 1:
```
ChuckHastings, line 112: This reference: `input_graph.renumber_map.implementation.col_names` exposes internal implementation wherever we use it.  I'd suggest adding a data member or method (either on the NumberMap or Graph class -- I'm inclined to think the latter) to indicate that it's a multi-column situation.  Otherwise if we eventually change how we implement multi-column (i.e. when we rework it to use an indirect hash table as we have discussed) then all references to the implementation of renumber_map will need to be changed throughout the python code.

ChuckHastings (commented):
Glad we added multi-column tests.

BradReesWork:
rerun tests

Iroy30:
rerun tests

python/cugraph/link_prediction/jaccard.py
```diff
@@ -108,13 +108,23 @@ def jaccard(input_graph, vertex_pair=None):
...
+                        vertex_pair, col, col
+                    )
+        else:
             if input_graph.renumbered:
                 vertex_pair = input_graph.add_internal_vertex_id(
-                    vertex_pair, col, col
+                    vertex_pair, "src", columns[:vertex_size]
+                )
+                vertex_pair = input_graph.add_internal_vertex_id(
+                    vertex_pair, "dst", columns[vertex_size:]
```
rlratzel, line 127: This pattern is repeated in several algorithms - can the entire if-elif-else block be replaced by a utility that can be reused everywhere? This would remove a lot of code and make what's happening more self-documenting IMO, and could also minimize the amount of places that would need to be updated when implementation changes.  Maybe something that can be used like:
```
if type(input_graph) is not Graph:
    raise Exception("input graph must be undirected")

vertex_pairs = prepare_vertex_pairs(input_graph, vertex_pairs)

df = jaccard_wrapper.jaccard(input_graph, None, vertex_pairs)
```
side note: it seems like an algo call shouldn't modify the graph like this (eg. adding internal vertex IDs to the graph), but that's probably a bigger problem for a different PR.

rlratzel (commented):
Just a question and a suggestion. I also agree with @ChuckHastings request to hide accessing the renumber_map implementation detail directly.

python/cugraph/tests/test_jaccard.py
```diff
@@ -222,3 +223,32 @@ def test_jaccard_nx(graph_file):
...
+def test_jaccard_multi_column(graph_file):
+    gc.collect()
+
+    M = utils.read_csv_for_nx(graph_file)
+
+    cu_M = cudf.DataFrame()
+    cu_M["src_0"] = cudf.Series(M["0"])
+    cu_M["dst_0"] = cudf.Series(M["1"])
+    cu_M["src_1"] = cu_M["src_0"] + 1000
+    cu_M["dst_1"] = cu_M["dst_0"] + 1000
```
jnke2016, line 238: Just wondering why two different patterns to create `cu_M` (test_force_atlas2  and test_jaccacrd). I would probably stick with one for consistency.  During refactoring the generation of a graph from cu_M will definitely be in a fixture file to avoid rebuilding the same graph multiple times when running multiple test algo at once (just a side comment)

jnke2016 (commented):
Overall looks good to me.

ChuckHastings (approved):
Thanks for the update.

python/cugraph/tests/test_jaccard.py
Iroy30: updated for consistency

jnke2016 (approved):

Iroy30:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
@gpucibot merge

PR #1572, opened 2021-05-03
Fix multi-GPU hang on graph generation

seunghwak:
Two bug fixes for multi-GPU graph creation.

- Add barrier to avoid overlap between different communicators
- NCCL bug workaround on DGX1


seunghwak:
@rlratzel There is a one more bug I am tracking but you can run this on other machines and see how this works (to see the current unresolved issue is DGX1 specific or not).

rlratzel (approved):
Thanks! Nothing jumped out, so LGTM.

aschaffer (approved):

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1573, opened 2021-05-04
Bindings for RMAT

### LINKED ISSUE #1473: [ENH] Bindings for R-MAT
We should have bindings for the RMAT generator #1411 to facilitate testing and benchmarking in the python layer.
Adding bindings for #1483 (the list of RMAT graphs) as well.

jnke2016:
Bindings for RMAT
Bindings for the list of RMAT graphs
includes some quick fixes to address some sphinx doc-building warnings
closes #1473


cpp/src/utilities/cython.cu
```diff
@@ -789,6 +790,65 @@ std::unique_ptr<cy_multi_edgelists_t> call_egonet(raft::handle_t const& handle,
...
+
+  graph_generator_t gg_vals{
+    std::make_unique<rmm::device_buffer>(std::get<0>(src_dst_tuple).release()),
+    std::make_unique<rmm::device_buffer>(std::get<1>(src_dst_tuple).release())};
+
+  return std::make_unique<graph_generator_t>(std::move(gg_vals));
+}
+
+template <typename vertex_t>
+std::unique_ptr<graph_generator_t*> call_generate_rmat_edgelists(
```
aschaffer, line 818: Why return pointer-to-pointer? This construct returns a `unique_ptr` to a raw pointer, `graph_generator_t*`. Why not just returning a `unique_ptr<graph_generator_t>` like at line 797? 

Also, you probably have a memory leak here, because for a `unique_ptr` to a pointer to get deallocated it needs an explicit `Deleter`, otherwise I don't see how that memory gets freed.

cpp/src/utilities/cython.cu
```diff
@@ -1360,5 +1420,51 @@ template std::unique_ptr<renum_quad_t<int64_t, int64_t>> call_renumber(
...
+  size_t scale,
+  size_t num_edges,
+  double a,
+  double b,
+  double c,
+  uint64_t seed,
+  bool clip_and_flip,
+  bool scramble_vertex_ids);
+
+template std::unique_ptr<graph_generator_t*> call_generate_rmat_edgelists<int32_t>(
```
aschaffer, line 1445: Same pointer-to-pointer comment as above.

cpp/src/utilities/cython.cu
```diff
@@ -1360,5 +1420,51 @@ template std::unique_ptr<renum_quad_t<int64_t, int64_t>> call_renumber(
...
+  size_t min_scale,
+  size_t max_scale,
+  size_t edge_factor,
+  cugraph::experimental::generator_distribution_t size_distribution,
+  cugraph::experimental::generator_distribution_t edge_distribution,
+  uint64_t seed,
+  bool clip_and_flip,
+  bool scramble_vertex_ids);
+
+template std::unique_ptr<graph_generator_t*> call_generate_rmat_edgelists<int64_t>(
```
aschaffer, line 1457: Same pointer-to-pointer comment as above.

python/cugraph/generators/rmat.pxd
```diff
@@ -0,0 +1,44 @@
...
+        size_t scale,
+        size_t num_edges,
+        double a,
+        double b,
+        double c,
+        int seed,
+        bool clip_and_flip,
+        bool scramble_vertex_ids) except +
+
+    cdef unique_ptr[graph_generator_t*] call_generate_rmat_edgelists[vertex_t] (
```
aschaffer, line 34: Same pointer-to-pointer comment as above.

aschaffer (changes requested):

cpp/src/utilities/cython.cu
jnke2016: The first one create a pointer to a struct containing `rmm::device_buffer` while the second one create a pointer to a pointer of struct containing `rmm::device_buffer`. The latter is pointing to a list of tuple (src,  dst) while the first is pointing to only one tuple (src, dst). There is definitely a leak and I am looking at how to incorporate some sort of unique pointer to my pointer which will automatically get deallocated at the end

BradReesWork:
@jnke2016 The dask-worker-space/global.lock file needs to be deleted

rlratzel (changes requested):
Looks good. I found a few minor typos in the docstrings (my fault!), as well as another quick change that needs to be made in the cython code.

aschaffer (approved):

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1580, opened 2021-05-06
[REVIEW] [FEA] Implement Graph Batching functionality

aschaffer:
This tracks work on graph object serialization and broadcasting functionality:
1. C++ support of un/serialization;
2. C++ / Python support for broadcasting graph;

Python / Dask orchestration for assembling all the functionality above will follow up in a future task, when a few more dependencies will become clearer (`graph_t` object Python wrapper, or opaque handle; decision on whether/when to implement dispatch layer; or even forgo the graph object in python, altogether, and just call a function stub after broadcasting: `distribute_and_call(functor, params)`, etc.);

The RW Batch functionality is currently under review. Consequently, the scope of this PR has been modified accordingly.


ChuckHastings (approved):

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1582, opened 2021-05-07
Move headers into `include/cugraph` dir

### LINKED ISSUE #1491: [ENH] Refactor C++ include directory
To better match the rest of RAPIDS, the structure of include should be "cpp/include/cugraph"

create a new directory structure and move existing files

trxcllnt:
This PR moves the headers in `cpp/include` into `cpp/include/cugraph` and updates C++ and Cython `#include` directives.

This change makes it easier for other libs to avoid naming conflicts with cuGraph public headers when linking `libcugraph.so`.

closes #1491


ChuckHastings (changes requested):
There are some copyright dates causing failures.

Thanks for these edits!

ChuckHastings (approved):

rlratzel (approved):
LGTM, thanks for all the work here!

BradReesWork:
@gpucibot merge

PR #1591, opened 2021-05-08
[REVIEW] multi seeds BFS with one seed per component

afender:
CUDA/++ part. It builds and runs. Test check sum matches what we get from running BFS in a loop
Draft: Needs <s>python bindings and</s> further validation. Could be cleaned up after the next round of generator updates


cpp/src/traversal/bfs.cu
```diff
@@ -59,50 +60,97 @@ void bfs(raft::handle_t const& handle,
...
                  GraphViewType::is_multi_gpu,
                  static_cast<size_t>(Bucket::num_buckets)>
     vertex_frontier(handle);
 
-  if (push_graph_view.is_local_vertex_nocheck(source_vertex)) {
-    vertex_frontier.get_bucket(static_cast<size_t>(Bucket::cur)).insert(source_vertex);
+  // insert local source(s) in the bucket
+  if (aggregate_n_sources == 1) {
+    vertex_t src;
+    // Fixme: this (cheap) transfer could be skiped when is_local_vertex_nocheck accpets device mem
```
ChuckHastings, line 142: Suggest an all-caps FIXME... at least that's what I usually see in the code.

ChuckHastings (commented):
A few minor suggestions

ChuckHastings (approved):

rlratzel:
@gpucibot merge

PR #1601, opened 2021-05-13
Added `uninstall` target which uninstalls libcugraph and cugraph from a prior build/install step

rlratzel:
Added `uninstall` target which uninstalls libcugraph and cugraph from a prior build/install step. Also extended the `clean` target to remove inplace artifacts (mainly cython-generated cpp files and inplace `.so` python extensions built from cython).

NOTE: since targets can be combined on the same command, a user can get the "scrub" behavior by running:
```
./build.sh uninstall clean
```

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -19,10 +19,11 @@ ARGS=$*
 REPODIR=$(cd $(dirname $0); pwd)
 LIBCUGRAPH_BUILD_DIR=${LIBCUGRAPH_BUILD_DIR:=${REPODIR}/cpp/build}
 
-VALIDARGS="clean libcugraph cugraph docs -v -g -n --allgpuarch --buildfaiss --show_depr_warn -h --help"
+VALIDARGS="clean scrub libcugraph cugraph docs -v -g -n --allgpuarch --buildfaiss --show_depr_warn -h --help"
 HELP="$0 [<target> ...] [<flag> ...]
  where <target> is:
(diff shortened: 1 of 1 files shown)


dantegd:
@rlratzel FYI: we do this cleaning as part of CI at least since 0.14 https://github.com/rapidsai/cuml/blob/branch-0.20/python/setup.py#L70 and it hasn't affected anything, I would recommend making this behavior just part of the clean target (just my 2 cents)

rlratzel:
> @rlratzel FYI: we do this cleaning as part of CI at least since 0.14 https://github.com/rapidsai/cuml/blob/branch-0.20/python/setup.py#L70 and it hasn't affected anything, I would recommend making this behavior just part of the clean target (just my 2 cents)

Thank you @dantegd -  I'll just make `scrub` the default `clean` behavior then since that simplifies things.

rlratzel:
@dantegd I actually decided I didn't feel comfortable uninstalling during a clean, so I made the uninstall steps a separate target but added the more thorough cleaning (inplace artifacts) part of the default clean like cuML.

ChuckHastings (approved):
This feels like it would have addressed the docker issue that @jnke2016 had today.  I like the idea.

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1602, opened 2021-05-13
Move the utility function to gather distributed vectors from tests/community to tests/utilities and update MG tests

seunghwak:
- Move the utility function to gather distributed vectors from tests/community/mg_louvain_helper.cu to tests/utilities/device_comm_wrapper.cu and rename the function to device_gatherv
- Update MG tests to gather the MG results in root and compare with the SG result in root.
- Few more minor updates (sort_by_key thrust wrapper return value, adding missing const in input parameter, add missing include statement, add barrier in MG performance measurement)

This update is necessary to simplify MG WCC testing.


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1603, opened 2021-05-13
[REVIEW] adding test graphs - part 2

### LINKED ISSUE #1543: [FEA] Support generation of test graphs as a library feature (C++)
Networkx supports a whole host of graph generation capabilities as part of the library that users can call when using networkx.

cuGraph has some graph generation capabilities within the testing component.  These graph generation capabilities are going to be extended as we migrate things to use the pattern accelerators.

This issue is to migrate the existing graph generation capabilities into `libcugraph.so` and expose them in the C++ API.

ChuckHastings:
This effort was originally targeted toward the WCC effort, but has been expanded a bit.  This supersedes #1545 which I will close.

The goal here is to create a means for constructing test graphs in an easier fashion.  Testing the capabilities of different graph algorithms might require a variety of graphs.  The objective of this PR is to better organize the graph generation components and to introduce some components to help in composing graphs out of multiple components.

This PR introduces the following capabilities:
* Create an ER graph
* Create a collection of Complete Graphs
* Create a collection of 2D mesh graphs
* Create a collection of 3D mesh graphs
* Create a random path graph (connect all vertices with a single randomly ordered path)
* Translate vertex ids of a graph
* Combine multiple edge lists into a single graph

Closes #1543


cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+  rmm::device_uvector<vertex_t> copy_src_v(d_src_v.size(), handle.get_stream());
+  rmm::device_uvector<vertex_t> copy_dst_v(d_dst_v.size(), handle.get_stream());
+
+  raft::copy(copy_src_v.begin(), d_src_v.begin(), d_src_v.size(), handle.get_stream());
+  raft::copy(copy_dst_v.begin(), d_dst_v.begin(), d_src_v.size(), handle.get_stream());
+
+  std::vector<rmm::device_uvector<vertex_t>> srcs_v{};
+  std::vector<rmm::device_uvector<vertex_t>> dsts_v{};
+
+  srcs_v.push_back(std::move(d_src_v));
```
aschaffer, line 186: Consider using `emplace_back()` instead of `push_back()`, as it's more efficient for rvalue references (which is the case here, because of `std::move()`).

cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+  if (optional_d_weights_v) {
+    std::vector<rmm::device_uvector<weight_t>> weights_v{};
+
+    rmm::device_uvector<weight_t> d_weights_v = std::move(optional_d_weights_v.value());
+    rmm::device_uvector<weight_t> copy_weights_v(d_weights_v.size(), handle.get_stream());
+
+    raft::copy(
+      copy_weights_v.begin(), d_weights_v.begin(), d_weights_v.size(), handle.get_stream());
+
+    weights_v.push_back(std::move(d_weights_v));
```
aschaffer, line 201: Consider using `emplace_back()` instead of `push_back()`, as it's more efficient for rvalue references (which is the case here, because of `std::move()`).

aschaffer (changes requested):

cpp/src/generators/erdos_renyi_generator.cu
```diff
@@ -0,0 +1,121 @@
...
+    thrust::make_counting_iterator<size_t>(0), [seed] __device__(size_t index) {
+      thrust::default_random_engine rng(seed);
+      thrust::uniform_real_distribution<float> dist(0.0, 1.0);
+      rng.discard(index);
+      return dist(rng);
+    });
+
+  size_t count = thrust::count_if(rmm::exec_policy(handle.get_stream()),
+                                  random_iterator,
+                                  random_iterator + num_vertices * num_vertices,
```
seunghwak, line 49: Just want to note that the Gnp model can be way more expensive than the Gnm model if ```num_vertices``` is large (but much simpler to implement as we don't need to worry about duplicates).

Shouldn't we add gnp in the function name so users can expect this will go over ```num_vertices``` * ```num_vertices``` potential edges?

cpp/src/generators/erdos_renyi_generator.cu
```diff
@@ -0,0 +1,121 @@
...
+  rmm::device_uvector<vertex_t> src_v(count, handle.get_stream());
+  rmm::device_uvector<vertex_t> dst_v(count, handle.get_stream());
+
+  thrust::transform(rmm::exec_policy(handle.get_stream()),
+                    indices_v.begin(),
+                    indices_v.end(),
+                    thrust::make_zip_iterator(thrust::make_tuple(src_v.begin(), src_v.end())),
+                    [num_vertices] __device__(size_t index) {
+                      size_t src = index / num_vertices;
+                      size_t dst = index % num_vertices;
```
seunghwak, line 71: This will break if vertex_t is 64 bit.

cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+}
+
+}  // namespace detail
+
+template <typename vertex_t>
+void scramble_vertex_ids(raft::handle_t const &handle,
+                         rmm::device_uvector<vertex_t> &d_src_v,
+                         rmm::device_uvector<vertex_t> &d_dst_v,
+                         vertex_t vertex_id_offset,
+                         uint64_t seed)
```
seunghwak, line 55: Have you checked that this function works if num_vertices is not a power of two?

IIRC, this function works if vertices are in the range [0, 2^scale), but I am not sure this still works otherwise. Need to check.

cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+#include <numeric>
+
+namespace cugraph {
+
+namespace detail {
+
+template <typename T>
+void append_all(raft::handle_t const &handle,
+                std::vector<rmm::device_uvector<T>> &&input,
+                rmm::device_uvector<T> &output)
```
seunghwak, line 38: Should we better return rmm::device_uvector<T> instead of taking ```rmm::device_uvector<T>& output```. Unless the ```ouput``` is in-out or there is another strong reason to do so, returning is more functional and side-effects-free than taking an lvalue reference.

cpp/src/generators/generator_tools.cu
seunghwak: I guess this will still work (but the scrambled vertex IDs are no longer contiguous integers but does not matter if we renumber or allow having many isolated vertices... but needs to double check).

cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+  detail::append_all<vertex_t>(handle, std::move(dests), dsts_v);
+
+  if (optional_d_weights) {
+    weights_v.resize(size, handle.get_stream());
+    detail::append_all(handle, std::move(optional_d_weights.value()), weights_v);
+  }
+
+  size_t number_of_edges{0};
+
+  if (optional_d_weights) {
```
seunghwak, line 127: So, should we always remove all multi-edges or better make this as an option? Graph500 input edge lists can possibly have multi-edges (and self-loops) and removing those is a task in the graph construction step.

cpp/src/generators/generator_tools.cu
```diff
@@ -0,0 +1,278 @@
...
+}
+
+template <typename vertex_t, typename weight_t>
+std::
+  tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>, rmm::device_uvector<weight_t>>
+  symmetrize_edgelist(raft::handle_t const &handle,
+                      rmm::device_uvector<vertex_t> &&d_src_v,
+                      rmm::device_uvector<vertex_t> &&d_dst_v,
+                      std::optional<rmm::device_uvector<weight_t>> &&optional_d_weights_v)
+{
```
seunghwak, line 176: I think the code below is simpler and more efficient (in memory footprint & the amount of data movement).
```
auto offset = d_src_v.size();
d_src_v.resize(offset * 2, handle.get_stream_view());
d_dst_v.resize(d_src_v.size(), handle.get_stream_view());
thrust::copy(rmm::exec_policy(handle.get_stream_view()), d_dst_v.begin(), d_dst_v.begin() + offset, d_src_v.begin() + offset);
thrust::copy(rmm::exec_policy(handle.get_stream_view()), d_src_v.begin(), d_src_v.begin() + offset, d_dst_v.begin() + offset);
if (optional_d_weights_v) {
  optional_d_weights_v.resize(d_src_v.size(), handle.get_stream_view());
  thrust::copy(rmm::exec_policy(handle.get_stream_view()), optional_d_weight_v.begin(), optinoal_d_weight_v.begin() + offset, optional_d_weight_v.begin() + offset);
}

return std::make_tuple(std::move(d_src_v), std::move(d_dst_v), optional_d_weights_v ? std::move(optional_d_weights_v) : std::nullopt);
```

cpp/src/generators/simple_generators.cu
```diff
@@ -0,0 +1,336 @@
...
+
+  for (auto tuple : component_parms_v) {
+    vertex_t num_vertices, base_vertex_id;
+    std::tie(num_vertices, base_vertex_id) = tuple;
+
+    auto transform_iter = thrust::make_transform_iterator(
+      thrust::make_counting_iterator<size_t>(0),
+      [base_vertex_id, num_vertices, invalid_vertex] __device__(size_t index) {
+        size_t graph_index = index / (num_vertices * num_vertices);
+        size_t local_index = index % (num_vertices * num_vertices);
```
seunghwak, line 267: This will break if vertex_t is 64 bit.

cpp/src/generators/generator_tools.cu
ChuckHastings: OBE with next push.

cpp/src/generators/generator_tools.cu
ChuckHastings: OBE with next push.

cpp/src/generators/erdos_renyi_generator.cu
```diff
@@ -0,0 +1,121 @@
...
+    });
+
+  size_t count = thrust::count_if(rmm::exec_policy(handle.get_stream()),
+                                  random_iterator,
+                                  random_iterator + num_vertices * num_vertices,
+                                  [p] __device__(float prob) { return prob < p; });
+
+  rmm::device_uvector<size_t> indices_v(count, handle.get_stream());
+
+  handle.get_stream_view().synchronize();
```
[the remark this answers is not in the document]
ChuckHastings: Residual from some earlier debugging.  Deleted.

cpp/src/generators/erdos_renyi_generator.cu
ChuckHastings: Changed function name.  Add a gnm function as well, although not implemented in this PR.

cpp/src/generators/erdos_renyi_generator.cu
ChuckHastings: Technically, it will break if num_vertices > (2^31 - 1).  If vertex_t is 64 bits but could be stored in 32 bits it would still work.

Since we don't want this variation called if num_vertices > (2^31 - 1)... even if it worked we don't want to do that much computation... I just added a CUGRAPH_EXPECTS at the beginning of the function.

cpp/src/generators/generator_tools.cu
ChuckHastings: Added a unit test for the scramble function.  The result is correct (it validates properly).  Your assertion is correct regarding the resulting ids no longer being contiguous.  This will (as you observe) result in a collection of isolated vertices if we do not renumber; although if we renumber that will be corrected.

cpp/src/generators/generator_tools.cu
ChuckHastings: Done in next push

cpp/src/generators/generator_tools.cu
ChuckHastings: Added a flag to control this behavior in next push.

cpp/src/generators/generator_tools.cu
ChuckHastings: Implemented this version.

cpp/src/generators/simple_generators.cu
ChuckHastings: Added a check (like above)

aschaffer (approved):

BradReesWork:
@gpucibot megre

BradReesWork:
@gpucibot merge

BradReesWork (approved):

PR #1604, opened 2021-05-14
SG & MG Weakly Connected Components

seunghwak:
- C++ SG & MG Weakly Connected Components
- This PR is to help python binding development
- Performance/memory footprint optimizations for SG, MG with # GPUs < (E/V)^2, MG with # GPUs > (E/V)^2 will be in separate PRs


aschaffer (changes requested):

cpp/src/components/weakly_connected_components.cu
```diff
@@ -0,0 +1,638 @@
...
+    new_roots.resize(num_new_roots + scan_size, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> scan_counts(0, handle.get_stream_view());
+    {
+      rmm::device_uvector<vertex_t> indices(scan_size, handle.get_stream_view());
+      auto input_pair_first = thrust::make_zip_iterator(thrust::make_tuple(
+                                candidate_first, thrust::make_counting_iterator(vertex_t{0}))) +
+                              num_scanned;
+      auto output_pair_first = thrust::make_zip_iterator(
+        thrust::make_tuple(new_roots.begin() + num_new_roots, indices.begin()));
+      new_roots.resize(
```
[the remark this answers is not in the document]
seunghwak: This resize is a down-sizing so does not involve memory reallocation or copying, but your comment is valid to the resize in https://github.com/rapidsai/cugraph/pull/1604/files/8d2a5c4ec2b3bf74386886c2b65a205054fb2d1b#diff-eea086cb1f3b4fd23ff600a41edce68425d996f05e3b80afdac36c7e20b273e8R84

I updated the code to potentially up-sizing new_roots multiple times (at the expense of initially allocating max_new_roots elements... this will likely to be over-allocating in most cases, but I think this will be OK as this will unlikely to be a memory footprint bottleneck).

Let me know if you have further concerns.

aschaffer (approved):

cpp/src/components/weakly_connected_components.cu
```diff
@@ -0,0 +1,741 @@
...
+                "GraphViewType::vertex_type should be integral.");
+  static_assert(!GraphViewType::is_adj_matrix_transposed,
+                "GraphViewType should support the push model.");
+
+  auto const num_vertices = push_graph_view.get_number_of_vertices();
+  if (num_vertices == 0) { return; }
+
+  // 1. check input arguments
+
+  CUGRAPH_EXPECTS(
```
ChuckHastings, line 234: Should this check be here, or would it be better up a level?

Thought being... SCC can be implemented as calling WCC on a directed graph, and then reversing the edges (transposing the matrix) and calling a customized WCC on the reversed/transposed graph.  Pushing this check up to the weakly_connected_components function would make that easier.

Of course, we can always do that when we get around to SCC.

ChuckHastings (approved):
LGTM

cpp/src/components/weakly_connected_components.cu
seunghwak: > Should this check be here, or would it be better up a level?
> 
Doesn't matter much, this can be here or one level up. But my rational for placing this check here is to avoid the upper level function simple (which is just a caller to the actual implementation function). Based on algorithms, checking for input arguments (especially when do_expensive_check is enabled) can become pretty lengthy.

And I haven't thought much about the behavior of this function if the input graph is asymmetric... or using this WCC as part of SCC implementation. I think we can revisit this later. Designing efficient SCC algorithm requires some additional thinking.

BradReesWork:
@gpucibot merge

PR #1609, opened 2021-05-21
Fea rmm device buffer change

ChuckHastings:
RMM is going to require specifying the stream in the device_buffer constructor.

This changes the one `cugraph` item and it updates to a version of `cuhornet` that makes the necessary changes.


seunghwak (approved):

BradReesWork:
@gpucibot merge

PR #1615, opened 2021-05-24
Undo disabling MG C++ testing outputs for non-root processes

seunghwak:
Disabling MG C++ testing outputs for non-root processes has an undesirable side effect of disabling error messages (e.g. exception outputs) as well. This makes failure diagnosis difficult in large-scale runs.

Delete the code disabling MG C++ testing outputs for non-root processes in this PR. In long run, we need a more elegant mechanism to disable only redundant test progress messages while still allowing error messages to be displayed.


seunghwak:
Submitted https://github.com/rapidsai/cugraph/issues/1616 for a longer term solution to disable only redundant test progress messages.

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1617, opened 2021-05-24
Pass rmm memory allocator to cuco::static_map

seunghwak:
Currently cuco::static_map instances in cuGraph are initialized with the cuco::static_map default allocator (which uses CUDA memory allocator).

Update to provide the RMM memory allocator instead.


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1607, opened 2021-05-19
Handle int64 in force atlas wrapper and update to uvector 

hlinsen:
- Update wrapper to handle int64
- Use `device_uvector`

Close: https://github.com/rapidsai/cugraph/issues/1581

PR #1625, opened 2021-05-25
[FIX] Initialize force_atlas2 `old_forces` device_uvector, use new `rmm::exec_policy`

trxcllnt:
Fixes an issue where the layout for all points converge to `(0,0)` when running multiple layout ticks after https://github.com/rapidsai/cugraph/pull/1607.

* Initializes the `d_old_forces` vector to fix layout issue.
* Updates to `barnes_hut.hpp` to use non-deprecated `rmm::exec_policy`.

### DIFF
--- a/cpp/src/layout/barnes_hut.hpp
+++ b/cpp/src/layout/barnes_hut.hpp
@@ -89,7 +89,7 @@ void barnes_hut(raft::handle_t const &handle,
   rmm::device_uvector<int> d_childl((nnodes + 1) * 4, stream);
   // FA2 requires degree + 1
   rmm::device_uvector<int> d_massl(nnodes + 1, stream);
-  thrust::fill(rmm::exec_policy(stream), d_massl.begin(), d_massl.end(), 1.f);
+  thrust::fill(rmm::exec_policy(stream), d_massl.begin(), d_massl.end(), 1);
 
   rmm::device_uvector<float> d_maxxl(blocks * FACTOR1, stream);
   rmm::device_uvector<float> d_maxyl(blocks * FACTOR1, stream);
--- a/cpp/src/layout/barnes_hut.hpp
+++ b/cpp/src/layout/barnes_hut.hpp
@@ -151,6 +151,8 @@ void barnes_hut(raft::handle_t const &handle,
   swinging   = d_swinging.data();
   traction   = d_traction.data();
 
+  thrust::fill(rmm::exec_policy(stream)->on(stream), d_old_forces.begin(), d_old_forces.end(), 0.f);
+
   // Sort COO for coalesced memory access.
   sort(graph, stream);
   CHECK_CUDA(stream);
--- a/cpp/src/layout/barnes_hut.hpp
+++ b/cpp/src/layout/barnes_hut.hpp
@@ -16,8 +16,8 @@
 
 #pragma once
 
-#include <rmm/thrust_rmm_allocator.h>
 #include <rmm/device_uvector.hpp>
+#include <rmm/exec_policy.hpp>
 
 #include <converters/COOtoCSR.cuh>
 #include <utilities/graph_utils.cuh>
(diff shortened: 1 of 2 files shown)


cpp/src/layout/barnes_hut.hpp
```diff
@@ -89,7 +89,7 @@ void barnes_hut(raft::handle_t const &handle,
   rmm::device_uvector<int> d_childl((nnodes + 1) * 4, stream);
   // FA2 requires degree + 1
   rmm::device_uvector<int> d_massl(nnodes + 1, stream);
-  thrust::fill(rmm::exec_policy(stream)->on(stream), d_massl.begin(), d_massl.end(), 1.f);
+  thrust::fill(rmm::exec_policy(stream), d_massl.begin(), d_massl.end(), 1.f);
```
ChuckHastings, line 92: Quick question.  Never noticed this in earlier reviews of this code.

Is it intentional that we use `float` for these values when the graph might have `float` or `double` input types?  Should we be templating these `float` instances to be `weight_t` do avoid the loss of value in conversions, or is it really the case that a graph with `double` weights should be using `float` for these values being computed?

I don't know enough about the context of this algorithm to know the answer, please enlighten.

trxcllnt: Looks like edge weights are only used in `apply_attraction()` which [discards](https://github.com/rapidsai/cugraph/blob/branch-21.06/cpp/src/layout/fa2_kernels.hpp#L55) the extra precision, so I don't think it matters.

It looks like we should be filling with `1` instead of `1.0f` though, because `d_massl` is a device_uvector of integers.

ChuckHastings: OK.  We should probably have a bigger conversation about this, but I don't think it should hold up progress here.

Could you add the following circa line 94:
```
// FIXME: Review whether these should be "float" or "weight_t"
```

That way we can come back at some point (perhaps when we refactor with the new graph) and revisit this.

trxcllnt: Two main reasons why I'm against it:

1. Double-precision arithmetic in the force integration kernels probably isn't worth the performance hit.
2. A user could accidentally opt in to a much slower implementation by accidentally using FP64 weights.

cpp/src/layout/barnes_hut.hpp
```diff
@@ -151,12 +151,14 @@ void barnes_hut(raft::handle_t const &handle,
   swinging   = d_swinging.data();
   traction   = d_traction.data();
 
+  thrust::fill(rmm::exec_policy(stream), d_old_forces.begin(), d_old_forces.end(), 0.f);
```
hlinsen, line 154: Thanks Paul. I think we should add it in the exact version also

trxcllnt: Updated!

cpp/src/layout/exact_fa2.hpp
```diff
@@ -64,9 +64,10 @@ void exact_fa2(raft::handle_t const &handle,
   rmm::device_uvector<float> repel(n * 2, stream);
   rmm::device_uvector<float> attract(n * 2, stream);
   rmm::device_uvector<float> old_forces(n * 2, stream);
+  thrust::fill(rmm::exec_policy(stream), old_forces.begin(), old_forces.end(), 0.f);
   // FA2 requires degree + 1.
   rmm::device_uvector<int> mass(n, stream);
-  thrust::fill(rmm::exec_policy(stream)->on(stream), mass.begin(), mass.end(), 1.f);
+  thrust::fill(rmm::exec_policy(stream), mass.begin(), mass.end(), 1.f);
```
hlinsen, line 70: We can pass int here as well. Otherwise lgtm, thanks for correcting my mistake

hlinsen (approved):

trxcllnt:
rerun tests

trxcllnt:
@gpucibot merge

cpp/src/layout/barnes_hut.hpp
```diff
@@ -52,9 +52,9 @@ void barnes_hut(raft::handle_t const &handle,
                 bool verbose                                  = false,
                 internals::GraphBasedDimRedCallback *callback = nullptr)
 {
-  cudaStream_t stream = handle.get_stream();
-  const edge_t e      = graph.number_of_edges;
-  const vertex_t n    = graph.number_of_vertices;
+  rmm::cuda_stream_view stream(handle.get_stream());
```
afender, line 55: Seems like we should use get_stream_view() for this, no? https://github.com/rapidsai/raft/blob/d5763480b899aa572afa92432fdf5bd6626a98a0/cpp/include/raft/handle.hpp#L102 
There are a couple of other occurrences in the PR

afender (approved):

trxcllnt:
@gpucibot merge

afender:
@gpucibot merge

PR #1633, opened 2021-05-28
transform_reduce_e bug fixes

ChuckHastings:
Discovered this bug during louvain testing.  The variable `idx` is an index controlling iteration over a subset of the vertices, while `major_offset` identifies a particular vertex offset in the overall data structure.  Several places where using `idx` where they needed to be using `major_offset`.


seunghwak (approved):
Thanks for finding & fixing this!!!

aschaffer (approved):

BradReesWork:
@gpucibot merge

Issue #1538, opened 2021-04-15
[ENH] RMM exec_policy


ChuckHastings:
RMM added a new feature in 0.18 that simplifies the `exec_policy` process.

We should modify graph code that uses `rmm::exec_policy` to include `#include <rmm/exec_policy.hpp>` instead of getting the version from `#include <rmm/thrust_rmm_allocator.h>` and use the simpler code.

For example:
```
// old code
#include <rmm/thrust_rmm_allocator.h>

...
thrust::copy(rmm::exec_policy(stream)->on(stream),
...
```
and the new version would look like:
```
// new code
#include <rmm/exec_policy.hpp>

...
thrust::copy(rmm::exec_policy(stream_view),
...
```

This is simpler and I imagine that the old technique will be deprecated eventually.

ChuckHastings:
Closed by #1707 and the corresponding raft changes.

Issue #1422, opened 2021-02-22
[ENH] update and refactor all cuGraph to use rmm's new cuda_stream_view


seunghwak:
**Describe the solution you'd like**
RMM added cuda_stream_view (https://github.com/rapidsai/rmm/blob/branch-0.19/include/rmm/cuda_stream_view.hpp#L34)

This wraps a native CUDA stream with convenience functions, and we should switch to this new feature.

```
CUDA_TRY(cudaStreamSynchronize(handle.get_stream()));
=>
handle.get_stream_view().synchronize()
```

```get_stream()``` call on raft::handle_t should be replaced with ```get_stream_view()```

and thrust execution policy

```rmm::exec_policy(handle_ptr_->get_stream())->on(handle_ptr_->get_stream())```

can now be ```rmm::exec_policy(handle.get_stream_view())``` for example.

PR #1653, opened 2021-06-07
Fea cleanup stream part1

ChuckHastings:
Addresses part of #1538 and #1422 

Breaking the work for these issues into smaller chunks to make it easier to test, review and approve.


seunghwak:
Is this PR pretty much complete (and just waiting for CI issues to be resolved) or you have more works to do?

I am worried that I will see a crazy number of merge conflicts if I try to merge my DCSR PR with this PR (both will touch many files in common).

If this PR is nearly complete, I can pull this to my PR first to avoid resolving merge conflicts later.

ChuckHastings:
I was going to do no more in this PR unless there were some review comments.  There's a bunch more files to do this with, but I was going to do them all in batches like this.

Just waiting for the CI issues to be resolved to request reviews.

cpp/include/cugraph/dendrogram.hpp
```diff
@@ -27,10 +27,11 @@ class Dendrogram {
  public:
   void add_level(vertex_t first_index,
                  vertex_t num_verts,
-                 cudaStream_t stream,
+                 rmm::cuda_stream_view const &stream_view,
```
seunghwak, line 30: Should we pass a const reference here?

Considering that
1) cuda_stream_view is just a wrapper for cudaStream_t.
https://github.com/rapidsai/rmm/blob/branch-21.08/include/rmm/cuda_stream_view.hpp#L108
2) and cuDF passes cuda_stream_view by value
https://github.com/rapidsai/cudf/blob/branch-21.08/cpp/include/cudf/detail/copy.hpp#L76
https://github.com/rapidsai/cudf/blob/branch-21.08/cpp/include/cudf/detail/groupby.hpp#L47

seunghwak (approved):
Not something critical, but we may need to review whether should we pass a cuda_stream_view object by value or const reference.

Besides that, looks good to me.

cpp/include/cugraph/dendrogram.hpp
ChuckHastings: Pushed the change to pass by value.

aschaffer (approved):

BradReesWork:
@gpucibot merge

PR #1673, opened 2021-06-14
Fea hungarian expose precision

### LINKED ISSUE #1645: [QST] Make the precision of the Hungarian algorithm controllable
Hi! Is there any way to manually control the precision of the Hungarian algorithm?  I checked the [RAFT library](https://github.com/rapidsai/raft/blob/a6f3e75938f4c4b607eadc4da05af4adf183b7e4/cpp/include/raft/lap/lap_kernels.cuh#L63) and it is currently fixed to 1e-6. I found this value not accurate enough for large-size dense assignment problems.

### LINKED ISSUE #1646: [ENH] Hungarian algorithm for rectangular matrix
Hi! I found that the current implementation of the Hungarian algorithm [does not support rectangular cost matrices](https://github.com/rapidsai/cugraph/blob/637c13947fb443d42985a62a7ec904066b267224/cpp/src/linear_assignment/hungarian.cu#L47). This problem is easy to solve by adding dummy rows/columns to the cost matrix without modifying the Hungarian algorithm itself. For example:

```python
import numpy as np
from scipy.optimize import linear_sum_assignment

if __name__ == '__main__':

    # A rectangular cost matrix
    cost = np.random.rand(10, 15)
    true_size = min(cost.shape)

    # solve for the rectangular LAP
    r1, c1 = linear_sum_assignment(cost)

    # add in dummy rows to make rectangular matrix square
    dummy = cost.max() * np.ones((cost.shape[1] - cost.shape[0], cost.shape[1]))
    cost_new = np.concatenate((cost, dummy), axis=0)
    r2, c2 = linear_sum_assignment(cost_new)
    r2, c2 = r2[:true_size], c2[:true_size]

    print(c1)
    print(c2)


```

ChuckHastings:
Closes #1645
Closes #1646 

Expose the precision parameter (epsilon in the Date/Nagi implementation) of the Hungarian algorithm to be controllable by the user.  Add support for rectangular matrices.

Will be enabled for CI after https://github.com/rapidsai/raft/pull/275 is merged.


ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

aschaffer (approved):

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1052,6 +1086,38 @@ weight_t hungarian(raft::handle_t const &handle,
...
+ * @param[in]  precision             parameter to define precision of comparisons
+ *                                   in reducing weights to zero.
+ */
+template <typename vertex_t, typename weight_t>
+weight_t hungarian(raft::handle_t const &handle,
+                   weight_t const *costs,
+                   vertex_t num_rows,
+                   vertex_t num_columns,
+                   vertex_t *assignment,
+                   weight_t precision);
```
seunghwak, line 1119: IIRC, we used epsilon in raft and we're using precision here, I got they are two different functions, but both epsilon and precision  are used to evaluate equality.... so better be consistent?

BradReesWork (approved):

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1052,6 +1086,38 @@ weight_t hungarian(raft::handle_t const &handle,
...
+ *                                   which vertex id (job) is assigned to that worker
+ * @param[in]  precision             parameter to define precision of comparisons
+ *                                   in reducing weights to zero.
+ */
+template <typename vertex_t, typename weight_t>
+weight_t hungarian(raft::handle_t const &handle,
+                   weight_t const *costs,
+                   vertex_t num_rows,
+                   vertex_t num_columns,
+                   vertex_t *assignment,
```
seunghwak, line 1118: assignment"s"? (as this is an array of values and we're using cost"s", without "s", one may think cost"s" is an array and assignment is a single value).

cpp/src/linear_assignment/hungarian.cu
```diff
@@ -35,28 +37,82 @@
...
+    //  Fill the extra rows/columns with max(d_original_cost)
+    //
+    index_t n         = std::max(num_rows, num_cols);
+    weight_t max_cost = thrust::reduce(rmm::exec_policy(handle.get_stream_view()),
+                                       d_original_cost,
+                                       d_original_cost + (num_rows * num_cols),
+                                       weight_t{0},
+                                       thrust::maximum<weight_t>());
+
+    rmm::device_uvector<weight_t> tmp_cost(n * n, handle.get_stream_view());
```
seunghwak, line 88: Why use _v for tmp_row_assignment_v and tmp_col_assignment_v but no _v for tmp_cost? IIRC, you use _v for vectors and better be consistent at least within a code block.

seunghwak (commented):
I have some minor comments about variable naming, but besides theses complaints, looks good to me.

kaatish (approved):

BradReesWork:
@gpucibot merge

PR #1681, opened 2021-06-21
Delete legacy renumbering implementation

ChuckHastings:
This code is no longer used.


seunghwak (approved):

seunghwak:
Why's this breaking?

ChuckHastings:
I labeled it breaking because it deletes a function from `functions.hpp`.  Anyone that wrote code that depended on that function will break.

It was unclear to me which things that break outside users warrant that label.  It doesn't break any python users... only if we have users of our legacy C++ API.

BradReesWork (approved):

aschaffer (approved):

Iroy30 (approved):
LGTM.

BradReesWork:
@gpucibot merge

PR #1683, opened 2021-06-22
Use the 21.08 branch of rapids-cmake as rmm requires it

robertmaynard:
Now that rmm uses rapids-cmake we need to update to the 21.08 branch to get the new `rapids_cmake_write_version_file` function


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1685, opened 2021-06-23
[REVIEW] Improve Random Walks performance

aschaffer:
This tracks work on adding additional Random Walks traversal strategy: horizontal. This strategy is more efficient, but requires more GPU memory.

The previous strategy (vertical) is only used when not enough GPU memory is available to support the horizontal one. The decision on which to use is based on GPU memory availability: if enough memory is available then the more performant (horizontal) strategy is used.


cpp/src/sampling/random_walks.cuh
```diff
@@ -1058,19 +1102,60 @@ random_walks(raft::handle_t const& handle,
...
+
+  // GPU memory requirements:
   //
-  return std::make_tuple(std::move(std::get<0>(quad_tuple)),
-                         std::move(std::get<1>(quad_tuple)),
-                         std::move(std::get<2>(quad_tuple)));
+  size_t coalesced_v_count = num_paths * max_depth;
+  auto coalesced_e_count   = coalesced_v_count - num_paths;
+  size_t req_mem_common    = sizeof(vertex_t) * coalesced_v_count +
+                          sizeof(weight_t) * coalesced_e_count +  // coalesced_v + coalesced_w
```
kaatish, line 1125: I don't understand the comments here without the context. Is it dead code or are you explaining something here?

cpp/src/sampling/rw_traversals.hpp
```diff
@@ -0,0 +1,297 @@
...
+  template <typename graph_t,
+            typename random_walker_t,
+            typename index_t,
+            typename real_t,
+            typename seed_t>
+  void operator()(
+    graph_t const& graph,                // graph being traversed
+    random_walker_t const& rand_walker,  // random walker object for which traversal is driven
+    seed_t seed0,                        // initial seed value
+    device_vec_t<typename graph_t::vertex_type>& d_coalesced_v,  // crt coalesced vertex set
```
kaatish, line 124: crt implies current here?

aschaffer: Yes.

cpp/src/sampling/random_walks.cuh
```diff
@@ -1058,19 +1112,60 @@ random_walks(raft::handle_t const& handle,
...
-  // ignore last element of the quad, seed,
-  // since it's meant for testing / debugging, only:
+  // GPU memory availability:
   //
-  return std::make_tuple(std::move(std::get<0>(quad_tuple)),
-                         std::move(std::get<1>(quad_tuple)),
-                         std::move(std::get<2>(quad_tuple)));
+  size_t free_mem_sp_bytes{0};
+  size_t total_mem_sp_bytes{0};
+  cudaMemGetInfo(&free_mem_sp_bytes, &total_mem_sp_bytes);
```
ChuckHastings, line 1128: Is this going to work if we're using RMM pool memory?  What about if we're using unified memory?

Do we have use cases that drive wanting to use this optimization?  I guess my thought is this... is it better to solve this at a higher level in the call stack?  I worry about the maintainability of code where the low level functions make decisions based on memory availability.  We already ran into an issue where a low level function had code like this and when we migrated to a newer GPU with more memory that code no longer worked.

aschaffer: I don't know. I asked the same question on our team channel. No answer. BTW, similar code is already in use in triangle counting:
```
./src/community/triangles_counting.cu:759:  cudaMemGetInfo(&free_bytes, &total_bytes);
./src/community/triangles_counting.cu:794:  cudaMemGetInfo(&free_bytes, &total_bytes);
```
Why wasn't that a problem there but it's a problem here?

Previous comments (on slack) suggested _not_ to use environment variables. Which is why I've chosen this approach. Migrating to other GPUs that use a completely different API would actually affect the entire code base.

ChuckHastings: In fact the `triangle_counting` example is exactly the code that had to be hot fixed in release 20.6 to fix an issue when we migrated to the Ampere chip.  I fixed the erroneous computation to allow that code to continue to work.  I think it *is* a problem there, but we haven't fixed it yet.  I left the code that way because we'll be rewriting triangle counting this summer and I don't believe we'll have this feature in the new code.

I guess my thought was to perhaps send in the choice of whether to do horizontal or vertical traversal as a parameter to the rw algorithm.  Perhaps we set the default to be horizontal but allow the user to specify vertical if they would rather have it consume less memory (and be slower).

cpp/src/sampling/random_walks.cuh
aschaffer: I am explaining how the `req_mem_common` was calculated.

cpp/src/sampling/random_walks.cuh
aschaffer: https://nvidia.slack.com/archives/CV0B740TW/p1624394626285100

aschaffer: Yes, _eventually_ a special runtime parameter might be added to leave the decision on the user (although I find that questionable: if they choose horizontal but not enough memory is available...then what? They'll probably open a bug).

But anyway, adding such a parameter would affect the Python API. And for now the decision was made to leave the Python API in place so that Xin Huang can use it, since there currently are more urgent priorities.

aschaffer: If we want to insulate ourselves from CUDA low level API changes and renaming (which should not affect such routines as `cudaMemGetInfo()`, imho) then the solution is to provide an _adapter_ class (or suite of classes) through which we can query low level GPU hardware settings. The adapter would hide the low level code. And we would call the adapter(s) members, which we control, so we wouldn't need to change them as frequently (if ever).

_But_ such low level query routines are necessary. I doubt we can make the assumption that we can completely avoid them. That, to me, seems an unreasonable constraint.

ChuckHastings: I'm not a big fan of reasoning in hypotheticals.  Concrete use cases are a much better engineering tool for understanding these kind of decisions.  I imagine we can construct a wide variety of hypothetical situations that would compel a different design decision.

I'm perfectly comfortable saying that if a user can choose between horizontal and vertical and there's not enough memory to do horizontal but that's what they choose then the program throws an exception and fails.  If the user asks us to create a graph that requires 64 GB of memory on a GPU that only has 32 GB of memory then we will fail.  I don't see a huge difference other than the fact that it's less obvious that they would not have enough memory.  If they want to open a bug then we'll respond that they should not ask for us to use too much memory.  We can, perhaps, put more detail in documentation to help them make a better parameter choice, but ultimately they have the best information of the total situation to make the choice.

We could expose the parameter through the cython layer and let python default it.  That wouldn't change the python API but would make it easier to expose it later if desired.

I also agree we shouldn't spend a great deal more time on this.  I'm just concerned that as soon as someone turns on RMM memory pools or uses unified memory that this computation will likely generate the wrong decision.  I don't know that we have a good solution in those situations.  Making it be a parameter would allow this to work efficiently in all of these situations, although it would require the user to make a decision that we're not sure how to guide them through yet.

ChuckHastings (approved):
We can continue discussion/changes related to the `cudaGetMemInfo` call separately and address anything we decide to do in a separate PR.  Otherwise I think this looks good.

cpp/src/sampling/rw_traversals.hpp
aschaffer: Hey @kaatish , can this be resolved so we can converge towards an approval? TIA.

cpp/src/sampling/random_walks.cuh
aschaffer: Hey @kaatish , can this be resolved so we can converge towards an approval? TIA.

kaatish (approved):

BradReesWork:
@gpucibot merge

PR #1687, opened 2021-06-24
[FIX] `CHECK_CUDA` macros in debug builds

trxcllnt:


cpp/src/utilities/graph_utils.cuh
```diff
@@ -180,7 +180,7 @@ void scatter(size_t n, T *src, T *dst, M *map)
                   thrust::device_pointer_cast(src + n),
                   thrust::device_pointer_cast(map),
                   thrust::device_pointer_cast(dst));
-  CHECK_CUDA(stream);
+  CHECK_CUDA(stream.value());
```
ChuckHastings, line 183: Shouldn't these next 4 be `stream_view.value()`?

seunghwak (approved):
We're switching from cudaStream_t to rmm::cuda_stream_view, so I think we need to eventually update CHECK_CUDA to take rmm::cuda_stream_view, but at this moment, this looks like a proper fix.

kaatish (approved):
LGTM

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1692, opened 2021-06-30
Louvain on int64 crashes

ChuckHastings:
Running this unit test results in:
```
unknown file: Failure
C++ exception with description "parallel_for failed: cudaErrorInvalidDeviceFunction: invalid device function" thrown in the test body.
```


seunghwak:
https://github.com/rapidsai/cugraph/pull/1696 should replace this.

PR #1694, opened 2021-07-03
[FIX] Always add `faiss` library alias if it's missing

trxcllnt:
Always add the `FAISS::FAISS` library target alias if it doesn't exist. This can happen if cuML is built and installs FAISS before cuGraph.

https://github.com/rapidsai/cuml/pull/4028


ChuckHastings (approved):
Looks fine to me.

Why do we need FAISS?

trxcllnt:
@ChuckHastings it's a raft dependency, though now that you mention it looks like raft needs this change too. And if raft gets it, we can probably remove cuML's and cuGraph's `get_faiss.cmake` file.

BradReesWork:
@trxcllnt if we can put this in RAFT it would make more sense and solve both graph and ml.  Putting this PR on hold until you responde

trxcllnt:
@BradReesWork The RAFT PR is here: https://github.com/rapidsai/raft/pull/287. This commit was included in PR https://github.com/rapidsai/cugraph/pull/1698, so I'll close this one.

PR #1695, opened 2021-07-06
Fix the Louvain "invalid device function" error with 64 bit vertex IDs

seunghwak:
* Louvain currently fails with the 64 bit vertex IDs (thrust::copy with thrust::make_transform_iterator & device lambda returns cudaErrorInvalidDeviceFunction), this PR fixes this.


seunghwak:
This displays the updates in another PR, I will create a new PR instead.

PR #1698, opened 2021-07-07
[FIX] Use `library_dirs` for cython linking, link cudatoolkit libs, allow setting UCX install location

trxcllnt:
This PR is a continuation of https://github.com/rapidsai/cugraph/pull/1694. Similar to https://github.com/rapidsai/cuml/pull/4015, this PR updates setup.py to:

* Use `library_dirs` instead of `runtime_library_dirs` when linking Cython.
* Allow overriding UCX lib and include dirs via a `UCX_HOME` envvar.
* Link `cudart`, `cusparse`, and `cusolver`.

These are necessary to compile the Cython via `pip` when not inside a conda environment and when UCX is installed to a location other than `/usr` or `/usr/local`.


rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #1709, opened 2021-07-13
Updated for CUDA-specific py packages

### LINKED ISSUE #1708: [BUG] Need to create CUDA-specific conda packages for cugraph
Since the cugraph python package from conda now links against several CUDA-specific libs (from [this PR](https://github.com/rapidsai/cugraph/pull/1698)), we need CUDA-specific conda packages to ensure the conda solver installs the correct dependencies based on CUDA version.  This was the same situation faced by cuML and cuDF.

rlratzel:
Updated cugraph recipe to include CUDA version in build string and depend on a specific cudatoolkit based on CUDA version. This pattern was copied from cuML's recipe.

closes #1708


raydouglass (approved):

BradReesWork:
@gpucibot merge

PR #1710, opened 2021-07-13
ReduceV test

kaatish:
Added test for reduce_v primitive.


cpp/tests/prims/mg_reduce_v.cu
```diff
@@ -0,0 +1,220 @@
...
+    }
+
+    vertex_t const* data = (*d_mg_renumber_map_labels).data();
+    rmm::device_uvector<result_t> test_property(d_mg_renumber_map_labels->size(),
+                                                handle.get_stream());
+    thrust::transform(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+                      data,
+                      data + test_property.size(),
+                      test_property.begin(),
+                      property_transform<vertex_t, result_t>(hash_bin_count));
```
seunghwak, line 136: Better move this before  the performance measured segment (as we're measuring the performance of reduce_v, not thrust::transform or device_uvector creation).

seunghwak (commented):
Looks great!!!

I just left one comment, and besides that one, I have no complaint.

cpp/tests/prims/mg_reduce_v.cu
```diff
@@ -0,0 +1,220 @@
...
+      std::tie(sg_graph, std::ignore) =
+        input_usecase.template construct_graph<vertex_t, edge_t, weight_t, store_transposed, false>(
+          handle, true, false);
+      auto sg_graph_view         = sg_graph.view();
+      auto expected_vertex_count = thrust::transform_reduce(
+        rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+        thrust::make_counting_iterator(sg_graph_view.get_local_vertex_first()),
+        thrust::make_counting_iterator(sg_graph_view.get_local_vertex_last()),
+        property_transform<vertex_t, result_t>(hash_bin_count),
+        result_t{0},
```
seunghwak, line 162: Umm... yeah... and we may test with non-zero initial value to check this is properly working.

cpp/tests/prims/mg_reduce_v.cu
```diff
@@ -0,0 +1,220 @@
...
+    vertex_t const* data = (*d_mg_renumber_map_labels).data();
+    rmm::device_uvector<result_t> test_property(d_mg_renumber_map_labels->size(),
+                                                handle.get_stream());
+    thrust::transform(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+                      data,
+                      data + test_property.size(),
+                      test_property.begin(),
+                      property_transform<vertex_t, result_t>(hash_bin_count));
+    auto vertex_count =
+      reduce_v(handle, mg_graph_view, test_property.begin(), test_property.end(), result_t{0});
```
seunghwak, line 138: Umm... yeah... and we may test with non-zero initial value to check this is properly working.

cpp/include/cugraph/prims/reduce_v.cuh
```diff
@@ -54,7 +54,7 @@ T reduce_v(raft::handle_t const& handle,
   auto ret = thrust::reduce(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
                             vertex_value_input_first,
                             vertex_value_input_first + graph_view.get_number_of_local_vertices(),
-                            init);
+                            (handle.get_comms().get_rank() == 0) ? init : T{0});
```
seunghwak, line 57: This code will seg-fault in SG. No comms in SG. T{0} will not work if T is a thrust::tuple.

You may do something like

```
(GraphView::is_multi_gpu && handle.get_comms().get_rank() == 0) ? init : T{}
```
Instead.

cpp/include/cugraph/prims/reduce_v.cuh
```diff
@@ -86,8 +86,10 @@ T reduce_v(raft::handle_t const& handle,
            InputIterator input_last,
            T init)
 {
-  auto ret = thrust::reduce(
-    rmm::exec_policy(handle.get_stream())->on(handle.get_stream()), input_first, input_last, init);
+  auto ret = thrust::reduce(rmm::exec_policy(handle.get_stream())->on(handle.get_stream()),
+                            input_first,
+                            input_last,
+                            (handle.get_comms().get_rank() == 0) ? init : T{0});
```
seunghwak, line 92: Same here.

cpp/include/cugraph/prims/reduce_v.cuh
seunghwak: And reduce_v should work for tuple of arithmetic types as well, so I think we'd better add a test case with thrust::zip_iterator (of arithmetic types).

kaatish: I am not sure that reduce_v works with a zip iterator.
This is what I get 
```
thrust-src/thrust/functional.h(234): error: no operator "+" matches these operands
            operand types are: const thrust::tuple<int, int, float, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type> + const thrust::tuple<int, int, float, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>
          detected during:
            instantiation of "T thrust::plus<T>::operator()(const T &, const T &) const [with T=thrust::tuple<int, int, float, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type, thrust::null_type>]" 
```
My understanding is that the call to thrust::reduce inside reduce_v implicitly tries to instantiate the thrust::plus<thrust::tuple<...>> which does not make sense because operator+ is not defined for thrust::tuple as far as I am aware.

seunghwak: Oh, yes, you're right. Let's submit a cuGraph issue to fix this.

seunghwak: Or if you want to fix this in this PR, you can refer to 

https://github.com/rapidsai/cugraph/blob/branch-21.08/cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh#L127

```plus_edge_op_result``` should work with thrust::tuple.

File: cpp/include/cugraph/prims/reduce_v.cuh [hunk 6]
```diff
@@ -27,6 +27,33 @@
...
+ public:
+  __device__ constexpr auto operator()(const Type& t1, const Type& t2)
+  {
+    return sum_impl(t1, t2, std::make_index_sequence<thrust::tuple_size<Type>::value>());
+  }
+};
+
+template <typename... Args>
+struct ValueAdd<std::tuple<Args...>> : public ValueAdd<thrust::tuple<Args...>> {
+};
```
seunghwak, line 56: Why did you implement ValueAdd instead of reusing the existing one in https://github.com/rapidsai/cugraph/blob/branch-21.08/cpp/include/cugraph/prims/edge_op_utils.cuh?

kaatish: The existing functionality plus_edge_op_result is not a functor so it does not work directly with the thrust call here. I would have had to wrap it around another struct. The functor being used internally plus_thrust_tuple could be used though.

ValueAdd implementation was an attempt to modernize it a bit.

seunghwak: Can't you just call ```plus_edge_op_result``` inside a functor? Why do we need ```sum_impl```?

And this functor should be used in transform_reduce_v as well, so we may place this functor elsewhere... (maybe edge_op_utils.cuh possibly renaming the file and functions to reflect that now these are used for both custom vertex & edge op outputs.

seunghwak: And I may better not strongly recommend using a device lambda here (as device lambdas often fail to work properly for very esoteric reasons), but doing something like

```
[]__device__(auto lhs, auto rhs) {
  return plus_edge_op_result(lhs, rhs);
}
```

might be a simpler option (if this works... and I guess this will work here). If you want to avoid using device lambdas, then we may go for a simple wrapper functor than trying to reinvent the existing functionality.

kaatish: > Can't you just call plus_edge_op_result inside a functor? Why do we need sum_impl?

Yes, that's what I was referring to here : I would have had to wrap it around another struct.

My point was that I would just be increasing the number of abstraction levels. The new functor would wrap the function plus_edge_op_result, which wraps a functor plus_thrust_tuple. I was hoping to replace the existing calls to plus_edge_op_result (wherever applicable) to this since it is slightly more compact and modern since it does not use the recursive variadic implementation but index sequence instead.

And yes perhaps move it to some other generalized location would be a good idea.

seunghwak: OK, then we should remove plus_edge_op_result rather than keeping both and make our codebase more redundant.

cpp/include/cugraph/prims/count_if_e.cuh
```diff
@@ -16,7 +16,7 @@
 #pragma once
 
 #include <cugraph/experimental/graph_view.hpp>
-#include <cugraph/prims/edge_op_utils.cuh>
+#include <cugraph/prims/property_op_utils.cuh>
```
seunghwak, line 19: Is this include necessary?

cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh
```diff
@@ -75,6 +75,7 @@ __global__ void for_all_major_for_all_nbr_hypersparse(
 
   auto dcs_nzd_vertex_count = *(matrix_partition.get_dcs_nzd_vertex_count());
 
+  property_add<T> edge_property_add;
```
seunghwak, line 78: ```property_add<T> edge_property_add;```=>```property_add<T> edge_property_add{};``` (and this appears in multiple places).

seunghwak (approved):
This looks great!!! I have few minor comments but beside those, this looks great to me.

cpp/include/cugraph/prims/count_if_e.cuh
kaatish: This file is using `cast_edge_op_bool_to_integer` which is in `property_op_utils.cuh` but that file itself is included in `transform_reduce_e.cuh` so I guess it can be removed.

seunghwak: > This file is using cast_edge_op_bool_to_integer which is in property_op_utils.cuh

I didn't realize this, and if it's the case, better directly include this file than relying on implementation details of transform_reduce_e.cuh.

ChuckHastings (approved):

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

BradReesWork:
@gpucibot merge

PR #1711, opened 2021-07-14
[REVIEW] Cascaded dispatch for type-erased API

aschaffer:
This enhancement PR provides the C++ infrastructure for a type-erased API to be exposed to C-only, Cython, and potentially other glue languages. This will be the first in a series of such API PRs to expose cugraph functionality to other languages.


cpp/CMakeLists.txt
```diff
@@ -204,6 +204,9 @@ add_library(cugraph SHARED
     src/components/weakly_connected_components.cu
     src/structure/create_graph_from_edgelist.cu
     src/utilities/host_barrier.cpp
+    src/visitors/graph_envelope.cpp
```
ChuckHastings, line 207: Something to think about: Where do we want this code to reside?

We've talked about making this a separate C-style API library.  It feels like that should have its own high level directory structure.  Perhaps something like: `src/api/visitors` and `include/cugraph/api`?  Just suggestions, I'm not sure what would work best.  But I think organizing the code so that all of the code for a particular library is grouped together will make the most sense.

cpp/include/cugraph/experimental/graph_view.hpp
```diff
@@ -228,9 +232,9 @@ size_t constexpr num_sparse_segments_per_vertex_partition{3};
 
 // Common for both graph_view_t & graph_t and both single-GPU & multi-GPU versions
 template <typename vertex_t, typename edge_t, typename weight_t>
-class graph_base_t {
+class graph_base_t : public graph_envelope_t::base_graph_t /*<- visitor logic*/ {
```
ChuckHastings, line 235: Not directly applicable to this line...

This (and much of the CD code) is in the experimental namespace.  It feels like this should go into a permanent namespace.  Again, not sure of a good name, `api` perhaps?  Probably should be consistent with whatever we name the directory.

cpp/include/cugraph/visitors/graph_envelope.hpp
```diff
@@ -0,0 +1,206 @@
...
+  //     p_impl_fact_.first->print();
+  //   else
+  //     throw std::runtime_error("ERROR: Implementation not allocated.");
+  // }
+
+  std::unique_ptr<base_graph_t> const& graph(void) const { return p_impl_fact_.first; }
+
+  std::unique_ptr<visitor_factory_t> const& factory(void) const { return p_impl_fact_.second; }
+
+  graph_envelope_t(DTypes vertex_tid,
```
ChuckHastings, line 102: Question to consider: Does hiding all parameters in the `erased_pack_t` make the code more or less maintainable?

I wonder if we would be better off (from an ease of use perspective) having separate SG and MG constructors for the graph_envelope_t, specifying all non-type-erased parameters explicitly and only including the type erased parameters in the erased_pack_t.

cpp/src/visitors/bfs_visitor.cu
```diff
@@ -0,0 +1,138 @@
+/*
```
ChuckHastings, line 1: Any chance this could be a .cpp file?  Think about what it might take to accomplish that.  I don't see any direct thrust calls here.

cpp/include/cugraph/visitors/bfs_visitor.hpp
```diff
@@ -0,0 +1,83 @@
...
+  return_t const& get_result(void) const override { return result_; }
+
+ private:
+  erased_pack_t& ep_;
+  return_t result_;
+};
+
+// wrapper:
+//
+return_t bfs_wrapper(graph_envelope_t const& g, erased_pack_t& ep);
```
ChuckHastings, line 81: Question to consider: Does hiding all parameters in the erased_pack_t make the code more or less maintainable?

Would it make sense to have the non-type-erased parameters explicitly included in the call and put the type erased parameters in the erased_pack_t?

ChuckHastings: Second question... should we consider defining the wrapper in a different file?  In the C++ API we define all algorithms in a single header file (`algorithms.hpp`).  Would it make sense to do the same so that the caller only needs to include a single file to access all algorithms?

ChuckHastings: Third question... does the suffix `_wrapper` actually add value?  If we move (as I suggested in a different comment) these into a separate namespace (e.g. cugraph::api), couldn't this just be `cugraph::api::bfs`?

cpp/include/cugraph/experimental/graph_view.hpp
aschaffer: Similar answer to above. However, I am opened to the idea of placing all this logic in a separate namespace. But for reasons discussed above, I would not call it `api`. Maybe `namespace visitor` or `namespace dispatch` ?

cpp/include/cugraph/visitors/graph_envelope.hpp
aschaffer: The idea for `graph_envelope` is to hide the template params on which the graph depends: `vertex_t, edge_t, weight_t, bool st, bool mg`. By "specifying  all non-type-erased parameters explicitly" you meant their names and number? Or, their types, too? If you meant types then that defeats the purpose of `graph_envelope_t`. 

If you meant just names and number, well that will create problems in the future: as we add new graph types and/or graph constructors, we'd need to do the same for `graph_envelope_t`. That, again, defeats the purpose of an "envelope". 

That being said, I agree we do need some hand-shake agreement / check between what's passed to the `graph_envelope_t` constructor and the correct graph constructor to be dispatched. I will think about that.

cpp/include/cugraph/visitors/bfs_visitor.hpp
aschaffer: Question 1: See answer here: https://github.com/rapidsai/cugraph/pull/1711#discussion_r670572920;

Question 2: Yes. Once I add at least one more wrapper I will move all into their own header.

Question 3: It probably does not (add any value). I think moving this into its own namespace and calling it by its proper name (`bfs()`) make sense. I'm not sure what to call this separate namespace, though. Maybe `namespace wrappers`?

aschaffer: Question 3 follow-up: yes, `cugraph::api` should work, but as said I'm planning to also add a `cugraph/api` directory. I'm not sure if that would confuse things.

cpp/src/visitors/bfs_visitor.cu
aschaffer: I will look into this.

aschaffer: Done.

cpp/include/cugraph/experimental/graph_view.hpp
ChuckHastings: Perhaps based on your above argument `namespace visitor` would be appropriate here.

cpp/include/cugraph/visitors/graph_envelope.hpp
ChuckHastings: I appreciate the goal of supporting type erasure.  I guess my thought was that parameter types that don't vary might be more clearly specified explicitly rather than hidden in the `erased_pack_t`.

The problem we're addressing, I believe, is that `vertex_t`, `edge_t` and `weight_t` need to be specified by some calling languages (python being the original motivator) dynamically.  If we look your C++ unit test (in `bfs_test.cpp`), we pass into the graph envelope constructor an erased pack (`ep_graph`) that contains things we need to type erase as well as things we don't really need to type erase.  We certainly *can* type erase everything, as this does.  My question is, do we think we have a *better* API by type erasing everything.  Type erasure eliminates the ability to check types at compile time, which loses something in the code quality.

Specifically, I would be concerned about:
* handle - this is a known type, not sure there's value in type erasing this
* num_vertices - this could be passed as a `size_t` to make it work for any value of vertex_t
* graph_props - this is a structure... I don't imagine we will have different versions of this structure that we want to support
* sorted and check are boolean values... again, don't imagine that we need to type erase those.

aschaffer: All of these would need to then be erased in the C-only API level (except maybe the booleans). Also, remember, num_vertices is `vertex_t`. So, you'd need an integer type large enough to cover all `vertex_t` types we support. So, probably `size_t`. Which still is type erasure (because it still needs to be casted to `vertex_t` inside, otherwise you're at th every least allocating too much; most likely would not compile). Not to mention C-only might not have a `size_t`. Etc.

So, since you'd have to erase the types of all these parameters anyway, I'd rather do it in a place where I actually _can_ retrieve their actual types; i.e., in the C++ code.

When calling from the C++ side, we don't need to go through this type-erasure mechanism (I'm only doing it to be able to test it, which is why you see that in my unit test).

aschaffer: More importantly, however, is that each algorithm call would require bolierplate code to materialize the type-erasure (specifically, casts). Well, that's exactly what the type-erased pack `ep` does under the hood. Instead of delegating that code to the caller `erased_pack_t` does the actual erasure of types for them, transparently.

ChuckHastings: Again, I'm not disparaging the `erased_pack_t` approach.  This works very well for much of what we're putting through the interface.

My concern is whether *all* things should be type erased in this fashion.  boolean's don't need to be type erased at all.  graph_props probably doesn't need to be type erased at all.  I hadn't thought about the handle class and the C issue... clearly that one must be type erased.  Promoting vertex_t to a size_t and casting back to a smaller size is a common C idiom... although as you observe is also type erasure - just via a different mechanism.

I'm OK going down with this approach.  I'm just raising the question to be sure we're comfortable that we've thought about the implications.  The caller from C will have to meet this API, I'm just questioning whether forcing everything to be type erased (even if they don't technically need it) is the best option for that caller.  Will it be intuitive - with adequate documentation - to call the C-only api, or will it be awkward and confusing even with good documentation?

Perhaps we go with this approach and revisit the decision later when we have more concrete examples.  It's relatively easy to move parameters around if we find it's more intuitive later to structure things differently.

aschaffer: Resolving for now.

cpp/include/cugraph/experimental/graph_view.hpp
aschaffer: Done.

cpp/include/cugraph/visitors/bfs_visitor.hpp
aschaffer: Done: Isolated API in its own header; added `cugraph::api` namespace;

cpp/CMakeLists.txt
aschaffer: Done (most of it + see latest discussions / decisions on C-API)

ChuckHastings (approved):

aschaffer:
rerun tests.

kaatish (approved):
LGTM

BradReesWork:
@gpucibot merge

PR #1721, opened 2021-07-20
[REVIEW] revert cuco to latest dev branch, issues should be fixed

ChuckHastings:
We had an issue in CI that uncovered the fact that raft is referencing the `dev` branch of coco while `cugraph` has it pinned to a previous commit.  CI is using the newer `dev` branch version, which doesn't match what happens when we do local builds.

The original motivation for pinning was a temporary fix because of a bug introduced in `coco` which has since been resolved.  This PR reverts us back to using the `dev` branch.


ChuckHastings:
rerun tests

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1731, opened 2021-07-23
Updated MG python tests to run in single and multi-node environments

rlratzel:
* Removed all `client_connection` module fixtures and replaced with a single `dask_client` session fixture which detects a single or multi-node environment and creates the appropriate cluster/client.
* Removed all hardcoded references to `../datasets` in MG tests and replaced with `RAPIDS_DATASET_ROOT_DIR`-based paths.
* Updated `npartitions` to use the number of workers from the client instead of just visible devices on the node.
* Added `FIXME`s
* Added option to skip building cpp gtests, which speeds up the nightly builds for benchmarks and python tests that don't use the cpp gtest binaries.


python/cugraph/tests/utils.py
```diff
@@ -37,41 +37,40 @@
 CUPY_MATRIX_TYPES = [cp_coo_matrix, cp_csr_matrix, cp_csc_matrix]
 SCIPY_MATRIX_TYPES = [sp_coo_matrix, sp_csr_matrix, sp_csc_matrix]
 
+RAPIDS_DATASET_ROOT_DIR = os.getenv("RAPIDS_DATASET_ROOT_DIR", "../datasets")
+RAPIDS_DATASET_ROOT_DIR_PATH = Path(RAPIDS_DATASET_ROOT_DIR)
```
jnke2016, line 41: I am just wondering why you replaced `PurePath` with `Path`. My understanding so far of `RAPIDS_DATASET_ROOT_DIR` is to allow the tests to find the datasets (no systems call). But `Path` provide method to do systems call on path object. Are we doing systems call or planning to do so?

rlratzel: I was thinking we might have test or setup code that wants to do things like `assert RAPIDS_DATASET_ROOT_DIR.is_dir()` or possibly `if (RAPIDS_DATASET_ROOT_DIR / "foo").exists(): ...`, which aren't available on a `PurePath` object.

jnke2016: Make sense

jnke2016 (approved):

Iroy30 (approved):

BradReesWork:
@gpucibot merge

PR #1732, opened 2021-07-24
[REVIEW] Biased Random Walks for GNN

aschaffer:
This tracks work on implementing Biased Random Walks for GNN.


cpp/include/cugraph/algorithms.hpp
```diff
@@ -1344,6 +1344,8 @@ extract_ego(raft::handle_t const& handle,
  * (compressed) format; when padding is used the output is a matrix of vertex paths and a matrix of
  * edges paths (weights); in this case the matrices are stored in row major order; the vertex path
  * matrix is padded with `num_vertices` values and the weight matrix is padded with `0` values;
+ * @param selector_type identifier for sampling strategy: uniform, biased, etc.; defaults to
```
ChuckHastings, line 1347: Should this be an enum?  We can have an enum for the public API that is separate from the internal (detail) enum if there's a reason not to expose the internal details.

If we're going to use an int, should the documentation identify the integer values that correspond to the current supported types.

cpp/src/sampling/rw_traversals.hpp
```diff
@@ -96,6 +104,243 @@ value_t const* raw_const_ptr(device_const_vector_view<value_t>& dv)
...
+              weight_t const* ptr_d_sum_weights)
+      : row_offsets_(ro), col_indices_(ci), values_(w), ptr_d_sum_weights_(ptr_d_sum_weights)
+    {
+    }
+
+    // pre-conditions:
+    //
+    // 1. (indices, weights) are assumed to be reshuflled
+    //    so that weights(neighborhood(src_v)) are ordered
+    //    increasingly; too expesnive to check this here;
```
ChuckHastings, line 286: `expesnive` should be `expensive`

Is this still a precondition?  I know we discussed how we probably don't need this precondition, but I'm not sure if this implementation still requires that.

cpp/include/cugraph/algorithms.hpp
aschaffer: I don't mind an `enum` but that enum would reside where? What header? We cannot place it in the implementation headers, because that would create circular header inclusions. It's also probably not that helpful to create a separate header only for that enum. 

I don't see any other such headers being included in `algorithms.hpp`. Also, there's only one other `enum` being defined in `algorithms.hpp`. And that is a scoped `enum`, which must be type-erased to an `int` anyway for either C-only API or Python. So, why not use an `int` directly? (at least in the public API that other languages must see and use)

(And, yes I'll add dox about the one other possible value for sampler selection, which is `1`. The default, `0` is already docummented).

cpp/src/sampling/rw_traversals.hpp
aschaffer: No, it's not a pre-condition anymore. I'll remove that comment.

cpp/include/cugraph/algorithms.hpp
BradReesWork: @aschaffer I agree that picking a place for the enum to be defined is an issue.  Maybe just define the int values in the docs and push enum definition to the python side for now.  We should have a later discussion on enums since this is not the only instance

cpp/src/sampling/rw_traversals.hpp
aschaffer: Done.

cpp/include/cugraph/algorithms.hpp
aschaffer: Done (documented better the integer values).

aschaffer: Resolving this for now.

ChuckHastings: Generally, the algorithms exposed in `algorithms.hpp` don't take enumerated types that control the behavior of the algorithm.  The only one that I know of (before this) is the legacy connected components implementation, which defines the enumerated type in this file.  Note that the new weakly connected components implementation (the MNMG version) is a separate entry point.  It is unclear, at this time, if the MNMG strongly connected component implementation will be a separate entry point or add the enumerated type to the weakly connected components implementation.

Other places where this might be appropriate we create separate entry points in the `algorithms.hpp` and reuse the code in the implementation.  For example, the spectral clustering algorithms are each exposed as separate entry points.  They could be exposed as a single entry point with an enumerated type identifying which variation to run.  I don't know what motivated the choice between approaches... probably left to the preference of the developer creating the algorithms.

I'm OK with just using an int as long as we define it in the documentation.

ChuckHastings (approved):

cpp/include/cugraph/algorithms.hpp
seunghwak: +1 for Chuck's idea to "create separate entry points in the algorithms.hpp and reuse the code in the implementation'', but yeah... just a recommendation.

cpp/include/cugraph/graph.hpp
```diff
@@ -176,6 +176,25 @@ class graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enab
...
+  // Notes:
+  // (1.) criterion is mutable (non-const)
+  //      to allow for sorter obejcts for which
+  //      the sorting operation fills additional structures
+  //      (later to be retrieved; e.g., for debugging);
+  // (2.) sorting object is responsible for updating "in-place"
+  //      any of the (offsets, indices, weights) arrays;
+  //
+  template <typename in_place_sorter_t>
+  void sort(in_place_sorter_t& criterion)
```
seunghwak, line 193: Is this function still necessary?

aschaffer: No, but since it's there, why remove it? It's a piece of functionality that might turn out to be necessary. At least useful. I am testing it.

seunghwak: So, sorting (based on specifics) will affect many other algorithms. For example, we may need to sort neighbors of each vertex to facilitate neighbor intersection (necessary for triangle counting and so on).

So, if we allow sorting of internal edge data (with the functor that allows sorting in many different ways), we should go for lengthy discussion about how this affects other algorithms.

Unless we want to go for that discussion in this PR, better delete this at this moment.

cpp/include/cugraph/algorithms.hpp
aschaffer: I'm not so sure about that. Unless I'm misunderstanding what's proposed here, when you have many moving parts (i.e., policies) the number of "separate entry points" would explode. For example, if we were to expose RW with (not even) all its moving parts: `random_engine_t` (let's say you have R possible choices), `traversal_t` (T possible choices), `sampling_t` (S possible choices); you would have to add function names for each combination, resulting in R*T*S "separate entry points". 

This is moving the named choice...into the name of the function. You're essentially unwrapping function instantiations (in the case of template functions, which is our case). Which really is compiler's job, not programmer's.

cpp/src/sampling/rw_traversals.hpp
```diff
@@ -96,6 +104,240 @@ value_t const* raw_const_ptr(device_const_vector_view<value_t>& dv)
...
+                                                                              real_t rnd_val) const
+    {
+      auto crt_out_deg = ptr_d_cache_out_degs_[src_v];
+      if (crt_out_deg == 0) return thrust::nullopt;  // src_v is a sink
+
+      real_t max_ub     = static_cast<real_t>(crt_out_deg - 1);
+      auto interp_vindx = rnd_val * max_ub + real_t{.5};
+      vertex_t v_indx   = static_cast<vertex_t>(interp_vindx);
+
+      auto col_indx  = v_indx >= crt_out_deg ? crt_out_deg - 1 : v_indx;
```
seunghwak, line 134: Assuming that ```rnd_val``` is a random number in [0, 1.0) and ```crt_out_deg``` is the vertex degree, this code wants to uniformly sample an integer in [0, vertex_degree - 1], right?

Say, the vertex degree is 3. Then, max_ub is 2.0 and interp_vindx will be in [0.5, 2.5). Considering that static_cast truncates, [0.5, 1.0) will be 0, [1.0, 2.0) will be 1, and [2.0, 2.5) will be 2. So, the index 1 has two times higher chance to be selected than 0 or 2, right? Or am I missing something?

cpp/include/cugraph/algorithms.hpp
aschaffer: Moreover, imagine over time you add one or more policies. Say, `neighborhood_discovery_t`. You must now go back and change the names of all the previous separate entry points, as opposed to just adding an additional `enum` or `int` argument. And while an additional argument might not break existent client code (if, for example, you make that argument optional); changing names on existent API certainly would (break caller's code).

cpp/src/topology/topology.cuh
```diff
@@ -45,25 +59,28 @@ namespace detail {
...
-  const IndexT* ptr_r_o = thrust::raw_pointer_cast(&d_row_offsets.front());
-  const IndexT* ptr_c_i = thrust::raw_pointer_cast(&d_col_indices.front());
-  BoolT* start_flags    = thrust::raw_pointer_cast(&d_flags.front());  // d_flags.begin();
-  BoolT* end_flags      = start_flags + nrows;
+  BoolT* start_flags = d_flags.data();  // d_flags.begin();
+  BoolT* end_flags   = start_flags + nrows;
   BoolT init{1};
   return thrust::transform_reduce(
-    thrust::device,
+    rmm::exec_policy(handle.get_stream_view()),
```
seunghwak, line 83: So in the below, you're checking only in case ```if (col_indx > row_indx) {```.

Is this sufficient?

Say ```row_indx``` is 3. Will this cover the case ```row_indx, col_indx``` = (3, 2) exists but (2, 3) missing?

cpp/include/cugraph/graph.hpp
BradReesWork: I agree with both of you.  this might be a useful feature later on, so preserving the code has benefit.  On the other hand, we do not want dead or unused code floating around in the code base.  

simply commenting it out for now, and adding a FIXME tag, is a good middle ground for now so that the PR can get merged.  We can then have a sorting sync and then either delete or improve the code.

cpp/src/sampling/random_walks.cuh
```diff
@@ -121,8 +112,12 @@ struct rrandom_gen_t {
   // if out_deg(v) > 0
   //   d_col_indx[v] = random index in [0, out_deg(v))
   //}
-  void generate_col_indices(device_vec_t<vertex_t>& d_col_indx) const
+  // d_crt_out_deg is non-coalesced;
+  //
+  void generate_col_indices(device_vec_t<edge_t> const& d_crt_out_deg,
+                            device_vec_t<vertex_t>& d_col_indx) const
   {
+    auto const* d_ptr_out_degs_ = d_crt_out_deg.data();
```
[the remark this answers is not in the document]
aschaffer: 1. That logic is unchanged from previous check-ins. In order to translate a number `rnd_val` from interval `[0.0, 1.0]` into a number `v_indx` in discrete set `{0,...,M=crt_out_deg-1}`, the logic is to scale `rnd_val` by `M`, `rnd_val*M` and then take either floor() function `(static_cast<int>(rnd_val*M)`) or ceiling ceil() function, `static_cast<int>(rnd_val*M + 0.5)`. I have opted for ceil(). The last statement comparison is just a sanity check.

But I think you're right, I either have to: (a) use floor(), or (b) change the "sanity check" into:
```
return (v_indx >= crt_out_deg ? 0 : v_indx);
```
I'll opt for (a), unless a different course of action is suggested.

2. BTW, this function is not used anymore for the overall RW algorithm(s). It's still there for convenience (debugging / low-level testing). This logic, however is replicated in the uniform sampler in another form

cpp/src/sampling/rw_traversals.hpp
aschaffer: Same as above:
"I either have to: (a) use floor(), or (b) change the "sanity check" into:

return (v_indx >= crt_out_deg ? 0 : v_indx);

I'll opt for (a), unless a different course of action is suggested."

cpp/src/topology/topology.cuh
aschaffer: So, this is an old function, to which I only changed the `exec` policy to make it compliant to `rmm`. I'd have to double check.

cpp/src/sampling/random_walks.cuh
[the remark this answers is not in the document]
aschaffer: Fixed. (removed the ceiling logic `0.5`)

cpp/src/sampling/rw_traversals.hpp
aschaffer: Fixed. (removed ceiling 0.5 logic).

cpp/include/cugraph/graph.hpp
aschaffer: Done. Commented out.

seunghwak (approved):

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1734, opened 2021-07-26
Use the new RAPIDS.cmake to fetch rapids-cmake

robertmaynard:
The original approach of using FetchContent naively has a subtle
bug when multiple projects that use rapids-cmake are combined together inside as sibling projects. This bug causes any
`include(rapids-*)` commands to fail, causing CMake errors.

Bug using `RAPIDS.cmake` we can resolve this issue and remove
the new complex logic from each consumer.


BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1742, opened 2021-07-28
Fea renumbering test

ChuckHastings:
Adds a C++ renumbering test for integer column renumbering.

Also refactored some of the test utilities (some tech debt from the original development of the test utilities).  Test graph construction now relies upon the edgelist construction calls instead of replicating functionality.


cpp/tests/experimental/renumbering_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+    std::vector<vertex_t> h_original_src_v{};
+    std::vector<vertex_t> h_original_dst_v{};
+    std::vector<vertex_t> h_final_src_v{};
+    std::vector<vertex_t> h_final_dst_v{};
+
+    rmm::device_uvector<vertex_t> src_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> dst_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> renumber_map_labels_v(0, handle.get_stream_view());
+    cugraph::experimental::partition_t<vertex_t> partition{};
```
seunghwak, line 77: Is this ever used?

seunghwak, line 77: Is this variable ever used?

ChuckHastings: Not in this SG test :-)

I'll remove it.

cpp/tests/experimental/renumbering_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+    std::vector<vertex_t> h_original_dst_v{};
+    std::vector<vertex_t> h_final_src_v{};
+    std::vector<vertex_t> h_final_dst_v{};
+
+    rmm::device_uvector<vertex_t> src_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> dst_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> renumber_map_labels_v(0, handle.get_stream_view());
+    vertex_t number_of_vertices{};
+    edge_t number_of_edges{};
+    bool symmetric{};
```
seunghwak, line 79: This variable is not used after the assignment in line 81. So, you can remove this variable and replace

```std::tie(src_v, dst_v, std::ignore, number_of_vertices, symmetric) =```
with
```std::tie(src_v, dst_v, std::ignore, number_of_vertices, std::ignore) =```

File: cpp/tests/experimental/renumbering_test.cpp [hunk 3]
```diff
@@ -0,0 +1,181 @@
...
+        renumber_map_labels_v.data(),
+        0,
+        static_cast<vertex_t>(renumber_map_labels_v.size()));
+      cugraph::experimental::unrenumber_local_int_vertices(
+        handle,
+        dst_v.data(),
+        dst_v.size(),
+        renumber_map_labels_v.data(),
+        0,
+        static_cast<vertex_t>(renumber_map_labels_v.size()));
```
seunghwak, line 123: This doesn't matter in SG, but if we consider MG extension, ```unrenumber_local_int_vertices``` works only with vertices in the range of ```[graph_view.get_local_vertex_first(), graph_view.get_local_vertex_last())```; so unrenumber_int_vertices (https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/include/cugraph/experimental/graph_functions.hpp#L212) could be more extensible.

Also note the comment in https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/include/cugraph/experimental/graph_functions.hpp#L185. If we need to unrenumber local edges in the performance critical path, we need to add addition unrenumber functions for better performance.

cpp/tests/utilities/test_graphs.hpp
```diff
@@ -163,11 +165,121 @@ class Rmat_Usecase : public detail::TranslateGraph_Usecase {
              bool>
   construct_edgelist(raft::handle_t const& handle, bool test_weighted) const
   {
-    // TODO: Tease through generate_graph_from_rmat_params
-    //       to extract the edgelist part
-    // Call cugraph::translate_vertex_ids(handle, d_src_v, d_dst_v, base_vertex_id_);
+    constexpr bool symmetric{true};
```
seunghwak, line 168: Do we always symmetrize R-mat graphs? This is necessary for WCC or Louvain, but there are many other algorithms that work with asymmetric R-mat graphs.

cpp/tests/experimental/renumbering_test.cpp
```diff
@@ -0,0 +1,182 @@
...
+    std::vector<vertex_t> h_original_dst_v{};
+    std::vector<vertex_t> h_final_src_v{};
+    std::vector<vertex_t> h_final_dst_v{};
+
+    rmm::device_uvector<vertex_t> src_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> dst_v(0, handle.get_stream_view());
+    rmm::device_uvector<vertex_t> renumber_map_labels_v(0, handle.get_stream_view());
+    cugraph::experimental::partition_t<vertex_t> partition{};
+    vertex_t number_of_vertices{};
+    edge_t number_of_edges{};
```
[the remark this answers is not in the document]
ChuckHastings: Removed

cpp/tests/experimental/renumbering_test.cpp
ChuckHastings: Done

cpp/tests/utilities/test_graphs.hpp
seunghwak: > Do we always symmetrize R-mat graphs? This is necessary for WCC or Louvain, but there are many other algorithms that work with asymmetric R-mat graphs.

@ChuckHastings I'm waiting for your answer for this to approve this PR.

ChuckHastings: That was the intention of my FIXME comment.  Right now the symmetrization flag is inside the test graph classes.  I think the goal would be to set the symmetrization flag to false for *all* of the test graph classes (except file which would read it from the MM file and return the intended value) and add a test graph wrapper that will symmetrize any test graph so that Louvain and WCC could use that wrapper around rmat and other generators.

seunghwak: Where is the FIXME comment?

```Rmat_Usecase``` has private member variable ```  bool undirected_{};```

So, this should be used instead of redefining ```symmetric``` here and ignoring the member variable. I assume users will be surprised if they set Rmat_Usecase.undirected_ = false and get symmetric graphs.

And it seems like ```construct_graph``` is still calling
```
return generate_graph_from_rmat_params<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>(
      handle,
      scale_,
      edge_factor_,
      a_,
      b_,
      c_,
      seed_,
      undirected_,
      scramble_vertex_ids_,
      test_weighted,
      renumber,
      partition_ids,
      comm_size);
````
So, ```construct_edgelist``` here is actually unused. Are you planning to replace ```generate_graph_from_rmat_params``` but haven't done the job, yet? Why not replace in this PR if it's the intention?

Comment (ChuckHastings) [reply cpp/tests/utilities/test_graphs.hpp:170]: The original implementation work here was truncated and a partially implemented capability was pushed many months ago.  This is unfinished tech debt.  I will look into making that change as part of this PR.

**Review by aschaffer — APPROVED** [review]

**Comment by ChuckHastings** [comment]
rerun tests

**Comment by rlratzel** [comment]
rerun tests

**Review by BradReesWork — APPROVED** [review]

**Review by aschaffer — APPROVED** [review]

File: cpp/tests/structure/renumbering_test.cpp [hunk 6]
```diff
@@ -0,0 +1,176 @@
...
+#include <iterator>
+#include <limits>
+#include <numeric>
+#include <random>
+#include <vector>
+
+// do the perf measurements
+// enabled by command line parameter s'--perf'
+//
+static int PERF = 0;
```
seunghwak, line 44: We're no longer using PERF, this gets replaced with cugraph::test::g_perf

cpp/tests/structure/renumbering_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+    if (renumbering_usecase.check_correctness) {
+      h_original_src_v.resize(src_v.size());
+      h_original_dst_v.resize(dst_v.size());
+
+      raft::update_host(h_original_src_v.data(), src_v.data(), src_v.size(), handle.get_stream());
+      raft::update_host(h_original_dst_v.data(), dst_v.data(), dst_v.size(), handle.get_stream());
+    }
+
+    if (PERF) {
+      handle.get_stream_view().synchronize();  // for consistent performance measurement
```
seunghwak, line 91: So, we're using CUDA_TRY(cudaDeviceSynchronize()) elsewhere.
This will be lower overhead and proper if every operation is executed on ```stream``` (this should be true... but may not be always true...) but we may better not mix the two in our tests. So, do you think we'd better use stream synchronize in our tests or we'd better continue using cudaDeviceSynchronize?

cpp/tests/structure/renumbering_test.cpp
```diff
@@ -0,0 +1,176 @@
...
+INSTANTIATE_TEST_SUITE_P(
+  rmat_small_tests,
+  Tests_Renumbering_Rmat,
+  ::testing::Combine(
+    // enable correctness checks
+    ::testing::Values(Renumbering_Usecase{}),
+    ::testing::Values(cugraph::test::Rmat_Usecase(10, 16, 0.57, 0.19, 0.19, 0, false, false))));
+
+INSTANTIATE_TEST_SUITE_P(
+  rmat_large_tests,
```
seunghwak, line 169: See https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/tests/prims/mg_count_if_v.cu#L225
We renamed rmat_large_tests to rmat_benchmark_test to use this in C++ benchmarking.

seunghwak (commented):
Looks good to me except for few minor complaints.

cpp/tests/structure/renumbering_test.cpp
ChuckHastings: That's what happens when you start something and then set it aside :-)

Fixed in next push.

cpp/tests/structure/renumbering_test.cpp
ChuckHastings: I switched to `CUDA_TRY(cudaDeviceSynchronize())` in the next push.

I'm not particularly concerned about overhead in tests.  We can consider that change as a general possibility later.

cpp/tests/structure/renumbering_test.cpp
ChuckHastings: Addressed in latest push.

cpp/tests/structure/renumbering_test.cpp
seunghwak: Yeah...  me, neither. Just wasn't happy about mixing cudaDeviceSynchronize() and stream_view.synchronize(). Either is fine for me as long as we're consistent.

seunghwak (approved):

ChuckHastings:
rerun tests

ChuckHastings:
@gpucibot merge

PR #1746, opened 2021-07-29
Changed cuco cmake function to return early if cuco has already been added as a target

rlratzel:
Changed cuco cmake function to return early if cuco has already been added as a target.  This matches the technique used by raft [here](https://github.com/rapidsai/raft/blob/a3af3895410c19f3e713caa608ea2024f6008350/cpp/cmake/thirdparty/get_cuco.cmake#L19).

Tested by doing a build and install of cuML, followed by a build of cuGraph and observing a CPM error about the alias target `cuco::cuco` already existing.  Made the change to return early if cuco is already a target and observed the cuGraph `libcugraph.so` build succeed.

### DIFF
--- a/cpp/cmake/thirdparty/get_cuco.cmake
+++ b/cpp/cmake/thirdparty/get_cuco.cmake
@@ -16,6 +16,10 @@
 
 function(find_and_configure_cuco VERSION)
 
+    if(TARGET cuco::cuco)
+      return()
+    endif()
+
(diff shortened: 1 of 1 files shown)


ChuckHastings (approved):

dillon-cullinan (approved):

BradReesWork:
@gpucibot merge

PR #1752, opened 2021-08-03
Move legacy implementations into legacy directories

ChuckHastings:
This is phase 1 of the plan to move the now `non-experimental` things in the experimental namespace into the `cugraph` namespace.  Unfortunately, some of the legacy implementations are still required, so we're going to move the files into legacy directories so they can be removed later.

NOTE: This is not moving everything that is legacy into legacy directories.  Only things that are in the way of new things or were in the same directories as things that were in the way of new things.  We will move other things into legacy directories when/if needed.


seunghwak (approved):
Looks good to me (can't we delete some of the unused legacy implementations? e.g. are we using katz_centrality?)

ChuckHastings:
> Looks good to me (can't we delete some of the unused legacy implementations? e.g. are we using katz_centrality?)

I think in one of my previous tries at this I tried deleting some of these things.  I believe katz_centrality was one of the things that I could delete.  I'll test that out on my next PR and delete what I can.

ChuckHastings:
@gpucibot merge

PR #1755, opened 2021-08-04
C++ benchmarking for MG PageRank

seunghwak:
- [x] Update google tests to take additional command line arguments (to report performance measurements and control R-mat graph size in command line)
- [x] Store the parsed command line arguments in globally accessible variables
- [x] Update MG PageRank test code to behave based on the globally accessible variables storing command line inputs.

Exemplar benchmark scripts
```
// 32 bit vertex & edge IDs
mpirun -n 2 --tag-output --output-filename log ./tests/MG_PAGERANK_TEST --gtest_filter=rmat_large_tests/Tests_MGPageRank_Rmat.CheckInt32Int32* --perf --rmat_scale=25 --rmat_edge_factor=16
// 32bit vertex ID & 64 bit edge ID
mpirun -n 2 --tag-output --output-filename log ./tests/MG_PAGERANK_TEST --gtest_filter=rmat_large_tests/Tests_MGPageRank_Rmat.CheckInt32Int64* --perf --rmat_scale=25 --rmat_edge_factor=16
// 64 bit vertex & edge IDs
mpirun -n 2 --tag-output --output-filename log ./tests/MG_PAGERANK_TEST --gtest_filter=rmat_large_tests/Tests_MGPageRank_Rmat.CheckInt64Int64* --perf --rmat_scale=25 --rmat_edge_factor=16
```


seunghwak:
@rlratzel @ChuckHastings Please review this based on our previous discussion about C++ benchmarking. Once we agree on the methodology, I will apply this to other tests as well (in a separate PR).

@kaatish You may also take a look as MG graph primitive tests should follow this.

cpp/tests/pagerank/mg_pagerank_test.cpp
```diff
@@ -315,8 +310,38 @@ TEST_P(Tests_MGPageRank_File, CheckInt32Int32FloatFloat)
 
 TEST_P(Tests_MGPageRank_Rmat, CheckInt32Int32FloatFloat)
 {
-  auto param = GetParam();
-  run_current_test<int32_t, int32_t, float, float>(std::get<0>(param), std::get<1>(param));
+  auto param            = GetParam();
+  auto pagerank_usecase = std::get<0>(param);
```
ChuckHastings, line 314: This approach will override *all* rmat tests, right?  In this case we have a small rmat use case that has the correctness test enabled and a large rmat use case that has the correctness disabled.  I believe this change would consequently run the test twice for each combination (32/32, 32/64, 64/64) once with and once without validation.  Not sure that's what we want.

seunghwak:
rerun tests

cpp/tests/pagerank/mg_pagerank_test.cpp
ChuckHastings: Perhaps we create a separate test to address this.  I'm bad at naming things, so imagine better names:

1. Create a test suite called Benchmark_MGPageRank_Rmat.  Disable validation in that test suite and provide a single rmat test (something small so that when we're doing unit tests in CI it won't take long)
2. Add this logic to *that* unit test, but not the basic unit tests
3. When you want to run with these overriding parameters you can run:  `mpirun -np 2 tests/MG_PAGERANK_TEST --gtest_filter=Benchmark_MGPageRank_Rmat*` and append whatever overriding parameters are desired

That would allow you to get a clean run without the extra runs.

seunghwak: > This approach will override _all_ rmat tests, right?

Yes, without ```--gtest_filter```, no, with the filter. This was intended to use in the benchmarking mode with something like ```--gtest_filter=rmat_large_tests/Tests_MGPageRank_Rmat.CheckInt32Int32*```.

seunghwak: See the "Exemplar benchmark scripts" part of this PR description.

seunghwak: With the filter to run only rmat_large_tests, I think the current code can achieve what we can do this Benchmark_MGPageRank_Rmat without actually adding additional code for that (and with minimum interruption in our C++ testing work process).

ChuckHastings: OK.  So I should read *all* of the description when reviewing the PR, just not jump to the code :-)

It would be nice to encapsulate this logic somehow rather than having to repeat it in every test.  Could we create a function that updates the rmat_usecase that can be reusable across all of the tests (SG or MG)?  Perhaps it could be a method in the rmat_usecase class?

ChuckHastings: Also, did we want to allow for specifying other input files (an analog for the file_usecase)?  That *might* require some refactoring of the tests since most of our tests don't have a single file test without validation.

seunghwak: > It would be nice to encapsulate this logic somehow rather than having to repeat it in every test.

Agreed, I will push this update in the next commit.

> Also, did we want to allow for specifying other input files (an analog for the file_usecase)?

Yeah... eventually, but let's think about this later (I guess we may need to add something like file_large_tests... but let's worry about this when we have a specific usecase).

ChuckHastings (approved):

cpp/tests/pagerank/mg_pagerank_test.cpp
```diff
@@ -316,7 +311,22 @@ TEST_P(Tests_MGPageRank_File, CheckInt32Int32FloatFloat)
 TEST_P(Tests_MGPageRank_Rmat, CheckInt32Int32FloatFloat)
 {
   auto param = GetParam();
-  run_current_test<int32_t, int32_t, float, float>(std::get<0>(param), std::get<1>(param));
+  run_current_test<int32_t, int32_t, float, float>(
+    std::get<0>(param), override_Rmat_Usecase_with_cmd_line_arguments(std::get<1>(param)));
```
rlratzel, line 315: A downside to this approach is that gtest doesn't know the override is happening, and as a result it may end up producing redundant combinations of input params (since each param is being overridden the same way), if there are >1 rmat usecase instances in the param list.  I'm guessing `--gtest_list_tests` may not print the correct list of tests when one of the rmat options are passed either (eg. `TestSuite/TestCase/2` will be associated with the default params and not the overridden ones), but that's relatively minor.

Unfortunately I don't have a better suggestion that's in the scope of this PR though.

kaatish (approved):

rlratzel (approved):
Approving but I do have some feedback that, if you can address would be nice, but I think it might be out-of-scope for this PR.

ChuckHastings:
@gpucibot merge

cpp/tests/pagerank/mg_pagerank_test.cpp
seunghwakang: Yeah... maybe it might be better to rename ```rmat_larget_tests``` to something like ```rmat_benchmarks``` (@ChuckHastings  also mentioned about creating Benchmark_MGPageRank_Rmat) and enforce rmat_benchmarks to have only one Rmat_Usecase (and in the benchmark mode, just select rmat_benchmarks using --gtest_filter).

seunghwak: Yeah... maybe it might be better to rename rmat_larget_tests to something like rmat_benchmarks (@ChuckHastings also mentioned about creating Benchmark_MGPageRank_Rmat) and enforce rmat_benchmarks to have only one Rmat_Usecase (and in the benchmark mode, just select rmat_benchmarks using --gtest_filter).

PR #1757, opened 2021-08-04
Move all new graph objects out of experimental namespace

ChuckHastings:
Moved new graph object headers out of experimental directory.

Updated all references to new graph objects to move them into the cugraph namespace.

NOTE: There will be at least one more PR to finish this job (moving a few more files around mostly)


seunghwak (commented):
Looks good to me, but any idea why python tests are failing?

ChuckHastings:
> Looks good to me, but any idea why python tests are failing?

It turns out that the SG python code is still calling the legacy `katz_centrality` implementation.  I reverted the commit that removed `katz_centrality`.  We'll have to wait for the python team to update the python code before we can remove the legacy implementation.

aschaffer (approved):

seunghwak (approved):

rlratzel (approved):
I *think* it looks good, but my eyes got tired after the 40th file or so.

I'm assuming the experimental C++ tests and the `EXPERIMENTAL_` prefix given to test binaries will be addressed in the next PR.

ChuckHastings:
@gpucibot merge

PR #1762, opened 2021-08-05
C++ benchmarking for additional algorithms

seunghwak:
Apply the updates in PR #1755 (for MG PageRank) to

MG: Katz Centrality, BFS, SSSP, WCC, and primitive tests
SG: PageRank, Katz Centrality, BFS, SSSP, and WCC tests

@ChuckHastings I will defer Louvain test updates to you :-) once Louvain tests get updated to take R-mat graphs.

### DIFF
--- a/cpp/tests/layout/force_atlas2_test.cu
+++ b/cpp/tests/layout/force_atlas2_test.cu
@@ -29,7 +29,6 @@
 #include <iostream>
 
 // iterations for perf tests
-// enabled by command line parameter '--perf-iters"
 static int PERF_MULTIPLIER = 5;
 
 typedef struct Force_Atlas2_Usecase_t {
--- a/cpp/tests/layout/force_atlas2_test.cu
+++ b/cpp/tests/layout/force_atlas2_test.cu
@@ -28,10 +28,6 @@
 #include <fstream>
 #include <iostream>
 
-// do the perf measurements
-// enabled by command line parameter s'--perf'
-static int PERF = 0;
-
 // iterations for perf tests
 // enabled by command line parameter '--perf-iters"
 static int PERF_MULTIPLIER = 5;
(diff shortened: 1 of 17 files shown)


rlratzel (approved):
This looks good, thank you. I think the new name expresses the intent better and the comments will prevent surprises to future maintainers. 👍

seunghwak:
rerun tests

cpp/tests/traversal/legacy/sssp_test.cu
```diff
@@ -108,11 +108,8 @@ void ref_sssp(const std::vector<MaxEType>& rowPtr,
   }
 }
 
-// do the perf measurements
-// enabled by command line parameter s'--perf'
-static int PERF = 0;
-
 // iterations for perf tests
+
 // enabled by command line parameter '--perf-iters"
```
ChuckHastings, line 113: This comment is no longer true, correct?   Probably should at least delete the comment if we want to leave in the `run 5 times` logic.

cpp/tests/layout/force_atlas2_test.cu
```diff
@@ -28,10 +28,6 @@
 #include <fstream>
 #include <iostream>
 
-// do the perf measurements
-// enabled by command line parameter s'--perf'
-static int PERF = 0;
-
 // iterations for perf tests
 // enabled by command line parameter '--perf-iters"
```
ChuckHastings, line 32: This comment is no longer true, correct? Probably should at least delete the comment if we want to leave in the run 5 times logic.

ChuckHastings (commented):
Minor comments

cpp/tests/traversal/legacy/sssp_test.cu
seunghwak: Correct, and done.

cpp/tests/layout/force_atlas2_test.cu
seunghwak: Correct, and done.

ChuckHastings (approved):

ChuckHastings:
@gpucibot merge

PR #1763, opened 2021-08-06
Move experimental source files and a few implementation headers

ChuckHastings:
Last phase of migrating things out of the experimental namespace.  This PR moves all of the source and test files out of the experimental directory and renames the tests.


cpp/tests/CMakeLists.txt
```diff
@@ -327,51 +327,51 @@ ConfigureTest(MST_TEST tree/mst_test.cu)
...
 
 ###################################################################################################
 # - Experimental SSSP tests -----------------------------------------------------------------------
-ConfigureTest(EXPERIMENTAL_SSSP_TEST experimental/sssp_test.cpp)
+ConfigureTest(SSSP_TEST traversal/sssp_test.cpp)
 
 ###################################################################################################
 # - Experimental PAGERANK tests -------------------------------------------------------------------
-ConfigureTest(EXPERIMENTAL_PAGERANK_TEST experimental/pagerank_test.cpp)
+ConfigureTest(PAGERANK_TEST pagerank/pagerank_test.cpp)
```
seunghwak, line 370: To be consistent with python, shouldn't this better go to link_analysis (after creating the directory)?

seunghwak (approved):
Looks good to me. I have left one minor comment about placing PageRank tests.

cpp/tests/CMakeLists.txt
ChuckHastings: Except I disagree with that python choice :-)

I'll make that change.

ChuckHastings: The `src` directory also uses `link_analysis`.  Now src, tests and python all match.

ChuckHastings:
@gpucibot merge

Issue #1477, opened 2021-03-22
[ENH] Improve memory footprint scaling with many GPUs for graphs with relative small E / V


seunghwak:
**Describe the solution you'd like**
Improvement 1:
Say
E: number of edges
V: number of vertices
P: number of GPUs

With the 2D partitioning, the edges of every vertex are distributed in sqrt(P) GPUs. If the average vertex degree is lower than sqrt(P), V_local > E_local in each local partition and the local partition becomes hypersparse. For hypersparse graphs and with the CSR (CSC) representation, the offset array will dominate the memory footprint. DCSR (DCSC) will be more memory efficient if E / V is relatively small and P is large.

Our renumbering scheme sorts vertices within a local partition based on vertex degrees and segments the local vertices based on the vertex degrees; this allows applying different representations to different segments; we can use CSR (CSC) for non-hypersparse segments and DCSR (DCSC) for hypersparse segments.

Improvement 2:
For PageRank style algorithms, we currently keep V / sqrt(P) old PageRank values in a consecutive array in each GPU. If V / sqrt(P) becomes larger than E / P (which can happen if E / V is relatively small and P is large), the V / sqrt(P) part will dominate the memory footprint and after this point, running N times larger graphs will require N^2 times more memory. But if V / sqrt(P) > E / P, we don't need the entire V / sqrt(P) old PageRank values; we need at most E / P old PageRank values. If we keep (source vertex, old PageRank value for the source vertex) pairs in each GPU, we can cut memory footprint form V / sqrt(P) to E / P (if E / P is significantly smaller than V / sqrt(P)). We can rely on either binary search or hash table for accessing old PageRank values in this case.

PR #1769, opened 2021-08-12
MNMG memory footprint improvement for low average vertex degree graphs (part 1)

seunghwak:
Partially addresses Issue https://github.com/rapidsai/cugraph/issues/1477

With the R-mat edge factor of 1 and a DGX machine with 8 GV100 GPUs (32 GB memory each), this PR improves the maximum R-mat scale we can run on the system from 30 to 31 (64 bit vertex & edge IDs).

This PR also includes few code refinements.

Key updates related to memory footprint improvements are in 
1) ```groupby_and_count_edgelist_by_local_partition_id``` (to allow optionally further groupby within a local partition by applying the compute_gpu_id_from_vertex_t to minor vertex IDs)
2) https://github.com/rapidsai/cugraph/pull/1769/files#diff-b8f50bff203b70295af567b06f68707dbd163c791e2630f8f21d4ad4b56f5262R754-R794 (renumber minor vertex IDs of each local partition in segments to limit the memory requirement).

One additional PR will be necessary to push memory bottleneck to row/column properties allocation (which currently requires O(V/sqrt(P)).
And there will be a separate PR to improve memory footprint of row/column properties when O(V/sqrt(P)) > O(E/P).

### DIFF
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -22,53 +22,43 @@ namespace cugraph {
 namespace detail {
 
 /**
- * @brief    Shuffle edgelist using the edge key function which returns the target GPU ID.
+ * @brief Shuffle edgelist using the edge key function which returns the target GPU ID.
  *
- * NOTE:  d_edgelist_rows, d_edgelist_cols and d_edgelist_weights
- *        are modified within this function (data is sorted)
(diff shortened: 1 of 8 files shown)


ChuckHastings (approved):
I like the solution to the note I added.  Move syntax makes that cleaner.

seunghwak:
rerun tests

BradReesWork:
@gpucibot merge

PR #1774, opened 2021-08-13
MNMG memory footprint improvement for low average vertex degree graphs (part 2)

seunghwak:
This PR has the following 2 updates.

1. Update the type of the create_graph_from_edgelist functions's optional vertex list input parameter from std::tuple of a pointer and the array size to rmm::device_uvector r-value (to be consistent with the edge list input parameters and to allow releasing the memory earlier).
2. Avoid storing counts (vertex degrees) for isolated vertices (which implicitly have 0 degrees) to save memory.


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1779, opened 2021-08-17
[BUG] force_atlas2 to support nx hypercube_graph

### LINKED ISSUE #1767: [BUG] AttributeError: 'Graph' object has no attribute 'edgelist'
Is there any particular reason why I can't do this? This is the error I get:

_AttributeError: 'Graph' object has no attribute 'edgelist'_

I figure since cugraph and networkx are supposed to be compatible, this would be a fairly standard operation.
```
import networkx as nx
import cudf
import cugraph

G = nx.hypercube_graph(4)

ITERATIONS=500
THETA=1.0
OPTIMIZE=True

pos_gdf = cugraph.layout.force_atlas2(G,
                                  max_iter=ITERATIONS,
                                  pos_list=None,
                                  outbound_attraction_distribution=True,
                                  lin_log_mode=False,
                                  edge_weight_influence=1.0,
                                  jitter_tolerance=1.0,
                                  barnes_hut_optimize=OPTIMIZE,
                                  barnes_hut_theta=THETA,
                                  scaling_ratio=2.0,
                                  strong_gravity_mode=False,
                                  gravity=1.0,
                                  verbose=False,
                                  callback=None)
```

jnke2016:
Allow `force_atlas2` to support hypercube_graph
The example provided in the issue linked to this PR works
closes #1767

### DIFF
--- a/python/cugraph/utilities/nx_factory.py
+++ b/python/cugraph/utilities/nx_factory.py
@@ -37,6 +37,7 @@ def convert_from_nx(nxG, weight=None):
 
     pdf = nx.to_pandas_edgelist(nxG)
     # Convert vertex columns to strings if they are not integers
+    # This allows support for any vertex input type
     if pdf["source"].dtype not in [np.int32, np.int64] or \
             pdf["target"].dtype not in [np.int32, np.int64]:
         pdf['source'] = pdf['source'].astype(str)
--- a/python/cugraph/utilities/nx_factory.py
+++ b/python/cugraph/utilities/nx_factory.py
@@ -36,6 +36,7 @@ def convert_from_nx(nxG, weight=None):
         raise ValueError("nxG does not appear to be a NetworkX graph type")
 
     pdf = nx.to_pandas_edgelist(nxG)
+    # Convert vertex columns to strings if they are not integers
     if pdf["source"].dtype not in [np.int32, np.int64] or \
             pdf["target"].dtype not in [np.int32, np.int64]:
         pdf['source'] = pdf['source'].astype(str)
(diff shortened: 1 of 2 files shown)


BradReesWork (approved):

BradReesWork:
rerun tests

python/cugraph/utilities/nx_factory.py
```diff
@@ -35,6 +36,11 @@ def convert_from_nx(nxG, weight=None):
         raise ValueError("nxG does not appear to be a NetworkX graph type")
 
     pdf = nx.to_pandas_edgelist(nxG)
+    if pdf["source"].dtype not in [np.int32, np.int64] or \
+            pdf["target"].dtype not in [np.int32, np.int64]:
+        pdf['source'] = pdf['source'].astype(str)
+        pdf['target'] = pdf['target'].astype(str)
```
rlratzel, line 42: Can you add a comment here explaining why this needs to be done? It's not clear to the reader why this is important, why it can't be removed or changed, etc.

python/cugraph/utilities/nx_factory.py
```diff
@@ -35,6 +36,11 @@ def convert_from_nx(nxG, weight=None):
         raise ValueError("nxG does not appear to be a NetworkX graph type")
 
     pdf = nx.to_pandas_edgelist(nxG)
+    if pdf["source"].dtype not in [np.int32, np.int64] or \
```
rlratzel, line 39: Would this operation be faster if done to the `gdf` instead?

rlratzel (changes requested):

python/cugraph/utilities/nx_factory.py
rlratzel: answer: maybe, but typical inputs to this function are small enough that the difference isn't significant in practice.

rlratzel (approved):

PR #1790, opened 2021-08-24
Update with rapids cmake new features

robertmaynard:
This combines some general CMake style cleanup and brings new rapids-cmake features to cugraph including:

- Usage of `rapids_cmake_write_version_file` to simplify cugraph version header writing
- Usage of `rapids_cmake_install_lib_dir` to make sure we install raft correctly on non-debain based distro's ( lib64 ), while also handling conda installation requirements ( always lib no matter the distro )
- Usage of `rapids_cpm` pre-configured pacakges
- Removal of early termination before `rapids_cpm_find` since a better solution now exists ( https://github.com/rapidsai/rapids-cmake/issues/49 )


cpp/cmake/thirdparty/get_cuco.cmake
```diff
@@ -16,12 +16,8 @@
 
 function(find_and_configure_cuco VERSION)
 
-    if(TARGET cuco::cuco)
-      return()
-    endif()
```
rlratzel, line 21: This was added by #1746 since we were seeing CPM errors about the alias target `cuco::cuco` already existing, since I believe raft also defines this.  Is this now safe to remove if we use raft?  Better question: if the target is already defined, will these changes result in an error?  If not, is there still an advantage in returning early if this was already defined?

cpp/cmake/thirdparty/get_gtest.cmake
```diff
@@ -14,30 +14,11 @@
 # limitations under the License.
 #=============================================================================
 
-function(find_and_configure_gtest VERSION)
+function(find_and_configure_gtest )
```
rlratzel, line 17: nit: space before the `)`

rlratzel (commented):
Thanks as always for keeping us up-to-date. A question and minor comment below:

cpp/cmake/thirdparty/get_cuco.cmake
robertmaynard: It is now safe to remove these guards when the target ( `cuco::cuco` ) is part of the `GLOBAL_TARGETS` list of `rapids_cpm_find`

This is due to an improvement to rapids-cmake ( https://github.com/rapidsai/rapids-cmake/pull/51 ) where it now rolls these checks into `rapids_cpm_find`.

rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #1792, opened 2021-08-25
[HOTFIX] Remove `-g` from cython compile commands

ajschmidt8:
Removes `-g` from the compile commands generated by distutils to compile Cython files.

This will make our container images, conda packages, and python wheels smaller.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -16,17 +16,26 @@ import sys
 import sysconfig
 import shutil
 
-from setuptools import setup, find_packages, Command
-from setuptools.extension import Extension
-from setuputils import use_raft_package, get_environment_option
+# Must import in this order:
+#   setuptools -> Cython.Distutils.build_ext -> setuptools.command.build_ext
+# Otherwise, setuptools.command.build_ext ends up inheriting from
+# Cython.Distutils.old_build_ext which we do not want
+import setuptools
 
 try:
-    from Cython.Distutils.build_ext import new_build_ext as build_ext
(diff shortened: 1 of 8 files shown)


python/setup.py
```diff
@@ -105,42 +140,50 @@ def run(self):
         os.system('find . -name "*.cpython*.so" -type f -delete')
 
 
+class build_ext_no_debug(_build_ext):
```
rlratzel, line 143: minor: I've always seen "release" used as the term for "no debug" (ie. it's either a release build or a debug build).

rlratzel (changes requested):
Looks good, thanks for doing this. I had one minor naming suggestion, and this will also need to be applied to the latest cugraph changes which add `pylibcugraph` (it has its own setup.py) and change the directory structure. I can help with that if you want.

python/setup.py
rlratzel: I just noticed this is a hotfix for 21.08, so I don't mind if this change isn't made (I'm fine either way).  I'll approve.

rlratzel (approved):

python/setup.py
BradReesWork: @ajschmidt8 is 21.08 and main the correct branches?

rlratzel:
rerun tests

python/setup.py
ajschmidt8: yes, these are the correct branches. this needs to go into `branch-21.08`.

PR #1793, opened 2021-08-27
updated to new doc theme 

BradReesWork:


ajschmidt8 (approved):

BradReesWork:
rerun tests

BradReesWork:
rerun tests

BradReesWork:
rerun tests

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -554,6 +612,9 @@ def __init__(self, m_graph=None):
 
 
 class MultiGraph(Graph):
+    """
+    A Multigraph, can be more than one edges between vertex pairs
```
rlratzel, line 616:
```suggestion
    A Multigraph; a Graph containing more than one edge between vertex pairs.
```

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -274,22 +295,43 @@ def from_pandas_edgelist(
         >>> G.from_pandas_edgelist(df, source='0', destination='1',
                                  edge_attr='2', renumber=False)
         """
+        if not isinstance(pdf, pandas.core.frame.DataFrame):
+            raise Exception("pdf input is not a Pandas DataFrame")
```
rlratzel, line 299: This should be raising `TypeError`

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -17,6 +17,7 @@
                                    npartiteGraphImpl)
 import cudf
 import warnings
+import pandas
```
rlratzel, line 20: This is a new dependency, so our conda recipe has to be updated.  Maybe we should look into a different way to do the type checking that doesn't require depending on Pandas?

This would be a good candidate for an optional import, since users passing Pandas types should already have Pandas installed, and those without Pandas probably wouldn't pass those types.  [Here's](https://github.com/rapidsai/cugraph/blob/df2d81dbb26904775386a6a25f20b776cd1897cb/python/cugraph/cugraph/utilities/utils.py#L311) the utility, and [here](https://github.com/rapidsai/cugraph/blob/df2d81dbb26904775386a6a25f20b776cd1897cb/python/cugraph/cugraph/utilities/nx_factory.py#L19) and [here](https://github.com/rapidsai/cugraph/blob/df2d81dbb26904775386a6a25f20b776cd1897cb/python/cugraph/cugraph/utilities/nx_factory.py#L28) are examples.

docs/cugraph/source/api_docs/structure.rst
```diff
@@ -1,31 +1,40 @@
...
    :toctree: api/
 
    Graph
+   MultiGraph
+   BiPartiteGraph
 
 
-Methods
--------
+Adding Data
```
rlratzel, line 16: This is kind of a nit pick so feel free to ignore: this list also includes `clear()`, so maybe the header is "Adding and Removing Data"? This could also be a good placeholder for future filtering and dynamic graph APIs.

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -274,22 +295,43 @@ def from_pandas_edgelist(
         >>> G.from_pandas_edgelist(df, source='0', destination='1',
                                  edge_attr='2', renumber=False)
         """
+        if not isinstance(pdf, pandas.core.frame.DataFrame):
```
rlratzel, line 298: Are there new tests for this and all other new checks in another PR?

rlratzel (changes requested):

rlratzel (approved):

BradReesWork:
@gpucibot merge

BradReesWork:
@gpucibot merge

PR #1795, opened 2021-08-30
UPDATED PYTHON to PYTHON_VER

jolorunyomi:


cpp/src/components/connectivity.cu
```diff
@@ -38,7 +38,7 @@ namespace detail {
  * This implementation comes from [1] and solves component labeling problem in
  * parallel on CSR-indexes based upon the vertex degree and adjacency graph.
  *
- * [1] Hawick, K.A et al, 2010. "Parallel graph component labelling with GPUs and CUDA"
+ * [1] Hawick, K.A et al, 2010. "Parallel graph component labelling with GPUs and CUDA_VER"
```
ChuckHastings, line 41: This is not a valid change.  `CUDA` appears in the title of the paper being referenced.

cpp/src/components/weak_cc.cuh
```diff
@@ -197,7 +197,7 @@ void weak_cc_label_batched(vertex_t* labels,
  * This implementation comes from [1] and solves component labeling problem in
  * parallel on CSR-indexes based upon the vertex degree and adjacency graph.
  *
- * [1] Hawick, K.A et al, 2010. "Parallel graph component labelling with GPUs and CUDA"
+ * [1] Hawick, K.A et al, 2010. "Parallel graph component labelling with GPUs and CUDA_VER"
```
ChuckHastings, line 200: This is not a valid change.  `CUDA` appears in the title of the paper being referenced.

cpp/src/components/weak_cc.cuh
```diff
@@ -253,7 +253,7 @@ void weak_cc_batched(vertex_t* labels,
...
```
ChuckHastings, line 256: This is not a valid change.  `CUDA` appears in the title of the paper being referenced.

cpp/src/components/weak_cc.cuh
```diff
@@ -301,7 +301,7 @@ void weak_cc(vertex_t* labels,
...
```
ChuckHastings, line 304: This is not a valid change.  `CUDA` appears in the title of the paper being referenced.

python/cugraph/cugraph/community/__init__.py
```diff
@@ -55,10 +55,10 @@ def __call__(self, *args, **kwargs):
 else:
     __kuvs = ".".join([str(n) for n in __ktruss_unsupported_cuda_version])
     k_truss = __UnsupportedModule(
-        NotImplementedError("k_truss is not currently supported in CUDA"
+        NotImplementedError("k_truss is not currently supported in CUDA_VER"
```
ChuckHastings, line 58: @rlratzel - Leaving this for you.  Do we want to change the text of an error message to say CUDA_VER rather than CUDA?

ChuckHastings (changes requested):

rlratzel (changes requested):
Thanks for updating this.  In addition to Chuck's feedback, you'll also need to get the latest version and update the new `pylibcugraph` python package (recipe, script updates, etc.) too before this can be merged.

python/cugraph/cugraph/community/__init__.py
rlratzel: No, `CUDA` is used here as the acronym in a user-facing error message, and is not a shell variable to be replaced.

PR #1797, opened 2021-08-31
Fix MG Louvain test compile errors

seunghwak:
Fix MG Louvain test compile errors after recent PR merges involving thrust execution policies.

### DIFF
--- a/cpp/tests/community/mg_louvain_helper.cu
+++ b/cpp/tests/community/mg_louvain_helper.cu
@@ -85,8 +85,7 @@ compressed_sparse_to_edgelist(edge_t const* compressed_sparse_offsets,
   // FIXME: this is highly inefficient for very high-degree vertices, for better performance, we can
   // fill high-degree vertices using one CUDA block per vertex, mid-degree vertices using one CUDA
   // warp per vertex, and low-degree vertices using one CUDA thread per block
-  auto execution_policy = handle.get_thrust_policy();
-  thrust::for_each(execution_policy,
+  thrust::for_each(rmm::exec_policy(stream),
                    thrust::make_counting_iterator(major_first),
                    thrust::make_counting_iterator(major_last),
                    [compressed_sparse_offsets,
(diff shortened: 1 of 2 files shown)


cpp/tests/community/mg_louvain_helper.cu
```diff
@@ -85,8 +85,7 @@ compressed_sparse_to_edgelist(edge_t const* compressed_sparse_offsets,
   // FIXME: this is highly inefficient for very high-degree vertices, for better performance, we can
   // fill high-degree vertices using one CUDA block per vertex, mid-degree vertices using one CUDA
   // warp per vertex, and low-degree vertices using one CUDA thread per block
-  auto execution_policy = handle.get_thrust_policy();
-  thrust::for_each(execution_policy,
+  thrust::for_each(rmm::exec_policy(stream),
```
ChuckHastings, line 88: I imagine this was missed because CI doesn't compile MG analytics.

Would it be better to change the function signature to pass in `handle`?  That would make it more consistent with the rest of the code base.

ChuckHastings: @kaatish addressed this in #1798 by changing the function signature.   We should decide which is better and be consistent.

seunghwak: > Would it be better to change the function signature to pass in `handle`? That would make it more consistent with the rest of the code base.

Ideally, this code should be deleted as this file is pretty much copy-and-paste (plus little modification) of https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/src/structure/coarsen_graph.cu

The reason this function is designed to take stream instead of handle is to facilitate multi-stream execution. Different matrix partitions can be concurrently executed in multiple CUDA streams and such functions take stream instead of handle (another approach might be passing handle + stream index if we see multiple cases using other handle features).

ChuckHastings (approved):

ChuckHastings:
@gpucibot merge

PR #1798, opened 2021-08-31
Transform Reduce E test

kaatish:
This PR adds multi-gpu testing for transform_reduce_e primitive.

A bug was fixed in transform_reduce_e where shared memory was being allocated inside a device function for the purpose of using it with cub::BlockReduce.

Utilities in dataframe_buffer.cuh were cleaned up. dataframe_buffer related functions with the exception of allocation functions no longer require an explicit template parameter.


cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,445 @@
...
+      auto output_property = std::make_tuple(rmm::device_uvector<Args>(graph_view.get_number_of_local_adj_matrix_partition_rows(), handle.get_stream())...);
+      copy_property(property, output_property,
+                    [&handle, &graph_view] (const auto& in, auto& out) {
+                      copy_to_adj_matrix_row(handle, graph_view, in.begin(), out.begin());
+                    }
+                   );
+      return output_property;
+    }
+
+};
```
seunghwak, line 175: FYI:

Actually, I am adding

```
template <typename GraphViewType, typename T>
class row(or col)_properties_t;
```
And this will be the last input parameter of copy_to_adj_matrix_row(...);

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,445 @@
...
+    return data;
+  }
+  template <typename label_t>
+  static std::tuple<rmm::device_uvector<Args>...> vertex_property(thrust::counting_iterator<label_t> begin,
+                                                           thrust::counting_iterator<label_t> end,
+                                                           int hash_bin_count,
+                                                           raft::handle_t const& handle)
+  {
+    auto length = thrust::distance(begin, end);
+    auto data   = std::make_tuple(rmm::device_uvector<Args>(length, handle.get_stream())...);
```
seunghwak, line 111: This pretty much overlaps with https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/include/cugraph/utilities/dataframe_buffer.cuh#L112 and https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/include/cugraph/utilities/dataframe_buffer.cuh#L118

If you using allocate_dataframe_buffer, I think you can merge the two different versions of vertex_property to one.

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,445 @@
...
+};
+template <typename vertex_t, template <typename...> typename Tuple, typename... Args>
+struct property_transform<vertex_t, Tuple<Args...>> : public property_transform<vertex_t, Args...> {
+};
+
+template <typename Tuple, std::size_t... I>
+auto make_iterator_tuple(Tuple& data, std::index_sequence<I...>)
+{
+  return thrust::make_tuple((std::get<I>(data).begin())...);
+}
```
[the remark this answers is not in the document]
kaatish: This isn't about converting an std tuple to thrust tuple. It is about getting a zipped iterator from an std::tuple of device_vectors. I will move it to the tuple_utils file. As I have mentioned in an earlier PR, an std::tuple of device_vectors is better suited sometimes because they do not mangle the template parameter pack.

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,445 @@
...
+  int mod{};
+  property_transform(int mod_count) : mod(mod_count) {}
+  constexpr __device__ auto operator()(const vertex_t& val)
+  {
+    cuco::detail::MurmurHash3_32<vertex_t> hash_func{};
+    auto value = hash_func(val) % mod;
+    return thrust::make_tuple(static_cast<Args>(value)...);
+  }
+};
+template <typename vertex_t, template <typename...> typename Tuple, typename... Args>
```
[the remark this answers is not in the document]
kaatish: > is this a mechanism to specify that Tuple will require an additional parameter pack to be fully defined?

Yes. We are expressing that the Tuple template will take a parameter pack. I believe the term for this pattern 
`template <template <typename> typename T>` is just called template template parameter. There are some links on [cppreference](https://en.cppreference.com/w/cpp/language/template_parameters) which mention it in passing but I think [this book](https://www.amazon.com/C-Templates-Complete-Guide-2nd/dp/0321714121) will be a better source for it.

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,445 @@
...
+
+#include <random>
+
+template <typename vertex_t, typename... Args>
+struct property_transform : public thrust::unary_function<vertex_t, thrust::tuple<Args...>> {
+  int mod{};
+  property_transform(int mod_count) : mod(mod_count) {}
+  constexpr __device__ auto operator()(const vertex_t& val)
+  {
+    cuco::detail::MurmurHash3_32<vertex_t> hash_func{};
```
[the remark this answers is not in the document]
kaatish: Yes, this can be run on the host.

cpp/tests/prims/mg_transform_reduce_e.cu
seunghwak: It will be completely covered by #1801 (but its actually behavior will not change, the actual memory footprint scaling improvements will happen in another PR, but this will be transparent to primitives users).

ChuckHastings (approved):

cpp/include/cugraph/prims/transform_reduce_e.cuh
```diff
@@ -117,9 +121,7 @@ __global__ void for_all_major_for_all_nbr_hypersparse(
     idx += gridDim.x * blockDim.x;
   }
 
-  e_op_result_sum =
-    block_reduce_edge_op_result<e_op_result_t, transform_reduce_e_for_all_block_size>().compute(
-      e_op_result_sum);
+  e_op_result_sum = BlockReduce(temp_storage).Reduce(e_op_result_sum, edge_property_add);
```
seunghwak, line 124: As we discussed, if block_reduce_edge_op_result can serve what we need, we'd better delete this and directly use cub everywhere for consistency.

cpp/include/cugraph/utilities/dataframe_buffer.cuh
```diff
@@ -122,100 +81,73 @@ auto allocate_dataframe_buffer(size_t buffer_size, rmm::cuda_stream_view stream_
...
-size_t size_dataframe_buffer(BufferType& buffer)
+template <typename Type>
+size_t size_dataframe_buffer(Type& buffer)
 {
-  return std::get<0>(buffer).size();
+  if constexpr (is_std_tuple_of_arithmetic_vectors<Type>::value) {
+    return std::get<0>(buffer).size();
+  } else if constexpr (is_arithmetic_vector<Type, rmm::device_uvector>::value) {
+    return buffer.size();
+  }
```
seunghwak, line 115: Didn't you see a warning about missing return statement?

See
https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/include/cugraph/prims/update_frontier_v_push_if_out_nbr.cuh#L80

Here, I am forced to use SFINAE instead of the simpler ```if constepxr``` approach to silence the warning due to an nvcc bug.

cpp/include/cugraph/utilities/thrust_tuple_utils.cuh
```diff
@@ -132,19 +133,44 @@ template <typename... Ts>
...
-template <typename TupleType, typename Enable = void>
+template <typename TupleType>
 struct is_thrust_tuple_of_arithmetic : std::false_type {
 };
 
-template <typename TupleType>
-struct is_thrust_tuple_of_arithmetic<TupleType,
-                                     std::enable_if_t<is_thrust_tuple<TupleType>::value>> {
-  static constexpr bool value =
-    detail::is_thrust_tuple_of_arithemetic_impl<TupleType,
```
seunghwak, line 143: So, if we're not using detail::is_thrust_tuple_of_arithmetic_impl anymore, we should delete unused functions.

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,448 @@
...
+#include <raft/handle.hpp>
+#include <rmm/device_scalar.hpp>
+#include <rmm/device_uvector.hpp>
+
+#include <gtest/gtest.h>
+
+#include <random>
+
+template <typename... Args>
+struct PropertyType {
```
seunghwak, line 46: Any reason for naming this struct as "P"roperty"T"ype instead of property_type (as you're using property_transform and generate_impl below)?

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,448 @@
...
+  using Type = std::conditional_t<(sizeof...(Args) > 1),
```
seunghwak, line 47: I think "t"ype is more idiomatic.

cpp/tests/prims/mg_transform_reduce_e.cu
```diff
@@ -0,0 +1,448 @@
...
+
+template <typename vertex_t, template <typename...> typename Tuple, typename... Args>
+struct property_transform<vertex_t, Tuple<Args...>> : public property_transform<vertex_t, Args...> {
+};
+
+template <typename... Args>
+struct generate_impl {
+ private:
+  using Type               = typename PropertyType<Args...>::Type;
+  using PropertyBufferType = std::conditional_t<
```
seunghwak, line 86: I think "t"ype and property_buffer_type are more idiomatic.

seunghwak (commented):
I really like the updates you made to dataframe_buffer.cuh!!!

It might be better to update the test code to be readable by users without strong expertise in template metaprogramming (don't go too far if it requires significantly sacrificing code quality... and I understand it might not be possible... but just in case if possible). I am just a bit worried about scaring potential graph primitives users who are not very familiar with TMP.

This PR looks really good otherwise.

cpp/include/cugraph/utilities/dataframe_buffer.cuh
kaatish: It would have been a problem but I am returning size_t{} in the next line to avoid that warning.

seunghwak: Ah, yeah... here the return types are same so easy to fix!!!

cpp/include/cugraph/prims/transform_reduce_e.cuh
kaatish: `block_reduce_edge_op_result` is still being used by `copy_v_transform_reduce_nbr` and therefore cannot be removed yet. It should be removed though because it is buggy. Perhaps in another PR?

seunghwak: Yeah... you can remove it when you write test cases for copy_v_transform_reduce_nbr.

seunghwak (approved):
This looks great and thanks for the updates!!!

BradReesWork:
@gpucibot merge

PR #1799, opened 2021-08-31
[WIP] Bindings for Biased RW 

aschaffer:
This PR tracks work on C-API / Python Bindings for Biased RW. The design is meant to be general enough to support future sampling techniques.


cpp/CMakeLists.txt
```diff
@@ -205,6 +205,7 @@ add_library(cugraph SHARED
     src/visitors/graph_envelope.cpp
     src/visitors/visitors_factory.cpp
     src/visitors/bfs_visitor.cpp
+    src/visitors/rw_visitor.cpp
```
seunghwak, line 208: So, do we plan to place bindings related files under the cugrpah/cpp directory? Or we'd better separate the core libcugraph from the bindings related files? (i.e. placing bindings related files under something like cugraph/bindings)

aschaffer: The visitor implementation is not restricted to bindings only. It's meant to be (and it is) used for implementation purposes, too. These _are_ cpp files. There will be a separate branch for the C-only API.

seunghwak: Then, what about creating a separate directory under the cpp (similar to tests)? Is this directly relevant to libcugraph users? (someone willing to directly use our C++ API). If not, I am not sure it is a good idea to include this in our core libcugraph_(sg|mg).so.

aschaffer: Yes, it is. As you probably know Visitor is useful to enrich the functionality of an object (visitee) without having to add new methods to it, which may inflate the code and cause merge conflicts on high traffic headers, like `graph*.hpp` .

The visitor abstraction is being used in RW, for example, for exactly such reasons.

Also, this structuring of this code has been in place for a while. And it has been extensively discussed before-hand (at least 2 presentations and demos have been done). Constantly shuffling code around _after the fact_ is counterproductive, confusing, and wasteful. If new code restructuring is in the works, then this is _not_ the PR for it. This PR has different goals. We've already agreed on small, targeted PRs as opposed to all-inclusive major code restructuring ones.

There's also the question of what exactly is the criterion for code cut-off to be accepted in `cugraph`. If this is to be discussed, then, again, maybe this is something to discuss _before-fact_, not after. I am starting to get the feeling that it's increasingly hard to get code accepted in `cugraph` and a lot of effort is potentially wasted with code shuffling and reshuffling on what appears to be quite ad-hoc and last-moment criteria.

aschaffer:
The `cugraph_api.h` and `cugraph_api.cpp` locations are (currently assumed as being) temporary. They most likely will be moved by the end of this PR. They're here just for me to be able to compile/build/test the C-API flow.

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -0,0 +1,97 @@
...
+ *      interpret return in a different (typed) way;
+ *
+ *      Example:
+ *      c_device_buffer_t* extract_rw_ret_vertex_path(c_type vertex_t_id, c_rw_ret_t* rw_result);
+ *      c_device_buffer_t* extract_rw_ret_weight_path(c_type vertex_t_id, c_rw_ret_t* rw_result);
+ */
+
+/* C algorithm specific wrapper declarations: : should go into separate corresponding headers */
+
+cugraph_error_t c_random_walks(const c_raft_handle_t* ptr_handle,
```
rlratzel, line 86: I think this may have been mentioned before, but now would be a good time to change the naming convention if we want to do that. Within RAPIDS, cuML follows the convention of using the project name as a prefix (as seen [here](https://github.com/rapidsai/cuml/blob/branch-21.10/cpp/include/cuml/cuml_api.h)), as do other CUDA libraries such as [this](https://docs.nvidia.com/cuda/cusolver/index.html#cusolverDNXgesv).  Related to that is the dreaded topic of case style; cuML and the other CUDA libraries use camelCase and not snake_case.

aschaffer: I'm not so sure about camelCase. All our classes (types) use snake_case in the cugraph code (and `_t` suffix).

rlratzel: I don't have a strong opinion about snake vs. camel, so leaving it snake_case is also fine by me if that's what we want. @ChuckHastings @BradReesWork any thoughts?

ChuckHastings: I'd prefer to keep things as snake_case unless there's a compelling reason to change them.  Personally not a fan of the `_t` as widely as we are using it, but it's OK.

I do think that rather than `c_` as the prefix, making it `cugraph_` would probably be more informative.  The fact that it's a C API is immaterial to the calling user.  The fact that it's part of a cugraph package is far more informative.  It would also reduce the likelihood of name collision, since we would likely be the only source of C functions that start with `cugraph_`.

ajschmidt8 (approved):

rlratzel:
@robertmaynard - can you also review my cmake changes, since they probably need improvement?

cpp/CMakeLists.txt
```diff
@@ -334,17 +334,57 @@ target_link_libraries(cugraph PRIVATE
...
+        "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+        "$<INSTALL_INTERFACE:include>"
+)
+
+################################################################################
+# - C-API link libraries -------------------------------------------------------
+target_link_libraries(cugraph_c
+        PUBLIC
+                cugraph::cugraph
+                FAISS::FAISS
```
robertmaynard, line 375:
```suggestion
                FAISS::FAISS
    PRIVATE
        CUDA::cublas
        CUDA::curand
        CUDA::cusolver
        CUDA::cusparse             
```

To resolve issues not find cublas / cusolve / etc issues

robertmaynard: Lets mark FAISS private as well

rlratzel: Thanks @robertmaynard . We discussed offline and decided to also try making FAISS private. I'll push that change and make FAISS public again if it results in problems.

cpp/tests/c_api/create_sg_graph_test.c
```diff
@@ -0,0 +1,128 @@
...
+  bool_t is_symmetric                  = 0;
+  bool_t is_multigraph                 = 0;
+
+  /*
+   * FIXME: populate GPU memory with a small (4 verts, 3 edges) graph COO.
+   * FIXME: return success until this test is finished.
+   */
+  return 0;
+
+  G = cugraph_make_sg_graph(&handle,
```
rlratzel, line 46: I just realized I named the test and test file with the word "create" instead of "make".  Similar to Chuck's comment re: func and file names matching, I'd make these match too (either change the API to "create" or the file and test to "make").

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,134 @@
...
+      ++p_path_v; /* iterate p_path_v*/
+    }
+
+    flag_passed = is_one_of(crt_path_accumulator, c_ps_array, N_PATHS);
+    if (flag_passed) ++count_passed;
+  }
+
+  return (count_passed == num_paths);
+}
+int test_random_walks_1()
```
rlratzel, line 73: You may not be finished updating this test, but just so you don't forget:  this was just meant to be a placeholder name - I'd change it to something more descriptive so the user can identify what failed from the test output.

rlratzel:
rerun tests

cpp/src/c_api/cugraph_api.cpp
```diff
@@ -0,0 +1,360 @@
...
+      return CUGRAPH_UNKNOWN_ERROR;
+    }
+}
+}  // namespace helpers
+
+extern "C" {
+int data_type_sz[] = {4, 8, 4, 8};
+}
+
+bool_t runtime_assert(bool_t statement_truth_value, const char* error_msg)
```
rlratzel, line 84: I think the name "assert" here is misleading.  When reading code, I expect execution to stop when a false value is passed to this function.

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,434 @@
...
+  cugraph_free_device_buffer(&dbuf_wgt);
+
+  cugraph_free_device_buffer(&dbuf_dst);
+
+  cugraph_free_device_buffer(&dbuf_src);
+
+  cugraph_free_handle(p_handle);
+
+  /* because bools and err return codes from programs have opposite meaning...*/
+  if (flag_failed)
```
rlratzel, line 419: I think this is still inverted: tests need to return 0 on success, but this will return 0 if flag_failed is true, meaning it's a failure? Maybe I'm confused...

edit: oh, this flag is only used for the check above which must "fail". I would possibly rename the variable, or add something to the comment, or change the name of the test, or all of the above.

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,434 @@
...
+
+  cugraph_free_device_buffer(&dbuf_src);
+
+  cugraph_free_handle(p_handle);
+
+  return 0;
+}
+
+/* negative test RW call flow*/
+int test_random_walks_3()
```
rlratzel, line 315: this tests and others still using the placeholder name need to have a more descriptive name.

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,434 @@
...
+   using 1-based indexing for vertices,
+   to avoid confusion between, for example,
+   `012` and `12`, which result in same number*/
+#define NUM_MAX_PATHS 30
+static int32_t c_ps_array[NUM_MAX_PATHS] = {
+  1,  2,  3,  4,   5,    6,    12,    124,   125, 1246, 1256, 24,   25,  246, 256,
+  31, 32, 34, 312, 3124, 3125, 31246, 31256, 324, 325,  3246, 3256, 346, 46,  56};
+
+/* linear search of `value` inside `p_cmprsd_path[max_num_paths]`*/
+bool_t is_one_of(int32_t value, int32_t* p_cmprsd_path, int max_num_paths)
```
rlratzel, line 152: I think some of these abbreviations are harder to type and read than the non-abbreviated words.  Can this just say "compressed" instead of "cmprsd" (assuming that's what it's abbreviating)?

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,434 @@
...
+
+  cugraph_free_device_buffer(&dbuf_wgt);
+
+  cugraph_free_device_buffer(&dbuf_dst);
+
+  cugraph_free_device_buffer(&dbuf_src);
+
+  cugraph_free_handle(p_handle);
+
+  return 0;
```
rlratzel, line 139: Because the asserts don't fail the test and their return values aren't being captured and checked, this test will always pass, and therefore CI will not catch anything.

cpp/tests/c_api/random_walks_test.c
```diff
@@ -0,0 +1,434 @@
...
+  cugraph_rw_ret_t rw_ret;
+  cugraph_device_buffer_t dbuf_start;
+  size_t num_paths                          = 2;
+  size_t max_depth                          = 4;
+  bool_t flag_use_padding                   = FALSE;
+  cugraph_unique_ptr_t* p_sampling_strategy = NULL;
+  vertex_t h_start[]                        = {0, 2};
+
+  p_handle = cugraph_create_handle();
+  runtime_assert(p_handle != NULL, "raft handle creation failed.");
```
rlratzel, line 69: since this isn't a real assert (the implementation I saw above seems to just print to stderr), the final return value of the test needs to reflect if this fails, but it's just being ignored.  Same for all other `runtime_assert` calls that aren't setting a flag of some sort.

cpp/tests/c_api/create_sg_graph_test.c
```diff
@@ -0,0 +1,161 @@
...
+  size_t num_edges                     = 3;
+  bool_t do_expensive_check            = 1;
+  bool_t store_transposed              = 0;
+  bool_t is_symmetric                  = 0;
+  bool_t is_multigraph                 = 0;
+
+  /*
+   * FIXME: return success until this test is finished.
+   */
+  return 0;
```
rlratzel, line 131: Can this test at least be enabled and passing prior to merge?  It seems like this should pass.

rlratzel (changes requested):
Still reviewing but a few things jumped out (I've only looked at the tests) so I'll post what I have so far.
My biggest concern is that it looks like some tests are hardcoded to return success, which should be easy to fix so that CI will actually catch problems.

cpp/tests/c_api/random_walks_test.c
aschaffer: We can modify this in the next release.

cpp/tests/c_api/create_sg_graph_test.c
aschaffer: This has been addressed.

cpp/tests/c_api/random_walks_test.c
aschaffer: Yes, this is just debug info. We can modify this in the next release.

cpp/src/c_api/cugraph_api.cpp
aschaffer: I don't necessarily agree, but you're free to change that.

cpp/tests/c_api/random_walks_test.c
aschaffer: It's correct. This is a negative test.

cpp/tests/c_api/random_walks_test.c
aschaffer: You provided these names. You're free to change them in the next release.

cpp/tests/c_api/random_walks_test.c
aschaffer: Again, question of style/taste. You're free to change them in the next release.

BradReesWork:
rerun tests

cpp/include/cugraph/visitors/graph_make_visitor.hpp
```diff
@@ -0,0 +1,82 @@
...
+namespace visitors {
+// primary empty template:
+//
+template <typename vertex_t,
+          typename edge_t,
+          typename weight_t,
+          bool st,
+          bool mg,
+          typename Enable = void>
+struct graph_maker_visitor;
```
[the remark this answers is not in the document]
aschaffer: This was done on purpose, because there's more functionality than just the visitor in there. The file name suggests content pertaining to the `visitor` design (verb = `make appropriate visitor` + related functionality) while the `visitor` name says: "I'm a visitor that makes graphs"...but in that same file there's more than that. The 2 names are too close, I agree, but I wanted to make that clear distinction. Can I please address that after merge (since every little change takes almost a whole day for CI to process)?

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -0,0 +1,170 @@
...
+cugraph_unique_ptr_t* cugraph_create_sampling_strategy(int sampling_type_id, double p, double q);
+
+/* Sampling data deallocator*/
+void cugraph_free_sampling_strategy(cugraph_unique_ptr_t* p_sampling);
+
+/* deallocate result returned by RW wrapper*/
+void cugraph_free_rw_result(cugraph_rw_ret_t* p_rw_ret);
+
+/* RW result vertex extractor*/
+void extract_vertex_rw_result(cugraph_rw_ret_t* p_rw_ret, cugraph_device_buffer_t* p_d_buf_v);
```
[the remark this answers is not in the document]
aschaffer: I thought about that. It's just becoming too long of a name. I'd rather keep the name as informative and short as possible. Adding `cugraph_` prefix would have probably require having to truncate / abbreviate the rest of the name. I will address that after merge, if I may (again to not be chocked by CI on what seems to be mergeable in last moment).

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -0,0 +1,170 @@
...
+
+/* RW result weights extractor*/
+void extract_weight_rw_result(cugraph_rw_ret_t* p_rw_ret, cugraph_device_buffer_t* p_d_buf_w);
```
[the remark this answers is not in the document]
aschaffer: Same FIXME I'll handle later.

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -0,0 +1,170 @@
...
+
+/* RW result size extractor*/
+void extract_size_rw_result(cugraph_rw_ret_t* p_rw_ret, cugraph_device_buffer_t* p_d_buf_sz);
```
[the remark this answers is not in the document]
aschaffer: Same FIXME I'll handle later.

cpp/src/c_api/cugraph_api.cpp
```diff
@@ -0,0 +1,319 @@
...
+#include <cugraph/visitors/rw_visitor.hpp>
+
+#include <cugraph/visitors/erased_api.hpp>
+
+#include <raft/handle.hpp>
+
+#include <rmm/device_buffer.hpp>
+#include <rmm/device_uvector.hpp>
+
+namespace helpers {
```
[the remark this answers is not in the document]
aschaffer: No. I thought about that. Anonymous spaces aren't really anonymous (compiler generates names for them, too). But, more importantly, I want to collect all these "helpers" under a common...abstraction. So, different C++ sources pertaining to the C-API would enrich this same `namespace` rather than shovel them in the same `hpp` header, which then could not / should not define unnamed namespaces, anyway.

cpp/src/visitors/graph_make_visitor.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <cugraph/algorithms.hpp>
+#include <cugraph/visitors/graph_make_visitor.hpp>
+
+namespace cugraph {
+namespace visitors {
+//
+// wrapper code:
+//
+template <typename vertex_t, typename edge_t, typename weight_t, bool st, bool mg>
+void graph_maker_visitor<vertex_t,
```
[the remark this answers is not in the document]
aschaffer: See above.

cpp/src/visitors/graph_make_visitor.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    result_ = return_t{std::move(graph_envelope)};
+  } else {
+    CUGRAPH_FAIL("Graph factory visitor not currently supported (multi_gpu == true).");
+  }
+}
+
+// EIDir's:
+//
+template class graph_maker_visitor<int, int, float, true, true>;
```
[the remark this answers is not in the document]
aschaffer: In the spirit of keeping PRs short and targetted I'd rather not handle this piece of functionality in this PR. It's unrelated and would trigger additional levels of separation (I cannot just divorce different SFINAEs of the same class into separate files without additional intermediary headers in between, etc.) Again, this is a lateral goal, separate from this PR. I can address that change in bulk afterwards.

cpp/src/visitors/rw_visitor.cpp
```diff
@@ -0,0 +1,145 @@
...
+    result_ = return_t{std::move(tpl_erased_result)};
+  } else {
+    CUGRAPH_FAIL(
+      "Unsupported RandomWalks algorithm (store_transposed == true or multi_gpu == true).");
+  }
+}
+
+// EIDir's:
+//
+template class rw_visitor<int, int, float, true, true>;
```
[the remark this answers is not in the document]
aschaffer: Same answer as above.

cpp/include/cugraph/visitors/ret_terased.hpp
```diff
@@ -28,23 +29,33 @@ struct return_t {
   struct base_return_t {
     virtual ~base_return_t(void) {}
 
-    virtual void copy(return_t const&)                       = 0;
+    virtual void copy(return_t&&)                            = 0;
```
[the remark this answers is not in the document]
aschaffer: It's specialized on `if constexpr (std::is_copy_constructible_v<T>)`, in which case it does a copy. The argument is just passed by r-value reference. From the user perspective they don't need to be exposed to the level of detail of whether a copy or a move is involved.

cpp/src/visitors/graph_make_visitor.cpp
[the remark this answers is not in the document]
ChuckHastings: Certainly a lateral goal.  Certainly the change I reference here (splitting things into separate files) happened after you began work on this.  I'm OK deferring in the interest of getting this PR merged.

Generally, I think if a PR is open for a long time and things change in the baseline that we should reflect those changes into the PR rather than add to the backlog of unfinished work.  Agree that this falls into the gray area since the MG/SG separation is still an incomplete task and this step can easily be added to either the next PR on this topic or the next MG/SG separation PR.  But if the MG/SG separation task had been completed during this time window then merging this PR could potentially have broken things.

ChuckHastings (approved):
Created https://github.com/rapidsai/cugraph/issues/1867 to remind us to address the review feedback that is being deferred.

rlratzel (approved):
Approving now that #1867 has been created.

ChuckHastings:
rerun tests

BradReesWork:
@gpucibot merge

PR #1808, opened 2021-09-07
Add new new raft symlink path to .gitignore

trxcllnt:


.gitignore
```diff
@@ -76,7 +76,7 @@ datasets/*
 cpp/doxygen/html
 
 # Raft symlink
-python/cugraph/raft
+python/cugraph/cugraph/raft
```
rlratzel, line 79:
```suggestion
python/cugraph/cugraph/raft
python/pylibcugraph/pylibcugraph/raft
```

rlratzel (changes requested):
We have another python package now that uses raft the same way.

rlratzel (approved):

rlratzel:
Thanks for this update!

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #1811, opened 2021-09-09
Changed code to disable `k_truss` on CUDA 11.4 differently

rlratzel:
Changed code to disable k_truss on CUDA 11.4 to not use `numba.cuda.runtime.get_version()` at import time since this creates a CUDA context which breaks dask `LocalCUDACluster` init (causes a nccl init invaid usage exception).

This code change results in almost the same UX as before, except explicit imports of the k_truss module no longer raise the exception immediately, and instead raise only when a k_truss algo is called.

Tested by ensuring unit tests for the conditional enabling of k_truss still work and also checked that a `LocalCUDACluster` could properly be created.

Also changed `dask_client` fixture to `module` since `session` was not compatible with tests that tried to create a `LocalCUDACluster` independently within the same session (rmat tests).

cc @cjnolet @pentschev

### DIFF
--- a/python/cugraph/conftest.py
+++ b/python/cugraph/conftest.py
@@ -24,9 +24,9 @@ from cugraph.comms import comms as Comms
 from cugraph.dask.common.mg_utils import get_visible_devices
 
 
-# session-wide fixtures
+# module-wide fixtures
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="module")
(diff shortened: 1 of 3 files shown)


BradReesWork (approved):

jnke2016 (approved):

BradReesWork:
@gpucibot merge

PR #1816, opened 2021-09-10
Group return values of renumber_edgelist and input parameters of graph_t & graph_view_t constructors.

seunghwak:
Group return values of ```renumber_edgelist``` (except for the main renumber_map) and return as a renumber_meta_t object.
Group input parameters of ```graph_t``` & ```graph_view_t``` constructors (except for the raft::handle and input parameters for the actual graph data) as a single graph_(view_)meta_t variable.

This is to make it easier to add more meta data in the future.


cpp/include/cugraph/graph_view.hpp
```diff
@@ -286,6 +286,33 @@ class graph_base_t : public graph_envelope_t::base_graph_t /*<- visitor logic*/
 
 }  // namespace detail
 
+template <typename vertex_t, typename edge_t, bool multi_gpu, typename Enable = void>
```
aschaffer, line 289: `graph_view_meta_t` seems repetitive of a functionality that `graph_meta_t` already offers. Is there a strong motivation for this apparent code duplication? Would it be possible to delegate this _grouping_ functionality to only one such `meta` pair?

cpp/src/structure/create_graph_from_edgelist_impl.hpp
```diff
@@ -108,21 +105,16 @@ create_graph_from_edgelist_impl(raft::handle_t const& handle,
...
-        handle,
-        local_vertex_span
-          ? std::optional<std::tuple<vertex_t const*, vertex_t>>{std::make_tuple(
-              (*local_vertex_span).data(), static_cast<vertex_t>((*local_vertex_span).size()))}
-          : std::nullopt,
-        major_ptrs,
-        minor_ptrs,
-        edgelist_edge_counts,
-        edgelist_intra_partition_segment_offsets);
+    std::tie(renumber_map_labels, meta) = cugraph::renumber_edgelist<vertex_t, edge_t, multi_gpu>(
```
aschaffer, line 108: Perhaps time to progressively replace `std::tie()` by C++17 structural bindings, which might be a bit more readable?

cpp/src/structure/create_graph_from_edgelist_impl.hpp
```diff
@@ -171,17 +163,16 @@ create_graph_from_edgelist_impl(raft::handle_t const& handle,
...
-    std::tie(*renumber_map_labels, *segment_offsets) =
-      cugraph::renumber_edgelist<vertex_t, edge_t, multi_gpu>(
-        handle,
-        vertex_span ? std::optional<std::tuple<vertex_t const*, vertex_t>>{std::make_tuple(
-                        (*vertex_span).data(), static_cast<vertex_t>((*vertex_span).size()))}
-                    : std::nullopt,
-        store_transposed ? edgelist_cols.data() : edgelist_rows.data(),
-        store_transposed ? edgelist_rows.data() : edgelist_cols.data(),
-        static_cast<edge_t>(edgelist_rows.size()));
+    std::tie(*renumber_map_labels, meta) = cugraph::renumber_edgelist<vertex_t, edge_t, multi_gpu>(
```
aschaffer, line 168: Same comment on `std::tie()` vs. structural bindings.

aschaffer (changes requested):

cpp/include/cugraph/graph_view.hpp
seunghwak: Yes, they are currently pretty much the same but this restructuring is mainly to pass additional parameters. When we support (key, value) pairs for row/column properties, pointers to unique rows/columns will be only passed to graph_view_meta_t.

So, these two will diverge (and possibly diverge even more over time), and that's the key motivation for having two separate classes.

cpp/src/structure/create_graph_from_edgelist_impl.hpp
seunghwak: This is not possible because renumber_map_labels & meta are defined outside the ```if (renumber) {}``` block.

cpp/src/structure/create_graph_from_edgelist_impl.hpp
seunghwak: Yeah... Here I used std::tie instead of structural binding as renumber_map_labels & meta are defined outside the ```{}``` block, but I don't see much need for this block (besides, releasing the host memory hold by ```std::vector<vertex_t const*> major_ptrs & minor_ptrs``` little early... I will update this.

aschaffer (approved):

cpp/include/cugraph/graph_view.hpp
seunghwak: There will be an overlap but the two will have non-overlapping parts.

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #1817, opened 2021-09-14
Updated to enforce only supported dtypes, changed to use legacy connected_components API

rlratzel:
* Updated to enforce only supported dtypes
* Changed to use legacy `connected_components` API so consistent arg types are used (to be updated to new `connected_components` APIs when available for both strongly and weakly CC).
* Updated tests to improve coverage and make easier to maintain.
* Added modified karate dataset for additional coverage.
* Moved `conftest.py` file to be closer to tests it is used in.

Tested by running the updated unit tests and comparing the hard-coded expected values to cuGraph, NetworkX, and SciPy.


BradReesWork (approved):

rlratzel:
rerun tests

BradReesWork:
@gpucibot merge

PR #1819, opened 2021-09-14
Updated python and cuda in scripts

jolorunyomi:
PYTHON variables and CUDA variables updated

### DIFF
--- a/ci/cpu/prebuild.sh
+++ b/ci/cpu/prebuild.sh
@@ -18,7 +18,7 @@ if [[ -z "$PROJECT_FLASH" || "$PROJECT_FLASH" == "0" ]]; then
     export BUILD_LIBCUGRAPH=1
 fi
 
-if [[ "$PYTHON" == "3.7" ]]; then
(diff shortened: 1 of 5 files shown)


rlratzel:
What happened to #1795 ? It looks like that one was closed without merging, but has a slightly different set of files with similar changes.  If they're related, can you describe how they differ, why the old one was closed, etc.? Also, can you update the description with why the variables are changing and if developers need to update anything in their environment?

BradReesWork:
rerun tests

BradReesWork:
@jolorunyomi the PR is not passing CI

jolorunyomi:
> What happened to #1795 ? It looks like that one was closed without merging, but has a slightly different set of files with similar changes. If they're related, can you describe how they differ, why the old one was closed, etc.? Also, can you update the description with why the variables are changing and if developers need to update anything in their environment?

The first PR had unrelated automated changes and so the work had to be re-done.

BradReesWork:
rerun tests

PR #1823, opened 2021-09-16
[REVIEW] Improve memory scaling for low average vertex degree graphs & many GPUs

seunghwak:
Assuming sqrt(P) is an integer, matrix row/column property array sizes in each GPU scale as V/sqrt(P) while the storage requirement for graph edges scale as E/P. If E/V is small and P is large, the O(V/sqrt(P)) part will dominate the memory requirement and analyzing N times larger graphs will require N^2 times more GPUs; this is unacceptable. However, in this case, at most E/P elements of the V/sqrt(P) sized array will be accessed, so no need to store the whole V/sqrt(P) values. Instead, we can store row/column properties in (key, value) pairs limiting the memory requirement to be the minimum of V/sqrt(P) and E/P.

This PR supports storing matrix row/column properties in (key, value) pairs if the percentage of actually accessed elements is lower than the threshold value (the code has been tested only up to 8 GPUs, and there was no clear benefit at this scale; currently the threshold value is set to 0 and (key, value) pair support is never enabled, but the threshold value will be adjusted later after large scale testing).


seunghwak:
rerun tests

seunghwak:
rerun tests

ChuckHastings (approved):
The description field for this PR should be updated with an actual description of the work.

BradReesWork:
@gpucibot merge

PR #1830, opened 2021-09-20
Make sure we keep the rapids-cmake and cugraph cal version in sync

robertmaynard:
When we make a new cugraph version, we need to also bump the rapids-cmake version at the same time. Otherwise we will get the previous releases dependencies by mistake.


ajschmidt8 (approved):

BradReesWork (approved):

BradReesWork:
@gpucibot merge

Issue #1753, opened 2021-08-03
[ENH] Create function to extract paths from BFS output


ChuckHastings:
BFS output is a data frame that contains, for every vertex in the graph, the distance from the seed and a back pointer to the previous element in the path.  Extracting the path from the seed to a particular vertex requires looking up values in the data frame to back track back to the seed.

If there are multiple vertices that you want to extract paths for this can be a very inefficient mechanism.

What would be helpful would be to create an efficient function that takes the data frame from BFS and a list of vertices and returns a list of the paths to the vertices.

Input:
* Dataframe from BFS call
* A list of vertices for which we want the path from the seed to each vertex
Output:
* Dataframe with one row for each vertex in the input list containing the path
* Columns (labeled something simple, like '0', '1', '2', ... 'n') representing step in the path
* Nulls (at the python level) indicating entries in the Dataframe are invalid

A few notes.
1. Not all paths will be the same length.  If the maximum path length is 5 and a path is only length 2, the unused vertices will be null
2. There is no requirement that there be a path to all vertices in the specified set in the input.  If there is no path then all elements in that row will be null.  Note that if the first column (column '0' as suggested above) is null then there is no path to the vertex for that row.
3. Observe that all nulls will be on the "right".  That is, in no row will there be a column 'i' that is null followed by a column 'j' that is not null (where j > i)

Implementation thoughts:
* Need a C++ function that does this.  Should be fairly straightforward
* Need a Python wrapper to encapsulate the C++ function.

PR #1838, opened 2021-09-23
[WIP] [skip-ci] Extract BFS paths implementation

ChuckHastings:
Partially addresses #1753 

Creates a new C++ function to extract BFS paths.

<still need to write unit tests and validate results>


cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,396 @@
...
+
+template <typename vertex_t>
+struct mg_partition_vertices {
+  int rank_;
+  vertex_t const* d_partition_offsets_begin_;
+  vertex_t const* d_partition_offsets_end_;
+
+  int __device__ operator()(vertex_t v) const
+  {
+#if 0
```
BradReesWork: This needs either a FIXME to drop/cleanup or should be a DEBUG or other defined type rather than "0"

ChuckHastings: Already pushed an update to remove this.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1125,6 +1125,47 @@ void bfs(raft::handle_t const& handle,
...
+ * @param graph_view Graph view object.
+ * @param distances Pointer to the distance array constructed by bfs.
+ * @param predecessors Pointer to the predecessor array constructed by bfs.
+ * @param destinations Destination vertices, extract path from source to each of these destinations
+ * @param n_destinations number of destinations (one source per component at most).
+ * @param paths Pointer to the output path array (preallocated to max_path_length * n_destinations elements).
+ * Paths will be placed into this array in the same order as the destinations input array.  Unused elements
+ * in the paths * will be set to invalid_vertex_id (-1 for a signed vertex_t,
+ * std::numeric_limits<vertex_t>::max() for an unsigned vertex_t type).
+ * @param max_path_length The maximum length of a path to extract.
```
seunghwak, line 1157: What do you think about

taking vertex_t * paths & max_path_length
vs
returning rmm::device_uvector<vertex_t> for paths and rmm::device_uvector<size_t> (or std::vector<size_t>) for offsets?

The latter sounds more functional to me and also users don't need to worry about max_path_length.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+  }
+};
+
+template <typename vertex_t>
+struct map_index_to_path_offset {
+  size_t max_path_length_;
+  vertex_t const* distances_;
+  vertex_t const* destinations_;
+
+  size_t __device__ operator()(vertex_t idx)
```
seunghwak, line 65: shouldn't this be ```size_t idx```?

cpp/tests/CMakeLists.txt
```diff
@@ -310,67 +310,73 @@ ConfigureTest(WCC_TEST components/wcc_test.cpp)
 ConfigureTest(HUNGARIAN_TEST linear_assignment/hungarian_test.cu)
 
 ###################################################################################################
-# - MST tests ----------------------------------------------------------------------------
+# - MST tests -------------------------------------------------------------------------------------
```
seunghwak, line 313: Thanks for fixing these :-)

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+  }
+};
+
+template <typename vertex_t>
+struct mg_lookup_distance {
+  vertex_t v_local_first_;
+  vertex_t const* distances_;
+
+  thrust::tuple<vertex_t, vertex_t, int> __device__
+  operator()(thrust::tuple<vertex_t, vertex_t, int> const& tuple)
```
seunghwak, line 92: What's the third tuple element (type int)?

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+  }
+};
+
+template <typename vertex_t>
+struct mg_lookup_predecessor {
+  vertex_t v_local_first_;
+  vertex_t const* predecessors_;
+
+  thrust::tuple<vertex_t, vertex_t, int> __device__
+  operator()(thrust::tuple<vertex_t, vertex_t, int> const& tuple)
```
seunghwak, line 106: What's the third tuple element (type int)?

cpp/tests/traversal/extract_bfs_paths_utils.cu
```diff
@@ -0,0 +1,100 @@
...
+#include <thrust/transform.h>
+
+template <bool multi_gpu, typename vertex_t>
+rmm::device_uvector<vertex_t> randomly_select_destinations(
+  raft::handle_t const& handle,
+  vertex_t number_of_vertices,
+  rmm::device_uvector<vertex_t> const& d_predecessors,
+  size_t num_paths_to_check,
+  uint64_t seed)
+{
```
seunghwak, line 33: So, if I am not mistaken, this randomly selects ```num_paths_to_check``` vertices (that are not sources).

I think we can achieve this by

thrust::copy_if();
thrust::shuffle();

seunghwak: And as this is only for test, another option is to do this in CPU, so no need to add this additional file and keep our codebase size in check.

seunghwak (commented):
Review part 1 (I am still reading extract_bfs_paths_impl.cuh).

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I definitely like the idea of returning the result rather than making the user pre-allocate it.  That is awkward in the unit test and will be awkward in the python implementation, I suspect.  The implementation can compute max_path_length much more easily.

The current expected uses of this function are python users.  I believe that returning a `num_destinations x max_path_length` dense matrix will be a more appropriate return for these users.

It would probably be useful to either:
* Create an API method to convert a CSR-style result like this into a dense matrix, or
* Create an API method to convert a dense matrix like this into a CSR-style result (or both)

My inclination, given the initial use will be easier with a dense result, would be to return:
`std::tuple<rmm::device_uvector<vertex_t>, vertex_t>`
(the dense matrix of the proper dimensions and the computed value of max_path_length).

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+      thrust::make_zip_iterator(
+        thrust::make_tuple(tmp_frontier.begin(), original_position.begin())),
+      tmp_frontier.size(),
+      [max_path_length, d_current_position = current_position.data()] __device__(auto tuple) {
+        auto distance = thrust::get<0>(tuple);
+        auto position = thrust::get<1>(tuple);
+
+        size_t tmp                   = position;
+        d_current_position[position] = (tmp * max_path_length) + distance;
+      });
```
seunghwak, line 262: So, the code here is basically to collect vertex properties for remote vertices.

I think this will be a recurring pattern, and we may better add something like

collect_values_for_vertices

mimicking the existing collecting_values_for_keys (this assumes that keys are arbitrary rather than contiguous, so using this as is will be somewhat inefficient, but we can copy-and-paste and replace the hash based query with contiguous array accesses with offsets).

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: The original rank.  Each rank has to retrieve results from another rank.  The current implementation is:
1. Construct the tuple (current_vertex, offset_in_current_frontier_array my_rank)
2. Shuffle the tuples based on which rank contains the predecessor information for current_vertex
3. Replace current_vertex with previous_vertex
4. Shuffle the tuples back to the original rank where the data is required
5. Update current_frontier in the proper location with the predecessor value

I could achieve the same result without the int parameter by writing a more sophisticated communication pattern (sort the vertices by which rank they are assigned to, do block transfers keeping track of the offsets for each rank, etc).  I suspect this would be more time efficient as well as memory efficient.  But I was trying to reuse existing functions as much as possible to get something working.  We can optimize this more later.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+  thrust::for_each_n(handle.get_thrust_policy(),
+                     thrust::make_zip_iterator(
+                       thrust::make_tuple(current_frontier.begin(), current_position.begin())),
+                     current_frontier.size(),
+                     detail::update_paths<vertex_t>{paths, invalid_vertex});
+
+  size_t count =
+    multi_gpu
+      ? cugraph::host_scalar_allreduce(
+          handle.get_comms(), current_frontier.size(), raft::comms::op_t::MAX, handle.get_stream())
```
seunghwak, line 284: Shouldn't this better be SUM reduction? (to minimize the difference in SG & MG logics)

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+      thrust::for_each_n(handle.get_thrust_policy(),
+                         thrust::make_zip_iterator(
+                           thrust::make_tuple(tmp_frontier.begin(), original_position.begin())),
+                         tmp_frontier.size(),
+                         [d_current_frontier = current_frontier.data()] __device__(auto tuple) {
+                           auto v        = thrust::get<0>(tuple);
+                           auto position = thrust::get<1>(tuple);
+
+                           d_current_frontier[position] = v;
+                         });
```
seunghwak, line 351: So, if I am not mistaken, this collects predecessors from the distributed predecessors array. And if we implement collect_values_for_vertices (as I mentioned in the code collecting remote distances from the distributed distances array), I think this code can be largely simplified.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+
+    thrust::for_each_n(handle.get_thrust_policy(),
+                       thrust::make_zip_iterator(
+                         thrust::make_tuple(current_frontier.begin(), current_position.begin())),
+                       current_frontier.size(),
+                       detail::update_paths<vertex_t>{paths, invalid_vertex});
+
+    count = multi_gpu ? cugraph::host_scalar_allreduce(handle.get_comms(),
+                                                       current_frontier.size(),
+                                                       raft::comms::op_t::MAX,
```
seunghwak, line 372: Shouldn't this better be SUM reduction? (to minimize the difference in SG & MG logics)

seunghwak (commented):
Review Part 2

cpp/tests/traversal/extract_bfs_paths_utils.cu
ChuckHastings: I can look into shuffle, that's one I haven't used before.

I can certainly look at a CPU implementation.  I was imagining that we might have other tests that might benefit from a function to randomly select vertices - although I started with it here because it was easier.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+template <typename vertex_t>
+std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<size_t>> shrink_extraction_list(
+  raft::handle_t const& handle,
+  rmm::device_uvector<vertex_t>&& vertex_list,
+  rmm::device_uvector<size_t>&& path_offset)
+{
+  auto constexpr invalid_vertex = invalid_vertex_id<vertex_t>::value;
+
+  rmm::device_uvector<vertex_t> new_vertex_list(vertex_list.size(), handle.get_stream());
+  rmm::device_uvector<size_t> new_path_offset(path_offset.size(), handle.get_stream());
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

cpp/include/cugraph/algorithms.hpp
seunghwak: > My inclination, given the initial use will be easier with a dense result, would be to return:
std::tuple<rmm::device_uvector<vertex_t>, vertex_t>
(the dense matrix of the proper dimensions and the computed value of max_path_length).

I second this, this sounds better than asking users to preallocate.

cpp/src/traversal/extract_bfs_paths_impl.cuh
seunghwak: Ah... OK, so I guess this code can go away if we implement collect_values_for_vertices.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+                  "Invalid input argument: destinations cannot be null");
+
+  //
+  //  Now what we want to do is to walk back the paths.
+  //
+  auto constexpr invalid_vertex = invalid_vertex_id<vertex_t>::value;
+
+  thrust::fill(handle.get_thrust_policy(),
+               paths,
+               paths + n_destinations * (max_path_length + 1),
```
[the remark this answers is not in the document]
ChuckHastings: The + 1 is incorrect.  Removed in the next push.  Another advantage to having the allocation occur inside the function :-)

cpp/tests/traversal/extract_bfs_paths_utils.cu
seunghwak: Yeah... so, for personalized PageRank, we're randomly selecting personalization vertices as well (it's currently implemented as host code).

https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/tests/link_analysis/pagerank_test.cpp#L178
https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/tests/link_analysis/mg_pagerank_test.cpp#L104

We may better combine these and add something under tests/utilities?

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: I think MAX is semantically what I want, although SUM would suffice.

I'm actually thinking I'll change this loop to a for loop from over the range [0, max_path_length), since we already know how many iterations are required.

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: Same as above.

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: Yes.  Do you think this simplification should be done in this PR, or can we defer it to the MG PR which will be in 21.12?

I could strip out all of the MG code from this PR and make the suggested improvements in the MG PR.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1125,6 +1125,47 @@ void bfs(raft::handle_t const& handle,
...
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * or multi-GPU (true).
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @param distances Pointer to the distance array constructed by bfs.
+ * @param predecessors Pointer to the predecessor array constructed by bfs.
+ * @param destinations Destination vertices, extract path from source to each of these destinations
```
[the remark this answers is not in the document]
seunghwak: I think we may still enforce this for consistency with other APIs (and we can avoid communication for the first step).

cpp/src/traversal/extract_bfs_paths_impl.cuh
seunghwak: > I'm actually thinking I'll change this loop to a for loop from over the range [0, max_path_length), since we already know how many iterations are required.

+1 for this, this will be clearer.

cpp/src/traversal/extract_bfs_paths_impl.cuh
seunghwak: Doesn't matter for me. You can either add a FIXME, strip out the code, or make fix in this release, or whatever... I just want to make sure we eventually make this code simpler & more reusable.

ChuckHastings:
Pushing out to 21.12 to incorporate more of the discussed changes.

cpp/src/traversal/extract_bfs_paths_impl.cuh
```diff
@@ -0,0 +1,378 @@
...
+        size_t tmp                   = position;
+        d_current_position[position] = (tmp * max_path_length) + distance;
+      });
+  } else {
+    thrust::transform(
+      handle.get_thrust_policy(),
+      thrust::make_counting_iterator<size_t>(0),
+      thrust::make_counting_iterator<size_t>(n_destinations),
+      current_position.data(),
+      detail::map_index_to_path_offset<vertex_t>{max_path_length, distances, destinations});
```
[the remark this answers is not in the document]
ChuckHastings: Done in next push

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Changed this back to this assumption.
 1. Added a check to validate that destinations are on the correct gpu
 2. Eliminated the initial shuffle logic (outside the loop in the implementation)
 3. Updated the documentation (including the other locations)

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: Changed to a for loop, definitely much clearer and saves the allreduce cost.

cpp/src/traversal/extract_bfs_paths_impl.cuh
ChuckHastings: Same fix as above

ChuckHastings:
@gpucibot merge

PR #1839, opened 2021-09-23
Changed value of the expensive check param to `false` in `populate_graph_container`

rlratzel:
Changed the value of the expensive check param to false in the call to `populate_graph_container`.

Since the expensive check is still valuable for test and debugging, the option will be exposed to the python user in a separate PR as part of a debug mode API, and for now it's assumed there's test coverage using the expensive check in the C++ tests.

see also #1840 

NOTES:
* This PR has the `bug` label since it prevents a bug from surfacing in the C++ expensive check, but it was not the root cause of the recent test failures. Setting expensive check to false needs to happen regardless for performance reasons.
* This was not built and tested locally yet!


seunghwak (approved):

ChuckHastings (approved):

rlratzel:
@gpucibot merge

PR #1842, opened 2021-09-23
add try/catch for python-louvain

BradReesWork:
added a catch for the test importing an uncommon module


rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #1853, opened 2021-09-28
Removed NetworkX requirement for type checks, fixed docstring, added new docstrings, import cleanups

### LINKED ISSUE #1782: [FEA] initial `pylibcugraph` support
- [ ] Make initial source tree changes (change `python` dir to have `cugraph` and `pylibcugraph` subdirs, move `setup.py`-related files, etc.)
- [ ] Update top-level `build.sh` script to add `pylibcugraph` as an optional target
- [ ] Add initial test (`python/pylibcugraph/pylibcugraph/tests`) that just tests that `import pylibcugraph` works
- [ ] Update `ci/gpu/build.sh` and `ci/test.sh` to include `pylibcugraph` build and test
- [ ] Add conda recipe for `pylibcugraph`, do test build and install and ensure `import pylibcugraph` works
- [x] Contact Ops team about these changes
- [x] Contact CuPy team about these changes
- [ ] Add support for `WCC` algo to `pylibcugraph`
- [ ] Update test to import and call `WCC`

### LINKED ISSUE #1807: [DOC] RMAT documentation example is badly formatted.
## Report incorrect documentation
RMAT generator example is badly formatted.

**Location of incorrect documentation**
https://docs.rapids.ai/api/cugraph/stable/api.html#module-cugraph.generators

### LINKED ISSUE #1851: [BUG] Pagerank requires NetworkX
Calling Pagerank results in an input type check that requires NetworkX.  This should be an optional check performed only if NetworkX is already installed.

```
# pagerank_demo.py

import cudf,cugraph
df=cudf.read_csv("/Projects/cugraph/datasets/dolphins.csv", delimiter=" ",
                 dtype=["int32", "int32", "float64"],header=None)
G = cugraph.DiGraph()
G.from_cudf_edgelist(df, source="0", destination="1", edge_attr="2")
ranks = cugraph.pagerank(G)
```

```
(rapids) [root@923fab474292 cugraph]# python pagerank_demo.py
  File "pagerank_demo.py", line 5, in <module>
    ranks = cugraph.pagerank(G)
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/cugraph-0+untagged.1.gc41ffd4-py3.8-linux-x86_64.egg/cugraph/link_analysis/pagerank.py", line 98, in pagerank
    G, isNx = cugraph.utilities.check_nx_graph(G, weight)
  File "/opt/conda/envs/rapids/lib/python3.8/site-packages/cugraph-0+untagged.1.gc41ffd4-py3.8-linux-x86_64.egg/cugraph/utilities/nx_factory.py", line 99, in check_nx_graph
    raise RuntimeError("NetworkX could not be imported, cannot check G")
RuntimeError: NetworkX could not be imported, cannot check G
```

cc @ChuckHastings

rlratzel:
* Changed type check code to no longer require NetworkX and instead only check Nx types if installed. #1851
* Cleaned up imports, removed unused imports.
* Changed `import_optional` to use `importlib` instead, and return `MissingModule` instances for easier error reporting.
* Fixed format of example in `rmat` docstring #1807
* Added docstrings for `pylibcugraph` `*_connected_components` APIs. #1782 

Tested changes to optional import and type checking code by uninstalling `NetworkX`, observing the failure shown in #1851, applying change, then testing to observe the same script runs to completion without Nx installed.

closes #1851 
closes #1807 
closes #1782

### DIFF
--- a/python/cugraph/cugraph/sampling/random_walks.py
+++ b/python/cugraph/cugraph/sampling/random_walks.py
@@ -57,6 +57,11 @@ def random_walks(G,
     if max_depth is None:
         raise TypeError("must specify a 'max_depth'")
 
+    # FIXME: supporting Nx types should mean having a return type that better
+    # matches Nx expectations (eg. data on the CPU, possibly using a different
+    # data struct like a dictionary, etc.). The 2nd value is ignored here,
+    # which is typically named isNx and used to convert the return type.
+    # Consider a different return type if Nx types are passed in.
     G, _ = ensure_cugraph_obj_for_nx(G)
 
     if start_vertices is int:
--- a/python/cugraph/cugraph/sampling/random_walks.py
+++ b/python/cugraph/cugraph/sampling/random_walks.py
@@ -13,7 +13,7 @@
 
 import cudf
 from cugraph.sampling import random_walks_wrapper
-import cugraph
+from cugraph.utilities import ensure_cugraph_obj_for_nx
 
 
 def random_walks(G,
@@ -57,7 +57,7 @@ def random_walks(G,
     if max_depth is None:
         raise TypeError("must specify a 'max_depth'")
 
-    G, _ = cugraph.utilities.check_nx_graph(G)
+    G, _ = ensure_cugraph_obj_for_nx(G)
 
     if start_vertices is int:
         start_vertices = [start_vertices]
--- a/python/cugraph/cugraph/centrality/betweenness_centrality.py
+++ b/python/cugraph/cugraph/centrality/betweenness_centrality.py
@@ -16,9 +16,10 @@ import numpy as np
 import cudf
 from cugraph.centrality import betweenness_centrality_wrapper
 from cugraph.centrality import edge_betweenness_centrality_wrapper
-from cugraph.utilities import df_edge_score_to_dictionary
-from cugraph.utilities import df_score_to_dictionary
-from cugraph.utilities import ensure_cugraph_obj_for_nx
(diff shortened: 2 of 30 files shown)


python/cugraph/cugraph/sampling/random_walks.py
```diff
@@ -57,7 +57,7 @@ def random_walks(G,
     if max_depth is None:
         raise TypeError("must specify a 'max_depth'")
 
-    G, _ = cugraph.utilities.check_nx_graph(G)
+    G, _ = ensure_cugraph_obj_for_nx(G)
```
BradReesWork, line 60: NetworkX does not have a "random_walks" function, which is why I assume the "isNx" flag is ignored.  But in that cases, shouldn't the whole check also be skipped?

BradReesWork (approved):

python/cugraph/cugraph/sampling/random_walks.py
rlratzel: offline discussion: if we support Nx types, then our return type should look closer to something Nx would return (Nx does not have a RW API, so we can't copy an API signature for this).
One option is to restrict that API to just cuGraph types, then open it up to Nx when we can later provide a conversion for the return type that is closer to Nx style (ie. on the CPU at least).  Another is do nothing and add a FIXME with these points while we decide.  We decided the FIXME would be best for now.

jnke2016 (approved):
Looks good

jnke2016 (approved):

BradReesWork:
@gpucibot merge

PR #1856, opened 2021-09-28
Copy v transform reduce out test

kaatish:
- Added testing for copy_v_transform_reduce_out_nbr primitive.
- Fixed bug in copy_v_transform_reduce_in_out_nbr kernels.

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -465,6 +465,10 @@ if(BUILD_CUGRAPH_MG_TESTS)
         ###########################################################################################
         # - MG PRIMS COUNT_IF_E tests -------------------------------------------------------------
         ConfigureTestMG(MG_COUNT_IF_E_TEST prims/mg_count_if_e.cu)
+
+        ###########################################################################################
+        # - MG PRIMS COPY_V_TRANSFORM_REDUCE_OUT tests --------------------------------------------
+        ConfigureTestMG(MG_COPY_V_TRANSFORM_REDUCE_OUT_TEST prims/mg_copy_v_transform_reduce_out_nbr.cu)
     else()
        message(FATAL_ERROR "OpenMPI NOT found, cannot build MG tests.")
     endif()
--- a/cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh
+++ b/cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh
@@ -267,6 +267,10 @@ __global__ void for_all_major_for_all_nbr_mid_degree(
   auto major_start_offset = static_cast<size_t>(major_first - matrix_partition.get_major_first());
   auto idx                = static_cast<size_t>(tid / raft::warp_size());
 
+  using WarpReduce = cub::WarpReduce<e_op_result_t>;
+  __shared__ typename WarpReduce::TempStorage
+    temp_storage[copy_v_transform_reduce_nbr_for_all_block_size / raft::warp_size()];
+
   property_add<e_op_result_t> edge_property_add{};
   while (idx < static_cast<size_t>(major_last - major_first)) {
     auto major_offset = major_start_offset + idx;
@@ -310,7 +314,8 @@ __global__ void for_all_major_for_all_nbr_mid_degree(
       }
     }
     if (update_major) {
-      e_op_result_sum = warp_reduce_edge_op_result<e_op_result_t>().compute(e_op_result_sum);
+      e_op_result_sum = WarpReduce(temp_storage[threadIdx.x / raft::warp_size()])
+                          .Reduce(e_op_result_sum, edge_property_add);
       if (lane_id == 0) { *(result_value_output_first + idx) = e_op_result_sum; }
     }
 
@@ -345,6 +350,10 @@ __global__ void for_all_major_for_all_nbr_high_degree(
 
   auto major_start_offset = static_cast<size_t>(major_first - matrix_partition.get_major_first());
   auto idx                = static_cast<size_t>(blockIdx.x);
+
+  using BlockReduce =
+    cub::BlockReduce<e_op_result_t, copy_v_transform_reduce_nbr_for_all_block_size>;
+  __shared__ typename BlockReduce::TempStorage temp_storage;
 
   property_add<e_op_result_t> edge_property_add{};
   while (idx < static_cast<size_t>(major_last - major_first)) {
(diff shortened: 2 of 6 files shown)


cpp/tests/CMakeLists.txt
```diff
@@ -465,6 +465,10 @@ if(BUILD_CUGRAPH_MG_TESTS)
         ###########################################################################################
         # - MG PRIMS COUNT_IF_E tests -------------------------------------------------------------
         ConfigureTestMG(MG_COUNT_IF_E_TEST prims/mg_count_if_e.cu)
+
+        ###########################################################################################
+        # - MG PRIMS COPY_V_TRANSFORM_REDUCE_OUT tests --------------------------------------------
+        ConfigureTestMG(MG_COPY_V_TRANSFORM_REDUCE_OUT_TEST prims/mg_copy_v_transform_reduce_out_nbr.cu)
```
seunghwak, line 471: So, no test for in_nbr?

seunghwak: And I think the name should better be ...OUT_NBR_TEST than ...OUT_TEST.

Better format this to multiple lines like

https://github.com/rapidsai/cugraph/pull/1856/files#diff-7caf2bdc6aa44337cf6e4b04392bc3e5940a7356512b4d525aba3bcda0be6c3bR436

seunghwak (commented):
Shouldn't we have a test for in_nbr as well?

Besides this & (some minor issues about formatting), this looks great to me.

cpp/tests/CMakeLists.txt
seunghwak: I think for each test, we can run both in_nbr & out_nbr, and rename test files to something like ...INOUT_NBR_TEST

ChuckHastings:
rerun tests

cpp/tests/CMakeLists.txt
```diff
@@ -468,7 +468,7 @@ if(BUILD_CUGRAPH_MG_TESTS)
 
         ###########################################################################################
         # - MG PRIMS COPY_V_TRANSFORM_REDUCE_OUT tests --------------------------------------------
```
seunghwak, line 470: COPY_V_TRANSFORM_REDUCE_OUT=>COPY_V_TRANSFORM_REDUCE_INOUT

cpp/tests/prims/mg_copy_v_transform_reduce_inout_nbr.cu
```diff
@@ -0,0 +1,506 @@
...
+          if (row_property < col_property) {
+            return row_property;
+          } else {
+            return col_property;
+          }
+        },
+        property_initial_value,
+        cugraph::get_dataframe_buffer_begin(global_out_result));
+      auto expected_out_result =
+        permute_result(handle, global_out_result, (*d_mg_renumber_map_labels));
```
seunghwak, line 393: Can't we use cugraph::test::sort_by_key (https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/tests/centrality/mg_katz_centrality_test.cpp#L148) instead (so no need to reinvent the wheel)?

cpp/tests/prims/mg_copy_v_transform_reduce_inout_nbr.cu
```diff
@@ -0,0 +1,506 @@
...
+          if (row_property < col_property) {
+            return row_property;
+          } else {
+            return col_property;
+          }
+        },
+        property_initial_value,
+        cugraph::get_dataframe_buffer_begin(global_in_result));
+      auto expected_in_result =
+        permute_result(handle, global_in_result, (*d_mg_renumber_map_labels));
```
seunghwak, line 413: Can't we use cugraph::test::sort_by_key (https://github.com/rapidsai/cugraph/blob/branch-21.10/cpp/tests/centrality/mg_katz_centrality_test.cpp#L148) instead (so no need to reinvent the wheel)?

seunghwak (commented):
Looks great besides few minor cosmetic issues.

And thanks a lot for finding & fixing the bug related to handling initial values.

cpp/tests/CMakeLists.txt
```diff
@@ -465,6 +465,10 @@ if(BUILD_CUGRAPH_MG_TESTS)
         ###########################################################################################
         # - MG PRIMS COUNT_IF_E tests -------------------------------------------------------------
         ConfigureTestMG(MG_COUNT_IF_E_TEST prims/mg_count_if_e.cu)
+
+        ###########################################################################################
+        # - MG PRIMS COPY_V_TRANSFORM_REDUCE_OUT tests --------------------------------------------
+        ConfigureTestMG(MG_COPY_V_TRANSFORM_REDUCE_INOUT_NBR_TEST prims/mg_copy_v_transform_reduce_inout_nbr.cu)
```
seunghwak, line 471: Yeah... and you can format this to multiple lines.

cpp/tests/prims/mg_copy_v_transform_reduce_inout_nbr.cu
kaatish: I will need to change sort_by_key to accept dataframe iterators instead of pointers. And it will need to be converted into a header only templated function since we cannot possible specialize for every tuple type we need for testing. Was the motivation behind specializing test::sort_by_key to reduce compile time? Is it a deal breaker if its not specialized anymore? or do you want to specialize it for a select few tuple types and restrict testing for those types in mg tests?

seunghwak: > I will need to change sort_by_key to accept dataframe iterators instead of pointers. And it will need to be converted into a header only templated function since we cannot possible specialize for every tuple type we need for testing. Was the motivation behind specializing test::sort_by_key to reduce compile time? Is it a deal breaker if its not specialized anymore? or do you want to specialize it for a select few tuple types and restrict testing for those types in mg tests?

The key motivation was to compile with g++ instead of requiring nvcc. This doesn't apply for primitives but for our C++ algorithms, we want to make sure they are useable with GCC. Let me think more about how we should handle this.

seunghwak: > do you want to specialize it for a select few tuple types and restrict testing for those types in mg tests?

Yeah... I think is this is a reasonable solution considering that this is only for testing. We may revisit this issue if we see more use cases for this (so specializing for every used types becomes to messy) later.

seunghwak: Maybe later... if we really need to provide this as a feature to users, we need to implement this by something like
```
convert std::tuple of rmm::device_uvector to std::vector of pointers

  for (size_t i = 0; i < tuple_size; ++i) {
    if (sizeof tuple element type == 1) {
      shuffle in a type agnostic way.
    }
    ...
    else if (sizeof tuple element type == 8) {
      shuffle in a type agnostic way.
    }
    else {
      CUGRAPH_FAIL(...);
    }
  }
```
But this sounds like a little too much work right at this moment.

seunghwak (approved):
This looks great!!!

BradReesWork:
rerun tests

BradReesWork:
rerun tests

ChuckHastings:
@gpucibot merge

PR #1882, opened 2021-10-13
Fixing Nx and Graph/DiGraph issues

BradReesWork:
Fixed a number of issues with conversion of Nx directed graphs.  Fix nx tests, fixed issues with using new Graph(directed=true) but checking for DiGraph


BradReesWork:
@gpucibot merge

PR #1870, opened 2021-10-06
Disable WCC test until we get get on an A100 to debug on

ChuckHastings:
CI testing on A100 is detecting a segmentation fault in the WCC implementation.  We currently can't get developer access to an A100 in order to debug this.  This PR will disable the WCC test.  Once we can resolve the A100 issue we can enable this test again.

PR #1883, opened 2021-10-14
Temporary workaround for CI issues with 11.0

ChuckHastings:
#1870 disabled some tests to allow us to push some things through CI.  While investigating this issue we discovered that we are having problem compiling the code in a container that uses the delivered 11.0 thrust.

This PR is a different temporary workaround.  We force downloading the desired version of thrust to be used as part of the cugraph build.  The PR also reactivates all of the WCC testing.


ChuckHastings:
@gpucibot merge

PR #1894, opened 2021-10-18
Raft Handle Updates to cuGraph

divyegala:
Depends on https://github.com/rapidsai/raft/pull/291


rlratzel (approved):

benchmarks/python_e2e/main.py
```diff
@@ -130,6 +130,7 @@ def run(algos,
         success = benchmark.run()
 
         algo_name = benchmark.results[1].name
+        algo_name = f"benchmarks.{algo_name}"
```
ChuckHastings, line 133: This seems like a questionable merge issue.

divyegala: Looks like it came from the auto-merge. Which one should I keep?

ChuckHastings: @rlratzel ?

ChuckHastings:
This PR includes merges that aren't in 22.02 yet (broken auto merge).  I will finish reviewing once the auto merge has brought things in sync.

rlratzel:
rerun tests

benchmarks/python_e2e/main.py
rlratzel: I think the code was just written across two lines using the same var name, it could have been combined into a single statement. You should keep the code with both lines 132 and 133 as-is.

ChuckHastings: I'm confused.  Doesn't this set `algo_name` to a value and then set `algo_name` to a different value?  What is the benefit of setting the same variable to two different values on successive lines?

rlratzel: The second assignment to `algo_name` is using the value from first assignment.  I'm guessing it was done this way to make it easier to see the two steps that form the final name.  It could have been written in one line as: 
```
# prepend the string "benchmark." to the name attribute of the benchmark result obj
algo_name = f"benchmarks.{benchmark.results[1].name}"
```

ChuckHastings: It does look like this is not part of your PR but something that was broken (if broken) somewhere else.  I would ignore this for now, we can correct it in a separate PR.

ChuckHastings (approved):
Still needs to be updated after the dependency is merged to point to the real raft.

Otherwise this looks good.  It would be good to get this merged soon, it's really ugly to review with all of the intervening commits.

rlratzel:
This PR will need #1968 merged in to branch 22.02.

rlratzel:
rerun tests

PR #1968 was merged in to 21.12 and the forward merge to 22.02 has finished.

rlratzel:
rerun tests

The latest cudf should fix the seg fault in the failing unit tests.

rlratzel:
@gpucibot merge

PR #1898, opened 2021-10-20
HITS primitive based implementation

### LINKED ISSUE #1657: [ENH] Implement HITS using the graph primitives
Our current implementation of HITS uses the gunrock implementation which is limited to 1 CPU.

Inherently HITS is the same computational pattern as Pagerank and Katz Centrality.  It should be possible to implement HITS using the graph primitives we use for Pagerank and Katz, giving us a HITS computation that can scale to multiple GPUs.

Suggest a sequence of PRs for this:

- [ ] Move the existing C++ implementation into a `cugraph::legacy` namespace, updating the python to reference the legacy implementation.  Define the new C++ API, define C++ unit tests using the C++ API.
- [ ] Implement the algorithm in C++, all unit tests should pass
- [ ] Implement the cython/python changes to support the new API
- [ ] Delete the legacy implementation

kaatish:
Depends on #1903. Fixes #1657


ChuckHastings (commented):
The approach looks correct.  But at this point the implementation isn't tied in (although the PR is still marked in-progress).

Be sure to instantiate hits for SG and MG in separate files - as Seunghwa mentioned in his comments.

It would be nice to have an MG test also in this PR, but if that has to be deferred until later, I'm OK with adding an issue to the backlog to add an MG test for hits and we can address that in a separate PR.

BradReesWork:
rerun tests

BradReesWork:
rerun tests

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1256,6 +1256,45 @@ void pagerank(raft::handle_t const& handle,
...
+ * @param authorities Pointer to the output authorities score array.
+ * @param epsilon Error tolerance to check convergence. Convergence is assumed if the sum of the
+ * differences in hub values between two consecutive iterations is less than @p epsilon
+ * @param max_iterations Maximum number of HITS iterations.
+ * @param has_initial_guess If set to `true`, values in the hubs output array (pointed by @p
+ * hubs) is used as initial hub values. If false, initial hub values are set to 1.0
+ * divided by the number of vertices in the graph.
+ * @param normalize If set to `true`, final hub and authority scores are normalized (the L1-norm of
+ * the returned hub and authority score arrays is 1.0) before returning.
+ * @param do_expensive_check A flag to run expensive checks for input arguments (if set to `true`).
```
seunghwak, line 1284: @return is missing here.

cpp/include/cugraph/prims/property_op_utils.cuh
```diff
@@ -171,6 +171,30 @@ constexpr auto op_dispatch(raft::comms::op_t op, F&& f)
...
+T identity(raft::comms::op_t op)
+{
+  switch (op) {
+    case raft::comms::op_t::SUM: {
+      return T{0};
+    } break;
+    case raft::comms::op_t::PROD: {
+      CUGRAPH_FAIL("raft::comms::op_t::PROD is not supported for op_dispatch");
+      return T{0};
+    } break;
```
seunghwak, line 184: Just better delete this as this will be caught in the default case.

cpp/include/cugraph/prims/property_op_utils.cuh
```diff
@@ -171,6 +171,30 @@ constexpr auto op_dispatch(raft::comms::op_t op, F&& f)
   };
 }
 
+template <typename T>
+T identity(raft::comms::op_t op)
```
seunghwak, line 175: I feel like this is a bit of misnomer...

identity sounds like an identity function... I will also think about a better name, but if you have a better one, I think we should rename this.

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -0,0 +1,201 @@
...
+#include <cugraph/algorithms.hpp>
+#include <cugraph/graph_view.hpp>
+#include <cugraph/prims/copy_to_adj_matrix_row_col.cuh>
+#include <cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh>
+#include <cugraph/prims/count_if_v.cuh>
+#include <cugraph/prims/reduce_v.cuh>
+#include <cugraph/prims/row_col_properties.cuh>
+#include <cugraph/prims/transform_reduce_v.cuh>
+
+// TODO : delete
```
seunghwak, line 29: delete :-)

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -0,0 +1,201 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include <thrust/fill.h>
+#include <thrust/transform.h>
```
seunghwak, line 19: Include statement for thrust should go between cugraph includes and standard library includes.

seunghwak: Our convention is to include lower level headers first.

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -0,0 +1,201 @@
...
+               GraphViewType const& graph_view,
+               result_t* hubs,
+               raft::comms::op_t op)
+{
+  auto hubs_norm = reduce_v(handle,
+                            graph_view,
+                            hubs,
+                            hubs + graph_view.get_number_of_local_vertices(),
+                            identity<result_t>(op),
+                            op);
```
seunghwak, line 45: Just to double check, hubs_norm can't be 0, right? If it can be 0, the transform below can be problematic and we need to be able to handle the case separately.

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -0,0 +1,201 @@
...
+
+  CUGRAPH_EXPECTS(epsilon >= 0.0, "Invalid input argument: epsilon should be non-negative.");
+
+  // Check validity of initial guess if supplied
+  if (has_initial_hubs_guess && do_expensive_check) {
+    auto num_negative_values =
+      count_if_v(handle, graph_view, hubs, [] __device__(auto val) { return val < 0.0; });
+    CUGRAPH_EXPECTS(num_negative_values == 0,
+                    "Invalid input argument: initial guess values should be non-negative.");
+    detail::normalize(handle, graph_view, hubs, raft::comms::op_t::SUM);
```
seunghwak, line 87: Should we normalize even when the input argument ```normalize``` is false? And ```do_expensive_check = true``` shouldn't change the program behavior. If the input is expected to be normalized, users should provide normalized values. If not, we should normalize outside the ```if (do_expensive_check)```.

cpp/tests/prims/mg_copy_v_transform_reduce_inout_nbr.cu
```diff
@@ -64,7 +64,8 @@ struct property_transform
     operator()(const vertex_t& val)
   {
     cuco::detail::MurmurHash3_32<vertex_t> hash_func{};
-    auto value = hash_func(val) % mod;
+    // auto value = hash_func(val) % mod;
+    auto value = 1;
```
seunghwak, line 68: What's the motivation for this change? And if you intended to make this change, better delete the commented out code.

cpp/tests/prims/mg_copy_v_transform_reduce_inout_nbr.cu
```diff
@@ -73,7 +74,8 @@ struct property_transform
     const vertex_t& val)
   {
     cuco::detail::MurmurHash3_32<vertex_t> hash_func{};
-    auto value = hash_func(val) % mod;
+    // auto value = hash_func(val) % mod;
+    auto value = 1;
```
seunghwak, line 78: What's the motivation for this change? And if you intended to make this change, better delete the commented out code.

cpp/tests/link_analysis/hits_test.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+template <typename T, typename L>
+std::vector<T> random_vector(L size, unsigned seed = 0)
+{
+  std::default_random_engine gen(seed);
+  std::uniform_real_distribution<T> dist(0.0, 1.0);
+  std::vector<T> v(size);
+  std::generate(v.begin(), v.end(), [&] { return dist(gen); });
+  return v;
+}
```
seunghwak, line 63: Should we better move these to tests/utilities?

cpp/tests/link_analysis/mg_hits_test.cpp
```diff
@@ -0,0 +1,661 @@
+#if 0
```
seunghwak, line 1: Better delete this.

cpp/tests/link_analysis/mg_hits_test.cpp
```diff
@@ -0,0 +1,661 @@
...
+
+template <typename T, typename L>
+std::vector<T> random_vector(L size, unsigned seed = 0)
+{
+  std::default_random_engine gen(seed);
+  std::uniform_real_distribution<T> dist(0.0, 1.0);
+  std::vector<T> v(size);
+  std::generate(v.begin(), v.end(), [&] { return dist(gen); });
+  return v;
+}
```
seunghwak, line 396: I guess this should better go to tests/utilities

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -0,0 +1,201 @@
...
+  using vertex_t = typename GraphViewType::vertex_type;
+  static_assert(std::is_integral<vertex_t>::value,
+                "GraphViewType::vertex_type should be integral.");
+  static_assert(std::is_floating_point<result_t>::value,
+                "result_t should be a floating-point type.");
+  static_assert(GraphViewType::is_adj_matrix_transposed,
+                "GraphViewType should support the pull model.");
+
+  auto const num_vertices = graph_view.get_number_of_vertices();
+  result_t hubs_error{std::numeric_limits<result_t>::max()};
```
seunghwak, line 74: I am a bit uncomfortable with this name. ```hubs_error``` sounds like something is wrong. This isn't really an error but very natural in many numerical methods with iterative solvers... I will also think about a better name.

seunghwak (commented):
Looks good in overall but I have some minor complaints about style & naming. I will also think about better names for few variables I complained.

cpp/src/link_analysis/hits_impl.cuh
kaatish: The term residual is used in numerical algorithms for calculating a stopping point.

seunghwak: I initially thought users will see this name, but this is used only inside the implementation, and I am less concerned about this name now. I may still use something like ```diff_sum``` instead, but I will stop complaining :-)

cpp/include/cugraph/prims/property_op_utils.cuh
kaatish: An operator also has the concept of identity.

See :
"Identity element - Wikipedia" https://en.m.wikipedia.org/wiki/Identity_element#Definitions

Basically any element when applied to a binary operator yields the other element, it's called an identity element. If this is confusing then sure we could change the name.

seunghwak: https://en.wikipedia.org/wiki/Identity_element

I think we'd better use ```identity_element``` or ```neutral_element```

seunghwak: You meant an operand? Yeah.... both ```idenity``` and ```ideneity_element``` can be used, but ```identity_element``` is clearer.

kaatish: How about passing an optional in place of init so that the internal implementation selects the appropriate init when nullopt is passed.

cpp/src/link_analysis/hits_impl.cuh
seunghwak: ```residual``` means in solving Ax=b, residual = b - Ax_approx.

Here, ```hubs_error``` is used for sum(diff(cur_hubs[] - prev_hubs[])). I guess these two are not same.

cpp/include/cugraph/prims/property_op_utils.cuh
seunghwak: Any motivation for that? Neither thrust nor std library does this, so I am negative unless there is a strong motivation for doing that.

cpp/src/link_analysis/hits_impl.cuh
kaatish: Norm cannot be 0. I can put a CUGRAPH_EXPECTS to throw an exception.

cpp/src/link_analysis/hits_impl.cuh
kaatish: How about this.,
We check if the input is normalized inside this block and throw logical error if it is not.

We normalize the input outside of this check block when has_initial_hubs_guess and normalize is true.

seunghwak (approved):
Looks great!!!

ChuckHastings (approved):

seunghwak:
> Depends on #1903. Fixes #1657
You need to update this comment.

ChuckHastings:
@gpucibot merge

Issue #1906, opened 2021-10-26
[ENH] Add graph creation methods to C API


ChuckHastings:
The new transpose method added in #1834 allows us to create a graph in one orientation and transpose it to another efficiently without needing to recreate everything.

In order to take advantage of this in the python layer, we need to create some create graph creation methods in the C API to return a type erased pointer to a graph.  Ultimately this will replace the current graph container/graph envelope mechanism in use by the current Cython.

PR #1907, opened 2021-10-26
C API for creating a graph

ChuckHastings:
Partially addresses #1906

This PR defines the API for graph creation and the pagerank and bfs calls that we will use to experiment with transposing a graph.

Some notes on the design here.

1. The intention is that the C API will handle renumbering (when set to true on graph creation).  This means that the opaque cugraph_graph_t pointer being populated by cugraph_sg_graph_create will contain the renumbering device vector and that the C API implementation of algorithms (pagerank and bfs demonstrated here) will unrenumber the result before returning
2. The intention is that the C API will understand whether the algorithm wants `store_transposed=true` or `store_transposed=false` and will call the transpose method if required.


cpp/src/c_api/array.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+    size_t nbytes = n_elems * (::data_type_sz[dtype]);
+
+    c_api::cugraph_type_erased_host_array_t* ret_value =
+      new c_api::cugraph_type_erased_host_array_t{new std::byte[nbytes], n_elems, nbytes, dtype};
+
+    *array = reinterpret_cast<cugraph_type_erased_host_array_t*>(ret_value);
+    return CUGRAPH_SUCCESS;
+  } catch (...) {
+    return CUGRAPH_UNKNOWN_ERROR;
```
rlratzel, line 115: Should there be a `FIXME` here to add an exception handler that checks for more error situations and sets the code with better information (eg. ALLOC error, etc.).

cpp/tests/c_api/c_test_utils.h
```diff
@@ -17,6 +17,12 @@
 #include <stdio.h>
 #include <time.h>
 
+#define TEST_ASSERT(RETURN_VALUE, STATEMENT, MESSAGE)                    \
+  {                                                                      \
+    (RETURN_VALUE) = !(STATEMENT);                                       \
+    if ((RETURN_VALUE)) { printf("ASSERTION FAILED: %s\n", (MESSAGE)); } \
+  }
```
rlratzel, line 24: This might be a good place to have the macro take an additional arg that can be used for the overall test return value:
```suggestion
#define TEST_ASSERT(RETURN_VALUE, STATEMENT, MESSAGE, OVERALL_VALUE)                    \
  {                                                                      \
    (RETURN_VALUE) = !(STATEMENT);                                       \
    OVERALL_VALUE |= RETURN_VALUE; \
    if ((RETURN_VALUE)) { printf("ASSERTION FAILED: %s\n", (MESSAGE)); } \
  }
```
(I don't think `OVERALL_VALUE` is a great name, but you get the idea)

rlratzel (commented):
Looks good, I just had a few suggestions.

aschaffer (approved):

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -127,7 +128,7 @@ cugraph_error_t cugraph_random_walks(const cugraph_raft_handle_t* ptr_handle,
                                      cugraph_rw_ret_t* ret);
 
 /* SG graph allocator*/
-cugraph_graph_envelope_t* cugraph_make_sg_graph(const cugraph_raft_handle_t* p_handle,
+cugraph_graph_envelope_t* cugraph_make_sg_graph(const cugraph_handle_t* p_handle,
```
rlratzel, line 131: Is the idea that this will be replaced by `cugraph_sg_graph_create()`, or are they different?

cpp/include/cugraph_c/cugraph_api.h
```diff
@@ -42,16 +43,16 @@ extern int data_type_sz[];
 
 /* C stub declarations */
 
-typedef struct cugraph_raft_handle_ {
-  int allign_;
-} cugraph_raft_handle_t;
+typedef struct cugraph_handle_ {
```
rlratzel, line 46: I like this change. I noticed that this type is described in the API doxygen comments as a "Handle for accessing resources".  Would a more self-documenting name be `cugraph_resource_handle_t`?  This would also make it less likely to be confused with a handle to a `cugraph_graph_t` or something similar.

cpp/include/cugraph_c/cugraph_api.h
ChuckHastings: Yes.  I left the version that makes a graph envelope, it is currently used by the prototype.  My expectation is that we will migrate away from the envelope version since transpose will be more efficient overall.

cpp/include/cugraph_c/algorithms.h
```diff
@@ -0,0 +1,204 @@
...
+ *                          in the graph multiplied by @p epsilon.
+ * @param [in]  max_iterations Maximum number of PageRank iterations.
+ * @param [in]  has_initial_guess If set to `true`, values in the PageRank output array (pointed by
+ * @p pageranks) is used as initial PageRank values. If false, initial PageRank values are set
+ * to 1.0 divided by the number of vertices in the graph.
+ * @param do_expensive_check A flag to run expensive checks for input arguments (if set to `true`).
+ * @param [out] result      Opaque pointer to pagerank results
+ * @return error code
+ */
+cugraph_error_t cugraph_pagerank(
```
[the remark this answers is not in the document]
ChuckHastings: I like this idea.  Let me add some methods to the API for this, hopefully in the next push.

cpp/tests/c_api/c_test_utils.h
ChuckHastings: I threw this in here primarily to simplify the test I wrote.  I actually feel like we probably should make exploring this a separate issue/PR.  I'm hesitant to make changes to this without thinking about the big picture.

The C++ unit tests all use RAAI constructs so that we don't leak resources and can run a series of tests within the same executable and see all of the failures in a single run of the test.  Without RAII constructs we can leak resources (most important is GPU memory) and potentially introduce phantom failures later in the sequence of tests (e.g. tests that fail because they run out of memory, although they would succeed if the memory had been freed).

The C unit tests can't use RAAI constructs.  I think we need to explore the following ideas:
* How extensively do we want to test the C API?
* Do we want to be able to have multiple tests and see all of the failures?
* How much effort do we want to go to in avoiding resource leaks in the tests when tests fail?

Based on the answers to these questions we can take some of these test cases and identify what macros we want to create that will make the unit tests easy to develop, read and maintain while satisfying the need.

rlratzel: I created this:  https://github.com/rapidsai/cugraph/issues/1923

cpp/src/c_api/array.cpp
rlratzel: Looks like my main concern was handled by setting the exception message on the new error type.

rlratzel (approved):

cpp/tests/c_api/c_test_utils.h
seunghwak: Not a request to change anything, but there should be no resource leak in tests code (we should at least try hard to achieve that. We shouldn't assume that resource leaks in test code is acceptable).

ChuckHastings:
rerun tests

seunghwak (approved):

BradReesWork:
@gpucibot merge

ChuckHastings:
rerun tests

rlratzel:
rerun tests

PR #1915, opened 2021-10-28
Added separate helpers for moving buffers to either cudf column and series objects

rlratzel:
This is a follow-up PR to address [this feedback](https://github.com/rapidsai/cugraph/pull/1901#discussion_r737550127) from #1901 

This adds a separate cython helper for moving a buffer to a column to allow for a slightly more efficient call where column objects can be used.


BradReesWork (approved):

ChuckHastings (approved):

jnke2016 (approved):

BradReesWork:
@gpucibot merge

PR #1918, opened 2021-11-01
Enable crosslink to rmm

AyodeAwe:
This PR updates cpp/doxygen/Doxyfile to consume the generated Doxygen tags from rmm (see rapidsai/rmm#672). This will enable linking between the cugraph docs and rmm docs.

### DIFF
--- a/cpp/doxygen/Doxyfile
+++ b/cpp/doxygen/Doxyfile
@@ -2075,7 +2075,7 @@ SKIP_FUNCTION_MACROS   = YES
 # the path). If a tag file is not located in the directory in which doxygen is
 # run, you must also specify the path to the tagfile here.
 
-TAGFILES               = rmm.tag=https://docs.rapids.ai/api/librmm/0.18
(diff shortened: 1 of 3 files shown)


cpp/doxygen/Doxyfile
```diff
@@ -2075,7 +2075,7 @@ SKIP_FUNCTION_MACROS   = YES
 # the path). If a tag file is not located in the directory in which doxygen is
 # run, you must also specify the path to the tagfile here.
 
-TAGFILES               =
+TAGFILES               = rmm.tag=https://docs.rapids.ai/api/librmm/0.18
```
ajschmidt8, line 2078:
```suggestion
TAGFILES               = rmm.tag=https://docs.rapids.ai/api/librmm/21.12
```

This should be the latest branch version, which is now `21.12`.

ajschmidt8 (changes requested):
Looks good. One change requested. Also please add a description to the PR describing the changes.

ajschmidt8 (approved):

PR #1920, opened 2021-11-01
Fix metadata mismatch

jnke2016:
A recent change causes some of the dask's partition dataframe to have different types. 
This PR ensures the dask partition after renumbering are the same as the initial dask dataframe

Note: Still looking for the root cause of this change


python/cugraph/cugraph/structure/number_map.py
```diff
@@ -480,10 +481,12 @@ def renumber_and_segment(
...
             renumber_type = 'experimental'
 
-        renumber_map = NumberMap()
         if not isinstance(src_col_names, list):
             src_col_names = [src_col_names]
             dst_col_names = [dst_col_names]
+
+        id_type = df[src_col_names[0]].dtype
+        renumber_map = NumberMap(id_type=id_type)
         if type(df) is cudf.DataFrame:
```
BradReesWork, line 490: why not use isinstance?

BradReesWork (approved):

rlratzel (approved):
Approved assuming Brad's feedback will be resolved.

ChuckHastings (approved):

rlratzel:
rerun tests

BradReesWork:
@gpucibot merge

PR #1921, opened 2021-11-02
improved Nx conversion performance

BradReesWork:
refactored how to convert from Nx to cuGraph to improve performance

### DIFF
--- a/python/cugraph/cugraph/utilities/nx_factory.py
+++ b/python/cugraph/cugraph/utilities/nx_factory.py
@@ -82,7 +82,7 @@ def convert_from_nx(nxG, weight=None, do_renumber=True):
     elif isinstance(nxG, nx.classes.graph.Graph):
         G = cugraph.Graph()
     else:
-        raise ValueError(
+        raise TypeError(f"nxG must be either a NetworkX Graph or DiGraph, got {type(nxG)}")
             "input is not a supported NetworkX graph type")
 
     is_weighted = nx.is_weighted(nxG)
--- a/python/cugraph/cugraph/utilities/nx_factory.py
+++ b/python/cugraph/cugraph/utilities/nx_factory.py
@@ -160,7 +160,6 @@ def df_edge_score_to_dictionary(df, k, src="src", dst="dst"):
     dst : str
         destination column name
 
-
     Returns
     -------
     dict : Dictionary of vertices and score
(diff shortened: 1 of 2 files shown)


python/cugraph/cugraph/utilities/nx_factory.py
```diff
@@ -28,7 +29,82 @@
 nx = import_optional("networkx")
 
 
+def convert_unweighted_to_gdf(NX_G):
+    _edges = NX_G.edges(data=False)
+    src = [s for s, _ in _edges]
+    dst = [d for _, d in _edges]
```
rlratzel, line 35: You can do this in one pass using `zip()`
```suggestion
    (src, dst) = zip(*_edges)
```

python/cugraph/cugraph/utilities/nx_factory.py
```diff
@@ -28,7 +29,82 @@
...
+
+    return _gdf
+
+
+def convert_weighted_named_to_gdf(NX_G, weight):
+    _edges = NX_G.edges(data=weight)
+
+    src = [s for s, _, _ in _edges]
+    dst = [d for _, d, _ in _edges]
+    wt = [w for _, _, w in _edges]
```
rlratzel, line 49: same feedback as above re: `zip()`

python/cugraph/cugraph/utilities/nx_factory.py
```diff
@@ -28,7 +29,82 @@
...
+    weight: weight column name. Only used if
+    nxG.is_weighted() is True
+    """
+
+    if isinstance(nxG, nx.classes.digraph.DiGraph):
+        G = cugraph.Graph(directed=True)
+    elif isinstance(nxG, nx.classes.graph.Graph):
+        G = cugraph.Graph()
+    else:
+        raise ValueError(
```
rlratzel, line 85: This should probably be a `TypeError`

python/cugraph/cugraph/utilities/nx_factory.py
BradReesWork: I tried that and you can do it in one pass but the performance is not as good

BradReesWork: two pass takes 500ms,  the zip approach take 1.2 sec

rlratzel: ah ok, I'll just mark as resolved then.

python/cugraph/cugraph/utilities/nx_factory.py
BradReesWork: good catch

rlratzel (approved):

BradReesWork:
rerun tests

rlratzel:
rerun tests

BradReesWork:
@gpucibot merge

PR #1914, opened 2021-10-28
Update graph_classes.py

bmg-pcl:
Clarify exception message on Graph constructor. 

Colleague lost some time tracing this!

PR #1925, opened 2021-11-03
Updated error message and using a proper TypeError exception when an invalid MultiGraph is passed in

rlratzel:
* Updated error message and using a proper TypeError exception when an invalid MultiGraph is passed in.
* Added test to verify.

This PR replaces #1914 since that one is not passing the style check and the author has not responded.


rlratzel:
rerun tests

BradReesWork:
@gpucibot merge

PR #1926, opened 2021-11-04
Ensure empty shuffled columns have the appropriate dtype

jnke2016:
Workers having no data after shuffling end up holding empty columns(series) of type object which are not supported by `__cuda_array_interface__`

Ensure that the empty columns have the appropriate dtype


python/cugraph/cugraph/structure/renumber_wrapper.pyx
```diff
@@ -49,14 +49,20 @@ cdef renumber_helper(shuffled_vertices_t* ptr_maj_min_w, vertex_t, weights):
         move(pair_s_minor.first), vertex_t, "shuffled_minor")
 
     shuffled_df = cudf.DataFrame()
+    # Some workers might have no data therefore ensure the empty column have the appropriate
+    # vertex_t or weight_t. Failing to do that will create am empty column of type object
+    # which is not supported by '__cuda_array_interface__'
     shuffled_df['major_vertices']=shuffled_major_series
+    shuffled_df['major_vertices']=shuffled_df['major_vertices'].astype(vertex_t)
```
BradReesWork, line 56: I guess you were not able to just add the "astype" to the series?   What is the performance hit of doing an "astype"?

jnke2016: Performance wised, it is not too bad because this operation mainly target to change the type of a `None` column to the appropriate dtype.  As this PR stated, some workers end up with no data after shuffling hence `shuffled_major_series` end up being None which translate to a column of type object(causing the error). Regarding the case where `shuffled_major_series` is a series with data, it will already have the appropriate dtype hence I will be doing a cast of the same dtype which takes 10^-3 seconds for a series where the dataset is of scale 24. I can add a condition to only cast when `shuffled_major_series` is None

python/cugraph/cugraph/structure/renumber_wrapper.pyx
```diff
@@ -49,14 +49,20 @@ cdef renumber_helper(shuffled_vertices_t* ptr_maj_min_w, vertex_t, weights):
         move(pair_s_minor.first), vertex_t, "shuffled_minor")
```
rlratzel, line 50: Can you do this instead (and also for `shuffled_weight_series` below)? This might be more self-documenting, and should also address @BradReesWork 's concern about always doing two operations on the df.
```suggestion
    if shuffled_major_series is None:
        shuffled_major_series = cudf.Series(dtype=vertex_t)
    if shuffled_minor_series is None:
        shuffled_minor_series = cudf.Series(dtype=vertex_t)
```

python/cugraph/cugraph/structure/renumber_wrapper.pyx
jnke2016: Addressed

python/cugraph/cugraph/structure/renumber_wrapper.pyx
jnke2016: Addressed

BradReesWork (approved):

BradReesWork:
rerun tests

rlratzel (approved):

BradReesWork:
@jnke2016 any insight into why tests are failing (gpuCI-cuda11.5/cugraph/gpu/cuda/11.5/python/3.8/centos7 )?

rlratzel:
rerun tests

python/cugraph/cugraph/structure/renumber_wrapper.pyx
jnke2016: @BradReesWork regarding this comment (I didn't see a better way to respond to your last comments)

> any insight into why tests are failing (gpuCI-cuda11.5/cugraph/gpu/cuda/11.5/python/3.8/centos7 )?

A lot of the 11.5 tests are failing but I though we could still merge PR for now though. I will look into this and see what's going on

rlratzel:
@gpucibot merge

PR #1927, opened 2021-11-04
add new demo notebook for louvain

ChuckHastings:
Added a demonstration notebook for running MNMG Louvain

### DIFF
--- a/notebooks/demo/mg_louvain.ipynb
+++ b/notebooks/demo/mg_louvain.ipynb
@@ -211,7 +211,7 @@
     "# Start Pagerank timer\n",
     "t_start = time.time()\n",
     "\n",
-    "# Get the pagerank scores\n",
+    "# Get the Louvain partition assignments for each vertex and the global modularity score.\n",
     "(louvain_df, modularity) = dask_cugraph.louvain(G)\n",
     "\n",
     "# Print time\n",
--- a/notebooks/demo/mg_louvain.ipynb
+++ b/notebooks/demo/mg_louvain.ipynb
@@ -29,7 +29,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Import needed libraries. We recommend using cugraph_dev env through conda\n",
+    "# Import needed libraries. We recommend using the [cugraph_dev](https://github.com/rapidsai/cugraph/tree/branch-21.12/conda/environments) env through conda\n",
     "from dask.distributed import Client, wait\n",
     "from dask_cuda import LocalCUDACluster\n",
     "import cugraph.comms as Comms\n",
--- a/notebooks/demo/mg_louvain.ipynb
+++ b/notebooks/demo/mg_louvain.ipynb
@@ -1,0 +1,403 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Multiple GPU Louvain in cuGraph\n",
+    "#### Author : Chuck Hastings\n",
+    "\n",
+    "In this notebook, we will show how to use multiple GPUs in cuGraph to compute the PageRank of each user in Twitter's dataset.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


notebooks/demo/mg_louvain.ipynb
```diff
@@ -0,0 +1,403 @@
...
+    "### Basic setup"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Import needed libraries. We recommend using cugraph_dev env through conda\n",
```
rlratzel, line 32:
```suggestion
    "# Import needed libraries. We recommend using the [cugraph_dev](https://github.com/rapidsai/cugraph/tree/branch-21.12/conda/environments) env through conda\n",
```

notebooks/demo/mg_louvain.ipynb
```diff
@@ -0,0 +1,403 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The Hollywood dataset is in our S3 bucket and zipped.  \n",
+    "1. We'll need to create a folder for our data in the `/data` folder\n",
+    "1. Download the zipped data into that folder from S3 (it will take some time as it it 6GB)\n",
+    "1. Decompress the zipped data for use (it will take some time as it it 26GB)"
```
rlratzel, line 56: Do these get rendered properly in jupyter or do they all show up as `1.` as seen here?

notebooks/demo/mg_louvain.ipynb
```diff
@@ -0,0 +1,403 @@
...
+    }
+   ],
+   "source": [
+    "import urllib.request\n",
+    "import os\n",
+    "\n",
+    "data_dir = '../data/'\n",
+    "if not os.path.exists(data_dir):\n",
+    "    print('creating data directory')\n",
+    "    os.system('mkdir ../data')\n",
```
rlratzel, line 81: A more modern way would be to use `pathlib`:
```
import pathlib
pathlib.Path("../data").mkdir()
```
but then you'd be better off updating all the path operations to use `pathlib` for consistency, so if you'd rather stick with the `os` module already imported:
```
os.mkdir("../data")
```

notebooks/demo/mg_louvain.ipynb
```diff
@@ -0,0 +1,403 @@
...
+     "text": [
+      "Louvain:  8.262660264968872 s\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Start Pagerank timer\n",
+    "t_start = time.time()\n",
+    "\n",
+    "# Get the pagerank scores\n",
```
rlratzel, line 214:
```suggestion
    "# Get the Louvain partition assignments for each vertex and the global modularity score.\n",
```

rlratzel (changes requested):

notebooks/demo/mg_louvain.ipynb
ChuckHastings: We had briefly discussed in slack the idea of making a set of utility functions to support this in a more general way (retrieving test data).  Would you like me to update this, or do you want to defer until we have a better tool?

notebooks/demo/mg_louvain.ipynb
ChuckHastings: They get rendered properly in jupyter.

ChuckHastings:
rerun tests

notebooks/demo/mg_louvain.ipynb
rlratzel: Ah, thanks for the reminder.  I've added [this issue](https://github.com/rapidsai/cugraph/issues/1935) to track it, so feel free to either implement it if you'd like or ignore it in favor of a future utility call.

ChuckHastings:
@gpucibot merge

PR #1928, opened 2021-11-04
Updates to support correct comparisons of cuDF Series with different names

rlratzel:
cuDF PR https://github.com/rapidsai/cudf/pull/9542 fixes a problem where comparisons between Series objects would return `True` even if the Series names were different, which is inconsistent with Pandas.  Some cuGraph tests relied on the old behavior which ignored names and were failing.

This PR updates those tests to use the proper comparison utility provided by cuDF to restore the intended behavior of the test assertions.

This PR also updates the `gpu` CI script to use `mamba` for an install step to hopefully speed up CI runs.


BradReesWork (approved):

ChuckHastings (approved):

jnke2016 (approved):

BradReesWork:
@gpucibot merge

ajschmidt8 (approved):

PR #1932, opened 2021-11-05
Use collections.abc.Sequence instead of deprecated collections.Sequence.

bdice:
Hi cuGraph team! I was looking at a couple things across RAPIDS related to Python 3.10 compatibility. There is one spot where the deprecated name `collections.Sequence` is used. This was moved to `collections.abc.Sequence` in Python 3.3 and has been deprecated for a while. In Python 3.10 the deprecated names for abstract base classes like `collections.Sequence` were removed.

This PR fixes that module path to help with eventual compatibility with Python 3.10+. I think this is the only instance of this deprecated name across RAPIDS packages.

(I don't have permissions to set labels or project boards, so the label checker will fail until someone else does that.)


rlratzel (approved):
Thank you for this PR @bdice

rlratzel:
@gpucibot merge

PR #1937, opened 2021-11-09
Accounting for raft::random detail changes

divyegala:
Depends on https://github.com/rapidsai/raft/pull/356/

PR #1944, opened 2021-11-11
Fix compile error due to recent RAFT update (rng.cuh => rng.hpp)

seunghwak:
In the recent raft PR (https://github.com/rapidsai/raft/pull/356), rng.cuh is renamed to rng.hpp and cugraph is no longer compiling.

This PR fixes the compile error.


BradReesWork (approved):

seunghwak:
Duplicate of #1937

PR #1863, opened 2021-09-30
[FEA] Remove FAISS dependency, inherit other common dependencies from raft

trxcllnt:
Originally this PR was about building FAISS shared libs via CPM, but @rlratzel mentioned cuGraph doesn't need FAISS anymore, and it'd be better if we remove it.

If we need FAISS again in the future, we can add `faiss::faiss` back to `target_link_libraries` without needing extra CPM configuration as it will be available [via raft](https://github.com/rapidsai/raft/pull/345).

Edit: Removed more dependencies that we inherit from raft and/or rmm. Depends on https://github.com/rapidsai/raft/pull/345.

Edit 2: I think the CUDA 11.0 thrust issue will be solved by https://github.com/rapidsai/rapids-cmake/pull/98

PR #1945, opened 2021-11-11
cugraph bring in the same cuco as raft and cudf

robertmaynard:


ChuckHastings (commented):
Does this mean we won't be merging #1863 into 21.12?  That PR would make this change obsolete.

robertmaynard:
I mainly proposed this to correct issues people might see before #1863 has been merged.

ChuckHastings (approved):

ChuckHastings:
@gpucibot merge

PR #1950, opened 2021-11-17
Fix the difference in 2D partitioning of GPUs in python and C++

seunghwak:
If sqrt(# GPUs) is not an integer, prows > pcols in python and prows < pcols in C++.

The C++'s choice leads to better performance (especially for BFS & SSSP and on the systems with low inter-GPU communication bandwidth) and this reduces the difference between python and C++ testing/benchmarking.


rlratzel:
rerun tests

BradReesWork (approved):

python/cugraph/cugraph/comms/comms.py
```diff
@@ -26,10 +26,10 @@
 
 
 def __get_2D_div(ngpus):
-    pcols = int(math.sqrt(ngpus))
-    while ngpus % pcols != 0:
-        pcols = pcols - 1
-    return int(ngpus/pcols), pcols
+    prows = int(math.sqrt(ngpus))
+    while ngpus % prows != 0:
```
BradReesWork, line 30: this will be a very fast loop, but you should have just been able to subtract the modulo result from prows rather than looping.

rlratzel (approved):

jnke2016 (approved):

python/cugraph/cugraph/comms/comms.py
seunghwak: Can we do this? I am not sure I am fully understanding your intention, but here `prows` is varying.

Say `ngpus` = 14. Then, `int(sqrt(14))` will be 3. `14 % 3` will be 2. `prows (3) - 2` will be 1. But we want `prows` to be 2.

This code is based on the discussions from https://stackoverflow.com/questions/16266931/input-an-integer-find-the-two-closest-integers-which-when-multiplied-equal-th

BradReesWork:
@gpucibot merge

PR #1951, opened 2021-11-17
Match python R-mat parameters to Graph 500

seunghwak:
Python tests currently use a=0.1, b=0.2, c=0.3 and this doesn't match with the Graph 500 specification. This matches the parameters with Graph 500.


seunghwak:
Duplicate of #1952

PR #1954, opened 2021-11-18
Explicitly install cusolver version with the correct ABI version

robertmaynard:
Starting with CUDA 11.1, cusolver updated the major SOVERSION from 10 to 11. To allow execution with the cudatoolkit=11.0 we
need to package the ABI 11 version.

### DIFF
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -47,6 +47,7 @@ requirements:
     - ucx-proc=*=gpu
     - faiss-proc=*=cuda
     - libfaiss 1.7.0 *_cuda
+    - libcusolver>=11.2.1
 
 about:
   home: http://rapids.ai/


robertmaynard:
@dantegd @quasiben

BradReesWork (approved):

BradReesWork:
@robertmaynard one minor file conflict

cpp/CMakeLists.txt
```diff
@@ -290,7 +300,7 @@ target_link_libraries(cugraph
         cuco::cuco
         CUDA::cublas
         CUDA::curand
-        CUDA::cusolver
+	${cusolver_target}
```
rlratzel, line 303: minor suggestion: indent

rlratzel (approved):

robertmaynard:
Revised our approach as the static linking required us to bring in cublas, cublasLT and some other components increasing the binary size by more compared to the dynamic libcusolver library.

Instead we state our conda package requires the new libcusolver ABI version explicitly so we still work with 11.0

ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

Issue #1891, opened 2021-10-18
[ENH] Add extract_bfs_paths to C API


ChuckHastings:
We are migrating to having python calling the C API.  We should add the new `extract_bfs_paths` algorithm to the C API to facilitate the creation of python bindings.

PR #1955, opened 2021-11-18
Propose extract_bfs_paths C API

ChuckHastings:
This PR defines the C API for extract_bfs_paths.  Implementation will follow once the API is approved.

Partially addresses #1891

### DIFF
--- a/cpp/include/cugraph_c/algorithms.h
+++ b/cpp/include/cugraph_c/algorithms.h
@@ -153,7 +153,8 @@ typedef struct {
  * @param [in]   result   The result from bfs or sssp
  * @return type erased array of vertex ids
  */
-cugraph_type_erased_device_array_t* cugraph_paths_result_get_vertices(cugraph_paths_result_t* result);
+cugraph_type_erased_device_array_t* cugraph_paths_result_get_vertices(
+  cugraph_paths_result_t* result);
 
 /**
  * @brief     Get the distances from the paths result
@@ -161,7 +162,8 @@ cugraph_type_erased_device_array_t* cugraph_paths_result_get_vertices(cugraph_pa
  * @param [in]   result   The result from bfs or sssp
  * @return type erased array of distances
  */
-cugraph_type_erased_device_array_t* cugraph_paths_result_get_distances(cugraph_paths_result_t* result);
+cugraph_type_erased_device_array_t* cugraph_paths_result_get_distances(
+  cugraph_paths_result_t* result);
 
 /**
  * @brief     Get the predecessors from the paths result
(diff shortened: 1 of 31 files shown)


ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

rlratzel:
rerun tests

rlratzel:
I talked with @ChuckHastings offline, and something we discussed is that these new APIs could be used for general path extraction - not just for extracting paths from BFS results - and should be renamed without `_bfs` in the names (and other changes to decouple from BFS, if possible). For example, a general path extraction API could be immediately useful when we add SSSP to the C API, which returns results in the same format.

cpp/include/cugraph_c/algorithms.h
```diff
@@ -194,20 +197,82 @@ void cugraph_bfs_result_free(cugraph_bfs_result_t* result);
...
+  bool_t compute_predecessors,
+  cugraph_bfs_result_t** result,
+  cugraph_error_t** error);
+
+/**
+ * @brief     Opaque extract_bfs_paths result type
+ */
+typedef struct {
+  int align_;
+} cugraph_extract_bfs_paths_result_t;
```
rlratzel, line 222: Reminder for either a `FIXME` or an update in this PR to address [this](https://github.com/rapidsai/cugraph/pull/1955#issuecomment-974584672).

rlratzel (commented):
Just added a reminder for a `FIXME` we discussed.

Also, this PR could add a unit test too, which calls a stubbed out implementation. I personally would benefit from seeing how the API is meant to be used in the context of an actual BFS call.

rlratzel:
rerun tests

CI failed due to what looks like a problem Jenkins had in accessing github. Rerunning in case this was temporary.

rlratzel (approved):
Thanks, LGTM. I had some small suggestions which need not hold up my approval.

cpp/tests/c_api/extract_paths_test.c
```diff
@@ -0,0 +1,180 @@
...
+  cugraph_error_t* ret_error;
+
+  cugraph_resource_handle_t* p_handle                    = NULL;
+  cugraph_graph_t* p_graph                               = NULL;
+  cugraph_paths_result_t* p_paths_result                 = NULL;
+  cugraph_extract_paths_result_t* p_extract_paths_result = NULL;
+  cugraph_type_erased_device_array_t* p_sources          = NULL;
+  cugraph_type_erased_device_array_t* p_destinations     = NULL;
+
+  p_handle = cugraph_create_handle();
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in latest push

cpp/tests/c_api/extract_paths_test.c
```diff
@@ -0,0 +1,180 @@
...
+  TEST_ASSERT(test_ret_value, p_handle != NULL, "raft handle creation failed.");
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in latest push

BradReesWork:
rerun tests

cpp/include/cugraph_c/algorithms.h
```diff
@@ -136,45 +138,50 @@ cugraph_error_code_t cugraph_personalized_pagerank(
...
 
 /**
- * @brief     Opaque bfs result type
+ * @brief     Opaque paths result type
+ *
+ * Store the output of BFS or SSSP, computing predecessors and distances
+ * from a seed.
  */
 typedef struct {
   int align_;
```
seunghwak, line 147: Should we better use int32_t instead of int if the intention here is to really enforce 4 byte alignment?

seunghwak (approved):
Looks good to me, I have just a minor comment to think about.

cpp/include/cugraph_c/algorithms.h
ChuckHastings: Good point.  Will fix in the implementation PR.

ChuckHastings:
@gpucibot merge

Ethyling (approved):

PR #1971, opened 2021-11-23
Add rmat tests to Louvain C++ unit tests

ChuckHastings:
Testructure louvain tests to use new test configuration, add rmat tests

### DIFF
--- a/cpp/tests/community/louvain_test.cpp
+++ b/cpp/tests/community/louvain_test.cpp
@@ -9,19 +9,15 @@
  *
  */
 #include <utilities/base_fixture.hpp>
+#include <utilities/test_graphs.hpp>
 #include <utilities/test_utilities.hpp>
+
+#include <cugraph/algorithms.hpp>
+#include <cugraph/graph.hpp>
 
 #include <raft/cudart_utils.h>
 #include <raft/handle.hpp>
 #include <rmm/device_uvector.hpp>
-#include <rmm/mr/device/cuda_memory_resource.hpp>
-
-#include <cugraph/graph.hpp>
-
-#include <cugraph/algorithms.hpp>
-
-#include <raft/cudart_utils.h>
-#include <raft/handle.hpp>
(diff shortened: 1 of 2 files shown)


cpp/tests/community/louvain_test.cpp
```diff
@@ -33,28 +29,15 @@
 #include <vector>
 
 struct Louvain_Usecase {
-  std::string graph_file_full_path_{};
   bool test_weighted_{false};
+  bool check_correctness_{false};
   int expected_level_{0};
   float expected_modularity_{0};
```
seunghwak, line 35: Should we better place check_correctness as the last member variable and set the default value to true? So, this way, in every test, we may consistently skip setting the last variable except for `rmat_benchmark_test`? This convention is used at least in multiple algorithms.

seunghwak (approved):
Looks good to me except for a minor complaints.

cpp/tests/community/louvain_test.cpp
ChuckHastings: Louvain is an unusual test because it approximates the solution to this NP-hard problem.  We don't currently have a CPU reference implementation that we compare against.  The way the tester currently works is that I have hand verified a result for the `karate` data that is used for validation.  If check_correctness is true then expected_level and expected_modularity need to be set to the expected results.  If we add results, currently, we would need to compute the expected level and modularity.

I agree that's a good paradigm and try to follow it.  To do that in this case we would need to create a serial CPU implementation of Louvain that mimicked the behavior of our parallel version (the reference serial implementation gets an entirely different answer because it randomizes the order of the vertices for every iteration of the outer loop).  We also perturb the serial order to enable parallelism by only allowing vertices to move up or down in the cluster sequence.  Mimicking that in serial code would require some extra bookkeeping.

ChuckHastings:
rerun tests

BradReesWork:
@gpucibot merge

PR #1973, opened 2021-12-01
Extract paths C API implementation

### LINKED ISSUE #1891: [ENH] Add extract_bfs_paths to C API

ChuckHastings:
C API implementation for extract paths.

Closes #1891


cpp/src/c_api/array.hpp
```diff
@@ -41,9 +41,10 @@ struct cugraph_type_erased_device_array_t {
   }
 
   template <typename T>
-  cugraph_type_erased_device_array_t(rmm::device_uvector<T>&& vec, data_type_id_t type)
-    : data_(vec.release()), size_(vec.size()), type_(type)
+  cugraph_type_erased_device_array_t(rmm::device_uvector<T>& vec, data_type_id_t type)
+    : size_(vec.size()), type_(type)
```
seunghwak, line 45: Isn't it cleaner to just change the order of size_ and data_ in the struct definition? (so : size_(vec.size()), data_(vec.release()) to work?)

seunghwak (approved):
Looks good to me, I have just one minor comment to consider.

cpp/src/c_api/array.hpp
ChuckHastings: Done in latest push.

I had tried changing the order of initialization without changing the order of the data in the struct that did not work.  Changing both was the missing trick.  Thanks!

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #1975, opened 2021-12-01
Replace all references to CHECK_CUDA with RAFT_CHECK_CUDA

ChuckHastings:
RAFT is providing a separate implementation that we want to use.

### DIFF
--- a/cpp/include/cugraph/prims/update_frontier_v_push_if_out_nbr.cuh
+++ b/cpp/include/cugraph/prims/update_frontier_v_push_if_out_nbr.cuh
@@ -957,7 +957,7 @@ void update_frontier_v_push_if_out_nbr(
                           d_offsets.begin());
       std::vector<vertex_t> h_offsets(d_offsets.size());
       raft::update_host(h_offsets.data(), d_offsets.data(), d_offsets.size(), handle.get_stream());
-      CUDA_TRY(cudaStreamSynchronize(handle.get_stream()));
+      RAFT_CHECK_CUDA(cudaStreamSynchronize(handle.get_stream()));
       h_offsets.push_back(matrix_partition_frontier_size);
       // FIXME: we may further improve performance by 1) concurrently running kernels on different
       // segments; 2) individually tuning block sizes for different segments; and 3) adding one more
--- a/cpp/src/centrality/betweenness_centrality.cu
+++ b/cpp/src/centrality/betweenness_centrality.cu
@@ -95,7 +95,7 @@ vertex_t get_total_number_of_sources(raft::handle_t const& handle, vertex_t loca
                                  raft::comms::op_t::SUM,
                                  handle.get_stream());
     total_number_of_sources_used = d_number_of_sources.value(handle.get_stream());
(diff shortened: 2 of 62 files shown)


cjnolet (approved):
LGTM. Haha, I guess I should have looked more closely at where you were using the macro and just made the change for you guys.

seunghwak:
So, we're mostly using CUDA_TRY instead of CUDA_CHECK, and is there any reason we should keep both?

seunghwak:
And maybe... shouldn't we better rename CUDA_TRY to either CUDA_CHECK or rename CUDA_TRY to RAFT_CUDA_TRY?

ChuckHastings:
> And maybe... shouldn't we better rename either CUDA_TRY to CUDA_CHECK or rename CUDA_TRY to RAFT_CUDA_TRY?

I can change the CUDA_TRY references to RAFT_CUDA_CHECK if that's the preferred spelling :-)

I will make that change and push an update.

seunghwak:
> 

I forgot the details, AFAIK, there was some discussion about CUDA_TRY and CUDA_CHECK in the past... I might be completely wrong, but I guess we were using CUDA_CHECK in the past, but we switched to CUDA_TRY when we decided to handle errors using exceptions. Let me dig into the history more.

seunghwak:
@cjnolet @ChuckHastings 

So, in raft, we have both CUDA_TRY, CUDA_CHECK, and CHECK_CUDA (https://github.com/rapidsai/raft/blob/branch-22.02/cpp/include/raft/cudart_utils.h).

For CUBLAS, we have both CUBLAS_CHECK & CUBLAS_TRY. For NCCL, we have only NCCL_TRY. I haven't checked for other CUDA libraries but I guess the situation might be similar.

I guess all these exist due to some backward compatibility reasons, but I think we should first decide between _TRY and _CHECK first and better be consistent.

BradReesWork:
rerun tests

BradReesWork (approved):

cjnolet:
@seunghwak, I think in the end, `CUDA_CHECK` is just calling `CUDA_TRY` in RAFT. I'm okay w/ either name we choose, though I'm thinking we should consider also prefixing the macros for the other library wrappers w/ `RAFT_`. This will avoid future potential conflicts but will also allow us to directly make use of the RAFT logging once that's moved over.

seunghwak:
> @seunghwak, I think in the end, `CUDA_CHECK` is just calling `CUDA_TRY` in RAFT. I'm okay w/ either name we choose, though I'm thinking we should consider also prefixing the macros for the other library wrappers w/ `RAFT_`. This will avoid future potential conflicts but will also allow us to directly make use of the RAFT logging once that's moved over.
> 
> Whichever names we choose, I'll also update RAFT and cuML to use them.

I'm also fine with either choice as long as consistent.

cjnolet:
PR for renaming updates has been merged on the raft side. `RAFT_CUDA_TRY` should be used now when running any cuda functions and `RAFT_CHECK_CUDA` should be used with a stream to check errors (and sync in debug mode).

PR #1980, opened 2021-12-03
Shutdown the connected scheduler and workers

jnke2016:
The MNMG tests/benchmarks runs rely on killing the dask cluster process ID to stop all python/dask processes
This PR uses `client.shutdown` to shutdown the connected scheduler and workers


rlratzel (approved):

rlratzel:
@gpucibot merge

PR #1989, opened 2021-12-10
make C++ tests run faster (fewer tests)

### LINKED ISSUE #1555: [FEA] Reduce the testing done by CI
The amount of testing we currently do in CI is causing timeout issues.  We need to reevaluate what we are testing as part of CI to reduce the total time.

As an example, I ran `make test` on my linux box and saw the following two results appear:
```
      Start  2: BETWEENNESS_TEST
 2/38 Test  #2: BETWEENNESS_TEST ...........................   Passed  843.43 sec
      Start  3: EDGE_BETWEENNESS_TEST
 3/38 Test  #3: EDGE_BETWEENNESS_TEST ......................   Passed  409.24 sec
```
It seems like 20 minutes spent running these two C++ unit tests is really too high.  The point of the unit tests is to verify the functionality of the code, not to measure performance (that should be a different suite of tests).

ChuckHastings:
Modified the most expensive C++ tests to run fewer tests.  Closes #1555 

Added an option (like in the rmat tests) to run a specific test file if the developer wants to manually run larger tests.  For example:

```tests/COARSEN_GRAPH_TEST --gtest_filter=file_ben* --test_file_name=test/datasets/ljournal-2008.mtx```

The smaller graphs *should* be small enough to test things.  Once we add C++ code coverage we should be able to verify this.

On my local workstation this reduced the time spent executing the C++ tests by about 25 minutes.

### DIFF
--- a/cpp/tests/link_analysis/hits_test.cpp
+++ b/cpp/tests/link_analysis/hits_test.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021, NVIDIA CORPORATION.
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/cpp/tests/link_analysis/hits_test.cpp
+++ b/cpp/tests/link_analysis/hits_test.cpp
@@ -144,7 +144,7 @@ class Tests_Hits : public ::testing::TestWithParam<std::tuple<Hits_Usecase, inpu
   template <typename vertex_t, typename edge_t, typename weight_t>
   void run_current_test(std::tuple<Hits_Usecase const&, input_usecase_t const&> const& param)
   {
-    constexpr bool renumber = true;
+    constexpr bool renumber            = true;
     auto [hits_usecase, input_usecase] = param;
 
     // 1. initialize handle
--- a/cpp/tests/link_analysis/hits_test.cpp
+++ b/cpp/tests/link_analysis/hits_test.cpp
@@ -144,13 +144,8 @@ class Tests_Hits : public ::testing::TestWithParam<std::tuple<Hits_Usecase, inpu
   template <typename vertex_t, typename edge_t, typename weight_t>
   void run_current_test(std::tuple<Hits_Usecase const&, input_usecase_t const&> const& param)
   {
-    run_current_test<vertex_t, edge_t, weight_t>(std::get<0>(param), std::get<1>(param));
-  }
-
-  template <typename vertex_t, typename edge_t, typename weight_t>
-  void run_current_test(Hits_Usecase const& hits_usecase, input_usecase_t const& input_usecase)
-  {
     constexpr bool renumber = true;
+    auto [hits_usecase, input_usecase] = param;
 
     // 1. initialize handle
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Review by kaatish — APPROVED** [review]
lgtm

File: cpp/tests/link_analysis/hits_test.cpp [hunk 1]
```diff
@@ -141,6 +141,12 @@ class Tests_Hits : public ::testing::TestWithParam<std::tuple<Hits_Usecase, inpu
   virtual void SetUp() {}
   virtual void TearDown() {}
 
+  template <typename vertex_t, typename edge_t, typename weight_t>
+  void run_current_test(std::tuple<Hits_Usecase const&, input_usecase_t const&> const& param)
+  {
+    run_current_test<vertex_t, edge_t, weight_t>(std::get<0>(param), std::get<1>(param));
+  }
+
```
seunghwak, line 149: What's the intention of creating this version of `run_current_test`? If we don't need to keep both, shouldn't we better remove one?

cpp/tests/structure/coarsen_graph_test.cpp
```diff
@@ -238,6 +238,14 @@ class Tests_CoarsenGraph
   virtual void SetUp() {}
   virtual void TearDown() {}
 
+  template <typename vertex_t, typename edge_t, typename weight_t, bool store_transposed>
+  void run_current_test(
+    std::tuple<CoarsenGraph_Usecase const&, input_usecase_t const&> const& param)
+  {
+    run_current_test<vertex_t, edge_t, weight_t, store_transposed>(std::get<0>(param),
+                                                                   std::get<1>(param));
+  }
```
seunghwak, line 247: As I mentioned above, should we keep both versions?

cpp/tests/structure/graph_test.cpp
```diff
@@ -143,99 +147,254 @@ class Tests_Graph : public ::testing::TestWithParam<Graph_Usecase> {
...
-    }
+      if (h_cugraph_weights) {
+        raft::update_host((*h_cugraph_weights).data(),
+                          *(graph_view.get_matrix_partition_view().get_weights()),
+                          graph_view.get_number_of_edges(),
+                          handle.get_stream());
+      }
 
-    CUDA_TRY(cudaStreamSynchronize(handle.get_stream()));
+      CUDA_TRY(cudaStreamSynchronize(handle.get_stream()));
```
seunghwak, line 172: I assume now we should better use `handle.sync_stream()` with the recent raft update.

cpp/tests/link_analysis/hits_test.cpp
ChuckHastings: The new signature (`std::tuple<...>` instead of sending in the two input arguments) makes things a little cleaner.

I have only applied it to the more expensive tests as part of this PR.  I can certainly combine the two `run_current_test` calls in these tests, it would be a little cleaner.

cpp/tests/structure/graph_test.cpp
ChuckHastings: Will look at this.  raft changes were not merged when I started working on this.

ChuckHastings: We have a number of calls like this in the code base, I'm inclined to address this in a separate PR that picks up all of them.

cpp/tests/link_analysis/hits_test.cpp
ChuckHastings: Fixed in latest push.

cpp/tests/structure/coarsen_graph_test.cpp
ChuckHastings: Fixed in latest push.

ChuckHastings:
rerun tests

seunghwak (approved):
LGTM

ChuckHastings:
@gpucibot merge

PR #1992, opened 2021-12-16
fix bug when calculating the number of vertices

jnke2016:
The number of vertices calculation in the previous implementation relied on getting the max of the source and destination's max. Each worker computes the max of its partition however, workers having empty partition attempt to do the same causing an error. This seems to be caused by a recent change on how dask handle empty partitions with the `max()` function. 

This PR calculates the total number of vertices by summing each partition's number of vertices which is inspired from our code base.


rlratzel:
Build timed out for some reason.

rerun tests

rlratzel (approved):

rlratzel:
@gpucibot merge

rlratzel:
@gpucibot merge

PR #1996, opened 2021-12-21
[Review] Add optimized 2x string column renumbering code

chirayuG-nvidia:


chirayuG-nvidia:
CC: @rlratzel  @ChuckHastings

cpp/libcugraph_etl/src/renumbering.cu
```diff
@@ -16,9 +16,954 @@
...
+
+    rmm::device_uvector<accum_type> atomic_agg(32, exec_strm);  // just padded to 32
+    CHECK_CUDA(cudaMemsetAsync(atomic_agg.data(), 0, sizeof(accum_type), exec_strm));
+
+    auto cuda_map_obj = cudf_map_type::create(std::max(
+                                                  static_cast<size_t>(
+                                                    static_cast<double>(num_rows) / load_factor),
+                                                    (size_t)num_rows + 1),
+                                              exec_strm,
+                                              str_hash_value{
```
ChuckHastings, line 743: Your default constructor already uses these limits, you should just be able to use `str_hash_value{}` here.

cpp/libcugraph_etl/src/renumbering.cu
```diff
@@ -16,9 +16,954 @@
...
+                                                std::numeric_limits<size_type>::max(),
+                                                std::numeric_limits<accum_type>::max(),
+                                                std::numeric_limits<int32_t>::max() }
+                                                ).release();
+    dim3 block(512, 1, 1);
+    dim3 grid((num_rows -  1)/ block.x + 1, 1, 1);
+
+    int32_t num_multiprocessors = 80; // get from cuda properties
+
+    std::cout << "hash map size: " << cuda_map_obj->capacity()  << " rows_: "<< num_rows << std::endl;
```
ChuckHastings, line 753: We don't want print statements like this in the final code.

cpp/libcugraph_etl/src/renumbering.cu
```diff
@@ -16,9 +16,954 @@
...
+    cudaOccupancyMaxActiveBlocksPerMultiprocessor(&num_blocks, assign_histogram_idx, block.x, 0);
+    grid.x = num_multiprocessors * num_blocks;
+    assign_histogram_idx<<<grid, block, 0, exec_strm>>>(*cuda_map_obj, cuda_map_obj->capacity(),
+                                                  sort_key.data(), sort_value.data(),
+                                                  atomic_idx.data());
+
+    // can release original histogram memory here
+
+    // cub doesnt have custom comparator sort
+    // new cub release will have sort with custom comparator - @ToDo
```
ChuckHastings, line 805: cugraph convention is to label this with the prefix FIXME:

```
// FIXME: new cub release ...
```

cpp/libcugraph_etl/src/renumbering.cu
```diff
@@ -16,9 +16,954 @@
...
+    }
+
+    for (int i = 0; i < dst_view.num_columns(); i++) {
+      auto str_col_view = cudf::strings_column_view(dst_view.column(i));
+      dst_vertex_chars_ptrs.push_back(const_cast<char_type*>(str_col_view.chars().data<char_type>()));
+      dst_vertex_offset_ptrs.push_back(const_cast<str_offset_type*>(
+                                            str_col_view.offsets().data<str_offset_type>()));
+    }
+
+    accum_type *hist_insert_counter = nullptr;
```
ChuckHastings, line 728: Can we make this an std::vector?  The RAAI construct is generally cleaner and easier to maintain.

The cugraph code uses raft utilities to support easy transfers between host and device.  https://github.com/rapidsai/raft/blob/6a8c7a3bebe85d8fef34951e6f09c93fa733b06f/cpp/include/raft/cudart_utils.h#L251

ChuckHastings: In order to efficiently use raft utilities, we *might* need to add the `raft::handle_t` as a parameter to the renumber_cudf_tables method.

ChuckHastings (commented):
This is a great start!

I have a bunch of comments, largely related to making this a bit more consistent with other cugraph software which will ultimately make this easier to maintain.

ChuckHastings:
A couple of other notes...
 1. The copyright headers will need to be updated to include 2022 for any files that are new or modified.
 2. In order to pass CI you will need to format the code using clang-format.  When you install the cugraph condo environment you should get clang-format installed into your conda environment.  From the top level directory (cugraph) you should be able to run: `python cpp/scripts/run-clang-format.py -inplace` and it will modify any files that don't match our clang-format rules.

cpp/libcugraph_etl/src/renumbering.cu
chirayuG-nvidia: Changing this to a raft::mr::host::buffer for RAII. I directly write some values to pinned memory so can't use a std::vector.

cpp/libcugraph_etl/include/hash/hash_constants.hpp
```diff
@@ -0,0 +1,64 @@
+/*
```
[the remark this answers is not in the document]
chirayuG-nvidia: removed.

cpp/libcugraph_etl/include/hash/hashing.cu
```diff
@@ -0,0 +1,138 @@
+/*
```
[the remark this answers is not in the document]
chirayuG-nvidia: removed.

cpp/libcugraph_etl/src/renumbering.cu
```diff
@@ -16,9 +16,954 @@
...
 
 namespace cugraph {
 namespace etl {
+using size_type = cudf::size_type;  // size_type is int32
+using accum_type = uint32_t;
+
+constexpr uint32_t hash_inc_constant = 9999;
+
+typedef struct str_hash_value{
+  __host__ __device__ str_hash_value() {
```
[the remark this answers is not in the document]
chirayuG-nvidia: Switched to initialization, however empty default constructor still required otherwise thrust::sort is complaining.

cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
```diff
@@ -0,0 +1,545 @@
+/*
```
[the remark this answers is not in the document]
chirayuG-nvidia: Added comment.

chirayuG-nvidia:
@ChuckHastings all the suggested edits are done. Please take a look.

BradReesWork:
@gpucibot merge

ChuckHastings (approved):

PR #2009, opened 2022-01-11
Removing faiss from build

cjnolet:


cjnolet:
rerun tests

cjnolet:
I'm able to build this locally without problem, even without having FAISS in my conda environment. Trying to figure out what's different

PR #2017, opened 2022-01-19
Removed unused CI files

rlratzel:
After an internal discussion, it was determined the `Dockerfile` has not been in use since RAPIDS `devel` and `runtime` Docker images have been released.

The `ci/local` dir is also not working as originally intended (it was originally created to allow devs to reproduce CI failures locally) because the script does not support Project Flash. Since the script cannot be easily updated for Project Flash, it'll just be removed.


ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

PR #2023, opened 2022-01-20
Python API updates to enable explicit control of internal `graph_t` creation and deletion

rlratzel:
Python API updates to enable explicit control of internal `graph_t` creation and deletion. These changes will allow users who need more control over how the internal `graph_t` object is managed to have additional APIs to do so.  Current cugraph algos construct the appropriate `graph_t` at the C++ boundary, run the algo, and destroy the object each time.  This is more convenient and safer in some cases since it does not require the user to understand the necessary data format requirements for each algo (CSC, CSR, etc.), but it adds overhead to each algo call. If a user knows which algos will be called ahead of time and the relevant graph creation options to use, they can reuse the underlying `graph_t` object and eliminate the redundant creation/deletion expense.  These APIs also allow for benchmarks to measure only the algo run time without the initial graph creation time.

These changes are grouped into a category of enhancements sometimes referred to as "expert mode"

Changes here include:
* Updates to `pylibcugraph` to add the APIs from the `libcugraph_c` library for `graph_t` management and calling Pagerank and SSSP
* Addition of the `experimental` namespace to `pylibcugraph`, which is where some of the new APIs will be until we finalize decisions on names, signatures, etc.
* <s>Updates to cugraph Graph classes, Pagerank, and SSSP algos to call in to `pylibcugraph`</s>


python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -0,0 +1,169 @@
...
+        of the ith edge of the graph.
+
+    store_transposed : bool
+        Set to True if the graph should be transposed. This is required for some
+        algorithms, such as pagerank.
+
+    renumber : bool
+        Set to True to indicate the vertices used in src_array and dst_array are
+        not appropriate for use as internal array indices, and should be mapped
+        to continuous integrers starting from 0.
```
ChuckHastings, line 90: integrers => integers

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -0,0 +1,169 @@
...
+                            f"{type(store_transposed)}")
+        if not(isinstance(renumber, (int, bool))):
+            raise TypeError("expected int or bool for renumber, got "
+                            f"{type(renumber)}")
+        if not(isinstance(expensive_check, (int, bool))):
+            raise TypeError("expected int or bool for expensive_check, got "
+                            f"{type(expensive_check)}")
+        assert_CAI_type(src_array, "src_array")
+        assert_CAI_type(dst_array, "dst_array")
+        assert_CAI_type(weight_array, "weight_array")
```
ChuckHastings, line 118: Should there be an assertion that src_array and dst_array have the same type?  FIXME would be fine for now.

python/pylibcugraph/pylibcugraph/pagerank.pyx
```diff
@@ -0,0 +1,256 @@
...
+        import numpy
+    except ModuleNotFoundError:
+        raise RuntimeError("pagerank requires the numpy package, which could "
+                           "not be imported")
+
+    if has_initial_guess is True:
+        raise ValueError("has_initial_guess must be False for the current "
+                         "release.")
+
+    assert_CAI_type(precomputed_vertex_out_weight_sums,
```
ChuckHastings, line 167: I *believe* that the type of this must match the weight type of the graph.  Probably worth asserting that here (eventually).

python/pylibcugraph/pylibcugraph/sssp.pyx
```diff
@@ -0,0 +1,249 @@
...
+
+    error_code = cugraph_sssp(c_resource_handle_ptr,
+                              c_graph_ptr,
+                              source,
+                              cutoff,
+                              compute_predecessors,
+                              do_expensive_check,
+                              &result_ptr,
+                              &error_ptr)
+    assert_success(error_code, error_ptr, "cugraph_sssp")
```
ChuckHastings, line 168: Something to address:  If `cugraph_sssp` (or any of the C API functions) returns an error then `error_ptr` will have a value.  That will need to be freed using `cugraph_error_free`.

ChuckHastings (approved):
A few minor observations.  I think it's fine to go forward, but we probably should address these early in 22.04

cpp/src/c_api/graph_sg.cpp
```diff
@@ -173,26 +175,28 @@ struct destroy_graph_functor : public abstract_functor {
...
+  const cugraph_type_erased_device_array_view_t* weights,
   bool_t store_transposed,
   bool_t renumber,
   bool_t check,
   cugraph_graph_t** graph,
   cugraph_error_t** error)
 {
   constexpr bool multi_gpu = false;
-  constexpr size_t int32_threshold{2 ^ 31 - 1};
+  constexpr size_t int32_threshold{std::numeric_limits<int32_t>::max()};
```
seunghwak, line 188: Why int32 here? Are we assuming 32 bit vertex IDs? (for SG) This might be OK for now (may be not if managed memory) but doesn't sound future-proof.

cpp/tests/c_api/bfs_test.c
```diff
@@ -54,15 +55,17 @@ int generic_bfs_test(vertex_t* h_src,
     p_handle, h_src, h_dst, h_wgt, num_edges, store_transposed, &p_graph, &ret_error);
 
   ret_code =
-    cugraph_type_erased_device_array_create(p_handle, INT32, num_seeds, &p_sources, &ret_error);
+    cugraph_type_erased_device_array_create(p_handle, num_seeds, INT32, &p_sources, &ret_error);
```
seunghwak, line 58: So, we're using (vertex/edge/weight)_tid in the create_graph_test.c, but why are we using INT32 here? Shouldn't we better take the same approach for flexibility & consistency?

python/pylibcugraph/pylibcugraph/_cugraph_c/graph.pxd
```diff
@@ -0,0 +1,80 @@
...
+
+    cdef cugraph_error_code_t \
+        cugraph_mg_graph_create(
+            const cugraph_resource_handle_t* handle,
+            const cugraph_graph_properties_t* properties,
+            const cugraph_type_erased_device_array_view_t* src,
+            const cugraph_type_erased_device_array_view_t* dst,
+            const cugraph_type_erased_device_array_view_t* weights,
+            const cugraph_type_erased_host_array_view_t* vertex_partition_offsets,
+            const cugraph_type_erased_host_array_view_t* segment_offsets,
```
seunghwak, line 67: Better remove `vertex_partition_offsets` and `segment_offsets`. These are unnecessary if we use `create_graph_from_edgelist` and this involves too much implementation details.

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -0,0 +1,169 @@
...
+    """
+    def __cinit__(self,
+                  EXPERIMENTAL__ResourceHandle resource_handle,
+                  EXPERIMENTAL__GraphProperties graph_properties,
+                  src_array,
+                  dst_array,
+                  weight_array,
+                  store_transposed,
+                  renumber,
+                  expensive_check):
```
seunghwak, line 104: Naming mismatch. `do_expensive_check` in the above documentation. `expensive_check` here.

python/pylibcugraph/pylibcugraph/pagerank.pyx
```diff
@@ -0,0 +1,256 @@
...
+
+    Parameters
+    ----------
+    resource_handle : ResourceHandle
+        Handle to the underlying device resources needed for referencing data
+        and running algorithms.
+
+    graph : SGGraph
+        The input graph. The graph must be created with the store_transposed
+        option set to True.
```
seunghwak, line 83: So, is this what we're planning to do?

I thought our plan was: if the input graph is created with store_transposed = false, we transpose the graph inside (so normal users do not need to explicitly think about this, but for performance critical use cases, users can create a graph with the expected format to avoid this conversion overhead).

python/pylibcugraph/pylibcugraph/tests/conftest.py
```diff
@@ -11,15 +11,196 @@
...
+    resource_handle = ResourceHandle()
+    graph_props = GraphProperties(is_symmetric=False, is_multigraph=False)
+
+    g = SGGraph(resource_handle,
+                graph_props,
+                device_srcs,
+                device_dsts,
+                device_weights,
+                store_transposed=transposed,
+                renumber=False,
```
seunghwak, line 131: So, are we disabling renumbering for SG? This has negative performance implications.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -26,25 +26,27 @@
...
                        cugraph_graph_properties_t const* properties,
-                       c_api::cugraph_type_erased_device_array_t const* src,
-                       c_api::cugraph_type_erased_device_array_t const* dst,
-                       c_api::cugraph_type_erased_device_array_t const* weights,
+                       c_api::cugraph_type_erased_device_array_view_t const* src,
+                       c_api::cugraph_type_erased_device_array_view_t const* dst,
+                       c_api::cugraph_type_erased_device_array_view_t const* weights,
                        bool_t renumber,
                        bool_t check,
                        data_type_id_t edge_type)
```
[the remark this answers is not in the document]
ChuckHastings: Vertex_type is embedded in the type erased array.  Edge type cannot be deduced from the other parameters.

python/pylibcugraph/pylibcugraph/graphs.pyx
rlratzel: fixed in next commit.

python/pylibcugraph/pylibcugraph/graphs.pyx
rlratzel: added FIXME in next commit.

python/pylibcugraph/pylibcugraph/pagerank.pyx
rlratzel: added FIXME in next commit.

python/pylibcugraph/pylibcugraph/sssp.pyx
rlratzel: added FIXME to the util function that handles those in next commit.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -19,11 +19,11 @@
...
-  const cugraph_type_erased_device_array_t* src,
-  const cugraph_type_erased_device_array_t* dst,
-  const cugraph_type_erased_device_array_t* weights,
-  const cugraph_type_erased_host_array_t* vertex_partition_offsets,
-  const cugraph_type_erased_host_array_t* segment_offsets,
+  const cugraph_type_erased_device_array_view_t* src,
+  const cugraph_type_erased_device_array_view_t* dst,
+  const cugraph_type_erased_device_array_view_t* weights,
+  const cugraph_type_erased_host_array_view_t* vertex_partition_offsets,
+  const cugraph_type_erased_host_array_view_t* segment_offsets,
```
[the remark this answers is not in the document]
ChuckHastings: Yes, update from my PR.  @rlratzel, can you add a FIXME with this into the code.  I will address this in 22.04 when we actually implement the MG version.

python/pylibcugraph/pylibcugraph/graphs.pyx
rlratzel: fixed in next commit.

cpp/src/c_api/graph_sg.cpp
ChuckHastings: This threshold is identifying whether to create a 32 or 64-bit edge_t.  If the number of edges is less than the threshold then we will use edge_t as int32_t, otherwise we will use int64_t.  The constant is used further down in the code to make this choice.

This does assume an int32_t vertex_t.  I suppose we could have the case where the vertex_t is int64_t but the number of edges is less than 2^31-1.  The if statement below could be modified to catch this case.  @rlratzel can we add a FIXME for this?

python/pylibcugraph/pylibcugraph/pagerank.pyx
rlratzel: Yes, the docstring in place was from an older API policy decision we decided not to go with for now. I'll remove that to match the current implementation in the next commit.

cpp/tests/c_api/bfs_test.c
ChuckHastings: It would be better to be consistent.  Of course in `create_graph_test.c` we're defining variables but then hard-coding them to the constant INT32.  Better would be to pass the types into the functions in both cases so that the test cases could be parameterized in the main.

These unit tests are relatively new and our standards for the C API testing are not well established.

python/pylibcugraph/pylibcugraph/tests/conftest.py
rlratzel: It's not disabled in the API; the code here is just for creating graphs for tests.  I'll add a FIXME to add coverage for both renumbering enabled and disabled in the next commit.

python/pylibcugraph/pylibcugraph/_cugraph_c/graph.pxd
ChuckHastings: Suggest a FIXME for this.  The MG version will be worked on next release and we'll figure out what changes are required here.

cpp/src/c_api/graph_mg.cpp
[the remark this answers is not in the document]
rlratzel: added FIXME in next commit.

cpp/src/c_api/graph_sg.cpp
rlratzel: added FIXME in the next commit.

cpp/tests/c_api/bfs_test.c
rlratzel: added FIXME in the next commit.

python/pylibcugraph/pylibcugraph/_cugraph_c/graph.pxd
rlratzel: added a FIXME in the next commit.

seunghwak (approved):
Approving (assuming FIXMEs will be addressed quickly in the next release).

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #2030, opened 2022-01-22
Mods for pylibcugraph

ChuckHastings:
Changes to support #2023
* Adds view semantic to the cugraph_type_erased_device_array
* Add device to device copy

Should not be merged until #2023 has been tested with it.

UPDATE:  The commits from this PR were merged into #2023.  So once 2023 is merged then all of these changes will be in the baseline.  I will close this PR without merging once 2023 is merged.


cpp/include/cugraph_c/array.h
```diff
@@ -55,44 +63,87 @@ cugraph_error_code_t cugraph_type_erased_device_array_create(
  */
 void cugraph_type_erased_device_array_free(cugraph_type_erased_device_array_t* p);
 
+#if 0
```
seunghwak, line 66: Better delete if no longer necessary?

cpp/include/cugraph_c/array.h
```diff
@@ -103,43 +154,84 @@ cugraph_error_code_t cugraph_type_erased_host_array_create(const cugraph_resourc
  */
 void cugraph_type_erased_host_array_free(cugraph_type_erased_host_array_t* p);
 
+#if 0
```
seunghwak, line 157: Better delete if no longer necessary?

seunghwak (approved):
LGTM

cpp/include/cugraph_c/array.h
ChuckHastings: Added a FIXME in the latest push.  This was going to be part of the API, but it can't currently be implemented.  We need to have a discussion about whether we can implement this or whether we should address this a different way.

cpp/include/cugraph_c/array.h
ChuckHastings: See above comment

BradReesWork:
fixes rolled into #2023

PR #2032, opened 2022-01-24
Fix raft git ref

Ethyling:

PR #2033, opened 2022-01-24
Using RAFT main branch

rlratzel:
Updated the RAFT cmake file to use the main branch again.

This change is similar to https://github.com/rapidsai/cuml/pull/4508 

Note: did NOT test locally yet.


rlratzel:
duplicate of https://github.com/rapidsai/cugraph/pull/2032

PR #2038, opened 2022-01-26
Improve MG PageRank scalability

seunghwak:
Improve MG PageRank performance & scalability in multi-node many GPU systems


seunghwak:
rerun tests

cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh
```diff
@@ -513,23 +516,98 @@ void copy_v_transform_reduce_nbr(raft::handle_t const& handle,
...
+      major_tmp_buffer_sizes[i] = GraphViewType::is_adj_matrix_transposed
+                                    ? graph_view.get_number_of_local_adj_matrix_partition_cols(i)
+                                    : graph_view.get_number_of_local_adj_matrix_partition_rows(i);
+    }
+    if (stream_pool_indices) {
+      auto num_concurrent_loops = (*stream_pool_indices).size() / max_segments;
+      major_tmp_buffers.reserve(num_concurrent_loops);
+      for (size_t i = 0; i < num_concurrent_loops; ++i) {
+        size_t max_size{0};
+        for (size_t j = i; j < graph_view.get_number_of_local_adj_matrix_partitions();
```
aschaffer, line 581: Why not do a `max_element()` here  (lines 581-584)? It's more readable and possibly faster. I understand you have a stride, but you can pass a sequence and calculate the stride inside a lambda comparer, etc.

aschaffer (changes requested):

cpp/include/cugraph/prims/copy_v_transform_reduce_in_out_nbr.cuh
seunghwak: How can I pass a sequence?

Can we do this without using thrust/boost (i.e. without using counting_iterator/transform_iterator).

We can create an additional vector storing a sequence, but then, I am not sure the code will be more readable.

Could you show me the code?

aschaffer: You can use counting_iterator/transform_iterator on std::vector<> with thrust::host policy and, say, thrust::reduce() or thrust::maximum_element() but the counter arithmetic can be messy. You're right the resulting code would probably _NOT_ be any more readable.

aschaffer (approved):

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2044, opened 2022-01-27
Improve MG graph creation

seunghwak:
Improve multi-node multi-GPU scalability of graph creation (especially the code computing renumber_map)
Fix an overflow bug when creating a graph with more than 2^31 vertices


seunghwak:
rerun tests

ChuckHastings (approved):

cpp/include/cugraph/utilities/shuffle_comm.cuh
```diff
@@ -36,6 +37,21 @@ namespace cugraph {
...
+  GroupIdIterator group_id_first{};
+  GroupIdIterator group_id_last{};
+
+  __device__ thrust::tuple<int, size_t> operator()(size_t i) const
+  {
+    auto lower_it =
+      thrust::lower_bound(thrust::seq, group_id_first, group_id_last, static_cast<int>(i));
+    auto upper_it = thrust::upper_bound(thrust::seq, lower_it, group_id_last, static_cast<int>(i));
+    return thrust::make_tuple(static_cast<int>(i),
+                              static_cast<size_t>(thrust::distance(lower_it, upper_it)));
```
kaatish, line 51: Instead of casting `i` to `int` can you cast it to
`typename std::iterator_traits<GroupIdIterator>::value_type` while you are using it for binary_search?
It could be cast to `int` while returning in the tuple.

kaatish (changes requested):

cpp/include/cugraph/utilities/shuffle_comm.cuh
seunghwak: Yeah...

Actually, `typename std::iterator_traits<GroupIdIterator>::value_type` should be `int` here (Group IDs or GPU IDs or process IDs are all assumed to be `int`). Using `GroupIdIterator` intead of `int` type pointer because I am using thrust::transform_iterator to compute group ID on the fly.

I'll add

`static_assert(std::is_same_v(typename std::iterator_traits<GroupIdIterator>::value_type, int));` to make this clearer.

And I may modify `operator()(size_t i)` to `operator()(int i)` to avoid type casting... but I have mixed thoughts about this (I assume `thrust::tabulate` computes the ranges of `i` from thrust::distance() and `i` can potentially overflow `int`. Shouldn't happen unless we have more than 2^31-1 GPUs and that's the assumption for using 'int' here for GPU IDs. I guess explicitly casting this makes this intention more clearer than relying on implicit casting).

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2046, opened 2022-01-27
Dgl graph store

BradReesWork:
new GraphStore for DGL intergration
Code will need to be added to DGL


python/cugraph/cugraph/gnn/graph_store.py
```diff
@@ -0,0 +1,247 @@
...
+    @property
+    def gdata(self):
+        return self.__G
+
+    def __init__(self, graph=None, cluster=None):
+        if graph is not None:
+            if isinstance(graph, PropertyGraph):
+                self.__G = graph
+            else:
+                raise ValueError("graph must be a PropertyGraph")
```
rlratzel, line 50: It looks like this is the only place a caller can set the input graph, so maybe it shouldn't be optional?
```suggestion
    def __init__(self, graph, cluster=None):
        if isinstance(graph, PropertyGraph):
            self.__G = graph
        else:
            raise ValueError("graph must be a PropertyGraph")
```

python/cugraph/cugraph/gnn/graph_store.py
```diff
@@ -0,0 +1,247 @@
...
+
+        self.__client = cluster
+
+    ######################################
+    # Utilities
+    ######################################
+    @property
+    def num_vertices(self):
+        if self.__G is None:
+            raise ValueError("graph has not been set")
```
rlratzel, line 60: if the above suggestion for the ctor is done, this and all others should be able to be safely removed.

python/cugraph/cugraph/gnn/graph_store.py
```diff
@@ -0,0 +1,247 @@
...
+    then adds functions to match the DGL GraphStorage API.
+    However, this class only return cuGraph types.  Conversion to other types,
+    like DGLGraph, needs to be done in DGL or other external packages.
+
+    homogeneous graphs, graphs with no attributes - use Property Graph
+    hetrogeneous graphs - use PropertyGraph
+    """
+
+    __G = None
+    __clinet = None
```
rlratzel, line 31: I think best practice is to only use class variable for constants and set all the instance variables in `__init__`.

python/cugraph/cugraph/gnn/graph_store.py
BradReesWork: dropping checks

python/cugraph/cugraph/gnn/graph_store.py
BradReesWork: updated

rlratzel (approved):

BradReesWork:
@gpucibot merge

wangxiaoyunNV:
Hi Brad, 
When I review the graphStorage class, I noticed that here is no output_device parameter in the sampling functions. But the DGL graphstorage sampling APIs have output_device parameter. Andrei's code has that as well. Could wee add that parameter to the samplings functions? or have another function called to_device in the GraphStorage class? 
https://github.com/dmlc/dgl/blob/701b4fccc2eed979ae3db801fabb6bf7bc03940c/examples/pytorch/__temporary__/dglnew/graph/graph.py

Thank you,
Xiaoyun

wangxiaoyunNV (commented):
I have a concern about CuGraphStore class, it should have a to_device function or parameter in the sampling functions. 

Thank you!

PR #2050, opened 2022-01-30
Updates to enable NumberMap to generate unique src/dst column names

rlratzel:
* Updates to enable `NumberMap` to generate unique src/dst column names instead of assuming the names `src` and `dst`. This is needed since cudf no longer allows for duplicate column names, and some user input may contain "src" and "dst" as pre-existing column names intended for other purposes.
* This commit also cleans up various tech debt items (exception types, docstrings, etc.).

Tested by ensuring the existing python unit tests for SG and MG (2 GPUs) passed, and also ran all notebooks as tests. New tests to emphasize support for specific column names that previously failed with the latest cudf were also added.


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #2054, opened 2022-02-01
Update SG and MG Louvain tests to support Rmat and benchmark tests

ChuckHastings:
Some general cleanup of C++ Louvain tests.
* Updated to use the latest structure/override features
* Added Rmat tests for the MG tests

### DIFF
--- a/cpp/tests/community/louvain_test.cpp
+++ b/cpp/tests/community/louvain_test.cpp
@@ -29,7 +29,8 @@
 #include <vector>
 
 struct Louvain_Usecase {
-  bool test_weighted_{false};
+  size_t max_level_{100};
+  double resolution_{1};
   bool check_correctness_{false};
   int expected_level_{0};
   float expected_modularity_{0};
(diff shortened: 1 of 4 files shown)


cpp/tests/community/louvain_test.cpp
```diff
@@ -29,7 +29,8 @@
 #include <vector>
 
 struct Louvain_Usecase {
-  bool test_weighted_{false};
```
seunghwak, line 32: So, are we always testing for weighted graphs only?

seunghwak (approved):
Looks good to me.

cpp/tests/community/louvain_test.cpp
ChuckHastings: Yes.  Louvain requires a weighted graph.  All of the computations operate on the edge weights.

I'll add a `CUGRAPH_EXPECTS` in the algorithm implementation to check this, since we don't currently check that.

seunghwak: Can't we implicitly assume edge weights of 1 if the input graph is unweighted? Or no real use case for this?

seunghwak: I assume finding communities in unweighted graphs to be something practical.

BradReesWork:
@gpucibot merge

cpp/tests/community/louvain_test.cpp
ChuckHastings: The python layer will add a weight of 1 for edges if the input graph is unweighted.

I think it would make the C++ code unwieldy to allow the input graph to be unweighted.  Once we coarsen the graph it must be weighted, so we'd end up having to have different code paths for the original graph than from the coarsened graphs.

seunghwak: Gotcha, just wanted to point out that there is memory requirement vs implementation complexity trade-off.

PR #2057, opened 2022-02-02
Add CMake `install` rules for tests

ajschmidt8:
This PR adds a CMake `install` rule for test targets. This step is a prerequisite to being able to package tests in their own `conda` package, which will enable us to deprecate _Project Flash_.

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -136,11 +136,12 @@ function(ConfigureTest CMAKE_TEST_NAME)
 
     set_target_properties(
         ${CMAKE_TEST_NAME}
-            PROPERTIES INSTALL_RPATH "\$ORIGIN/../lib")
+            PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
 
     install(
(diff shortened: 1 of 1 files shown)


cpp/tests/CMakeLists.txt
```diff
@@ -133,6 +133,15 @@ function(ConfigureTest CMAKE_TEST_NAME)
     endif(OpenMP_CXX_FOUND)
```
robertmaynard, line 133: The following targets need to be installed as well:
- targets made via ConfigureTestMG()
- targets made via ConfigureCTest()

robertmaynard (changes requested):

cpp/tests/CMakeLists.txt
ajschmidt8: @robertmaynard, unless I'm mistaken, I think I've already covered those targets here. I just rebased on the latest changes and pushed a commit to update the `DESTINATION` and `RPATH` path for each target.

robertmaynard (approved):

rlratzel (approved):

PR #2059, opened 2022-02-05
Add node2vec to C API - API PR

ChuckHastings:
This adds a new node2vec method to the C API.

Implementation will follow in a subsequent PR.


rlratzel (approved):
We had some discussion offline about the new `cugraph_random_walk_result_t` type vs. using the existing "paths type" used by BFS/SSSP, but decided a bigger discussion may be needed and we can refactor if the type changes later.

BradReesWork:
@gpucibot merge

Issue #2003, opened 2022-01-05
[ENH] Improve graph primitives API for better usability


seunghwak:
**Describe the solution you'd like**
Thank you very much @kaatish for providing various inputs. Additional suggestions will be appreciated.

Graph primitives includes classes/functions

under
https://github.com/rapidsai/cugraph/tree/branch-22.02/cpp/include/cugraph/prims

and in

https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/graph.hpp
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/graph_view.hpp
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/graph_functions.hpp
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/vertex_partition_view.hpp
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/vertex_partition_device_view.cuh
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/matrix_partition_view.hpp
https://github.com/rapidsai/cugraph/blob/branch-22.02/cpp/include/cugraph/matrix_partition_device_view.cuh

- [x] Improve primitives names; especially, `in/out_nbr` in `copy_v_transform_reduce_in/out_nbr` is very confusing.
- `transform_reduce_in_nbr` here actually means iterate over the incoming edges of each vertex; produce one value per edge (by transforming edge source/destination IDs, edge weights, and edge source/destination properties); and reduce the produced values to update the (source) vertex (of the incoming edges) properties. Better rename to make this clearer.
- [x] We're mixing vertex/edge-centric terminologies and matrix-centric terminologies. Better be consistent and consistently use vertex/edge-centric terminologies.
- row/col=>src/dst, matrix_partition=>edge_partition
- e.g. `row/col_properties_t` => `edge_src/dst_properties_t`, `matrix_partition_(device_)view_t` => `edge_partition_(device_)view_t`, `is_adj_matrix_transposed` => `storage_transposed`
- [x] Better group `graph(_view)_t`'s member functions about vertex partition & edge partition.
- [x] `v_op` of `count_if_v` or `transform_reduce_v` may better take a vertex ID as an input parameter (as `e_op` takes source/destination vertex IDs and to be more consistent).
[...]

PR #2064, opened 2022-02-08
Multi gpu sample edges utilities

kaatish:
Add utilities to enable multi gpu gathering to be used for mnmg sampling.


ChuckHastings (approved):
LGTM

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+ * [vertex_input_first, vertex_input_last)
+ * @param global_degree_offset Global degree offset to local adjacency list for every source
+ * represented by current gpu
+ * @return A tuple of device vector containing the sources and destinations gathered locally
+ */
+template <typename GraphViewType, typename EdgeIndexIterator, typename gpu_t>
+std::tuple<rmm::device_uvector<typename GraphViewType::vertex_type>,
+           rmm::device_uvector<typename GraphViewType::vertex_type>,
+           rmm::device_uvector<gpu_t>>
+gather_local_edges(
```
seunghwak, line 402: Should we place this here or better move to graph_functions.hpp? Most functions under the include/cugraph/prims are the ones that take custom functors and cannot be explicitly instantiated (and we should expose implementation in the header).

I think this function can be explicitly instantiated like other functions in graph_functions.hpp (which have their implementations and explicit instantiation code under cpp/src).

And we may just use int (process ID) for gpu_t (unless we really want to allow users to use 8 bit and 16 bit GPU IDs).

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+ * gpus in the column communicator
+ * @param global_out_degrees Global out degrees for every source represented by current gpu
+ * @return Global out degrees of all sources in active_sources
+ */
+template <typename GraphViewType>
+rmm::device_uvector<typename GraphViewType::edge_type> get_active_source_global_degrees(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  rmm::device_uvector<typename GraphViewType::vertex_type>& active_sources,
+  const rmm::device_uvector<typename GraphViewType::edge_type>& global_out_degrees)
```
seunghwak, line 183: And should we reveal the detail of edge partitioning to end-users through our public API? We may just let users think about vertex partitioning and create a second version of compute_out_degrees
(https://github.com/rapidsai/cugraph/blob/branch-22.04/cpp/include/cugraph/graph_view.hpp#L591) taking a vertex list.

Creating a source global out degrees from local vertex's global out degrees should be handled in the detail space (or inside the implementation of public API).

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+ * @param graph_view Non-owning graph object.
+ * @return Tuple of two device vectors. The first one contains per source edge-count encountered
+ * by gpus in the column communicator before current gpu. The second device vector contains the
+ * global out degree for every source represented by current gpu
+ */
+template <typename GraphViewType>
+std::tuple<rmm::device_uvector<typename GraphViewType::edge_type>,
+           rmm::device_uvector<typename GraphViewType::edge_type>>
+get_global_degree_information(raft::handle_t const& handle, GraphViewType const& graph_view)
+{
```
seunghwak, line 260: What's the difference between this function and compute_out_degrees (https://github.com/rapidsai/cugraph/blob/branch-22.04/cpp/include/cugraph/graph_view.hpp#L591)?

seunghwak (commented):
More review comments are coming.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+    input_iter,
+    thrust::remove_if(
+      handle.get_thrust_policy(),
+      input_iter,
+      input_iter + destinations.size(),
+      destinations.begin(),
+      [invalid_vertex_id] __device__(auto dst) { return (dst == invalid_vertex_id); }));
+  sources.resize(compacted_length, handle.get_stream());
+  destinations.resize(compacted_length, handle.get_stream());
+  destination_gpu_ids.resize(compacted_length, handle.get_stream());
```
seunghwak, line 490: And should we really return destination_gpu_ids? We can return just the number of edges for each vertex partition.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+      //  v - seg[partition_id]
+      auto location_in_segment = source - vertex_id_seg[partition_id];
+
+      // csr offset value for vertex v that belongs to partition (partition_id)
+      auto csr_offset       = graph_offsets[partition_id][location_in_segment];
+      auto local_out_degree = graph_offsets[partition_id][location_in_segment + 1] - csr_offset;
+      edge_t const* adjacency_list = graph_indices[partition_id] + csr_offset;
+      // read location of global_degree_offset needs to take into account the
+      // partition offsets because it is a concatenation of all the offsets
+      // across all partitions
```
seunghwak, line 467: And should we really iterate over each edge?

This is beneficial comparing to iterating over vertices if vertex degrees vary a lot, but here loop count per vertex is `indices_per_source` so no thread-divergence issue.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+ * Collect all the edges that are present in the adjacency lists on the current gpu
+ *
+ * @tparam GraphViewType Type of the passed non-owning graph object.
+ * @tparam EdgeIndexIterator Type of the iterator for edge indices.
+ * @tparam GPUIdIterator  Type of the iterator for gpu id identifiers.
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Non-owning graph object.
+ * @param active_sources_in_row Device vector containing all the vertex id that are processed by
+ * gpus in the column communicator
```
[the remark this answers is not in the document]
kaatish: We can move all of these functions to detail namespace since they will be used for sampling.

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: This gives two results. The exclusive_scan of the degrees up till that gpu and the total out degrees

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+  using edge_t   = typename GraphViewType::edge_type;
+  using weight_t = typename GraphViewType::weight_type;
+
+  rmm::device_uvector<edge_t> local_degrees(
+    graph_view.get_number_of_local_adj_matrix_partition_rows(), handle.get_stream());
+
+  // get_number_of_local_adj_matrix_partition_rows == summation of get_major_size() of all
+  // partitions belonging to the gpu
+  vertex_t partial_offset{0};
+  for (size_t i = 0; i < graph_view.get_number_of_local_adj_matrix_partitions(); ++i) {
```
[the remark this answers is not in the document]
kaatish: I need local out degrees and not just the total out degrees. I could pull out the code from that function and make compute_local_out_degrees.

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: This is a requirement for sampling. The destination_gpu_ids are more like properties associated with each source. It has nothing to do with the gpu id of where this code is run.

cpp/include/cugraph/prims/gather_edges.cuh
seunghwak: Then, can't you 1) compute "total out degrees" first and 2) sum total out degrees in each GPU, and 3) run exclusive scan? This sounds much simpler than the code here.

seunghwak: Umm... please disregard, I think I misunderstood your explanation.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+        recv_data.begin(), recv_data.size(), comm_rank - row_size, handle.get_stream());
+      thrust::transform(handle.get_thrust_policy(),
+                        temp_input.begin(),
+                        temp_input.end(),
+                        recv_data.begin(),
+                        temp_input.begin(),
+                        thrust::plus<edge_t>());
+    }
+    handle.get_comms().barrier();
+  }
```
seunghwak, line 298: Just FYI for future optimization.

So, in this code, only two GPUs (out of col_comm_size GPUs) will be active in each iteration, only one GPU has computation. This will become a scalability bottleneck especially when col_comm_size is large.

What you may do instead is

There are col_comm_size number of blocks in local_degrees.

col_comm_rank = 0 starts with the block0
col_comm_rank = 1 starts with the block1
and so on....

Then, each GPU will send the data for one block and receive and compute the data for one block as well in each iteration. Every GPU will do a comparable amount of send/receive and computation.

You may add a FIXME statement for future reference.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+      comm.device_recv(
+        recv_data.begin(), recv_data.size(), comm_rank - row_size, handle.get_stream());
+      thrust::transform(handle.get_thrust_policy(),
+                        temp_input.begin(),
+                        temp_input.end(),
+                        recv_data.begin(),
+                        temp_input.begin(),
+                        thrust::plus<edge_t>());
+    }
+    handle.get_comms().barrier();
```
seunghwak, line 297: Do you really need barrier on the global communicator? Wont' col_comm.barrier() sufficient?

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+#include <cugraph/utilities/host_scalar_comm.cuh>
+
+#include <raft/handle.hpp>
+
+#include <numeric>
+#include <thrust/distance.h>
+#include <utility>
+
+#define DEBUG
+#ifdef DEBUG
```
seunghwak, line 32: Don't forget to delete this once this code has been validated. Or, if you really want to keep this for future debugging, you may better move this to the utilities directory. Better not mix long debug code with the main code for code readability and maintainability.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+  }
+}
+}  // namespace debug
+#endif
+namespace cugraph {
+
+namespace detail {
+
+/**
+ * @brief Calculate local out degrees of the sources belonging to the adjacency matrices
```
seunghwak, line 167: And we're using major/minor in the detail namespace and source & destination (currently, we're mixing row&column as well, but in long run, I think we'd better not mix, and just use source & destination) in the public namespace, this same code can be used to compute in-degrees if the storage is transposed. This code computes major degrees (which are out-degrees in CSR + DCSR and in-degrees in CSC + DCSC).

Better rename this to `compute_local_major_degrees()` to be consistent with the functions in graph_impl.cuh (e.g. https://github.com/rapidsai/cugraph/blob/branch-22.04/cpp/src/structure/graph_view_impl.cuh#L86).

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+rmm::device_uvector<typename GraphViewType::edge_type> calculate_local_degrees(
+  raft::handle_t const& handle, GraphViewType const& graph_view)
+{
+  static_assert(GraphViewType::is_adj_matrix_transposed == false);
+  using vertex_t = typename GraphViewType::vertex_type;
+  using edge_t   = typename GraphViewType::edge_type;
+  using weight_t = typename GraphViewType::weight_type;
+
+  rmm::device_uvector<edge_t> local_degrees(
+    graph_view.get_number_of_local_adj_matrix_partition_rows(), handle.get_stream());
```
seunghwak, line 189: Better be `GraphViewTypes::is_adj_matrix_transposed ? graph_view.get_number_of_local_adj_matrix_partition_cols() : graph_view.get_number_of_local_adj_matrix_partition_rows();` for better re-useability.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+                        offsets       = matrix_partition.get_offsets(),
+                        local_degrees = thrust::raw_pointer_cast(csr_begin)] __device__(auto i) {
+                         auto d = offsets[(major_hypersparse_first - major_first) + i + 1] -
+                                  offsets[(major_hypersparse_first - major_first) + i];
+                         auto v                         = vertex_ids[i];
+                         local_degrees[v - major_first] = d;
+                       });
+    } else {
+      auto csr_begin = local_degrees.begin() + partial_offset;
+      auto csr_end   = local_degrees.begin() + partial_offset + matrix_partition.get_major_size();
```
seunghwak, line 245: If we compute local major degrees, this is no longer necessarily a CSR but can be either CSR or CSC. Maybe use sparse_begin() and sparse_end() instead?

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+        ? ((*segment_offsets).size() > (detail::num_sparse_segments_per_vertex_partition + 1))
+        : false;
+
+    if (use_dcs) {
+      auto major_hypersparse_first =
+        matrix_partition.get_major_first() +
+        (*segment_offsets)[detail::num_sparse_segments_per_vertex_partition];
+      // Calculate degrees in csr region
+      auto csr_begin = local_degrees.begin() + partial_offset;
+      auto csr_end   = local_degrees.begin() + partial_offset +
```
seunghwak, line 212: If we compute local major degrees, this is no longer necessarily a CSR but can be either CSR or CSC. Maybe use sparse_begin() and sparse_end() instead?

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+      ;
+
+      thrust::tabulate(handle.get_thrust_policy(),
+                       csr_begin,
+                       csr_end,
+                       [offsets = matrix_partition.get_offsets()] __device__(auto i) {
+                         return offsets[i + 1] - offsets[i];
+                       });
+
+      // Calculate degrees in dcsr region
```
seunghwak, line 223: Better use hypersparse instead of dcsr (as this can be dcsc as well if the storage is transposed).

cpp/include/cugraph/prims/gather_edges.cuh
seunghwak: And if we move this function to the detail space, source here should be renamed to major.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+  rmm::device_uvector<vertex_t> vertex_id_end(id_end.size(), handle.get_stream());
+  rmm::device_uvector<vertex_t> vertex_count_offsets(count_offsets.size(), handle.get_stream());
+  raft::update_device(
+    vertex_id_begin.data(), id_begin.data(), id_begin.size(), handle.get_stream());
+  raft::update_device(vertex_id_end.data(), id_end.data(), id_end.size(), handle.get_stream());
+  raft::update_device(
+    vertex_count_offsets.data(), count_offsets.data(), count_offsets.size(), handle.get_stream());
+
+  auto& comm           = handle.get_comms();
+  auto const comm_rank = comm.get_rank();
```
seunghwak, line 416: So, is this code intended to use only for the multi-GPU cases? If not, this should be guarded inside `if constepxr (GraphViewType::is_multi_gpu)`. If to be used only for the multi-GPU use cases, there should be either something like `std::enable_if<GraphViewType::is_multi_gpu>` or `static_assert(GrpahViewType::is_multi_gpu);`

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+ * @param active_sources Device vector containing all the vertex id that are processed by
+ * gpus in the column communicator
+ * @param global_out_degrees Global out degrees for every source represented by current gpu
+ * @return Global out degrees of all sources in active_sources
+ */
+template <typename GraphViewType>
+rmm::device_uvector<typename GraphViewType::edge_type> get_active_source_global_degrees(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  rmm::device_uvector<typename GraphViewType::vertex_type>& active_sources,
```
seunghwak, line 380: Can't this be `const`?

And if this is read-only, now we should consider using `raft::device_span` (and actually, if we don't need to resize, we can use device_span).

cpp/include/cugraph/prims/gather_edges.cuh
seunghwak: What you need is just comm_rank, you can do `auto const comm_rank = GraphViewType::is_multi_gpu ? handle.get_comms().get_rank() : int{0};`.

`handle.get_comms().get_rank()` will seg-fault in single-GPU.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+                    active_source_degrees.begin(),
+                    [id_begin = vertex_id_begin.data(),
+                     id_end   = vertex_id_end.data(),
+                     comm_rank,
+                     global_out_degrees   = global_out_degrees.data(),
+                     vertex_count_offsets = vertex_count_offsets.data(),
+                     count                = vertex_id_end.size()] __device__(auto v) {
+                      // Find which partition id did the source belong to
+                      auto partition_id = thrust::distance(
+                        id_end, thrust::upper_bound(thrust::seq, id_end, id_end + count, v));
```
seunghwak, line 430: And if `active_sources` are sorted, you really don't need to run thrust::upper_bound for every element. But there is sorting overhead instead.... So, not sure which is better in small-scale, but in large-scale, see the discussion below.

And if I am not mistaken, you're getting `global_out_degrees` (or I'd better say `global_major_degrees`) for the entire major range (which is V/P * col_comm_size). Beware that this can be a memory capacity bottleneck if col_comm_size gets large.

What you can do instead is loop over matrix_partitions (or edge_partitions as we have discussed about improving the primitives API https://github.com/rapidsai/cugraph/issues/2003), broadcast global degrees for the major range, and update global degrees for the sources that belong to the matrix partition range.

And here, we can manage memory footprint vs parallelism trade-offs by processing multiple edge partitions concurrently using multiple streams.

You may not need to do this from the beginning, but if you really want to go to very large scale, you may consider making this updates at some point.

seunghwak (commented):
More reviews coming.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+  }
+  // Get global degrees
+  device_bcast(col_comm,
+               temp_input.begin(),
+               temp_input.begin(),
+               temp_input.size(),
+               col_size - 1,
+               handle.get_stream());
+
+  return std::make_tuple(std::move(recv_data), std::move(temp_input));
```
seunghwak, line 512: Note that sizes of recv_data & temp_input are V/P * col_comm_size, so this can be a scalability bottleneck (in terms of memory capacity especially when col_comm_size > E/V).

And should we really store the major_degrees and exclusive_scan of the local major degrees for the entire major range?

I am not sure how the sampling will be implemented end-to-end, but if I make a guess.

1) store major_degrees (or major weight sums) for V/P vertices (not for the V/ P * col_comm_size) in each GPU.

2) Once the active source vertices are given, shuffle sources to the corresponding vertex partition, then, based on the global degrees (or weight sums), generate random numbers within the ranges.

Broadcast the random numbers (only for the active sources) to the GPUs in the same column communicator. Compute exclusive scan of the local degrees "only for the active source vertices". Based on the random numbers, sample edges. Then, collect edges back to the final destination GPUs.

This may involve some recomputing (but only for the active source vertices) but memory footprint will be much more scalable. And recomputing cost may be pretty moderate when # active sources << V.

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,710 @@
...
+    std::accumulate(external_source_counts.begin(), external_source_counts.end(), size_t{0});
+  std::vector<size_t> displacements(external_source_counts.size(), size_t{0});
+  std::exclusive_scan(
+    external_source_counts.begin(), external_source_counts.end(), displacements.begin(), size_t{0});
+
+  rmm::device_uvector<vertex_t> active_sources(total_external_source_count, handle.get_stream());
+  rmm::device_uvector<gpu_t> active_source_gpu_ids(total_external_source_count,
+                                                   handle.get_stream());
+  // Get the sources other gpus on the same row are working on
+  // TODO : replace with device_bcast for better scaling
```
seunghwak, line 559: +1 for this, and AFAIK, we have agreed to use FIXME instead of TODO.

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: We will be getting a sense of the workload when this is used in sampling and I will be inclined to make changes to address those bottlenecks.

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: I am not sure I fully understand your approach now but I'll get there. Having a more performant vector exclusive sum will decide if we can avoid gathering all of the global out degrees and global out offsets.

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: > And recomputing cost may be pretty moderate when # active sources << V.

I am not sure about this. For the first call this might be true. But the subsequent calls will probably increase the number of active vertices.

As per my current understanding, my aim was to calculate these values (degrees and offsets) once because the number of gather calls could be quite large.
Once I get a sense of what the memory vs communication trade off looks like I will fix this.

See [here](https://github.com/rapidsai/cugraph/pull/2064/files/258b7d4c1e4a185c1b454dd8590f0031dd61496f#r804779751).

cpp/include/cugraph/prims/gather_edges.cuh
kaatish: I was in the process of changing it when I found out that you have to explicitly create a span out of the device_uvector. Perhaps we should add a constructor that instantiates from supported containers like cudf.

seunghwak: I see, not a high priority thing at all, if this is a tricky thing, then, we can revisit this later in a separate PR (maybe create a PR just updating cugraph public C++ API functions to use span).

cpp/include/cugraph/prims/gather_edges.cuh
```diff
@@ -0,0 +1,495 @@
...
+           rmm::device_uvector<typename GraphViewType::vertex_type>,
+           rmm::device_uvector<gpu_t>>
+gather_local_edges(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  rmm::device_uvector<typename GraphViewType::vertex_type>& active_sources_in_row,
+  rmm::device_uvector<gpu_t>& active_source_gpu_ids,
+  EdgeIndexIterator edge_index_first,
+  typename GraphViewType::vertex_type invalid_vertex_id,
+  int indices_per_source,
```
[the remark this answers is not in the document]
kaatish: I have changed it to edge type. We do not expect this value to exceed the maximum vertex degree.

cpp/include/cugraph/detail/graph_functions.cuh
```diff
@@ -0,0 +1,584 @@
...
+           rmm::device_uvector<typename GraphViewType::vertex_type>,
+           rmm::device_uvector<gpu_t>>
+gather_local_edges(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  const rmm::device_uvector<typename GraphViewType::vertex_type>& active_majors_in_row,
+  const rmm::device_uvector<gpu_t>& active_major_gpu_ids,
+  EdgeIndexIterator edge_index_first,
+  size_t indices_per_major,
+  const rmm::device_uvector<typename GraphViewType::edge_type>& global_degree_offsets)
```
seunghwak, line 486: Should we expose all these implementation details in the header file under `cpp/include`?

Cant' we just expose the header for `gather_local_edges` and place implementations under `src/sampling/gather_local_edges_impl.cuh` (or better place/naming under the `src` directory)? I assume users will just call `gather_local_edges`???

cpp/include/cugraph/detail/graph_functions.cuh
```diff
@@ -0,0 +1,191 @@
...
+ * @tparam GraphViewType Type of the passed non-owning graph object.
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Non-owning graph object.
+ * @return A single vector containing the local out degrees of the sources belong to the adjacency
+ * matrices
+ */
+template <typename GraphViewType>
+rmm::device_uvector<typename GraphViewType::edge_type> compute_local_major_degrees(
+  raft::handle_t const& handle, GraphViewType const& graph_view);
```
seunghwak, line 58: Sorry for nitpicking but we need a space here between two function declarations.

seunghwak (approved):

ChuckHastings:
@gpucibot merge

PR #2069, opened 2022-02-11
add node2vec C API implementation

ChuckHastings:
Provides C API implementation wrapping node2vec.

### DIFF
--- a/cpp/include/cugraph/api_helpers.hpp
+++ b/cpp/include/cugraph/api_helpers.hpp
@@ -28,6 +28,14 @@ enum class sampling_strategy_t : int { UNIFORM = 0, BIASED, NODE2VEC };
 struct sampling_params_t {
   sampling_params_t(void) {}
 
+  sampling_params_t(sampling_strategy_t sampling_type, double p = 1.0, double q = 1.0, bool use_alpha_cache = false)
+    : sampling_type_(sampling_type),
+      p_(p),
+      q_(q),
+      use_alpha_cache_(use_alpha_cache)
+  {
+  }
+
   sampling_params_t(int sampling_type, double p = 1.0, double q = 1.0, bool use_alpha_cache = false)
     : sampling_type_(static_cast<sampling_strategy_t>(sampling_type)),
       p_(p),
--- a/cpp/include/cugraph_c/algorithms.h
+++ b/cpp/include/cugraph_c/algorithms.h
@@ -328,6 +328,8 @@ typedef struct {
  *                           needs to be transposed
  * @param [in]  sources      Array of source vertices
  * @param [in]  max_depth    Maximum length of the generated path
+ * @param [in]  compress_result If true, return the paths as a compressed sparse row matrix,
+ *                              otherwise return as a dense matrix
  * @param [in]  p            The return parameter
  * @param [in]  q            The in/out parameter
  * @param [in]  result       Output from the node2vec call
@@ -339,7 +341,7 @@ cugraph_error_code_t cugraph_node2vec(const cugraph_resource_handle_t* handle,
                                       cugraph_graph_t* graph,
                                       const cugraph_type_erased_device_array_view_t* sources,
                                       size_t max_depth,
-                                      bool_t flag_use_padding,
+                                      bool_t compress_result,
                                       double p,
                                       double q,
                                       cugraph_random_walk_result_t** result,
@@ -359,8 +361,8 @@ size_t cugraph_random_walk_result_get_max_path_length(cugraph_random_walk_result
 /**
  * @brief     Get the matrix (row major order) of vertices in the paths
  *
- * @param [in]   result   The result from extract_paths
- * @return type erased array pointing to the matrix in device memory
+ * @param [in]   result   The result from a random walk algorithm
+ * @return type erased array pointing to the path matrix in device memory
  */
 cugraph_type_erased_device_array_view_t* cugraph_random_walk_result_get_paths(
   cugraph_random_walk_result_t* result);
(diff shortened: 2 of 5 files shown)


rlratzel (approved):

aschaffer (approved):

cpp/include/cugraph_c/algorithms.h
```diff
@@ -339,7 +341,7 @@ cugraph_error_code_t cugraph_node2vec(const cugraph_resource_handle_t* handle,
                                       cugraph_graph_t* graph,
                                       const cugraph_type_erased_device_array_view_t* sources,
                                       size_t max_depth,
-                                      bool_t flag_use_padding,
+                                      bool_t compress_result,
```
aschaffer, line 344: Okay, this is a problem: the semantics of `flag_use_padding` is the _opposite_ of `compress_result`: if you use padding, the result is not compressed; it is compressed when you do _not_ use padding.

aschaffer (changes requested):

cpp/include/cugraph_c/algorithms.h
ChuckHastings: Yes.  I changed the semantic meaning because it seemed more intuitive to me.  I tend to think of using compression to shrink the result as a behavior I want to enable or disable.

In the implementation I negate the value since, as you observe, it is the opposite of the C++ call.  So I get the correct behavior through this API.

aschaffer: Oh, okay, as long as you negate the flag all is well. Thank you!

cpp/include/cugraph/api_helpers.hpp
```diff
@@ -28,6 +28,14 @@ enum class sampling_strategy_t : int { UNIFORM = 0, BIASED, NODE2VEC };
...
 
+  sampling_params_t(sampling_strategy_t sampling_type,
+                    double p             = 1.0,
+                    double q             = 1.0,
+                    bool use_alpha_cache = false)
+    : sampling_type_(sampling_type), p_(p), q_(q), use_alpha_cache_(use_alpha_cache)
+  {
+  }
+
   sampling_params_t(int sampling_type, double p = 1.0, double q = 1.0, bool use_alpha_cache = false)
```
seunghwak, line 39: Why are we taking `int` `sampling_type` here but `sampling_strategy_t` `sampling_type` above? Better be consistent.

And more fundamentally, why do we have two functions seemingly doing the same thing?

cpp/src/c_api/random_walks.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+namespace cugraph {
+namespace c_api {
+
+struct cugraph_random_walk_result_t {
+  bool result_compressed_;
+  size_t max_path_length_;
+  cugraph_type_erased_device_array_t* paths_;
+  cugraph_type_erased_device_array_t* weights_;
+  cugraph_type_erased_device_array_t* sizes_;
```
seunghwak, line 38: Better add `{}` at the end to avoid holding garbage values.
```
  bool result_compressed_{false};
  size_t max_path_length_{0};
  cugraph_type_erased_device_array_t* paths_{nullptr};
  cugraph_type_erased_device_array_t* weights_{nullptr};
  cugraph_type_erased_device_array_t* sizes_{nullptr};
  ```

File: cpp/src/c_api/random_walks.cpp [hunk 4]
```diff
@@ -0,0 +1,193 @@
...
+
+struct node2vec_functor : public abstract_functor {
+  raft::handle_t const& handle_;
+  cugraph_graph_t* graph_;
+  cugraph_type_erased_device_array_view_t const* sources_;
+  size_t max_depth_;
+  bool compress_result_;
+  double p_;
+  double q_;
+  cugraph_random_walk_result_t* result_{};
```
seunghwak, line 49: Isn't it our convention to define member variables before functions? And don't forget {} at the end of variable names.

cpp/include/cugraph/api_helpers.hpp
ChuckHastings: The original code (developed for C++) included a constructor that takes an `int sampling_type` and converts it into the enum.  This original code is used by the current python/cython interface.

In implementing the C code, I wanted to use the enum directly (I think it's more clear in the calling code to use the enum).  I added the new constructor to allow for construction from the enum directly.  I intend to remove the old constructor once the python/cython code uses the new C API.

seunghwak: Gotcha, and in that case, FIXME statement stating the intention will make this clearer.

cpp/src/c_api/random_walks.cpp
ChuckHastings: Added {}

cpp/src/c_api/random_walks.cpp
ChuckHastings: Added the {}.  Not sure what you mean by your question.

seunghwak: What we typically have is
```
struct {
  functions(); ...
  variables; ...
};
```
but here
```
struct {
  variables;...
  functions(); ...
 };
```

seunghwak: Any reason for this?

ChuckHastings: Ah.  I interpreted your question as the opposite - hence my confusion.

What I have generally seen in our code base is that when we define a class we define it as:

class {
public:
   functions;
private:
   private_functions;
   variables;
};

But when we define a struct (essentially the same, only real difference is that everything is public) we define it as:

struct {
  variables;
  functions;
};

Admittedly, *most* of the structs that we define are to work around the nvcc problem with lambdas.  My feeling (from a design perspective) was that this functor is really a heavier weight lambda.  But you are correct in observing that it is more class-like.  It does derive from an abstract struct, hence the need for a constructor as well as an `operator()()`.

This is consistent across all of the C API functions.  We can certainly change it if we like, but if we do we should fix the entire C API (and better to do it before we add too many more calls).

ChuckHastings: And I guess the sampling_params_t is a struct but follows the class description in my above comment.  So we have inconsistency there as well.

seunghwak: Oh, yes, you're right, actually, I also do {vairables; operators();} for struct (i.e. functors). Seeing https://github.com/rapidsai/cugraph/pull/2069/files#diff-1264d5c0712616b080bb085897241ca6ad0bd8d4d132c83a75be4edf229ce988L28, I erroneously thought that we do functions first (do we consider constructors as something special, or should this better be class, or should we resolve inconsistencies in ordering?).

aschaffer (approved):

BradReesWork:
@gpucibot merge

BradReesWork (approved):

cpp/src/c_api/random_walks.cpp
ChuckHastings: @seunghwak - I'll make a note to consider this and write up some options for us to discuss.

PR #2071, opened 2022-02-14
Allow PropertyGraph `default_edge_weight` to be used to add an edge weight value on extracted Graphs even when a weight property wasn't specified

rlratzel:
Allows a Graph to be extracted with a default weight value even if an edge property for a weight is not specified.
This PR also changes the arg names that are intended to take column name strings to make it more obvious that strings are expected and not Column/Series objects.


BradReesWork (approved):

acostadon (approved):
didn't find any issues. I noticedthe column_name variable name change.

rlratzel:
> didn't find any issues. I noticedthe column_name variable name change.

Yes, thanks, I changed that to make it more obvious that it's a name and not an obj.  I'll update the PR description too.

jnke2016 (approved):

rlratzel:
@gpucibot merge

PR #2074, opened 2022-02-15
Skips ktruss docstring example for CUDA version 11.2, 11.4

betochimas:
Addresses unit test failures affecting CUDA 11.2 and 11.4 builds, skips ktruss doctest and avoids assertion error


rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #2077, opened 2022-02-15
C API code cleanup

ChuckHastings:
Pull out common cascading logic/error handling into a template function.  This cleans up the existing C API implementations and will make new C API implementations a bit simpler.

Note: #2069 uses the old paradigm.  Depending on merge order we might want to modify the node2vec implementation to use this technique also.  Can decide whether to do that in 2069 if this PR merges first, or if 2069 merges first we can either add that to this PR or leave it as tech debt to address later.


ChuckHastings:
rerun tests

seunghwak (approved):

ChuckHastings:
@gpucibot merge

PR #2080, opened 2022-02-17
Coarsening symmetric graphs leads to slightly asymmetric edge weights

seunghwak:
If you coarsen a symmetric (i.e. undirected) graph, the output graph should be symmetric as well.

However, due to limited floating point resolution, edge weights can be slightly asymmetric after coarsening (e.g. for a triplet of src, dst, weight, we may see (1, 2, 1.0) and its reverse edge (2, 1, 1.0 + 1e-7), this is only approximately symmetric and not strictly symmetric).

This PR fixes this by coarsening using only the lower triangular part (including the diagonal) after relabeling and reconstructing a symmetric graph from the lower triangular part (if the input graph is symmetric).


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #2081, opened 2022-02-17
Reduce peak memory requirement in graph creation (part 2/2)

seunghwak:
This PR reduces peak memory usage and replaces one large allocation with few smaller allocations with the same aggregate size (this works better with the pool allocator) in graph creation. Now we can create a graph from 2^30 edges (64 bit vertex IDs & 4 byte edge weight, 20B per edge) limiting the maximum pool size to 36 GB.


ChuckHastings (approved):

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2086, opened 2022-02-19
API Doc Namespace Edits + SimpleGraphImpl methods

betochimas:
Increases consistency between the API docs and the namespace in which the method can be called from, reducing potential confusion. Ex: cugraph.centrality.betweenness_centrality.betweenness_centrality can now be found as cugraph.betweenness_centrality. Note: both are valid ways of calling the same method.

Includes methods and attributes of a graph object once data is loaded (from SimpleGraphImpl), such as number_of_vertices, has_edge, get_two_hops_neighbors, etc.

Ready for review.


BradReesWork:
@gpucibot merge

PR #2087, opened 2022-02-19
Automatically clone raft when the raft pinned tag changes

cjnolet:
Now that RAFT has official conda packages, it'll be much easier to work locally on cugraph features that rely on the RAFT nightly. 

Working on features that span raft and cugraph should be fairly straightforward to do locally as well, because CPM provides the `CPM_raft_SOURCE` variable that can passed into the `cmake` call to tell the build to use a local working copy of RAFT.

The raft conda packages are now ready to add to the CI conda dependencies and once added, the updating the raft pinned tag in `get_raft.cmake` will no longer work. The option added in this PR will force clone a local version of the pinned tag (both locally in CI) and set the `CPM_raft_SOURCE` to that path.


cjnolet:
rerun tests

ChuckHastings:
Can you include an example [tested :-)] of how to use this, perhaps in `SOURCEBUILD.md`?

cjnolet:
> Can you include an example [tested :-)] of how to use this, perhaps in SOURCEBUILD.md?

Yes absolutely. I did test this locally and the changes work on the cmake side. Unfortunatelly I found the cuml Python (which I believe the cugraph is also derived) was broken so I also used the opportunity to integrate `pyraft` into the python side. 

I added some light instructions in [here](https://github.com/rapidsai/raft/blob/a9744076347ef3bd2c2537a49bc130432cfa7395/BUILD.md#building-raft-c-from-source-in-cmake) in raft. I can also add instructions to cugraph if it will be useful.

conda/recipes/pylibcugraph/meta.yaml
```diff
@@ -33,8 +33,10 @@ requirements:
     - ucx-proc=*=gpu
     - cudatoolkit {{ cuda_version }}.*
     - rmm {{ minor_version }}.*
+    - libraft-headers {{ minor_version }}
   run:
     - python x.x
+    - libraft-headers {{ minor_version }}
```
ajschmidt8, line 39: FYI, `conda` supports a recipe key called `run_exports` which will ensure that a package is automatically added to the `run` dependency list if it's used in the `host` dependency list. The `run_exports` key would need to be added to the `libraft-headers` package and then it could be safely removed from the `run` list here. `libcudf` does this.

- `conda` docs - https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements
- `libcudf` example - https://github.com/rapidsai/cudf/blob/c1638869116aae2c6dde6024394279a2fb79e685/conda/recipes/libcudf/meta.yaml#L34-L35

ajschmidt8 (approved):
Approving `ops-codeowner` file changes.

For what it's worth, the `cudf` repo has removed all `conda/environments/` files besides the CUDA `11.5` one ([link](https://github.com/rapidsai/cudf/tree/branch-22.04/conda/environments)) due to enhanced compatibility. It might not hurt for `cugraph` to follow suit. Then there would be less `.yml` files to maintain.

conda/recipes/pylibcugraph/meta.yaml
ajschmidt8: It looks like there's a few recipes in this PR that this comment could apply to.

cjnolet:
@ChuckHastings, I updated the `SOURCEBUILD.MD` . Once @robertmaynard's `rapids-cmake` [PR](https://github.com/rapidsai/rapids-cmake/pull/158) is merged, this should be ready.

cjnolet:
rerun tests

ChuckHastings (approved):
LGTM

SOURCEBUILD.md
```diff
@@ -144,6 +144,26 @@ This project uses cmake for building the C/C++ library. To configure cmake, run:
...
+                        PINNED_TAG <your_git_branch_or_tag>
+
+                        # When PINNED_TAG above doesn't match cugraph,
+                        # force local raft clone in build directory
+                        # even if it's already installed.
+                        CLONE_ON_PIN     ON
+                        )
+```
+
+If the above change is pushed to a pull request, the continuous integration servers will use the specified RAFT branch to run the cuGraph tests. For local development, the `CPM_raft_SOURCE=<path/to/raft/source>` option can be passed to the `cmake` command to enable `libcugraph` to use the local RAFT branch instead.
```
rlratzel, line 166: Thanks for this often overlooked update to the docs!

> For local development, the `CPM_raft_SOURCE=<path/to/raft/source>` option can be passed to the `cmake` command to enable `libcugraph` to use the local RAFT branch instead.

I wonder if this should be mentioned first, before the example showing how to change `get_raft.cmake`, since (I hope) just passing `CPM_raft_SOURCE` is the more common use case for changing the version of RAFT being used.
And I know this is somewhat vague, but can you also elaborate on when a user should need to change `get_raft.cmake` vs. just passing in the cmake option? The descriptions for both currently sound very similar ("working on new features" vs. "local development")

rlratzel (commented):
Thanks, Corey, I just have a couple minor doc-related suggestions: one for `SOURCEBUILD.md`, and also a request to add a `FIXME` somewhere (possibly in the `TODO` file you already started?) to mention that `pylibcugraph` should be able to remove the copy of RAFT code added here once the `connected_components` API is updated to use the new cugraph C API. The only reason RAFT is being pulled in is because the legacy connected_components in pylibcugraph uses it.

rlratzel (approved):
Changes look great, thanks!

cjnolet:
rerun tests

SOURCEBUILD.md
```diff
@@ -144,6 +144,30 @@ This project uses cmake for building the C/C++ library. To configure cmake, run:
   ```
 The default installation locations are `$CMAKE_INSTALL_PREFIX/lib` and `$CMAKE_INSTALL_PREFIX/include/cugraph` respectively.
 
+#### Updating the RAFT branch
+
+`libcugraph` uses the [RAFT](https://github.com/rapidsai/raft) library and there are times when it might be desirable to build against a different RAFT branch, such as when working on new features that might span both RAFT and cuGraph. 
+
+For local development, the `CPM_raft_SOURCE=<path/to/raft/source>` option can be passed to the `cmake` command to enable `libcugraph` to use the local RAFT branch.
```
robertmaynard, line 151: We really have two approaches to this.

1. CPM_raft_SOURCE which means for raft changes to be seen by cugraph you just need to build cugraph.  
2. CMAKE_PREFIX_PATH to the raft build directory.  This will allow you to use an already built version of raft, but any modifications of raft will require you to rebuild it ( unless the changes are header only )

rlratzel:
@gpucibot merge

PR #2089, opened 2022-02-22
Fix bug Random Walk in array sizes

ChuckHastings:
Random walk implementation returns a list of paths (vertex ids) and a list of edge weights for the edges on those paths.

@betochimas discovered that the size of the returned arrays is the same, even though the contents are different sizes.  In his example, if the return path contains one path:  `[ 0, 1, 3, 5 ]` it returns a weights array with 4 elements `[0.1, 2.1, 7.2, ???]` even though only the first three elements are valid.  The weights are edge weights, the path `[0,1,3,5]` only has 3 edges.  The first 3 elements in the returned array are correct.  Interpreting the results as intended will never reference that value.  However the sizes should be correct.

This PR fixes the initialization of the arrays to the correct size, which seems to correct the bug.

### DIFF
--- a/cpp/src/sampling/random_walks.cuh
+++ b/cpp/src/sampling/random_walks.cuh
@@ -776,10 +776,10 @@ random_walks_impl(raft::handle_t const& handle,
 
   // pre-allocate num_paths * max_depth;
   //
-  auto coalesced_sz = num_paths * max_depth;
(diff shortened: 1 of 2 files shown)


betochimas (approved):
Looks good to me!

seunghwak (approved):

BradReesWork:
@gpucibot merge

PR #2085, opened 2022-02-18
Add `node2vec` wrappers to pylibcugraph

betochimas:
Helps close #2058, when combined with a future PR for the cugraph layer implementation of node2vec, which depends on this PR. Ready for review. Disregard #2084 and #2063.

PR #2093, opened 2022-02-23
Add `node2vec` wrapper to cugraph

### LINKED ISSUE #2058: [ENH] Add `node2vec` API to cugraph
node2vec has been implemented in libcugraph, in the C++ library. The task is to add the API to `pylibcugraph` and `cugraph`, as well as writing tests and writing clear documentation for these node2vec methods

- [x] Add initial set of unit tests in `cugraph` for desired cugraph node2vec API
- [x] Write initial implementation in cugraph for `node2vec`, which either calls stubbed-out `pylibcugraph.node2vec()` or hardcodes fake data.
- [x] Add initial set of unit tests in `pylibcugraph`. These will be written to exercise the requirements of the `cugraph.node2vec()` implementation.
- [x] Write initial implementation in pylibcugraph for `node2vec`, which either calls a stubbed-out `node2vec` C API or hardcodes fake data.
- [x] Review https://github.com/rapidsai/cugraph/pull/2059
- [x] `cugraph.node2vec()` docstring
- [x] `pylibcugraph.node2vec()` docstring

betochimas:
Helps close #2058, when combined with #2085. Requires #2085 to be merged before merging, as the cugraph implementation requires the pylibcugraph node2vec wrapper


betochimas:
rerun tests

cjnolet:
rerun tests

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,147 @@
...
+
+    srcs = G.edgelist.edgelist_df['src']
+    dsts = G.edgelist.edgelist_df['dst']
+    weights = G.edgelist.edgelist_df['weights']
+
+    srcs = cupy.asarray(srcs)
+    dsts = cupy.asarray(dsts)
+    weights = cupy.asarray(weights)
+    start_vertices = cupy.asarray(start_vertices)
+
```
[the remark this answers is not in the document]
betochimas: I talked with Rick and we agreed that these conversions are unnecessary, will be fixed in the next commit

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,147 @@
...
+    >>> M = cudf.read_csv(datasets_path / 'karate.csv', delimiter=' ',
+    ...                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> G = cugraph.Graph()
+    >>> G.from_cudf_edgelist(M, source='0', destination='1', edge_attr='2')
+    >>> start_vertices = cudf.Series([0, 2])
+    >>> paths, weights, path_sizes = cugraph.node2vec(G, start_vertices, 3,
+    ...                                               True, 0.8, 0.5)
+
+    """
+    if (type(max_depth) != int) or (max_depth < 1):
```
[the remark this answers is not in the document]
betochimas: Addressed in a640a54

python/cugraph/cugraph/sampling/node2vec.py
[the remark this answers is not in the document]
betochimas: Addressed in a640a54

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+    A Grover, J Leskovec: node2vec: Scalable Feature Learning for Networks,
+    Proceedings of the 22nd ACM SIGKDD International Conference on Knowledge
+    Discovery and Data Mining, https://arxiv.org/abs/1607.00653
+
+    Parameters
+    ----------
+    G : cuGraph.Graph or networkx.Graph
+        The graph can be either directed (DiGraph) or undirected (Graph).
+
+    start_vertices: int or list or cudf.Series
```
jnke2016, line 36: Are we not supporting multi-column vertices for  `start_vertices` like the random walk implementation?

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pylibcugraph
+import cudf
+from cugraph.utilities import ensure_cugraph_obj_for_nx
+
+
+def node2vec(G, start_vertices, max_depth, use_padding, p=1.0, q=1.0):
```
jnke2016, line 19: No default value for both max_depth and use_padding which are both optional. If the user doesn't pass either or both, wouldn't you get a `required positional argument missing error`?

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+    Parameters
+    ----------
+    G : cuGraph.Graph or networkx.Graph
+        The graph can be either directed (DiGraph) or undirected (Graph).
+
+    start_vertices: int or list or cudf.Series
+        A single node or a list or a cudf.Series of nodes from which to run
+        the random walks
+
+    max_depth: int, optional
```
jnke2016, line 40: follow up with the declaration from the function declaration I commented

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+        The maximum depth of the random walks
+
+    use_padding: bool, optional
```
jnke2016, line 43: follow up with the declaration from the function declaration I commented

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+        If True, padded paths are returned else coalesced paths are returned
+
+    p: double, optional
```
jnke2016, line 46: Probably let the user know that the default value is `1.0`

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+        Return factor, which represents the likelihood of backtracking to
+        a previous node in the walk. A higher value makes it less likely to
+        sample a previously visited node, while a lower value makes it more
+        likely to backtrack, making the walk "local"
+
+    q: double, optional
```
jnke2016, line 52: Same comment as above

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+    >>> G = cugraph.Graph()
+    >>> G.from_cudf_edgelist(M, source='0', destination='1', edge_attr='2')
+    >>> start_vertices = cudf.Series([0, 2])
+    >>> paths, weights, path_sizes = cugraph.node2vec(G, start_vertices, 3,
+    ...                                               True, 0.8, 0.5)
+
+    """
+    if (not isinstance(max_depth, int)) or (max_depth < 1):
+        raise ValueError("'max_depth' must be a positive integer")
+    if (not isinstance(use_padding, bool)):
```
jnke2016, line 83: This will require `use_padding` to be set to a bool value by default in case the user doesn't provide this `optional` argument

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+    p: double, optional
+        Return factor, which represents the likelihood of backtracking to
+        a previous node in the walk. A higher value makes it less likely to
+        sample a previously visited node, while a lower value makes it more
+        likely to backtrack, making the walk "local"
+
+    q: double, optional
+        In-out factor, which represents the likelihood of visiting nodes
+        closer or further from the outgoing node. If q > 1, the random walk
+        is likelier to visit nodes closer to the outgoing node. If q < 1, the
```
jnke2016, line 55: if the `Return` and `In-out factor` are required to be positive double, probably mention it in the docsctring like 
`0 <= q < 1` so that the user is aware of it?

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+        start_vertices = cudf.Series(start_vertices)
+
+    if G.renumbered is True:
+        if isinstance(start_vertices, cudf.DataFrame):
+            start_vertices = G.lookup_internal_vertex_id(
+                start_vertices, start_vertices.columns)
+        else:
+            start_vertices = G.lookup_internal_vertex_id(start_vertices)
+
+    srcs = G.edgelist.edgelist_df['src']
```
jnke2016, line 105: This call will return the renumbered dataframe and will raise an error if the user didn't create a cugraph.Graph from a renumbered dataframe. You should probably first check if G is renumbered to avoid this

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,140 @@
...
+    dsts = G.edgelist.edgelist_df['dst']
+    weights = G.edgelist.edgelist_df['weights']
+
+    resource_handle = pylibcugraph.experimental.ResourceHandle()
+    graph_props = pylibcugraph.experimental.GraphProperties(
+                    is_multigraph=G.is_multigraph())
+    store_transposed = False
+    renumber = G.renumbered
```
jnke2016, line 113: This line is ambiguous because the dataset is already renumbered. I assume you should pass False to pylibcugraph if it is already the case. This is probably a design question at this point to see if we want pylibcugraph of the cugraph python to do the renumbering

python/pylibcugraph/pylibcugraph/node2vec.pyx
```diff
@@ -0,0 +1,184 @@
...
+    assert_success,
+    copy_to_cupy_array,
+    assert_CAI_type,
+    get_c_type_from_numpy_type,
+)
+
+
+def EXPERIMENTAL__node2vec(EXPERIMENTAL__ResourceHandle resource_handle,
+                           _GPUGraph graph,
+                           src_array,
```
jnke2016, line 60: Probably rename it `seed _array` to avoid confusion. In fact `src_array` already has a meaning internally in `EXPERIMENTAL__SGGraph(EXPERIMENTAL__Graph)`

python/pylibcugraph/pylibcugraph/node2vec.pyx
```diff
@@ -0,0 +1,184 @@
...
+    ----------
+    resource_handle : ResourceHandle
+        Handle to the underlying device resources needed for referencing data
+        and running algorithms.
+
+    graph : SGGraph
+        The input graph.
+
+    src_array: device array type
+        Device array containing the pointer to the array of source vertices.
```
jnke2016, line 78: Probably replace `source` by `seed`

python/pylibcugraph/pylibcugraph/node2vec.pyx
```diff
@@ -0,0 +1,184 @@
...
+    # Extract individual device array pointers from result and copy to cupy
+    # arrays for returning.
+    cdef cugraph_type_erased_device_array_view_t* paths_ptr = \
+        cugraph_random_walk_result_get_paths(result_ptr)
+    cdef cugraph_type_erased_device_array_view_t* weights_ptr = \
+        cugraph_random_walk_result_get_weights(result_ptr)
+    cdef cugraph_type_erased_device_array_view_t* path_sizes_ptr = \
+        cugraph_random_walk_result_get_path_sizes(result_ptr)
+
+    cupy_paths = copy_to_cupy_array(c_resource_handle_ptr, paths_ptr)
```
jnke2016, line 177: I liked how the function `copy_to_cupy_array` eased the code from the previous implementation

python/pylibcugraph/pylibcugraph/tests/test_node2vec.py
```diff
@@ -0,0 +1,131 @@
...
+# =============================================================================
+# fixtures used in this test module are defined in conftest.py
+
+
+# =============================================================================
+# Tests
+# =============================================================================
+
+@pytest.mark.parametrize("compress_result", [True, False])
+def test_node2vec(sg_graph_objs, compress_result):
```
jnke2016, line 74: How is this working. I don't see you creating the fixture `sg_graph_objs` in `pylibcugraph/tests/conftest.py` unless I am missing something

python/pylibcugraph/pylibcugraph/tests/test_node2vec.py
```diff
@@ -0,0 +1,131 @@
...
+
+# =============================================================================
+# Test data
+# =============================================================================
+# The result names correspond to the datasets defined in conftest.py
+# Note: the only deterministic path(s) in the following datasets
+# are contained in Simple_1
+_test_data = {"karate.csv": {
+                  "seeds": cp.asarray([0, 0], dtype=np.int32),
+                  "paths": cp.asarray([0, 8, 33, 29, 26, 0, 1, 3, 13, 33],
```
jnke2016, line 27: I think this is good for now but I am a little concern with this approach of testing because of the randomness of the results.
1) did you make sure there are only 2 paths with depth 5 starting from seed 0?
2) Even if it is the case, the testing seems hardcoded and only a few seeds are tested. 
An alternative could be to pick random seeds and verify the accuracy of the results by checking the predecessors.  To illustrate this, let's take the first path 
[0, 8, 33, 29, 26]. you verify that 0 is a predecessor of 8, then 8 being a predecessor of 33 and so on... but I am not sure how to easily do this without involving cugraph.

python/cugraph/cugraph/sampling/node2vec.py
betochimas: It is supported, I'll update the documentation to reflect this

python/cugraph/cugraph/sampling/node2vec.py
betochimas: I assume you mean the description in the docstring right? I think I did that because the default value is None, but as you mentioned, `max_depth` certainly isn't optional, so I'll go ahead and remove that from node2vec and random_walks

python/cugraph/cugraph/sampling/node2vec.py
betochimas: Because both p and q represent likelihoods and not probabilities, they aren't bound like `0 <= q < 1` but rather `0 < p, q`. I can include it in the argument description

jnke2016: I was saying `0 <= q < 1` because I thought it needed to be a positive double?

from your code
```
if (q is None) or (q <= 0.0):
        raise ValueError("'q' must be a positive double")
```

python/pylibcugraph/pylibcugraph/tests/test_node2vec.py
betochimas: This approach of testing was also in the cugraph level, but yes when this was done I did not check that the walks taken were in fact valid, just putting placeholder values and focusing on making sure the start of each path was each of the seeds. I've since made the testing more robust on the cugraph side, but for pylibcugraph it's more complicated, I'm still figuring that out right now but I can include a FIXME or add an issue to revisit later

python/cugraph/cugraph/sampling/node2vec.py
betochimas: Both p and q can be greater than 1, so as long as they aren't None or a negative value, the ValueError wouldn't be raised

python/cugraph/cugraph/sampling/node2vec.py
betochimas: Investigating this right now, the pylibcugraph wrapper does not work as intended if renumbered is False and returns an incorrect input. I added a fixme on line 124 describing this, and I believe this has to be resolved from the pylibcugraph or c layer

python/cugraph/cugraph/sampling/node2vec.py
jnke2016: Got it. I didn't understand your comment at first. That it should just be positive double

betochimas:
rerun tests

acostadon (approved):
Great job on this. Testing especially.

betochimas:
rerun tests

python/pylibcugraph/pylibcugraph/node2vec.pyx
```diff
@@ -0,0 +1,183 @@
...
+    try:
+        import cupy
+    except ModuleNotFoundError:
+        raise RuntimeError("node2vec requires the cupy package, which could not "
+                           "be imported")
+    try:
+        import numpy
+    except ModuleNotFoundError:
+        raise RuntimeError("node2vec requires the numpy package, which could not "
+                           "be imported")
```
rlratzel, line 137: It doesn't look like you're using `numpy` in this function. If that's the case, then this should be removed.

rlratzel (commented):
This is looking good, thanks. I have a few comments below based on what I found so far. Aside from that, I'd also like to look into the renumbering comments Joseph posted so I may have a follow-up review.

python/cugraph/cugraph/sampling/node2vec.py
betochimas: Yes, it's a positive double in pylibcugraph bc of using C types, but in python it's a float (same precision). Either way, it's been clarified in a later description

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,151 @@
...
+    use_padding: bool, optional (default=False)
+        If True, padded paths are returned else coalesced paths are returned
+
+    p: double, optional (default=1.0, [0 < p])
+        Return factor, which represents the likelihood of backtracking to
+        a previous node in the walk. A higher value makes it less likely to
+        sample a previously visited node, while a lower value makes it more
+        likely to backtrack, making the walk "local". A positive double.
+
+    q: double, optional (default=1.0, [0 < q])
```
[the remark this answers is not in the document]
betochimas: Yup, I automatically looked at C type def that's double in pylibcugraph and thought it was a Python type. The description's been changed to saying 'float'

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -0,0 +1,181 @@
...
+        path_start += vertex_path_sizes[i]
+    assert err == 0
+
+
+@pytest.mark.parametrize("graph_file", utils.DATASETS_SMALL)
+@pytest.mark.parametrize("directed", DIRECTED_GRAPH_OPTIONS)
+@pytest.mark.parametrize("max_depth", [None, -1])
+@pytest.mark.parametrize("p", [None, -1])
+@pytest.mark.parametrize("q", [None, -1])
+def test_node2vec_invalid(
```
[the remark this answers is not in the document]
betochimas: I modified the invalid inputs test to a series of checks instead of all possible combinations, and also removed the variability of the external parameters `graph_file` and `directed`, greatly reducing the number of tests generated but with the same coverage (more coverage bc I added strings and floats in the case of max_depth), going from 60 to 13 tests

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,151 @@
...
+    ...                   dtype=['int32', 'int32', 'float32'], header=None)
+    >>> G = cugraph.Graph()
+    >>> G.from_cudf_edgelist(M, source='0', destination='1', edge_attr='2')
+    >>> start_vertices = cudf.Series([0, 2])
+    >>> paths, weights, path_sizes = cugraph.node2vec(G, start_vertices, 3,
+    ...                                               True, 0.8, 0.5)
+
+    """
+    if (max_depth is None) or (max_depth < 1):
+        raise ValueError("'max_depth' must be a positive integer")
```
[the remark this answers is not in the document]
betochimas: Done with other inputs as well

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -0,0 +1,151 @@
...
+    if (not isinstance(use_padding, bool)):
+        raise ValueError("'use_padding' must be a bool")
+    if (p is None) or (p <= 0.0):
+        raise ValueError("'p' must be a positive double")
+    if (q is None) or (q <= 0.0):
```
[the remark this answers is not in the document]
betochimas: I think the initial reason was for separating ValueErrors (like -1 being less than 0) and TypeErrors (None isn't a float), but they've been combined a potential ValueError being raised for each of the different parameters (a TypeError is still a ValueError)

python/pylibcugraph/pylibcugraph/node2vec.pyx
betochimas: I was looking at the doctest and thought "I need to import numpy still". I didn't, and it's been changed.

BradReesWork (approved):

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -0,0 +1,185 @@
...
+    ----------
+    G : cuGraph.Graph or networkx.Graph
+
+    start_vertices : int or list or cudf.Series
+
+    max_depth : int
+
+    use_padding : bool
+
+    p : double
```
rlratzel, line 57: minor: this and other places should be updated to say float for consistency.

rlratzel (changes requested):
I noticed one other minor change to make, plus verifying the renumbering questions, then it looks ready to approve.

rlratzel (approved):

ChuckHastings (approved):

jnke2016 (approved):
Looks good to me

BradReesWork:
@gpucibot merge

PR #2096, opened 2022-02-26
extract_if_e bug fix.

seunghwak:
extract_if_e primitive bug fix.

This primitive takes a device_view of row(or column)_properties_t objects, and in the internal implementation, we need to properly set the local edge partition ID to properly retrieve row properties (if CSR-ish data structure is used).

This cod was missing and this PR fixes the bug.


ChuckHastings (approved):

kaatish:
> extract_if_e primitive bug fix.

Could you document what the bug was?

ChuckHastings:
@gpucibot merge

PR #2098, opened 2022-02-28
Get openmpi from conda instead of system

Ethyling:


rlratzel:
Updated build env has compatible NetworkX/SciPy packages.

rerun tests

PR #2100, opened 2022-03-02
Graph prmitives API update

seunghwak:
This is one of the series of planned PRs to address https://github.com/rapidsai/cugraph/issues/2003

We want to consistently use source & destination (or src & dst as abbreviation) instead of mixing source/destination and row/column and better group function/class/parameter names based on the concept of vertex partition and edge partition.

This PR renames

`row/col_properties_t` to `edge_partition_src/dst_property_t`
and
`copy_to_adj_matrix_partition_row/col` to `update_edge_partition_src/dst_property`
and
`transform_reduce_by_adj_matrix_row|col_key_e` to `transform_reduce_by_src|dst_key_e`

and fix resulting compile errors.

Possibly a breaking change (even though I am not aware of any external projects to be affected).


seunghwak:
@kaatish @aschaffer FYI: this may possibly break graph-ops if they are directly using graph primitives.

aschaffer (approved):

ChuckHastings (approved):

BradReesWork:
@gpucibot merge

PR #2101, opened 2022-03-02
Use conda compilers

Ethyling:
This PR enables the usage of conda compilers to build conda packages. This is required to use `mambabuild`

### DIFF
--- a/conda/recipes/cugraph/meta.yaml
+++ b/conda/recipes/cugraph/meta.yaml
@@ -19,34 +19,36 @@ build:
   number: {{ GIT_DESCRIBE_NUMBER }}
   string: cuda{{ cuda_major }}_py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
   script_env:
-    - CC
-    - CXX
-    - CUDAHOSTCXX
     - PARALLEL_LEVEL
+  ignore_run_exports_from:
+    - nvcc_linux-64 # [linux64]
+    - nvcc_linux-aarch64 # [aarch64]
 
 requirements:
   build:
-    - python x.x
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }} {{ cuda_version }}
+    - sysroot_linux-64 {{ sysroot_version }}  # [linux64]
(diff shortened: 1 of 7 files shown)


Ethyling:
rerun tests

rlratzel:
Updated build env has compatible NetworkX/SciPy packages.

rerun tests

Ethyling:
rerun tests

python/pylibcugraph/setup.py
```diff
@@ -122,7 +122,7 @@ def run(self):
                   os.path.join(os.sys.prefix, "lib")
               ],
               libraries=['cudart', 'cusparse', 'cusolver', 'cugraph', 'nccl',
-                         'cugraph_c'],
+                         'cugraph_c', 'cublas'],
```
Ethyling, line 125: As our CI system force us to rebuild Python packages in GPU tests stage, we are compiling using CUDA 11.0 and we are linking to `libcusolver.so.10` which does not contain the `libcublas`. We need to explicitly link to `libcublas` to fix this

ajschmidt8 (approved):
Approving, pending the removal of the `-conda-comp` suffix for the `SCCACHE_S3_KEY_PREFIX`

conda/recipes/cugraph/meta.yaml
```diff
@@ -19,34 +19,36 @@ build:
...
+    - python
     - pylibcugraph={{ version }}
     - libcugraph={{ version }}
     - libraft-headers {{ minor_version }}
     - pyraft {{ minor_version }}
     - cudf={{ minor_version }}
-    - dask-cudf {{ minor_version }}
-    - dask-cuda {{ minor_version }}
-    - dask>=2022.03.0
-    - distributed>=2022.03.0
```
rlratzel, line 49: cugraph still directly depends on all of these packages and I prefer to avoid assuming that they'll all be properly installed transitively.  Is there a reason they're being a removed here?  If so, can it be documented somewhere in this file that we still use these packages?

Ethyling: Thank you for catching this Rick, this was not expected at all. PR updated to fix this

rlratzel (approved):

Ethyling:
@gpucibot merge

PR #2116, opened 2022-03-14
Remove major/minor from renumber_edgelist public functions.

seunghwak:
Partially addresses #2003

We should not use major,minor (instead of src, dst) in the public non-detail API and should use major,minor in the detail space.

The public non-detail namespace renumber_edgelist currently uses major, minor and this PR fixes this.

This PR also replaces row/col to src/dst in `single_gpu_renumber_edgelist_given_number_map` as we're aiming to consistently use src/dst instead of row/col in our public C++ API.

This is possibly breaking as this PR changes the public API (but I am not aware of any non-cugraph libraries directly calling renumber_edgelist).

### DIFF
--- a/python/cugraph/cugraph/structure/renumber_wrapper.pyx
+++ b/python/cugraph/cugraph/structure/renumber_wrapper.pyx
@@ -122,8 +122,8 @@ def renumber(input_df,           # maybe use cpdef ?
     cdef uintptr_t c_major_vertices = major_vertices.__cuda_array_interface__['data'][0]
     cdef uintptr_t c_minor_vertices = minor_vertices.__cuda_array_interface__['data'][0]
 
-    cdef uintptr_t shuffled_major = <uintptr_t>NULL
-    cdef uintptr_t shuffled_minor = <uintptr_t>NULL
+    cdef uintptr_t shuffled_src = <uintptr_t>NULL
+    cdef uintptr_t shuffled_dst = <uintptr_t>NULL
 
     # FIXME: Fix fails when do_check = True
     cdef bool do_check = False # ? for now...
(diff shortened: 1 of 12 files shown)


ChuckHastings (approved):

BradReesWork (approved):

python/cugraph/cugraph/structure/renumber_wrapper.pyx
```diff
@@ -186,13 +186,18 @@ def renumber(input_df,           # maybe use cpdef ?
                     shuffled_df = input_df
                     edge_counts_32 = make_unique[vector[int]](1, num_local_edges)
 
-                shuffled_major = major_vertices.__cuda_array_interface__['data'][0]
-                shuffled_minor = minor_vertices.__cuda_array_interface__['data'][0]
+                if not transposed:
```
jnke2016, line 189: checking `transposed` again makes it redundant since it is already done in 179. I believe idea was just to swap the column name which will swap the `src column` and `dst column` if `transpose=True`. However, unless I am misunderstanding something, `major_vertices` and `minor_vertices` should be assigned shuffled_df column after checking for `transposed`. If not , when `transpose=True`  `major_vertices` and `minor_vertices` does not get update. Unless `major_vertices` and `minor_vertices` point to the same address as shuffled_df columns which I doubt.

jnke2016: And sg doesn't seem to account for transpose in the current implementation

rlratzel (dismissed):

jnke2016 (approved):
Looks good to me.

BradReesWork:
@gpucibot merge

PR #2118, opened 2022-03-15
Add MG neighborhood sampling to pylibcugraph & cugraph APIs

### LINKED ISSUE #2108: [ENH] Add MG neighborhood sampling to cugraph Python API
- [ ] cugraph API signature discussed and defined
- [ ] cugraph tests in place
- [ ] cugraph docstring
- [ ] cugraph API implementation defined (calling pylibcugraph API stub)
- [ ] pylibcugraph tests in place
- [ ] pylibcugraph docstring
- [ ] pylibcugraph API implementation defined (calling C API stub)
- [ ] replace stubs with actual calls everywhere when C API finished, ensure tests with actual data pass

This depends on: #2073

<hr>
<h2>Notes</h2>

* A separate graph service for use by DGL would operate in a more parallel batch-type mode by
  * Collecting several calls from different clients and assigning each client a unique label (an integer type) and associating the starting vertices for that client with that label
  * The starting vertices arrays are combined into a single flat array of tuples of (starting_vertex, label)
  * When enough client calls are received to exceed some threshold, the service makes a single call to the pylibcugraph/C API `uniform_neighborhood_sample()` with the array of tuples.
  * The `uniform_neighborhood_sample()` is run (multi-GPU) and the result is processed.  The result is a flat list of tuples of tuples: ((src, dst, label, index), rx_counts) where index is used to reconstruct the path (eg. if index=n, that edge is then the nth edge in the path for that label) and rx_counts is ???
* Let start with this API for `cugraph`:

```
cugraph.dask.sampling.uniform_neighborhood(G, start_info_list, fanout_vals, with_replacement=True)
```
where `G` is a cugraph Graph instance created using `from_dask_cudf_edgelist()`

* Lets start with this API for `pylibcugraph`:
```
pylibcugraph.uniform_neighborhood_sampling(G, start_info_list, fanout_vals, with_replacement=True)
```
where `G` is a `MGGraph` instance.

betochimas:
Closes #2108 when merged. Requires both #2088 and #2156 to be merged before, the former because this uses MGGraph, and the later because of the C implementation of neighborhood sampling.


python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
```diff
@@ -0,0 +1,165 @@
...
+    """
+    Does neighborhood sampling.
+
+    Parameters
+    ----------
+    input_graph : cugraph.DiGraph
+        cuGraph graph, which contains connectivity information as dask cudf
+        edge list dataframe
+
+    start_info_list : list
```
jnke2016, line 88: Can you add the `docstring `and `cudf.Series` as an option?

python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
```diff
@@ -0,0 +1,165 @@
...
+            Contains the destination vertices from the sampling result
+        ddf['labels']: dask_cudf.Series
+            Contains the start labels from the sampling result
+        ddf['index']: dask_cudf.Series
+            Contains the indices from the sampling result
+        ddf['counts']: dask_cudf.Series
+            Contains the transaction counts from the sampling result
+    """
+
+    print("Hello from cugraph/dask!")
```
jnke2016, line 115: I won't comment on the debug print and comment since it looks like you are still working on it

python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
```diff
@@ -0,0 +1,165 @@
...
+                            with_replacement,
+                            workers=[wf[0]])
+              for idx, wf in enumerate(data.worker_to_parts.items())]
+
+
+    wait(result)
+
+    cudf_result = [client.submit(convert_to_cudf,
+                                 cp_arrays)
+                   for cp_arrays in cupy_result]
```
jnke2016, line 157: cupy_result is not reference anywhere. I believe you forgot to update the variable `result` to `cupy_result`

python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
```diff
@@ -0,0 +1,165 @@
...
+    Creates a cudf DataFrame from cupy arrays from pylibcugraph wrapper
+    """
+    cupy_sources, cupy_destinations, cupy_labels, cupy_indices = cp_arrays
+    # cupy_sources, cupy_destinations, cupy_labels, cupy_indices, cupy_counts = cp_arrays
+    df = cudf.DataFrame()
+    df["sources"] = cupy_sources
+    df["destinations"] = cupy_destinations
+    df["labels"] = cupy_labels
+    df["indices"] = cupy_indices
+    df["counts"] = cupy_counts
```
jnke2016, line 71: `cupy_counts`  is not referenced

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
```diff
@@ -0,0 +1,190 @@
...
+    Parameters
+    ----------
+    resource_handle: ResourceHandle
+        Handle to the underlying device and host resources needed for
+        referencing data and running algorithms.
+
+    input_graph: MGGraph
+        The input graph, for Multi-GPU operations.
+
+    start_info_list: device array type
```
jnke2016, line 79: My understanding is that there are 2 distinct variables (`start`, `start_label`) based on the C and C++ API. where is `start_info_list` coming from ?

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
```diff
@@ -0,0 +1,190 @@
...
+    Returns
+    -------
+    A tuple of device arrays, where the first and second items in the tuple
+    are device arrays containing the starting and ending vertices of each
+    walk respectively, the third item in the tuple is a device array
+    containing the start labels, the fourth item in the tuple is a device
+    array containing the indices for ..., and the fifth item in the tuple
+    is a device containing the counts for ...
+
+    Examples
```
jnke2016, line 105: This example won't work because you need to distribute the edges across workers(GPUs) and for each partition, call `pylibcugraph.experimental.MGGraph` . We do this by leveraging `dask`. I think you should remove the example for now until we find a solution for that.

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
```diff
@@ -0,0 +1,190 @@
...
+    cdef cugraph_sample_result_t* result_ptr
+    cdef cugraph_error_code_t error_code
+    cdef cugraph_error_t* error_ptr
+
+    cdef uintptr_t cai_start_ptr = \
+        start_info_list.__cuda_array_interface__["data"][0]
+    cdef uintptr_t cai_labels_ptr = \
+        start_info_list.__cuda_array_interface__["data"][1]
+    cdef uintptr_t cai_fan_out_ptr = \
+        h_fan_out.__cuda_array_interface__["data"][0]
```
jnke2016, line 136: if `h_fan_out` is a `list` , it doesn't have `__cuda_array_interface__` attribute.

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
jnke2016: Maybe you are consolidating those two?

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
```diff
@@ -0,0 +1,190 @@
...
+    cdef cugraph_graph_t* c_graph_ptr = input_graph.c_graph_ptr
+
+    cdef cugraph_sample_result_t* result_ptr
+    cdef cugraph_error_code_t error_code
+    cdef cugraph_error_t* error_ptr
+
+    cdef uintptr_t cai_start_ptr = \
+        start_info_list.__cuda_array_interface__["data"][0]
+    cdef uintptr_t cai_labels_ptr = \
+        start_info_list.__cuda_array_interface__["data"][1]
```
jnke2016, line 134: If you combined these two variables into one `start_info_list`, this won't work because `__cuda_array_interface__["data"][1]` refers to something totally different I think. Based on this[ documentation](https://numba.pydata.org/numba-doc/dev/cuda/cuda_array_interface.html), `__cuda_array_interface__["data"][1]` is read-only flag as a Python bool

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
betochimas: Yes, in the issue tied to this PR, issue#2108, the API combines the start vertices along with their client labels into `start_info_list`, which is separated into `start` and `start_label` later on

acostadon (changes requested):
Made a couple small suggestions. 
I saw some tests missing. Are those gonna be part of the initial release?

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
betochimas: Yeah, I agree about the part where you said to remove an example, the cugraph wrapper too wouldn't benefit from an example if it can't be tested using doctest

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
betochimas: start_list, info_list, and h_fan_out should all have `__cuda_array_interface` attributes, and so I've updated that wrapper to check all three for that using `assert_CAI_type`. `fanout_vals` has been renamed to `h_fan_out` because in lower level its pointer belongs in host memory, and 'h' helps indicate that

python/pylibcugraph/pylibcugraph/uniform_neighborhood_sampling.pyx
betochimas: This has been separated into two variables in pylibcugraph. It is still combined as `start_info_list`in the cugraph layer

python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
betochimas: Thanks for catching that.

python/cugraph/cugraph/dask/sampling/neighborhood_sampling.py
betochimas: It's been temporarily disabled, until a solution has been found for handing this variable which does not move in device memory

python/cugraph/cugraph/dask/__init__.py
```diff
@@ -11,10 +11,16 @@
...
+from .link_analysis.hits import hits
 from .traversal.bfs import bfs
 from .traversal.sssp import sssp
 from .common.read_utils import get_chunksize
 from .community.louvain import louvain
 from .centrality.katz_centrality import katz_centrality
 from .components.connectivity import weakly_connected_components
+
+from .sampling.neighborhood_sampling import EXPERIMENTAL__uniform_neighborhood
+uniform_neighborhood = experimental_warning_wrapper(EXPERIMENTAL__uniform_neighborhood)
```
rlratzel, line 26: We should move this to the `experimental` namespace, so this would go in `cugraph.experimental.dask`

I think we should also call it `uniform_neighborhood_sampling` since the `sampling` namespace isn't always used (as seen here).

python/pylibcugraph/pylibcugraph/utils.pyx
```diff
@@ -18,6 +18,7 @@ from libc.stdint cimport uintptr_t
 
 import numpy
 import cupy
+import cudf
```
rlratzel, line 21: There shouldn't be any cudf calls in pylibcugraph - was this an artifact of an old merge?

rlratzel (changes requested):

python/pylibcugraph/pylibcugraph/utils.pyx
betochimas: Yes, from the first MG Hits merge. It's been removed in this next commit

acostadon (approved):
My issues were addressed

rlratzel (approved):

jnke2016 (approved):

ChuckHastings (approved):

Ethyling (approved):

BradReesWork:
@gpucibot merge

PR #2120, opened 2022-03-16
Renumbering causing inconsistent outputs in node2vec

betochimas:
Renumbering is done in two different layers when using the python package of cugraph, one in cugraph and one in libcugraph. Because of a renumbering bug, edges and vertices are not being renumbered correctly, causing multiple vertices and edges to be potentially identified as the same (i.e. take node ‘0’ and ‘6’ - they could both appear as node ‘0’). Testing was added to try to identify the exact source, although that is currently a WIP.

### DIFF
--- a/python/cugraph/cugraph/tests/test_node2vec.py
+++ b/python/cugraph/cugraph/tests/test_node2vec.py
@@ -25,7 +25,9 @@ import cugraph
 # =============================================================================
 DIRECTED_GRAPH_OPTIONS = [False, True]
 DATASETS_SMALL = [pytest.param(d) for d in utils.DATASETS_SMALL]
-KARATE = DATASETS_SMALL[0][0][0]
+# KARATE = DATASETS_SMALL[0][0][0]
+KARATE = utils.RAPIDS_DATASET_ROOT_DIR_PATH/"karate.csv"
+ARROW = utils.RAPIDS_DATASET_ROOT_DIR_PATH/"small_arrow.csv"
 
 
 # =============================================================================
@@ -92,7 +94,7 @@ def test_node2vec_coalesced(
     err = 0
     for i in range(k):
         for j in range(max_depth - 1):
-            # weight = edge_weights[i * (max_depth - 1) + j]
+            weight = edge_weights[i * (max_depth - 1) + j]
             u = vertex_paths[i * max_depth + j]
             v = vertex_paths[i * max_depth + j + 1]
             # Walk not found in edgelist
@@ -100,9 +102,13 @@ def test_node2vec_coalesced(
                 err += 1
             # FIXME: Checking weights is buggy
             # Corresponding weight to edge is not correct
-            # expr = "(src == {} and dst == {})".format(u, v)
-            # if not (G.edgelist.edgelist_df.query(expr)["weights"] == weight):
-            #    err += 1
+            expr = "(src == {} and dst == {})".format(u, v)
+            edge_query = G.edgelist.edgelist_df.query(expr)
+            if not edge_query:
+                err += 1
+            else:
+                if edge_query["weights"] != weight:
+                    err += 1
     assert err == 0
 
 
@@ -135,7 +141,7 @@ def test_node2vec_padded(
     path_start = 0
     for i in range(k):
         for j in range(max_depth - 1):
-            # weight = edge_weights[i * (max_depth - 1) + j]
+            weight = edge_weights[i * (max_depth - 1) + j]
             u = vertex_paths[i * max_depth + j]
             v = vertex_paths[i * max_depth + j + 1]
             # Walk not found in edgelist
@@ -143,9 +149,13 @@ def test_node2vec_padded(
                 err += 1
             # FIXME: Checking weights is buggy
             # Corresponding weight to edge is not correct
-            # expr = "(src == {} and dst == {})".format(u, v)
-            # if not (G.edgelist.edgelist_df.query(expr)["weights"] == weight):
-            #    err += 1
+            expr = "(src == {} and dst == {})".format(u, v)
+            edge_query = G.edgelist.edgelist_df.query(expr)
+            if not edge_query:
+                err += 1
+            else:
+                if edge_query["weights"] != weight:
+                    err += 1
         # Check that path sizes matches up correctly with paths
         if vertex_paths[i * max_depth] != seeds[i]:
             err += 1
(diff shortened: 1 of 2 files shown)


betochimas:
As of this commit, CI fails because a specific reason: querying edges in the form of walks back to the original graph data does not work as intended. To confirm this, a temporary line graph file was created and placed within the testing directory, and at first the testing suite was failing because of misconceptions about the coalesced and padded outputs. The cugraph API has been revised to ensure consistency, and the testing was redone to fit this API change. The line graph tests all pass, but the same cannot be said of the tests using the karate dataset. The failing tests in question have some details on which edges were not found.

rlratzel:
raft update published which should resolve thrust cmake issue in CI.

rerun tests

betochimas:
When using cudf to import graph data, there is a bug where double padding occurs, blocking out half of the requested paths. This happens in both pylibcugraph and cugraph, and `test_node2vec_renumber_cudf` in both testing files shows this. In contrast, `test_node2vec_renumber_cupy` returns the correct paths, with the only difference being that cupy arrays are used.

python/cugraph/cugraph/sampling/node2vec.py
```diff
@@ -115,24 +124,22 @@ def node2vec(G,
...
-                                           store_transposed, renumber,
-                                           do_expensive_check)
-
-    vertex_set, edge_set, sizes = pylibcugraph.experimental.node2vec(
-                                    resource_handle, sg, start_vertices,
-                                    max_depth, use_padding, p, q)
+    sg = SGGraph(resource_handle, graph_props, srcs, dsts, weights,
+                 store_transposed, renumber, do_expensive_check)
+
+    vertex_set, edge_set, sizes = \
```
jnke2016, line 139: I am seeing you are not doing the conversion when `compress_result=False` (like you did in your previous PR when `use_padding=True`). Is that taken care of the the C++ layer when passing the parameter `compress_result`?

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -35,10 +37,20 @@ def setup_function():
     gc.collect()
 
 
+def _get_param_args(param_name, param_values):
+    """
+    Returns a tuple of (<param_name>, <pytest.param list>) which can be applied
+    as the args to pytest.mark.parametrize(). The pytest.param list also
+    contains param id string formed from teh param name and values.
```
jnke2016, line 44: There is typo.
And nice helper function. Usually we directly pass the list of parameters to @pytest.mark.parametrize for instance like this `DIRECTED = [pytest.param(d) for d in DIRECTED_GRAPH_OPTIONS]`.

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
     for bad_depth in invalid_max_depths:
```
jnke2016, line 96: Instead of having these `for loops` going over each parameter values,  why don't you just pass them to @pytest.mark.parametrize? isn't what it is meant for?

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
...
+
+    G = cugraph.Graph(directed=directed)
+
+    G.from_cudf_edgelist(cu_M, source="0", destination="1", edge_attr="2",
+                         renumber=False)
+    num_verts = G.number_of_vertices()
+    k = random.randint(6, 12)
+    start_vertices = random.sample(range(num_verts), k)
+    max_depth = 5
+    df, seeds = calc_node2vec(
```
jnke2016, line 148: Why `df` when it's a tuple of cudf.Series (vertex_paths, edge_weights, vertex_path_sizes)?

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
...
+        # necessary to use in order to track starts of paths
+        assert vertex_paths.size == vertex_path_sizes.sum()
+        if directed:
+            # directed graphs may be coalesced at any point
+            assert vertex_paths.size - k == edge_weights.size
+            # This part is for checking to make sure each of the edges
+            # in all of the paths are valid and are accurate
+            idx = 0
+            for path_idx in range(vertex_path_sizes.size):
+                for j in range(vertex_path_sizes[path_idx] - 1):
```
jnke2016, line 169: `j` is never used. you could probably do something like 
```
start_path_idx = 0
end_path_idx = 0
for path_idx in range(vertex_path_sizes.size):
      end_path_idx = start_path_ix + path_idx - 1
      for idx in range (start_path_idx, end_path_idx):
```
It's just a recommendation. Or substitute `j` with `_` so that we know it's not going to be used

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
...
+                    weight = edge_weights[idx]
+                    u = vertex_paths[idx + path_idx]
+                    v = vertex_paths[idx + path_idx + 1]
+                    edge_found = G.has_edge(u, v)
```
jnke2016, line 173: I like that you leveraged cuGraph. When I did something similar with random walks I was directly using cudf to check if the edge exist

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
...
+                    if not edge_found:
+                        raise ValueError("Edge {},{} not found".format(u, v))
+                    expr = "(src == {} and dst == {})".format(u, v)
+                    edge_query = G.edgelist.edgelist_df.query(expr)
+                    if edge_query.empty:
```
jnke2016, line 178: Why are you checking again if the edge (`src` and `dst`) exists since you did it earlier? can't you directly check for (`src`, `dst` , `values`). If you look at the implementation of `G.has_edge()`, you will see it is doing something similar to `edge_query = G.edgelist.edgelist_df.query(expr)`

jnke2016: It would have been easier if `G.has_edge()` was supporting `weight` as an argument :)

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,202 @@ def test_node2vec_invalid(
...
+                         dtype=['int32', 'int32', 'float32'], header=None)
+    G = cugraph.Graph(directed=True)
+    G.from_cudf_edgelist(cu_M, source="0", destination="1", edge_attr="2",
+                         renumber=renumber)
+
+    # if graph_file == KARATE:
+    #     start_vertices = [12, 28, 20, 23]
+    # else:
+    start_vertices = [8, 0, 7, 1, 6, 2]
+    num_seeds = 4
```
jnke2016, line 281: Why is `len(start_vertices) = 6` and `num_seeds = 4`? Did you forget to update `num_seeds` after you commented part of your code?

python/cugraph/cugraph/tests/test_node2vec.py
betochimas: Using @pytest.mark.parametrize would create tests for all combinations of invalid_max_depths, invalid_pqs, creating more tests than is necessary. This was a choice to reduce the number of tests without compromising coverage, since the pattern would be identical.

python/cugraph/cugraph/tests/test_node2vec.py
betochimas: Yes, that's correct, it's not a dataframe. It's been changed to result

cpp/src/c_api/random_walks.cpp
```diff
@@ -123,6 +123,14 @@ struct node2vec_functor : public abstract_functor {
         // std::make_unique<sampling_params_t>(2, p_, q_, false));
         std::make_unique<sampling_params_t>(cugraph::sampling_strategy_t::NODE2VEC, p_, q_));
 
+      //
+      // Need to unrenumber the vertices in the resulting paths
+      // FIXME: this is not necessary (wasteful) if the input graph was not
+      // renumbered.
+      //
```
seunghwak, line 130: We may just delete this comment, input graphs are 'by default' renumbered except for some special cases (e.g. debugging).

seunghwak (approved):
Approving the C/C++ part.

python/cugraph/cugraph/tests/test_node2vec.py
betochimas: substituted `j` with `_`

python/cugraph/cugraph/tests/test_node2vec.py
jnke2016: Make sense

python/cugraph/cugraph/sampling/node2vec.py
jnke2016: Got it. Thanks

python/cugraph/cugraph/tests/test_node2vec.py
betochimas: It was that. Updating the value of num_seeds did not affect the correctness of the test

python/cugraph/cugraph/tests/test_node2vec.py
betochimas: Something like `G.has_edge().has_weight()` or `G.has_edge_w_weight()` would have been nice. I removed the duplicate edge check

cpp/src/c_api/random_walks.cpp
ChuckHastings: Agree we should delete the comment.

* `unrenumber_local_int_vertices` is very fast (no comms, just a very efficient call to `thrust::transform`)
* Renumbering must be enabled for MG graphs
* For SG graphs it can be turned off in some circumstances, but transposing the graph will turn it on.  There is no bookkeeping in the C API to keep track of what kind of graph transformations have occurred.  So there's no good way to tell, even if you didn't initially renumber the graph, what the state of this current graph is.

ChuckHastings (approved):

cpp/src/c_api/random_walks.cpp
betochimas: It will be removed in the next commit

python/pylibcugraph/pylibcugraph/node2vec.pyx
```diff
@@ -172,7 +173,9 @@ def EXPERIMENTAL__node2vec(EXPERIMENTAL__ResourceHandle resource_handle,
     cupy_weights = copy_to_cupy_array(c_resource_handle_ptr, weights_ptr)
     cupy_path_sizes = copy_to_cupy_array(c_resource_handle_ptr,
                                            path_sizes_ptr)
-
+    # raise ValueError("{}".format(cupy_paths))
```
rlratzel, line 176: dead code?

rlratzel (commented):
just some minor things:

python/cugraph/cugraph/tests/test_node2vec.py
```diff
@@ -172,14 +96,183 @@ def test_node2vec_invalid(
...
+                                      compress_result=compress, p=bad_p, q=q)
     # Tests for invalid q
     for bad_q in invalid_pqs:
         with pytest.raises(ValueError):
             df, seeds = calc_node2vec(G, start_vertices, max_depth=max_depth,
-                                      use_padding=use_padding, p=p, q=bad_q)
+                                      compress_result=compress, p=p, q=bad_q)
+
+
+# Remove once bug is resolved
```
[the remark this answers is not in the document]
betochimas: At that moment, I was under the impression that the toy line graph would not remain once the PR left draft status, but since its utility can be used in other algos, it can go either way. I removed the comment for this next commit, but I can remove the test if the line graph example is also to be gone.

python/pylibcugraph/pylibcugraph/node2vec.pyx
betochimas: Yes, it's been removed.

rlratzel (approved):

PR #2148, opened 2022-03-22
[FIX] Use `rapids_find_package` to get `cugraph-ops`

trxcllnt:
Depends on https://github.com/rapidsai/raft/pull/583 and https://github.com/rapidsai/cugraph-ops/pull/70

This PR uses `CPMFindPackage` to find (or build) `cugraph-ops`.

This enables building from source against a local `cugraph-ops` build via `-D cugraph-ops_ROOT=/cugraph-ops/build`.

### DIFF
--- a/cpp/cmake/thirdparty/get_libcugraphops.cmake
+++ b/cpp/cmake/thirdparty/get_libcugraphops.cmake
@@ -15,35 +15,20 @@
 #=============================================================================
 
 set(CUGRAPH_MIN_VERSION_cugraph_ops "${CUGRAPH_VERSION_MAJOR}.${CUGRAPH_VERSION_MINOR}.00")
-set(CUGRAPH_BRANCH_VERSION_cugraph_ops "${CUGRAPH_VERSION_MAJOR}.${CUGRAPH_VERSION_MINOR}")
 
 function(find_and_configure_cugraph_ops)
 
-    set(oneValueArgs VERSION FORK PINNED_TAG CLONE_ON_PIN)
-    cmake_parse_arguments(PKG "" "${oneValueArgs}" "" ${ARGN} )
+    set(oneValueArgs VERSION)
+    cmake_parse_arguments(PKG "" "${oneValueArgs}" "" ${ARGN})
 
-    if(PKG_CLONE_ON_PIN AND NOT PKG_PINNED_TAG STREQUAL "branch-${CUGRAPH_BRANCH_VERSION_cugraph_ops}")
-        message("Pinned tag found: ${PKG_PINNED_TAG}. Cloning cugraph-ops locally.")
-        set(CPM_DOWNLOAD_cugraph-ops ON)
-    endif()
-
-    rapids_cpm_find(cugraph-ops ${PKG_VERSION}
+    rapids_find_package(cugraph-ops ${PKG_VERSION} REQUIRED
       GLOBAL_TARGETS      cugraph-ops::cugraph-ops++
       BUILD_EXPORT_SET    cugraph-exports
       INSTALL_EXPORT_SET  cugraph-exports
-        CPM_ARGS
-            GIT_REPOSITORY https://github.com/${PKG_FORK}/cugraph-ops.git
-            GIT_TAG        ${PKG_PINNED_TAG}
-            OPTIONS
-                "BUILD_CUGRAPH_OPS_CPP_LIBRARY ON"
-                "BUILD_CUGRAPH_OPS_CPP_TESTS OFF"
-                "BUILD_CUGRAPH_OPS_PYTHON_LIBRARY ${PKG_BUILD_CUGRAPH_OPS_PYTHON_LIBRARY}"
     )
-
 endfunction()
 
 ###
-# Change pinned tag and fork here to test a commit in CI
 #
 # To use a locally-built cugraph-ops package, set the CMake variable
 # `-D cugraph-ops_ROOT=/path/to/cugraph-ops/build`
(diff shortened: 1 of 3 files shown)


rlratzel:
Since cugraph-ops is closed-source, can the community member somehow get messaging to direct them to install cugraph-ops prior to building cugraph? I think this will result in a failed github login message which could be confusing.

ChuckHastings:
> Since cugraph-ops is closed-source, can the community member somehow get messaging to direct them to install cugraph-ops prior to building cugraph? I think this will result in a failed github login message which could be confusing.

Following up on Rick's comment above, we certainly want to make the default behavior that an external community member can checkout `cugraph` source code, create a conda environment and build `cugraph` from source and have everything work smoothly.

If someone wants to also build from `cugraph-ops` source (and has the ability), there needs to be a way to support that.

It's not clear to me how this PR would change the build experience for a person that doesn't have access to the `cugraph-ops` source code when they try and build `cugraph`.  Can we explain that?

cpp/cmake/thirdparty/get_libcugraphops.cmake
```diff
@@ -14,26 +14,50 @@
...
-    )
+    set(oneValueArgs VERSION FORK PINNED_TAG CLONE_ON_PIN)
+    cmake_parse_arguments(PKG "" "${oneValueArgs}" "" ${ARGN} )
+
+    if(PKG_CLONE_ON_PIN AND NOT PKG_PINNED_TAG STREQUAL "branch-${CUGRAPH_BRANCH_VERSION_cugraph_ops}")
+        message("Pinned tag found: ${PKG_PINNED_TAG}. Cloning cugraph-ops locally.")
+        set(CPM_DOWNLOAD_cugraph-ops ON)
+    endif()
 
-    rapids_find_package(cugraphops
```
robertmaynard, line 31: We should keep this component so we can find versions of cugraph-ops that don't provide the cmake `cugraph-ops-config.cmake` file. Once that has been merged we can come back and only use `rapids_cpm_find`.

cpp/cmake/thirdparty/get_libcugraphops.cmake
```diff
@@ -14,26 +14,50 @@
...
-        REQUIRED
-        BUILD_EXPORT_SET    cugraph-exports
-        INSTALL_EXPORT_SET  cugraph-exports
+    rapids_cpm_find(cugraph-ops ${PKG_VERSION}
```
robertmaynard, line 30: What we might want to is provide more control over building cugraph-ops from source which would allow us to provide error messages and instructions as needed.

```cmake
option(CUGRAPH_BUILD_CUGRAPHOPS_FROM_SOURCE OFF)

if(NOT CUGRAPH_BUILD_CUGRAPHOPS_FROM_SOURCE)
  rapids_find_generate_module(cugraph-ops)
  rapids_find_package(cugraph-ops)
  if(NOT TARGET cugraph-ops::cugraph-ops++)
    message(FATAL_ERROR "The current enviornment you are building in doesn't have cugraph-ops. If using conda please read <X.Y> to properly configure the enviornment. If you are building from source and have access to the cugraph-ops repositry please enable `CUGRAPH_BUILD_CUGRAPHOPS_FROM_SOURCE`)
  endif()
endif()

if(CUGRAPH_BUILD_CUGRAPHOPS_FROM_SOURCE AND
   NOT TARGET cugraph-ops::cugraph-ops++)
    rapids_cpm_find(cugraph-ops ${PKG_VERSION} .... )
endif()
```

trxcllnt: As I mentioned in slack, I worry this would be introducing code that is effectively never tested -- neither CI nor devs who primarily use conda will set `-D CUGRAPH_BUILD_CUGRAPHOPS_FROM_SOURCE=ON`

trxcllnt:
@rlratzel I updated the logic to use `rapids_find_package` instead of `rapids_cpm_find`. This won't provide an option to download `cugraph-ops` from GH, and instead will fail if cugraph-ops is not installed or the package root provided via `-D cugraph-ops_ROOT`.

Not using `rapids_cpm_find` also means devs can't use `-D CPM_cugraph-ops_SOURCE=` to provide a source directory that CPM should attempt to build, but I assume that's fine? If devs have access to `cugraph-ops` and want to use a local clone, they can build it first and pass the build dir via `-D cugraph-ops_ROOT`.

I'm still partial to using `rapids_cpm_find` only, because it would provide us with more options both locally (`-D CPM_cugraph-ops_SOURCE=`) and in CI (we could add the option for PRs to change the `cugraph-ops` branch and CI would clone and build that branch from source).

We could try to search for it first via `rapids_find_package` and fall back to `rapids_cpm_find` -- but in the case of an external contributor who can't clone `rapidsai/cugraph-ops`, that will still fail with the error of not having correct GH access rights, which AFAIK is the error you want to avoid.

The state of cugraph to me seems to be this:
1. External contributors must retrieve `cugraph-ops` (install via conda, `.tgz` eventually)
2. CMake must be able to find `cugraph-ops` (either implicitly or explicitly)

If neither of the above conditions are met, an external contributor building from source will fail to configure. If they cannot install `cugraph-ops` via conda (and avoiding conda is presumably why they are building `cugraph` from source), and there's no other way to get the `cugraph-ops` binaries today, what difference does the error message make? The end result is they cannot configure `cugraph` and have no recourse to resolve it.

trxcllnt:
rerun tests

trxcllnt:
rerun tests

rlratzel (approved):

zbjornson:
Just wanted to chime in to add a voice of support for a non-conda option here. We can't use conda in our computing environment, so we manually download and unpack libcugraphops (and libcumlprims) from anaconda.

BradReesWork:
@gpucibot merge

PR #2150, opened 2022-03-23
HITS C API implementation

### LINKED ISSUE #2092: [FEA] Expose HITS through the C API
Create a C API for the HITS algorithm.

ChuckHastings:
Closes #2092 

Provide the C API implementation for HITS.  Adds a unit tester for MG, updates the unit tests for both SG and MG.

### DIFF
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -93,5 +93,29 @@ rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partition_id(
   std::optional<rmm::device_uvector<weight_t>>& d_edgelist_weights,
   bool groupby_and_count_local_partition_by_minor = false);
 
+/**
+ * @brief Collect vertex values (represented as k/v pairs across cluster) and update the
+ *        local value arrays on the GPU responsible for each vertex.
+ *
+ * Data will be shuffled and d_local_values[d_vertices[i]] = d_values[i]
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam value_t  Type of value associated with the vertex.
+ * @tparam bool     multi_gpu flag
+ *
+ * @param[in] handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator,
+ * and handles to various CUDA libraries) to run graph algorithms.
+ * @param[in/out] d_vertices Vertex IDs for the k/v pair
+ * @param[in/out] d_values Values for the k/v pair
+ * @param[out] d_local_values The device vector on each GPU that should be updated
+ * @param[in] local_vertex_first The first vertex id assigned to the local GPU
+ */
+template <typename vertex_t, typename value_t, bool multi_gpu>
+void collect_vertex_values_to_local(raft::handle_t const& handle,
+                                    rmm::device_uvector<vertex_t>& d_vertices,
+                                    rmm::device_uvector<value_t>& d_values,
+                                    rmm::device_uvector<value_t>& d_local_values,
+                                    vertex_t local_vertex_first);
+
 }  // namespace detail
 }  // namespace cugraph
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -94,10 +94,11 @@ rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partition_id(
   bool groupby_and_count_local_partition_by_minor = false);
 
 /**
- * @brief Collect vertex values (represented as k/v pairs across cluster) and update the
+ * @brief Collect vertex values (represented as k/v pairs across cluster) and return
  *        local value arrays on the GPU responsible for each vertex.
  *
- * Data will be shuffled and d_local_values[d_vertices[i]] = d_values[i]
+ * Data will be shuffled, renumbered and initialized with the default value,
+ * then:  return_array[d_vertices[i]] = d_values[i]
  *
  * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
  * @tparam value_t  Type of value associated with the vertex.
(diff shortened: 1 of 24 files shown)


BradReesWork (approved):

cpp/src/detail/shuffle_wrappers.cu
```diff
@@ -335,5 +335,103 @@ template rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partiti
...
+    rmm::device_uvector<vertex_t> d_rx_vertices(0, handle.get_stream());
+    rmm::device_uvector<value_t> d_rx_values(0, handle.get_stream());
+
+    std::tie(d_rx_vertices, d_rx_values, std::ignore) =
+      cugraph::groupby_gpu_id_and_shuffle_kv_pairs(
+        comm,
+        d_vertices.begin(),
+        d_vertices.end(),
+        d_values.begin(),
+        [key_func = cugraph::detail::compute_gpu_id_from_vertex_t<vertex_t>{comm_size}] __device__(
```
seunghwak, line 358: Have you checked with initial guess hub values? If I am not mistaken, you're calling this after renumbering, but `cugrpah::detail::compute_gpu_id_from_vertex_t` expects original (before renumbering) vertex IDs.

cpp/src/c_api/hits.cpp
```diff
@@ -113,40 +114,46 @@ struct hits_functor : public cugraph::c_api::abstract_functor {
...
+        cugraph::renumber_ext_vertices<vertex_t, multi_gpu>(handle_,
+                                                            guess_vertices.data(),
+                                                            guess_vertices.size(),
+                                                            number_map->data(),
+                                                            graph_view.get_local_vertex_first(),
+                                                            graph_view.get_local_vertex_last(),
+                                                            do_expensive_check_);
+
+        cugraph::detail::collect_vertex_values_to_local<vertex_t, weight_t, multi_gpu>(
+          handle_, guess_vertices, guess_values, hubs, graph_view.get_local_vertex_first());
```
seunghwak, line 141: If I am not mistaken, this code will not work as intended and looks fragile.

First, `hubs` is not initialized, so if user provides `guess_vertices` only for a subset of vertices, hubs can hold garbage values. If `guess_vertices` has duplicates, I am concerned potential concurrent writes to the same location can happen, and not sure this is not an undefined behavior. But I don't see these requirements in the documentation of `collect_vertex_values_to_local`.

Second, if I am not mistaken, `collect_vertex_values_to_local` may not work as intended with renumbered vertices.

Third, we may update this function to take `guess_vertices` & `guess_values` as R-value input parameters and return `hubs`; I think this is more functional and easier to understand.

And we may be able to re-implement this in a bit different way.

Say, first shuffle guess_vertices & guess_values kv pairs to target GPUs; renumber (and here, we can renumber just using local `number_map`); then 0 initialize hubs then scatter. I think this will be more efficient.

ChuckHastings: Yes, I think this would be more efficient.  I will refactor this as you suggested.

I will add a test that specifies initial values so that we can verify that it works on the conditions you mention.

cpp/src/detail/shuffle_wrappers.cu
ChuckHastings: adding a test in the next push.  Refactored this code as you suggested.

cpp/src/c_api/hits.cpp
seunghwak: I don't see these updates yet (if I am not mistaken), so is this PR still targeting 22.04?

ChuckHastings: Just pushed these changes about an hour ago.

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -93,5 +93,39 @@ rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partition_id(
...
+template <typename vertex_t, typename value_t, bool multi_gpu>
+rmm::device_uvector<value_t> collect_renumber_ext_vertex_values_to_local(
+  raft::handle_t const& handle,
+  rmm::device_uvector<vertex_t>&& d_vertices,
+  rmm::device_uvector<value_t>&& d_values,
+  rmm::device_uvector<vertex_t> const& number_map,
+  vertex_t local_vertex_first,
+  vertex_t local_vertex_last,
+  value_t default_value,
+  bool do_expensive_check);
```
seunghwak, line 128: Do we still need this function?

cpp/src/detail/shuffle_wrappers.cu
```diff
@@ -335,5 +336,145 @@ template rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partiti
...
+rmm::device_uvector<value_t> collect_renumber_ext_vertex_values_to_local(
+  raft::handle_t const& handle,
+  rmm::device_uvector<vertex_t>&& d_vertices,
+  rmm::device_uvector<value_t>&& d_values,
+  rmm::device_uvector<vertex_t> const& number_map,
+  vertex_t local_vertex_first,
+  vertex_t local_vertex_last,
+  value_t default_value,
+  bool do_expensive_check)
+{
```
seunghwak, line 349: We may no longer need this function?

cpp/include/cugraph/detail/shuffle_wrappers.hpp
ChuckHastings: We are using this function, I think it addresses what you are suggesting needs to be done.
`collect_renumber_ext_vertex_values_to_local` does:
* shuffle to get ext vertex values to the proper GPU
* local renumber of ext vertices
* scatter to populate our dense vertex property arrays

If I can rename the function to make it more clear that these steps are what is done, I'm open to a better name.

seunghwak: Gotcha, so this function gets actually updated to implement the new logic.

seunghwak (approved):

cpp/include/cugraph/detail/shuffle_wrappers.hpp
ChuckHastings: Sounds good.  I'll update the function name.

ChuckHastings:
@gpucibot merge

PR #2176, opened 2022-03-30
Pin cmake in conda recipe to <3.23

dantegd:
CMake 3.23 has a bug that breaks our conda-build based builds in CI, this avoids that issue.


ajschmidt8 (approved):

BradReesWork:
@gpucibot merge

PR #2179, opened 2022-03-30
[FEA] Enable building static libs

trxcllnt:
This PR allows building libcugraph static libs via CMake option `-DBUILD_SHARED_LIBS=ON|OFF`.

I was seeing a linker error due to not having `-fPIC` enabled, but I wouldn't have expected this to affect static libs. I'll investigate some more and turn it off if possible, but for now `-fPIC` is enabled.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -247,6 +247,7 @@ set_target_properties(cugraph
                CXX_STANDARD_REQUIRED               ON
                CUDA_STANDARD                       17
                CUDA_STANDARD_REQUIRED              ON
+               POSITION_INDEPENDENT_CODE           ON
                INTERFACE_POSITION_INDEPENDENT_CODE ON
 )
 


robertmaynard:
> I was seeing a linker error due to not having `-fPIC` enabled, but I wouldn't have expected this to affect static libs. I'll investigate some more and turn it off if possible, but for now `-fPIC` is enabled.

All static code that will be embedded in to a shared library needs to be compiled with fPIC. Code that will be embedded into a binary that isn't compiled with fPIE has no such requirement ( statement presumes x86_64 arch ).

robertmaynard (approved):
LGTM

trxcllnt:
rerun tests

trxcllnt:
rerun tests

trxcllnt:
@rlratzel Right on, thanks for noticing. Pushed a commit with changes for `libcugraph_etl` as well.

rlratzel (approved):

BradReesWork:
@gpucibot merge

Issue #2146, opened 2022-03-22
[ENH] Eigenvector Centrality


BradReesWork:
Create a new Eigenvector Centrality function

- [X] C++
- [X] C
- [ ] pylibcugraph
- [ ] cugraph

PR #2180, opened 2022-03-30
Define C API for eigenvector centrality

ChuckHastings:
This PR defines the C API for eigenvector centrality.  It is the first of a series of PRs to address #2146

It also does the following:
* Renames `cugraph_pagerank_result_t` to `cugraph_centrality_result_t` to allow it to be reused for all centrality algorithms
* Reorganizes the code for the results into a separate file
* Provides an initial implementation of eigenvector centrality that can be used for testing the API.  NOTE: It literally calls pagerank with some hardcoded parameters.  The return formats for the algorithms are the same, so this will mimic the API behavior (although the scores will be different once eigenvector centrality is implemented
* Provided a C unit test that validates the results (will need to be updated to validate eigenvector centrality results once the real algorithm is in place)

Note that this is a breaking change since it modifies the type of the pagerank result.

### DIFF
--- a/cpp/include/cugraph_c/algorithms.h
+++ b/cpp/include/cugraph_c/algorithms.h
@@ -25,36 +25,36 @@ extern "C" {
 #endif
 
 /**
- * @brief     Opaque pagerank result type
+ * @brief     Opaque centrality result type
  */
 typedef struct {
   int32_t align_;
-} cugraph_pagerank_result_t;
+} cugraph_centrality_result_t;
 
 /**
- * @brief     Get the vertex ids from the pagerank result
+ * @brief     Get the vertex ids from the centrality result
  *
- * @param [in]   result   The result from pagerank
+ * @param [in]   result   The result from a centrality algorithm
  * @return type erased array of vertex ids
  */
-cugraph_type_erased_device_array_view_t* cugraph_pagerank_result_get_vertices(
-  cugraph_pagerank_result_t* result);
+cugraph_type_erased_device_array_view_t* cugraph_centrality_result_get_vertices(
+  cugraph_centrality_result_t* result);
 
 /**
- * @brief     Get the pagerank values from the pagerank result
+ * @brief     Get the centrality values from a centrality algorithm result
  *
- * @param [in]   result   The result from pagerank
- * @return type erased array of pagerank values
+ * @param [in]   result   The result from a centrality algorithm
+ * @return type erased array view of centrality values
  */
-cugraph_type_erased_device_array_view_t* cugraph_pagerank_result_get_pageranks(
-  cugraph_pagerank_result_t* result);
+cugraph_type_erased_device_array_view_t* cugraph_centrality_result_get_centralities(
+  cugraph_centrality_result_t* result);
 
 /**
- * @brief     Free pagerank result
+ * @brief     Free centrality result
  *
- * @param [in]   result   The result from pagerank
+ * @param [in]   result   The result from a centrality algorithm
  */
-void cugraph_pagerank_result_free(cugraph_pagerank_result_t* result);
+void cugraph_centrality_result_free(cugraph_centrality_result_t* result);
 
 /**
  * @brief     Compute pagerank
@@ -90,7 +90,7 @@ cugraph_error_code_t cugraph_pagerank(
   size_t max_iterations,
   bool_t has_initial_guess,
   bool_t do_expensive_check,
-  cugraph_pagerank_result_t** result,
+  cugraph_centrality_result_t** result,
   cugraph_error_t** error);
 
 /**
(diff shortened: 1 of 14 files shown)


cpp/src/c_api/centrality_result.hpp
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+#include <c_api/array.hpp>
+
+namespace cugraph {
+namespace c_api {
+
+struct cugraph_centrality_result_t {
+  cugraph_type_erased_device_array_t* vertex_ids_;
+  cugraph_type_erased_device_array_t* centralities_;
```
seunghwak, line 26: In line with the previous comment, centrality_values_ (or just values_ as centrality appears in the struct name) may sound a bit less ambiguous?

cpp/src/c_api/eigenvector_centrality.cpp
```diff
@@ -0,0 +1,125 @@
...
+#include <optional>
+
+namespace {
+
+struct eigenvector_centrality_functor : public cugraph::c_api::abstract_functor {
+  raft::handle_t const& handle_;
+  cugraph::c_api::cugraph_graph_t* graph_;
+  double epsilon_;
+  size_t max_iterations_;
+  bool do_expensive_check_;
```
seunghwak, line 38: Better add `{}` to avoid holding garbage values?

cpp/src/c_api/centrality_result.hpp
seunghwak: And to add '{}' to avoid holding garbage values.

cpp/src/c_api/eigenvector_centrality.cpp
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  template <typename vertex_t,
+            typename edge_t,
+            typename weight_t,
+            bool store_transposed,
+            bool multi_gpu>
+  void operator()()
+  {
+    // FIXME: Think about how to handle SG vice MG
```
seunghwak, line 62: Should we do anything different? (compared to PageRank & Katz centrality?)

cpp/src/c_api/pagerank.cpp
```diff
@@ -27,33 +28,28 @@
...
   raft::handle_t const& handle_;
-  cugraph_graph_t* graph_;
-  cugraph_type_erased_device_array_view_t const* precomputed_vertex_out_weight_sums_;
-  cugraph_type_erased_device_array_view_t* personalization_vertices_;
-  cugraph_type_erased_device_array_view_t const* personalization_values_;
+  cugraph::c_api::cugraph_graph_t* graph_;
+  cugraph::c_api::cugraph_type_erased_device_array_view_t const*
+    precomputed_vertex_out_weight_sums_;
+  cugraph::c_api::cugraph_type_erased_device_array_view_t* personalization_vertices_;
+  cugraph::c_api::cugraph_type_erased_device_array_view_t const* personalization_values_;
```
seunghwak, line 39: Add `{}`

seunghwak (commented):
Looks good to me except for few nitpicking comments.

cpp/include/cugraph_c/algorithms.h
```diff
@@ -25,36 +25,36 @@ extern "C" {
...
+ * @brief     Get the centrality values from a centrality algorithm result
  *
- * @param [in]   result   The result from pagerank
- * @return type erased array of pagerank values
+ * @param [in]   result   The result from a centrality algorithm
+ * @return type erased array view of centrality values
  */
-cugraph_type_erased_device_array_view_t* cugraph_pagerank_result_get_pageranks(
-  cugraph_pagerank_result_t* result);
+cugraph_type_erased_device_array_view_t* cugraph_centrality_result_get_centralities(
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

cpp/src/c_api/eigenvector_centrality.cpp
ChuckHastings: Fixed in next push

cpp/src/c_api/eigenvector_centrality.cpp
ChuckHastings: Regarding the FIXME?

I should probably create an issue and delete all of the FIXMEs that are propagating here.  The issue is to come up with a way to have an SG only version of cugraph.  Some customers don't need an MG version and an SG only package of cugraph would be lighter weight for them to install.  But this aspect of the C API implementation would need to be addressed somehow.

However it is addressed, it would be addressed the same in each API.

cpp/src/c_api/pagerank.cpp
ChuckHastings: Done in next push

seunghwak (approved):

rlratzel (approved):
Just some minor observations which need not hold up approval.

cpp/src/c_api/centrality_result.cpp
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
ChuckHastings: Yes.  Pushed update.

cpp/src/c_api/eigenvector_centrality.cpp
```diff
@@ -0,0 +1,125 @@
+/*
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
ChuckHastings: Yes.  Pushed update.

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2187, opened 2022-04-04
Update graph/graph primitives API to consistently use vertex/edge centric terminologies instead of matrix centric terminolgies

seunghwak:
Partially address https://github.com/rapidsai/cugraph/issues/2003

cuGrpah public API has been mixing vertex/edge-centric terminologies and vector/matrix-centric terminologies.

- matrix_partition => edge_partition
- row => source (src), column (col) => destination (dst)
- rename `get_property_name()` functions to `property_name()` for brevity and better follow naming conventions for C++ standard library
- other updates to better group/clarify vertex/edge partition related functions.


ChuckHastings (approved):

BradReesWork:
@gpucibot merge

Issue #1759, opened 2021-08-05
[ENH] SG python implementation of `katz_centrality` still uses legacy.


ChuckHastings:
We should migrate to using the new primitive based version of `katz_centrality` so that we can delete the legacy implementation.

PR #2185, opened 2022-04-01
Labeling algorithm updates for C API

ChuckHastings:
Several things bundled here:
 * Added C API definition, implementation and tests for weakly connected components
 * Added C API definition for strongly connected components
 * Updated the C API test functions for supporting the `is_symmetric` property
 * Added C unit tests for SCC, and an implementation that returns `NOT_IMPLEMENTED`

I could add some mock data returns, although the output from SCC is the same as the output from WCC.  So python testing could just call WCC instead until there is an SCC implementation.

PR #2192, opened 2022-04-05
Add Katz Centrality to the C API

ChuckHastings:
Add Katz Centrality to the C API.  This will allow the work for #1759 to be done using the C API which should make things simpler and avoid doing the work twice.

This PR both defines and implements the Katz Centrality algorithm through the C API.


cpp/include/cugraph_c/centrality_algorithms.h
```diff
@@ -0,0 +1,306 @@
...
+  cugraph_centrality_result_t* result);
+
+/**
+ * @brief     Get the centrality values from a centrality algorithm result
+ *
+ * @param [in]   result   The result from a centrality algorithm
+ * @return type erased array view of centrality values
+ */
+cugraph_type_erased_device_array_view_t* cugraph_centrality_result_get_values(
+  cugraph_centrality_result_t* result);
```
seunghwak, line 51: Isn't this an update happened in one of the previous PRs? It seems like this PR includes updates for PageRank, HITS, and so on in addition to the updates for Katz. Is this because this PR is showing updates from other PRs or these are indeed new updates in this PR?

ChuckHastings: I felt like the `algorithms.h` file was getting too large to keep track of.  This PR (as well as #2185) starts moving like algorithms into a separate file.  In this case I'm moving all of the centrality based algorithms into `centrality_algorithms.h`.

So the definition of this function was in an earlier PR, this is just moving it to a new file.

seunghwak: Gotcha, makes sense. Not for this PR, but to be consistent, PageRank is currently under `link_analysis`. Maybe in a separate PR, shouldn't we better move PageRank to the `centrality` directory? (in C/C++/Python).

seunghwak: And HITS as well (https://github.com/rapidsai/cugraph/tree/branch-22.06/cpp/src/link_analysis)

seunghwak: Yeah...and we also should start thinking about splitting algorithms.hpp to put the C & C++ parts in sync.

seunghwak (approved):
LGTM.

cpp/include/cugraph_c/centrality_algorithms.h
betochimas: Unlike the previous PR where you defined the centrality result types in both C and pylib, here you're splitting up `algorithms.h` into itself and `centrality_algorithms.h`. I think it would be better and easier to follow along if I do the same in my pylibcugraph wrapper PR for `algorithms.pxd` and `centrality_algorithms.pxd`. It would likely prevent a merge conflict and I can keep pulling from your branch whenever changes are being made

ChuckHastings: @betochimas - that makes sense.  Did you want me to split that in this PR, or are you going to handle it in #2201?

betochimas: @ChuckHastings I can handle it in #2201

rlratzel: yep, I think @betochimas is saying this too: we've been keeping the .pxd files in `pylibcugraph/_cugraph_c` in sync with the C API headers to make it easy to follow, so we'd want the same reorg in pylibcugraph. I think it can be done in #2201.

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2216, opened 2022-04-11
Enable MG support for small datasets

### LINKED ISSUE #2196: [BUG] Failure when running MG algos on very small datasets
Our MG implementation assumes that each workers has a partition, which is unlikely to occur for very small datasets and a large number of GPUs

Run MG Neighborhood sampling with 3+ GPUs on the `small_tree.csv` datasets

jnke2016:
This PR enables MG support for very small datasets where the number of partitions is smaller than the number of workers

Dependent on issue #2217 to be resolved before merging
closes #2196


rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #2219, opened 2022-04-12
Fix error from two conflicting merges

ChuckHastings:
I added a new algorithm while Seunghwa modified a graph_view method name.  The two PRs passed and merged but introduced this error.


seunghwak (approved):

ChuckHastings:
@gpucibot merge

PR #2220, opened 2022-04-12
Graph primitives API updates

seunghwak:
Partially address #2003.

1. Renumber VertexFrontier & SortedUniqueKeyBucket to vertex_frontier_t & sorted_unique_key_bucket_t to be consistent with the naming schemes for graph_t & graph_view_t.
2. vertex_frontier_t to take number of buckets as an input parameter (instead of non-type template parameter) and rename `get_bucket` to `bucket`.
3. Use `constexpr size_t` instead of `enum calss` for bucket indices to avoid unnecessary type casting.
4. Update `fill()` of `edge_partition_src|dst_property_t` to take `handle` instead of `stream` to be consistent with other member functions (e.g. `clear()`)
5. Remove `..._v` primitives that working on a subset of local vertices.
6. Update `v_op` to take vertex ID (to be consistent with `e_op` which takes source & destination IDs).
7. Other miscellaneous clean-ups.


ChuckHastings (approved):

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2224, opened 2022-04-13
Address some MNMG issues in cython.cu

### LINKED ISSUE #2217: [BUG] Failure at renumbering when passing empty dataframe
Some partitions have their column values set to `<NA>` after shuffling when passing empty dataframe as described below.

```
    major_vertices  minor_vertices value
0               3               5  <NA>
1               2               0  <NA>
2               2               1  <NA>
3               4               5  <NA>
```
There is also an OOM/hang reported when calling renumbering after shuffling. There is no hang/error when the datasets is unweighted

Build PR #2216 and run any MG algos with the following datasets on 4+ GPUs

```
"src" = [0, 1, 1, 2, 2, 2, 3, 4, 5]
"dst" = [1, 3, 4, 0, 1, 3, 5, 5, 6]
"value" = [0.1, 2.1, 1.1, 5.1, 3.1, 4.1, 7.2, 3.2, 1.2]
```

ChuckHastings:
Closes #2217 

The issue identifies some issues if the partitioning ends up with some empty partitions.  There were two issues here:
 1. The `call_shuffle` method determines whether a graph has weights by comparing the weight pointer to `nullptr`.  But if an MNMG partition is empty then the weight pointer will be `nullptr` even if the graph has weights.  Added a parameter to identify if the graph is weighted.
 2. The `renumber_helper` method was missing an else and was overwriting the correct value with an invalid value if the return from shuffling on a gpu had no entries.

These are corrected in the PR.

### DIFF
--- a/cpp/include/cugraph/utilities/cython.hpp
+++ b/cpp/include/cugraph/utilities/cython.hpp
@@ -594,9 +594,10 @@ std::unique_ptr<major_minor_weights_t<vertex_t, edge_t, weight_t>> call_shuffle(
     edgelist_major_vertices,  // [IN / OUT]: groupby_gpu_id_and_shuffle_values() sorts in-place
   vertex_t* edgelist_minor_vertices,  // [IN / OUT]
   weight_t* edgelist_weights,         // [IN / OUT]
-  edge_t num_edgelist_edges);
+  edge_t num_edgelist_edges,
+  bool is_weighted);
 
-// Wrapper for calling renumber_edeglist() inplace:
+// Wrapper for calling renumber_edgelist() inplace:
 //
 template <typename vertex_t, typename edge_t>
 std::unique_ptr<renum_tuple_t<vertex_t, edge_t>> call_renumber(
--- a/python/cugraph/cugraph/structure/renumber_wrapper.pyx
+++ b/python/cugraph/cugraph/structure/renumber_wrapper.pyx
@@ -54,10 +54,12 @@ cdef renumber_helper(shuffled_vertices_t* ptr_maj_min_w, vertex_t, weights):
     # which is not supported by '__cuda_array_interface__'
     if shuffled_major_series is None:
         shuffled_major_series = cudf.Series(dtype=vertex_t)
+    else:
+        shuffled_df['major_vertices']= shuffled_major_series
     if shuffled_minor_series is None:
         shuffled_minor_series = cudf.Series(dtype=vertex_t)
-    shuffled_df['major_vertices']= shuffled_major_series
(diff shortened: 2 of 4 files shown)


cpp/include/cugraph/utilities/cython.hpp
```diff
@@ -594,9 +594,10 @@ std::unique_ptr<major_minor_weights_t<vertex_t, edge_t, weight_t>> call_shuffle(
     edgelist_major_vertices,  // [IN / OUT]: groupby_gpu_id_and_shuffle_values() sorts in-place
   vertex_t* edgelist_minor_vertices,  // [IN / OUT]
   weight_t* edgelist_weights,         // [IN / OUT]
-  edge_t num_edgelist_edges);
+  edge_t num_edgelist_edges,
+  bool is_weighted);
```
seunghwak, line 598: Just FYI, we're using std::optional elsewhere to address this instead of adding an additional flag. Not sure it is worth the effort in applying the same strategy here as this code will become obsolete once we switch to the C-API based approach.

seunghwak: And yeah... this is cython, so we may not be able to use std optional here.

seunghwak (approved):

jnke2016 (approved):

python/cugraph/cugraph/structure/renumber_wrapper.pyx
```diff
@@ -65,7 +65,8 @@ cdef renumber_helper(shuffled_vertices_t* ptr_maj_min_w, vertex_t, weights):
             move(pair_s_weights.first), weight_t, "shuffled_weights")
         if shuffled_weights_series is None:
             shuffled_df['value']= cudf.Series(dtype=weight_t)
-        shuffled_df['value']= shuffled_weights_series
+        else:
+            shuffled_df['value']= shuffled_weights_series
```
jnke2016, line 69: I believe we should do the same for `shuffled_minor_series` and `shuffled_major_series`

jnke2016 (changes requested):
I left a comment regarding adding the else statement for both shuffle_major and shuffle_minor too

python/cugraph/cugraph/structure/renumber_wrapper.pyx
ChuckHastings: Good catch.  Wasn't causing an error, but the code is, in fact, incorrect.  Updated in the last push.

jnke2016 (approved):

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2230, opened 2022-04-15
Use conda to build python packages during GPU tests

Ethyling:
This PR convert the `from sources` build we are doing in GPU test job to a `conda build`. This is done for the following reasons:
- This is required step to improve the Ops CI/CD setup to a more convenient pipeline
- This is required to start using `conda` compilers and `mamba` to build RAPIDS packages
- This prevent us from manually managing and installing the dependencies in GPU job
- This ensure the packages can be installed
- This ensure the tests are running and working against the package content and not the build results. Currently the Python packages are not tested.

This may increase the global pipeline time, but the usage of `mamba` should resolve this as `mamba` is faster than `conda` to build packages


Ethyling:
rerun tests

Ethyling:
rerun tests

ajschmidt8 (approved):
Approving `ops-codeowner` file changes.

python/cugraph/cugraph/tests/mg/test_mg_batch_betweenness_centrality.py
```diff
@@ -12,15 +12,18 @@
...
 import numpy as np
 
 from cugraph.dask.common.mg_utils import is_single_gpu
-from cugraph.tests.utils import RAPIDS_DATASET_ROOT_DIR_PATH
+from cugraph.testing.utils import RAPIDS_DATASET_ROOT_DIR_PATH
 
 # Get parameters from standard betwenness_centrality_test
-from cugraph.tests.test_betweenness_centrality import (
+# As tests directory is not a module, we need to add it to the path
+sys.path.insert(0, '../')
```
rlratzel, line 25: A few things:
* this does not add the tests directory to the import path, it adds `..`, which will be interpreted as the parent directory of the CWD from where pytest is run.
* pytest automatically adds the directories containing the collected test modules to the import path, so this may be "accidentally" working because of that
* these changes are not being tested by CI since our test script skips everything in the `mg` directory, so have you manually tested these changes on a multi-GPU machine to ensure they work?
* needing to modify sys.path directly could be a symptom that we have a broken package hierarchy or some other code smell - is there a reason this directory is no longer a package (ie. had its `__init__.py` file removed)?

python/cugraph/cugraph/tests/mg/test_mg_batch_edge_betweenness_centrality.py
```diff
@@ -12,24 +12,27 @@
...
 
 from cugraph.dask.common.mg_utils import is_single_gpu
 
-from cugraph.tests.utils import RAPIDS_DATASET_ROOT_DIR_PATH
+from cugraph.testing.utils import RAPIDS_DATASET_ROOT_DIR_PATH
 
 # Get parameters from standard betwenness_centrality_test
-from cugraph.tests.test_edge_betweenness_centrality import (
+# As tests directory is not a module, we need to add it to the path
+sys.path.insert(0, '.')
```
rlratzel, line 26: (same feedback as above)

rlratzel (changes requested):
Most of the changes look okay, but I did see a few that could cause problems which I commented on below.  I also don't understand why some of these changes are needed for allowing the GPU tests to use a conda install instead of a from-source build (eg. why can't we continue to have a `cugraph.tests` package? Why do we need a separate `cugraph.testing` package?, etc.).  Maybe we can chat offline about it and then update the PR or report back here, if necessary?

python/cugraph/cugraph/tests/mg/test_mg_batch_betweenness_centrality.py
rlratzel: @Ethyling based on [our conversation](https://github.com/rapidsai/cugraph/pull/2230#issuecomment-1123923864), would you mind removing the `sys.path.insert()` calls and adding a `FIXME` stating that those tests need to be refactored to not import other tests and should instead put shared test code in a util module? Same for the other comment below.

rlratzel (approved):

Ethyling:
@gpucibot merge

PR #2234, opened 2022-04-19
Rename primitive functions.

seunghwak:
Partially address https://github.com/rapidsai/cugraph/issues/2003
Rename
- copy_v_transform_reduce_in|out_nbr => transform_reduce_incoming|outgoing_e_of_v
- copy_v_transform_reduce_key_aggregated_out_nbr => transform_reduce_key_aggregated_outgoing_e_of_v
- update_frontier_v_push_if_out_nbr => update_frontier_from_outgoing_e_of_v


ChuckHastings:
rerun tests

ChuckHastings (approved):

cpp/include/cugraph/graph_view.hpp
```diff
@@ -59,10 +59,10 @@ class serializer_t;  // forward...
...
  * To be more specific, a GPU with (col_comm_rank, row_comm_rank) will be responsible for
  * col_comm_size rectangular partitions [a_i,b_i) by [c,d) where a_i =
- * vertex_partition_offsets[row_comm_size * i + row_comm_rank] and b_i =
- * vertex_partition_offsets[row_comm_size * i + row_comm_rank + 1]. c is
- * vertex_partition_offsets[row_comm_size * col_comm_rank] and d =
- * vertex_partition_offsests[row_comm_size * (col_comm_rank + 1)].
+ * vertex_partition_range_offsets[row_comm_size * i + row_comm_rank] and b_i =
+ * vertex_partition_range_offsets[row_comm_size * i + row_comm_rank + 1]. c is
+ * vertex_partition_range_offsets[row_comm_size * col_comm_rank] and d =
+ * vertex_partition_range_offsets[row_comm_size * (col_comm_rank + 1)].
```
kaatish, line 65: Could we rename this to specify that it is a range of vertex ids?
vertex_id_ranges perhaps?

cpp/include/cugraph/prims/transform_reduce_incoming_outgoing_e_of_v.cuh
```diff
@@ -457,22 +456,23 @@ __global__ void copy_v_transform_reduce_nbr_high_degree(
...
           typename T,
           typename VertexValueOutputIterator>
-void copy_v_transform_reduce_nbr(raft::handle_t const& handle,
-                                 GraphViewType const& graph_view,
-                                 EdgePartitionSrcValueInputWrapper edge_partition_src_value_input,
-                                 EdgePartitionDstValueInputWrapper edge_partition_dst_value_input,
-                                 EdgeOp e_op,
-                                 T init,
-                                 VertexValueOutputIterator vertex_value_output_first)
+void transform_reduce_e_of_v(raft::handle_t const& handle,
```
kaatish, line 467: We could just use `transform_reduce_edges` here.

cpp/include/cugraph/prims/transform_reduce_incoming_outgoing_e_of_v.cuh
```diff
@@ -985,7 +982,7 @@ template <typename GraphViewType,
           typename EdgeOp,
           typename T,
           typename VertexValueOutputIterator>
-void copy_v_transform_reduce_out_nbr(
+void transform_reduce_outgoing_e_of_v(
```
kaatish, line 985: `transform_reduce_outgoing_edges`?
The of_v part seems unnecessary to me.

cpp/include/cugraph/prims/transform_reduce_key_aggregated_outgoing_e_of_v.cuh
```diff
@@ -204,7 +202,7 @@ template <typename GraphViewType,
           typename ReduceOp,
           typename T,
           typename VertexValueOutputIterator>
-void copy_v_transform_reduce_key_aggregated_out_nbr(
+void transform_reduce_key_aggregated_outgoing_e_of_v(
```
kaatish, line 205: `transform_reduce_key_aggregated_outgoing_edges`?

cpp/src/link_analysis/hits_impl.cuh
```diff
@@ -105,7 +105,7 @@ std::tuple<result_t, size_t> hits(raft::handle_t const& handle,
   }
   for (size_t iter = 0; iter < max_iterations; ++iter) {
     // Update current destination authorities property
-    copy_v_transform_reduce_in_nbr(
+    transform_reduce_incoming_e_of_v(
```
kaatish, line 108: `transform_reduce_incoming_edges`?

cpp/src/traversal/bfs_impl.cuh
```diff
@@ -222,7 +222,7 @@ void bfs(raft::handle_t const& handle,
         e_op.prev_visited_flags = prev_visited_flags.data();
       }
 
-      update_frontier_v_push_if_out_nbr(
+      update_frontier_from_outgoing_e_of_v(
```
kaatish, line 225: If we are using the sources/destinations nomenclature then this could be renamed to be
`update_frontier_from_destinations` because the frontier will only contain vertices that are destinations as opposed to sources.
Having said that `update_frontier_from_outgoing_edges` is fine too.

cpp/include/cugraph/prims/transform_reduce_key_aggregated_outgoing_e_of_v.cuh
seunghwak: See

https://github.com/rapidsai/cugraph/pull/2234#discussion_r858892190

cpp/src/link_analysis/hits_impl.cuh
seunghwak: See

https://github.com/rapidsai/cugraph/pull/2234#discussion_r858892190

cpp/include/cugraph/graph_view.hpp
seunghwak: To be more consistent with the rest, we may need to rename `vertex_partition_range` with `vertex_id_partition_range` but I feel like this is a bit too lengthy (I thought `vertex_partition` somewhat implies partitioning of vertex ID ranges but it seems like this implication is not clear enough).

I added additional comments below to make this point clearer. Do you think this is sufficient or we need to further discuss about renaming?

cpp/src/traversal/bfs_impl.cuh
seunghwak: What about `update_frontier_from_outgoing_e` to be more consistent with the rest? (we are using `e` to denote `edges` so I hope this abbreviation is not too confusing).

I initially used `e_of_v` instead of just `e` borrowing this convention from other primitives performing per vertex reduction, but here, we are not performing per vertex reduction. Starting from an input frontier, iterate over all the outgoing edges from the vertices in the frontier, then, compute a new frontier by filtering out destination vertices of the outgoing edges.

Something like `compute_new_frontier_from_outgoing_e_of_v_in_input_frontier` might be more descriptive (but too verbose) and I assume `update_frontier` part reasonably covers both `compute_new_frontier` and `of_v_in_input_frontier`.

cpp/include/cugraph/prims/transform_reduce_incoming_outgoing_e_of_v.cuh
kaatish: I would read transform_reduce_outgoing_e_of_v as 'transform reduce outgoing edges of vertices'.
The 'of vertices' doesn't really communicate that a per source reduction is being done. We should think of something that does make this clear.

We could name it exactly as transform_reduce_e to denote that every edge is being iterated on. The fact that this variant is doing per source reduction could be explained by differing function signature. This could be confusing to some users.

We could also use `transform_reduce_outgoing_edges_per_source` or '..._e_per_src'.

kaatish: Another point was that if we are talking in terms of edges, in my opinion we should use the term source/destination (src/dst) instead of vertex.
Perhaps I am overlooking some nuances when dealing with transpose (csc) graphs.

seunghwak: Agreed, I like the `per_src` idea. I will make this change.

cpp/src/traversal/bfs_impl.cuh
kaatish: > What about update_frontier_from_outgoing_e to be more consistent with the rest?

Yes, this works

cpp/include/cugraph/prims/transform_reduce_incoming_outgoing_e_of_v.cuh
kaatish: Will `per_src` still make sense even when the graph is transposed?

seunghwak: I assume so.

But `transform_reduce_outgoing_e_of_v` and `transform_reduce_e_per_src` may have slightly different mental models.

I initially hoped `outgoing_e_of_v` to be interpreted as "for every vertex, iterate over outgoing edges and reduce the edge functor outputs per vertex"; so this is more vertex centric.

"e_per_src" might be interpreted as iterate over every directed edge, apply a functor to each edge, than reduce functor outputs per source vertex ID; so this is more edge centric.

Both are pretty much doing the same thing, but nuances might be slightly different, and I feel like the latter is a bit clearer.

seunghwak: Whether the graph is transposed or not affects efficiency of the primitive implementation. Per source reduction can be more efficient with a source-major format while per destination reduction can be more efficient with a destination-major format (by avoiding atomics, and some primitives allow only a source-major format... this is more to avoid implementing something that will not be used in the foreseeable future).

seunghwak: My main concern about `per_src` nomenclature is about 0-(in|out)-degree vertices.

So these primitives are designed to serve `properties_of_v = some_initial_value + sum_over_incoming|outgoing_e (calculated per edge value)` (e.g. PageRank, Katz centrality, HITS, summing per vertex incoming/outgoing edge weights).

This doesn't align very well with the edge-centric mental model.

kaatish (approved):

seunghwak:
@ChuckHastings @kaatish Added "do not merge" label.

Still needs to figure out how to properly re-name `copy_v_transform_reduce_in|out_nbr` (or other two level primitives).

`transform_reduce_incoming|outgoing_e_of_v` may not be interpreted as per vertex reduction.
`transform_reduce_e_per_src|dst` may have an issue with 0 in|out-degree vertices.

I'm now thinking more about `for_each_v_transform_reduce_incoming|outgoing_e`... but still thinking about better names.

ChuckHastings:
@gpucibot merge

Issue #2226, opened 2022-04-14
Refactor libcugraph Neighborhood Sampling 


ChuckHastings:
The following steps are required (each a separate PR):
1. Update API as discussed in our meeting (remove batching, drop some of the return values)
2. Refactor the implementation to use more of the existing primitives and support both SG and MG

PR #2236, opened 2022-04-21
add API for the new uniform neighborhood sampling

ChuckHastings:
Added an API for the new uniform neighborhood sampling implementation.

Added the API with an experimental tag in its name so it won't break the current python code.  Plan is to have the new API replace the existing API during this release.

Plan for this release will be to have the edge weight be an edge index (passed by the calling python code), so the return "index" value will be the edge weight.  Down the road we will release this restriction and allow real edge weights.

Addresses part 1 of #2226

### DIFF
--- a/cpp/include/cugraph_c/algorithms.h
+++ b/cpp/include/cugraph_c/algorithms.h
@@ -307,11 +307,40 @@ typedef struct {
  *                           be populated if error code is not CUGRAPH_SUCCESS
  * @return error code
  */
+// FIXME: This older API will be phased out this release in favor of the experimental one below
 cugraph_error_code_t cugraph_uniform_neighbor_sample(
   const cugraph_resource_handle_t* handle,
   cugraph_graph_t* graph,
   const cugraph_type_erased_device_array_view_t* start,
   const cugraph_type_erased_device_array_view_t* start_label,
+  const cugraph_type_erased_host_array_view_t* fan_out,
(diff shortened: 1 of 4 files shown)


seunghwak:
@ChuckHastings So, in using the wedge weight as the edge index, are you planning to set weight_t to integer type? Or you plan to store integer in floating point format? I guess the latter will be less invasive (some cugraph code assumes edge weights to be floating point type).

seunghwak (approved):
LGTM.

ChuckHastings:
> @ChuckHastings So, in using the wedge weight as the edge index, are you planning to set weight_t to integer type? Or you plan to store integer in floating point format? I guess the latter will be less invasive (assuming there is no resolution issue for a very large number of edges).

I was imagining I would instantiate `graph_t<int32_t, int32_t, int32_t>`, `graph_t<int32_t, int64_t, int64_t>` and `graph_t<int64_t, int64_t, int64_t` and use them for use in only this algorithm.  But if the cost is too great in the implementation - since we do no computation using the weight values for this algorithm - we could use the float for int32 and the double for int64 and everything should work fine.  I'll work that out during implementation.

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2252, opened 2022-04-27
changed the nodes type in sampler

wangxiaoyunNV:
I changed the nodes type in the sampler in graphstore, now it works for pytorch tensor.


wangxiaoyunNV:
Changed it a another way.

PR #2254, opened 2022-04-28
Add MG Hits and MG Neighborhood_sampling to benchmarks

jnke2016:
This PR 

1. Adds MG Hits and MG Neighborhood sampling to our list of algos to benchmarks in the nightlies
2. Drops the legacy renumbering not needed in the `C/Pylibcugraph` path


python/cugraph/cugraph/dask/link_analysis/hits.py
```diff
@@ -154,9 +154,9 @@ def hits(input_graph, tol=1.0e-5, max_iter=100,  nstart=None, normalized=True):
 
     client = default_client()
 
-    # FIXME Still compute renumbering at this layer in case str
-    # vertex ID are passed
-    input_graph.compute_renumber_edge_list(transposed=False)
+    # FIXME: 'legacy_renum_only' will not trigger the C++ renumbering
```
rlratzel, line 157: Can you explain the FIXME?  Is the idea that we don't want to have to use this arg in the future?

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -440,7 +453,9 @@ def neighbors(self, n):
         ddf = self.edgelist.edgelist_df
         return ddf[ddf["src"] == n]["dst"].reset_index(drop=True)
 
-    def compute_renumber_edge_list(self, transposed=False):
+    def compute_renumber_edge_list(self,
+                                   transposed=False,
+                                   legacy_renum_only=False):
```
rlratzel, line 458: You should probably update the docstring since you added a new arg.

rlratzel (changes requested):

rlratzel (approved):

PR #2260, opened 2022-05-04
[FEA] Raft RNG updated API

MatthiasKohl:
This PR updates usage of RAFT RNG functionality throughout the cugraph code-base, including using the new RAFT RngState to pass state between cugraph and cugraph-ops.
Finally, I removed unnecessary includes from `raft/random` where applicable.

### DIFF
--- a/cpp/src/detail/utility_wrappers.cu
+++ b/cpp/src/detail/utility_wrappers.cu
@@ -31,7 +31,7 @@ void uniform_random_fill(rmm::cuda_stream_view const& stream_view,
                          value_t max_value,
                          uint64_t seed)
 {
-  raft::random::RngState rng_state(seed, raft::random::PhiloxGenerator);
+  raft::random::RngState rng_state(seed, raft::random::GeneratorType::GenPhilox);
   raft::random::uniform<value_t, size_t>(
     rng_state, d_value, size, min_value, max_value, stream_view.value());
 }


cpp/src/sampling/neighborhood.cu
```diff
@@ -20,21 +20,23 @@
...
+#include <raft/random/rng_state.hpp>
+
 namespace cugraph {
 
 template <typename graph_t>
 std::tuple<rmm::device_uvector<typename graph_t::edge_type>,
            rmm::device_uvector<typename graph_t::vertex_type>>
 sample_neighbors_adjacency_list(raft::handle_t const& handle,
-                                ops::gnn::graph::Rng& rng,
+                                raft::random::RngState& rng_state,
```
kaatish, line 31: You will have to make the corresponding changes to cpp/include/cugraph/algorithms.hpp file. The functions sample_neighbors* do not have a python binding yet so CI should not break because of this.
@rlratzel thoghts?

MatthiasKohl: Thanks for catching this. I've updated the public API in algorithms.
I don't see any uses in python or binding code, so I think this should be good enough for now.

rlratzel: > You will have to make the corresponding changes to cpp/include/cugraph/algorithms.hpp file. The functions sample_neighbors* do not have a python binding yet so CI should not break because of this. @rlratzel thoghts?

This should be okay since the python binding for neighborhood sampling will use the C API, which does not expose this function signature to the binding code.

ChuckHastings: I will merge this change (once it is merged into the code base) into my refactor branch for uniform neighborhood sampling.  I've been talking with Corey about wanting to make this change, so I'm ready.

ChuckHastings (approved):
Do we have an issue somewhere to track the issue we were seeing with the PC Generator (the default for raft)?

MatthiasKohl:
I'll open an issue

MatthiasKohl:
I've opened issue #2266 for this and the tests here seem to be passing.
@kaatish can we merge this PR?

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2267, opened 2022-05-10
[BUG] Update some clustering algos to only support undirected graphs

### LINKED ISSUE #2263: [QST] Result of Leiden algorithm being different for renumbered and non-renumbered graph
Having a graph represented by the `edgelist` that has a structure like the following:
```
1,0
1,3
2,1
4,0
4,2
```
Vertices are having labels in the range [0, V - 1] therefore they are kind of renumbered already. Loading a graph with `renumber=True` and `renumber=False` gains two completely different results for the Leiden `(cugraph.leiden(G)) algorithm` call. What is the reasoning behind this?

jnke2016:
Our current implementation of `Louvain`, `Leiden`, `KTruss` and `Triangle Count` only support `undirected graphs`. To enforce that, we are currently raising an `exception` if a graph is not of `type Graph` but this is ineffective if a user create a `cugraph.Graph` with the parameter` directed=True`

This PR
- Raises an exception if `directed=True` when creating a `cugraph.Graph`
- Update the docstrings for those algos
- Add tests

closes #2263


python/cugraph/cugraph/community/ktruss_subgraph.py
```diff
@@ -159,7 +160,7 @@ def ktruss_subgraph(G, k, use_weights=True):
     _ensure_compatible_cuda_version()
 
     KTrussSubgraph = Graph()
-    if type(G) is not Graph:
+    if G.is_directed():
         raise Exception("input graph must be undirected")
```
rlratzel, line 164: It would be good to clean these up when we see them: the exception type for these checks should be `ValueError`.

python/cugraph/cugraph/tests/mg/test_mg_louvain.py
```diff
@@ -103,26 +110,19 @@ def uddaskGraphFromDataset(request, dask_client):
...
-    # FIXME: either call Nx with the same dataset and compare results, or
-    # hardcode golden results to compare to.
-    print()
-    print(parts.compute())
-    print(mod)
-    print()
+def test_mg_louvain_with_edgevals_directed_graph(daskGraphFromDataset):
+    # Directed graphs are not supported by Louvain and an exception should be
+    # raised
+    with pytest.raises(Exception):
```
rlratzel, line 116: This should be changed to something more specific too (`ValueError`). Catching `Exception` will catch everything and will hide bugs.

rlratzel (changes requested):
Looks good, just a few minor requests.

python/cugraph/cugraph/tests/mg/test_mg_louvain.py
jnke2016: Addressed

BradReesWork (approved):

rlratzel (approved):

BradReesWork:
@gpucibot merge

PR #2271, opened 2022-05-13
Define triangle_count C API

ChuckHastings:
This defines the C API for triangle counting.

This PR is independent of #2253 and can be merged independently.  The change here defines the C API and returns the `CUGRAPH_NOT_IMPLEMENTED` error when called.  Once #2253 is completed and merged, a follow-up PR will fill in the C API implementation for triangle counting (although the code is written and untested in this PR).

### DIFF
--- a/cpp/include/cugraph_c/community_algorithms.h
+++ b/cpp/include/cugraph_c/community_algorithms.h
@@ -1,0 +1,76 @@
+/*
+ * Copyright (c) 2021-2022, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cugraph_c/error.h>
+#include <cugraph_c/graph.h>
+#include <cugraph_c/resource_handle.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief     Opaque triangle counting result type
+ */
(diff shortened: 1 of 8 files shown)


cpp/include/cugraph_c/community_algorithms.h
```diff
@@ -0,0 +1,76 @@
...
+  int32_t align_;
+} cugraph_triangle_count_result_t;
+
+/**
+ * @brief     Triangle Counting
+ *
+ * @param [in]  handle       Handle for accessing resources
+ * @param [in]  graph        Pointer to graph.  NOTE: Graph might be modified if the storage
+ *                           needs to be transposed
+ * @param [in]  start        Device array of vertices we want to count triangles for
```
seunghwak, line 40: Add a comment saying "if this is NULL, the entire set of vertices is assumed"?

seunghwak (approved):
LGTM (except for one comment about documentation).

cpp/include/cugraph_c/community_algorithms.h
ChuckHastings: Will fix in next push.

cpp/include/cugraph_c/community_algorithms.h
```diff
@@ -0,0 +1,77 @@
...
+ * @brief     Triangle Counting
+ *
+ * @param [in]  handle       Handle for accessing resources
+ * @param [in]  graph        Pointer to graph.  NOTE: Graph might be modified if the storage
+ *                           needs to be transposed
+ * @param [in]  start        Device array of vertices we want to count triangles for.  If NULL
+ *                           the entire set of vertices in the graph is processed
+ * @param [in]  do_expensive_check
+ *                           A flag to run expensive checks for input arguments (if set to true)
+ * @param [in]  result       Output from the triangle_count call
```
rlratzel, line 44: Why is `result` labelled `in` but `error` is an `out`?  I figured they'd both be out.

rlratzel (approved):
LGTM. I focused on the public API part and less on the (commented out) implementation.  I had one questions which need not hold up approval.

cpp/include/cugraph_c/community_algorithms.h
ChuckHastings: Good catch.  Should both be out.  I will correct in a follow-up PR.

ChuckHastings:
@gpucibot merge

PR #2273, opened 2022-05-16
add bindings for triangle counting

jnke2016:
This PR add python bindings for triangle counting.
`cuGraph` and `pylibcuGraph` tests completed

Fails when sampling some vertices from the graph that will be used as parameter (`start_list`) for the MG implementation

The SG implementation works with and without `start_list`


BradReesWork:
rerun tests

python/cugraph/cugraph/dask/community/triangle_count.py
```diff
@@ -0,0 +1,167 @@
...
+import cugraph.dask.comms.comms as Comms
+import dask_cudf
+import cudf
+
+from pylibcugraph.experimental import (
+    ResourceHandle,
+    GraphProperties,
+    MGGraph,
+    triangle_count as pylicugraph_triangle_count
+    )
```
rlratzel, line 28: `ResourceHandle`, `GraphProperties`, and `MGGraph` have all been promoted so we shouldn't be importing them from experimental, otherwise users will get the "experimental-promoted" warning.

Also, can you promote `triangle_count` in pylibcugraph to stable in this PR too?  I think it's unlikely to go away/move/change in a way that would break users so I think we can safely promote it now.

python/cugraph/cugraph/dask/community/triangle_count.py
```diff
@@ -0,0 +1,167 @@
...
+
+import cugraph.dask.comms.comms as Comms
+import dask_cudf
+import cudf
+
+from pylibcugraph.experimental import (
+    ResourceHandle,
+    GraphProperties,
+    MGGraph,
+    triangle_count as pylicugraph_triangle_count
```
rlratzel, line 27:
```suggestion
    triangle_count as pylibcugraph_triangle_count
```

python/cugraph/cugraph/dask/community/triangle_count.py
```diff
@@ -0,0 +1,167 @@
...
+    start_list : list or cudf.Series (int32), optional (default=None)
+        list of starting vertices for triangle count.
+
+    Returns
+    -------
+    result : dask_cudf.DataFrame
+        GPU distributed data frame containing 2 dask_cudf.Series
+
+    ddf['vertex']: dask_cudf.Series
+            Contains thetriangle counting vertices
```
rlratzel, line 100:
```suggestion
            Contains the triangle counting vertices
```

python/pylibcugraph/pylibcugraph/triangle_count.pyx
```diff
@@ -0,0 +1,137 @@
...
+)
+from pylibcugraph.utils cimport (
+    assert_success,
+    copy_to_cupy_array,
+    assert_CAI_type,
+    get_c_type_from_numpy_type,
+)
+
+
+def EXPERIMENTAL__triangle_count(ResourceHandle resource_handle,
```
rlratzel, line 57: Reminder: don't forget to remove the `EXPERIMENTAL_` prefix after it's been promoted.

python/pylibcugraph/pylibcugraph/triangle_count.pyx
```diff
@@ -0,0 +1,137 @@
...
+    Parameters
+    ----------
+    resource_handle: ResourceHandle
+        Handle to the underlying device and host resources needed for
+        referencing data and running algorithms.
+
+    graph: MGGraph
+        The input graph, for Multi-GPU operations.
+
+    start_list: device array type
```
rlratzel, line 74: It looks like from the code that `start_list` can also be `None`.  Can you describe the behavior here if the user passes `None` for `start_list`?

rlratzel (changes requested):
Looks pretty good, just a few requests.

rlratzel (changes requested):
Also, we should probably have some minimal test coverage in `pylibcugraph` for triangle count - I didn't see a new test in pylibcugraph.

python/pylibcugraph/pylibcugraph/triangle_count.pyx
jnke2016: My understanding is if `start_list=None`, `triangle_counts` will compute the number of triangle for each vertex in the graph as opposed to a subset of the graph if `start_list !=None` like NetworkX

python/cugraph/cugraph/dask/community/triangle_count.py
jnke2016: Done

python/cugraph/cugraph/dask/community/triangle_count.py
jnke2016: Done

python/cugraph/cugraph/dask/community/triangle_count.py
jnke2016: Done

python/pylibcugraph/pylibcugraph/triangle_count.pyx
jnke2016: Done

jnke2016: Done

python/pylibcugraph/pylibcugraph/experimental/__init__.py
```diff
@@ -71,3 +71,6 @@
 
 from pylibcugraph.katz_centrality import EXPERIMENTAL__katz_centrality
 katz_centrality = experimental_warning_wrapper(EXPERIMENTAL__katz_centrality)
+
+from pylibcugraph.triangle_count import triangle_count
```
betochimas, line 75: Triangle count is a new pylibcugraph binding, so it wouldn't have been in experimental for a release. Is the promoted from experimental warning necessary in that case / having it also in experimental?

PR #2282, opened 2022-05-17
Adding optional mask to graph view

cjnolet:


cpp/include/cugraph/graph.hpp
```diff
@@ -156,10 +156,12 @@ class graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enab
 
   bool is_weighted() const { return edge_partition_weights_.has_value(); }
 
-  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> view() const
+  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> view(
+    std::optional<std::vector<graph_mask_t<bool> const*>> mask = std::nullopt) const
```
seunghwak, line 160: What's the purpose of passing `mask` as an input parameter?

Who owns the memory block for masks?

How users will update the mask values?

Based on the answer, `enable_vertex_mask = false` & `enable_edge_mask` might be sufficient?

cpp/src/link_analysis/pagerank_impl.cuh
```diff
@@ -55,6 +55,13 @@ void pagerank(
   bool has_initial_guess,
   bool do_expensive_check)
 {
+  /**
+   * TODO: Add masking to the following prims:
+   * 1. count_if_v
+   * 2. count_if_e
+   * 3. transform_reduce_v
+   * 4.
+   */
```
seunghwak, line 64: We have test code for `count_if_v` (https://github.com/rapidsai/cugraph/blob/branch-22.06/cpp/tests/prims/mg_count_if_v.cu).

Maybe for the first step, we may just update `count_if_v` to work with masks (this will require much less work than PageRank).
Then, we work on `count_if_e` before working on something more complex.

cjnolet:
Hey @seunghwak, I thank you for taking a look at this. It's still a little early for feedback at this stage. If it's okay with you, I'll tag you when I feel it's in a good state and we can further discuss some of the specifics if they aren't already obvious by then.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,426 @@
...
+
+ protected:
+  raft::handle_t const& handle_;
+  bool has_vertex_mask_{false};
+  bool has_edge_mask_{false};
+  vertex_t n_vertices_;
+  edge_t n_edges_;
+  bool complement_{false};
+  mask_t* vertices_;
+  mask_t* edges_;
```
seunghwak, line 219: I am not having a single mask_t array will work for edges in multi-GPU. Each GPU holds multiple edge partitions, and shouldn't we better have one array per edge partition?

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,426 @@
...
+  using size_type   = std::size_t;
+
+  ~graph_mask_t()                       = default;
+  graph_mask_t(graph_mask_t&&) noexcept = default;
+
+  graph_mask_t() = delete;
+
+  explicit graph_mask_t(raft::handle_t const& handle,
+                        vertex_t n_vertices,
+                        edge_t n_edges,
```
seunghwak, line 252: Same here, 

I am not having a single mask_t array will work for edges in multi-GPU. Each GPU holds multiple edge partitions, and shouldn't we better have one array per edge partition?

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,426 @@
...
+    if (edges_.size() == 0) {
+      edges_.resize(n_edges_, handle_.get_stream());
+      clear_edge_mask();
+    }
+  }
+
+  /**
+   * Initializes a vertex mask by allocating the device memory
+   */
+  void initialize_vertex_mask()
```
seunghwak, line 336: "initialize" might be interpreted as filling mask values to some initial values. And should we have a separate function for allocation/initialization or can't we allocate & initialize in a constructor?

I think graph_view_t can have a member function to return a graph_mask_t object.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,426 @@
...
+      clear_vertex_mask();
+    }
+  }
+
+  /**
+   * Clears out all the masked bits of the edge mask
+   */
+  void clear_edge_mask()
+  {
+    if (edges_.size() == 0) {
```
seunghwak, line 349: `==` => `>`?

cpp/include/cugraph/graph_view.hpp
```diff
@@ -661,6 +664,21 @@ class graph_view_t<vertex_t,
   edge_t count_self_loops(raft::handle_t const& handle) const;
   edge_t count_multi_edges(raft::handle_t const& handle) const;
 
+  void attach_mask(std::vector<graph_mask_t<vertex_t, edge_t>*>& graph_mask)
```
seunghwak, line 667: Should we better attach a view object?

cpp/include/cugraph/graph_mask.hpp
cjnolet: I don't mind having separate methods for `allocate()` and `initialize()` if that makes it easier to use. Rather than different constructors, I'd prefer if we use a either static factory methods or builders to construct and populate a mask, for example when it is known at construction time that an edge mask and/or vertex mask will be needed.

Thinking out loud, I'm kind of partial to something like this very rough idea:

```c++
graph_mask_t<vertex_t, edge_t, mask_t> graph_mask_builder_t(handle, n_vertices, n_edges)
        .edge_mask(e_mask_func mask_function, bool init = 0)   // init will first do a thrust::fill or cudamemset
        .vertex_mask(v_mask_func mask_function, bool init = 0)  // init will first do a thrust::fill or cudamemset
        .build();
```

This way, the underlying mask owning object can stay mutable even wrt its memory allocations but all the internals and complexities can remain hidden behind the builder API. This also allows a user to build a mask object with edge and/or vertex  masks without having to maintain that logic in multiple constructors (nor make a single constructor more complex than it needs to be).

cpp/include/cugraph/graph_view.hpp
cjnolet: I do think attaching the view object here can make more sense- that way we can assume the necessary mask elements (e.g. vertex / edge mask) have been properly allocated before being used in a `graph_view` object.

seunghwak:
@cjnolet Is this PR ready for review?

cjnolet:
@seunghwak yes, initial single GPU implementation for uniform neighborhood sampling is ready. At this point it's not the most optimal, but it should be correct. I am going to continue optimizing it for 22.10 and finish the MG path.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Compile-time fast lookup of log2(num_bits(mask_t)) to eliminate
+ * the log2 computation for powers of 2.
+ * @tparam mask_t
+ */
+template <typename mask_t>
+__host__ __device__ constexpr int log_bits()
+{
+  switch (std::numeric_limits<mask_t>::digits) {
+    case 8: return 2;
```
seunghwak, line 39: Shouldn't this be 3?

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+    case 16: return 4;
+    case 32: return 5;
+    case 64: return 6;
+    default: return log2(std::numeric_limits<mask_t>::digits);
```
seunghwak, line 43: If I am not mistaken, this will use https://en.cppreference.com/w/cpp/numeric/math/log2.

Will this work when std::numeric_limits<mask_t>::digits == 9?

log2(9) will be 3.xxx and if this gets implicitly converted to int, it will return 3. If this is the intention,

I think

`return static_cast<int>(floor(log2(std::numeric_limits<mask_t>::digits))); ` will be clearer.

Or if you're assuming that the size of mask_t is power of two, should we better add an assert statement to make the intention clear? (e.g. https://stackoverflow.com/questions/108318/how-can-i-test-whether-a-number-is-a-power-of-2).

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/**
+ * Uses bit-shifting to perform a fast mod operation. This
+ * is used to compute the index of a specific bit
+ * @tparam mask_t
+ * @tparam T
+ */
+template <typename mask_t, typename T>
+__host__ __device__ int bit_mod(T numer)
```
seunghwak, line 54: `numer`=>`number`?

seunghwak (commented):
More reviews coming.

cpp/include/cugraph/graph_mask.hpp
cjnolet: Yep that should be 3. I must have made a typo when I was cleaning this up.

cpp/include/cugraph/graph_mask.hpp
cjnolet: Really, I think we could drop the explicit log2 from this function altogether. It's just a little utility meant to provide a fast lookup for log2 of (a very limited set of) powers of two. It's also used almost exclusively on device and so I didn't want to perform strict input checking. I really think the function belongs in the detail namespace but I use it in the sampling utils so I exposed it as part of the masking public API. I'll think about this one a bit more.

cpp/tests/structure/graph_mask_test.cpp
```diff
@@ -0,0 +1,262 @@
...
+  bool multigraph{false};
+  bool check_correctness{true};
+};
+
+template <typename input_usecase_t>
+class Tests_GraphMask
+  : public ::testing::TestWithParam<std::tuple<Graph_Usecase, input_usecase_t>> {
+ public:
+  Tests_GraphMask() {}
+  static void SetupTestCase() {}
```
seunghwak, line 82: Set"U"pTestCase, this is a typo in the current cuGraph implementation, should be fixed after PR 2439 gets merged for the rest part.

seunghwak (commented):
More reviews coming.

cpp/tests/sampling/sg_uniform_neighbor_sampling.cu
```diff
@@ -60,26 +84,64 @@ class Tests_Uniform_Neighbor_Sampling
       std::cout << "construct_graph took " << elapsed_time * 1e-6 << " s.\n";
     }
 
-    auto graph_view                           = graph.view();
-    constexpr edge_t indices_per_source       = 2;
-    constexpr vertex_t repetitions_per_vertex = 5;
-    constexpr vertex_t source_sample_count    = 3;
+    auto graph_view = graph.view();
+
+    printf("Creating mask\n");
```
seunghwak, line 89: Better delete this. And we don't use printf for the host side code.

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -49,8 +49,14 @@ std::tuple<rmm::device_uvector<typename graph_view_t::vertex_type>,
            rmm::device_uvector<typename graph_view_t::edge_type>>
 uniform_nbr_sample_impl(
   raft::handle_t const& handle,
+
   graph_view_t const& graph_view,
+
+  // List of vertices to sample.
   rmm::device_uvector<typename graph_view_t::vertex_type>& d_in,
+
```
seunghwak, line 57: No need for this empty line.

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -49,8 +49,14 @@ std::tuple<rmm::device_uvector<typename graph_view_t::vertex_type>,
            rmm::device_uvector<typename graph_view_t::edge_type>>
 uniform_nbr_sample_impl(
   raft::handle_t const& handle,
+
   graph_view_t const& graph_view,
+
```
seunghwak, line 54: No need for this empty line.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -18,10 +18,12 @@
 
 #include <cugraph/detail/decompress_edge_partition.cuh>
 #include <cugraph/edge_partition_device_view.cuh>
+#include <cugraph/graph_mask_functions.cuh>
```
seunghwak, line 21: I feel like it is better to roll-back the mask updates in the sampling code. These sampling code should be re-written with a selection primitive. AFAIK, there is no immediate user to use this masking support for sampling. So, this work will likely to be discarded in the future without actually being used. @ChuckHastings What do you think?

In this PR, we may better focus on defining the graph_mask_t and graph_mask_view_t classes and attaching masks to a graph_view object.

cpp/src/prims/edge_partition_src_dst_property.cuh
```diff
@@ -94,6 +94,8 @@ class edge_partition_major_property_device_view_t {
     }
 
     assert((partition_idx == size_t{0}) || edge_partition_major_value_start_offsets_);
+    // This is failing in the debug build...
+    //    assert((partition_idx == size_t{0}) || edge_partition_value_firsts_);
```
seunghwak, line 98: This code is deleted in one of the previously merged PRs. Please merge with the current branch.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+   */
+  bool has_vertex_mask() const { return get_vertex_mask().has_value(); }
+
+  /**
+   * Returns the edge mask if it has been initialized on the instance
+   * @return
+   */
+  std::optional<mask_t const*> get_edge_mask() const
+  {
+    return edges_.size() > 0 ? std::make_optional<mask_t const*>(edges_.data()) : std::nullopt;
```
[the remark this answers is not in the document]
cjnolet: If a graph has no edges then how could it have an edge mask?

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+
+  vertex_t get_n_vertices() { return n_vertices_; }
+
+  edge_t get_n_edges() { return n_edges_; }
+
+  edge_t get_edge_mask_size() const { return n_edges_ >> log_bits<mask_t>(); }
+
+  vertex_t get_vertex_mask_size() const { return n_vertices_ >> log_bits<mask_t>(); }
+
+  void initialize_edge_mask(bool init = 0)
```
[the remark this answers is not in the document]
cjnolet: I'm drawing a very clear distinction here. To allocate a mask just means memory has been allocated (and cleared). The initialize actually sets the values to the user's choosing. This doesn't have to do with complement at all- for example in testing, I found I wanted to initialize a mask with all 1's so that I could unset just a few of the bits. I made that available in the initialization so that we can keep that logic in the mask itself.

cpp/include/cugraph/graph_view.hpp
```diff
@@ -1072,6 +1105,8 @@ class graph_view_t<vertex_t,
 
   // segment offsets based on vertex degree, relevant only if vertex IDs are renumbered
   std::optional<std::vector<vertex_t>> segment_offsets_{std::nullopt};
+
+  std::optional<graph_mask_view_t<vertex_t, edge_t>*> mask_{std::nullopt};
```
seunghwak, line 1109: Why pointer here? A view object should be cheap to copy.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+  {
+    if (!has_edge_mask()) {
+      allocate_edge_mask();
+      RAFT_CUDA_TRY(cudaMemsetAsync(edges_.data(),
+                                    edges_.size() * sizeof(mask_t),
+                                    std::numeric_limits<mask_t>::max() * init,
```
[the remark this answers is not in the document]
cjnolet: Other than being a member of the mask object, the complement is not supported at all currently. I can remove it in the near term if it makes it easier.

cpp/include/cugraph/graph_view.hpp
```diff
@@ -851,6 +870,8 @@ class graph_view_t<vertex_t,
 
   bool is_weighted() const { return weights_.has_value(); }
 
+  //  bool has_mask() const { return mask_.has_value(); }
```
seunghwak, line 873: Delete dead code.

cpp/include/cugraph/graph_view.hpp
```diff
@@ -958,6 +979,18 @@ class graph_view_t<vertex_t,
...
+    mask_ = std::make_optional<graph_mask_view_t<vertex_t, edge_t>*>(&mask);
+  }
+
+  std::optional<graph_mask_view_t<vertex_t, edge_t>> get_mask_view(size_t partition_idx = 0) const
+  {
+    assert(partition_idx == 0);
+    return mask_.has_value() ? std::make_optional<graph_mask_view_t<vertex_t, edge_t>>(**mask_)
+                             : std::nullopt;
+  }
+
```
seunghwak, line 993: And we should have something like `clear_mask_view` as well setting `mask_` to std::nullopt;

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+  }
+
+  /**
+   * Initializes an edge mask by allocating the device memory
+   */
+  void allocate_edge_mask()
+  {
+    if (edges_.size() == 0) {
+      edges_.resize(get_edge_mask_size(), handle_.get_stream());
+      clear_edge_mask();
```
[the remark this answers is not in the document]
cjnolet: This is really just for completeness- if we initialize a mask to some value (it can be 0 or 1 and doesn't depend at all on whether it's a complement), it's nice to set the mask back to 0. That said, it's not even being called anywhere.

cpp/include/cugraph/graph_view.hpp
```diff
@@ -679,6 +682,21 @@ class graph_view_t<vertex_t,
...
+  std::optional<graph_mask_view_t<vertex_t, edge_t>> get_mask_view(size_t partition_idx = 0) const
+  {
+    graph_mask_view_t<vertex_t, edge_t>* mask = (*mask_)[partition_idx];
+    return mask_.has_value()
+             ? std::make_optional<graph_mask_view_t<vertex_t, edge_t>>(*(*mask_)[partition_idx])
+             : std::nullopt;
+  }
+
+  std::optional<vertex_t const*> local_sorted_unique_edge_src_begin() const;
```
seunghwak, line 698: What's this? Why did you add this? It seems like this appeared here due to some erroneous merge conflict resolution.

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+                        edge_t n_edges,
+                        bool complement = false)
+    : handle_(handle),
+      n_vertices_(n_vertices),
+      n_edges_(n_edges),
+      edges_(0, handle.get_stream()),
+      vertices_(0, handle.get_stream()),
+      complement_(complement)
+  {
+  }
```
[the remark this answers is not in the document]
cjnolet: That seems really awkward to me to initialize an owning mask object with a graph view object just so we can then turn around and attach the masks view on the graph view.

Did you see my previous comments about using a builder or factory to hide the more complex bits?

cpp/include/cugraph/graph_view.hpp
```diff
@@ -821,6 +839,7 @@ class graph_view_t<vertex_t,
                      std::optional<raft::host_span<vertex_t const>>,
                      std::optional<std::byte> /* dummy */>
     local_sorted_unique_edge_dst_vertex_partition_offsets_{std::nullopt};
+  std::optional<std::vector<graph_mask_view_t<vertex_t, edge_t>*>> mask_{std::nullopt};
```
seunghwak, line 842: The difference between SG & MG should be encapsulated inside the graph_mask_view_t object. This will end-up replicating vertex mask multiple times if there are multiple edge partitions.

cpp/include/cugraph/graph_view.hpp
```diff
@@ -958,6 +979,18 @@ class graph_view_t<vertex_t,
     return vertex_partition_view_t<vertex_t, false>(this->number_of_vertices());
   }
 
+  void attach_mask_view(graph_mask_view_t<vertex_t, edge_t>& mask)
+  {
+    mask_ = std::make_optional<graph_mask_view_t<vertex_t, edge_t>*>(&mask);
+  }
+
+  std::optional<graph_mask_view_t<vertex_t, edge_t>> get_mask_view(size_t partition_idx = 0) const
```
seunghwak, line 987: This should not take `partition_idx`. We'd better not use one graph_mask_view_t object per edge partition; note that graph_mask_view_t object has vertex mask as well. In case of MG, a graph_mask_view_t object may have more than one edge mask arrays internally.

seunghwak (commented):
Need to review graph_mask_functions.cuh as well.

I think we'd better roll-back the updates in the sampling code. The current sampling code is pretty convoluted and updating this to support masking can be very challenging and bug-prone. This should be re-written with a selection primitive. I am worried that the updates in the sampling part will just be discarded without really being used. @ChuckHastings What do you think?

cpp/include/cugraph/graph_mask_functions.cuh
```diff
@@ -0,0 +1,281 @@
...
+                    graph_mask_view_t<vertex_t, edge_t, mask_type> const& mask,
+                    int major_range_first,
+                    vertex_t const* vertex_ids,
+                    edge_t const* indptr)
+{
+  CUGRAPH_EXPECTS(mask.get_edge_mask().has_value(), "Edge mask is required to compute degrees.");
+  detail::masked_degree_kernel<vertex_t, edge_t, mask_type, 128>
+    <<<size, 128, 0, handle.get_stream()>>>(
+      degrees_output, mask.get_edge_mask().value().data(), major_range_first, vertex_ids, indptr);
+}
```
seunghwak, line 279: I assume this code is unnecessary if we roll-back the updates in neighbor sampling.

Ideally, all the complexities related to masking should be handled inside the primitives, and should not be exposed in non-detail space public API.

cjnolet:
Hey @seunghwak I think I'm going to go ahead and let you take on the masking. The priority of the masking in the neighborhood sampling was set by @BradReesWork and I've been discussing this design with @ChuckHastings. But that said, I'm about done here, so I'm going to close this PR and let someone else implement this.

cjnolet:
Hey @seunghwak would you prefer to implement the masking instead? It seems from the discussion in this PR thus far that you might have your mind set on a design already and would just prefer it's done your way.

@BradReesWork requested that I start with uniform neighborhood sampling and I've been discussing the approach in this PR with @ChuckHastings. Though it's not yet the most optimal, I plan to continue improving the performance throughout 22.10. I've also implemented most of the feedback you provided in this PR and in prior conversations we have had.

A lot of work went into getting this initial functionality for 22.08 and I'm a bit frustrated with the idea of reverting the main focus of this PR (the sampling). I understand we are very close to burndown and I'm not opposed to pushing this to an early 22.10 task, but I would like to respectfully request that you consider my ideas here as well so we can have a two-sided discussion.

cjnolet:
Opened #2679 with just the graph mask class.

PR #2287, opened 2022-05-19
Implement eigenvector centrality

ChuckHastings:
This PR implements Eigenvector Centrality in C++ using the graph primitives.  It also provides the C API implementation.

There are unit tests for C++ and C both SG and MG.

Partially addresses #2146


cpp/include/cugraph/algorithms.hpp
```diff
@@ -1221,6 +1221,38 @@ void pagerank(raft::handle_t const& handle,
...
+ * differences in eigenvector centrality values between two consecutive iterations is less than the
+ * number of vertices in the graph multiplied by @p epsilon.
+ * @param max_iterations Maximum number of power iterations.
+ * @param do_expensive_check A flag to run expensive checks for input arguments (if set to `true`).
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+void eigenvector_centrality(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, weight_t, true, multi_gpu> const& graph_view,
+  raft::device_span<weight_t> centralities,
```
seunghwak, line 1251: Just for the sake of discussion,

So, what do you think about passing `raft::device_span<weight_t> centralities` as an input argument vs returning rmm::device_uvector<weight_t> holding centrality values?

The former might be more natural when we're passing initial values and we may be able to reduce memory allocations (when we are running PageRank with different personalization vectors, but with the rmm pool allocator, memory allocation overhead might be insignificant) while the latter might be more functional.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1221,6 +1221,38 @@ void pagerank(raft::handle_t const& handle,
...
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * or multi-GPU (true).
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @param centralities Device span where we should store the eigenvector centralities
```
seunghwak, line 1240: Can we pass initial values?

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I got the idea of using the span from looking at your new `triangle_count` implementation.  The [in/out] of centralities is more consistent with what we have been doing.  Our paradigm thus far has been to specify the output storage a priori if we can know it, and to allocate it dynamically if we can't know it.

What you are suggesting would be a paradigm shift for the API.  I'm not opposed to changing the paradigm.

It seems to me the current paradigm has the following advantages:
 * Less memory allocation.  The new strategy would require temporarily having an extra vector of length V.
 * The caller can use any memory allocator that they choose to allocate the device memory

The new paradigm would have the following advantages:
 * More functional in nature
 * More consistency (all algorithms would return results the same way, whether the size is predictable or not)

In the grand scheme of memory things, I'm not all that concerned over allocating an extra result array temporarily.  It seems to me that the functional feel of the proposed paradigm is useful and consistency in how algorithms behave across the interface is always better.

ChuckHastings: In this case I can certainly change `raft::device_span<weight_t> centralities` to `std::optional< raft::device_span<weight_t>> centralities` to support an optional input, and make the return value `rmm::device_uvector<weight_t>`

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I will add that support.  Missed that.

cpp/src/centrality/eigenvector_centrality_impl.cuh
```diff
@@ -0,0 +1,155 @@
...
+#include <cugraph/prims/reduce_v.cuh>
+#include <cugraph/prims/transform_reduce_v.cuh>
+#include <cugraph/prims/update_edge_partition_src_dst_property.cuh>
+#include <cugraph/utilities/error.hpp>
+
+#include <raft/handle.hpp>
+#include <rmm/exec_policy.hpp>
+
+#include <thrust/fill.h>
+#include <thrust/for_each.h>
```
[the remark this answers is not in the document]
ChuckHastings: Probably not, copy/paste.  I'll check all the headers.

cpp/src/centrality/eigenvector_centrality_impl.cuh
```diff
@@ -0,0 +1,155 @@
...
+                   [] __device__(vertex_t, vertex_t, weight_t w, auto, auto) { return w <= 0.0; });
+      CUGRAPH_EXPECTS(num_nonpositive_edge_weights == 0,
+                      "Invalid input argument: input graph should have postive edge weights.");
+    }
+  }
+
+  thrust::fill(handle.get_thrust_policy(),
+               centralities.begin(),
+               centralities.end(),
+               weight_t{1.0} / static_cast<weight_t>(num_vertices));
```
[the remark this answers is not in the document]
ChuckHastings: Will add, missed that.

ChuckHastings:
Pushed an update to address @seunghwak comments

seunghwak (approved):

cpp/src/centrality/eigenvector_centrality_impl.cuh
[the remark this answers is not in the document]
ChuckHastings: Done

ChuckHastings:
@gpucibot merge

PR #2288, opened 2022-05-19
Add and test mechanism for creating graph with edge index as weight

ChuckHastings:
Missed this in the last uniform neighborhood sampling updates.

Added a mechanism for creating a graph using an edge index as the weight.  This is a temporary mechanism until we add direct support for edge indexes in the graph.

### DIFF
--- a/cpp/tests/c_api/mg_test_utils.cpp
+++ b/cpp/tests/c_api/mg_test_utils.cpp
@@ -191,15 +191,15 @@ extern "C" int create_mg_test_graph(const cugraph_resource_handle_t* handle,
 
   return test_ret_value;
 }
-extern "C" int create_mg_test_graph_with_ids(const cugraph_resource_handle_t* handle,
-                                             int32_t* h_src,
-                                             int32_t* h_dst,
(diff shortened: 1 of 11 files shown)


cpp/tests/c_api/mg_test_utils.cpp
```diff
@@ -191,3 +191,102 @@ extern "C" int create_mg_test_graph(const cugraph_resource_handle_t* handle,
 
   return test_ret_value;
 }
+extern "C" int create_mg_test_graph_with_ids(const cugraph_resource_handle_t* handle,
```
seunghwak, line 194: `with_edge_ids` might be more intuitive?

seunghwak (approved):
LGTM.

cpp/tests/c_api/mg_test_utils.cpp
ChuckHastings: Fixed in latest push

jnke2016 (approved):
Reviewed and tested. It looks good to me

BradReesWork:
@gpucibot merge

PR #2292, opened 2022-05-19
Add Louvain to the C API

ChuckHastings:
This PR adds Louvain to the C API.  This PR is a full implementation including:
 * C API definition
 * C API implementation calling the C++ version of Louvain
 * SG and MG unit tests exercising the Louvain implementation
 * A small documentation update to triangle counting that was discovered in the final review of the triangle counting C API PR


seunghwak (approved):
So, when can we delete the legacy Louvain? After python updates?

PR #2295, opened 2022-05-20
Migrate SG and MG SSSP to pylibcugraph

alexbarghi-nv:
Uses the pylibcugraph C API to make these calls instead of the wrapper functions.  Removes wrapper functions from the code.


alexbarghi-nv:
Needs to be merged after #2305

ChuckHastings (approved):

python/cugraph/cugraph/dask/traversal/sssp.py
```diff
@@ -16,41 +16,67 @@
...
-                           handle,
-                           segment_offsets,
-                           start)
-
-
-def sssp(input_graph, source):
+from pylibcugraph import sssp as pylibcugraph_sssp
+from pylibcugraph.experimental import (ResourceHandle,
+                                       GraphProperties,
+                                       MGGraph)
```
rlratzel, line 27: We should be importing the non-experimental APIs now.

python/cugraph/cugraph/traversal/sssp.py
```diff
@@ -15,13 +15,17 @@
...
                                is_matrix_type,
                                is_cp_matrix_type,
                                is_nx_graph_type,
                                cupy_package as cp,
                                )
+from pylibcugraph import sssp as pylibcugraph_sssp
+from pylibcugraph.experimental import (ResourceHandle,
+                                       GraphProperties,
+                                       SGGraph,
+                                       )
```
rlratzel, line 28: Same here: we should be importing the non-experimental APIs now.

rlratzel (changes requested):
Looks good, just two quick changes.

python/cugraph/cugraph/dask/traversal/sssp.py
alexbarghi-nv: fixed

python/cugraph/cugraph/traversal/sssp.py
alexbarghi-nv: fixed

jnke2016 (approved):
Looks good to me

alexbarghi-nv:
@BradReesWork Rick's requested changes are done, you can override the blocker and merge this.

PR #2305, opened 2022-05-23
Add test to reproduce issue with double weights, fix issue (graph cre…

ChuckHastings:
@alexbarghi-nv discovered issue with creating graphs through the C API with double weights.

Added CAPI SSSP test which demonstrates the error.  Added fix to the error in the graph creation logic (both SG and MG).

### DIFF
--- a/cpp/src/c_api/graph_mg.cpp
+++ b/cpp/src/c_api/graph_mg.cpp
@@ -211,7 +211,7 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
                "Invalid input arguments: src type != dst type.",
                *error);
 
-  CAPI_EXPECTS(!weights || (p_weights->size_ == p_src->size_),
+  CAPI_EXPECTS((weights == nullptr) || (p_weights->size_ == p_src->size_),
                CUGRAPH_INVALID_INPUT,
                "Invalid input arguments: src size != weights size.",
                *error);
@@ -225,7 +225,7 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
     edge_type = data_type_id_t::INT64;
   }
 
-  if (!weights) {
+  if (weights != nullptr) {
     weight_type = p_weights->type_;
   } else {
     weight_type = data_type_id_t::FLOAT32;
--- a/cpp/include/cugraph_c/graph.h
+++ b/cpp/include/cugraph_c/graph.h
@@ -40,7 +40,8 @@ typedef struct {
  * @param [in]  properties  Properties of the constructed graph
  * @param [in]  src         Device array containing the source vertex ids.
  * @param [in]  dst         Device array containing the destination vertex ids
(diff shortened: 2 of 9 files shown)


alexbarghi-nv:
I just ran some tests and it looks like ints as weights are being rejected as well, which causes some of the tests to fail.  Is there any reason integer weights aren't supported?  Right now it return this error:
`RuntimeError: non-success value returned from cugraph_sg_graph_create(): CUGRAPH_UNSUPPORTED_TYPE_COMBINATION`

ChuckHastings:
> I just ran some tests and it looks like ints as weights are being rejected as well, which causes some of the tests to fail. Is there any reason integer weights aren't supported? Right now it return this error:
> `RuntimeError: non-success value returned from cugraph_sg_graph_create(): CUGRAPH_UNSUPPORTED_TYPE_COMBINATION`

Integer weights are not supported primarily because it would explode the compilation time (which is already very long).

Our code uses C++ templates and are templated on the vertex type (int32_t or int64_t), the edge type (int32_t or int64_t), the weight type (float or double), whether or not the graph is stored in transposed format, and whether or not the graph is represented on a single GPU or multiple GPUs.  This results in 16 different supported combinations, each of which is generated and compiled.  Adding support for integer weight types would double the compile time of the C++ code.  To this point we have had very few requests to support integer types.

It is certainly something we can add if there is sufficient user interest.

alexbarghi-nv:
Ok, in that case I will force float/double weights.  That must have been what it was doing before if we were never supporting integer weights.

alexbarghi-nv (approved):
LGTM, tests are passing.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -211,7 +211,7 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
                "Invalid input arguments: src type != dst type.",
                *error);
 
-  CAPI_EXPECTS(!weights || (p_weights->size_ == p_src->size_),
+  CAPI_EXPECTS((weights == nullptr) || (p_weights->size_ == p_src->size_),
                CUGRAPH_INVALID_INPUT,
                "Invalid input arguments: src size != weights size.",
                *error);
```
seunghwak, line 217: Question, `weights` can be `nullptr` for either unweighted graphs or weighted but # (local) edges == 0. How do you distinguish these two cases?

ChuckHastings: The assumption in the C API is the following:
 * For an unweighted graph, `weights == nullptr`
 * For a weighted graph, the weights will be passed, but that it will be an empty array if the number of local edges is 0.  So the `cugraph_type-erased_device_array_view_t *` would be passed if the graph is weighted, but the data pointer will be `nullptr` and the size will be 0

I will add that to the documentation of the graph creation functions.

seunghwak (approved):

BradReesWork:
@gpucibot merge

PR #2318, opened 2022-05-31
Fix BFS Docstring

alexbarghi-nv:
Corrects the BFS docstring by clarifying that multiple start vertices are accepted as long as they are from different components, and that passing start vertices from the same component will result in undefined behavior.


python/cugraph/cugraph/traversal/bfs.py
```diff
@@ -183,8 +183,11 @@ def bfs(G,
         information. Edge weights, if present, should be single or double
         precision floating point values.
 
-    start : Integer, optional (default=None)
-        The index of the graph vertex from which the traversal begins
+    start : Integer or list, optional (default=None)
+        The id of the graph vertex from which the traversal begins, or
+        if a list, the vertex from which the traversal begins in each
+        component of the graph.  Passing in multiple start vertices
```
betochimas, line 189: If the current MG bfs call supports only 1 seed vertex per component, would it be better to instead say that the list option is unsupported as of this moment and add a FIXME in the code instead? Stating that certain parameters results in undefined behavior would be better suited for GitHub issues, rather than the user-friendly docs

betochimas: Same for the SG bfs call. Overall the PR looks good, so I can approve it later today

alexbarghi-nv: list is supported so long as no two vertices in the list are in the same component.  I can remove the mention of undefined behavior and just state that.

alexbarghi-nv: I just changed it.

betochimas (approved):
Looks good!

jnke2016 (approved):
Looks good to me

BradReesWork:
@gpucibot merge

PR #2328, opened 2022-06-02
Performance-optimize storing edge partition source/destination properties in (key, value) pairs

seunghwak:
Using (key, value) pairs instead of linear array for the entire source/destination range cuts memory footprint significantly when # GPUs >> (V/E)^2.

This PR performance-optimizes finding the value for the given key and also enables (key, value) pairs (this feature was previously disabled by default).

The original implementation binary searched over the entire set of local source/destination keys. This PR compared a hash table based approach (using cuCollection) and the auxiliary array based approach (described in the original DCSC paper) and selected the auxiliary array based approach. 

Breaking due to few graph primitives API changes.


cpp/src/structure/graph_impl.cuh
```diff
@@ -469,202 +473,239 @@ update_local_sorted_unique_edge_majors_minors(
...
                           static_cast<double>(num_local_unique_edge_majors) /
                             static_cast<double>(aggregate_major_range_size),
                           raft::comms::op_t::MAX,
                           handle.get_stream());
-  auto max_minor_properties_fill_ratio = host_scalar_allreduce(
-    comm,
-    static_cast<double>(num_local_unique_edge_minors) / static_cast<double>(minor_size),
-    raft::comms::op_t::MAX,
-    handle.get_stream());
+  std::cout << "max_major_properties_fill_ratio=" << max_major_properties_fill_ratio
```
ChuckHastings, line 635: Residual debug statement?

seunghwak: Oops, I though I deleted this but it seems like I didn't :-) I will delete this.

kaatish (approved):

ChuckHastings:
@gpucibot merge

PR #2332, opened 2022-06-02
update versions to 22.08

ChuckHastings:
Update conda files to upgrade raft versions from 22.06 to 22.08

### DIFF
--- a/conda/environments/cugraph_dev_cuda11.2.yml
+++ b/conda/environments/cugraph_dev_cuda11.2.yml
@@ -11,8 +11,8 @@ dependencies:
 - libcudf=22.08.*
 - rmm=22.08.*
 - librmm=22.08.*
-- libraft-headers=22.06.*
-- pyraft=22.06.*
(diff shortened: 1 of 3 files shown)


ChuckHastings:
> @ChuckHastings, can you add entries for these libraries in the automatic update script below?
> 
> I think the script could be generalized a bit to something like the following to make sure it captures all current and future libs. I'll leave this up to your discretion.
> 
> ```shell
> sed_runner "s/=${CURRENT_SHORT_TAG}/=${NEXT_SHORT_TAG}/g" ${FILE};
> ```
> 
>  * https://github.com/rapidsai/cugraph/blob/492245009cd2075054573b450a602422ae8f4a78/ci/release/update-version.sh#L52-L60

I added a pair of sed expressions to address those libraries.  I get nervous about generalizing too much, although none of our current external dependencies overlap in our versioning convention.

ChuckHastings:
@gpucibot merge

PR #2337, opened 2022-06-03
Update the list of algos to benchmark

jnke2016:
This PR
1. Update the way `uniform neighbor sample` is imported( it has been removed from experimental)
2. Ping `libraft-headers` and `pyraft` to 22.08
3. Add `Triangle count` to the list of algos to benchmarks


rlratzel:
rerun tests

python/cugraph/cugraph/tests/mg/test_mg_neighborhood_sampling.py
```diff
@@ -36,6 +36,7 @@ def setup_function():
 # FIXME: Do more testing for this datasets
 # [utils.RAPIDS_DATASET_ROOT_DIR_PATH/"email-Eu-core.csv"]
 datasets = utils.DATASETS_UNDIRECTED
+# datasets = utils.DATASETS_UNDIRECTED_WEIGHTS
```
rlratzel, line 39: Can this be removed?

python/cugraph/cugraph/tests/mg/test_mg_neighborhood_sampling.py
```diff
@@ -120,30 +121,43 @@ def test_mg_neighborhood_sampling_simple(dask_client, input_combo):
...
         join2 = input_df.merge(
-            result_nbr, how='left', left_on=[*input_df.columns],
+            result_nbr, how='right', left_on=[*input_df.columns],
             right_on=[*result_nbr.columns])
+
         pd.set_option('display.max_rows', 500)
-        print('df1 = \n', input_df.sort_values([*input_df.columns]))
-        print('df2 = \n', result_nbr.sort_values(
+        # print('input_df = \n', input_df.sort_values([*input_df.columns]))
+        print('result_df = \n', result_nbr.sort_values(
```
rlratzel, line 129: I can see the usefulness of print statements in tests, but it also seems like there should be a better way to get this information if we think it's needed by everyone (eg. a dedicated test that fails if a particular result is unexpected).  If this was just a debug print though, I think it should be removed.

rlratzel (commented):
Looks good, I just had some minor feedback on one of the test files.

rlratzel (approved):

PR #2339, opened 2022-06-06
Clean up some unused code in the C API (and beyond)

ChuckHastings:
Cleaning up some unused code in the C API, and some old visitor methods that are no longer used once this code is gone.


seunghwak (approved):

ChuckHastings:
@gpucibot merge

PR #2343, opened 2022-06-06
Fix sampling bug

ChuckHastings:
@jnke2016 found a bug in the sampling code.

Fix weight lookup to match index lookup.  The weight array mirrors the indices array in its structure, but the computation to do the lookup was correct for the indices array but incorrect for the weights array.


seunghwak (approved):

ChuckHastings:
@gpucibot merge

PR #2357, opened 2022-06-15
Remove topology header

ChuckHastings:
Remove the header file `topology.cuh` which is no longer used.


cpp/src/components/connectivity.cu
```diff
@@ -27,8 +27,6 @@
 #include <type_traits>
 #include <utilities/graph_utils.cuh>
```
seunghwak, line 28: May better update this to

```
#include <utilities/graph_utils.cuh>

#include <type_traits>
```

to be consistent with the rest of the cuGraph.

seunghwak (approved):
LGTM (I added a comment to clean-up one more our code debt).

cpp/src/components/connectivity.cu
ChuckHastings: Reorganized includes.

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2358, opened 2022-06-15
[REVIEW]Fix Fanout -1

VibhuJawa:
This PR fixes sampling  when for the default value of -1 which is called below: 

https://github.com/rapidsai/cugraph/blob/92f6ba451b2d9e6c3f60dbccfa05bbf3c480e43a/python/cugraph/cugraph/gnn/graph_store.py#L129-L133

In workflows this is called during inference so we need this to work for inference to work .with CuGraphStorage .

### DIFF
--- a/python/cugraph/cugraph/gnn/graph_store.py
+++ b/python/cugraph/cugraph/gnn/graph_store.py
@@ -85,7 +85,7 @@ class CuGraphStore:
             Node IDs to sample neighbors from.
         fanout : int
             The number of edges to be sampled for each node on each edge type.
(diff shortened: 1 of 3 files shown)


rlratzel (approved):
The code change and test LGTM and I'll just approve now, but I have one small request: could you also update the docstring [here](https://github.com/rapidsai/cugraph/blob/92f6ba451b2d9e6c3f60dbccfa05bbf3c480e43a/python/cugraph/cugraph/gnn/graph_store.py#L87) to indicate that `-1` now returns all samples?

alexbarghi-nv (approved):

Issue #1348, opened 2021-01-21
[ENH] Add a `datasets` API and cleanup existing datasets


rlratzel:
CuGraph could benefit from a new `datasets` API that allows users to easily create Graph objects from pre-defined datasets.  Currently, users have to read a file using cuDF to create an edgelist, then create a Graph instance from the edgelist.  This is a multi-step process that is very common, yet not always easy to remember.  Streamlining this very common process will benefit users as well as our tests and benchmarks.

The datasets directory also needs to be cleaned up a bit and the README updated.

* Consolidate redundant subdirs and remove unused: “ref”, “test/ref”, “highbench_small/1”
* Update README (if necessary)
* Update download script (if necessary)

For all of the above, the Python and C++ test and benchmarks need to be checked to see if they are referencing any dataset that might be a moved or renamed as part of the cleanup.

betochimas:
Some tasks for this project:
- [ ] Write Dataset and Metadata class
- [ ] Write datasets/__init__.py and meta-data files for each dataset
- [ ] Remove unused datasets across all layers of cugraph
- [ ] Testing
- [ ] Update docstrings
- [ ] Update notebooks
- [ ] If necessary, write notebook and/or section in html docs demoing API

oorliu:
I fleshed out the `__init__` method for `MetaData` to open a JSON file and store its info.

betochimas:
@cjnolet @dantegd @wphicks 
As of now, these are current requirements for the `datasets` API. The issue addressing the first set of changes is #2351 
We appreciate any feedback as we work on making this tool compatible across cugraph and others

betochimas:
In today's sync, we agreed on ensuring that the metadata and configuration `.yaml` files are input-only (user provides them if they want to load custom datasets or modify configs) and are not writeable. The path to the datasets will then be stored as an attribute to the `datasets` object, and will not be part of the metadata in the `.yaml` file. We introduced `load_all`, a utility function outside of `datasets` class that fetches all datafiles from the metadata files from within the metadata directory. This can be used for validation, and downloading dataset all together

betochimas:
Status update:

- get_path returns the path to where the dataset is stored in the form of a PosixPath object, if downloaded. Otherwise will return None, indicating that the dataset could not be found
- load_all is a utility function that loads all metadata yamls from within metadata and downloads them if missing. Custom metadata yamls can be dropped in that directory in addition to the defaults.
[...]

PR #2367, opened 2022-06-22
Add datasets API to import graph data from configuration/metadata files

betochimas:
Addresses issue [#1348](https://nvidia.slack.com/archives/C01SCT7ELMR). A working version of the datasets API has been added under the "experimental" module of cuGraph. This API comes with the ability to import a handful of built-in datasets to create graphs and edge lists. Each dataset comes with its own metadata file in the format of a YAML file. These files contain general information about the dataset, as well as formatting information about their columns and datatypes.


raydouglass:
rerun tests

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+# limitations under the License.
+
+import cugraph
+import cudf
+import yaml
+import os
+from pathlib import Path
+
+
+class DefaultDownloadDir:
```
rlratzel, line 21: Can you add a docstring for this class too?

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        The metadata file for the specific graph dataset, which includes
+        information on the name, type, url link, data loading format, graph
+        properties
+
+    """
+    def __init__(self, meta_data_file_name):
+        self.__dir_path = Path(__file__).parent.absolute()
+
+        self._meta_data_file_name = meta_data_file_name
+        self.__read_meta_data_file(self._meta_data_file_name)
```
rlratzel, line 61: It looks like `self._meta_data_file_name` is only used here, so it probably doesn't need to be a member.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+    Parameters
+    ----------
+    meta_data_file_name : yaml file
+        The metadata file for the specific graph dataset, which includes
+        information on the name, type, url link, data loading format, graph
+        properties
+
+    """
+    def __init__(self, meta_data_file_name):
+        self.__dir_path = Path(__file__).parent.absolute()
```
rlratzel, line 58: It looks like `self.__dir_path` is only used in the call to read the meta-data file, so you could just move this to the `__read_meta_data_file()` method and remove it here as a member.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+
+        self._meta_data_file_name = meta_data_file_name
+        self.__read_meta_data_file(self._meta_data_file_name)
+
+        self._dd = default_download_dir.dd
```
rlratzel, line 63: This could probably also use a better name than `dd`, plus it would be better if you assigned it the `default_download_dir` instance itself instead of the value of `.dd`, since that will allow you to always easily lookup the current value instead of having a `__refresh_dd()` method.
```suggestion
        self._download_dir = default_download_dir
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        self._edgelist = None
+        self._graph = None
+        self._path = None
+
+    def __read_meta_data_file(self, meta_data_file):
+        metadata_path = self.__dir_path / meta_data_file
+        with open(metadata_path, 'r') as file:
+            self.metadata = yaml.safe_load(file)
+            file.close()
```
rlratzel, line 72: The `open()` context manager will call `.close()` automatically, so you should be able to remove this line.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # check for updates for download_dir location within obj
+    def __refresh_dd(self):
```
rlratzel, line 75: I think this method should be removed in favor of having the class access `self._dd.path` directly each time a path is needed.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+        if self._dd != default_download_dir.dd:
+            self._dd = default_download_dir.dd
+
+    def __download_csv(self, url):
+        if not os.path.isdir(self._dd):
```
rlratzel, line 80: so for this with the above changes, you'd just add:
```
download_dir = self._dd.path
```
Then replace `self._dd` in this method with `download_dir` and you'll always have the most current value of download_dir without needing a separate call to `self.__refresh_dd()`

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+            os.makedirs(self._dd)
+
+        filename = self.metadata['name'] + self.metadata['file_type']
+        if os.path.isdir(self._dd):
+            df = cudf.read_csv(url)
+            df.to_csv(self._dd / filename, index=False)
+
+        else:
+            raise RuntimeError("The directory " + str(self._dd.absolute())
+                               + " does not exist")
```
rlratzel, line 90: With the above suggestion:
```suggestion
            raise RuntimeError(f"The directory {download_dir.absolute()} does not exist")
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+            full_path = self._dd / (self.metadata['name'] +
+                                    self.metadata['file_type'])
+
+            if not os.path.isfile(full_path):
+                if fetch:
+                    self.__download_csv(self.metadata['url'])
+                else:
+                    raise RuntimeError("The datafile does not exist. Try" +
+                                       " get_edgelist(fetch=True) to" +
+                                       " download the datafile")
```
rlratzel, line 114: You should include the path to the file it's trying to read, and you also don't need `+` here:
```suggestion
                    raise RuntimeError(f"The datafile {full_path} does not exist."
                                       " Try get_edgelist(fetch=True) to"
                                       " download the datafile")
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,212 @@
...
+
+        self._graph = cugraph.Graph(directed=self.metadata['is_directed'])
+        self._graph.from_cudf_edgelist(self._edgelist, source='src',
+                                       destination='dst')
+
+        return self._graph
+
+    def get_path(self):
+        """
+            Returns the location of the stored dataset file
```
rlratzel, line 145: This indent is different than the others.

python/cugraph/cugraph/tests/test_dataset.py
```diff
@@ -0,0 +1,173 @@
...
+from pathlib import Path
+from tempfile import NamedTemporaryFile, TemporaryDirectory
+from cugraph.experimental.datasets import (ALL_DATASETS)
+
+
+# =============================================================================
+# Pytest Setup / Teardown - called for each test function
+# =============================================================================
+def setup_function():
+    gc.collect()
```
rlratzel, line 28: I think we should remove this since it shouldn't be necessary for these tests.

rlratzel (commented):
I would also remove the `metadata/__init__.py` file if that directory does not contain any .py files.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -19,23 +19,30 @@
 
 
 class DefaultDownloadDir:
+    """
+    An object that is used to store the location of the dataset files. It is
+    shared across all dataset objects. Upon creation, it checks for the
+    existence of the environment variable RAPIDS_DATASET_ROOT_DIR. Otherwise,
+    defaults to <HOME>/.cugraph/datasets.
+    """
```
rlratzel, line 27: It might be better to just briefly describe the class and how it's intended to be used, then let other parts of the code/comments describe the specifics with the env var, etc.
```suggestion
    """
    Maintains the path to the download directory used by Dataset instances.
    Instances of this class are typically shared by several Dataset instances
    in order to allow for the download directory to be defined and updated by
    a single object.
    """
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -19,23 +19,30 @@
...
+                                         Path.home() / ".cugraph/datasets"))
 
     @property
-    def dd(self):
-        if self._dd is None:
-            self._dd = Path(os.environ.get("RAPIDS_DATASET_ROOT_DIR",
-                                           Path.home() / ".cugraph/datasets"))
-        return self._dd
+    def path(self):
+        if self._path is None:
```
rlratzel, line 34: Around here could be a good place where you describe how `path` is returned (eg. returns the current setting, and if not set returns the env var RAPIDS_DATASET_ROOT_DIR, else <home>/.cugraph/datasets)
Although, feel free to ignore if you think it's fairly self-documenting as-is.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -55,39 +62,29 @@ class Dataset:
...
-    # check for updates for download_dir location within obj
-    def __refresh_dd(self):
-        if self._dd != default_download_dir.dd:
-            self._dd = default_download_dir.dd
 
     def __download_csv(self, url):
-        if not os.path.isdir(self._dd):
-            os.makedirs(self._dd)
+        if not os.path.isdir(self._dl_path.path):
+            os.makedirs(self._dl_path.path)
```
rlratzel, line 78: I think you should be able to replace these lines with a single call:
```suggestion
        self._dl_path.path.mkdir(parents=True, exist_ok=True)
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -55,39 +62,29 @@ class Dataset:
...
 
         filename = self.metadata['name'] + self.metadata['file_type']
-        if os.path.isdir(self._dd):
+        if os.path.isdir(self._dl_path.path):
```
rlratzel, line 81: Might be better to use the methods on the `Path` object instead of `os.path` functions when possible:
```suggestion
        if self._dl_path.path.isdir():
```

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -101,17 +98,16 @@ def get_edgelist(self, fetch=False):
         """
 
         if self._edgelist is None:
-            self.__refresh_dd()
-            full_path = self._dd / (self.metadata['name'] +
-                                    self.metadata['file_type'])
+            full_path = self._dl_path.path / (self.metadata['name'] +
+                                              self.metadata['file_type'])
 
             if not os.path.isfile(full_path):
```
rlratzel, line 104: same here, you should be able to use `Path.is_file()`

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -159,20 +155,19 @@ def load_all(force=False):
     Parameters
         ----------
     """
-    if not os.path.isdir(default_download_dir.dd):
-        os.makedirs(default_download_dir.dd)
+    if not os.path.isdir(default_download_dir.path):
+        os.makedirs(default_download_dir.path)
```
rlratzel, line 159: same as above: you could use `Path.mkdir(parents=True, exist_ok=True)`

rlratzel (changes requested):
Just a few more minor requests

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+    provided in each YAML file.
+
+    Parameters
+    force : Boolean (default=False)
+        Overwrite any existing copies of datafiles.
+    """
+    default_download_dir.path.mkdir(parents=True, exist_ok=True)
+
+    meta_path = Path(__file__).parent.absolute() / "metadata"
+    for file in os.listdir(meta_path):
```
rlratzel, line 166: Here's another place you could use the `Path` methods:
```suggestion
    for file in meta_path.iterdir():
```
and `Path.iterdir()` returns `Path` objects.

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -0,0 +1,210 @@
...
+        meta = None
+        if file.endswith('.yaml'):
```
rlratzel, line 168: if you do the above suggestion, then this becomes:
```suggestion
        if file.suffix == '.yaml':
```

rlratzel (approved):

jnke2016 (approved):
Looks good to me

rlratzel:
closes #1348

PR #2370, opened 2022-06-23
[WIP][skip-ci] MG C++ test  bug fixes.

seunghwak:

PR #2371, opened 2022-06-23
Delete old nbr sampling software

ChuckHastings:
Time ran out during 22.06 development to replace the old code.

This PR renames the experimental version of uniform neighborhood sampling to the regular name and eliminates the original implementation.

Tagged as `breaking` since we eliminate a function from the C++ API.


seunghwak (approved):

BradReesWork:
rerun tests

cpp/src/sampling/detail/sampling_utils_sg.cu
```diff
@@ -84,6 +85,7 @@ template rmm::device_uvector<int64_t> get_global_adjacency_offset(
   graph_view_t<int64_t, int64_t, double, false, false> const& graph_view,
   rmm::device_uvector<int64_t> const& global_degree_offsets,
   rmm::device_uvector<int64_t> const& global_out_degrees);
+#endif
```
seunghwak, line 88: Better just delete?

seunghwak (approved):

cpp/src/sampling/detail/sampling_utils_sg.cu
seunghwak: And can I just close PR https://github.com/rapidsai/cugraph/pull/2370?

ChuckHastings: I think you can just close PR 2370.

I deleted both of these blocks of code (in _sg and _mg).  Thanks for catching that.

ChuckHastings:
@gpucibot merge

PR #2387, opened 2022-07-04
Pin cuda-python.

vyasr:
Recent changes to cuda-python lead to incompatibilities with the CTK available via conda.

If necessary, these changes can be backported to 22.06.


vyasr:
Looks like this was done by #2380.

PR #2388, opened 2022-07-05
Moving Centrality notebooks to new structure and updating/testing

acostadon:


BradReesWork (changes requested):
- [x] algorithms/README - copyright date is wrong
- [x] centrality README - make the image a little smaller (use the size tags)
- [x] for each of the sub-READMEs - you can drop the “RAPIDS notebooks” portion
- [ ] Betweenness notebook - karate image broken
- [ ] Betweenness - might be better to drop the “print top score” function and just use the “sort + head” call that is done next
- [ ] Katz - same as above
- [ ] Katz - the the Numpy version still needed?

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #2394, opened 2022-07-07
Update `Graph` to store a Pylibcugraph Graph (SG/MG Graph)

### LINKED ISSUE #2373: [FEA] Update cugraph Graph classes to maintain a pylibcugraph SG/MGGraph instance
The current cugraph algorithm APIs create a `pylibcugraph` MG/SGGraph instance as part of the algo call, then delete it prior to returning.  This was a short-term put in place to start using `pylibcugraph` from the high-level `cugraph` API, but it's not efficient.

The intended longer-term design called for the higher-level cugraph `Graph` classes to create an instance of a `pyibcugraph` MG/SGGraph, then re-use it for algo calls, eliminating the need to create and destroy those instances each time.

Considerations:
* `pylibcugraph.MG/SGGraph` instances need to be created with `store_transposed` either `True` or `False` based on the algo they're to be used on.  `cugraph` algo calls will need to check this and either raise an exception if the orientation is incorrect for the algo, or auto-convert.  auto-convert will result in a possibly unexpected performance hit, but might be the preferred behavior for some use cases.  We may need to add flags to algos and/or Graph classes for this.
* "lazily" creating the internal `pylibcugraph.MG/SGGraph` may also be related to the above
* `cugraph.Graph` classes could create the `pylibcugraph.MG/SGGraph` as part of their ctor, but questions around creating the `pylibcugraph.MG/SGGraph` instance transposed or not need to be answered, and/or if a user option is added to the `cugraph.Graph` ctor should be added for this.
* This issue is intended to precede #2374

---

  - [] Code passes CI
  - [] Code uses Graph Primitives
  - [] Code in C++/CUDA layer
  - [] Code in C layer
  - [] Code in cugraph Python layer
  - [] Documentation

alexbarghi-nv:
Closes #2373


alexbarghi-nv:
rerun tests

BradReesWork (approved):

rlratzel (commented):
Looks good, I just had some minor comments and questions.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -89,6 +46,24 @@ def convert_to_cudf(cp_arrays, weight_t):
     return df
 
 
+def _call_plc_uniform_neighbor_sample(sID,
```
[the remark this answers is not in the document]
alexbarghi-nv: yeah, I tried using a lambda expression but unfortunately that didn't seem to work.  It hanged indefinitely.

python/cugraph/cugraph/dask/traversal/bfs.py
```diff
@@ -147,27 +115,14 @@ def bfs(input_graph,
...
-    dst_col_name = input_graph.renumber_map.renumbered_dst_col_name
+    client = input_graph._client
 
     if not isinstance(start, (dask_cudf.DataFrame, dask_cudf.Series)):
         if not isinstance(start, (cudf.DataFrame, cudf.Series)):
             start = cudf.Series(start)
         if isinstance(start, (cudf.DataFrame, cudf.Series)):
             # convert into a dask_cudf
-            start = dask_cudf.from_cudf(start, ddf.npartitions)
+            start = dask_cudf.from_cudf(start, input_graph.npartitions)
```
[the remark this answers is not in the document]
alexbarghi-nv: sure, let me correct that

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -129,6 +170,37 @@ def __from_edgelist(
         self.source_columns = source
         self.destination_columns = destination
 
+        self.compute_renumber_edge_list(
+            transposed=False, legacy_renum_only=True
+        )
```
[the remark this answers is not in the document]
alexbarghi-nv: sure

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -751,7 +763,35 @@ def _degree(self, vertex_subset, direction=Direction.ALL):
 
         return df
 
-    def to_directed(self, DiG):
+    def make_plc_graph(self, value_col=None, store_transposed=False):
```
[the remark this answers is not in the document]
alexbarghi-nv: yes, that's correct.  Now that I think about it, I should probably have the same function for both graphs for consistency, and it should be protected

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -153,39 +121,45 @@ def uniform_neighbor_sample(input_graph,
...
-              for idx, wf in enumerate(data.worker_to_parts.items())]
+
+    '''
+    start_list = dask_cudf.from_cudf(
+        start_list,
+        npartitions=input_graph.npartitions
+    )
+    start_list = get_distributed_data(start_list)
+    wait(start_list)
+    '''
```
jnke2016, line 139: The part you commented is something we have to address in the future .
I had a discussion with Chuck regarding distributing the seeds(for performance reason) like `BFS` because our current implementation send a copy of the seeds to each rank which results in `n` sampling for `n` seeds:

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -153,39 +121,45 @@ def uniform_neighbor_sample(input_graph,
...
+            _call_plc_uniform_neighbor_sample,
+            Comms.get_session_id(),
+            mg_graph,
+            start_list,
+            fanout_vals,
+            with_replacement,
+            workers=[w],
+        )
+        for mg_graph, (w, _) in zip(
+            input_graph._plc_graph, dummy.worker_to_parts.items()
```
jnke2016, line 160: if the sole goal of `dummy` is just to get the workers addresses, you can just call `Comms.get_workers()` because that's what `get_distributed_data` does under the hood

jnke2016: May be I am overthinking but I am worried the worker initially carrying a partition might not be the one we are passing in `line 157` regardless if we get the worker addresses from `dummy` or `Comms.get_workers()`: The old implementation ensured it doesn't happen. This will still work but it might cause some performance regression because it defeat the purpose of data locality.

alexbarghi-nv: I'll make the change to Comms.get_workers().

I thought the API was consistent with the data being partitioned the same way each time.  But if it isn't I can store the workers along with each _plc graph, right?

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: Right, I know this is something we're working on.  I will make a note of that here.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
jnke2016: > I thought the API was consistent with the data being partitioned the same way each time
You are right. i just took a look and the API consistently [persist](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/dask/common/part_utils.py#L93) the data been partitioned. So each worker will persist the data in the order they appear in Comms.get_workers.

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -129,6 +170,37 @@ def __from_edgelist(
         self.source_columns = source
         self.destination_columns = destination
 
+        self.compute_renumber_edge_list(
+            transposed=False, legacy_renum_only=True
```
jnke2016, line 174: I second what Rick mentioned. Furthermore, having renumbering at the graph creation(`__from_edgelist`)  forces to add `legacy_renum_only ` to `__from_edgelist` because some algos (the legacy algo) require the C++ renumbering and others(pylibcugraph, DGL) don't.

jnke2016: Maybe I am missing something but I don't get why the edge list needs to be renumbered as `pylibcugraph` renumbers the edgelist internally when creating the `pylibcugraph` graph. Can you explain why?

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -129,6 +170,37 @@ def __from_edgelist(
...
+        )
+
+        self.properties.renumbered = self.renumber_map.implementation.numbered
+        ddf = self.edgelist.edgelist_df
+
+        num_edges = len(ddf)
+        edge_data = get_distributed_data(ddf)
+        src_col_name = self.renumber_map.renumbered_src_col_name
+        dst_col_name = self.renumber_map.renumbered_dst_col_name
+        graph_props = GraphProperties(is_multigraph=self.properties.multi_edge)
```
jnke2016, line 184: the parameter `is_symmetric` is missing in the `GraphProperty`. For instance it needs to be sets to `True` for `core_number` or `triangle_count` otherwise it causes a crash somewhere in the C/C++ API. I think it should be set to `self.properties.directed`.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -88,7 +94,8 @@ def __from_edgelist(
         destination="destination",
         edge_attr=None,
         renumber=True,
-        legacy_renum_only=False,
+        legacy_renum_only=True,
```
jnke2016, line 97: Setting this parameter to `True` skips the `C++ renumbering` and I don't think we want this by default.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -751,7 +763,35 @@ def _degree(self, vertex_subset, direction=Direction.ALL):
 
         return df
 
-    def to_directed(self, DiG):
+    def make_plc_graph(self, value_col=None, store_transposed=False):
+        if value_col is None:
+            value_col = cudf.Series(
+                cupy.ones(len(self.edgelist.edgelist_df), dtype='float32')
```
jnke2016, line 769: I like that you added the edge values column for edge lists that don't

alexbarghi-nv: well, I believe the plc graph creation fails otherwise...

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: I figured out the correct way to do this and I'm fixing it now.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
jnke2016: Right. So far we have been luck to process datasets that have edge values

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
alexbarghi-nv: I did that because the original version of simpleDistributedGraph said renumbering was always required.  But I'll drop it and see what happens.

alexbarghi-nv: I believe you are correct.  This was done just for consistency as some users may expect that the renumbering occurs.

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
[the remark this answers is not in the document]
alexbarghi-nv: Transposed shouldn't be hardcoded and legacy_renum_only should be set based on the provided arguments.  I'm fixing that now.

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
jnke2016: > I did that because the original version of simpleDistributedGraph said renumbering was always required'

Right,  renumbering([C++ renumbering](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/structure/number_map.py#L590)) is `still` required for the algos following the legacy path like `MG louvain`(once we move everything to `pylibcugraph`, we will drop the C++ renumbering). In fact, those algos require the [aggregate_segment_offset](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/structure/number_map.py#L650) which can only. be provided after computing the C++ renumbering

alexbarghi-nv: Yes, so what I'm doing is keeping renumber=True as the default and adding an option for legacy renumbering, set to False by default.  I assume any user modifying those options knows what they're doing and is aware of the API differences.

jnke2016: I added the `legacy_renum_only` to skip the `C++ renumbering` which is not needed for the `pylibcugraph` algos. It only does a `python` renumbering of the edgelist if the vertices are non integers. Sorry if the naming is confusing and should probably find a better name for that

jnke2016: > I did that because the original version of simpleDistributedGraph said renumbering was always required. But I'll drop it and see what happens

I think if you drop it, it should still be find. But probably let's have a discussion regarding it during sync with the rest of the team

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: Update: who_has doesn't seem to work right, but Comms.get_workers does.  I'm not sure why this is the case.

jnke2016: That's weird. Did you copy the script I shared above?

alexbarghi-nv: I did.

jnke2016: Got it. I will take a look at tit tomorrow

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
alexbarghi-nv: Legacy_renum_only=False appears to break the pylibcugraph algorithms...

jnke2016: Yes it needs to be set  to `True` for the `pylibcugraph` algos

alexbarghi-nv: Well this causes a problem, basically the graph is now only usable on one API or the other depending on whether or not it's set to true.

alexbarghi-nv: Credit to you for catching this; since all of our tests call only one of the APIs they did not fail.

jnke2016: > Well this causes a problem, basically the graph is now only usable on one API or the other depending on whether or not it's set to true

Got it. We can have a post stand up on it after sync.

jnke2016: I may have an idea of what is going on. Prior to this change, skipping the C++ algos was solely at the discretion of each algo which is no longer the case now. 

1. If you have `renumbering` in `__from_edgelist`, it will be skipped at the algo's call
2. We might also have to refactor the tests involved too.
I think the a solution could be do remove the renumbering in `__from_edgelist` but I am yet to test that

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -134,6 +135,15 @@ def from_cudf_edgelist(
...
             vertex IDs.
 
+        store_transposed : bool, optional (default=False)
+            If True, stores the transpose of the adjacency matrix.  Required
+            for certain algorithms.
+
+        legacy_renum_only : bool, optional (default=False)
+            If True, skips the C++ renumbering step.  Must be true for
+            pylibcugraph algorithms.  Must be false for algorithms
+            not yet converted to the pylibcugraph C API.
```
rlratzel, line 145: We should revisit this in a separate PR since I'm thinking most users won't understand it, but for now we'll want this PR merged to unblock others.

rlratzel (approved):

python/cugraph/cugraph/structure/graph_classes.py
rlratzel: I filed [this issue](https://github.com/rapidsai/cugraph/issues/2437) so we don't forget.

rlratzel:
@gpucibot merge

PR #2396, opened 2022-07-08
Run relevant CI tests based on what's changed in the ChangeList

anandhkb:
cpp_changed==cpp+python+notebooks, python_changed==python+notebooks, notebooks_changed==notebooks

### DIFF
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -140,6 +140,15 @@ do
    fi
 done
 ################################################################################
+# PRINT SUMMARY OF TESTS to BE EXECUTED
+################################################################################
+gpuci_logger "Summary of CI tests to be executed"
+gpuci_logger "Run cpp tests=$run_cpp_tests"
+gpuci_logger "Run python tests=$run_python_tests"
+gpuci_logger "Run notebook tests=$run_nb_tests"
+gpuci_logger "Docs updated=$doc_changed"
+
+################################################################################
 # TEST
 ################################################################################
 
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -122,6 +122,8 @@ fi
 ################################################################################
 
 fnames=()
+# Initialize all the run variables to true, as we want to run all the tests if the build_mode is NOT a pull-request
+run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
 if [ "$BUILD_MODE" == "pull-request" ]; then
     PR_ENDPOINT="https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files"
     fnames=(
@@ -134,11 +136,8 @@ if [ "$BUILD_MODE" == "pull-request" ]; then
       jq -r '.[].filename'
       )
     )
-    # Initialize all the run variables to false, later in the code below, based on what's changed, these variables will be set to true
+    # Initialize all the run variables to false, for pull-requests only, later, based on what's changed, these variables will be set to true
     run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
-else
-    # Initialize all the run variables to true, as we want to run all the tests if the build_mode is NOT a pull-request
-    run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
 fi
 # this will not do anything if the 'fnames' array is empty
 for fname in "${fnames[@]}"
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -134,8 +134,10 @@ if [ "$BUILD_MODE" == "pull-request" ]; then
       jq -r '.[].filename'
       )
     )
+    # Initialize all the run variables to false, later in the code below, based on what's changed, these variables will be set to true
     run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
 else
+    # Initialize all the run variables to true, as we want to run all the tests if the build_mode is NOT a pull-request
     run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
 fi
 # this will not do anything if the 'fnames' array is empty
--- a/ci/gpu/build.sh
+++ b/ci/gpu/build.sh
@@ -121,25 +121,31 @@ fi
 # Identify relevant testsets to run in CI based on the ChangeList
 ################################################################################
 
-run_cpp_tests="false" run_python_tests="false" run_nb_tests="false" doc_changed="false" 
 fnames=()
 if [ "$BUILD_MODE" == "pull-request" ]; then
-    PRHTTP=https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files
-    fnames=`curl -sb -X GET -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $GHTK"  $PRHTTP | python3 -c "import sys, json; print([labels['filename'] for labels in json.load(sys.stdin)])"`
+    PR_ENDPOINT="https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files"
+    fnames=(
+      $(
+      curl -s \
+      -X GET \
+      -H "Accept: application/vnd.github.v3+json" \
+      -H "Authorization: token $GHTK" \
+      "$PR_ENDPOINT" | \
+      jq -r '.[].filename'
+      )
+    )
+    run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
 else
     run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
 fi
 # this will not do anything if the 'fnames' array is empty
-for fname in ${fnames[@]}
+for fname in "${fnames[@]}"
 do
    if [[ "$fname" == *"cpp/cmake/"* || "$fname" == *"cpp/CMakeLists.txt"* || "$fname" == *"cpp/src/"* || "$fname" == *"cpp/include/"* || "$fname" == *"cpp/tests/"* || "$fname" == *"cpp/libcugraph_etl/"* || "$fname" == *"cpp/scripts/"* ]]; then
-      run_cpp_tests="true" run_python_tests="true" run_nb_tests="true" doc_changed="true"
+      run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
    fi
    if [[ "$fname" == *"python/"* ]]; then
-      run_python_tests="true" run_nb_tests="true" doc_changed="true"
-   fi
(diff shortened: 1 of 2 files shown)


ci/test.sh
```diff
@@ -26,6 +26,31 @@ GTEST_ARGS="--gtest_output=xml:${CUGRAPH_ROOT}/test-results/"
...
+      nb_changed="true"
+   fi
+done
+
+if [[ "$nb_changed" == "false" ]]; then
+   export NB_CHANGED="FALSE"
+else
+   export NB_CHANGED="TRUE"
+fi
+
```
rlratzel, line 53: I think this code would be more appropriate in `ci/gpu/build.sh` and have the variables exported and then referenced in the lower-level scripts (like you're doing for the `test-notebooks.sh` script).  The main reason being that `test.sh` is sometimes also used by devs outside of CI, and this code would prevent that due to how it's trying to access info about a PR.  It would also mean you wouldn't have to `source test.sh` (which can sometimes have unintended side-effects) since the variables would be set in the top-level calling script.

Once this code is moved out, `test.sh` should be written to also handle cases where the various `python_changed`, `nb_changed`, etc. variables are not set, so devs can continue to run `test.sh` outside of CI.

ci/test.sh
```diff
@@ -26,6 +26,31 @@ GTEST_ARGS="--gtest_output=xml:${CUGRAPH_ROOT}/test-results/"
...
+PRHTTP=https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files
+fnames=`curl -sb -X GET -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $GHTK"  $PRHTTP | python3 -c "import sys, json; print([labels['filename'] for labels in json.load(sys.stdin)])"`
+cpp_changed="false" python_changed="false" nb_changed="false" doc_changed="false" 
+for fname in ${fnames[@]}
+do
+   if [[ "$fname" == *"cpp/cmake/"* || "$fname" == *"cpp/CMakeLists.txt"* || "$fname" == *"cpp/src/"* || "$fname" == *"cpp/include/"* || "$fname" == *"cpp/tests/"* || "$fname" == *"cpp/libcugraph_etl/"* || "$fname" == *"cpp/scripts/"* ]]; then
+      cpp_changed="true" python_changed="true" nb_changed="true" doc_changed="true"
+   fi
+   if [[ "$fname" == *"python/"* ]]; then
+      python_changed="true" nb_changed="true" doc_changed="true"
```
rlratzel, line 38: It might be better to make the variable names a little more accurate, since `python_changed` implies python code changed when it could actually be set by way of a C++ code change.  Perhaps the naming convention could be `run_python_tests`, `run_nb_tests`, etc.?

ci/gpu/build.sh
```diff
@@ -146,13 +146,15 @@ else
...
-    python ${WORKSPACE}/ci/utils/nbtestlog2junitxml.py nbtest.log
+    if [[ $NB_CHANGED == "TRUE" ]]; then
+        gpuci_logger "Running cuGraph notebook test script..."
+        ${WORKSPACE}/ci/gpu/test-notebooks.sh 2>&1 | tee nbtest.log
+        gpuci_logger "Ran cuGraph notebook test script : return code was: $?, gpu/build.sh exit code is now: $EXITCODE"
+        python ${WORKSPACE}/ci/utils/nbtestlog2junitxml.py nbtest.log
+    fi
 fi
 
 if [ -n "${CODECOV_TOKEN}" ]; then
```
rlratzel, line 160: I believe the codecov step below will fail if python tests aren't run, since the python tests generate the output needed by codecov.  You might try adding something to this `if` block to prevent codecov from running if python tests did not run.

ci/test.sh
```diff
@@ -26,6 +26,31 @@ GTEST_ARGS="--gtest_output=xml:${CUGRAPH_ROOT}/test-results/"
...
+   fi
+done
+
+if [[ "$nb_changed" == "false" ]]; then
+   export NB_CHANGED="FALSE"
+else
+   export NB_CHANGED="TRUE"
+fi
+
 export RAPIDS_DATASET_ROOT_DIR=${RAPIDS_DATASET_ROOT_DIR:-${CUGRAPH_ROOT}/datasets}
```
rlratzel, line 54: Based on the CI failures, it looks like this isn't being set properly beause `CUGRAPH_ROOT` isn't being set properly, because (I think) this script is now being `source`'d .  I think my other suggestion would eliminate the need to `source` this script and should fix that problem.

rlratzel (changes requested):
Thanks for doing this.  I have a few suggestions and change requests to hopefully fix the CI failures.

ci/test.sh
rlratzel, line 53: Somewhere around here might be a good place to print a summary of what's going to be run, what's going to be skipped, and (most importantly) why, since it might be confusing to devs who are expecting certain tests to run which end up getting skipped.

rlratzel (changes requested):

ci/test.sh
anandhkb: Got it, thanks Rick

ci/test.sh
anandhkb: Make sense

ci/gpu/build.sh
anandhkb: Got it, thanks Rick

ci/test.sh
anandhkb: Sounds good

ci/test.sh
anandhkb: Fair ask, will do

ci/gpu/build.sh
```diff
@@ -117,6 +117,37 @@ else
     gpuci_mamba_retry install cugraph pylibcugraph -c ${CONDA_BLD_DIR} -c ${CONDA_ARTIFACT_PATH}
 fi
 
+################################################################################
+# Identify relevant testsets to run in CI based on the ChangeList
+################################################################################
+
+PRHTTP=https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files
```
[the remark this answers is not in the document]
anandhkb: Thanks @ajschmidt8 , how to distinguish the nightly builds from the PR builds then? Based on that, I can add an 'if condition' and hide the new code under that 'if block'.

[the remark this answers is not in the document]
anandhkb: if [ "$BUILD_MODE" = "pull-request" ]; then
<new code>
fi

ajschmidt8 (changes requested):
Left some suggestions. Also, I don't see the `doc_changed` value being used anywhere throughout the PR. Can this be removed?

ci/gpu/build.sh
```diff
@@ -117,6 +117,43 @@ else
...
+################################################################################
+
+run_cpp_tests="false" run_python_tests="false" run_nb_tests="false" doc_changed="false" 
+fnames=()
+if [ "$BUILD_MODE" == "pull-request" ]; then
+    PRHTTP=https://api.github.com/repos/rapidsai/cugraph/pulls/${PR_ID}/files
+    fnames=`curl -sb -X GET -H "Accept: application/vnd.github.v3+json" -H "Authorization: token $GHTK"  $PRHTTP | python3 -c "import sys, json; print([labels['filename'] for labels in json.load(sys.stdin)])"`
+else
+    run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
+fi
```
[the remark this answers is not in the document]
anandhkb: @ajschmidt8 @rlratzel 
RE: "First, the else statement can be removed as long as the proper defaults values are set above the if statement"

That's wrong logic. The logic we want to implement is:
a. If the build_mode is NOT a pull-request, then run ALL tests (cpp_tests = true, python_tests=true, nb_tests=true)
b. Else (build_mode is a pull-request), then initialize cpp_tests = false, python_tests=false, nb_tests=false
c. Walk through the filenames changed by the ChangeList and based on what is changed, update the cpp_tests/python_tests/nb_tests variables to 'true'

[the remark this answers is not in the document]
ajschmidt8: @anandhkb, I understand what you're saying. the two example code blocks below are the same however. We can eliminate the `else` statement by setting the default values for the `run_*_tests` variables above the `if` block.

```sh
if [ "$BUILD_MODE" == "pull-request" ]; then
    run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
else
    run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
fi
```

```sh
run_cpp_tests="true" run_python_tests="true" run_nb_tests="true"
if [ "$BUILD_MODE" == "pull-request" ]; then
    run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
fi
```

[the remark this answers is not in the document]
anandhkb: Yes, you are absolutely right @ajschmidt8 and my latest update addresses your point.

rlratzel (approved):
Looks good. I have one suggestion but it need not hold up my approval.

ajschmidt8 (approved):
Approved pending the resolution of Rick's comment

ci/gpu/build.sh
```diff
@@ -117,6 +117,49 @@ else
...
+      jq -r '.[].filename'
+      )
+    )
+    # Initialize all the run variables to false, for pull-requests only, later, based on what's changed, these variables will be set to true
+    run_cpp_tests="false" run_python_tests="false" run_nb_tests="false"
+fi
+# this will not do anything if the 'fnames' array is empty
+for fname in "${fnames[@]}"
+do
+   if [[ "$fname" == *"cpp/cmake/"* || "$fname" == *"cpp/CMakeLists.txt"* || "$fname" == *"cpp/src/"* || "$fname" == *"cpp/include/"* || "$fname" == *"cpp/tests/"* || "$fname" == *"cpp/libcugraph_etl/"* || "$fname" == *"cpp/scripts/"* ]]; then
```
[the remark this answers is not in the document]
rlratzel: I would personally err on the side of running the tests rather than skipping them when in doubt. If you enumerate every subdir and file that triggers testing, and we add a subdir or file that isn't listed, the tests won't get run.  By explicitly listing the cpp dir then only items to exclude, you don't risk missing test runs on newly added source dirs or files (for instance, in the past few months we added libcugraph_etl). Missing test runs could potentially go unnoticed for a while in that situation and lead to a broken upstream.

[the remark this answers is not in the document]
anandhkb: You are correct Rick and I have update the commit based on your advise

anandhkb:
@gpucibot merge

PR #2409, opened 2022-07-14
[REVIEW] Optimize has_duplicate_edges 

VibhuJawa:
This PR fixes drop duplicates scalability by removing apply which does serial processing. 

**Benchmark Data**
```python
n_nodes = 100_000 
n_rows  = 1_500_000
df = cudf.DataFrame({'src':cp.random.randint(0,n_nodes,n_rows),
                    'dst':cp.random.randint(0,n_nodes,n_rows)})
```

**### After PR:**
```python
17.8 ms ± 536 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

**### Before PR:**
```python
26.3 s ± 78.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/python/cugraph/cugraph/dask/structure/mg_property_graph.py
+++ b/python/cugraph/cugraph/dask/structure/mg_property_graph.py
@@ -654,7 +654,7 @@ class EXPERIMENTAL__MGPropertyGraph:
         if len(df.columns) == 0:
             return False
 
-        unique_pair_len = df.drop_duplicates(split_out=df.npartitions).shape[0]
+        unique_pair_len = df.drop_duplicates(split_out=df.npartitions, ignore_index=True).shape[0]
         # if unique_pairs == len(df)
         # then no duplicate edges
         return unique_pair_len != df.shape[0]
--- a/python/cugraph/cugraph/dask/structure/mg_property_graph.py
+++ b/python/cugraph/cugraph/dask/structure/mg_property_graph.py
@@ -656,10 +656,14 @@ class EXPERIMENTAL__MGPropertyGraph:
             return False
 
         def has_duplicate_dst(df):
-            return df[cls.dst_col_name].nunique() != \
-                df[cls.dst_col_name].size
-
-        return df.groupby(cls.src_col_name).apply(has_duplicate_dst).any()
+        # may need to split out for dask frames
+        # to scale
+        unique_pair_len = len(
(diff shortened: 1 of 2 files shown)


BradReesWork (approved):

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -650,16 +650,18 @@ def has_duplicate_edges(cls, df):
...
-        def has_duplicate_dst(df):
-            return df[cls.dst_col_name].nunique() != \
-                df[cls.dst_col_name].size
-
-        return df.groupby(cls.src_col_name).apply(has_duplicate_dst).any()
+        # may need to split out for dask frames
+        # to scale
+        unique_pair_len = len(
+            df[[cls.src_col_name, cls.dst_col_name]].drop_duplicates()
+        )
```
rlratzel, line 661: Just curious, since I think the runtime performance seems worth it, but is there a tradeoff here in the temporary DataFrame using potentially more memory than the groupby object returned from `groupby()`?

rlratzel (approved):

seunghwak:
Just FYI: I am not sure this is relevant here, but cuGraph C++ graph_view_t has count_multi_edges() (https://github.com/rapidsai/cugraph/blob/branch-22.08/cpp/include/cugraph/graph_view.hpp#L680).

python/cugraph/cugraph/dask/structure/mg_property_graph.py
VibhuJawa: Thanks for  pointing out the problem around the memory usage . We indeed seem to use more memory here as we now store `unique dst and src combinations` while previously we were only saving  `unique src`  columns.  

That said, with dask i am not too worried as we will release it for each partition as we create it so it should not be too much. 

Also previous to this review we were saving index too, now with `ignore_index=True` we wont save that so that should save on some memory.  Thanks for that.

rlratzel:
@gpucibot merge

VibhuJawa:
> Just FYI: I am not sure this is relevant here, but cuGraph C++ graph_view_t has count_multi_edges() (https://github.com/rapidsai/cugraph/blob/branch-22.08/cpp/include/cugraph/graph_view.hpp#L680).

@seunghwak , Thanks for the input , raised an issue here to track it. https://github.com/rapidsai/cugraph/issues/2417

PR #2419, opened 2022-07-16
Add options to `extract_subgraph()` to bypass renumbering and adding edge_data, exclude internal `_WEIGHT_` column from `edge_property_names`

rlratzel:
Add options to `extract_subgraph()` to bypass renumbering and adding edge_data, exclude internal `_WEIGHT_` column from `edge_property_names`

Added unit tests to verify new `extract_subgraph()` options work, and `_WEIGHT_` columns names aren't included, for both SG and MG versions.

closes #2418 
closes #2410

PR #2434, opened 2022-07-20
Add get_num_vertices and get_num_edges methods to PropertyGraph.

### LINKED ISSUE #2422: [FEA] API to get `num_nodes` and `num_edges` per `type` from the PropertyGraph

API to get `num_nodes` and `num_edges` per `type` from the PropertyGraph

```python
def num_nodes(type=None)
```

```python

def num_edges(type=None)

```

eriknw:
Closes #2422.  I'll add this to MGPG when we finalize the API and behavior.


eriknw:
This keeps `pG.num_vertices` and`pG.num_edges` as properties, which I think is convenient, and adds `pG.get_num_vertices(type)` and `pG.get_num_edges(type)`.

If `type` is `None`, should we return the same as e.g. `pG.num_vertices`, or could we use `dataframe[self.type_col_name].value_counts()` to return the number for _all_ types.

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -195,6 +195,46 @@ def _vertex_prop_dataframe(self):
...
+
+        See Also
+        --------
+        PropertyGraph.num_vertices
+        PropertyGraph.get_num_edges
+        """
+        if type is None:
+            return self.num_vertices
+        if self.__vertex_prop_dataframe is None:
+            return 0
```
BradReesWork, line 214: this should call             return self.num_vertices

BradReesWork (changes requested):

python/cugraph/cugraph/structure/property_graph.py
eriknw: I don't think so.  My understanding is that vertices only referred to in the edge table don't have a type associated with them.  It would be surprising and probably incorrect to use edge type as vertex type.  So, if there is no vertex property table, then there are no typed vertices, hence 0.

rlratzel (changes requested):
@eriknw and I had a conversation offline and we decided that it would be better to keep the API surface area smaller and minimize confusion by getting rid of the `num_vertices` and `num_edges` python property attributes, and just used the new getters here.  The new getters should also support the caching behavior the python property attributes had.

python/cugraph/cugraph/tests/test_property_graph.py
```diff
@@ -338,7 +352,13 @@ def test_num_vertices(df_type):
                        property_columns=None)
 
     assert pG.num_vertices == 9
+    assert pG.get_num_vertices() == 9
+    assert pG.get_num_vertices('merchants') == 5
+    assert pG.get_num_vertices('users') == 4
+    assert pG.get_num_vertices('taxpayers') == 7  # These nodes have two types
```
rlratzel, line 358: TL;DR - [this PR](https://github.com/rapidsai/cugraph/pull/2419) will remove "taxpayers", so it might be easier to just remove these assertions here too.

There was an offline discussion (sorry for not including the outcome on GH) where it was decided that the "taxpayer" input was considered illegal input, since it results in the same vertex showing up as two different types.  I think we decided that users should format the input to have a 1:1 for types and vertex IDs, and entities that show up with multiple properties (eg. "merchants" and "taxpayers") should format their input to have bool columns such as "is_merchant" and "is_taxpayer" instead, for example.

alexbarghi-nv: PG should eventually enforce vertex id uniqueness, but that should not hold up this PR.

alexbarghi-nv (approved):
👍

BradReesWork (approved):

rlratzel (approved):

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -195,6 +166,33 @@ def edge_property_names(self):
             return props
         return []
 
+    @property
```
rlratzel, line 169: Apologies for not bringing this up sooner, but I'm now wondering if we can take advantage of [@cached_property](https://docs.python.org/3/library/functools.html?highlight=cached_property#functools.cached_property) here and elsewhere.  In this case, whenever `add_vertex_data` is called, the `vertex_types` attr would be deleted to cause this code to run and cache on the new data.

The downside is that @cached_property allows writes, but there's an easy and fairly clean workaround for that.
Thoughts?
```python
>>> from functools import cached_property
>>> class Foo:
...    some_expensive_computation_result = 33
...    @cached_property
...    def __foo(self):
...       print("running")
...       return self.some_expensive_computation_result
...    @property
...    def foo(self):
...       return self.__foo
...    def update_data(self):
...       del self.__foo
...
>>> f=Foo()
>>> f.foo
running
33
>>> f.foo
33
>>> f.foo=22
  File "<stdin>", line 1, in <module>
AttributeError: can't set attribute
>>> f.update_data()
>>> f.foo
running
33
>>> f.foo
33
```

rlratzel (commented):
Looks good, I had one suggestion which I was wondering about.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
eriknw: Good question.  I would only use `@cached_property` for truly immutable results.  Since we need to clear the property, I don't think using `@cached_property` buys us anything, since it's still complex.  I also prefer to avoid adding or removing keys to an objects `__dict__` so that we get the advantages of [PEP 412](https://peps.python.org/pep-0412/).

Perhaps we should try to be more consistent with how we cache properties.

BradReesWork:
@gpucibot merge

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: You convinced me with PEP 412 and the fact that we have to `del self.__prop` to clear it which would constantly change the instance keys.

I do think that we should eventually (soon?) look into a general caching pattern of some kind for PG (and possibly others), since the usage pattern seems to be 1. add data in one or several add steps, 2. access data often. We do support adding data after accessing it of course, but I wouldn't want to penalize the most common case (at least as we've seen so far for GNNs) by not having any caching at all.

PR #2439, opened 2022-07-22
Refactor MG C++ tests (handle initialization)

seunghwak:
Update MG C++ tests to initialize a RAFT handle_t object in SetUpTestCase().

Previously, a handle_t object is initialized inside individual tests. If an exception is thrown inside a single test, a stack unwinding process starts, and this calls handle_t's destructor, and the destructor invokes ncclCommDestroy(). If only a subset of workers fail, some processes will still advance and can be blocked on a NCCL call, This bars ncclCommDestroy() from finishing, and the stack unwinding process will hang. The outcome is no output in the console and testers may have no clue that exceptions are thrown.

With this restructuring, exceptions in individual tests will not destroy a handle_t object, so testers will see the exception messages.


ChuckHastings (approved):

kaatish (approved):

jnke2016 (approved):

BradReesWork:
@gpucibot merge

PR #2441, opened 2022-07-25
Use Datasets API to Update Docstring Examples

### LINKED ISSUE #2361: [ENH] `datasets` API: Update docstrings
Milestone for https://github.com/rapidsai/cugraph/issues/1348

Update all Python docstrings to use the updated import pattern. This also includes updating test_docstrings.

oorliu:
closes #2361

Docstring examples now use the new method of creating graphs by using the `datasets` API. This change cleans up the code by eliminating the usage of `cuDF`.

Old docstring example:
```
>>> M = cudf.read_csv(datasets_path / 'karate.csv', delimiter=' ',
...                   dtype=['int32', 'int32', 'float32'], header=None)
>>> G = cugraph.Graph()
>>> G.from_cudf_edgelist(M, source='0', destination='1')
```

Updated docstring example:
```
>>> from cugraph.experimental.datasets import karate
>>> G = karate.get_graph()
```


rlratzel:
I only see an update to bfs.py in this PR. Shouldn't the other docstrings be updated the same way, or have they been updated elsewhere in another PR?

rlratzel:
rerun tests

https://github.com/rapidsai/cugraph/pull/2473 was merged which should resolve test failures.

python/cugraph/cugraph/centrality/betweenness_centrality.py
```diff
@@ -106,10 +106,8 @@ def betweenness_centrality(
 
     Examples
     --------
-    >>> gdf = cudf.read_csv(datasets_path / 'karate.csv', delimiter=' ',
```
rlratzel, line 109: I'm thinking we can remove [this line](https://github.com/rapidsai/cugraph/blob/2a57740c8eac484ddd1a200f0e93f0dbb4062679/python/cugraph/cugraph/tests/test_doctests.py#L153) now too, since the docstrings won't be using the `dataset_path` variable.  Removing that will also make sure we didn't forget anything, since the old docstring pattern will fail.

oorliu: There are some docstrings that still rely on this path and I believe shouldn't be replaced by the datasets API. For example, there are some MG examples in the `cugraph/dask` directory, as well as functions like `from_cudf_edgelist` in `cugraph/structures` that demonstrate more granular behavior that shouldn't be overwritten by the datasets API

oorliu: Removing the line in `test_doctests.py` causes these tests to fail unless the `datasets_path` is given to them elsewhere, or included in the respective docstring example.

rlratzel: > functions like from_cudf_edgelist in cugraph/structures that demonstrate more granular behavior that shouldn't be overwritten by the datasets API

That's a good reason to keep it.

rlratzel:
@gpucibot merge

rlratzel (approved):

PR #2444, opened 2022-07-26
Added `get_vertex_data()` and `get_edge_data()` to SG/MG PropertyGraph

### LINKED ISSUE #2421: [FEA] API for PropetyGraphs for get_edge_data and get_vertex_data

We should create an API for PropetyGraphs to get_edge_data and get_vertex_data 

Requested API :

```python
def get_edge_data(edge_ids, columns, etype):
     Parameters:
              edge_ids (cudf.Series) 
              columns (list[string])  (Can include `SRC` and `DST` columns)
              etype: String 
    Returns:
             df (cudf.DataFrame): Dataframe corresponding to the `edge_ids` with the corresponding  `columns` 
```

```python
def get_vertex_data(vertex_ids, columns, vtype):
      Parameters:
              vertex_ids (cudf.Series) 
              col_names (list[string]) 
              vtype: String 
     
     Returns:
             df (cudf.DataFrame): Dataframe corresponding to the `vertex_ids` with the corresponding `columns` 
```

rlratzel:
closes #2421 

Added `get_vertex_data()` and `get_edge_data()` to SG and MG PropertyGraph, and corresponding tests.

Prior to these methods, users had to either call `pG.annotate_dataframe()` to get properties for edges or access the internal debug dataframes directly via `pG._vertex_prop_dataframe` and `pG._edge_prop_dataframe`.

Users can now call `pG.get_vertex_data(vertex_ids, types, columns)` to get vertex properties for the vertices specified by `vertex_ids`, and 'types', with data for each column specified. All args are optional and default to "all" for each category.  `pG.get_edge_data(edge_ids, types, columns)` works the same for edges.  The return value for both is a dataframe.


python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -331,6 +331,34 @@ def add_vertex_data(self,
...
 
+    def get_vertex_data(self, vertex_ids=None, types=None, columns=None):
+        """
+        Return a dataframe containing vertex properties for only the specified
+        vertex_ids, columns, and/or types, or all vertex IDs if not specified.
+        """
+        if self.__vertex_prop_dataframe is not None:
+            if vertex_ids is not None:
+                df_mask = self.__vertex_prop_dataframe[self.vertex_col_name]\
+                              .isin(vertex_ids)
```
eriknw, line 342: Style-wise, I prefer to not use `\`.  Instead, method chaining like this should be done using `(` and `)` as e.g.
```python
                df_mask = (
                    self.__vertex_prop_dataframe[self.vertex_col_name]
                    .isin(vertex_ids)
                )
```
I think this is how `black` would reformat this.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -331,6 +331,34 @@ def add_vertex_data(self,
...
+            if types is not None:
+                # FIXME: coerce types to a list-like if not?
+                df_mask = df[self.type_col_name].isin(types)
+                df = df.loc[df_mask]
+
+            if columns is None:
+                # Note: this includes the "internal" pG.vertex_col_name and
+                # pG.type_col_name columns, since they are assumed to be needed
+                # by the caller.
+                columns = self.__vertex_prop_dataframe.columns
```
eriknw, line 356: Does this just use all the columns?  Why is this necessary?

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -331,6 +331,34 @@ def add_vertex_data(self,
...
+                df = df[columns]
```
eriknw, line 357: Is this the correct indentation?  Did you mean to dedent it so it always runs?  For example, what happens if the user-provided `columns` is not None?

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: Thanks for this suggestion. I made this change and I also prefer it now that I've seen it.  I'll be keeping this in mind in the future.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: I think I copied the pattern from `get_edge_data` that did some additional handling of the column names without noticing it wasn't necessary here. I've changed this here and in `get_edge_data`.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: Thank you for noticing this. It's definitely a bug, and it also exposed a problem in my tests since they didn't catch it.
I updated my tests, verified that they caught this problem, updated the code here and in `get_edge_data` since it also had the same problem, then verified that the new tests passed.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -667,13 +734,14 @@ def edge_props_to_graph(self,
...
+        # renumbering.  The workaround is to always pass renumber=True, but set
+        # legacy_renum_only based on if actual renumbering (ie. changing the
+        # vertex IDs) should happen or not.
+        renumber = True
         if renumber_graph is False:
-            renumber = True
+            legacy_renum_only = True
         else:
-            renumber = renumber_graph
+            legacy_renum_only = False
```
alexbarghi-nv, line 744: I think most of the plc algorithms fail with a memory error if legacy_renum_only=False.  Should we just always make it true for now since the property graph is really only being used for GNNs at this point?

alexbarghi-nv (changes requested):

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -411,22 +443,20 @@ def add_edge_data(self,
...
-        # FIXME: Find a better way to create the edge id
-        prev_eid = -1 if self.__last_edge_id is None else self.__last_edge_id
         tmp_df[self.src_col_name] = tmp_df[vertex_col_names[0]]
         tmp_df[self.dst_col_name] = tmp_df[vertex_col_names[1]]
+        tmp_df[self.type_col_name] = type_name
+
+        # Add unique edge IDs to the new rows
+        prev_eid = -1 if self.__last_edge_id is None else self.__last_edge_id
         starting_eid = prev_eid + 1
         data_size = len(tmp_df.compute().index)
```
VibhuJawa, line 453: 1. To get the the length we should not call compute on the whole dataframe. I recommend we switch the data_size to :

```python 
len(tmp_df.index)
```

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -411,22 +443,20 @@ def add_edge_data(self,
...
-            dask_cudf.from_cudf(cudf_series, self.__num_workers)
-        dask_series = dask_series.reset_index(drop=True)
-        self.__last_edge_id = starting_eid + data_size
-        tmp_df = tmp_df.reset_index(drop=True)
-        tmp_df[self.edge_id_col_name] = dask_series
-        tmp_df[self.type_col_name] = type_name
+        new_eids = cudf.Series(range(starting_eid, starting_eid + data_size))
+        new_eids_dask = dask_cudf.from_cudf(new_eids, self.__num_workers)
+        self.__last_edge_id = starting_eid + data_size - 1
+        tmp_df[self.edge_id_col_name] = new_eids_dask
```
VibhuJawa, line 457: I think we can just do this via below API and avoid the `cuDF` and client calls and also keeps it delayed. 

```python
tmp_df[self.edge_id_col_name] = 1
tmp_df[self.edge_id_col_name] = tmp_df[self.edge_id_col_name]].cumsum()+starting_eid 

self.__last_edge_id = starting_eid + data_size - 1
```

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -442,9 +475,17 @@ def add_edge_data(self,
...
         tmp_df[self.dst_col_name] = tmp_df[vertex_col_names[1]]
-        # FIXME: handle case of a type_name column already being in tmp_df
         tmp_df[self.type_col_name] = type_name
 
+        # Add unique edge IDs to the new rows
+        prev_eid = -1 if self.__last_edge_id is None else self.__last_edge_id
+        starting_eid = prev_eid + 1
+        data_size = len(tmp_df.index)
+        new_eids = self.__series_type(range(starting_eid,
+                                            starting_eid + data_size))
```
VibhuJawa, line 485: I suggest using cupy.arange instead to keep the compute on GPUs. 
```
cp.arange(starting_eid, starting_eid + data_size)
```

VibhuJawa (changes requested):
Have requested changes to avoid cudf/client side calls. 

Thanks for working on this.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: Thanks to a long offline conversation with @alexbarghi-nv and @jnke2016 , I think this is resolved. Please review though to make sure I got it right.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: Done, thanks.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
rlratzel: This is much better, thanks. I had to change it slightly to keep the edge IDs starting at 0, but all tests are passing. I also used this same pattern for SG to better accommodate a refactoring where SG and MG are sharing the core implementation more (cc @eriknw) and SG tests are also passing.

VibhuJawa (approved):
LGTM to me.

rlratzel:
rerun tests

There was a compile failure from a cython-generated cpp file, so rerunning while I investigate in case it's somehow a transient failure on the CI side.

alexbarghi-nv (approved):
👍

eriknw (approved):
LGTM 🚀

BradReesWork:
@gpucibot merge

BradReesWork (approved):

PR #2447, opened 2022-07-26
Fix pandas SettingWithCopyWarning, which really shouldn't be ignored.

eriknw:
See: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy

This implementation requires column names to be strings, which is hopefully fine (but could be changed).

The downside is that this sometimes copies data when unnecessary.  It's pretty hard to know when objects are views or not when using public APIS.  Pandas objects have a `._is_view` attribute that tell us this.  I don't know about cugraph, or if this is even an issue in cugraph.

CC @rlratzel

### DIFF
--- a/python/cugraph/cugraph/tests/test_property_graph.py
+++ b/python/cugraph/cugraph/tests/test_property_graph.py
@@ -118,6 +118,21 @@ def setup_function():
 # =============================================================================
 # Pytest fixtures
 # =============================================================================
+@pytest.fixture(scope="function", autouse=True)
+def raise_on_pandas_warning():
+    """Raise when pandas gives SettingWithCopyWarning warning"""
+    # Perhaps we should put this in pytest.ini, pyproject.toml, or conftest.py
+    import warnings
+
+    filters = list(warnings.filters)
+    warnings.filterwarnings(
+        "error",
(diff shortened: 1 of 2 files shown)


rlratzel (approved):

python/cugraph/cugraph/tests/test_property_graph.py
```diff
@@ -118,6 +118,21 @@ def setup_function():
...
+
+    filters = list(warnings.filters)
+    warnings.filterwarnings(
+        "error",
+        category=pd.core.common.SettingWithCopyWarning
+    )
+    yield
+    warnings.filters = filters
+
+
```
eriknw, line 135: pytest does some weird things with warnings between each test, but this fixture works and is only applied to this file (and I don't know if we need to reset the filters).  If we want to make this warning raise _everywhere_, there are better ways to achieve it.

Without this fixture, a way to make pytest raise on this warning is:
```
pytest -Werror::pandas.core.common.SettingWithCopyWarning cugraph/cugraph/tests/test_property_graph.py
```

BradReesWork:
@gpucibot merge

PR #2453, opened 2022-07-27
Datasets API Update: Add Extra Params and Improve Testing

oorliu:
Adding two new parameters to the `get_graph()` method within the datasets API.

- `default_direction` allows users to get only undirected graph objects.
- `weights` will specify whether or not the cugraph.Graph object has an `edge_attr` field.

`test_dataset.py` has been updated to:
1. Add coverage for the new parameters
2. Removing fetching datasets from the web when it's unnecessary. 

Docstrings have also been updated for clarity.


python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -118,22 +117,35 @@ def get_edgelist(self, fetch=False):
...
+        weights : Boolean (default=True)
+            Include weights.
         """
         if self._edgelist is None:
             self.get_edgelist(fetch)
 
         self._graph = cugraph.Graph(directed=self.metadata['is_directed'])
-        self._graph.from_cudf_edgelist(self._edgelist, source='src',
-                                       destination='dst')
+        if (len(self.metadata['col_names']) > 2 and weights):
```
rlratzel, line 140: If a user requests weights by passing `weights=True`, but the metadata doesn't have >2 columns, this will silently not add weights which might be confusing. 

Perhaps the arg should be `ignore_weights` which defaults to False. This is subtle but I think it helps make it more obvious that weighted graphs are returned whenever possible.

In either case, the docstring above should mention that weights will not be present if they aren't included in the dataset.

```suggestion
        if (len(self.metadata['col_names']) > 2 and not(ignore_weights)):
```

rlratzel (changes requested):

python/cugraph/cugraph/experimental/datasets/dataset.py
```diff
@@ -126,27 +126,37 @@ def get_graph(self, fetch=False, default_direction=True, weights=True):
...
+        create_using: cugraph.Graph (instance or class), optional
+        (default=Graph)
+            Specify the type of Graph to create. Can pass in an instance to
+            create a Graph instance with specified 'directed' attribute.
 
-        weights : Boolean (default=True)
-            Include weights.
+        ignore_weights : Boolean (default=False)
+            Weights will be present by default, unless weights are not
+            included in the dataset.
```
rlratzel, line 136: Minor change to make it slightly more clear.
```suggestion
        ignore_weights : Boolean (default=False)
            Ignores weights in the dataset if True, resulting in an
            unweighted Graph. If False (the default), weights from the
            dataset -if present- will be applied to the Graph. If the
            dataset does not contain weights, the Graph returned will
            be unweighted regardless of ignore_weights.
```

rlratzel (commented):
Just a couple very minor changes, otherwise looks almost done.

rlratzel (approved):

rlratzel:
@gpucibot merge

PR #2454, opened 2022-07-27
Only build cugraphmgtestutil when requested

robertmaynard:
Guard the compilation of cugraphmgtestutil to only happen when BUILD_CUGRAPH_MG_TESTS is enabled so you can build a subset of tests without MPI

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -381,7 +381,7 @@ ConfigureTest(TRIANGLE_COUNT_TEST community/triangle_count_test.cpp)
 if(BUILD_CUGRAPH_MG_TESTS)
     ###############################################################################################
     # - find MPI - only enabled if MG tests are to be built
-    find_package(MPI REQUIRED)
+    find_package(MPI REQUIRED COMPONENTS CXX)
 
     add_library(cugraphmgtestutil STATIC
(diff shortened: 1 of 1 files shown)


ChuckHastings (approved):

cpp/tests/CMakeLists.txt
```diff
@@ -400,167 +379,182 @@ ConfigureTest(TRIANGLE_COUNT_TEST community/triangle_count_test.cpp)
...
+            MPI::MPI_CXX
+    )
+
     # Set the GPU count to 1.  If the caller wants to execute MG tests using
     # more than 1, override from the command line using -DGPU_COUNT=<gpucount>
     if (NOT DEFINED GPU_COUNT)
       set(GPU_COUNT "1")
     endif()
 
-    if(MPI_CXX_FOUND)
```
rlratzel, line 414: I'm curious why this if/else block with the informative error message is being removed.

robertmaynard: Since we requie MPI to be found, that will error out before this if branch is reached.
To be extra sure, I have listed `CXX` as a required COMPONENT to `find_package(MPI)`

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2460, opened 2022-07-28
Change default return type `PropertyGraph.extract_subgraph() -> cugraph.Graph(directed=True)`

### LINKED ISSUE #2459: [FEA] Change the default for Graph instances returned by `PropertyGraph.extract_subgraph()` to `directed`
Most known use cases for `PropertyGraph.extract_subgraph()` require a directed graph (eg. GNN sampling pipelines). The current default is undirected, and users are required to pass a graph instance to `create_using` to get a directed graph.

Perhaps since the Property Graph itself is directed by nature (ie. the edge data is not being symmetrized), and the majority of users (as far as we know) need directed, the default graph instance returned by `extract_subgraph` if `create_using` isn't specified should be directed.

NOTE: I've added the `breaking` label to this issue since changing a default is technically a breaking change.

See also #2457

eriknw:
Fixes #2459

The tests already specify the use of directed graphs most places, so this required minimal changes.

### DIFF
--- a/python/cugraph/cugraph/dask/structure/mg_property_graph.py
+++ b/python/cugraph/cugraph/dask/structure/mg_property_graph.py
@@ -480,7 +480,7 @@ class EXPERIMENTAL__MGPropertyGraph:
             edge_selection_series=selected_col)
 
     def extract_subgraph(self,
-                         create_using=cugraph.Graph,
+                         create_using=None,
                          selection=None,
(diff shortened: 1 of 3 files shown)


rlratzel (approved):
Thank you for the quick response!

rlratzel:
@gpucibot merge

PR #2463, opened 2022-07-28
Cmake bug fix (MG related part wasn't property guarded)

seunghwak:
MG related parts were not properly guarded with `if(BUILD_CUGRAPH_MG_TESTS)`, this leads to a CMake error when just building for SG.


seunghwak:
Addressed in https://github.com/rapidsai/cugraph/pull/2454. Closing.

PR #2467, opened 2022-07-28
Update PageRank to leverage pylibcugraph

### LINKED ISSUE #2430: [FEA] pylibcugraph error handling should include detailed error message from the C API
pylibcugraph APIs use a utility to check for success after calling C APIs, but do not extract the detailed error message available.  Having additional details would be very helpful in debugging problems, such as the one reported [here](https://github.com/rapidsai/cugraph/issues/2425#issuecomment-1189428034).

There's already a `FIXME` to address this [here](https://github.com/rapidsai/cugraph/blob/23f9444463d63ec19fae9ea70081dca1eac0691d/python/pylibcugraph/pylibcugraph/utils.pyx#L49).

cc @ChuckHastings 

---

  - [] Code passes CI

### LINKED ISSUE #2455: [BUG] MG PageRank requires the parameter store_transposed to be set to True
**Describe the bug**
A recent change to the python `cugraph-API` was introduced when integrating the `PLC Graph`. The legacy(`cython.cu`)`Renumbering` (which sets the parameter `store_transposed`) now occurs at the lazy graph creation and this parameter can no longer be set at the algo's call. PageRank is currently the only algo affected by the change. A solution would be to leverage `pylibcugraph` `PageRank` which is already implemented or raise an exception if the user doesn't set the parameter at the lazy graph creation.

Run the MG PageRank tests

jnke2016:
This PR 

1. Extends the capabilities of the `PageRank CAPI` to the `python cuGraph API`
2. Update or Add support for the parameters `personalization` `precomputed_vertex_out_weight` and `nstart` for both the SG and MG implementation

closes #2455
closes #2430


ChuckHastings (approved):

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -62,6 +62,25 @@ template <typename vertex_t>
...
+ *
+ * @param[in] handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator,
+ * @param[in] d_vertices Vertex IDs to shuffle
+ * @param[in] d_values Values to shuffle
+ *
+ * @return tuple containing device vector of shuffled vertices and device vector of shuffled values
+ */
+template <typename vertex_t, typename value_t>
+std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<value_t>>
+shuffle_ext_vertices_and_values_by_gpu_id(raft::handle_t const& handle,
```
seunghwak, line 80: So, we have collect_local_vertex_values_from_ext_vertex_value_pairs (https://github.com/rapidsai/cugraph/pull/2467/files#diff-08a102118cad7e0349c4888019ea4b596112841ec316c0ed93f0542ac55f01c3R156).

Should we better rename this to `shuffle_ext_vertex_value_pairs_by_gpu_id` to be more consistent in naming?

ChuckHastings: The behavior is subtly different, the shuffle functions take a set of inputs and return that set of inputs shuffled to the proper GPU.  The collect function does a shuffle operation, but then executes other logic (in this case a `thrust::scatter`) to generate the desired output.  So I'm not sure that using the same naming pattern is semantically meaningful.  Perhaps `shuffle_scatter_ext_vertex_value_pairs_by_gpu_id` or something like that, although I'm not sure that's particularly meaningful either.

Feels like a thing to defer to later, but perhaps to keep in mind as we review those functions.

seunghwak: I will take a closer look, but this function internally calls `cugraph::groupby_gpu_id_and_shuffle_kv_pairs`. So, I just assumed that this shuffle key, value pairs and where `key == external vertex ID`. `external_vertex_value_pairs` may sound more consistent with `kv_pairs`.

Anyways, not something super critical, so I'm approving if this needs to be merged faster for this release.

seunghwak (approved):

python/pylibcugraph/pylibcugraph/pagerank.pyx
```diff
@@ -46,16 +49,20 @@ from pylibcugraph.utils cimport (
...
     assert_CAI_type,
     copy_to_cupy_array,
+    get_c_type_from_numpy_type,
+    create_cugraph_type_erased_device_array_view_from_py_obj,
 )
 
 
 def pagerank(ResourceHandle resource_handle,
             _GPUGraph graph,
+            precomputed_vertex_out_weight_vertices,
```
rlratzel, line 59: It looks like this and other params are missing from the docstring.

python/cugraph/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -105,8 +157,20 @@ def pagerank(input_graph,
...
+        GPU Dataframe containing the initial guess for pagerank.
+        nstart['vertex'] : cudf.Series
+            Subset of vertices of graph for initial guess for pagerank values
+        nstart['values'] : cudf.Series
+            Pagerank values for vertices
+
+    weight: str, optional (default=None)
+        The attribute column to be used as edge weights if Graph is a NetworkX
+        Graph. This parameter is here for NetworkX compatibility and is ignored
+        in case of a cugraph.Graph
```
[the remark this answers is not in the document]
jnke2016: Right. I should probably remove that parameter because it was never supported in previous releases. I copied the `docstrings` from the SG implementation and I forgot to clean this

python/cugraph/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -105,8 +157,20 @@ def pagerank(input_graph,
...
+
+    dangling : dict, optional (default=None)
+        This parameter is here for NetworkX compatibility and ignored
```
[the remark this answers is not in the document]
jnke2016: This parameter should be removed too.

python/cugraph/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -84,19 +128,27 @@ def pagerank(input_graph,
...
             Subset of vertices of graph for personalization
         personalization['values'] : cudf.Series
             Personalization values for vertices
 
+    precomputed_vertex_out_weight : cudf.Dataframe, optional (default=None)
+        GPU Dataframe containing the precomputed vertex out weight
+        information.
+        precomputed_vertex_out_weight['vertex'] : cudf.Series
+            Subset of vertices of graph for precomputed_vertex_out_weight
+        personalization['sums'] : cudf.Series
```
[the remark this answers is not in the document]
jnke2016: Good catch

python/pylibcugraph/pylibcugraph/personalized_pagerank.pyx
```diff
@@ -0,0 +1,232 @@
...
+    increases when the tolerance descreases and/or alpha increases toward the
+    limiting value of 1.
+
+    Parameters
+    ----------
+    resource_handle : ResourceHandle
+        Handle to the underlying device resources needed for referencing data
+        and running algorithms.
+
+    graph : SGGraph
```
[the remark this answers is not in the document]
jnke2016: It should be `Graph` and we should update this for all pylibcugraph algo's docstring.

python/cugraph/cugraph/link_analysis/pagerank.py
```diff
@@ -89,24 +103,41 @@ def pagerank(
     PageRank : cudf.DataFrame
         GPU data frame containing two cudf.Series of size V: the vertex
         identifiers and the corresponding PageRank values.
-
         df['vertex'] : cudf.Series
             Contains the vertex identifiers
         df['pagerank'] : cudf.Series
             Contains the PageRank score
```
rlratzel, line 109: We should probably also include the following as part of the description of the return value:

`NOTE: if the input cugraph.Graph was created using the renumber=False option of any of the from_*_edgelist() methods, pagerank assumes that the vertices in the edgelist are contiguous and start from 0. If the actual set of vertices in the edgelist is not contiguous (has gaps) or does not start from zero, pagerank will assume the "missing" vertices are isolated vertices in the graph, and will compute and return pagerank values for each.`

python/cugraph/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -105,8 +157,20 @@ def pagerank(input_graph,
...
+    weight: str, optional (default=None)
+        The attribute column to be used as edge weights if Graph is a NetworkX
+        Graph. This parameter is here for NetworkX compatibility and is ignored
+        in case of a cugraph.Graph
+
+    dangling : dict, optional (default=None)
+        This parameter is here for NetworkX compatibility and ignored
 
     Returns
     -------
```
rlratzel, line 176: We should probably also include the following as part of the description of the return value:

`NOTE: if the input cugraph.Graph was created using the renumber=False option of any of the from_*_edgelist() methods, pagerank assumes that the vertices in the edgelist are contiguous and start from 0. If the actual set of vertices in the edgelist is not contiguous (has gaps) or does not start from zero, pagerank will assume the "missing" vertices are isolated vertices in the graph, and will compute and return pagerank values for each.`

python/cugraph/cugraph/link_analysis/pagerank.py
```diff
@@ -89,24 +103,41 @@ def pagerank(
...
-
     """
 
-    G, isNx = ensure_cugraph_obj_for_nx(G, weight)
+    initial_guess_vertices = None
+    initial_guess_values = None
+    pre_vtx_o_wgt_vertices = None
+    pre_vtx_o_wgt_sums = None
+
+    G, isNx = ensure_cugraph_obj_for_nx(G)
```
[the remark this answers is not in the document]
jnke2016: It already has a default value [weight](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/utilities/utils.py#L326) passed in and if `G` is of type `cugraph.Graph`, then the return is (`G`, `False`), otherwise, it will extract the [edges](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/utilities/nx_factory.py#L44)(`src, dst, weight`)  from the `nx.Graph` and return a [cudf.DataFrame](https://github.com/rapidsai/cugraph/blob/branch-22.08/python/cugraph/cugraph/utilities/nx_factory.py#L55) which will be used to create a `cugraph.Graph`

python/cugraph/cugraph/dask/link_analysis/pagerank.py
```diff
@@ -13,65 +13,131 @@
...
+    """
+    cupy_vertices, cupy_pagerank = cp_arrays
+    df = cudf.DataFrame()
+    df["vertex"] = cupy_vertices
+    df["pagerank"] = cupy_pagerank
+
+    return df
+
+
+def ensure_valid_dtype(input_graph, input_df, input_df_name):
```
rlratzel, line 42: Can you add a FIXME saying this should be moved to a utility module to be shared by both other algos (especially SG and MG)?

python/cugraph/cugraph/link_analysis/pagerank.py
```diff
@@ -11,17 +11,55 @@
...
+    else:
+        cols = 'vertex'
+    input_df = input_graph.add_internal_vertex_id(
+        input_df, "vertex", cols
+    )
+
+    return input_df
+
+
+def ensure_valid_dtype(input_graph, input_df, input_df_name):
```
rlratzel, line 39: Same FIXME as above.

rlratzel (commented):
Looks good, just a quick request for a FIXME.
Also, can you update the description to say it closes #2430 too?

python/cugraph/cugraph/link_analysis/pagerank.py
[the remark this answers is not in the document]
rlratzel: Resolved by reverting this line, so it's now passing `weight` as before.

python/pylibcugraph/pylibcugraph/personalized_pagerank.pyx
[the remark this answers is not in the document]
rlratzel: Resolved by changing the text to read SGGraph or MGGraph

rlratzel (approved):

rlratzel:
@gpucibot merge

rlratzel:
@gpucibot merge

PR #2475, opened 2022-08-01
Fix typos in Python CMakeLists CUDA arch file

vyasr:
These typos would prevent properly handling required CMake features as determined by `rapids_cuda_init_architectures`. This only affects compilations where `FIND_CUGRAPH_CPP=OFF`.


BradReesWork (approved):

rlratzel:
@gpucibot merge

PR #2477, opened 2022-08-02
fix non-deterministic bug in uniform neighborhood sampling

### LINKED ISSUE #2446: [BUG] Illegal Memory Access when running neighborhood samppling
**Describe the bug**
A bug was unveiled when running `uniform neighbor sampling` on a single GPU. Upon initial investigation, it appears to be caused by the `cython.cu` renumbering and skipping it (`legacy_renum_only=True`) seems to resolve the issue(yet to be verified). However,, skipping the C++ renumbering is not a viable solution.

```
import cudf
import cugraph
import numpy as np

df = cudf.DataFrame({'src':[2,3,1],
                       'dst':[4,5,2]}).astype(np.int32)
df['weight']=1.0
G.from_cudf_edgelist(df, source='src',destination='dst',edge_attr='weight')

for _ in range(0,2):
    seed_list = cudf.Series([2]).astype(np.int32)
    sampled_g_cugraph = cugraph.uniform_neighbor_sample(G,
                                         start_list=seed_list,
                                         fanout_vals=[-1],
                                         with_replacement=False)
```

ChuckHastings:
Found a bug in uniform neighborhood sampling.  Called a function that returns an optional.  In SG mode the optional is always null.  It was being dereferenced and passed to a function where random memory contents were being used.

Fixed the SG code to not use those values.

Closes #2446


ChuckHastings:
rerun tests

alexbarghi-nv (approved):
@ChuckHastings this has fixed the error on my side.

seunghwak (approved):

ChuckHastings:
@gpucibot merge

Issue #2479, opened 2022-08-02
[FEA] Vertex and edge property support.


seunghwak:
As pre-requsites for https://github.com/rapidsai/cugraph/issues/2143 and https://github.com/rapidsai/cugraph/issues/2433, we need a wrapper class to store vertex property values (which can be a simple 1D array without vertex masking, this does not hold anymore with vertex masking, need a wrapper class to hide complexity) and another class to store edge property values (in addition to the edge weight).

- [ ] Update the current edge_partition_src|dst_property_t to support vertex/edge source|destination/edge property values in a consistent and efficient way.
- [ ] Add a wrapper class for vertex property values
- [ ] Add a new class for edge property values.

ChuckHastings:
We have added support for edge properties.

We still manage vertex properties using `rmm::device_uvector`, which exposes internal implementation and prevents support for vertex masking and some other optimizations we might consider.

PR #2501, opened 2022-08-03
Rename multiple .cuh (.cu) files to .hpp (.cpp)

seunghwak:
Pre-requisites for https://github.com/rapidsai/cugraph/issues/2479

Several .cuh header files (.cuh indicates nvcc is required to compile) can be renamed to .hpp (.hpp indicates the file can be compiled with a host compiler such as g++) with no or minimum modifications.

This PR performs multiple renaming (eventually to separate edge_partition_src_dst_property.cuh to .hpp and .cuh files).

Breaking as public header files are renamed.


cpp/src/prims/property_op_utils.cuh
```diff
@@ -136,6 +137,36 @@ struct intersection_op_result_type<
...
+                           thrust::get<I>(value));
+    atomic_accumulate_thrust_tuple_impl<Iterator, TupleType, I + 1, N>().compute(iter, value);
+  }
+};
+
+template <typename Iterator, typename TupleType, size_t I>
+struct atomic_accumulate_thrust_tuple_impl<Iterator, TupleType, I, I> {
+  __device__ constexpr void compute(Iterator iter, TupleType const& value) const {}
+};
+
```
naimnv, line 169: Not related to to this PR: If the tuple has too many element (not sure if it can be the case), would it be a bit inefficient in terms of number of function instantiates it would generate?

naimnv (approved):

cpp/src/prims/property_op_utils.cuh
seunghwak: As we discussed offline, this will end up with N atomicAdd instructions. So, no in terms of execution time, but yes, this can increase compile time.

jnke2016 (approved):

ChuckHastings (approved):

ChuckHastings:
@gpucibot merge

PR #2516, opened 2022-08-05
Branch 22.10 fix mg bfs type mismatch

### LINKED ISSUE #2515: [BUG] Dask-cudf dataframes no longer merges dataframes of different type
**Describe the bug**
A recent change in `cudf` no longer allows inner merges of dataframes with different type and this is breaking `MG BFS`. In fact, when checking if the user is passing valid seeds to `MG BFS`, an inner merge between all the vertices in the graph and the `BFS seeds` is computed but this fails if the dataframes are of different type. In such case, `cudf` can return unexpected results as described in the error message below
```
UserWarning: Merging dataframes with merge column data type mismatches:
+---------------+------------+-------------+
| Merge columns |left dtype  | right dtype |
+---------------+------------+-------------+
|    (0, 0)     |   int32    |  int64      |
+---------------+------------+-------------+
Cast dtypes explicitly to avoid unexpected results.
```
Run `BFS` with seeds of different type than the graph vertices's type

jnke2016:
This PR coerces the seeds and vertices type, raise a warning and updates the tests accordingly. 
closes #2515.

### DIFF
--- a/python/cugraph/cugraph/dask/traversal/bfs.py
+++ b/python/cugraph/cugraph/dask/traversal/bfs.py
@@ -142,7 +142,7 @@ def bfs(input_graph,
 
         if invalid_dtype:
             raise ValueError("The 'start' values dtype must match "
-                                "the graph's vertices dtype.")
+                             "the graph's vertices dtype.")
 
         is_valid_vertex = G.has_node(start)
         if not is_valid_vertex:
--- a/python/cugraph/cugraph/dask/traversal/bfs.py
+++ b/python/cugraph/cugraph/dask/traversal/bfs.py
@@ -122,12 +122,28 @@ def bfs(input_graph,
 
     if not isinstance(start, (dask_cudf.DataFrame, dask_cudf.Series)):
         if not isinstance(start, (cudf.DataFrame, cudf.Series)):
-            start = cudf.Series(start)
+            vertex_dtype = input_graph.nodes().dtype
+            start = cudf.Series(start, dtype=vertex_dtype)
         if isinstance(start, (cudf.DataFrame, cudf.Series)):
             # convert into a dask_cudf
             start = dask_cudf.from_cudf(start, input_graph._npartitions)
 
     def check_valid_vertex(G, start):
+        invalid_dtype = False
+        if isinstance(start, dask_cudf.Series):
+            if start.dtype is not input_graph.nodes().dtype:
+                invalid_dtype = True
+        else:
+            # Multicolumn vertices case
+            start_dtype = start.dtypes.reset_index(drop=True)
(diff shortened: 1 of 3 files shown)


python/cugraph/cugraph/dask/traversal/bfs.py
```diff
@@ -122,12 +122,28 @@ def bfs(input_graph,
 
     if not isinstance(start, (dask_cudf.DataFrame, dask_cudf.Series)):
         if not isinstance(start, (cudf.DataFrame, cudf.Series)):
-            start = cudf.Series(start)
+            vertex_dtype = input_graph.nodes().dtype
+            start = cudf.Series(start, dtype=vertex_dtype)
         if isinstance(start, (cudf.DataFrame, cudf.Series)):
             # convert into a dask_cudf
             start = dask_cudf.from_cudf(start, input_graph._npartitions)
```
rlratzel, line 129: Seems like we didn't need the `if isinstance` block, or am I missing something?
```suggestion
        # convert into a dask_cudf
        start = dask_cudf.from_cudf(start, input_graph._npartitions)
```

python/cugraph/cugraph/dask/traversal/bfs.py
```diff
@@ -122,12 +122,28 @@ def bfs(input_graph,
...
 
     def check_valid_vertex(G, start):
```
rlratzel, line 131: It's not obvious why we need to define a function here just to call it once as part of the `if check_start` block below.  Should this code just be moved to the `if check_start` block?

python/cugraph/cugraph/tests/mg/test_mg_bfs.py
```diff
@@ -126,6 +126,11 @@ def test_dask_bfs_invalid_start(dask_client, directed):
     with pytest.raises(ValueError):
         dcg.bfs(G, source_vertex).compute()
 
+    # invalid dtype
```
rlratzel, line 129:
```suggestion
    # invalid dtype (the default cudf.Series() dtype is int64)
```

python/cugraph/cugraph/dask/traversal/bfs.py
jnke2016: Right. I am missing and `else` statement here which should raise a type error.

```
else:
     raise TypeError("'start' must be a list or a cudf object or a dask 
                     f"object, got :  {type(start)}")
```

jnke2016: Right. This extra check is unnecessary

rlratzel:
rerun tests

Rerunning due to a conda server error.

rlratzel (approved):

PR #2518, opened 2022-08-08
Use rapids-cmake 22.10 best practice for RAPIDS.cmake location

robertmaynard:
Removes possibility of another projects RAPIDS.cmake being used, and removes need to always download a version.


rlratzel (approved):

rlratzel:
@gpucibot merge

PR #2584, opened 2022-08-12
Update vertex_frontier_t to take unsorted (tagged-)vertex list with possible duplicates

seunghwak:
Partially address https://github.com/rapidsai/cugraph/issues/2580

Update vertex_frontier_t to take unsorted input (tagged-)vertex list with possible duplicates.

This will be used to define a selection primitive. The primitive needs to take a (tagged-)vertex list to support Node2Vec style algorithms (tagged with the predecessor vertex ID or some derived property from the current vertex predecessor vertex pairs).

Breaking as primitives API has been changed (`vertex_frontier_t` and `transform_reduce_v_frontier_outgoing_e_by_dst`).

Issue #2580, opened 2022-08-12
Create Selection primitive


ChuckHastings:
Create new Selection primitive for use by Neighbor sampling and Random walks.

PR #2586, opened 2022-08-15
Define a selection primtive API

seunghwak:
Partially address #2580.

This PR is dependent on #2584.

This PR defines API for two selection primitives, one for the biased sampling/random walk and another for uniform random sampling/random walk.

@ChuckHastings For Node2Vec style random walk,

We can compute intersections for each (previous vertex, current vertex pairs).

We need to first create a non-detail space primitive calling detail::nbr_intersection (https://github.com/rapidsai/cugraph/blob/branch-22.10/cpp/src/prims/detail/nbr_intersection.cuh#L492) for given vertex pairs (this can be used for Jaccard and Overlap coefficients as well).

In MG, each GPU should store neighbor intersection outputs for the relevant source/destination ranges (not sure whether should we create additional utility functions to handle this, or this may not be a recurring pattern, so just leave this task much more as a dark magic for advanced users who understand how the 2D partitioning actually works). May go for the latter till we see this pattern occurring in other places.

Once we have neighbor intersection outputs and previous vertex IDs for the relevant (previous vertex ID, current vertex ID) pairs, we can create `frontier` having tagged-current vertex ID values (tag is an index to access previous vertex ID and neighbor intersection outputs for the perv vertex, current vertex pair).

Then, `e_bias_op` can check whether the outgoing neighbor belongs to the intersection or coincides with the previous vertex to properly set bias values.


ChuckHastings (approved):

seunghwak:
rerun tests

jnke2016 (approved):

cpp/src/prims/per_v_random_select_and_transform_outgoine_e.cuh
```diff
@@ -0,0 +1,170 @@
...
+template <typename GraphViewType,
+          typename VertexFrontierBucketType,
+          typename EdgeSrcValueInputWrapper,
+          typename EdgeDstValueInputWrapper,
+          typename EdgeBiasOp,
+          typename EdgeOp,
+          typename T>
+std::tuple<std::optional<rmm::device_uvector<size_t>>,
+           decltype(allocate_dataframe_buffer<T>(size_t{0}, rmm::cuda_stream_view{}))>
+per_v_random_select_and_transform_outgoing_e(raft::handle_t const& handle,
```
seunghwak, line 78: @ChuckHastings What do you think about `select_and_transfrom` vs `select_transform` in the primitive name?

I initially picked former but now getting more inclined to the latter as it is less verbose and more in line with other std:: thrust:: functions like `partitoin_copy`, `transform_reduce`, and so on. I guess removing `and` here won't cause much confusion?

I have a similar issue in naming `extract_and_transform_if...` primitives (a modification of the existing `extract_if_e` primitive to allow not just extracting an edge (source, destination, weight) triplets but to allow including edge IDs and types and so on... I feel like `extract_transform` may work better than `extract_and_transform`...

ChuckHastings: I think removing the `_and` from the name is reasonable.  As you observe, the thrust model doesn't include and conjunctions, they are implied.

ChuckHastings:
@gpucibot merge

PR #2588, opened 2022-08-16
Fix MG Louvain C API test 

ChuckHastings:
Add extra logic to support arbitrary numbering of clusters.

Our Louvain implementation is deterministic based on the number of GPUs (that is, with k GPUs and a given graph it will always generate the same answer).  However, in the case of this test a different number of GPUs can generate a different result based on how the vertices are numbered.

We could unrenumber the result, but that's an extra step that's not necessary in the ordinary flow.  Added a more sophisticated test that will recreate the necessary subset of the number map and validate according to that.


ChuckHastings:
@gpucibot merge

PR #2607, opened 2022-08-23
Fix MG PLC algos intermittent hang

### LINKED ISSUE #2568: [BUG] Failure occurs when running mg PLC algos numerous times
**Describe the bug**
An intermittent failure occurs when running mg PLC algos like `core-number` and `PageRank` several times with the same `PLC Graph`.

run mg `core-number` or `PageRank` more than once with the same `PLC Graph`

jnke2016:
Dask doesn't always release some of the inactive futures fast enough. This can be problematic when running the same `algo` several times with the same `PLC graph` because those futures can be cache in the next iteration causing a hang if some get released midway.

This PR manually delete inactive futures.
closes #2568

### DIFF
--- a/python/cugraph/cugraph/dask/cores/core_number.py
+++ b/python/cugraph/cugraph/dask/cores/core_number.py
@@ -133,6 +133,7 @@ def core_number(input_graph,
     # the past iteration's futures and this can cause a hang if some
     # of those futures get released midway
     del result
+    del cudf_result
 
     ddf = dask_cudf.from_delayed(cudf_result).persist()
     wait(ddf)
--- a/python/cugraph/cugraph/dask/cores/core_number.py
+++ b/python/cugraph/cugraph/dask/cores/core_number.py
@@ -114,6 +114,7 @@ def core_number(input_graph,
             degree_type,
             do_expensive_check,
             workers=[w],
+            allow_other_workers=False,
         )
         for w in Comms.get_workers()
     ]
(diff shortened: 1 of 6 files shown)


rlratzel (approved):

python/cugraph/cugraph/dask/cores/core_number.py
```diff
@@ -126,7 +127,16 @@ def core_number(input_graph,
 
     wait(cudf_result)
 
-    ddf = dask_cudf.from_delayed(cudf_result)
+    # FIXME: Dask doesn't always release it fast enough.
+    # For instance if the algo is run several times with
+    # the same PLC graph, the current iteration might try to cache
+    # the past iteration's futures and this can cause a hang if some
+    # of those futures get released midway
+    del result
```
VibhuJawa, line 135: I think removing `result` is still a good practice as when `cudf_result`  is completed  we will still have `result` in memory so from a memory relief standpoint it makes sense.

python/cugraph/cugraph/dask/cores/core_number.py
```diff
@@ -126,7 +127,16 @@ def core_number(input_graph,
...
+
+    ddf = dask_cudf.from_delayed(cudf_result).persist()
+    wait(ddf)
+
```
VibhuJawa, line 139: I suggest we also remove `cudf_result` here to save on memory . We have a renumbered call after this so `result` does not go out of scope immediately.

VibhuJawa (changes requested):
I am requesting to delete `cudf_result`, looks good otherwise.

jnke2016:
rerun tests

jnke2016:
rerun tests

jnke2016:
rerun tests

jnke2016:
rerun tests

ajschmidt8 (approved):

PR #2523, opened 2022-08-09
PropertyGraph set index to vertex and edge ids

eriknw:
~Currently, this only does SG version for #2401.  MG is still TODO.~

Closes #2401

This also doesn't change anything user-facing (yet).

PR #2614, opened 2022-08-24
Allow setting the edge weights to edge ids as currently required by uniform_neighbor_sample

alexbarghi-nv:
Blocked by https://github.com/rapidsai/cugraph/pull/2523
Blocked by https://github.com/rapidsai/cugraph/pull/2510


alexbarghi-nv:
Dropping this in favor of the solution currently being developed using edge ids and edge types.

Issue #2600, opened 2022-08-19
Create new primitive to replace `gather_one_hop`


ChuckHastings:
In addition to random sampling, the neighborhood sampling code supports a feature (when fanout for a level is set to -1) to retrieve all neighbors of vertices encountered at that point in the sampling.

The current implementation uses a method named `gather_one_hop`, which has some issues related to how it handles the hyper-sparse areas in an MG graph.

Create a new primitive that will provide that functionality (given a set of vertices, return the set of edges that depart from that vertex).

BradReesWork:
closed by #2623

PR #2617, opened 2022-08-24
Code cleanup

seunghwak:
As a pre-requisite to address https://github.com/rapidsai/cugraph/issues/2600

- [x] add utility functions to compute local degrees (for the entire set of majors & a subset of the majors) and the sum of local degrees (for a subset of the majors) to `edge_partition_device_view_t`
- [x] Replace raw pointers with `raft::device_span` in multiple places of `graph_view_t`, `edge_partition_view_t`, and `edge_partition_device_view_t`
- [x] `cugraph::test::to_host` was previously defined, but used only in a subset of C++ tests, update `to_host` to take both `raft::device_span` and `rmm::device_uvector` as input arguments and use this (instead of update_host) in most of the C++ tests (except for some legacy sampling tests). 

Breaking as some functions in `graph_view_t`, `edge_partition_view_t` and `edge_partition_device_view_t` are updated to return `raft::device_span` instead of a raw pointer (these functions are used only internally AFAIK, but still a breaking change in the public API).


ChuckHastings:
rerun tests

naimnv (approved):
Thanks Seunghwa for pointing me to major changes.

ChuckHastings:
@gpucibot merge

PR #2623, opened 2022-08-26
Primitive to support gathering one hop neighbors

seunghwak:
- [x] Refactored `transform_reduce_v_frontier_outgoing_e_by_dst` (moved the first part of the implementation to `detail/extract_transform_v_frontier_e`, this moved code is used in the new primitive as well).
- [x] Added a new primitive `extract_transform_v_frontier_outgoing_e`

@ChuckHastings To gather one hop neighbors, in SG, you need to call this primitive and an edge_op passed to the primitive should return src,dst, (edge weight), and (edge ID or type once we added edge property support). In MG, the result should be properly shuffled in addition. Then, call `thrust::sort_by_key` and `thrust::reduce_by_key` to gather neighbors of each vertex together and compute degrees for each vertex. In SG, we might be able to avoid `thrust::sort_by_key` if we change the primitive internals and provide some additional guarantees about the extracted output ordering, but I guess this `thrust::sort_by_key` overhead is not too significant. If `thrust::sort_by_key` in gathering one hop neighbors really becomes a performance bottleneck, we can re-visit this assumption.


cpp/src/prims/detail/extract_transform_v_frontier_e.cuh
```diff
@@ -0,0 +1,1039 @@
+/*
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Is this new (should be just 2022) or moved from somewhere?

cpp/src/prims/extract_transform_v_frontier_outgoing_e.cuh
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright (c) 2020-2022, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Is this new (should be just 2022) or moved from somewhere?

ChuckHastings (approved):

cpp/src/prims/detail/extract_transform_v_frontier_e.cuh
seunghwak: Copied from transfrom_reduce_v_frontier_outgoing_e_by_dst.cuh

cpp/src/prims/extract_transform_v_frontier_outgoing_e.cuh
seunghwak: Yes, this is actually new. I will fix the copyright year.

naimnv (approved):
I looked carefully into some part of the code and skimmed over some other parts.

ChuckHastings:
@gpucibot merge

PR #2624, opened 2022-08-26
Define C and C++ APIs for Jaccard/Sorensen/Overlap

### LINKED ISSUE #2541: MNMG Jaccard - Define C API
Define a C API for Jaccard, Overlap and Sorensen algorithms.

### LINKED ISSUE #2542: MNMG Jaccard - Define C++ API
Define C++ API for Jaccard, Overlap and Sorensen

### LINKED ISSUE #2544: MNMG Jaccard - Define C unit tests
Define C API unit tests. These do not need to duplicate the C++ tests but should focus on testing all different mechanisms on the C API.

Tests should include both SG and MG tests.

### LINKED ISSUE #2545: MNMG Jaccard - Define C++ unit tests
Define C++ unit tests.  Tests should include both SG and MG tests.

ChuckHastings:
This PR defines C and C++ APIs for the similarity algorithms Jaccard, Sorensen and Overlap, as well as the C++ unit tests.

Legacy implementation of Jaccard and Overlap are moved to a legacy directory to make the transition clear.

Closes #2541 
Closes #2542 
Closes #2544
Closes #2545


cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
...
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @param first optional device span defining the first vertex in the desired vertex pairs
+ * @param second optional device span defining the second vertex in the desired vertex pairs
```
seunghwak, line 1699: Any requirements about pair shuffling in MG?

I guess we need a bit more descriptive variable name than `first` and `second`... Maybe something like `vertex_pair_firsts` and `vertex_pair_seconds`?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
...
```
seunghwak, line 1730: Any requirements about pair shuffling in MG?

I guess we need a bit more descriptive variable name than `first` and `second`... Maybe something like `vertex_pair_firsts` and `vertex_pair_seconds`?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
...
```
seunghwak, line 1761: Any requirements about pair shuffling in MG?

I guess we need a bit more descriptive variable name than `first` and `second`... Maybe something like `vertex_pair_firsts` and `vertex_pair_seconds`?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
...
+ * @param use_weights If true use the weights associated with the graph.  If false assume a weight
+ *                    of 1 for all edges.
+ * @return tuple defining vertex pairs and their Jaccard similarity coefficient.
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+std::
+  tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>, rmm::device_uvector<weight_t>>
+  overlap(raft::handle_t const& handle,
```
seunghwak, line 1769: I feel like `jaccard|sorensen|overlap_(similarity_)coefficients` are better names... `jaccard` and `sorensen` might be fine, but just seeing `overlap`, this can be interpreted in many different ways (something like finding an overlapping part of two graphs or whatever) before seeing the documentation and input parameter lists.

cpp/src/link_prediction/jaccard_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+#include <cugraph/algorithms.hpp>
+#include <link_prediction/similarity_impl.cuh>
+
+#include <raft/handle.hpp>
+
+namespace cugraph {
+namespace detail {
+
+struct jaccard_functor_t {
+  template <typename weight_t>
```
seunghwak, line 27: Why's this placed after struct?

cpp/src/link_prediction/jaccard_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+                                    weight_t,
+                                    weight_t)
+  {
+    return static_cast<weight_t>(cardinality_a_intersect_b) /
+           static_cast<weight_t>(cardinality_a + cardinality_b - cardinality_a_intersect_b);
+  }
+};
+
+struct weighted_jaccard_functor_t {
+  template <typename weight_t>
```
seunghwak, line 41: Why's this placed after struct?

cpp/src/link_prediction/overlap_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+#include <cugraph/algorithms.hpp>
+#include <link_prediction/similarity_impl.cuh>
+
+#include <raft/handle.hpp>
+
+namespace cugraph {
+namespace detail {
+
+struct overlap_functor_t {
+  template <typename weight_t>
```
seunghwak, line 27: Why's this placed after struct?

cpp/src/link_prediction/overlap_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+                                    weight_t,
+                                    weight_t)
+  {
+    return static_cast<weight_t>(cardinality_a_intersect_b) /
+           static_cast<weight_t>(std::min(cardinality_a, cardinality_b));
+  }
+};
+
+struct weighted_overlap_functor_t {
+  template <typename weight_t>
```
seunghwak, line 41: Why's this placed after struct?

cpp/src/link_prediction/sorensen_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+#include <cugraph/algorithms.hpp>
+#include <link_prediction/similarity_impl.cuh>
+
+#include <raft/handle.hpp>
+
+namespace cugraph {
+namespace detail {
+
+struct sorensen_functor_t {
+  template <typename weight_t>
```
seunghwak, line 27: Why's this placed after struct?

cpp/src/link_prediction/sorensen_impl.cuh
```diff
@@ -0,0 +1,72 @@
...
+                                    weight_t,
+                                    weight_t)
+  {
+    return static_cast<weight_t>(2 * cardinality_a_intersect_b) /
+           static_cast<weight_t>(cardinality_a + cardinality_b);
+  }
+};
+
+struct weighted_sorensen_functor_t {
+  template <typename weight_t>
```
seunghwak, line 41: Why's this placed after struct?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
...
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @param first optional device span defining the first vertex in the desired vertex pairs
+ * @param second optional device span defining the second vertex in the desired vertex pairs
+ * @param use_weights If true use the weights associated with the graph.  If false assume a weight
+ *                    of 1 for all edges.
+ * @return tuple defining vertex pairs and their Jaccard similarity coefficient.
```
seunghwak, line 1702: Should we return the vertex pairs as well? Or should this be optional and vertex pairs should be returned only when `first` and `second` are not provided?

And if we provide a function to easily compute two-hop neighbors, we may make `first` and `second` required parameters. If uses want to compute coefficients for two-hop neighbors, they call the function to compute two-hop neighbors first and call this function.

cpp/tests/link_prediction/similarity_compare.cpp
```diff
@@ -0,0 +1,157 @@
...
+               thrust::make_zip_iterator(src.begin(), dst.begin()),
+               thrust::make_zip_iterator(src.end(), dst.end()));
+  thrust::sort(
+    thrust::host,
+    thrust::make_zip_iterator(result_src.begin(), result_dst.begin(), result_score.begin()),
+    thrust::make_zip_iterator(result_src.end(), result_dst.end(), result_score.end()));
+
+  size_t result_pos = 0;
+
+  // Iterate over all (u,v) pairs
```
seunghwak, line 64: Should we do this??? This can be very expensive (even though this is just for testing).

seunghwak: (`src`, `dst`, `wgt`) is a graph edge list, right? I guess `result_src` and `result_dst` here are misnomers. They are first and second elements of vertex pairs, right?

If we take the CSR graph, and we can iterate over vertex pairs (`result_src` and `result_dst`), and for the first and second elements of the pair, we can extract neighbor lists from the CSR graph. Then, we can compute the expected score from the host, and compute this number with the `result_score`. I guess this will be much faster (and also easier to understand). Am I missing something?

cpp/include/cugraph/algorithms.hpp
seunghwak: I now feel like `std::tuple<raft::device_span<vertex_t const>, raft::device_span<vertex_t const>> vertex_pairs` sounds like a better name.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,99 @@ void triangle_count(raft::handle_t const& handle,
                     raft::device_span<edge_t> counts,
                     bool do_expensive_check = false);
 
+/**
+ * @brief     Compute Jaccard similarity coefficient
+ *
+ * Similarity is computed based for every pair of vertices specified, or (if no vertices
```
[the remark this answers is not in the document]
ChuckHastings: Yes, corrected in next push

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Added a statement about being local to this GPU in the next push.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Added a statement about being local to this GPU in the next push.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Done in next push

cpp/src/link_prediction/jaccard_impl.cuh
ChuckHastings: I was templating the method rather than the structure since I'm not defining any member data.

This allows me to pass the struct through different functions without needing to compile multiple variations, the multiple variations are only compiled in the place where the function is actually called.

I can change this if we really think it would be better on the struct itself.  I don't expect that I will actually pass the struct through a bunch of function calls when I finish the implementation.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I did it this way largely for this discussion :-)

The legacy implementation had a function to compute two-hop neighbors and then made this a required field.  We could mimic that behavior here.  I do think that's a bit cleaner.  I provided this definition as an alternative... essentially we embed the computation of the two-hop neighbors in the function.

If you're on board with the change, I'm happy to
* add a function to compute two-hop neighbors
* Make the vertex_pairs parameter required
* Only return the scores, ordered to match the vertex_pairs

cpp/tests/link_prediction/similarity_compare.cpp
ChuckHastings: Several thoughts mixed in this reply.
* Yes, the tuple (`src`, `dst`, `wgt`) is the graph edge list.  I could make that a tuple called graph edge list, which would make things more semantically clear.  Perhaps that will be easier to understand.
* With the above change, `result_src` and `result_dst` will become `result_vertex_pairs`

If we make the change to compute the 2-hop neighbors, I can change this to the following:
* Add a separate test for the 2-hop neighbor function and validate it works properly
* In this test, pass in the 2-hop neighbors, and just iterate over those and directly compute the scores

Ultimately I was going to wait until I started testing to consider other optimizations to this function (other than the above).  If we're testing a full set of 2-hop neighbors, I'm not sure that extracting everything on demand will actually be faster than just operating on the fully-extracted graph.

cpp/include/cugraph/algorithms.hpp
seunghwak: So, being local is based on the edge partitioning requirement, right? (even though vertex pairs are not necessarily edges). In that case, this aligns with the expectation of detail::nbr_intersection primitive, so I assume no additional shuffling will be necessary in implementation (https://github.com/rapidsai/cugraph/blob/branch-22.10/cpp/src/prims/detail/nbr_intersection.cuh#L475).

cpp/src/link_prediction/jaccard_impl.cuh
seunghwak: I see. Yeah, not a big deal, I just thought this was a mistake, but I now learned this is intended.

cpp/include/cugraph/algorithms.hpp
seunghwak: Yes, I am on board. I think the suggested design is much more composable and also better stick to the one-function-does-one-thing idea.

cpp/tests/link_prediction/similarity_compare.cpp
ChuckHastings: Perhaps we can add a parameter that allows us to test a subset of the edges rather than the whole.  For smaller graphs we can test everything, if you want to test on a bigger graph we just validate some of the vertex pairs.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I will put it in the next push.

cpp/tests/link_prediction/similarity_compare.cpp
seunghwak: Now I see why you did this. You want to test this for every possible vertex pair. Yeah... this might be great, but can be too expensive for non-tiny graphs. I guess testing for two-hop neighbors with limited K (for high-degree vertices, pick only up to K neighbors, similar to neighbor sampling) might be sufficient.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Yes, that is my intent.

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+ * @param [in]  use_weight   If true consider the edge weight in the graph, if false use an
+ *                           edge weight of 1
+ * @param [in]  do_expensive_check A flag to run expensive checks for input arguments (if set to
+ * `true`).
+ * @param [out] result       Opaque pointer to paths results
+ * @param [out] error        Pointer to an error object storing details of any error.  Will
+ *                           be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_jaccard(const cugraph_resource_handle_t* handle,
```
seunghwak, line 93: Add `_coefficients` to be consistent with the C++ naming?

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -584,4 +584,33 @@ create_graph_from_edgelist(
...
+ * @return tuple containing pairs of vertices that are 2-hops apart.
+ */
+template <typename vertex_t,
+          typename edge_t,
+          typename weight_t,
+          bool store_transposed,
+          bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>> get_two_hop_neighbors(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu> const& graph_view);
```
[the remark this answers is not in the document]
ChuckHastings: Good point.  I added an optional start_vertices parameter in the next push.

cpp/include/cugraph_c/similarity_algorithms.h
ChuckHastings: Done in next push.

rlratzel (changes requested):

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1677,6 +1679,90 @@ void triangle_count(raft::handle_t const& handle,
...
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @param vertex_pairs tuple of device spans defining the vertex pairs to compute similarity for
+ * In a multi-gpu context each vertex pair should be local to this GPU.
+ * @param use_weights If true use the weights associated with the graph.  If false assume a weight
+ *                    of 1 for all edges.
+ * @return tuple defining vertex pairs and their Jaccard similarity coefficient.
```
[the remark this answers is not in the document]
jnke2016: @ChuckHastings To make sure I am following your discussion with Seunghwa. The vertex pairs will be mandatory in the C++ API of `jaccard` but those will be derived internally after computing the `two-hop neighbors` and the python user won't be aware of it? So if the python user passes a vertex pair, only the results containing the vertex pairs will be filter out? and if the python user doesn't pass the vertex pairs there is no filter and the result of the entire graph are returned?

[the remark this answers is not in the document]
ChuckHastings: @rlratzel - the documentation should be corrected to reflect the proper return value.  If the user of the C/C++ API wants the entire graph then the first call will be to construct the list of 2-hop neighbors.  I notice I did not add that call to the C API, I will add that in the next push.

@jnke2016 - Hopefully my comment to Rick addresses your question, if not see my next comment.

I guess we should discuss the python API expectation.  At the C/C++ level the direction we're taking here is to provide the tools and let them be used to provide the desired higher level experience.  I suspect there will be other uses for a function to return 2-hop neighbors beyond this one.  My thought would be to extend this model through PLC and leave it up to the python layer as to whether to require the caller to construct a list of 2-hop neighbors with one python call and pass it in as another, or (as we do today) to have the python call make the 2-hop neighbor list optional and if not specified to first call the PLC function to generate that list and then call the function to compute jaccard and after the function is complete delete the 2-hop neighbor list.

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+  int32_t align_;
+} cugraph_similarity_result_t;
+
+/**
+ * @brief       Get the first vertex id array
+ *
+ * @param [in]     result   The result from a similarity algorithm
+ * @return type erased array of vertex ids
+ */
+cugraph_type_erased_device_array_view_t* cugraph_similarity_result_get_first(
```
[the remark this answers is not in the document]
ChuckHastings: I should delete this function in the next push.  I did not ripple the changes Seunghwa requested fully into the C API.

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+ *
+ * @param [in]    result    The result from a similarity algorithm
+ */
+void cugraph_similarity_result_free(cugraph_similarity_result_t* result);
+
+/**
+ * @brief     Perform Jaccard similarity computation
+ *
+ * If (first, second) are specified (non-NULL) then compute the similarity of
+ * the specified vertex pairs.  If (first, second) are NULL then compute the similarity
```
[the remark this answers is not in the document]
ChuckHastings: Yes

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+ * @return type erased array of vertex ids
+ */
+cugraph_type_erased_device_array_view_t* cugraph_similarity_result_get_first(
+  cugraph_similarity_result_t* result);
+
+/**
+ * @brief       Get the second vertex id array
+ *
+ * @param [in]     result   The result from a similarity algorithm
+ * @return type erased array of vertex ids
```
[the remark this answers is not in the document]
ChuckHastings: Cleaned up in next push

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+ */
+cugraph_type_erased_device_array_view_t* cugraph_similarity_result_get_second(
+  cugraph_similarity_result_t* result);
+
+/**
+ * @brief       Get the similarity score array
```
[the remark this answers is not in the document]
ChuckHastings: Changed in next push

cpp/include/cugraph_c/similarity_algorithms.h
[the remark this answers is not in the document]
ChuckHastings: Changed in next push

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
+ * @param [in]  graph        Pointer to graph
+ * @param [in]  first        Optional type erased device array identifying first vertex,
+ *                           if NULL compute for all vertex pairs
+ * @param [in]  second       Optional type erased device array identifying first vertex,
+ *                           if NULL compute for all vertex pairs
+ * @param [in]  use_weight   If true consider the edge weight in the graph, if false use an
+ *                           edge weight of 1
+ * @param [in]  do_expensive_check A flag to run expensive checks for input arguments (if set to
+ * `true`).
+ * @param [out] result       Opaque pointer to paths results
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,170 @@
...
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

naimnv (approved):

jnke2016 (approved):
Looks good to me

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,135 @@
...
+ * @param [out] error        Pointer to an error object storing details of any error.  Will
+ *                           be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_jaccard_coefficients(const cugraph_resource_handle_t* handle,
+                                                  cugraph_graph_t* graph,
+                                                  const cugraph_vertex_pairs_t* vertex_pairs,
+                                                  bool_t use_weight,
+                                                  bool_t do_expensive_check,
+                                                  cugraph_core_result_t** result,
```
jnke2016, line 76: This is a typo. I believe you meant `cugraph_similarity_result_t`

cpp/include/cugraph_c/similarity_algorithms.h
```diff
@@ -0,0 +1,135 @@
...
+ * @param [out] error        Pointer to an error object storing details of any error.  Will
+ *                           be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_sorensen_coefficients(const cugraph_resource_handle_t* handle,
+                                                   cugraph_graph_t* graph,
+                                                   const cugraph_vertex_pairs_t* vertex_pairs,
+                                                   bool_t use_weight,
+                                                   bool_t do_expensive_check,
+                                                   cugraph_core_result_t** result,
```
jnke2016, line 103: This is a typo. I believe you meant `cugraph_similarity_result_t`

cpp/src/c_api/similarity.cpp
```diff
@@ -0,0 +1,151 @@
...
+  delete internal_pointer;
+}
+
+extern "C" cugraph_error_code_t cugraph_jaccard_coefficients(
+  const cugraph_resource_handle_t* handle,
+  cugraph_graph_t* graph,
+  const cugraph_vertex_pairs_t* vertex_pairs,
+  bool_t use_weight,
+  bool_t do_expensive_check,
+  cugraph_core_result_t** result,
```
jnke2016, line 117: This is a typo. I believe you meant `cugraph_similarity_result_t`.This occurs in more than one occurence in this file but I am only mentioning it once to avoid repeating this comment.

jnke2016 (changes requested):
Sorry I approved this PR too soon. Upon writing the python bindings for the `similarity` algos, i uncovered this typo which breaks the `pylibcugraph` build.

cpp/include/cugraph_c/graph_functions.h
```diff
@@ -0,0 +1,104 @@
...
+ * @param [in]  graph          Pointer to graph
+ * @param [in]  start_vertices Optional type erased array of starting vertices
+ *                             If NULL use all, if specified compute two-hop
+ *                             neighbors for these starting vertices
+ * @param [out] vertex_pairs   Opaque pointer to resulting vertex pairs
+ * @param [out] error          Pointer to an error object storing details of any error.  Will
+ *                             be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_two_hop_neighbors(
```
jnke2016, line 95: I am having a concern as I am trying to expose this function not just to `pylibcugraph` but also [cugraph](https://github.com/rapidsai/cugraph/blob/branch-22.10/python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py#L505). in python `cugraph`, this function will returns `first` and `second`(in fact, `pylibcugraoh` `get_two_hop_neighbors` will call `cugraph_create_vertex_pairs` which is fine. But the similarity algos needs pylibcugraph `get_two_hop_neighbors` to return `cugraph_vertex_pairs_t*` which will there be freed in the similarity algo. However, I think this can cause a leak if the user doesn't call the similarity algos. I think for now I can't make these two calls modulable(seperate) and call `cugraph_two_hop_neighbors` within the similarity algos. any thoughts?

ChuckHastings:
rerun tests

rlratzel (approved):

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

ChuckHastings:
rerun tests

ChuckHastings:
@gpucibot merge

PR #2630, opened 2022-08-30
Define a C API for data masking

### LINKED ISSUE #2571: Data Masking - Define C API
Define a C API for data masking.

ChuckHastings:
Add a C API for data masking.

Closes #2571 

The objective here is to add a C API that will allow us to associated a data mask with a graph.  Basic functions:
* Create a data mask
* Destroy a data mask
* Associated a data mask with a graph
* Disassociated a data mask from a graph
* Discover what data mask is associated with a graph

Current plan is to allow for the creation of a mask by the caller (the python layer for our current use case) and have it passed in.  We may later add support for other techniques.

This PR can be approved independently, but implementation of the API is dependent on #2679


ChuckHastings:
rerun tests

naimnv (commented):
@ChuckHastings  I wonder, how the association between a graph and mask would be kept?  
If the caller explicitly maintains that association then it's understood.
Apart from that, the API looks very transparent to me.

BradReesWork (approved):

naimnv (approved):

ChuckHastings:
> @ChuckHastings I wonder, how the association between a graph and mask would be kept?
> If the caller explicitly maintains that association then it's understood.
> Apart from that, the API looks very transparent to me.

My intention was to support the C API's graph object to maintain that connection.  Sequence of operations would be:
1. Create a data mask with the call `cugraph_data_mask_create`
2. Associate data mask with a graph with the call `cugraph_graph_add_data_mask`
3. Caller can now forget the details of the data mask, as it is part of the graph object
4. If you need to fetch the data mask, the caller can call `cugraph_graph_get_data_mask` to rediscover that object
5. `cugraph_graph_release_data_mask` removes the association of the data mask with the graph
6. `cugraph_data_mask_destroy` destroys the data mask.  This is required if the mask has been released from the graph, but if the mask is still associated with the graph then when the graph is destroyed this would automatically be called

naimnv (approved):

BradReesWork:
@gpucibot merge

PR #2660, opened 2022-09-02
Create graph with edge property values

seunghwak:
- [x] Move the CSR (+ DCSR, or CSC +DCSC) compression and adjacency list sort parts from graph_t to create_graph_from_edgelist; for the main code path, there is another graph_t constructor used in cython.cu, this constructor is still performing compress and sort, this function will be deleted once cython.cu retires. Currently, the compression and sort functions are moved to detail/structure_utils.cuh to be used in both create_graph_from_edgelist_impl.cuh and graph_impl.cuh.
- [x] Added another graph_t constructor taking compressed and sorted CSR + DCSR graph data
- [x] Updated the create_graph_from_edgelist function to ooptionally take edge IDs and types.
- [x] Made few std::optional objects in multi-GPU struct/class non-optional (as they will always have a valid value in multi-GPU).
- [x] Replace few more raw-pointers with raft::device_span


seunghwak:
rerun tests

alexbarghi-nv:
Is this ready for me to test with #2629 ?

seunghwak:
> Is this ready for me to test with #2629 ?

Hopefully in few hours, I will mark this as ready-for-review once this is ready to be tested with #2629. I have locally tested, but we have two versions of create_graph_edgelist (one does not take edge IDs and types and one OPTIONALLY takes edge IDs and types, as the first can be implemented with the second, I want to remove the first).

seunghwak:
@alexbarghi-nv And I am currently assuming that edge types are 32 bit signed integers, and will this work for you?

alexbarghi-nv:
I was hoping they would be uint8 to save memory; right now we don't have any use cases with anything more than a handful of edge types.

alexbarghi-nv:
But if they need to be int32 that's ok, it will just not be very memory-efficient

seunghwak:
> But if they need to be int32 that's ok, it will just not be very memory-efficient

OK, then, let's stick to int32_t at this moment, but once we find this becomes a memory bottleneck, we will get back to uint8_t. We can instantiate this for both uint8_t and uint32_t but we're getting some complaints about compile time and binary size, so we may want to avoid instantiating for too many type combinations. uint8 may not be sufficient for some future use cases but if this limits the maximum size we can run, we can use uint8 instead.

alexbarghi-nv:
Ok, sounds good.  I'll update my code to use int32

seunghwak:
@alexbarghi-nv OK, this PR is now ready for testing.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -102,13 +102,19 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
       rmm::device_uvector<vertex_t>* number_map =
         new rmm::device_uvector<vertex_t>(0, handle_.get_stream());
 
-      std::tie(*graph, new_number_map) = cugraph::
-        create_graph_from_edgelist<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>(
+      std::tie(*graph, std::ignore, new_number_map) =
+        cugraph::create_graph_from_edgelist<vertex_t,
+                                            edge_t,
+                                            weight_t,
+                                            int32_t,
```
ChuckHastings, line 109: This hard-codes int32_t, which is fine for now (as you've discussed with @alexbarghi-nv.  Ultimately we should probably allow this to be arbitrary, but I think we can defer that to later.

ChuckHastings (approved):

cpp/src/c_api/graph_sg.cpp
seunghwak: What about doing `using edge_type_t = int32_t` then passing `edge_type_t`? I guess this will make the intention clearer.

naimnv (approved):

naimnv (approved):

cpp/src/c_api/graph_sg.cpp
ChuckHastings: I think that would be a bit better.  That would make the references easier to isolate when we ultimately add support for variable types.

VibhuJawa (approved):
LGTM

ChuckHastings:
@gpucibot merge

PR #2661, opened 2022-09-02
Move GaaS sources, tests, docs, scripts from the rapidsai/GaaS repo to the cugraph repo

### LINKED ISSUE #2652: Move GaaS source code to cugraph repo
The outcome of various discussion around the name of `GaaS` (cuGraph Graph-as-a-Service) is that it needs to be renamed. The new name will be simply be `cugraph_service`.  The code will also be moved into the `cugraph` repo alongside the cugraph python package which it depends on.  This will greatly simplify coordinating PRs containing changes to be cugraph and cugraph_service, allow CI runs to verify the service didn't break, and other benefits of a monorepo approach.

Note that the conda package name(s) may use a hyphen (`cugraph-service`) based on conda package naming conventions (need to confirm).  We're avoiding using hyphens in directory names though to be consistent with subdirs that that need to be imported as packages in python which cannot have hyphens in them (eg. `from cugraph_service_client import client`)

A PR to do the following should be submitted:
* Move the GaaS sources to the cugraph repo as a separate top-level python package named `cugraph_service` (hopefully preserving git commit history)

One or more separate issues and PRs will be open for the following:
* Ensure all `cugraph_service` unit tests pass
* Update CI scripts, use commit meta-data features (already in place) to skip tests for cugraph, PLC, C, C++ when `cugraph_service` changes are made.

rlratzel:
closes #2652

**_Note to reviewers:_**
> The code here is based on the latest 22.10 GaaS sources which include [this last PR to GaaS](https://github.com/rapidsai/GaaS/pull/24), and has only been changed to update names ("gaas" -> "cugraph_service") and pass style checks, but the diffs will obviously show everything as new content.
>
> In other words, don't feel obligated to review anything beyond name changes if you've already reviewed the GaaS sources.
>
> Future changes that update functionality related to this move will be done in separate PRs so the diff can clearly show what's being changed.

This is the initial PR which only moves source files and changes names in sources, docs, directories, etc.  The unit tests were run (SG and MG) after the name change and they all pass.

One or more followup PRs will be open to add appropriate support for CI (ie. no MG tests run by default, limit to handler tests to simplify automating server subprocesses if necessary, etc.).

The entire GaaS repo is now under `python/cugraph_service` as a peer to `cugraph` and `pylibcugraph`.


BradReesWork (approved):

VibhuJawa (approved):
LGTM

BradReesWork:
@gpucibot merge

PR #2671, opened 2022-09-08
Promote triangle count from experimental

### LINKED ISSUE #2667: Promote `experimental.triangle_count` to stable
- [ ] Replace the current stable `triangle_count` API with `experimental.triangle_count`.  This will get us back to a single `triangle_count` API implementation, as intended.
- [ ] In order to maintain backwards compatibility for users that may have started using `experimental.triangle_count`, use the `promoted_experimental_warning_wrapper` to give those users 1-2 releases to stop using the experimental version.  After that, the experimental version will be removed completely.
- [ ] Update docs and tests if necessary

jnke2016:
In addition to replacing the current stable `triangle_count` API with `experimental.triangle_count` and updating the docs and test, this PR also adds support for the parameter `start_list` for both the SG and the MG implementation. It also renames `triangle_count` to `triangles` to be consistent with the previous API

closes #2667


BradReesWork:
@gpucibot merge

PR #2679, opened 2022-09-09
`graph_mask_t` and separating raft includes for `host_span` and `device_span` 

### LINKED ISSUE #2322: [FEA] Prototype vertex and edge masking approach
#2218 defined the context of the problem and some initial designs.

The goal in 22.08 is to prototype something so that we can build a capability to deliver in 22.10.

cjnolet:
This depends on https://github.com/rapidsai/raft/pull/810. 

Closes #2322


cjnolet:
rerun tests

cpp/include/cugraph/graph_mask.hpp
```diff
@@ -0,0 +1,382 @@
...
+}
+
+/**
+ * Uses bit-shifting to perform a fast mod operation. This
+ * is used to compute the index of a specific bit
+ * @tparam mask_t
+ * @tparam T
+ */
+template <typename mask_t, typename T>
+__host__ __device__ int bit_mod(T numer)
```
seunghwak, line 54: I assume `numer` here is a typo of `number`.

cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -55,8 +55,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION    ${CUGRAPH_MIN_VERSION_raft}
-                        FORK       rapidsai
-                        PINNED_TAG branch-${CUGRAPH_BRANCH_VERSION_raft}
+                        FORK       cjnolet
+                        PINNED_TAG imp-2210-host_device_mdspan
```
seunghwak, line 59: Don't forget to revert this.

seunghwak (approved):
Looks good to me (but I assume this should be merged after the raft side PR gets merged and the change in `get_raft.cmake` gets reverted.

cpp/include/cugraph/graph_mask.hpp
cjnolet: It's actually the `numerator` of the division for the mod but i can change it to `number` if that's what you prefer

seunghwak: I see, then, this is not a big issue, but I slightly prefer either `number` or `numerator` than `numer`. It is little difficult to infer `numerator` from `numer`.

cjnolet:
rerun tests

cjnolet:
rerun tests

ChuckHastings (approved):

rlratzel (approved):

raydouglass (approved):

rlratzel:
@gpucibot merge

PR #2684, opened 2022-09-12
Added `SamplingResult` cdef class to return cupy "views" for PLC sampling algos instead of copying result data

rlratzel:
## Summary
* (partially) closes rapidsai/GaaS#23
* Added `SamplingResult` cython extension class that corresponds to the C API's `cugraph_sample_result_t` type, which removes a copy operation in sampling algos.
* Added new testing APIs and unit tests to validate correctness and that memory is cleaned up as expected when all refs to results are removed (see `test_neighborhood_sampling.py::test_sample_result`)
* Updated `build.sh` to use `ninja` by default.

## Background
As part of a larger effort to provide [GPU-GPU data transfer in `cugraph_service`](https://github.com/rapidsai/GaaS/issues/23) to improve performance, the need to minimize redundant data copy operations was also something to address.

Prior to this change, PLC algos would copy their results from the C API device data structures into one or more cupy ndarrays, which allowed the algo to then delete the result data owned by the C API and leave the lifecycle management of the results returned to python to the python interpreter/GC.  The copy operation not only slowed down sampling algo calls (slightly) but also caused spikes in GPU memory usage where two copies of results existed in GPU memory for a short time immediately after the copy but before the original was deleted.  

This PR adds a cython extension class that corresponds to the C API's `cugraph_sample_result_t` type, which is used to pass ownership of the sampling results to python.  The extension class calls `cugraph_sample_result_free()` when all references to the result data in python are removed and the garbage collector runs.  The owning `SamplingResult` object allows PLC sampling algos to instead return cupy ndarray "views" for results instead of copying them to new cupy ndarrays.

New test APIs to create sampling result objects in C and PLC were added and used in the new tests to verify that the ownership of the result data was properly transferred to Python, and that the lifecycle management of results was done properly when clients keep references to data (eg. pass results to other parts of code for use long after the algo completes).

A benchmark was also added to verify the (minor) speedup. Runs `0001-2` were before the change, runs `0003-5` (red box) were after:
image


rlratzel:
I believe the CI test failures are related to the changes in this PR, so I'm looking into those.

cpp/src/c_api/uniform_neighbor_sampling.cpp
```diff
@@ -222,6 +222,80 @@ extern "C" cugraph_type_erased_host_array_view_t* cugraph_sample_result_get_coun
...
+    if (!handle) {
+      *error = reinterpret_cast<cugraph_error_t*>(
+        new cugraph::c_api::cugraph_error_t{"invalid resource handle"});
+      return CUGRAPH_INVALID_HANDLE;
+    }
+
+    auto p_handle = reinterpret_cast<cugraph::c_api::cugraph_resource_handle_t const*>(handle);
+    auto host_srcs =
+      reinterpret_cast<const cugraph::c_api::cugraph_type_erased_host_array_view_t*>(srcs);
+    auto new_device_srcs = new cugraph::c_api::cugraph_type_erased_device_array_t(
```
ChuckHastings, line 248: I would be inclined to do this a bit differently.  The values `new_device_srcs`, `new_device_dsts`, `new_device_wgts` and `new_device_counts` that you're creating here are  a potential for a memory leak, since they are not in RAII objects.

You could:
1. Do type dispatching and create these as `rmm::device_uvector<appropriate_type>` and then convert to the return value.  This would mimic the pattern of the algorithm implementation.
2. Create these as `std::unique_ptr` values and release them when you convert to the return value
3. Create the return value as a `std::unique_ptr` (releasing upon return) with the new calls for these operators inside and then just use the values from the return object.  We do know the sizes a priori

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -268,6 +268,30 @@ cugraph_type_erased_host_array_view_t* cugraph_sample_result_get_counts(
...
+ * @param [in]   weights        Host array view to populate weights
+ * @param [in]   counts         Host array view to populate counts
+ * @param [out]  result         Pointer to the location to store the
+ *                              cugraph_sample_result_t*
+ * @param [out]  error          Pointer to an error object storing details of
+ *                              any error.  Will be populated if error code is
+ *                              not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_sample_result_create(
```
ChuckHastings, line 286: Is the intention to leave this testing API in the final API, or to use this until the actual implementation is available?

If the plan is to leave it in the API permanently, I think we should put the word `test` somewhere in the function name.  Perhaps `cugraph_test_sample_result_create` to indicate that it's a test API.

cpp/src/c_api/uniform_neighbor_sampling.cpp
```diff
@@ -222,6 +222,80 @@ extern "C" cugraph_type_erased_host_array_view_t* cugraph_sample_result_get_coun
...
+      host_srcs->size_, host_srcs->num_bytes_, host_srcs->type_, p_handle->handle_->get_stream());
+    raft::update_device(reinterpret_cast<byte_t*>(new_device_srcs->data_.data()),
```
ChuckHastings, line 250: If you're not going to do type dispatching, consider the following.

You could actually call `cugraph_type_erased_device_array_create` and `cugraph_type_erased_device_array_view_copy_from_host` and leverage some of the already written code.  This also gets you around having to do so much type casting.  This might be a bit cleaner (untested):
```
cugraph_type_erased_device_array_t *new_src{nullptr};
err = cugraph_type_erased_device_array_create(handle, host_srcs->size_, host_srcs_->type_, &new_src, error);
// check the value of err, return if it's not CUGRAPH_SUCCESS
auto new_device_sources_view = cugraph_type_erased_device_array_view(new_device_sources);
cugraph_type_erased_device_array_view_copy_from_host(
  handle,
  & new_device_sources_view,
  host_srcs->data_->as_type<byte_t>(),
  error);
// check the value of err, return if it's not CUGRAPH_SUCCESS
```

You might not need, then, to create p_handle and a few of the other things you're doing.

cpp/include/cugraph_c/sampling_algorithms.h
rlratzel: I was thinking it would be kept in the library for unit testing code that handles sample results, without requiring that a sampling algo be called.  I'll rename the function `cugraph_test_sample_result_create` as suggested.

cpp/src/c_api/uniform_neighbor_sampling.cpp
rlratzel: Instead of this suggestion, I changed the implementation to use other cugraph C API calls (as suggested below), which included using `cugraph_type_erased_device_array_create_from_view()` from #2712 . Let me know if you'd prefer I use the `rmm:device_uvector`/type dispatching approach instead.

cpp/src/c_api/uniform_neighbor_sampling.cpp
rlratzel: I did this in the most recent commit, and I also used cugraph_type_erased_device_array_create_from_view()` from #2712 , based on our discussion offline.

rlratzel:
rerun tests

reason: looks like a gpuCI failure trying to access the repo, trying again...

rlratzel:
rerun tests

reason: one last try before assuming Ops needs to intervene

raydouglass:
rerun tests

rlratzel:
rerun tests

reason: ops indicated they fixed the gpuCI issue.

cpp/src/c_api/uniform_neighbor_sampling.cpp
```diff
@@ -222,6 +222,62 @@ extern "C" cugraph_type_erased_host_array_view_t* cugraph_sample_result_get_coun
...
+    *error = reinterpret_cast<cugraph_error_t*>(
+      new cugraph::c_api::cugraph_error_t{"invalid resource handle"});
+    return CUGRAPH_INVALID_HANDLE;
+  }
+
+  // copy srcs to new device array
+  cugraph_type_erased_device_array_t* new_device_srcs{nullptr};
+  error_code =
+    cugraph_type_erased_device_array_create_from_view(handle, srcs, &new_device_srcs, error);
+  if (error_code != CUGRAPH_SUCCESS) return error_code;
```
ChuckHastings, line 249: This looks much cleaner.

I think this still potentially leaks memory.  Putting the comment here, but probably this one is fine, the rest would be potentially a problem.  If one of the `cugraph_type_erased_device_array_create_from_view` functions fails, then you will return an error code here, but any of the previously allocated device arrays will leak.

Seems like the best option would be to make that an `std::unique_ptr` and define it's freeing function as the `cugraph_type_erased_device_array_free` function (and do that for each of the device array's you create in this function).

Then in the line where we create the return value you can release the unique pointer when calling the constructor.

NOTE: This probably exists in other functions as well, I just noticed it here.

rlratzel: Thanks, I added this to my latest commit.

ChuckHastings (approved):

Ethyling (approved):

BradReesWork:
is there, or should there, be a macro for `if (error_code != CUGRAPH_SUCCESS) return error_code;`?

BradReesWork (approved):

galipremsagar (approved):

galipremsagar:
rerun tests

rlratzel:
@gpucibot merge

PR #2688, opened 2022-09-13
Fixes bug in `NumberMap` preventing use of string vertex IDs for MG graphs

### LINKED ISSUE #2686: How to read the graph edge if the node value is non-numerical
Hi team, I am exploring the cugraph, and the base image is `rapidsai/rapidsai:cuda11.0-base-ubuntu20.04-py3.8`. 

I have tested on the input with node value being numerical, and it worked fine. But it failed when the input become a string. the error occured inside `dcg.pagerank(G_dask, tol=1e-4)`

```
import time
import cudf
import cugraph
import cugraph.dask as dcg
import dask_cudf
import pandas as pd
from cugraph.dask.comms import comms as Comms
from dask.distributed import Client, wait
from dask_cuda import LocalCUDACluster


def setup():
    cluster = LocalCUDACluster(threads_per_worker=1)
    client = Client(cluster)
    Comms.initialize(p2p=True)
    return (client, cluster)


def teardown(client, cluster=None):
    Comms.destroy()
    client.close()
    if cluster:
        cluster.close()

if __name__ == "__main__":

    # setting up the cluster
    setup_objs = setup()

    # dummy data
    df = pd.DataFrame(
        {
            "src": ['a1', 'a1', 'a2', 'a3'],
            "dst": ['a2', 'a3', 'a4', 'a4'],
        }
    )
    input_file = "dummy.csv"
    df.to_csv(input_file, index=False, header=False, sep="\t")

    chunksize = dcg.get_chunksize(input_file)
    e_list = dask_cudf.read_csv(
        input_file, chunksize=chunksize, delimiter="\t", names=["src", "dst"], dtype=[str, str]
    )

    G_dask = cugraph.DiGraph()
    G_dask.from_dask_cudf_edgelist(e_list, source="src", destination="dst")
    pr_df = dcg.pagerank(G_dask, tol=1e-4)
    logger.info(f"print out {pr_df.head(4)}")

    # tearing fown the cluster
    teardown(*setup_objs)
```

rlratzel:
fixes #2686 

### Summary
* Moved assignment of a variable that sets a dtype to a point _after_ the mappings from vertex strings to integer IDs are applied in order to capture the correct dtype of the internal dataframes. This bug was not detected earlier since integer vertex IDs had matching dtypes before and after any initial mappings were applied, and there was no test coverage for MG string vertex IDs.
* Added test that uses a graph containing string vertex IDs then runs `pagerank` using both SG and MG versions and ensures the results are identical.  This new test was based on the code provided in issue #2686 
* Fixed several bugs in MG test code that prevented `pytest` from collecting tests when the entire `mg` suite was specified. This probably would have been uncovered sooner if the MNMG nightly tests were configured to run differently (they run individual test files instead of having `pytest` collect all tests from the `mg` directory).

### DIFF
--- a/python/cugraph/cugraph/tests/mg/test_mg_renumber.py
+++ b/python/cugraph/cugraph/tests/mg/test_mg_renumber.py
@@ -15,13 +15,15 @@
 
 import gc
 import pytest
+import tempfile
+from pathlib import Path
 
 import pandas
 import numpy as np
 import dask_cudf
 import dask
 import cudf
-from cudf.testing import assert_series_equal
+from cudf.testing import assert_frame_equal, assert_series_equal
 
 import cugraph.dask as dcg
 import cugraph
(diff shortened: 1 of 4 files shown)


python/cugraph/cugraph/tests/mg/test_mg_renumber.py
```diff
@@ -287,3 +289,57 @@ def test_mg_renumber_common_col_names(graph_file, dask_client):
...
+    Note: the dask_client fixture sets up and tears down a LocalCUDACluster.
+    See ../conftest.py
+    """
+    # Use pandas and to_csv() to create a CSV file that can be read in by both
+    # dask_cudf and cudf.
+    df = pandas.DataFrame({"src": ['a1', 'a1', 'a2', 'a3'],
+                           "dst": ['a2', 'a3', 'a4', 'a4'],
+                           }
+                          )
+    tempdir_object = tempfile.TemporaryDirectory()
```
ChuckHastings, line 310: Is this mechanism necessary?  Isn't it possible to take a pandas Dataframe and convert it to a dask Dataframe directly without going to disk first?

rlratzel: yes, I debated if we should do it this way in the tests or if we should just go directly from an inline pandas/cudf DataFrame to a dask_cudf one (as is done elsewhere).  I ultimately did it this way to match the user's script [here](https://github.com/rapidsai/cugraph/issues/2686), just to be thorough. However, I can remove this in favor of our standard mechanism which doesn't involve a temp file.

rlratzel: On second thought, I think I will simplify the test and not use a tempfile with `read_csv()`

BradReesWork:
LGTM - would like to see single GPU dask testing

BradReesWork (approved):

rlratzel:
> LGTM - would like to see single GPU dask testing

Good idea, I filed #2714 for that.

jnke2016 (approved):

rlratzel:
@gpucibot merge

PR #2697, opened 2022-09-14
Renumber PG to be contiguous per type

### LINKED ISSUE #2627: [FEA] Renumber PropertyGraphs to be contiguous per type basis

We should have a utility to renumber PropertyGraphs to be contiguous per type basis. 

**Motivation**
In frameworks like DGL, PyG the  Nodes/Edges of different types have independent ID space and feature storage. 
See below example from [DGL documentation](https://docs.dgl.ai/en/latest/guide/graph-heterogeneous.html#guide-graph-heterogeneous), where the user and game node IDs both start from zero and they have different features.  
image

To easily convert b/w cugraphs PG to DGL if we renumber it to be  contiguous per type the cost of conversion just becomes a subtract operation.  (See below):

```python
def dgl_n_id_to_cugraph_id(index, nid_d, ntype):
    return index + nid_d[ntype]
    

def cugraph_n_id_to_dgl_id(index, nid_d, ntype):
     return index - nid_d[ntype]
````

**Requested API**

```python
def pg.renumber_per_ntype():
       '''
       Renumbers node ids per node type
       
       Returns:
          node_type_offset_d (dict):  A dictionary containing the start and stop ids for each node type  
       '''
                    
```

```python
def pg.renumber_per_etype():
       '''
       Renumbers node ids per edge type
       
       Returns:
          edge_type_offset_d (dict):  A dictionary containing the start and stop ids for each edge type  
       '''              
```

CC:  @alexbarghi-nv: For inputs from the PyG side

CC: @eriknw , @rlratzel ,  @BradReesWork

eriknw:
Closes #2627 and CC @VibhuJawa 

Currently only SG as we hammer out the API and behavior.

This returns a dataframe with start and stop for each type.  Should stop be inclusive or exclusive?

How should we handle vertex ids that only exist in edge data?  Should we raise (for now) if this condition exists?  I think we can handle this without _too_ much difficulty, but it will take more work.

Since we number edge data, I think edge IDs will often be added in a way that is already contiguous per type.  We could keep track of this to avoid unnecessary computation.

Also, I want to confirm that we _cannot_ have multiple rows for a single vertex ID, right?  I think we settled on this.  Multiple rows with the same ID would cause a problem with the current implementation--it currently gives each row a unique ID.

### DIFF
--- a/python/cugraph/cugraph/dask/structure/mg_property_graph.py
+++ b/python/cugraph/cugraph/dask/structure/mg_property_graph.py
@@ -877,6 +877,94 @@ class EXPERIMENTAL__MGPropertyGraph:
 
         return G
 
+    def renumber_vertices_by_type(self):
+        """Renumber vertex IDs to be contiguous by type.
+
+        Returns a DataFrame with the start and stop IDs for each vertex type.
+        Stop is *inclusive*.
+        """
+        # Check if some vertex IDs exist only in edge data
+        default = self._default_type_name
+        if (
+            self.__edge_prop_dataframe is not None
+            and self.get_num_vertices(default, include_edge_data=True)
+            != self.get_num_vertices(default, include_edge_data=False)
+        ):
+            raise NotImplementedError(
+                "Currently unable to renumber vertices when some vertex "
+                "IDs only exist in edge data"
+            )
+        if self.__vertex_prop_dataframe is None:
+            return None
+        # We'll need to update this when index is vertex ID
+        df = (
+            self.__vertex_prop_dataframe
+            .sort_values(by=self.type_col_name)
+        )
+        if self.__edge_prop_dataframe is not None:
+            new_name = f"new_{self.vertex_col_name}"
+            df[new_name] = 1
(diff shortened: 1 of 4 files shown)


VibhuJawa:
> How should we handle vertex ids that only exist in edge data? Should we raise (for now) if this condition exists? I think we can handle this without _too_ much difficulty, but it will take more work.

I think we dont need it for DGL usecases .


> I want to confirm that we cannot have multiple rows for a single vertex ID, right? I think we settled on this. Multiple rows with the same ID would cause a problem with the current implementation--it currently gives each row a unique ID.

Yup. One vertex id for now for our storage.

eriknw:
Great, thanks @VibhuJawa.  This is ready for review.

`stop` is now inclusive.  Renumbering vertex IDs raises if there are vertex IDs that only exist in edge data.

VibhuJawa (approved):
LGTM

alexbarghi-nv (approved):
👍

rlratzel (approved):

rlratzel:
@gpucibot merge

PR #2698, opened 2022-09-14
PG allow get_vertex_data to accept single type or id

### LINKED ISSUE #2692: [FEA] Property Graph get_vertex_data accept int
currently get_vertex_data only accepts an array of vertex IDs, it should also support a single int as input

---

  - [] Code passes CI
  - [] Code uses Graph Primitives
  - [] Code in C++/CUDA layer
  - [] Code in C layer
  - [] Code in cugraph Python layer
  - [] Documentation

eriknw:
Closes #2692.

Currently SG only.  ~MG version coming shortly!~  MG version done.

### DIFF
--- a/python/cugraph/cugraph/dask/structure/mg_property_graph.py
+++ b/python/cugraph/cugraph/dask/structure/mg_property_graph.py
@@ -437,18 +437,20 @@ class EXPERIMENTAL__MGPropertyGraph:
         vertex_ids, columns, and/or types, or all vertex IDs if not specified.
         """
         if self.__vertex_prop_dataframe is not None:
+            df = self.__vertex_prop_dataframe
             if vertex_ids is not None:
-                df_mask = (
-                    self.__vertex_prop_dataframe[self.vertex_col_name]
-                    .isin(vertex_ids)
-                )
-                df = self.__vertex_prop_dataframe.loc[df_mask]
-            else:
-                df = self.__vertex_prop_dataframe
+                if isinstance(vertex_ids, int):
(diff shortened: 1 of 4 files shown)


python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -437,18 +437,20 @@ def get_vertex_data(self, vertex_ids=None, types=None, columns=None):
...
-            else:
-                df = self.__vertex_prop_dataframe
+                if isinstance(vertex_ids, int):
+                    df_mask = df[self.vertex_col_name] == vertex_ids
+                else:
+                    df_mask = df[self.vertex_col_name].isin(vertex_ids)
+                df = df.loc[df_mask]
 
             if types is not None:
                 # FIXME: coerce types to a list-like if not?
```
rlratzel, line 449: is this FIXME still relevant, since it seems like we decided we aren't coercing to a list?  Same question for all the other FIXMEs like this.

rlratzel (approved):
I had a question about a few old FIXMEs which need not hold up approval.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
eriknw: well, we check if it's a string, but we don't yet check if it's a list.  We defer to `isin`.  so if we don't have a clear answer to this FIXME, then it _may_ still be relevant.

rlratzel:
@gpucibot merge

PR #2707, opened 2022-09-19
suppress expansion of unused raft spectral templates

### LINKED ISSUE #2678: spectral_clustering build is suspiciously slow and contributes to CI build timeouts
We need to look at why `spectral_clustering.cu` takes ~10x the time to compile vs. all others.  On my personal workstation I observed >6 minutes to build this file vs. <0.5 minutes for others.
We have been seeing several CI build timeouts lately for specific platforms, and the excessively long time to build spectral_clustering is believed to be a contributing factor.  Plus, local development from a clean build is impacted for every developer.

ChuckHastings:
Closes #2678 

raft updated some k-means logic and dramatically increased the compile time of our spectral clustering implementation.  This PR adds an include that will suppress the expansion of templates that we are not using.

### DIFF
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -38,8 +38,8 @@ requirements:
   host:
     - doxygen {{ doxygen_version }}
     - cudatoolkit {{ cuda_version }}.*
-    - libraft-headers {{ minor_version }}.*
-    - libraft-distance {{ minor_version }}.*
+    - libraft-headers {{ minor_version }}
+    - libraft-distance {{ minor_version }}
     - libcugraphops {{ minor_version }}.*
     - librmm {{ minor_version }}.*
     - libcudf {{ minor_version }}.*
@@ -63,8 +63,8 @@ outputs:
         - cmake {{ cmake_version }}
       run:
         - cudatoolkit {{ cuda_spec }}
-        - libraft-headers {{ minor_version }}.*
-        - libraft-distance {{ minor_version }}.*
+        - libraft-headers {{ minor_version }}
+        - libraft-distance {{ minor_version }}
         - librmm {{ minor_version }}
         - nccl {{ nccl_version }}
         - ucx-proc=*=gpu
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -63,11 +63,8 @@ outputs:
         - cmake {{ cmake_version }}
       run:
         - cudatoolkit {{ cuda_spec }}
-<<<<<<< HEAD
-        - libraft-headers {{ minor_version }}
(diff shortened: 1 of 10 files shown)


ChuckHastings:
rerun tests

cpp/CMakeLists.txt
```diff
@@ -146,6 +146,7 @@ include(${rapids-cmake-dir}/cpm/cuco.cmake)
 rapids_cpm_cuco(BUILD_EXPORT_SET cugraph-exports)
 
 include(cmake/thirdparty/get_raft.cmake)
+find_package(raft COMPONENTS distance)
```
cjnolet, line 149: `get_raft.cmake` should be doing this now so there shouldn't be a need for this line

ChuckHastings: Fixed

conda/recipes/libcugraph/meta.yaml
```diff
@@ -63,6 +63,11 @@ outputs:
         - cmake {{ cmake_version }}
       run:
         - cudatoolkit {{ cuda_spec }}
+<<<<<<< HEAD
+        - libraft-headers {{ minor_version }}
+        - libraft-distance {{ minor_version }}
+=======
+>>>>>>> branch-22.10
```
rlratzel, line 70: Looks like a merge conflict was committed. This is why CI is failing.

rlratzel (changes requested):

ChuckHastings:
rerun tests

conda/recipes/libcugraph/meta.yaml
```diff
@@ -38,9 +38,9 @@ requirements:
   host:
     - doxygen {{ doxygen_version }}
     - cudatoolkit {{ cuda_version }}.*
+    - libraft-headers {{ minor_version }}.*
+    - libraft-distance {{ minor_version }}.*
```
cjnolet, line 42: I think this might also need to be added to pylibcugraph and cugraph

ChuckHastings: Done in latest push

cjnolet:
@vyasr do you have any ideas as to why this would still be failing when libraft-distance is installed as a dependency of cugraph/pylibcugraph? I'm scratching my head a little bit.

ChuckHastings:
> @cjnolet @ChuckHastings If you're pulling libraft-distances then I believe the target that you should be linking to is `raft::raft_distance_lib` instance of `raft::distance`.

Trying that now, thanks!

vyasr:
Hmm apparently not the issue? Can you reproduce this locally? I can give that a shot if you haven't been able to.

ChuckHastings:
Corey fixed this in #2739 ,  we'll merge that PR instead.

vyasr:
Ahhh I think I see the issue.

PR #2712, opened 2022-09-21
Define k-core API and tests

### LINKED ISSUE #2631: MNMG K-core - Define C++ API
Define C++ API for MNMG K-Core

### LINKED ISSUE #2632: MNMG K-core - Define C API
Define C API for MNMG K-core

### LINKED ISSUE #2633: MNMG K-core - Define C++ unit tests
Define C++ unit tests for MNMG K-core

### LINKED ISSUE #2635: MNMG K-core - Define C unit tests
Add C unit tests for MNMG K-core

ChuckHastings:
Define k-core C++ API and tests.

Closes #2631 
Closes #2632 
Closes #2633 
Closes #2635


cpp/include/cugraph/algorithms.hpp
```diff
@@ -1606,6 +1606,31 @@ void core_number(raft::handle_t const& handle,
...
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param  graph           cuGraph graph in coordinate format
+ * @param  k               Order of the core. This value must not be negative.
+ * @param  core_numbers    Output from core_number algorithm
+ *
+ * @return edge list for the graph
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
```
naimnv, line 1624: Would it be a better idea to pass `bool transpose=false` as template parameter and then use transpose as a named parameter to the function instead of implicit false?

cpp/src/cores/k_core_impl.cuh
```diff
@@ -0,0 +1,37 @@
...
+#pragma once
+
+#include <cugraph/algorithms.hpp>
+#include <cugraph/graph_view.hpp>
+#include <cugraph/utilities/error.hpp>
+
+namespace cugraph {
+
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
```
naimnv, line 25: Same here as mentioned in my previous comment.

naimnv (approved):

BradReesWork (approved):

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1606,6 +1606,31 @@ void core_number(raft::handle_t const& handle,
...
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<weight_t>>>
+k_core(raft::handle_t const& handle,
```
seunghwak, line 1628: Something to think about...

So, we're passing core numbers here. So in this case, should we really have both this function and induced subgraph?

seeing core numbers, we can easily extract a vertex list in a single thrust call (e.g. thrust::copy_if). Then, we can pass a vertex list to induced subgraph.

Not sure this function adds enough convenience to justify increase in compile time/binary size.

seunghwak: Maybe OK, if this calls explicitly instantiated induced_subgraph, increases in compile time/binary size might be minimal...

Still debating between whether this function should take core_numbers or just call core_numbers internally.... (to maximize user convenience, if not, a user can just separately call core_numbers, something like thurst::copy_if, and induced_subgraph).

cpp/include/cugraph/algorithms.hpp
ChuckHastings: We try to avoid excess compilation of different templated functions, as this increased compile time.

*Most* of our functions are implemented based on an assumption about whether the graph is stored transposed or not.  For example, `pagerank` assumes that `store_transposed=true`.  Implementing `pagerank` if `store_transposed=false` would result in a great deal more synchronization (and communication in a multi-gpu environment).  There are a few examples that will work on a graph in either orientation.

Supporting this would require providing an implementation that would work with either orientation of the graph.  It seems like as long as this matches the requirement for `core_number` (which has to be called first) then just one orientation is sufficient.

cpp/src/cores/k_core_impl.cuh
ChuckHastings: See above reaction.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I had that debate internally.  Ultimately I concluded that if we're going to expose a complete k-core implementation at the python level we should probably expose it at the lower levels as well.  It seems like something that would be useful.

I do wonder if we should make passing in `core_numbers` optional, and if they are not passed then also call `core_numbers`.  This would allow a simple caller that just wants to extract a 3-core from the graph to be able to call the function directly and get the complete answer, while a more sophisticated case might call `core_number` once and reuse the result to extract the different subgraphs as required.

seunghwak: Yeah... +1 for taking `core_numbers` as std::optional.

Still a bit not-sure about "Ultimately I concluded that if we're going to expose a complete k-core implementation at the python level we should probably expose it at the lower levels as well."

At the python layer, user convenience might be more important than anything else. And providing an additional python function that internally calls core_number, find vertex list, and induced subgraph has no toll on compile time/binary size as python is an interpreted language.

Not sure we should provide the same level of convenience for C++ users, as we assume C++ users are more advanced and they might be OK about finding k-core by composing core_numbers and induced subgraph.

But this is a much bigger topic, and we may have this discussion again in the future.

seunghwak (approved):

cpp/include/cugraph/algorithms.hpp
seunghwak: And another thing to consider is that this may incur some sort of dependency within a same algorithm level in our C++ software hierarchy. Most algorithms are composed of primitives and thrust calls. But now algorithms are composed of primitives, thrust calls, and other algorithms. I am not sure what kind of complications this can incur in the future.

Composing C++ algorithms in the C layer might be OK, but I am not sure whether we should support use cases that can be supported simply by composing few existing algorithms in the C++ level.

ChuckHastings: Fair point.

I will change `core_numbers` to be optional.

We can explore dropping the `k_core` from the C++ layer if having the algorithms depend on other algorithms becomes an issue.  I see the point that adding it at the C layer may be sufficient.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1606,6 +1606,31 @@ void core_number(raft::handle_t const& handle,
...
+/**
+ * @brief   Extract K Core of a graph
+ *
+ * @throws     cugraph::logic_error when an error occurs.
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param  graph           cuGraph graph in coordinate format
```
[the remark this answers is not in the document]
ChuckHastings: Fixed

jnke2016:
I don't see where these `k-core` functions are implemented: `cugraph_core_result_get_src_vertices`, `cugraph_core_result_get_dst_vertices`, `cugraph_core_result_get_weights` are implemented. Please can you point me to those? I am getting the error below in `pylibcugraph`
```
'cugraph_core_result_get_src_vertices' is not a constant, variable or function identifier
```

ChuckHastings:
Good catch, Joseph.  Adding the missing functions.

ChuckHastings:
I see another issue, need to fix something else also.

jnke2016:
Ok thanks. It looks like `cugraph_k_core` too is missing

jnke2016 (approved):

ChuckHastings:
rerun tests

ChuckHastings:
@gpucibot merge

PR #2715, opened 2022-09-22
Adds API option to `uniform_neighbor_sample()` and UCX-Py infrastructure to allow for a client-side device to directly receive results

rlratzel:
closes rapidsai/GaaS#23

### Summary
* Adds an API option to `uniform_neighbor_sample()` to allow clients to specify a client-side GPU (using device number) which will receive the results, eliminating the expensive server-side device-host copy in order to serialize and send via RPC.
* Adds test(s) and benchmarks to cover the new feature
* Also adds a `--pydevelop` option to `build.sh` for building python components that allow for making edits without re-installing (runs `setup.py` with the `develop` option)

### Details
* This PR depends on #2684 

### Benchmarks
New benchmarks have been added by this PR, with the results shown below. The results to take note of are in the red rectangles. `device=None` is the previous version using serialization and transfer from the server to the client using host memory (using Apache Thrift), and `device=0` is a GPU-GPU transfer using UCX-Py from the server on GPU1 to the client on GPU0.  A typical data transfer for many GNN use cases involves ~1M values, which is what the highlighted benchmark simulates.
For smaller transfers, the Thrift-based RPC is faster, but increases significantly as the payload increases (so much so that the largest size that could be run in a reasonable time for Thrift was 1e6, vs UCX-Py sizes of 2e9 still being much faster). UCX-Py seems to have some amount of fixed overhead, but the benefits are dramatic at larger sizes.

image


python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -169,4 +169,25 @@ def uniform_neighbor_sample(
...
-    return ddf
+    # FIXME: this will convert to a dask_cudf.DataFrame even if arrays are to
+    # be returned so the unrenumbering step above can be used. A better
+    # approach would be to update the renumbering code to accept cupy arrays to
+    # renumber.
+    if (
+        hasattr(uniform_neighbor_sample, "_return_type")
+        and uniform_neighbor_sample._return_type == "arrays"
+    ):
+        uniform_neighbor_sample._return_type = ""
```
VibhuJawa, line 180: Is this overwrite of `uniform_neighbor_sample._return_type` intentional ?

VibhuJawa: Just saw the test, Can you add a comment here to explain why you over-write `_return_type` here

python/cugraph_service/README.md
```diff
@@ -13,18 +13,25 @@
...
+Starting a server for single-GPU-only cuGraph, using server extensions in `/my/cugraph_service/extensions`:
+```
+$> PYTHONPATH=/Projects/cugraph/python/cugraph_service python -m cugraph_service_server.server --graph-creation-extension-dir=/my/cugraph_service/extensions
+```
+
+Starting a server for multi-GPU cuGraph, same extensions:
+```
+$> export SCHEDULER_FILE=/tmp/scheduler.json
+$> /Projects/cugraph/python/cugraph_service/scripts/run-dask-process.sh scheduler workers &
+$> PYTHONPATH=/Projects/cugraph/python/cugraph_service python -m cugraph_service_server.server --graph-creation-extension-dir=/my/cugraph_service/extensions --dask-scheduler-file=$SCHEDULER_FILE
```
VibhuJawa, line 32: I suggest using an environment variable to set path and not have hard_coded path here.

python/cugraph_service/README.md
```diff
@@ -13,18 +13,25 @@
...
 ## Client
 (description)
 ### Installing the `cugraph_service_client` conda package
 
     TBD
 
+### Example
+Starting a server for single-GPU-only cuGraph, using server extensions in `/my/cugraph_service/extensions`:
+```
+$> PYTHONPATH=/Projects/cugraph/python/cugraph_service python -m cugraph_service_server.server --graph-creation-extension-dir=/my/cugraph_service/extensions
```
VibhuJawa, line 25: I suggest using an environment variable to set path and not have hard_coded path here.

python/cugraph_service/cugraph_service_server/cugraph_handler.py
```diff
@@ -677,20 +679,100 @@ def uniform_neighbor_sample(
...
+                        "be set if either is set. Got: "
+                        f"{result_host=}, {result_port=}"
+                    )
+                asyncio.run(
+                    self.__ucx_send_results(
+                        result_host,
+                        result_port,
+                        uns_result.sources,
+                        uns_result.destinations,
+                        uns_result.indices,
```
VibhuJawa, line 702: Curious on how would we transfer a dict of indices: {type: ['sources' ,'destinations', 'indices'} something that looks like that

VibhuJawa (changes requested):
Have requested some clarifications and minor changes. Looks good otherwise.

jnke2016 (approved):

alexbarghi-nv (commented):
Looks good; just one comment on whether we still need the buffer wrappers.

python/pylibcugraph/pylibcugraph/device_array_pybuffer.pyx
```diff
@@ -0,0 +1,59 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
rlratzel: Thanks for noticing that, no it isn't being used anywhere so I removed it in the latest commit.

However, we did manage to remove a device-to-device copy in PLC's `uniform_neighbor_sample` by returning a CuPy array "view" instead of the old approach of copying the data to a new owning CuPy array [here](https://github.com/rapidsai/cugraph/pull/2684). That resulted in a small speedup (as shown by the before-after benchmarks in that PR) and the elimination of a memory spike while making the copy. The `DeviceArrayPyBuffer` was an earlier attempt to do the same thing, but I never removed that class after switching to the CuPy array approach.

python/cugraph_service/cugraph_service_server/cugraph_handler.py
rlratzel: @VibhuJawa and I discussed this offline. We'll need the ability to have extension APIs return various numbers of arrays, dtypes, etc. which are user-defined, so we'll likely allow extension authors to specify meta-data about the arrays being returned and used to reconstruct the correct number of arrays and dtypes client-side.  This will be part of a future PR.

alexbarghi-nv (approved):
👍

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #2729, opened 2022-09-24
Update raft import

### LINKED ISSUE #2727: [BUG] Raft import changed
**Describe the bug**
The new `raft` conda packages recently deployed are a breaking change for cuGraph as it substitutes `raft` with `raft-dask` and `pylibraft`. This breaks the way `raft` is currently imported as it can be observed from the error traceback below
```
>>> import raft
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'raft'
```
The new way of importing raft is described as below
```
>>> import raft_dask
>>> 

```
This change affect couple `raft` utils currently used by cugraph like the `default_client`, `raftComms`, `get_raft_comm_state`, and `get_client`

jnke2016:
This PR updates the way `raft` is imported following changes on the `raft` new conda packages.

closes #2727

### DIFF
--- a/conda/environments/cugraph_dev_cuda11.5.yml
+++ b/conda/environments/cugraph_dev_cuda11.5.yml
@@ -1,4 +1,4 @@
-name: cugraph_dev
+name: raft
 channels:
 - rapidsai
 - nvidia
--- a/conda/environments/cugraph_dev_cuda11.5.yml
+++ b/conda/environments/cugraph_dev_cuda11.5.yml
@@ -1,4 +1,4 @@
-name: raft
+name: cugraph_dev
 channels:
 - rapidsai
 - nvidia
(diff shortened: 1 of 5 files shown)


rlratzel (approved):
Looks okay to me. I also added @cjnolet just to make sure this is the new intended import.

rlratzel:
rerun tests

reason: build timed out. Timeouts should be less common after #2707 is merged, but trying a rerun now anyway.

alexbarghi-nv:
rerun tests

Ethyling:
Removing ops-codeowners from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

BradReesWork (approved):

rlratzel:
@gpucibot merge

PR #2738, opened 2022-09-26
Removes unneeded test dependency on cugraph from pylibcugraph tests

### LINKED ISSUE #2737: pylibcugraph depends on cugraph for certain tests
[This](https://github.com/rapidsai/cugraph/blob/branch-22.10/python/pylibcugraph/pylibcugraph/tests/test_node2vec.py#L21) pylibcugraph test is importing a cugraph util module, which forces pylibcugraph test runs to have cugraph installed.  These are used for getting a path to a test file and for creating a cugraph Graph to test how renumbering is handled.

This is causing problems for cugraph wheel builds, since the pylibcugraph wheel testing should not require cugraph to be installed.

This dependency should either be reversed (cugraph should depend on pylibcugraph for those utils and they should be moved into pylibcugraph) or just removed or copied into pylibcugraph if they're small enough.

cc @vyasr

rlratzel:
closes #2737 

* Uses the same util from `pylibcugraph.testing` instead of `cugraph.testing`
* Removes a renumbering test that uses cugraph/cudf and relies on the same test that uses cupy

### DIFF
--- a/python/pylibcugraph/pylibcugraph/tests/test_node2vec.py
+++ b/python/pylibcugraph/pylibcugraph/tests/test_node2vec.py
@@ -18,8 +18,7 @@ from pylibcugraph import (ResourceHandle,
                           GraphProperties,
                           SGGraph,
                           node2vec)
-from cugraph.testing import utils
-import cugraph
+from pylibcugraph.testing import utils
 
 
 COMPRESSED = [False, True]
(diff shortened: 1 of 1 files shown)


vyasr:
Thanks for the quick turnaround on this!

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #2739, opened 2022-09-27
suppress expansion of unused raft spectral templates

### LINKED ISSUE #2678: spectral_clustering build is suspiciously slow and contributes to CI build timeouts

cjnolet:
Closes #2678 

raft updated some k-means logic and dramatically increased the compile time of our spectral clustering implementation.  This PR adds an include that will suppress the expansion of templates that we are not using.

### DIFF
--- a/python/cugraph/CMakeLists.txt
+++ b/python/cugraph/CMakeLists.txt
@@ -36,6 +36,18 @@ option(FIND_CUGRAPH_CPP "Search for existing CUGRAPH C++ installations before de
 
 # If the user requested it,  we attempt to find CUGRAPH.
 if(FIND_CUGRAPH_CPP)
+
+  include(rapids-cpm)
+  rapids_cpm_init()
+  include(rapids-cuda)
+  rapids_cuda_init_architectures(cugraph-python)
+  enable_language(CUDA)
+
+  # We need to call get_raft due to cuML asking for raft::nn and
+  # raft::distance targets
+  # see issue https://github.com/rapidsai/cuml/issues/4843
+  include(../cpp/cmake/thirdparty/get_raft.cmake)
+
   find_package(cugraph ${cugraph_version} REQUIRED)
 else()
   set(cugraph_FOUND OFF)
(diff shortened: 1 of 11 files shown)


BradReesWork:
added labels just so CI will run

ChuckHastings (approved):

raydouglass (approved):

rlratzel (approved):

ChuckHastings:
@gpucibot merge

python/cugraph/CMakeLists.txt
```diff
@@ -36,6 +36,22 @@ option(FIND_CUGRAPH_CPP "Search for existing CUGRAPH C++ installations before de
...
+  enable_language(CUDA)
+
+  set(CUGRAPH_VERSION_MAJOR "${cugraph-python_VERSION_MAJOR}")
+  set(CUGRAPH_VERSION_MINOR "${cugraph-python_VERSION_MINOR}")
+
+
+  # We need to call get_raft due to cuML asking for raft::nn and
+  # raft::distance targets
+  # see issue https://github.com/rapidsai/cuml/issues/4843
+  include(../../cpp/cmake/thirdparty/get_raft.cmake)
```
vyasr, line 53: @cjnolet good find. The underlying issue here will be fixed by https://github.com/rapidsai/rapids-cmake/pull/154.

vyasr:
For future reference the issue this PR fixes that caused #2707 to fail is that rapids-cmake's export commands don't support `COMPONENTS`, so the distance component requirement of libcugraph isn't being propagated to the pylibcugraph and cugraph packages. Once rapidsai/rapids-cmake#154 is merged we should be able to get rid of the extra `get_raft.cmake` includes.

Issue #1733, opened 2021-07-26
[ENH] Clearer `None` warnings during chained use


lmeyerov:
**Scenario motivating the issue**
When (incorrectly) attempting to use the API in a pipeline (`cugraph.Graph().add_edges(...).louvain()..`), I hit some hard to decipher error messages. These mostly came down to deep internal messages due to passing in `None`, which was getting passed in because imperative commands are non-chainable so they return `None` instead of `G`, and I missed that.

**Describe the solution you'd like**
Methods should have a `None` check, at least for the `G` position

**Describe alternatives you've considered**
I was also thinking about how to make the library more compositional
- The naive solution would be to make methods return `G`
- ... but that will be confusing. There is value in keeping side-effect operations clearly imperative via `None` returns, as otherwise folks may mutate data they think immutable
- Some libraries like pandas try to track this -- `G' = G.effect()` and mark `G'` as a view of a stateful obj, so `G'.effect()` can give a warning that it is unsafe
- While others go truly functional, which would change the side-effecting methods to functional... which is a deeper change

So I figured, at least for interim, going compositional will be hard. So instead, should make it easier to tell users that they shouldn't be going compositional. So, a `is None` check on the `G` position would go far.

cc @BradReesWork

BradReesWork:
@lmeyerov I am trying to come up with a good solution to the problem posed.  We would just echo back the graph, but as you stated, that might lead developers to believe that the return is different (mutated).  A `Graph` is a stateful python class that adheres closer to  object-oriented programming than functional.  The problem that you mention is that we do mix our models

lmeyerov:
Yeah -- I think long-term it's smart to go functional, and agree that near-term tricky to clearly do both. 

The immediate issue is a rocky experience for new + intermittent users for the current API, which is a much easier hill to climb :) There are probably 20-50 top-level `cugraph.some_method(G, ...)` calls, so an intermediate fix is:

1. runtime type contract check, at least for G position <- best cost/reward ratio?

```python
def some_method(G):
  assert_type(G, cugraph.Graph, 'Method expected parameter G of type cugraph.Graph but received %s', type(G))
  ...
```

or even

```python
@assert_G
def some_method(G):
  ...
```

2. `typing` annotations on the signature. Good for docs, and some IDE users, but probably not remote IDE ones <-- may already be there, and if not, seems best for py3.7+, which is a judgement call..

```python
def some_method(G: cugraph.Graph, ...) -> None`
```

[...]

PR #2745, opened 2022-09-27
added support for pipelining functions

### LINKED ISSUE #2735: Update Python code to support pipelining of functions
There are a number of functions that currently prevent function pipelining since the function returns None.   
Update functions to always return an object 

Note: [This](https://stackoverflow.com/questions/73826904/cugraph-create-nonetype/73830932?noredirect=1#comment130391020_73830932) is a related SO question

BradReesWork:
closes #2735 

related to SO:  https://stackoverflow.com/questions/73826904/cugraph-create-nonetype/73830932?noredirect=1#comment130391020_73830932

### DIFF
--- a/python/cugraph/cugraph/structure/graph_classes.py
+++ b/python/cugraph/cugraph/structure/graph_classes.py
@@ -169,6 +169,7 @@ class Graph:
             renumber=renumber,
             store_transposed=store_transposed,
             legacy_renum_only=legacy_renum_only)
+        return self
 
     def from_cudf_adjlist(self, offset_col, index_col, value_col=None):
         """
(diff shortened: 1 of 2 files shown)


acostadon (approved):
It responds well to the stack overflow request/question.
Just found the one comment typo.

rlratzel:
I think we should move this to 22.12 so we have time to discuss some of the concerns raised in [this issue](https://github.com/rapidsai/cugraph/issues/1733).  If this gets merged now and we realize we need to change the behavior to support views (as mentioned in [the issue](https://github.com/rapidsai/cugraph/issues/1733)), we may be forced to make breaking API changes.

BradReesWork:
Closing - fixing the "from" API introduces another set of issues.   We need to redesign our graph classes

PR #2756, opened 2022-09-29
Raise a warning for certain algorithms

### LINKED ISSUE #2742: Add warning in python cugraph when auto-transpose will happen
Add warning in python cugraph when auto-transpose will happen

Algos that require transposed, and will auto-transpose:
* pagerank
* katz
* hits
* eigenvector_centrality

What about the other algos that require non-transposed?  Will they handle and auto-un-transpose a transposed graph?

jnke2016:
Certain algorithms (Katz, HITS, PageRank, Eigenvector centrality) require the flag `store_transposed` to be set to `True` for optimal performance. Although the CAPI internally `transposed` the graph at the algo's call if it wasn't at the graph creation, this adds extra overheads. This PR raises an exception if the user doesn't set the flag to `True` at the graph creation

closes #2742


alexbarghi-nv:
Shouldn't `simpleDistributedGraph` also include the new `store_transposed` property?

jnke2016:
@alexbarghi-nv [store_transposed](https://github.com/rapidsai/cugraph/blob/branch-22.10/python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py#L50) was already available in `simpleDistributedGraph`

alexbarghi-nv (approved):
👍

rlratzel (approved):

rlratzel:
@gpucibot merge

PR #2776, opened 2022-10-05
Bug fix when -1 is used as a valid external vertex ID

### LINKED ISSUE #2383: [BUG] Invalid vertex ID
**Describe the bug**
We're currently using `std::numeric_limits<vertex_t>::max()` (if vertex_t is unsigned) or -1 (if vertex_t is signed) as the invalid vertex ID. This works for internal vertex IDs (after renumbering) but for external vertex IDs (before renumbering), there is non-zero chance that users may use max() or -1 as a valid vertex ID.

This is not very likely in practice but still possible. We need a mechanism to properly handle this.

seunghwak:
Close #2383 

Our code assumed that -1 (if vertex_t is signed) or std::numeric_limits<vertex_t>::max() (if vertex_t is unsigned) is invalid vertex ID. This assumption does not necessarily hold for user provided edge list before renumbering.

This ended up erroneous renumbering output when invalid vertex ID (e.g. -1) is used as a valid vertex ID in the user input edge list.

This PR fixes this bug.

If renumbering is disabled, negative values cannot be used as  valid vertex IDs. This PR adds a check for this as well (only when do_expensive_check is set).


ChuckHastings (commented):
Implementation looks good.

Could we add a unit test that validates this behavior?  Perhaps a simple load karate and "replace vertex id 1 with vertex id -1", or for all v: "v = -v;".

seunghwak:
> Perhaps a simple load karate and "replace vertex id 1 with vertex id -1", or for all v: "v = -v;".

This is technically possible but test code gets dirty (I did something similar to test this, but just deleted the code before committing).

I tried to create karate-negative-id.mtx and change the ID there, but our mtx reader somewhat expects that vertex ID either starts from either 0 or 1, and this doesn't look like a very reliable mechanism to test this.

I remember we discussed about switching to csv files in testing and I think this is a good excuse to write a simple csv reader.

I will write a simple csv reader and create a csv file with weird vertex IDs like -1, std::numeric_limits<vertex_t>::lowest(), and std::numeric_limits<vertex_t>::max(). And we can work on gradually phasing out mtx files.

ChuckHastings:
There is a mechanism to specify a custom edge list in a unit test.  I don't think we have any code that does that.  We could potentially explore that.

seunghwak:
@ChuckHastings Added a CSV reader (https://github.com/rapidsai/cugraph/pull/2791) to add a unit test. This PR should be merged after https://github.com/rapidsai/cugraph/pull/2791.

ChuckHastings (approved):

ChuckHastings:
rerun tests

jnke2016 (approved):

BradReesWork:
@gpucibot merge

PR #2783, opened 2022-10-06
remove graph broadcast and serialization object, not used

ChuckHastings:
Graph broadcast and `serialization_t` object are no longer used, eliminate these from the cugraph build/code base.

Marked as breaking, as it does delete things from the C++ API, but nothing uses them.


seunghwak (approved):
Just to double check, aren't we using this in multi-GPU batch betweenness centrality? (replicate a SG graph, then run SG BC with different seeds in each GPU)

ChuckHastings:
> Just to double check, aren't we using this in multi-GPU batch betweenness centrality? (replicate a SG graph, then run SG BC with different seeds in each GPU)

Legacy implementation of the batch feature does replication at the python level.  When we implement MG betweenness we can decide how to grapple with that, but this code is a bit more complex then we probably need (we have more tools now, we can do something much simpler).

ChuckHastings:
rerun tests

ChuckHastings:
@gpucibot merge

PR #2798, opened 2022-10-12
Code cleanup (remove dead code and move legacy files to the legacy directory)

seunghwak:
Code clean-up before bringing bigger changes to the codebase.

Breaking because several replaced legacy functions are deleted (I guess no one is directly using them).


ChuckHastings:
Some of these changes are probably premature.

Legacy jaccard/overlap is still in use by python.  We also don't have an implementation for the weighted variation.

seunghwak:
> Legacy jaccard/overlap is still in use by python.

I wonder why compile didn't fail in my local system. Aren't we calling these in cython/c_api?

seunghwak:
It seems like `extract_subgraph_vertex` is also used in python as well (isn't this induced subgraph?) But it seems like this wasn't called in cython/c_api. Do we have another mechanism to call C++ function from python (without relying on c_api or cython)?

ChuckHastings:
There was, before `cython.cu` a mechanism that we used for calling legacy algorithms.  Not everything was migrated to use `python.cu`, because we decided to create the C API/PLC layer and stopped migrating things to `python.cu`.

naimnv (approved):

rlratzel:
rerun tests

reason: test timeout, try rerunning for now but may have to look into it if it persists.

ChuckHastings (approved):

naimnv (approved):

jnke2016 (approved):

seunghwak:
rerun tests

PR #2804, opened 2022-10-13
Fix concat with different index dtypes in SG PropertyGraph

eriknw:
@alexbarghi-nv want to give this a try?  I previously made these changes to MG (was there an issue for it?), but not SG.

### DIFF
--- a/python/cugraph/cugraph/structure/property_graph.py
+++ b/python/cugraph/cugraph/structure/property_graph.py
@@ -1334,11 +1334,6 @@ class EXPERIMENTAL__PropertyGraph:
         if epd is not None:
             vert_sers.append(epd[self.src_col_name])
             vert_sers.append(epd[self.dst_col_name])
-        if len(vert_sers) > 1 and not all(
-            cudf.api.types.is_dtype_equal(vert_sers[0].index.dtype, s.index.dtype)
-            for s in vert_sers
-        ):
-            vert_sers = [s.reset_index(drop=True) for s in vert_sers]
         return vert_sers
 
     @staticmethod
--- a/python/cugraph/cugraph/structure/property_graph.py
+++ b/python/cugraph/cugraph/structure/property_graph.py
@@ -262,9 +262,13 @@ class EXPERIMENTAL__PropertyGraph:
             vert_sers = self.__get_all_vertices_series()
             if vert_sers:
                 if self.__series_type is cudf.Series:
-                    self.__num_vertices = cudf.concat(vert_sers).nunique()
+                    self.__num_vertices = cudf.concat(
+                        vert_sers, ignore_index=True
+                    ).nunique()
                 else:
-                    self.__num_vertices = pd.concat(vert_sers).nunique()
+                    self.__num_vertices = pd.concat(
+                        vert_sers, ignore_index=True
+                    ).nunique()
             return self.__num_vertices
 
         value_counts = self._vertex_type_value_counts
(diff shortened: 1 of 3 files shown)


alexbarghi-nv (approved):
This fixes the bug.  👍

VibhuJawa:
Sorry missed this, just raised a PR to fix the same. Lets merge this in please.

VibhuJawa (changes requested):
Looks good to me.  Can we just add a test like this: 

```python

def test_get_num_vertices():

    pg = PropertyGraph()
    node_df = cudf.DataFrame()
    node_df["node_id"] = cudf.Series([0, 1, 2]).astype("int32")
    pg.add_vertex_data(node_df, "node_id", type_name="_N")

    edge_df = cudf.DataFrame()
    edge_df["src"] = cudf.Series([0, 1, 2]).astype("int32")
    edge_df["dst"] = cudf.Series([0, 1, 2]).astype("int32")
    pg.add_edge_data(edge_df, ["src", "dst"], type_name="_E")

    assert pg.get_num_vertices() == 3
   ```

**Comment by eriknw** [comment]
Thanks for the test @VibhuJawa! Added.  Sorry you missed this PR and did duplicate work.

I confirmed your test breaks w/o this PR and passes with this PR.

**Review by VibhuJawa — APPROVED** [review]
LGTM to me

File: python/cugraph/cugraph/tests/test_property_graph.py [hunk 1]
```diff
@@ -1845,6 +1845,27 @@ def test_add_data_noncontiguous(df_type):
         )
 
 
+@pytest.mark.parametrize("df_type", df_types, ids=df_type_id)
+def test_get_num_vertices(df_type):
```
rlratzel, line 1849: Rather than add another test with a very similar name, can `test_num_vertices()` be updated to add this coverage? Otherwise, it's not clear how the intent of `test_get_num_vertices()` differs from `test_num_vertices()`.

(I like this new name better since it matches the actual API name, so if you combine them my vote would be to use the new name)

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -1326,6 +1334,11 @@ def __get_all_vertices_series(self):
         if epd is not None:
             vert_sers.append(epd[self.src_col_name])
             vert_sers.append(epd[self.dst_col_name])
+        if len(vert_sers) > 1 and not all(
```
rlratzel, line 1337: Can you also update the docstring for this method to mention that the index for the returned Series objects will be normalized/reset if the dtypes differ?

eriknw: This code has been deleted. The test allowed me to confirm that it's not needed (unlike for MG, where it is).  `concat` with `ignore_index=True` works as expected.

Also, I don't think the indices here are meaningful.

python/cugraph/cugraph/tests/test_property_graph.py
eriknw: Thanks for the careful review.  Good catch.  My bad.

I changed the name and added a description.  It tests a very specific case.

rlratzel (approved):

rlratzel:
rerun tests

reason: timeout as described in https://github.com/rapidsai/cugraph/issues/2810 - rerunning for now.

rlratzel:
@gpucibot merge

PR #2819, opened 2022-10-17
Fix MG tests bugs

### LINKED ISSUE #2815: [BUG] MG tests failing
**Describe the bug**
The MG `batch_betweenness_centrality` and `batch_edge_betweenness_centrality` tests are failing with `AttributeError: 'str' object has no attribute 'get_graph'` and several MG replication tests are failing with an assertion error

Run the above MG tests

**Expected behavior**
tests should pass for all configuration of GPUs

jnke2016:
A 22.10 PR leveraging the datasets API broke the MG `batch_betweenness_centrality` and `batch_edge_betweenness_centrality`. In fact, both tests are calling their non batch SG counterparts tests but with non datasets API objects leading to `AttributeError: 'str' object has no attribute 'get_graph'`. Another 22.10 PR which repartitioned  the dask dataframe also broke the above tests including the MG replication tests. In fact, those algos/function with `batch_enable=True` expect only one [partition](https://github.com/rapidsai/cugraph/blob/branch-22.12/python/cugraph/cugraph/dask/structure/replication.pyx#L38) which will later be replicated across all workers during the replication call. 

This PR fixes the above issues by not repartitioning the dask dataframe when `batch_enable=True` and also leverages the datasets API for MG `batch_betweenness_centrality` and `batch_edge_betweenness_centrality`

closes #2815


jnke2016:
rerun tests

rlratzel (approved):

BradReesWork:
@gpucibot merge

Issue #2668, opened 2022-09-07
[DOC] Update and improve cuGraph apidocs to include current components


acostadon:
Add in cugraph _service and make the top level more of an overview to each component with links to more details.

Add in links to the C++ Docs

Add a section at the bottom linking to the C api page

BradReesWork:
succeeded and replaced by #2877

PR #2827, opened 2022-10-19
Pyg docs

acostadon:
updating/adding docstrings to graph_store in gnn/pyg code

addresses some content required for #2668


alexbarghi-nv:
These updates have now all been folded into other PRs.

PR #2829, opened 2022-10-19
[WIP][skip-ci] packed bool specialization to store edge endpoint|edge properties

### LINKED ISSUE #3472: Implement efficient mechanism to store Boolean (bit) property values
https://github.com/rapidsai/cugraph/issues/3446 Outlined an approach for implementing k-truss that depends on an implementation of edge masking.

Create an efficient data structure and corresponding utility methods to implement an edge bit mask.

seunghwak:
Pack 32 bools in uint32_t (1 bit per bool) instead of using 1 byte per bool (uint8_t).

This can benefit the current BFS implementation and also can be used to implement edge masking in the future.

closes #3472 

Remaining works:

update_edge_src|dst_property for packed bool.
Need to decide how to handle when bool is included in thrust::tuple.


seunghwak:
Replaced by https://github.com/rapidsai/cugraph/pull/3482

PR #2834, opened 2022-10-20
Avoid directly calling graph constructor (as code cleanup before edge property support in primitives)

seunghwak:
Delete dead code and avoid directly calling graph_t constructor (use create_graph_from_edgelist instead) as a pre-requsite for removing edge weights from graph_t class and edge property support in graph primitives.

`bool renumber` is added to `coarsen_graph` to support MG Louvain testing (instead of replicating all the code related graph coarsening to support graph coarsening without renumbering).

This PR also removed GRAPH_TEST as we don't expect users to directly call graph_t constructors.

Breaking as `bool renumber` is added to the public function `coarsen_graph`


cpp/include/cugraph/graph.hpp
```diff
@@ -556,7 +533,3 @@ __host__ __device__ std::enable_if_t<std::is_unsigned<vertex_t>::value, bool> is
 }  // namespace cugraph
 
 #include "eidecl_graph.hpp"
-
```
ChuckHastings, line 559: Bottom of the documentation group

cpp/include/cugraph/graph.hpp
```diff
@@ -483,25 +479,6 @@ class graph_t<vertex_t, edge_t, weight_t, store_transposed, multi_gpu, std::enab
  private:
   friend class cugraph::serializer::serializer_t;
```
ChuckHastings, line 457: Could we also get rid of `serializer_t`?  Seems like with `graph_bcast.hpp` gone this class is unused.

cpp/include/cugraph/graph.hpp
```diff
@@ -28,10 +28,6 @@
 #include <type_traits>
 #include <vector>
 
-/** @defgroup cpp_api cuGraph C++ API
```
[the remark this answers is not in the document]
seunghwak: Done.

cpp/include/cugraph/graph.hpp
seunghwak: Done.

cpp/include/cugraph/graph.hpp
seunghwak: Done.

ChuckHastings (approved):

PR #2832, opened 2022-10-20
Cugraph-Service Remote Graphs and Algorithm Dispatch

alexbarghi-nv:
Resolves https://github.com/rapidsai/graph_dl/issues/80

This PR combines `RemoteGraph` and `RemotePropertyGraph`, adds support for remote API calls to cugraph.Graph, adds subgraph extraction to the property graph wrappers, adds the ability to dispatch cuGraph algorithms (though currently only `uniform_neighbor_sample` is supported), better supports multigraphs for subgraph extraction, properly handles renumbering on the server, and supports implicit subgraph extraction for remote graphs.

PR #2850, opened 2022-10-26
Adds arbitrary server extension support to cugraph-service

### LINKED ISSUE #2822: Allow cugraph_service extensions to support more than graph creation
cugraph_service extensions currently only support graph creation, and make an assumption that they'll always return graph objects which can be referenced as graph IDs by clients.  However, other operations can benefit from executing completely server-side, and extensions should allow for these use cases.

Graph creation extensions don't need to change, but additional extension types should be added which don't assume graph objects are returned.  Other considerations:
* extensions which return data to the client (instead of just IDs)
* GPU-GPU data xfer

cc @VibhuJawa

rlratzel:
closes #2822 

This PR adds the ability for users to add extensions that accept and return arbitrary values that run server-side, in addition to the graph creation extensions already supported.

* Adds the new `load_extensions()`, and `call_extension()` APIs
* Replaces `unload_graph_creation_extensions()` with the more general `unload_extension_module()`
* Adds additional tests for the new extension types
* Added the ability to load extensions using a python-style module path, useful for loading server-side extensions that are importable but the path on disk is not known
* Added tests for extensions that access the server facade object, to serve as examples for extension writers
* Also added result_device_id option to new `call_extension()`, and benchmark
* Also updates dask init call to use defaults and removes problematic hardcoded transports

image
_`device=None` is a host RPC data transfer, `device=0` is a GPU-GPU data transfer from the server on device 1 to the client on device 0_

TODO:
* <s> Update new `call_extension()` API to allow caller to specify a device to directly transfer results to</s>
* <s> Add test/example of an extension that accesses a server-side graph</s>


rlratzel:
rerun tests

reason: gpuCI issue related to conda channel priority in the build image, which may have been resolved by now.

alexbarghi-nv:
Is this the preferred way of calling cuGraph algorithms, i.e. `uniform_neighbor_sample`, and if so, do we want to pre-load cuGraph algorithm extensions by default, or go through another route?

rlratzel:
> Is this the preferred way of calling cuGraph algorithms, i.e. `uniform_neighbor_sample`, and if so, do we want to pre-load cuGraph algorithm extensions by default, or go through another route?

I don't think this is the preferred way, but instead a way for clients to just add custom functionality that runs on the server (plugins). I suppose we can use this to easily prototype new functionality too (eg. @VibhuJawa will be using it to add APIs to support heterogeneous graphs more efficiently in the short-term), but for now I'm thinking we'd still want clients to call  `client.uniform_neighbor_sample`, etc. for the "standard" APIs. 

> Basically I'm looking ahead to what #2832 is trying to do, and see if I can use this extension support mechanism to dispatch any cuGraph algorithm.

You do bring up a good point, and the server could eventually be nothing more than a mechanism to call server-side functions from files on disk which could just make up the entire client API, and the client API could just be hidden behind the Remote classes and algo dispatcher.

alexbarghi-nv:
Ok, this looks good then, I just want to have one more look-over before I give the final thumbs up.

alexbarghi-nv (approved):
👍

BradReesWork:
rerun tests

VibhuJawa (approved):
LGTM to me

BradReesWork (approved):

BradReesWork:
rerun tests

BradReesWork:
@gpucibot merge

PR #2856, opened 2022-10-27
CAPI create graph from CSR

### LINKED ISSUE #2508: C API - Add Graph creation from CSR matrices
Some users of cugraph might already have a CSR structure when they need to create a graph.

Add a C API that accepts an indices and offsets array to create a graph.

ChuckHastings:
Add capability to C API to create an SG graph from an existing CSR.

Dependent on PR #2854 

Closes #2508


ChuckHastings:
rerun tests

cpp/include/cugraph_c/graph.h
```diff
@@ -76,6 +76,46 @@ cugraph_error_code_t cugraph_sg_graph_create(
...
+ *
+ * @param [in]  handle      Handle for accessing resources
+ * @param [in]  properties  Properties of the constructed graph
+ * @param [in]  offsets     Device array containing the CSR offsets array
+ * @param [in]  indices     Device array containing the destination vertex ids
+ * @param [in]  weights     Device array containing the edge weights.  Note that an unweighted
+ *                          graph can be created by passing weights == NULL.
+ * @param [in]  edge_ids    Device array containing the edge ids for each edge.  Optional
+                            argument that can be NULL if edge ids are not used.
+ * @param [in]  edge_types  Device array containing the edge types for each edge.  Optional
```
rlratzel, line 90: I wonder if it would be better if we started a naming convention that emphasizes these are arbitrary user-defined types, and not something to be confused with an `edge_t` used in so many places elsewhere.  I know most devs will see this and realize it can't really be an array of `edge_t` values, but it still might help with overall readability of our APIs IMO.

Maybe `edge_type_ids` instead of `edge_types`, with the "id" part to emphasize these are basically user-defined numerical enum values.

Related: this seems like it could have an impact on `edge_type_t` too?

cpp/src/c_api/graph_sg.cpp
```diff
@@ -190,6 +191,167 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+            bool multi_gpu>
+  void operator()()
+  {
+    if constexpr (multi_gpu || !cugraph::is_candidate<vertex_t, edge_t, weight_t>::value) {
+      unsupported();
+    } else {
+      // using edge_type_t = int32_t;
+
+      if (check_) {
+        // FIXME:  Need an implementation here.
```
seunghwak, line 243: Yes :-)

seunghwak (approved):

cpp/include/cugraph_c/graph.h
ChuckHastings: Updated in latest push

jnke2016 (approved):

naimnv (approved):

rlratzel (approved):

ChuckHastings:
@gpucibot merge

PR #2864, opened 2022-10-31
Define and implement C/C++ for MNMG Egonet

### LINKED ISSUE #2906: MNMG Egonet - Update C++ API definition for Egonet
Update C++ API definition for Egonoet to support MNMG

### LINKED ISSUE #2907: MNMG Egonet - Define C API for Egonet
Define C API for Egonet

### LINKED ISSUE #2908: MNMG Egonet - Define C unit tests
Define C unit tests for MNMG Egonet

### LINKED ISSUE #2909: MNMG Egonet - Define C++ unit tests
Define C++ unit tests for MNMG Egonet.

### LINKED ISSUE #2910: MNMG Egonet - Implement C++
Current implementation of egonet already uses the new graph structure, need to refresh some things in the code and provide the MNMG implementation (currently the code raises an exception if an MG graph is passed).

### LINKED ISSUE #2911: MNMG Egonet - Implement C API
Implement C API for Egonet

ChuckHastings:
Define and implement C API for egonet.  This will help us eliminate dependence on `cython.cu`.

Closes #2906 
Closes #2907
Closes #2908
Closes #2909
Closes #2910
Closes #2911


ChuckHastings:
rerun tests

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1380,6 +1384,37 @@ extract_ego(raft::handle_t const& handle,
             vertex_t n_subgraphs,
             vertex_t radius);
 
+/**
+ * @brief returns induced EgoNet subgraph(s) of neighbors centered at nodes in source_vertex within
+ * a given radius.
```
seunghwak, line 1389: Isn't this identical to k-hop neighbors?

seunghwak: Umm... maybe not... I assume identical only when k=1.

cpp/src/community/egonet_impl.cuh
```diff
@@ -92,67 +90,74 @@ extract(raft::handle_t const& handle,
...
+
+    cugraph::bfs<vertex_t, edge_t, weight_t, multi_gpu>(multi_gpu ? handle : light_handle,
+                                                        graph_view,
+                                                        reached[i].data(),
+                                                        nullptr,
+                                                        source_vertex.data() + i,
+                                                        1,
+                                                        direction_optimizing,
+                                                        radius,
+                                                        do_expensive_check);
```
seunghwak, line 120: Maybe just add a FIXME and address later, but this looks very inefficient (but not less efficient than what we had... so better kick out cython.cu first).

Sequentially running one BFS (with maximum distance of radius) per source vertex can largely under-utilize a GPU (or multiple GPUs in multi-GPU runs).

We may better either run multi-source BFS, or run 1-hop neighbors recursively (radius times).

ChuckHastings: There is a FIXME on line 93 (about 20 lines higher) observing that this is likely slow.  Is that sufficient to address this concern?

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Essentially the ego net for a vertex v1 and a radius of k is the following:
* compute the distance of every vertex in the graph from v1
* induce subgraph for the subset of vertices that are of distance <= k

So for radius = 1, that will include v1, all of its neighbors, the edges to its neighbors and the edges between its neighbors.
For radius = 2, that will include v1, all of its neighbors, all of its 2-hop neighbors, the edges to the 1-hop neighbors, the edges between the 1-hop neighbors, the edges from the 1-hop neighbors to the 2-hop neighbors and the edges between the 2-hop neighbors.
etc.

seunghwak: Gotcha, yeah... so this is basically finding [0, radius]-hop neighbors then running induced subgraph, so this can be a bit different from just running radius-hop neighbors then running induced subgraph.

cpp/src/community/egonet_impl.cuh
seunghwak: I saw that FIXME but wasn't sure about the `until the frontiers are large enough to use this approach` part. But we can defer this discussion to the future.

The reason I wasn't sure about the FIXME comment.

I think we'd better not use `this approach` (i.e. running single source BFS in a sequential loop). Running 1-hop neighbors recursively sounds like a simpler approach with optimal parallelism but we may revisit this in the future when we actually update this algorithm.

seunghwak (approved):

jnke2016 (approved):

ChuckHastings:
@gpucibot merge

PR #2868, opened 2022-11-01
Use new rapids-cmake functionality for rpath handling.

vyasr:
The new functionality creates a single source of truth defining the location at which a library will be, allowing rapids-cmake to handle the necessary relative paths in the RPATH and avoiding the current error-prone approach that requires each CMakeLists.txt to correctly assign the relative path.


BradReesWork (approved):

BradReesWork:
rerun tests

BradReesWork:
@gpucibot merge

PR #2895, opened 2022-11-07
Correct build failures when doing a local build

robertmaynard:
I found two errors when building locally that this PR fixes.

- The cugraph test util target relies on gtest headers but doesn't use the `GTest::gtest` target. This works in CI due to the conda env placing the GTest headers in a directory found by other `-I`
- The `cpp/tests/centrality/betweenness_centrality_validate.cu` test uses the wrong `nearly_equal` function which fails due to calling a host only function from a device/host function.


ChuckHastings:
@gpucibot merge

PR #2896, opened 2022-11-08
[REVIEW] Bring up cugraph_dgl_repo

VibhuJawa:
This PR resolves https://github.com/rapidsai/graph_dl/issues/92

- [x] Add code
- [x] Add tests
- [x] Add readme


alexbarghi-nv (approved):
👍

python/cugraph-dgl/cugraph_dgl/cugraph_storage.py
```diff
@@ -0,0 +1,857 @@
...
+    explanation about its usage.
+
+    """
+
+    def __init__(
+        self,
+        num_nodes_dict: dict,
+        single_gpu: bool = True,
+        cugraph_service_client=None,
+        device_id=0,
```
rlratzel, line 44: Throughout this file too are cases of some args with type hints, some without. Should we be enforcing in reviews that these be present on new code, or would this be a FIXME type of situation for now?

python/cugraph-dgl/cugraph_dgl/utils/test_utils.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
```
rlratzel, line 1: Are these utils used for the tests for this package, or are they intended to be used by your user's tests (or both)?  I'm just asking since if these are strictly for the tests for this package, then it might be better to move this to the `cugraph-dgl/tests` folder and rename it just `utils.py`.  One minor issue with this is that `pytest` may try to collect tests from here since it matches the standard file name for tests.

rlratzel, line 1: Also, same question about type hints, since if this module is meant to be user-facing then it should have type hints (if we're committing to those in this PR).

python/cugraph-dgl/setup.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.
```
rlratzel, line 1: I've made some important changes to this file for building the conda package [here](https://github.com/rapidsai/cugraph/blob/8bfcbd5f21742049056bf5fc298d613db170568b/python/cugraph-dgl/setup.py). I'm open to suggestion on how we should coordinate this wrt who merges first.

python/cugraph-dgl/tests/test_cugraph_storage.py
```diff
@@ -0,0 +1,143 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import pytest
+
+try:
+    import cugraph_dgl
+except ImportError:
```
rlratzel, line 18: For this and other places where this pattern is used in test files: I usually catch `ModuleNotFoundError`, since `ImportError` can also be raised if the importable is present but had another problem preventing it from being imported (missing symbol, SyntaxError, etc.), and catching `ImportError` will mask those errors we should probably know about.

rlratzel (changes requested):
I didn't see any showstoppers, but I did have several questions and minor requests.  I like that the test coverage seems good and that they are easy to follow.

python/cugraph-dgl/cugraph_dgl/cugraph_storage.py
```diff
@@ -0,0 +1,857 @@
...
+        See also ``cugraph_dgl.cugraph_storage_from_heterograph``
+        to convert from DGLHeteroGraph to CuGraphStorage
+
+        Parameters
+        ----------
+         single_gpu: bool
+            Wether to create the cugraph Property Graph
+            on a single GPU or multiple GPUs
+            single GPU = True creates PropertyGraph
+            single GPU = False creates MGPropertyGraph
```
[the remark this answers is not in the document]
VibhuJawa: Removed the MGPropertyGraph link here

python/cugraph-dgl/cugraph_dgl/convert.py
```diff
@@ -0,0 +1,34 @@
...
+import dgl
+from cugraph_dgl import CuGraphStorage
+from cugraph_dgl.utils.cugraph_conversion_utils import (
+    add_nodes_from_dgl_HeteroGraph,
+    add_edges_from_dgl_HeteroGraph,
+)
+
+
+def cugraph_storage_from_heterograph(
+    g: dgl.DGLHeteroGraph, single_gpu=True
```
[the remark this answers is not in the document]
VibhuJawa: Added type hint for single_gpu

python/cugraph-dgl/cugraph_dgl/utils/test_utils.py
VibhuJawa: Moved to `tests/utils.py`

VibhuJawa: This is not a user facing module so not adding type-hints here . Will add in a followup PR. Will file an issue to track it

https://github.com/rapidsai/cugraph/issues/2922

python/cugraph-dgl/tests/test_cugraph_storage.py
VibhuJawa: Changed to `ModuleNotFoundError` 
 https://github.com/rapidsai/cugraph/pull/2896/commits/aa394725f2ccab68257dfad8f8001c3e05d7a8b5

python/cugraph-dgl/cugraph_dgl/cugraph_storage.py
VibhuJawa: I have added all the type-hints that are possible without adding extra imports.  The missing ones are now `dgl objects` and `dataframe` objects that depends on the underlying library (for dgl bakcend, it can be pytorch, tensorflow,keras)  and for `dataframe` can be (`cuDF,dask_cuDF, pandas`).

VibhuJawa: Have added issue: https://github.com/rapidsai/cugraph/issues/2922 to track it.

python/cugraph-dgl/setup.py
VibhuJawa: Lets merge this PR in as it includes the most changes and has the things we all need.

rlratzel (approved):
LGTM, thanks.

sevagh (approved):

raydouglass (approved):

rlratzel:
@gpucibot merge

PR #2898, opened 2022-11-09
Fix bug in how is_symmetric is set when transposing storage

ChuckHastings:
The `tranpose_storage` function was incorrectly setting graph properties.

### DIFF
--- a/cpp/src/structure/transpose_graph_storage_impl.cuh
+++ b/cpp/src/structure/transpose_graph_storage_impl.cuh
@@ -62,7 +62,7 @@ transpose_graph_storage_impl(
   }
 
   auto is_multigraph = graph.is_multigraph();
-  auto is_symmetric = graph.is_symmetric();
+  auto is_symmetric  = graph.is_symmetric();
 
   // FIXME: if is_symmetric is true we can do this more efficiently,
   //        since the graph contents should be exactly the same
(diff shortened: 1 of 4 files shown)


cpp/src/structure/transpose_graph_storage_impl.cuh
```diff
@@ -123,6 +127,10 @@ transpose_graph_storage_impl(
   auto number_of_vertices = graph.number_of_vertices();
   auto is_multigraph      = graph.is_multigraph();
   bool renumber           = renumber_map.has_value();
+  auto is_symmetric       = graph.is_symmetric();
+
+  // FIXME: if is_symmetric is true we can do this more efficiently,
+  //        since the graph contents should be exactly the same
```
seunghwak, line 133: We may handle this inside the C++ function. We can check the flag and if symmetric, it might be possible to just `move` all the graph contents.

seunghwak: And why aren't you directly calling `transpose_graph_storage`? (https://github.com/rapidsai/cugraph/blob/branch-22.12/cpp/src/structure/transpose_graph_storage_impl.cuh#L168)

seunghwak: Oops :-( Sorry... I thought this is the C API function...

seunghwak (approved):

PR #2920, opened 2022-11-11
key, value store abstraction

seunghwak:
* create an abstract class to store (key, value) pairs and support searching with keys.
  - The abstract class can either use binary search or hash table (NVIDIA cuCollection) for search.
  - If value type is not directly supported by cuCollection (e.g. if not bitwise comparable or trivially copyable), we store key, value_index pairs in a cuCollection object and use the retrieved value_index to find a value.
* Update cugraph C++ code and per_v_transform_reduce_dst_key_aggregated_outgoing_e to use this abstraction (the latter is necessary to support (key, value) pairs with thrust::tuple type values; this is necessary for Leiden).


seunghwak:
@naimnv Please review this, and now the `per_v_transform_reduce_dst_key_aggregated_outoing_e` primitive is updated to support thrust::tuple type value.

naimnv (approved):

ChuckHastings (approved):

ChuckHastings:
@gpucibot merge

PR #2932, opened 2022-11-15
CuGraph-Service Asyncio Fix

### LINKED ISSUE #2875: [BUG]: cugraph_service client APIs that specify `result_device` cannot be used from Jupyter notebooks
### Version

22.12

_No response_

### Describe the bug.

When resut_device is specified, is calls an async function to start an event loop for the UCX-Py listener. Since Jupyter starts an event loop, calling a client API that uses a result_device results in a traceback (see below).

_No response_

### Relevant log output

```shell
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In [12], line 1
----> 1 gs.client.call_extension(func_name="add_node_data_from_parquet_remote",
      2     file_path='nt.a.parquet',
      3     node_col_name='node_id',
      4     node_offset=0,
      5     ntype='nt.a',
      6     graph_id=gs.gdata._graph_id,
      7     result_device=0
      8                          )

File ~/dgl/cugraph/python/cugraph_service/examples/../cugraph_service_client/client.py:109, in CugraphServiceClient.__server_connection.<locals>.wrapped_method(self, *args, **kwargs)
    107 self.open()
    108 try:
--> 109     ret_val = method(self, *args, **kwargs)
    110 finally:
    111     if not self.hold_open:

File ~/dgl/cugraph/python/cugraph_service/examples/../cugraph_service_client/client.py:437, in CugraphServiceClient.call_extension(self, func_name, result_device, *func_args, **func_kwargs)
    435 func_kwargs_repr = repr(func_kwargs)
    436 if result_device is not None:
--> 437     result_obj = asyncio.run(
    438         self.__call_extension_to_device(
    439             func_name, func_args_repr, func_kwargs_repr, result_device
    440         )
    441     )
    442     # result_obj is a cupy array or tuple of cupy arrays on result_device
    443     return result_obj

File /datasets/vjawa/miniconda3/envs/cugraph_dgl_22_12/lib/python3.9/asyncio/runners.py:33, in run(main, debug)
      9 """Execute the coroutine and return the result.
     10 
     11 This function runs the passed coroutine, taking care of
   (...)
     30     asyncio.run(main())
     31 """
     32 if events._get_running_loop() is not None:
---> 33     raise RuntimeError(
     34         "asyncio.run() cannot be called from a running event loop")
     36 if not coroutines.iscoroutine(main):
     37     raise ValueError("a coroutine was expected, got {!r}".format(main))

RuntimeError: asyncio.run() cannot be called from a running event loop
```

### Environment details

_No response_

### Other/Misc.

_No response_

### Code of Conduct

alexbarghi-nv:
Resolves #2875 

This resolves a bug where the CGS client tried to start a new event loop without checking whether it was in an existing event loop.  Running a new event loop on the same thread as an existing loop is not allowed, and this caused CGS to fail on Jupyter when using the CGS `result_device` option.

This PR resolves the bug by creating a new thread when necessary so that CGS can run its own event loop when one is already running in the current thread.


alexbarghi-nv:
rerun tests

python/cugraph-service/client/cugraph_service_client/client.py
```diff
@@ -32,6 +32,32 @@
...
+        self.args = args
+        self.kwargs = kwargs
+        self.result = None
+        super().__init__()
+
+    def run(self):
+        self.result = asyncio.run(self.func(*self.args, **self.kwargs))
+
+
+def run_async(func, *args, **kwargs):
```
rlratzel, line 47: This is nice. Can you also add a simple docstring to this function though (and possibly one for the `RunAsyncioThread` class too for completeness)?

python/cugraph-service/tests/test_e2e.py
```diff
@@ -515,6 +515,29 @@ def test_uniform_neighbor_sampling(client_with_edgelist_csv_loaded):
     )
 
 
+def test_inside_asyncio_event_loop(client_with_edgelist_csv_loaded):
```
rlratzel, line 518: Can you parameterize this test so `result_device` is tested with both a device ID and `None`?  That will ensure the same code paths work from an `async` caller both when the client `async` functions are used and not used. 
Also I've typically put tests for device data transfer in `test_mg_e2e.py` to ensure transfers happen between different GPUs as opposed to from GPU 0 to GPU 0 (although either should still exercise the same code paths, I think).

rlratzel (commented):
Looks good, I just had a couple of requests.

python/cugraph-service/client/cugraph_service_client/client.py
```diff
@@ -32,6 +32,32 @@
...
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        loop = None
+    if loop and loop.is_running():
```
rlratzel, line 52: I'm curious if the check for `loop.is_running()` is redundant here, since `asyncio.get_running_loop()` is supposed to return the running event loop.

BradReesWork (approved):

VibhuJawa (approved):
LGTM. Thanks for the clean implementation @alexbarghi-nv

BradReesWork:
@gpucibot merge

PR #2949, opened 2022-11-17
Remove legacy renumbering

### LINKED ISSUE #1679: [BUG] Intermittent Hangs on MNMG
There is an intermittent hang on Prom which goes away after rerunning. Here is a [link](https://gist.github.com/ayushdg/e2fb1cd8611bf5b17769f42bf2fbfa18) to the error which can be located in the log file. There is no clear way to reproduce the error if not running the benchmarks couple times. That error can be found in combination with an OOM (maximum pool size exceeded) if you run BFS (unweighted dataset) with 6 Nodes at scale 30.

### LINKED ISSUE #2437: [BUG] Fix exposed `legacy_renum_only` options in public graph APIs
A change (still to be merged at the time of this writing) exposes a `legacy_renum_only` which was previously an internal flag, but is now exposed in order to create graphs specifically for use with algos that use pylibcugraph Graph objects stored in the cuGraph graph objects.

This is confusing to users not familiar with the implementation details, and could cause problems if users don't understand when they should or shouldn't set it.

For reference: https://github.com/rapidsai/cugraph/pull/2394#discussion_r927139725

<hr>

Behavior based on `legacy_renum_only` setting vs. SG/MG/algo type

<table>
<tr><td></td><td><b>SG legacy algo</b></td><td><b>MG legacy algo</b></td><td><b>SG plc algo</b></td><td><b>MG plc algo</b></td></tr>
<tr><td><b>legacy_renum_only=True</b></td><td><b>ERROR</b></td><td><b>ERROR</b></td><td>optimal performance</td><td>optimal performance</td></tr>
<tr><td><b>legacy_renum_only=False</b></td><td>no error</td><td>no error</td><td>no error but reduced performance (redundant renumber step)</td><td>no error but reduced performance (redundant renumber step), and <b>potential hang</b></td></tr>
</table>

### LINKED ISSUE #2535: MNMG Induce Subgraph - Python
Modify the python induce subgraph calls to use the new PLC version of induce subgraph. Add MNMG support.

API definition in https://github.com/rapidsai/cugraph/issues/2529 blocks starting this work
Implementation in https://github.com/rapidsai/cugraph/issues/2532 blocks completing this work

### LINKED ISSUE #2536: MNMG Induce Subgraph - Python unit tests
Implement Induce Subgraph unit tests for python.

### LINKED ISSUE #2537: MNMG Induce Subgraph - Remove old Python
Remove old python implementation of induce subgraph.

jnke2016:
closes #2437 
closes #1679
closes #2535 
closes #2536 
closes #2537


BradReesWork (approved):

BradReesWork:
rerun tests

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -36,11 +36,11 @@ def dask_client():
         enable_tcp_over_ucx=True,
         enable_nvlink=True,
         enable_infiniband=True,
-        enable_rdmacm=True,
+        enable_rdmacm=False,
         # net_devices="mlx5_0:1",
```
VibhuJawa, line 40: I suggest setting everything to `None`:
I think this not aligned with the current conftest

https://github.com/rapidsai/cugraph/blob/6c89b6359f853d9558d72c47d98e16615e3b329a/python/cugraph/cugraph/testing/mg_utils.py#L26-L28


You will have to make align it here.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -195,9 +201,7 @@ def __from_edgelist(
             )
             source = renumber_map.renumbered_src_col_name
             destination = renumber_map.renumbered_dst_col_name
-            # Use renumber_map to figure out if renumbering was skipped or not
-            # This was added to handle 'legacy_renum_only' which will skip the
-            # old C++ renumbering  when running the pylibcugraph/C algos
+            # Use renumber_map to figure out if the python renumbering occured
             self.properties.renumbered = renumber_map.implementation.numbered
```
VibhuJawa, line 205: I seem to be getting below error  on this, can you recheck

```
stributedGraphImpl.__from_edgelist(self, input_ddf, source, destination, edge_attr, renumber, store_transposed, legacy_renum_only)
    232 # If renumbering is not enabled, this function will only create
    233 # the edgelist_df and not do any renumbering.
    234 # C++ renumbering is enabled by default for algorithms that
    235 # support it (but only called if renumbering is on)
    236 self.compute_renumber_edge_list(
    237     transposed=store_transposed, legacy_renum_only=legacy_renum_only
    238 )
--> 240 self.properties.renumbered = self.renumber_map.implementation.numbered
    241 ddf = self.edgelist.edgelist_df
    243 num_edges = len(ddf)

AttributeError: 'NoneType' object has no attribute 'implementation'
```

python/cugraph/cugraph/structure/number_map.py
```diff
@@ -550,8 +516,8 @@ def renumber_and_segment(
```
alexbarghi-nv, line 516: Can we get a comment here explaining that renumbering only occurs here if the column names are a list?

python/cugraph/cugraph/structure/number_map.py
```diff
@@ -481,7 +454,13 @@ def renumber_and_segment(
         store_transposed=False,
         legacy_renum_only=False,
     ):
```
alexbarghi-nv, line 456: I think it would be a good time to add a docstring for this function that explains exactly what it does and why.

python/cugraph/cugraph/tests/mg/test_mg_graph.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
```
alexbarghi-nv, line 1: Can you add an additional test to this file (and sg as well) to ensure renumbering works as expected when edge ids, edge types, and weights are passed in as the edge attributes?

alexbarghi-nv:
Shouldn't the renumber wrapper get thrown out in this PR?
(python/cugraph/cugraph/structure/renumber_wrapper.pyx)

alexbarghi-nv:
This PR appears to break personalized pagerank tests.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
rlratzel: @jnke2016 it might be better to add a proper method or bool property to NumberMap, like `is_renumbered` instead of trying to access the internal objects.

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -282,13 +282,8 @@ def from_dask_cudf_edgelist(
             If True, stores the transpose of the adjacency matrix.  Required
             for certain algorithms.
 
-        legacy_renum_only : bool, optional (default=False)
```
rlratzel, line 285: You probably shouldn't remove the description if the arg is still available to be used.  You should instead mention that it's deprecated and/or explain how it's supposed to be used.  If it's being deprecated here, then we should also be raising a warning.

Also, shouldn't we be updating `from_cudf_edgelist()` too?

rlratzel (changes requested):

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -36,11 +36,11 @@ def dask_client():
...
     )
 
     yield client
 
-    stop_dask_client(client)
+    stop_dask_client(*client)
```
eriknw, line 45: Why `*client` here? Perhaps `stop_dask_client(client, cluster)` instead?

seunghwak:
Are we using shuffle.py (https://github.com/rapidsai/cugraph/blob/branch-23.04/python/cugraph/cugraph/structure/shuffle.py) in anyplace? Can't we delete this file if no longer relevant?

ChuckHastings (approved):

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
jnke2016: I believe I resolved this issue per @rlratzel recommendation but feel free to test it again and let me know @VibhuJawa

python/cugraph/cugraph/tests/conftest.py
jnke2016: I believe this was resolved when I fetched the latest changes

python/cugraph/cugraph/structure/number_map.py
jnke2016: Sure. I added it.

python/cugraph/cugraph/structure/number_map.py
jnke2016: Done.

python/cugraph/cugraph/structure/graph_classes.py
jnke2016: Right. I believe this was resolved.

python/cugraph/cugraph/tests/conftest.py
jnke2016: Yep. resolved in a previous PR

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
VibhuJawa: I think your tests should catch if the error would have happened.

alexbarghi-nv (approved):
👍

jnke2016:
rerun tests

python/cugraph/cugraph/community/subgraph_extraction.py
```diff
@@ -56,33 +57,12 @@ def subgraph(G, vertices):
 
     """
 
-    G, isNx = ensure_cugraph_obj_for_nx(G)
-    directed = G.is_directed()
+    warning_msg = (
+        "This call is deprecated and will be refactored "
+        "in the next release. Please call 'cugraph.induced_subgraph()' instead"
+    )
+    warnings.warn(warning_msg, PendingDeprecationWarning)
```
rlratzel, line 64: Should use `DeprecationWarning` since it's now being officially deprecated.
```suggestion
    warnings.warn(warning_msg, DeprecationWarning)
```

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -153,6 +153,8 @@ def from_cudf_edgelist(
             pylibcugraph algorithms.  Must be false for algorithms
             not yet converted to the pylibcugraph C API.
 
+            This parameter is deprecated and will be removed.
```
rlratzel, line 156: Can you also raise a `DeprecationWarning` if the user sets `legacy_renum_only` to `True`?

rlratzel (changes requested):

VibhuJawa (approved):
LGTM

python/cugraph/cugraph/structure/graph_classes.py
jnke2016: So this function eventually calls `self._Impl._simpleGraphImpl__from_edgelist` which intrinsically raises the `DeprecationWarning`. I didn't add an extra one to avoid duplicates

rlratzel (approved):

raydouglass (approved):

PR #2951, opened 2022-11-17
Provide option to keep original vertex/edge IDs when renumbering

### LINKED ISSUE #2713: Provide an option for the PropertyGraph renumber API to preserve the original vertex/edge IDs
After https://github.com/rapidsai/cugraph/issues/2627 is closed, a user calling the renumber API will alter the property graph and no longer be able to retrieve the original vertex/edge IDs.

@eriknw proposed an enhancement to the renumber API to take an optional column name to be used to store the original IDs. This would solve the problem in a way that gives the user the option to consume more memory to store the originals only if they need it.

eriknw:
Closes #2713


alexbarghi-nv (approved):
👍

BradReesWork (approved):

BradReesWork:
@gpucibot merge

PR #2964, opened 2022-11-19
Add wheel builds

### LINKED ISSUE #3004: Add `ALLOW_CLONE_CUGRAPH_OPS` option to `CMakeLists.txt`
Based on the discussion [here](https://github.com/rapidsai/cugraph/pull/2964#discussion_r1034189329), we need to add an option to `CMakeLists.txt` to prevent the default behavior of running `cmake` to be to attempt to clone a closed-source repo.

vyasr:
This PR adds wheel building to cugraph.

Also resolves #3004.


vyasr:
rerun tests

vyasr:
rerun tests

jolorunyomi (approved):

BradReesWork (approved):

BradReesWork:
rerun tests

cpp/CMakeLists.txt
```diff
@@ -79,10 +79,21 @@ option(BUILD_CUGRAPH_MG_TESTS "Build cuGraph multigpu algorithm tests" OFF)
 option(CMAKE_CUDA_LINEINFO "Enable the -lineinfo option for nvcc (useful for cuda-memcheck / profiler" OFF)
 option(BUILD_TESTS "Configure CMake to build tests" ON)
 option(USE_CUGRAPH_OPS "Enable all functions that call cugraph-ops" ON)
+option(USE_RAFT_STATIC "Build raft as a static library" OFF)
+option(CUDA_STATIC_RUNTIME "Statically link the CUDA toolkit runtime and libraries" OFF)
+option(CUGRAPH_USE_CUGRAPH_OPS_STATIC "Build and statically link the cugraph-ops library" OFF)
+option(CUGRAPH_EXCLUDE_CUGRAPH_OPS_FROM_ALL "Exclude cugraph-ops targets from cuGraph's 'all' target" OFF)
```
rlratzel, line 85: How does this compare to the `USE_CUGRAPH_OPS` option above? Do we need both?

cpp/cmake/thirdparty/get_libcugraphops.cmake
```diff
@@ -16,20 +16,63 @@
...
 ###
-find_and_configure_cugraph_ops(VERSION ${CUGRAPH_MIN_VERSION_cugraph_ops})
+find_and_configure_cugraph_ops(VERSION      ${CUGRAPH_MIN_VERSION_cugraph_ops}
+                               REPO         ${CUGRAPH_CUGRAPH_OPS_REPOSITORY}
+                               PINNED_TAG   ${CUGRAPH_CUGRAPH_OPS_BRANCH}
+                               BUILD_STATIC     ${CUGRAPH_USE_CUGRAPH_OPS_STATIC}
+                               EXCLUDE_FROM_ALL ${CUGRAPH_EXCLUDE_CUGRAPH_OPS_FROM_ALL}
+                               # When PINNED_TAG above doesn't match cugraph,
+                               # force local cugraph-ops clone in build directory
+                               # even if it's already installed.
```
rlratzel, line 77: How does this affect non-RAPIDS devs that may not have access to the cugraph-ops (closed) source?  Rather than attempting a clone by default, should it output an informative error instead?  I haven't tried this, but I'm concerned that this would output a message about invalid credentials and the developer wouldn't know why or what happened.

rlratzel (commented):
Thank you very much for this, these changes look good and also clean up some tech debt.  I did have a couple of questions though.

BradReesWork:
rerun tests

cpp/CMakeLists.txt
vyasr: The first option controls whether we build against cugraph-ops at all. This option just controls whether we install compiled cugraph-ops libs, which we don't want to do in wheels (mostly for size reasons).

cpp/cmake/thirdparty/get_libcugraphops.cmake
vyasr: That is true. Previously we were just doing `rapids_find_package` here so there was no risk of that. One option would be to add a separation option like `ALLOW_CLONE_CUGRAPH_OPS` to control whether or not we use a `rapids_find_package` call instead of `rapids_cpm_find`. Do you think that's worth doing (and if so, do you think it's worth doing in this PR, or saved for a follow-up)?

cpp/CMakeLists.txt
rlratzel: Ah, I didn't notice the `CUGRAPH_USE_CUGRAPH_OPS_STATIC` option, which brings the total number of cugraph-ops-related options to 3. To be clear, I was asking about the `CUGRAPH_EXCLUDE_CUGRAPH_OPS_FROM_ALL` option vs. the older `USE_CUGRAPH_OPS` option we added separately a few weeks ago ([here](https://github.com/rlratzel/cugraph/blob/872fd1c2a1d2b09ecd8d55ffefe9331c66e03be2/python/cugraph/CMakeLists.txt#L36)).

It sounds like all 3 do different things, so to answer my own question, I think we need them all.

cpp/cmake/thirdparty/get_libcugraphops.cmake
rlratzel: I think an `ALLOW_CLONE_CUGRAPH_OPS` which defaults to `OFF` would be good to prevent unexpected clone attempts of the closed source repo, and possibly go a long way in preventing issues filed from community members building from source.  It's up to you if you think it's better in this PR or a separate one, since I image we want this merged soon.  I can file an issue for this if it's going in a separate PR.

cpp/CMakeLists.txt
vyasr: Yup I was answering the same one. To make sure we're on the same page:
- `USE_CUGRAPH_OPS`: Whether or not we build with cugraph-ops support. If we turn this off, some functionality in cugraph won't work.
- `CUGRAPH_USE_CUGRAPH_OPS_STATIC`: Whether to compile cugraph-ops as a static lib. We want this for wheels so that we can just compile all the necessary pieces directly into libcugraph rather than needing to ship around multiple libraries. In the more general case static linking also helps us avoid including parts of the statically linked library that are unused by the dependent library, but in this case I assume that concern is irrelevant since cugraph is the only consumer of cugraph-ops so it probably uses all of it.
- `CUGRAPH_EXCLUDE_CUGRAPH_OPS_FROM_ALL`: Turning this off prevents installation of libcugraphops.[so|a] into the install tree. We want this for wheels because once we have statically linked against it the Python package doesn't need it anymore. Since we're not making libcugraph.so publicly available to consumers (it's only there for pylibcugraph/cugraph to use) there's no reason to provide the underlying cugraph-ops library either, and removing this saves space in the wheel.

vyasr:
rerun tests

BradReesWork:
rerun tests

cpp/CMakeLists.txt
rlratzel: excellent, thanks for the explanations, I may include these in a future comment to the `CMakeLists.txt` file.

cpp/cmake/thirdparty/get_libcugraphops.cmake
rlratzel: https://github.com/rapidsai/cugraph/issues/3004

rlratzel (approved):

cpp/cmake/thirdparty/get_libcugraphops.cmake
vyasr: Since I had to fix merge conflicts and wait for one more round of CI I went ahead and implemented this flag in this PR.

BradReesWork:
@gpucibot merge

PR #2980, opened 2022-11-25
SG tested Leiden

naimnv:
### Leiden algorithm 
This PR adds Leiden implementation using cugraph primitives.  It also add an implementation of `maximal independent set ` using cugraph primitives that is used by the Leiden implementation. 

- It reuses code from Louvain implementation to find locally optimal moves for vertices. 
- Locally optimal movement creates  clusters of `Louvain communities` 
- In refinement phase, which is the main feature of Leiden algorithm, vertices are moved only within a `Louvain community` to subdivide a `Louvain community` in multiple `sub communities` or `Leiden communities`
- In graph contraction phase, vertices belonging to a `Leiden community` are merged to become a node in aggregated graph.

This code has been tested to work on SG. We aim to test MG version in next release.
Future improvement:  

- Move vertices randomly instead of greedily refinement phase
- Test for MNMG


cpp/src/community/detail/common_methods.cuh
```diff
@@ -183,10 +187,67 @@ weight_t compute_modularity(
...
+    thrust::transform(handle.get_thrust_policy(),
+                      keys_to_lookup.begin(),
+                      keys_to_lookup.end(),
+                      values_for_sought_keys.begin(),
+                      [d_cluster_weights = map_values.data(),
+                       d_cluster_keys    = map_keys.data(),
+                       num_clusters      = map_keys.size()] __device__(map_key_t cluster) {
+                        auto pos = thrust::lower_bound(
+                          thrust::seq, d_cluster_keys, d_cluster_keys + num_clusters, cluster);
+                        return d_cluster_weights[pos - d_cluster_keys];
```
seunghwak, line 239: assert(pos != d_cluster_keys + num_clusters);  <= this is executed only in the debug mode, this can happen if `keys_to_look_up` has keys missing in `map_keys`.
thrust::distance(d_cluster_keys, pos);

cpp/src/community/detail/common_methods.cuh
```diff
@@ -183,10 +187,67 @@ weight_t compute_modularity(
...
+      std::is_floating_point<map_value_t>::value ? std::numeric_limits<map_value_t>::max()
+                                                 : invalid_vertex_id<map_key_t>::value,
+      handle.get_stream());
+    values_for_sought_keys = cugraph::collect_values_for_keys(handle.get_comms(),
+                                                              cluster_key_weight_map.view(),
+                                                              keys_to_lookup.begin(),
+                                                              keys_to_lookup.end(),
+                                                              vertex_to_gpu_id_op,
+                                                              handle.get_stream());
+  } else {
```
seunghwak, line 223: You can use `cluster_key_weight_map.view().find(keys_to_look_up.begin(), keys_to_look_up.end(), values_for_sought_keys.begin(), handle.get_sream());` to minimize code bifurcation between MG and SG paths.

cpp/src/community/detail/common_methods.cuh
```diff
@@ -183,10 +187,67 @@ weight_t compute_modularity(
...
+  if constexpr (multi_gpu) {
+    cugraph::detail::compute_gpu_id_from_ext_vertex_t<map_key_t> vertex_to_gpu_id_op{
+      handle.get_comms().get_size()};
+
+    kv_store_t<map_key_t, map_value_t, false> cluster_key_weight_map(
+      map_keys.begin(),
+      map_keys.end(),
+      map_values.data(),
+      invalid_vertex_id<map_key_t>::value,
+      std::is_floating_point<map_value_t>::value ? std::numeric_limits<map_value_t>::max()
```
seunghwak, line 214: `std::is_floating_point<map_value_t>::value`=>`std::is_floating_point_v<map_value_t>` to be little more concise.

cpp/src/community/detail/common_methods.cuh
```diff
@@ -183,10 +187,67 @@ weight_t compute_modularity(
...
+                                                 : invalid_vertex_id<map_key_t>::value,
```
seunghwak, line 215: `invalid_vertex_id<map_key_t>::value` is `map_key_t` type and will this going to work say if `map_key_t` is `int32_t` and `map_value_t` is int16_t?

cpp/src/community/detail/common_methods.cuh
```diff
@@ -183,10 +187,67 @@ weight_t compute_modularity(
...
+template <typename map_key_t, typename map_value_t, bool multi_gpu>
+rmm::device_uvector<map_value_t>&& lookup_primitive_values_for_keys(
+  raft::handle_t const& handle,
+  rmm::device_uvector<map_key_t>& map_keys,
+  rmm::device_uvector<map_value_t>& map_values,
+  rmm::device_uvector<map_key_t>& keys_to_lookup)
+{
+  static_assert(std::is_integral<map_key_t>::value);
+  static_assert(std::is_same<map_key_t, std::int32_t>::value ||
+                std::is_same<map_key_t, std::int64_t>::value);
```
seunghwak, line 199: I assume `static_assert(std::is_integral<map_key_t>::value);` is redundant here.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,729 @@
...
+  raft::handle_t const& handle,
+  cugraph::graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  std::optional<cugraph::edge_property_view_t<edge_t, weight_t const*>> edge_weight_view)
+{
+  using GraphViewType = cugraph::graph_view_t<vertex_t, edge_t, false, multi_gpu>;
+  using FlagType      = vertex_t;
+
+  vertex_t number_of_vertices = graph_view.local_vertex_partition_range_size();
+
+  rmm::device_uvector<vertex_t> mis(number_of_vertices, handle.get_stream());
```
seunghwak, line 84: `mis` is the return value and this variable is not used till very close to the end of this function. Defining this variable at the beginning wastes memory (you don't need to reserve this memory till the very end of this function) and also makes code more cryptic. A general rule is to minimize the scope of a variable. This makes code easier to understand and also avoid wasting memory (keep the peak memory usage low).

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,729 @@
...
+    thrust::copy_if(handle.get_thrust_policy(),
+                    vertex_begin,
+                    vertex_end,
+                    thrust::make_zip_iterator(thrust::make_tuple(
+                      mis_inclusion_flags.begin(), discard_flags.begin(), out_degrees.begin())),
+                    remaining_candidates.begin(),
+                    [] __device__(auto flags) {
+                      return (thrust::get<0>(flags) == 0) && (thrust::get<1>(flags) == 0) &&
+                             (thrust::get<2>(flags) > 0);
+                    });
```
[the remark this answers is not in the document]
seunghwak: And I don't think this code will work as you intended. What's the size of `mis_inclusion_flags`? `remaining_vertices.size()` or `graph_view.local_vertex_partition_range()`? It's the latter if I am not mistaken.

I guess the code below is what you intended.

```
auto last = thrust::remove_if(
  handle.get_thrust_policy(), 
  remaining_vertices.begin(),
  reamining_vertices.end(),
  [mis_inclusion_flags = raft::device_span(...),
   discard_flags = raft::device_span(...),
   out_degrees =  raft::device_span(...)]__device__(auto v) {
    return mis_inclusion_flags[v] == 0 && discard_flags[v] == 0 && out_degrees[v] > 0;
  });
remaining_vertices.resize(thrust::distance(remaining_vertices.begin(), last), handle.get_stream());
```

And you can just remove 0 vertices with 0 out-degrees when you initialize `remaining_vertices` so no need to check out degrees again & again.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,729 @@
...
+  mis.resize(nr_vertices_in_mis, handle.get_stream());
+
+  // if (debug) {
+  cudaDeviceSynchronize();
+  std::cout << "Found mis of size " << mis.size() << std::endl;
+  cudaDeviceSynchronize();
+  // raft::print_device_vector("mis", mis.data(), mis.size(), std::cout);
+  // }
+
+  return mis;
```
seunghwak, line 726: And I think this code is overly complicated with unnecessarily many temporary state variables.

Can't you just do the following?
```
1. compute out degrees.
2. initialize remaining_vertices to all non-zero out degree vertices.
3. initialize ranks to out-weight-sums.
4. while (remaining_vertices.size() > 0)
4-1. Pick 50% of remaining_vertices as candidates.
4-2. Call per_v_transform_reduce_incoming_e and per_v_transform_reduce_outgoing_e to find the maximum rank neighbor.
If the maximum neighbor rank is std::numeric_limits<weight_t>::max(), remove from `remaining_vertices`.
If the maximum neighbor rank is lower than the vertex's own rank, remove from `remaining_vertices` and add to `mis`. Also, set the rank to std::numeric_limits<<weight_t>::max().
5. return `mis`. 
```

seunghwak: And you may create a temporary copy of ranks and set ranks to std::numeric_limits<weight_t>::lowest() if it's not already std::numeric_limits<weight_t>::max() (included in the MIS) and if the vertex is not a candidate.

Alternatively, you may maintain a flag vector indicating whether a vertex is a candidate or not.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,729 @@
...
+  auto vertex_begin =
+    thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first());
+  auto vertex_end = thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last());
+
+  rmm::device_uvector<vertex_t> included_list(graph_view.local_vertex_partition_range_size(),
+                                              handle.get_stream());
+  rmm::device_uvector<vertex_t> discard_list(graph_view.local_vertex_partition_range_size(),
+                                             handle.get_stream());
+
+  // Out degree must of 1 to be included in MIS
```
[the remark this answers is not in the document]
naimnv: Decision graph, which is a directed graph, is constructed from (src, dst, modularity gain) tuples and  each vertex can have at most one outgoing edge.

cpp/src/community/detail/mis_impl.cuh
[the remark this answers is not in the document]
naimnv: Size of `mis_inclusion_flags` is `graph_view.local_vertex_partition_range()`

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,422 @@
...
+      remaining_vertices.end(),
+      [temporary_ranks =
+         raft::device_span<weight_t>(temporary_ranks.data(), temporary_ranks.size()),
+       v_first = graph_view.local_vertex_partition_range_first()] __device__(auto v) {
+        //
+        // if rank of a non-candidate vertex is not +Inf (i.e. the vertex
+        // is not already in MIS), set it to -Inf
+        //
+        auto v_offset = v - v_first;
+        if (temporary_ranks[v_offset] < (std::numeric_limits<weight_t>::max() - EPSILON)) {
```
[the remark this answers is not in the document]
seunghwak: Yes, and look for std::numeric_limits<weight_t>::infinity();

I guess we can use

`constexpr weight_t max_weight = std::numeric_limits<weight_t>::has_infinity() ? std::numeric_limits<weight_t>::infinity() : std::numeric_limits<weight_t>::max()`;
`constexpr weight_t lowest_weight = std::numeric_limits<weight_t>::has_infinity() ? std::numeric_limits<weight_t>::infinity() * -1.0 : std::numeric_limits<weight_t>::lowest();`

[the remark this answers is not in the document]
seunghwak: Need to double check whether this work on device (I assume this will work, I've seen inf. in running device code, but need to double check).

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,422 @@
...
+      multi_gpu ? dst_rank_cache.view()
+                : detail::edge_minor_property_view_t<vertex_t, weight_t const*>(
+                    temporary_ranks.data(), vertex_t{0}),
+      edge_dummy_property_t{}.view(),
+      [] __device__(auto src, auto dst, auto src_rank, auto dst_rank, auto wt) {
+        return thrust::make_tuple(src_rank, src);
+      },
+      thrust::make_tuple(std::numeric_limits<weight_t>::lowest(),
+                         invalid_vertex_id<vertex_t>::value),
+      cugraph::reduce_op::maximum<thrust::tuple<weight_t, vertex_t>>{},
```
seunghwak, line 258: Note that `per_v_tansform_reduce_incoming|outgoing_e` has a bug for min/max reduction. This is fixed in https://github.com/rapidsai/cugraph/pull/3284.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,422 @@
...
+    //
+    // Select a random set of candidate vertices
+    //
+    thrust::default_random_engine g;
+    g.seed(0);
+    thrust::shuffle(
+      handle.get_thrust_policy(), remaining_vertices.begin(), remaining_vertices.end(), g);
+
+    vertex_t nr_candidates =
+      std::max(vertex_t{1}, static_cast<vertex_t>(0.60 * remaining_vertices.size()));
```
seunghwak, line 145: Shouldn't this be `std::min`?

cpp/src/community/detail/mis_impl.cuh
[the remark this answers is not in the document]
seunghwak: "Also, encodings to represent infinity and not-a-number (NaN) data are reserved. The IEEE 754 Standard [[2]](https://docs.nvidia.com/cuda/floating-point/index.html#references__2) describes floating point encodings in full."
https://docs.nvidia.com/cuda/floating-point/

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,707 @@
...
+    //
+    thrust::default_random_engine g;
+    g.seed(0);
+    thrust::shuffle(
+      handle.get_thrust_policy(), remaining_vertices.begin(), remaining_vertices.end(), g);
+
+    vertex_t nr_candidates = std::max(
+      vertex_t{1},
+      std::min(static_cast<vertex_t>((0.50 + 0.25 * loop_counter) * remaining_vertices.size()),
+               vertex_t{remaining_vertices.size()}));
```
seunghwak, line 144: And just a heads-up. In multi-GPU, this logic may no longer work. While the total number of remaining vertices might still be larger than 0, the number of local remaining vertices can be 0 in some GPUs. In such a case, `nr_candidates` based on this logic (due to std::max(vertex_t{1}, ...) can overflow `remaining_vertices`.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,793 @@
...
+  thrust::for_each(
+    handle.get_thrust_policy(),
+    vertex_begin,
+    vertex_end,
+    [out_degrees = raft::device_span<edge_t const>(out_degrees.data(), out_degrees.size()),
+     ranks       = raft::device_span<vertex_t>(ranks.data(), ranks.size()),
+     v_first     = graph_view.local_vertex_partition_range_first()] __device__(auto v) {
+      auto v_offset = v - v_first;
+
+      if (out_degrees[v_offset] == 0) { ranks[v_offset] = invalid_vertex_id<vertex_t>::value; }
```
seunghwak, line 128: This won't work if vertex_t is unsigned. Note that invalid_vertex_id = -1 if signed and max() if unsigned.

seunghwak: And you can use thrust::replace_if.

https://thrust.github.io/doc/group__replacing_ga9953891234efd44b09942f7d6381ff79.html#ga9953891234efd44b09942f7d6381ff79

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,793 @@
...
+                       // if rank of a non-candidate vertex is not +Inf (i.e. the vertex
+                       // is not already in MIS), set it to -Inf
+                       //
+                       auto v_offset = v - v_first;
+                       if (temporary_ranks[v_offset] < std::numeric_limits<vertex_t>::max()) {
+                         temporary_ranks[v_offset] = invalid_vertex_id<vertex_t>::value;
+                         if (debug)
+                           printf("Setting rank of %d to %d\n", v, temporary_ranks[v_offset]);
+                       }
+                     });
```
seunghwak, line 181: And you can use thrust::replace_if.

https://thrust.github.io/doc/group__replacing_ga9953891234efd44b09942f7d6381ff79.html#ga9953891234efd44b09942f7d6381ff79

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,793 @@
...
+          if (debug) { printf("---> to discard %d\n", v); }
+
+          // Maximum rank neighbor is alreay in MIS
+          // Discard current vertex by setting (global) rank to -Inf
+          return true;
+        }
+
+        if (rank_of_v > max_neighbor_rank) {
+          if (debug) { printf("---> to include %d\n", v); }
+          // Mark it included by setting (global) rank to +Inf
```
seunghwak, line 392: Where did you set rank to +Inf?

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,793 @@
...
+
+    std::cout << "check out degree of candidates\n";
+
+    thrust::for_each(handle.get_thrust_policy(),
+                     remaining_vertices.end() - nr_candidates,
+                     remaining_vertices.end(),
+                     [out_degrees = out_degrees.data(),
+                      v_first     = graph_view.local_vertex_partition_range_size()] __device__(auto v) {
+                       if (out_degrees[v - v_first] <= 0) {
+                         printf("?v=%d, out_degrees[%d] = \n", v, v, out_degrees[v - v_first]);
```
seunghwak, line 407: This requires two integers but you're providing 3. It is very likely that compiler is complaining about this. Please fix warnings as soon as you see them and don't wait till code cleanup.

cpp/src/community/detail/common_methods.cuh
naimnv: Removed the function and used kv find stead

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -0,0 +1,319 @@
...
+                      max_outgoing_ranks.begin(),
+                      thrust::maximum<vertex_t>());
+
+    max_incoming_ranks.resize(0, handle.get_stream());
+    max_incoming_ranks.shrink_to_fit(handle.get_stream());
+
+    //
+    // Count number of candidates to be removed
+    //
+    vertex_t nr_candidates_to_remove = thrust::count_if(
```
seunghwak, line 224: Where are you using `nr_candidates_to_remove`?

seunghwak (commented):
Review part 1.

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+#include <prims/transform_reduce_e_by_src_dst_key.cuh>
+#include <prims/update_edge_src_dst_property.cuh>
+#include <utilities/collect_comm.cuh>
+
+#include <cugraph/detail/shuffle_wrappers.hpp>
+#include <cugraph/detail/utility_wrappers.hpp>
+#include <cugraph/graph_functions.hpp>
+
+#include <algorithm>
+#include <cmath>
```
seunghwak, line 36: include for std c++ headers should appear after thrust headers.

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+
+namespace cugraph {
+namespace detail {
+
+// a workaround for cudaErrorInvalidDeviceFunction error when device lambda is used
+template <typename vertex_t, typename weight_t>
+struct reduce_op_t {
+  using type                          = thrust::tuple<vertex_t, weight_t>;
+  static constexpr bool pure_function = true;  // this can be called from any process
+  // inline static type const identity_element =
```
seunghwak, line 62: delete dead code.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -715,9 +715,26 @@ template <typename vertex_t, typename edge_t, typename weight_t>
 std::pair<size_t, weight_t> leiden(raft::handle_t const& handle,
                                    legacy::GraphCSRView<vertex_t, edge_t, weight_t> const& graph,
                                    vertex_t* clustering,
-                                   size_t max_iter     = 100,
+                                   size_t max_level    = 100,
```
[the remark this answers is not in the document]
naimnv: If it's not much work then 23.06 maybe?

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+
+  __device__ auto operator()(thrust::tuple<vertex_t, weight_t> p0,
+                             thrust::tuple<vertex_t, weight_t> p1) const
+  {
+    auto id0 = thrust::get<0>(p0);
+    auto id1 = thrust::get<0>(p1);
+    auto wt0 = thrust::get<1>(p0);
+    auto wt1 = thrust::get<1>(p1);
+
+    return (wt0 < wt1) ? p1 : ((wt0 > wt1) ? p0 : ((id0 < id1) ? p0 : p1));
```
seunghwak, line 74: Can't we use thrust::tuple<weight_t, vertex_t>? Then, we can just use a thrust's default compare functor instead of writing our own custom functor.

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+  }
+};
+
+// a workaround for cudaErrorInvalidDeviceFunction error when device lambda is used
```
seunghwak, line 78: Just FYI (not a request to change), but we may need to check this is still a case. This was necessary in the past, but with a more recent CUDA compiler, I haven't seen this for a while, so this may no longer be necessary.

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+           std::pair<rmm::device_uvector<typename GraphViewType::vertex_type>,
+                     rmm::device_uvector<typename GraphViewType::vertex_type>>>
+refine_clustering(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  std::optional<edge_property_view_t<typename GraphViewType::edge_type, weight_t const*>>
+    edge_weight_view,
+  weight_t total_edge_weight,
+  weight_t resolution,
+  rmm::device_uvector<weight_t> const& vertex_weights_v,  // ?
```
seunghwak, line 138: Why `// ?`?

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+  rmm::device_uvector<typename GraphViewType::vertex_type>&& cluster_keys_v,
+  rmm::device_uvector<weight_t>&& cluster_weights_v,
+  rmm::device_uvector<typename GraphViewType::vertex_type>&& louvain_assignment,
+  edge_src_property_t<GraphViewType, weight_t> const& src_vertex_weights_cache,  //?
```
seunghwak, line 142: Why `// ?`?

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+                      auto wdeg           = thrust::get<1>(wcut_wdeg_and_louvain_volume);
+                      auto louvain_volume = thrust::get<2>(wcut_wdeg_and_louvain_volume);
+                      return wcut > (gamma * wdeg * (louvain_volume - wdeg));
+                    });
+
+  // Update cluster weight, weighted degree and cut for edge sources
+
+  edge_src_property_t<GraphViewType, weight_t> src_louvain_cluster_weight_cache(handle);
+  edge_src_property_t<GraphViewType, thrust::tuple<weight_t, weight_t>>
+    src_wdeg_and_cut_to_Louvain_cache(handle);
```
seunghwak, line 238: And any reason to use uppercase "L" in "Louvain" here? (while using lowercase "l" in other places).

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+  kv_store_t<vertex_t, vertex_t, false> leiden_to_louvain_map(leiden_assignment.begin(),
+                                                              leiden_assignment.end(),
+                                                              louvain_assignment.begin(),
+                                                              invalid_vertex_id<vertex_t>::value,
+                                                              invalid_vertex_id<vertex_t>::value,
+                                                              handle.get_stream());
+
+  auto vertex_begin =
+    thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first());
+  auto vertex_end = thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last());
```
seunghwak, line 291: It is generally more advisable to minimize the scope of variable and avoid generating unnecessary temporaries unless there is a justification for doing this. This variable is declared here and used in Line 502, any reason to do this?

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+    if constexpr (GraphViewType::is_multi_gpu) {
+      update_edge_src_property(
+        handle, graph_view, leiden_assignment.begin(), src_leiden_assignment_cache);
+
+      update_edge_dst_property(
+        handle, graph_view, leiden_assignment.begin(), dst_leiden_assignment_cache);
+
+      update_edge_src_property(
+        handle, graph_view, singleton_and_connected_flags.begin(), src_active_flag_cache);
+    }
```
seunghwak, line 329: This to if statements can be combined to one.

cpp/src/community/detail/refine.cuh
```diff
@@ -0,0 +1,763 @@
...
+    rmm::device_uvector<vertex_t> leiden_keys_used_in_edge_reduction(0, handle.get_stream());
+    rmm::device_uvector<weight_t> refined_community_volumes(0, handle.get_stream());
+    rmm::device_uvector<weight_t> refined_community_cuts(0, handle.get_stream());
+
+    //
+    // For each refined community, compute its volume
+    // (i.e.sum of weighted degree of all vertices inside it, ||Cr||) and
+    // and cut between itself and its Louvain community (E(Cr, S-Cr))
+    //
+    // TODO: Can we update ||Cr|| and E(Cr, S-Cr) instead of recomputing?
```
seunghwak, line 356: Our convention is to use FIXME instead of TODO. We search for FIXME to find/track FIXMEs.

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+  rmm::device_uvector<vertex_t> louvain_of_refined_partition(0, handle.get_stream());     // #V
+
+  //
+  // Edge source/destination cache
+  //
+  edge_src_property_t<graph_view_t, weight_t> src_vertex_weights_cache(handle);
+  edge_src_property_t<graph_view_t, vertex_t> src_louvain_assignment_cache(handle);
+  edge_dst_property_t<graph_view_t, vertex_t> dst_louvain_assignment_cache(handle);
+
+// FIXME: delete, temporary code for debugging
```
seunghwak, line 102: Please delete.

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+    // if (current_graph_view.number_of_vertices() == prev_nr_of_vertices) { break; }
+
+    //
+    //  Initialize every cluster to reference each vertex to itself
+    //
+    dendrogram->add_level(current_graph_view.local_vertex_partition_range_first(),
+                          current_graph_view.local_vertex_partition_range_size(),
+                          handle.get_stream());
+
+// FIXME: delete, temporary code for debugging
```
seunghwak, line 122: Please delete.

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+
+#ifdef TIMING
+    detail::timer_stop<graph_view_t::is_multi_gpu>(handle, hr_timer);
+#endif
+
+    if (no_movement) { break; }
+    if (cur_Q <= best_modularity) { break; }
+    best_modularity = cur_Q;
+
+#if 1
```
seunghwak, line 399: Delete `#if 1` and matching endif.

cpp/tests/community/new_leiden_test.cpp
seunghwak, line 1: Better place old Leiden test to the legacy directory and rename this to `leiden_test`

cpp/tests/community/new_leiden_test.cpp
```diff
@@ -0,0 +1,410 @@
...
+    std::cout << "level:" << level << std::endl;
+
+    float compare_modularity = static_cast<float>(modularity);
+
+    if (check_correctness) {
+      ASSERT_FLOAT_EQ(compare_modularity, expected_modularity);
+      ASSERT_EQ(level, expected_level);
+    }
+  }
+};
```
seunghwak, line 154: This is called before the above `run_current_test` function, right? Then, this should appear before the `run_current_test` function.

cpp/tests/community/new_leiden_test.cpp
```diff
@@ -0,0 +1,410 @@
...
+    raft::handle_t handle{};
+
+    rmm::device_uvector<vertex_t> clustering_v(num_vertices, handle.get_stream());
+    size_t level;
+    weight_t modularity;
+
+    std::tie(level, modularity) = cugraph::leiden(
+      handle, graph_view, edge_weight_view, clustering_v.data(), max_level, resolution);
+
+    RAFT_CUDA_TRY(cudaDeviceSynchronize());  // for consistent performance measurement
```
seunghwak, line 143: What is this? This shouldn't be necessary for consistent performance measurement. And if this function is returning (level, modularity), these host variables should be valid right after this function returns.

cpp/tests/community/new_leiden_test.cpp
```diff
@@ -0,0 +1,410 @@
...
+    std::cout << "modularity: " << modularity << std::endl;
+    std::cout << "level:" << level << std::endl;
```
seunghwak, line 145: Better avoid print outs in the performance measured part.

cpp/src/community/detail/common_methods.cuh
```diff
@@ -212,7 +218,11 @@ graph_contraction(raft::handle_t const& handle,
     labels.size(),
     false);
```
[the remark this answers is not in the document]
naimnv: This only does graph contraction and return the renumber map (which are Leiden assignment of aggregate graph). Will make  `labels` read only in next commit.

cpp/src/community/detail/common_methods.cuh
```diff
@@ -212,7 +218,11 @@ graph_contraction(raft::handle_t const& handle,
...
 
-  return std::make_tuple(std::move(new_graph), std::move(new_edge_weights));
+  numbering_indices.resize(0, handle.get_stream());
+  numbering_indices.shrink_to_fit(handle.get_stream());
```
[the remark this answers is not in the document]
naimnv: Will fix in next commit.

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+#ifdef TIMING
+    detail::timer_start<graph_view_t::is_multi_gpu>(
+      handle, hr_timer, "compute_vertex_and_cluster_weights");
+#endif
+
+    vertex_weights = compute_out_weight_sums(handle, current_graph_view, *edge_weight_view);
+    cluster_keys.resize(vertex_weights.size(), handle.get_stream());
+    cluster_weights.resize(vertex_weights.size(), handle.get_stream());
+
+    if (first_iteration) {
```
seunghwak, line 153: Can't we replace this with `if (dendrogram->num_levels() == 1)`?

seunghwak: Tracking many variables with large scopes increases cognitive load.

cpp/src/community/detail/mis_impl.cuh
naimnv: Was part of debug code, will remove in next commit

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+        std::tie(tmp_keys_buffer, tmp_weights_buffer) =
+          shuffle_ext_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
+            handle, std::move(cluster_keys), std::move(cluster_weights));
+
+        thrust::sort_by_key(handle.get_thrust_policy(),
+                            tmp_keys_buffer.begin(),
+                            tmp_keys_buffer.end(),
+                            tmp_weights_buffer.begin());
+
+        cluster_keys.resize(tmp_keys_buffer.size(), handle.get_stream());
```
seunghwak, line 230: See the code here https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/structure/renumber_edgelist_impl.cuh#L449 (line 449-461).

You can avoid addition resize and shrink_to_fit function calls (This can be more costly than counting # uniques first and also use more memory).

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+
+        cluster_keys.resize(
+          static_cast<size_t>(thrust::distance(cluster_keys.begin(), pair_of_iterators_end.first)),
+          handle.get_stream());
+        cluster_weights.resize(static_cast<size_t>(thrust::distance(cluster_weights.begin(),
+                                                                    pair_of_iterators_end.second)),
+                               handle.get_stream());
+
+        cluster_keys.shrink_to_fit(handle.get_stream());
+        cluster_weights.shrink_to_fit(handle.get_stream());
```
seunghwak, line 253: This code becomes unnecessary if you count # uniques first and resize cluster_keys and cluster_weights properly.

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+  rmm::device_uvector<weight_t> vertex_weights(0, handle.get_stream());                   // #V
+  rmm::device_uvector<vertex_t> louvain_assignment_for_vertices(0, handle.get_stream());  // #V
+  rmm::device_uvector<vertex_t> louvain_of_refined_partition(0, handle.get_stream());     // #V
+
+  //
+  // Edge source/destination cache
+  //
+  edge_src_property_t<graph_view_t, weight_t> src_vertex_weights_cache(handle);
+  edge_src_property_t<graph_view_t, vertex_t> src_louvain_assignment_cache(handle);
+  edge_dst_property_t<graph_view_t, vertex_t> dst_louvain_assignment_cache(handle);
```
seunghwak, line 100: Can't we move this inside the while loop?

i.e. before
```
    if constexpr (graph_view_t::is_multi_gpu) {
      src_vertex_weights_cache =
        edge_src_property_t<graph_view_t, weight_t>(handle, current_graph_view);
      update_edge_src_property(
        handle, current_graph_view, vertex_weights.begin(), src_vertex_weights_cache);
      vertex_weights.resize(0, handle.get_stream());
      vertex_weights.shrink_to_fit(handle.get_stream());
    }
```

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+
+    // After call to relabel, cluster_assignment contains louvain partition of the aggregated graph
+    relabel<vertex_t, multi_gpu>(
+      handle,
+      std::make_tuple(static_cast<vertex_t const*>(leiden_to_louvain_map.first.begin()),
+                      static_cast<vertex_t const*>(leiden_to_louvain_map.second.begin())),
+      leiden_to_louvain_map.first.size(),
+      (*cluster_assignment).data(),
+      (*cluster_assignment).size(),
+      false);
```
seunghwak, line 525: Aren't we calling `relabel` inside `graph_contraction`, and do we need to call `relabel` here again?

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+    raft::copy(louvain_of_refined_partition.begin(),
+               (*cluster_assignment).begin(),
+               (*cluster_assignment).size(),
+               handle.get_stream());
+
+    if (cluster_assignment) {
+      (*cluster_assignment).resize(0, handle.get_stream());
+      (*cluster_assignment).shrink_to_fit(handle.get_stream());
+    }
+
```
seunghwak, line 539: So, here `cluster_keys` and `cluster_weights` are empty, right? If these variables do not carry any information across the while loop iteration, these variables can be located inside the loop to reduce their scope (and lower cognitive load accordingly).

cpp/src/community/detail/refine.cuh
seunghwak, line 1: Better rename to refine_impl.cuh to be consistent.

cpp/src/community/detail/refine.cuh
naimnv: Nope. Will be fixed in next commit

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+  using graph_view_t = cugraph::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>;
+
+  std::unique_ptr<Dendrogram<vertex_t>> dendrogram = std::make_unique<Dendrogram<vertex_t>>();
+
+  graph_t current_graph(handle);
+  graph_view_t current_graph_view(graph_view);
+
+  std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view(
+    input_edge_weight_view);
+  std::optional<edge_property_t<graph_view_t, weight_t>> coarsen_graph_edge_property(handle);
```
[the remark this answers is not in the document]
naimnv: Will rename from 
`edge_weight_view  `to `current_edge_weight_view` 
and 
`coarsen_graph_edge_property `to `coarsen_graph_edge_weight`
in the next commit

cpp/src/community/leiden_impl.cuh
naimnv: I removed the call to `relabel  `that was inside `graph_contraction`

cpp/src/community/leiden_impl.cuh
naimnv: Will reduce scope for few variable in the next commit

cpp/tests/community/new_leiden_test.cpp
naimnv: `run_current_test  `calls  function `leiden` in the test file,  which in turn calls `cugraph::leiden `

cpp/src/community/leiden_impl.cuh
```diff
@@ -42,17 +45,508 @@ void check_clustering(graph_view_t<vertex_t, edge_t, false, multi_gpu> const& gr
...
+  }
+
+#ifdef TIMING
+  detail::timer_display<graph_view_t::is_multi_gpu>(handle, hr_timer, std::cout);
+#endif
+
+  return std::make_pair(std::move(dendrogram), best_modularity);
 }
 
 // FIXME: Can we have a common flatten_dendrogram to be used by both
```
ChuckHastings, line 552: Sure.  You could move this function to `detail/common_methods.{app,cuh,sg,mg}` and have both Louvain and Leiden implementations call it.  This can be deferred to later if we're running out of time.

cpp/src/community/detail/refine.cuh
naimnv: The custom operator prefers ` (higher weight, lower id) `over `(lower weight, higher id) `,  i.e.  `(higher weight, lower id)  >(lower weight, higher id) `, but sure we can use `reduce_op::maximum<thrust::tuple<weight_t, vertex_t>>()`

cpp/include/cugraph/algorithms.hpp
```diff
@@ -715,9 +715,97 @@ template <typename vertex_t, typename edge_t, typename weight_t>
...
+ * @return                           a pair containing:
+ *                                     1) number of levels of the returned clustering
+ *                                     2) modularity of the returned clustering
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+std::pair<size_t, weight_t> leiden(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
+  vertex_t* clustering,
```
seunghwak, line 805: For new functions, we're switching from raw pointers to (device_)span. Better use `raft::device_span<vertex_t> clustering`

cpp/src/community/detail/common_methods.cuh
```diff
@@ -203,16 +209,8 @@ graph_contraction(raft::handle_t const& handle,
                         numbering_indices.size(),
                         new_graph_view.local_vertex_partition_range_first());
```
seunghwak, line 211: `new_graph_view` is used in nowhere.

We're filling `numbering_indices` but it seems like this is not used anywhere as well? Better delete this?

cpp/src/community/detail/common_methods.cuh
```diff
@@ -203,16 +209,8 @@ graph_contraction(raft::handle_t const& handle,
...
-    std::make_tuple(static_cast<vertex_t const*>((*numbering_map).begin()),
-                    static_cast<vertex_t const*>(numbering_indices.begin())),
-    new_graph_view.local_vertex_partition_range_size(),
-    labels.data(),
-    labels.size(),
-    false);
-
-  return std::make_tuple(std::move(new_graph), std::move(new_edge_weights));
+  return std::make_tuple(
+    std::move(new_graph), std::move(new_edge_weights), std::move(numbering_map));
```
seunghwak, line 213: So, what this function basically dose is just calling `coarsen_graph`. Do we really need this additional level of indirection?

```
  auto [new_graph, new_edge_weights, numbering_map] =
    coarsen_graph(handle, graph_view, edge_weights, labels.data(), true);
```
Isn't the code above sufficient to replace this graph_contraction function call? Then, better remove this function.

cpp/src/community/leiden_impl.cuh
```diff
@@ -14,19 +14,22 @@
...
+#include <community/flatten_dendrogram.hpp>
+#include <prims/update_edge_src_dst_property.cuh>
 
-#include <thrust/tuple.h>
+#include <cugraph/detail/shuffle_wrappers.hpp>
+#include <cugraph/detail/utility_wrappers.hpp>
+#include <cugraph/graph.hpp>
+#include <cugraph/utilities/high_res_timer.hpp>
+#include <thrust/sort.h>
+#include <thrust/unique.h>
```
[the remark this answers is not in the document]
seunghwak: You need a space between thrust headers and std headers.

cpp/src/community/leiden_impl.cuh
```diff
@@ -14,19 +14,23 @@
...
 #include <community/flatten_dendrogram.hpp>
-#include <cugraph/algorithms.hpp>
-#include <cugraph/graph_view.hpp>
-#include <cugraph/utilities/error.hpp>
-#include <cugraph/vertex_partition_device_view.cuh>
+#include <cugraph/detail/shuffle_wrappers.hpp>
+#include <cugraph/detail/utility_wrappers.hpp>
+#include <cugraph/graph.hpp>
+#include <cugraph/utilities/high_res_timer.hpp>
+#include <prims/update_edge_src_dst_property.cuh>
```
seunghwak, line 26: Headers under `cpp\include` and internal headers (`cpp\src`) need to be separated by space.

cpp/src/community/detail/common_methods.cuh
```diff
@@ -29,6 +29,7 @@
 #include <cugraph/detail/utility_wrappers.hpp>
 #include <cugraph/graph_functions.hpp>
 
+#include <algorithm>
```
ChuckHastings, line 32: Is this header still necessary?  I suspect you've removed things that used it, as no changes I see require that header.

If we do need the header, I'd suggest putting it after the thrust headers below.

cpp/src/community/detail/mis.hpp
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should this just be 2023?

cpp/src/community/detail/mis_sg.cu
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be only 2023

cpp/src/community/detail/refine.hpp
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be only 2023

cpp/src/community/detail/refine.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#ifdef TIMING
+#include <cugraph/utilities/high_res_timer.hpp>
```
ChuckHastings, line 19: This header (and the ifdef's) aren't referenced as part of this header.

cpp/src/community/detail/refine_impl.cuh
```diff
@@ -0,0 +1,767 @@
+/*
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be only 2023

cpp/src/community/detail/refine_mg.cu
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be only 2023

cpp/src/community/detail/refine_sg.cu
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be only 2023

cpp/src/community/leiden_mg.cu
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Should be 2023

cpp/tests/community/new_leiden_test.cpp
```diff
@@ -0,0 +1,238 @@
+/*
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.  All rights reserved.
```
ChuckHastings, line 2: Should be 2023

ChuckHastings (commented):
A bunch of little cosmetic comments.

cpp/src/community/detail/common_methods.cuh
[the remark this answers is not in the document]
seunghwak: Did you make this change?

Once `labels` becomes read only,

```
  auto new_graph_view = new_graph.view();

  rmm::device_uvector<vertex_t> numbering_indices((*numbering_map).size(), handle.get_stream());
  detail::sequence_fill(handle.get_stream(),
                        numbering_indices.data(),
                        numbering_indices.size(),
                        new_graph_view.local_vertex_partition_range_first());
  relabel<vertex_t, multi_gpu>(
    handle,
    std::make_tuple(static_cast<vertex_t const*>((*numbering_map).begin()),
                    static_cast<vertex_t const*>(numbering_indices.begin())),
    new_graph_view.local_vertex_partition_range_size(),
    labels.data(),
    labels.size(),
    false);
```
is no longer necessary? Right? And this function will just call coarsen_graph, so the caller can just directly call coarsen_graph, isn't it?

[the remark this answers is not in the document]
naimnv: `labels `are not read only. Sorry for the confusion. 
In case of Louvain, `labels `are reference to dendrogram and  they are renumber according to new indices of contracted graph. So we can't make labels read only. 

In case of Leiden we are not using this function.  
Rather we need two relabeling

1. To get back **louvain cluster keys** of contracted nodes (where nodes are contracted based on refined leiden cluster keys)
2.  renumber the dendrogram based on number of unique **louvain cluster keys** 

I can change the function to return 
```
std::tuple<
  cugraph::graph_t<vertex_t, edge_t, false, multi_gpu>,
  std::optional<edge_property_t<graph_view_t<vertex_t, edge_t, false, multi_gpu>, weight_t>>>
```
instead of additional `numbering_map`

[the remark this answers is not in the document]
naimnv: Changed in last commit.

ChuckHastings (approved):

seunghwak (approved):

PR #2996, opened 2022-11-29
Use Vertex ID Offsets in CuGraphStorage

### LINKED ISSUE #3069: [FIX] PropertyGraph & MGPropertyGraph renumber_vertices/edges_by_type Fails to Preserve Data Type
Currently, `renumber_vertices_by_type` and `renumber_edges_by_type` upcast the index to `int64`.  This causes errors when calling sampling since the index dtype doesn't match the src/dst dtype.

This issue should be fixed by saving the original dtype and forcing the dtype of the new index to match the old dtype using `astype`, or a similar solution.

alexbarghi-nv:
Users Vertex ID offsets in `CuGraphStore`.  Offers the option of using pre-computed offsets by setting `renumber_graph=False`.  Default behavior is to warn the user that it is computing offsets and save the original vertex ids in a new column.  If `renumber_graph=True`, there is no warning, `renumber_vertices_by_type()` is called, and the vertex ids are overwritten.

Due to a bug in `MGPropertyGraph`, if vertices are renumbered, edges must also be renumbered.

Closes https://github.com/rapidsai/graph_dl/issues/95
Closes #3069 

- [x] Update Remote Graph tests
- [x] Update CGS e2e tests
- [x] Update SG PyG Extension Tests
- [x] Update MG PyG Extension Tests
- [x] Verify tests that are currently broken in CI

Includes fix for #3069 
Merge after #3012


cpp/CMakeLists.txt
```diff
@@ -349,6 +349,7 @@ if (USE_CUGRAPH_OPS)
     target_link_libraries(cugraph
         PUBLIC
             rmm::rmm
+            rt
```
VibhuJawa, line 352: Dumb question, is this intention ?

cpp/CMakeLists.txt
```diff
@@ -363,6 +364,7 @@ else()
...
```
VibhuJawa, line 367: Dumb question, is this intention ? What does this do.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -598,7 +598,9 @@ def get_vertex_data(self, vertex_ids=None, types=None, columns=None):
                 # FIXME: invalid columns will result in a KeyError, should a
                 # check be done here and a more PG-specific error raised?
                 df = df[[self.type_col_name] + columns]
-            return df.reset_index()
+            df_out = df.reset_index().persist()
+            df_out.index = df_out.index.astype(df.index.dtype)
+            return df_out
```
VibhuJawa, line 603: This will add considerable memory over-head , would it be possible to break down this as a separate PR , we have a big memory overhead anyways ?

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -849,7 +849,9 @@ def get_vertex_data(self, vertex_ids=None, types=None, columns=None):
                 # FIXME: invalid columns will result in a KeyError, should a
                 # check be done here and a more PG-specific error raised?
                 df = df[[self.type_col_name] + columns]
-            return df.reset_index()
+            df_out = df.reset_index()
+            df_out.index = df_out.index.astype(df.index.dtype)
```
VibhuJawa, line 853: What does this do ?

VibhuJawa (changes requested):
The pyg side changes look good but i think we should the PG changes separately as they require more discussion and testing

python/cugraph-service/server/cugraph_service_server/testing/utils.py
```diff
@@ -34,6 +34,13 @@ def create_tmp_extension_dir(file_contents, file_name="my_extension.py"):
     return tmp_extension_dir
 
 
+_failed_subproc_err_msg = (
```
VibhuJawa, line 37: These are coming from https://github.com/rapidsai/cugraph/pull/3012 right ? 

We should let that merge in first .

cpp/CMakeLists.txt
alexbarghi-nv: This is intentional, for the time being.  Previously, this was necessary to build both local and CI but the problem may be fixed now.  Once that's confirmed I'll revert this change.

cpp/CMakeLists.txt
alexbarghi-nv: same as other one

cpp/CMakeLists.txt
alexbarghi-nv: This declares rt as a library dependency so it can be properly found by cmake.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
alexbarghi-nv: I could make this a separate PR, but that would block this PR.

alexbarghi-nv: I'd rather just fix the code.  I think I can safely drop `persist` entirely here.

alexbarghi-nv: I'm just worried the whole dataframe will get recomputed.  But I will add a check that skips this if there is no change in datatype.

python/cugraph/cugraph/structure/property_graph.py
alexbarghi-nv: This ensures the dtype of the returned vertex ids matches the actual vertex id dtype.  Otherwise, there are issues encountered when passing returned vertex ids into cugraph algorithms because the datatype of the input vertices does not match the datatypes of the graph.  I added some explanatory comments and also cleaned up the code a bit.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
VibhuJawa: `astype` wont  cause dataframe to be recomputed it will just be a task that gets added to the DAG.

python/cugraph-service/server/cugraph_service_server/testing/utils.py
alexbarghi-nv: right, this is dependent on 3012; I'll make a note of that.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
alexbarghi-nv: ok, so the changes I just pushed should resolve this then

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -849,7 +849,16 @@ def get_vertex_data(self, vertex_ids=None, types=None, columns=None):
                 # FIXME: invalid columns will result in a KeyError, should a
                 # check be done here and a more PG-specific error raised?
                 df = df[[self.type_col_name] + columns]
-            return df.reset_index()
+
+            df_out = df.reset_index()
```
VibhuJawa, line 853: I think you might want to drop the index and not have it lying around. 
```suggestion
            df_out = df.reset_index(drop=True)
```

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -1245,7 +1254,15 @@ def get_edge_data(self, edge_ids=None, types=None, columns=None):
                 df = df[
                     [self.src_col_name, self.dst_col_name, self.type_col_name] + columns
                 ]
-            return df.reset_index()
+            df_out = df.reset_index()
```
VibhuJawa, line 1257: I asume you dont want the index to be lying around in memory
```suggestion
            df_out = df.reset_index(drop=True)
```

VibhuJawa:
@alexbarghi-nv , ready for second round of reviews ?

alexbarghi-nv:
Go ahead

python/cugraph-pyg/cugraph_pyg/tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+from dask_cuda.initialize import initialize
+
+from cugraph.dask.comms import comms as Comms
+from cugraph.dask.common.mg_utils import get_visible_devices
+
+
+# module-wide fixtures
+
+
+# Spoof the gpubenchmark fixture if it's not available so that asvdb and
```
rlratzel, line 30: [This ](https://github.com/rapidsai/cugraph/blob/1cb20d10b64bd00537b4fa479953d07ee90c6abb/benchmarks/cugraph/pytest-based/bench_cugraph_uniform_neighbor_sample.py#L23-L33)is another pattern we've been using (although it still requires `pytest-benchmark` which is much more common).

python/cugraph-pyg/cugraph_pyg/tests/conftest.py
```diff
@@ -0,0 +1,77 @@
...
+@pytest.fixture(scope="module")
+def dask_client():
+    dask_scheduler_file = os.environ.get("SCHEDULER_FILE")
+    cluster = None
+    client = None
+    tempdir_object = None
+
+    if dask_scheduler_file:
+        # Env var UCX_MAX_RNDV_RAILS=1 must be set too.
+        initialize(
```
rlratzel, line 51: I know this was copied from existing fixtures elsewhere, but I think it might be better to just use all defaults for this call (cc @VibhuJawa). FYI [here's](https://github.com/rapidsai/cugraph/blob/c6028e1fb9c1dc07b5c1411ce5f6d7fd45ee7368/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py#L395-L407) the latest configurations for multi-node and `LocalCUDACluster` we've been using to get the latest benchmark numbers (and we'll probably need to update the other fixtures as well).

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -857,6 +870,7 @@ def add_edge_data(
                 tmp_df, self.__edge_prop_dataframe
             )
         self.__edge_prop_dtypes.update(new_col_info)
+        print("tmp df:", tmp_df.index.dtype)
```
rlratzel, line 873: I'm assuming we don't want this here?

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -1576,6 +1606,13 @@ def is_multi_gpu(self):
         """
         return True
 
+    def is_remote(self):
```
rlratzel, line 1609: This seems like a requirement from higher-level abstraction that doesn't apply generally to PG.  Is there a way we can avoid having this method?

python/cugraph-pyg/cugraph_pyg/tests/conftest.py
VibhuJawa: Yup, we should not enable anything here and stick to defaults.

alexbarghi-nv: I fixed this in another upcoming PR but I can make the change here too.

python/cugraph-pyg/cugraph_pyg/tests/conftest.py
alexbarghi-nv: changed to the new pattern

python/cugraph/cugraph/dask/structure/mg_property_graph.py
alexbarghi-nv: should be fixed now

alexbarghi-nv:
rerun tests

rlratzel (approved):

VibhuJawa (approved):
LGTM

alexbarghi-nv:
@rlratzel I think this can be merged now

PR #3009, opened 2022-11-30
Update dask-cuda version and disable wheel builds in CI

vyasr:
We need to install dask-cuda from source during wheel builds for cugraph CI to pass because dask evolves too fast to pin. We also don't want to build wheels on every PR, so we'll need that disabled. I'll make that change once I confirm that wheel builds pass tests here.

This PR also fixes an unrelated issue causing doctests to fail. I'm guessing that because the docstring is in tests/mg it isn't being discovered during conda gpuCI because we exclude that directory from the pytest command in ci/test.sh.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -1,9 +1,6 @@
 name: cuGraph wheels
 
 on:
-  push:
-    branches:
-      - 'pull-request/[0-9]+'
   workflow_call:
     inputs:
       versioneer-override:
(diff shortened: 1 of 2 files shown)


vyasr:
Once the full suite of tests passes with the proper dask-cuda pinning I'll disable the wheels CI and this PR can be merged.

BradReesWork (approved):

sevagh (approved):

PR #3014, opened 2022-12-01
Adds additional checks to skip PLC and cugraph tests, and dataset downloads based on the python sources modified

### LINKED ISSUE #3013: cugraph CI script should skip more tests and dataset downloads based on python sources changed
A somewhat recent change allows the cugraph CI script to skip C++ tests if no C++ sources were changed in a PR. This should be extended to skip certain python tests based on the various python packages edited and their hierarchy in the stack.  For example, a change to cugraph-service sources should not require the PLC and cugraph tests to be run.

Also, the expensive dataset download from S3 step is only needed for C++ tests, so that should be skipped if C++ tests are not being run.

rlratzel:
closes #3013 

This PR further streamlines CI testing based on the sources modified, skipping tests that cannot have been affected by the changes in the PR.  For example, a change to cugraph-service sources cannot affect the results of the C++, PLC, or cugraph tests.

Similarly, the dataset download step is only needed for C++ tests, so it is skipped if C++ tests are skipped.


rlratzel:
rerun tests

reason: [This PR](https://github.com/rapidsai/cugraph/pull/3036) was merged which should fix the failing PG tests.

rlratzel:
@bdice - I was wondering if you had any thoughts about the more aggressive test skipping in this PR, especially if it happens to be relevant to the upcoming move to GHA.

bdice:
@rlratzel The conditional logic for test skipping is rather complicated. I haven’t looked at this PR but that’s my feeling from seeing the existing logic. With GHA, we will have 4 separate jobs for C++ build/test and Python build/test. C++ tests will run in parallel with the Python build step, so it may not really matter if they are skipped for the total end-to-end runtime. For now @ajschmidt8 and I are planning to make GHA run all tests for all packages and let the cugraph team apply and maintain any logic more complex than that. We can also parallelize Python tests across multiple jobs. For example, cugraph in one job and pylibcugraph, cugraph-service in another.

rlratzel:
closing to possibly re-open later using new GitHub Actions CI scripts.

PR #3042, opened 2022-12-05
[skip-gpuci] Add dependencies.yaml for rapids-dependency-file-generator

ChuckHastings:
First cut at dependencies.yaml.

We could certainly decompose this more, but wanted to get something out there quickly.


dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+             packages:
+               - cudatoolkit=11.4
+           - matrix:
+               cuda: "11.2"
+             packages:
+               - cudatoolkit=11.2
+           - matrix:
+               cuda: "11.0"
+             packages:
+               - cudatoolkit=11.0
```
bdice, line 40: We can remove CUDA 11.0. ([Dropped in 22.10](https://docs.rapids.ai/notices/rsn0021/))
```suggestion
```

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+   common_build:
+     common:
+       - output_types: [conda]
+         packages:
+           - cmake>=3.23.1,!=3.25.0
+       - output_types: conda
+         packages:
```
bdice, line 47: Only one packages list is needed here, since all the outputs are conda packages.
```suggestion
```

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+           - ninja
+           - pip
+   cpp_build:
+     common:
+       - output_types: conda
+         packages:
+           - nvcc_linux-64=11.5
```
bdice, line 54: We'll need to split this by arch, since ARM doesn't use the same package name (it uses `nvcc_linux-aarch64`). See example here from cudf: https://github.com/rapidsai/cudf/blob/2048829ec66af1ceb7a9801702417d12ef13c77a/dependencies.yaml#L83-L94

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+           - c-compiler
+           - cxx-compiler
+           - librmm=23.02.*
+           - libraft-headers=23.02.*
+           - libraft-distance=23.02.*
+           - libcugraphops=23.02.*
+           - clang=11.1.0
+           - clang-tools=11.1.0
```
bdice, line 62: I think these can be dropped unless you know a reason why not. Historically these were used only to provide clang-format, as I understand it.

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+           - boost
```
bdice, line 63: Where do we depend on `boost`?

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+         packages:
+           - cmake>=3.23.1,!=3.25.0
+       - output_types: conda
+         packages:
+           - ninja
+           - pip
+   cpp_build:
+     common:
+       - output_types: conda
+         packages:
```
bdice, line 53: Let's sort the contents of all `packages:` lists alphabetically.

dependencies.yaml
```diff
@@ -0,0 +1,107 @@
...
+           - dask-cudf=23.02.*
+           - dask>=2022.11.1
+           - distributed>=2022.11.1
+           - nccl>=2.9.9
+           - ucx-py=0.30.*
+           - ucx-proc=*=gpu
+           - scipy
+           - networkx>=2.5.1
+           - scikit-build>=0.13.1
+           - python>=3.8,<3.10
```
bdice, line 86: Let's use a separate dependency list for Python versions. See cudf example: https://github.com/rapidsai/cudf/blob/2048829ec66af1ceb7a9801702417d12ef13c77a/dependencies.yaml#L190-L204

bdice (commented):
Thanks @ChuckHastings! I have some preliminary comments.

dependencies.yaml
ChuckHastings: I'm OK with removing things.  If we remove these from here, how does the C++ code formatting work?  Is there some other mechanism we are using?

dependencies.yaml
ChuckHastings: Looks like we used to depend on boost (circa version 0.6) but we may have eliminated that dependency.  I'm testing without it now.

dependencies.yaml
bdice: Many other RAPIDS repos have migrated to using `pre-commit` as a single source of truth for CI style checks. I will file a PR to update cugraph to match, since that's something we've been doing as part of the GitHub Actions migrations.

For clang-format specifically, switching to pre-commit lets us eliminate this section:
https://github.com/rapidsai/cugraph/blob/efcf6a8e9bd07936fd371d16eb2563502c383ca3/ci/checks/style.sh#L33-L36
as well as the entire `run-clang-format.py` file in favor of a single hook definition [like this example from cudf](https://github.com/rapidsai/cudf/blob/0c8eeb5a3caacf5881621f84da891a5d804d43d7/.pre-commit-config.yaml#L62-L67).

Then the style check script can also be simplified from [~79 lines](https://github.com/rapidsai/cugraph/blob/efcf6a8e9bd07936fd371d16eb2563502c383ca3/ci/checks/style.sh) to [~23 lines, like cudf](https://github.com/rapidsai/cudf/blob/branch-23.02/ci/check_style.sh).

dependencies.yaml
ChuckHastings: Removing boost

ChuckHastings:
@bdice - addressed your comments, ready for more feedback.
@rlratzel - suggestions/thoughts?  We talked about some other changes.

bdice (approved):
Thanks @ChuckHastings! This PR looks good enough to merge. We may have some additional tweaks as we work through GitHub Actions migrations, but this is a great starting point. Much appreciated!

ajschmidt8:
@ChuckHastings, I pushed some updates here. Let me know if you have any questions. It's mostly just some cleanup, but I also deleted the manually managed `conda/environment` files and replaced them with a file generated from `dependencies.yaml`. I only added the CUDA 11.5 variant and removed the old variants. That seems to be the convention we've gone with for other repositories.

Additionally I added a GitHub Action workflow to ensure that the generated files and source `dependencies.yaml` file stay in sync. This will be an optional GH check for now, but it will be required once we merge the subsequent GH Actions PR for `cugraph`.

bdice (approved):

BradReesWork (approved):

ajschmidt8:
@gpucibot merge

PR #3045, opened 2022-12-06
Fix cugraph_c target name in Python builds

vyasr:
pylibcugraph is currently linking to the `cugraph_c` target directly rather than via the exported alias `cugraph::cugraph_c`. I am unsure under exactly what circumstances the former should be working, but I don't think it is reliable. I see failures locally without the change. I have currently made this PR into 22.12 because it seems like an important fix, but I can push it to 23.02 if needed.

### DIFF
--- a/python/pylibcugraph/pylibcugraph/internal_types/CMakeLists.txt
+++ b/python/pylibcugraph/pylibcugraph/internal_types/CMakeLists.txt
@@ -15,7 +15,7 @@
 set(cython_sources
     sampling_result.pyx
 )
-set(linked_libraries cugraph::cugraph;cugraph_c)
+set(linked_libraries cugraph::cugraph;cugraph::cugraph_c)
 
 rapids_cython_create_modules(
   CXX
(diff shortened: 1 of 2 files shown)


vyasr:
I'd like @robertmaynard's take on whether this is critical to get into 22.12. I'm surprised that this isn't breaking more builds right now since I can reproduce failures without it locally (it won't affect wheel builds, or any other build where the C++ library is built as part of the Python package, but I would expect it to break conda builds). I am curious if the issue in #3041 was just blocking builds from getting far enough to notice this issue.

robertmaynard:
> I'd like @robertmaynard's take on whether this is critical to get into 22.12.

I would say this is critical only if it is blocking getting conda builds out the door. For people building from source we can provide a minimal set of instructions to unblock when building 22.12 locally.


I expect you are correct that #3041 was hiding this issue

rlratzel (approved):

raydouglass (approved):

PR #3054, opened 2022-12-06
Also use build isolation for editable builds.

vyasr:
This was an oversight in #3052. This change only impacts developers, so it is not necessary to hotfix like #3052.


vyasr:
Superseded by #3063 since we do need to hotfix some related issues anyway.

PR #3063, opened 2022-12-08
Don't install dependencies with pip in build.sh

### LINKED ISSUE #3057: Change developer install scripts to avoid redundant cupy installs
The latest 23.02 changes to the build/install scripts (top-level `build.sh` and related) build and install the sources using `pip` and other tools.  This results in both `cupy` and `cupy-cuda11x` being installed, and the following warning at runtime:
```
  CuPy may not function correctly because multiple CuPy packages are installed
  in your environment:

    cupy, cupy-cuda11x

  Follow these steps to resolve this issue:

    1. For all packages listed above, run the following command to remove all
       existing CuPy installations:

         $ pip uninstall <package_name>

      If you previously installed CuPy via conda, also run the following:

         $ conda uninstall cupy

    2. Install the appropriate CuPy package.
       Refer to the Installation Guide for detailed instructions.

         https://docs.cupy.dev/en/stable/install.html
```

The scripts should be changed so this condition doesn't happen.

vyasr:
build.sh should assume that all dependencies are already installed, for the same reasons as #3052.

Resolves #3057

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -104,7 +104,7 @@ BUILD_CPP_MG_TESTS=OFF
 BUILD_ALL_GPU_ARCH=0
 BUILD_WITH_CUGRAPHOPS=ON
 CMAKE_GENERATOR_OPTION="-G Ninja"
-PYTHON_ARGS_FOR_INSTALL="-m pip install --no-build-isolation ."
+PYTHON_ARGS_FOR_INSTALL="-m pip install --no-build-isolation --no-deps ."
 
 # Set defaults for vars that may not have been defined externally
 #  FIXME: if PREFIX is not set, check CONDA_PREFIX, but there is no fallback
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -175,7 +175,7 @@ if hasArg --cmake_default_generator; then
     CMAKE_GENERATOR_OPTION=""
 fi
 if hasArg --pydevelop; then
-    PYTHON_ARGS_FOR_INSTALL="-m pip install -e ."
+    PYTHON_ARGS_FOR_INSTALL="-m pip install -e --no-build-isolation --no-deps ."
```
rlratzel, line 178: Looks like `-e` has to be at the end.
```
ERROR: --no-build-isolation is not a valid editable requirement. It should either be a path to a local project or a VCS URL (beginning with bzr+http, bzr+https, bzr+ssh, bzr+sftp, bzr+ftp, bzr+lp, bzr+file, git+http, git+https, git+ssh, git+git, git+file, hg+file, hg+http, hg+https, hg+ssh, hg+static-http, svn+ssh, svn+http, svn+https, svn+svn, svn+file).
```

rlratzel (changes requested):

build.sh
rlratzel: FYI, I made this change in [my PR](https://github.com/rapidsai/cugraph/pull/3065) and tested it locally. We should be able to close this PR now if that's easier.

vyasr:
Closing since these changes have been pulled into #3065

PR #3049, opened 2022-12-06
Replace clock_gettime with std::chrono::steady_clock

seunghwak:
* Replace clock_gettime in HighResTimer with std::chrono::steady_clock in high_res_timer.hpp
* Move high_res_timer.hpp from src/utilities to include/cugraph/utilities
* Replace HighResClock (tests/utilities/high_res_clock.h) with HighResTimer

PR #3071, opened 2022-12-09
Benchmark tool for uniform neighbor sampling C API

ChuckHastings:
This PR contains a new benchmarking tool for timing the uniform neighbor sampling algorithm from the C API.

The code is a mix of C/C++ - to take advantage of the C++ test infrastructure that we have built.

This branch contains a hack (adding `rt` to the libraries to work around the problem corrected by #3049).  Once PR 3049 is merged we can undo that hack and this will be ready to review/merge.

### DIFF
--- a/cpp/tests/utilities/test_utilities_impl.cuh
+++ b/cpp/tests/utilities/test_utilities_impl.cuh
@@ -19,9 +19,17 @@
 #include <utilities/device_comm_wrapper.hpp>
 #include <utilities/test_utilities.hpp>
 
+#include <cugraph/detail/shuffle_wrappers.hpp>
 #include <cugraph/graph_functions.hpp>
+#include <cugraph/utilities/host_scalar_comm.hpp>
 
 #include <raft/core/device_span.hpp>
+#include <raft/random/rng.cuh>
+#ifndef NO_CUGRAPH_OPS
+#include <cugraph-ops/graph/sampling.hpp>
+#endif
+
+#include <rmm/device_scalar.hpp>
 
 namespace cugraph {
 namespace test {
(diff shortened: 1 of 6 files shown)


naimnv (dismissed):

cpp/tests/utilities/test_utilities_impl.cuh
```diff
@@ -187,5 +195,61 @@ mg_graph_to_sg_graph(
...
+    int rank           = handle.get_comms().get_rank();
+    if (rank > 0) base_index = vertex_ends[rank - 1];
+  }
+
+  if (!multi_gpu || (handle.get_comms().get_rank() == 0)) {
+    vertex_ids.resize(count, handle.get_stream());
+
+    rmm::device_scalar<vertex_t> d_num_vertices(number_of_vertices, handle.get_stream());
+    raft::random::RngState rng_state(seed);
+
```
seunghwak, line 228: We may need to consider when ```#ifdef NO_CUGRAPH_OPS``` is true.

And should this low level testing utility function be dependent on cugraph-ops? (and under the current cugraph-ops implementation, this will be served by 1 CUDA warp).

We may consider an alternative.

Say when count << num_vertices,

Generate count + alpha random indices and remove duplicates and take first count (generate additional random numbers if the resulting unique indices are fewer than count).

If count is in the same order of num_vertices, use thrust::shuffle to shuffle indices from 0 to num_vertices - 1 and take the first count indices.

cpp/tests/utilities/test_utilities_impl.cuh
```diff
@@ -187,5 +195,61 @@ mg_graph_to_sg_graph(
...
+      host_scalar_allgather(handle.get_comms(), number_of_vertices, handle.get_stream());
+
+    vertex_ends.resize(vertex_sizes.size());
+
+    std::inclusive_scan(vertex_sizes.begin(), vertex_sizes.end(), vertex_ends.begin());
+
+    number_of_vertices = vertex_ends.back();
+    int rank           = handle.get_comms().get_rank();
+    if (rank > 0) base_index = vertex_ends[rank - 1];
+  }
```
seunghwak, line 221: What will be the behavior of this function if count is larger than number_of_vertices?

jnke2016 (dismissed):

ChuckHastings:
Not going to merge this.  We don't have a real call for this capability - it was a temporary branch to do some performance testing.  If we do have a need to do performance testing at the C API we should create a more comprehensive mechanism for this.

PR #3062, opened 2022-12-08
Use pre-commit for CI style checks.

bdice:
This PR adopts `pre-commit` in CI style checks and updates a few of the hooks (flake8, clang-format, and the copyright checker).

This helps with the ongoing transition to GitHub Actions by centralizing style checks.

I also applied a significant set of changes from `clang-format` and `flake8` suggestions in commits cfe6395 and adf0741. To minimize the diff in this PR, we can split those two commits into their own PRs with some small tweaks to the `pre-commit` configuration. Let me know if you'd like me to do this.

PR #3076, opened 2022-12-13
Add GitHub Actions Workflows

bdice:
This PR adds GitHub Actions workflows to `cugraph`.

**Wait until after the 22.12 release is complete (and potentially until other repos are ready to migrate to GHA) to merge.**

### Task list

Coverage required for this PR:
- [x] C++ tests
  - [x] libcugraph
  - [x] libcugraph_c
- [x] Python tests
  - [x] pylibcugraph
  - [x] cugraph
  - [x] benchmarks
  - [x] cugraph_pyg
  - [x] cugraph-service
- [x] Codecov
- [x] Style checks (requires #3062)

Future work:
- Build Metrics Report (copy from cudf when it is added)
- JUnit output annotations? https://github.com/marketplace/actions/junit-report-action
- Reduce exit code handling verbosity (use global variable with `trap` function)
- Set up `rapids-dependency-file-generator` pre-commit hook
- Docs builds


ci/check_style.sh
```diff
@@ -0,0 +1,25 @@
...
+rapids-dependency-file-generator \
+  --output conda \
+  --file_key checks \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch);py=${RAPIDS_PY_VERSION}" | tee env.yaml
+
+rapids-mamba-retry env create --force -f env.yaml -n checks
+conda activate checks
+
+set +e
+
```
bdice, line 18:
```suggestion
```

conda/recipes/cugraph-dgl/meta.yaml
```diff
@@ -20,22 +20,22 @@ requirements:
   host:
     - python x.x
     # FIXME: this pin can be removed once we move to the GitHub Actions build process
-    - setuptools<=65.2.0
+    - setuptools <=65.2.0
```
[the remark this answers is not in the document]
ajschmidt8: We'll have to wait to remove that pin until we disable Jenkins.

For whatever reason, there was a certain version of `setuptools` that was not working correctly with the `--no-remove-work-dir` flags used in the Jenkins CI scripts below. However, the Jenkins process relies on those flags to function correctly. So it was easier to add a `setuptools` version specifier until we disable Jenkins.

- https://github.com/rapidsai/cugraph/blob/39c7f6ee3465d73bb2caac7c168b321b91121b1d/ci/cpu/build.sh#L123-L132

This isn't listed in the `FIXME`, but we should also remove any `preserve_egg_dir: True` lines in the recipes after we disable Jenkins as well.

dependencies.yaml
```diff
@@ -122,3 +138,14 @@ dependencies:
           - sphinxcontrib-websupport
           - sphinx-markdown-tables
           - sphinx-copybutton
+  test_python:
+    common:
+      - output_types: [conda, requirements]
+        packages:
+          - networkx>=2.5.1
```
bdice, line 145: We should add `networkx` back to the "build" dependencies above (as well as leave it here).

`networkx` should be considered a runtime dependency of the package, see https://github.com/rapidsai/cugraph/blob/df12979cd25f22b28c52417fc3b3a6a55fcdfc57/python/cugraph/cugraph/experimental/compat/nx/Graph.py#L14

This was also discussed in [Slack](https://nvidia.slack.com/archives/CALREMR0T/p1671132367806729?thread_ts=1671128173.982389&cid=CALREMR0T) and resulted in https://github.com/rapidsai/cugraph/issues/3084.

cc: @rlratzel

ajschmidt8: fixed in 9608f16

ajschmidt8:
I reverted the `pytest-xdist` commits for now.

It seems that they're causing some issues when running `pytest`.

We can look into re-adding them in a future PR.

ajschmidt8:
rerun tests

bdice:
While reviewing this PR, I looked at Python tests a bit further. `cugraph`'s pytests take the bulk of the time, at about 23:13 (mm:ss). The rest of the packages are `pylibcugraph` at 15 seconds, cugraph benchmarks at 23 seconds, `cugraph_pyg` at 55 seconds, and `cugraph-service` at about 1:40. We would see no significant benefit from splitting the test jobs as we did for cudf / cuml. However, we would see benefits from enabling `pytest-xdist` for `cugraph`. I did a bit of looking at the pytest-xdist failures. The major problem is that test ids are not deterministic. For example:

```
==================================== ERRORS ====================================
_____________________________ ERROR collecting gw4 _____________________________
Different tests were collected between gw5 and gw4. The difference is:
--- gw5
+++ gw4

@@ -678,62 +678,62 @@

 cugraph/tests/test_betweenness_centrality.py::test_betweenness_centrality_nx[True-True-graph_file0]
 cugraph/tests/test_betweenness_centrality.py::test_betweenness_centrality_nx[True-True-graph_file1]
 cugraph/tests/test_betweenness_centrality.py::test_betweenness_centrality_nx[True-True-graph_file2]
-cugraph/tests/test_bfs.py::test_bfs[seed=42,depth_limit=None-ds=<cugraph.experimental.datasets.dataset.Dataset object at 0x7f37ea6ccb50>,dirctd=True-cugraph.Graph]
-cugraph/tests/test_bfs.py::test_bfs_invalid_start[seed=42,depth_limit=None-ds=<cugraph.experimental.datasets.dataset.Dataset object at 0x7f37ea6ccb50>,dirctd=True-cugraph.Graph]
...
+cugraph/tests/test_bfs.py::test_bfs[seed=42,depth_limit=None-ds=<cugraph.experimental.datasets.dataset.Dataset object at 0x7f2423f95cd0>,dirctd=True-cugraph.Graph]
+cugraph/tests/test_bfs.py::test_bfs_invalid_start[seed=42,depth_limit=None-ds=<cugraph.experimental.datasets.dataset.Dataset object at 0x7f2423f95cd0>,dirctd=True-cugraph.Graph]
...
```
The issue is that test ids are process-specific because of the inclusion of `cugraph.experimental.datasets.dataset.Dataset object at 0x7f2423f95cd0` as a parameter. If we could alter this such that the pytest id used a dataset _name_ rather than its memory location (any unique string for each dataset that is consistent across processes would suffice), we could use `pytest-xdist` and save some time and maximize GPU utilization as we do for cudf, cuml, and other libraries.

@rlratzel It looks like you wrote the code for `gen_fixture_param_product`. In a follow-up PR, it would be good to alter the test generation logic to provide consistent ids. There are multiple approaches:
- Specify that the test ids should use some piece of the metadata (or the filename used to generate the `Dataset`?) instead of the raw result of stringifying the `Dataset` object (which exposes the process-specific memory address)
- Add a `def __repr__(self):` method or `def __str__(self):` to `Dataset` so that stringifying a `Dataset` doesn't produce a raw memory address

Once that is addressed, we should be able to turn `pytest-xdist` on and save some CI time. 🕥 
(cc: @ajschmidt8 since you were curious about the failures)

bdice (commented):
I reviewed the current state of the code and skimmed the latest CI logs. All looks good to merge, with two small comments.

ajschmidt8 (approved):

rlratzel (approved):

rlratzel:
> it would be good to alter the test generation logic to provide consistent ids. There are multiple approaches:
> ...
> Once that is addressed, we should be able to turn `pytest-xdist` on and save some CI time.

Yes, there is some tech debt in our test IDs in parametrized tests at the moment, with the use of the default `Dataset` repr being a big part.  Others include having IDs which aren't providing meaningful info to devs, such as `test_betweenness_centrality_nx[True-True-graph_file0]`.  We'll address this in a separate PR, and I filed [this issue](https://github.com/rapidsai/cugraph/issues/3094) for it.

ajschmidt8:
The segfaults that are causing the current failures are not specific to GH Actions.

Therefore, it was decided (via a Slack conversation) that we will admin merge this GH Actions PR now that it's ready and address the segfaults in a follow-up PR.

Admin merging now.

PR #2997, opened 2022-11-29
Update Uniform Neighborhood Sampling API

ChuckHastings:
Update the Uniform Neighborhood Sampling C API to address some shortcomings:
* Return edge id, edge type, edge weight directly from graph
* Get rid of the remove duplicates functionality
* Improve C API testing to cover a wider range of conditions (multigraph, batch capability, error conditions)

So far this PR only defines the API change.

partially addresses #2994 
closes #2597 
closes #2774

PR #2887, opened 2022-11-03
[REVIEW] Optimize dask.uniform_neighbor_sample

VibhuJawa:
This PR  closes https://github.com/rapidsai/cugraph/issues/2872 and part of https://github.com/rapidsai/graph_dl/issues/74 .  

On a preliminary benchmark i am seeing a ~`6x`~ ,  `3.5x` improvement  and i expect more improvement on bigger clusters. 

Before PR : 
```python
%%timeit
df = cugraph.dask.uniform_neighbor_sample(dask_g, start_list[:10_000], [20])
187 ms ± 40.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

After PR 
```python
%%timeit
df = cugraph.dask.uniform_neighbor_sample(dask_g, start_list[:10_000], [20])
50.1 ms ± 1.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Updated benchmark 

Before PR: 
```
---------------------------------------------------------------------- benchmark: 3 tests ----------------------------------------------------------------------
Name (time in ms, mem in bytes)                            Mean              GPU mem            GPU Leaked mem            Rounds            GPU Rounds          
----------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_uniform_neigbour_sample_email_eu_core[1000]      170.2833 (1.0)        300,224 (1.0)                   0 (1.0)           1           1
bench_uniform_neigbour_sample_email_eu_core[5000]      191.7691 (1.13)     1,200,944 (4.00)                  0 (1.0)           1           1
bench_uniform_neigbour_sample_email_eu_core[10000]     194.8658 (1.14)     1,200,944 (4.00)                  0 (1.0)           1           1
----------------------------------------------------------------------------------------------------------------------------------------------------------------

```
After PR: 
```
--------------------------------------------------------------------- benchmark: 3 tests --------------------------------------------------------------------
Name (time in ms, mem in bytes)                           Mean            GPU mem            GPU Leaked mem            Rounds            GPU Rounds          
-------------------------------------------------------------------------------------------------------------------------------------------------------------
bench_uniform_neigbour_sample_email_eu_core[1000]      54.0802 (1.0)            0 (1.0)                   0 (1.0)           1           1
bench_uniform_neigbour_sample_email_eu_core[5000]      63.0271 (1.17)           0 (1.0)                   0 (1.0)           1           1
[...]

PR #3082, opened 2022-12-15
Implement New Sampling API in Python

### LINKED ISSUE #2562: [BUG] is_edge_ids in uniform_sampling does not seem to do anything
**Describe the bug**

When we pass `is_edge_ids=True` we dont seem to do anything with it in the ` uniform_neighbor_sample` function (see below).   We should triage this and fix it accordingly. 

https://github.com/rapidsai/cugraph/blob/7d8f0fd63ad58ce6deada5508bfc08ee9aa46d36/python/cugraph/cugraph/sampling/uniform_neighbor_sample.py#L23-L50

### LINKED ISSUE #2871: [BUG]: unnecessary expensive dask compute call
### Version

22.12

Conda

### Describe the bug.

MG uniform neighbor sample takes as input the `start_list` which will be distributed across ranks. To achieve that, It leverages `dask ` by creating a `dask_cudf.Series` from a non dask `start_list`. If the `input_df` was renumbered, the `start list` which is of type dask or non dask  will also be renumbered resulting in a `dask_cudf.Series` regardless. However , there is an expensive and perhaps unnecessary `dask compute` [call](https://github.com/rapidsai/cugraph/blob/branch-22.12/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py#l129). 

Investigate if that call is necessary.

```shell
Run MG uniform neighbor sample with a multi column vertex or non integer datasets which will trigger renumbering
```

### LINKED ISSUE #3073: [FEA] Update Python Uniform Neighbor Sample API to Return Edge Data
Return edge id, edge type, and edge weight from `uniform_neighbor_sample` through the Python API.

- [ ] Verify https://github.com/rapidsai/cugraph/pull/2997
- [ ] Write Pylibcugraph code
- [ ] Write Python code to return these three arrays (will add a flag to `uniform_neighbor_sample` to return them so the current API is not broken)
- [ ] Python tests
- [ ] Update cugraph-PyG tests
- [ ] Update cugraph-DGL tests
- [ ] Update documentation

alexbarghi-nv:
Resolves #3073 
Resolves https://github.com/rapidsai/graph_dl/issues/72
Resolves #2562 
Resolves https://github.com/rapidsai/graph_dl/issues/49
Resolves #2871 

Takes Chuck's changes from #2997 and implements the necessary changes in Python.  Adds tests in Python for the handling of edge id, edge type, etc.  Also updates Python tests to reflect the removal of `count_and_remove_duplicates`


python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -145,22 +220,37 @@ def uniform_neighbor_sample(
         indices_t = numpy.int32
 
     if input_graph.renumbered:
-        start_list = input_graph.lookup_internal_vertex_id(start_list).compute()
+        start_list = input_graph.lookup_internal_vertex_id(start_list)
+
+    if isinstance(start_list, cudf.Series):
+        start_list = dask_cudf.from_cudf(start_list, npartitions=len(Comms.get_workers()))
```
VibhuJawa, line 226: So this can be quite slow, I suggest only doing this for start_list if it is all ready in dask. 

See PR: https://github.com/rapidsai/cugraph/pull/2887

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -100,19 +142,47 @@ def uniform_neighbor_sample(
...
+            df['edge_weight']: dask_cudf.Series
+                Contains the edge weights from the sampling result
+            df['edge_id']: dask_cudf.Series
+                Contains the edge ids from the sampling result
+            df['edge_type']: dask_cudf.Series
+                Contains the edge types from the sampling result
+            df['batch_id']: dask_cudf.Series
+                Contains the batch ids from the sampling result
+            df['hop_id']: dask_cudf.Series
+                Contains the hop ids from the sampling result
```
VibhuJawa, line 185: Does this work in case when we dont have all the properties, like I only want want `sources `, `destinations ` , `edge_id `, `batch_id` and  `hop_id ` for DGL ?

alexbarghi-nv: right now, if edge ids are provided, edge types and weights are also required

alexbarghi-nv: From simpleDistributedGraph.__from_edgelist:
```
if len(edge_attr) == 1:
       input_ddf = input_ddf.rename(columns={edge_attr[0]: self.edgeWeightCol})
       value_col_names = [self.edgeWeightCol]
elif len(edge_attr) == 3:
       weight_col, id_col, type_col = edge_attr
       input_ddf = input_ddf.rename(
           columns={
               weight_col: self.edgeWeightCol,
               id_col: self.edgeIdCol,
               type_col: self.edgeTypeCol,
           }
       )

       value_col_names = [self.edgeWeightCol, self.edgeIdCol, self.edgeTypeCol]
else:
       raise ValueError("Only 1 or 3 values may be provided" "for edge_attr")
```

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: resolved

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Gotcha, Is there a mechanism to get `hop-ids` if I provide say only 1 column (say `edgeIdCol `) ?  The reason i am asking is the memory overhead of adding extra dummy columns will be prohibitive for large scale usecases.

VibhuJawa (changes requested):
Have questions regarding the sampling API for non multiple types. The implimentation looks good otherwise.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: In the C++ code, either a tuple of <edge id, edge type>, or null must be provided

alexbarghi-nv: But I think you can skip weights, it's just not allowed by the Python API right now.

alexbarghi-nv: Also, I see that the Python API adds a weight column, even if one is not specified by the user.  I think this was done for some algorithms which require a default weight value.

alexbarghi-nv: So we would need to do refactoring at both the C++ and Python API level to support passing only edge ids.

alexbarghi-nv: Also, since we are now using `FeatureStore`, I thought there was less memory pressure on the sampling GPUs, giving us the flexibility to work with our current API, and eventually refactor graph creation to have less memory overhead in the future.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -145,22 +220,41 @@ def uniform_neighbor_sample(
...
             _call_plc_uniform_neighbor_sample,
             session_id,
             input_graph._plc_graph[w],
-            start_list[i],
+            ddf[w][0],
             fanout_vals,
             with_replacement,
             weight_t=weight_t,
+            with_edge_properties=with_edge_properties,
+            seed=seed+i, # ensure different seed per GPU
```
ChuckHastings, line 257: How is this seed assigned and updated over the course of the execution of the sampling code?  It will be important that the seeds don't get reused across calls.  For example, if the caller simply increments the seed by 1, then each GPU will be reseeded in the next call with the neighbor GPUs sequence of random numbers - which is not going to be very random overall.

Part of the reason that the `cugraph_rng_state_t` object was added was to allow the state to be initialized once and maintained across calls - which will guarantee good uniqueness across multiple calls to the sampling algorithm.

Might be worth trying to think about how to propagate that design through into this layer as well.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
```diff
@@ -17,10 +17,17 @@
...
 
 def uniform_neighbor_sample(
-    G, start_list, fanout_vals, with_replacement=True, is_edge_ids=False
+    G,
+    start_list,
+    fanout_vals,
+    with_replacement=True,
+    with_edge_properties=False,
+    batch_id_list=None,
+    seed=42,
```
ChuckHastings, line 30: Not sure if providing a default seed is a good idea as this is written.  If the approach suggested above were to be used (creating some sort of stateful randomization object), having a default seed for the stateful object might be reasonable.  But providing a default here will not provide good random samples except in the very first call.

ChuckHastings (commented):
Didn't explore the python rigorously.  The C/C++ code and the python looks good to me.

Big issue, I think, is the seed.  Left some comments, we could get together and discuss if that would be helpful.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Okay, I think for now, Let us stick with what we have but we will need to revisit this sooner rather than later as for large graphs memory becomes prohibitive quickly as we need 1.7x more memory, which can often mean some one requiring the double the nodes. 

The GNN use case from DGL perspective requires the following format: 
- no weights, id, no type .

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
alexbarghi-nv: So is the way to rectify this requiring that the user provide a seed?  Or do we instead associate a random state with a graph object?  Or do we use a random default seed?

ChuckHastings: I don't know what the right solution is from a user experience perspective.  My thought would be:
* Create a stateful python object that represents random number generator state.
* The constructor for that stateful python object would take a seed as a parameter, perhaps defaulting to some value if not provided
* The stateful python object would instantiate a `cugraph_rng_state_t`, initializing it with the seed.  In MG mode, the state would be initialized with `seed + rank`
* The stateful python object would be used by `uniform_neighbor_sample` to send the `cugraph_rng_state_t` into the C API.  Note that the `cugraph_rng_state_t` is modified by the C API, it is updated each time we use it to generate a random number to a new state to create a sequence of random numbers.  Reusing it keeps the sequence of random numbers flowing properly.

I'm not terribly familiar with python semantics.  This stateful python object could be managed by the user (required as a parameter and visible in the python API).  I suppose you could also make it a singleton and hide it from the API, just providing some public function to initialize the seed once.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
ChuckHastings: Without such a construct, you essentially rely on the caller understanding the mechanics of what is going on for the use of the random number generator and creating their own mechanism for providing unique seeds... which is an invitation for disaster.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -152,12 +152,20 @@ def __from_edgelist(
...
+                        "User-provided edge ids and edge "
+                        "types are not permitted for an "
+                        "undirected graph."
+                    )
+                if not legacy_renum_only:
+                    raise ValueError(
+                        "User-provided edge ids and edge "
+                        "types are only permitted when "
+                        "from_edgelist is called with "
+                        "legacy_renum_only=True."
```
VibhuJawa, line 167: Just ran into this. Why do we need this again ? I am running into some memory issues with `legacy_renum_only ` on SG Graphs on this PR

alexbarghi-nv: The non-legacy renumbering removes the additional edge properties.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
alexbarghi-nv: @ChuckHastings I exposed the random state at the PLC level, and added a global random state.  In `pylibcugraph.uniform_neighbor_sample`, the user can provide the random state, or it defaults to the global state, instantiating it if necessary.  Please confirm this is an acceptable solution.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -1204,7 +1224,9 @@ def edge_props_to_graph(
 
         # If a default_edge_weight was specified but an edge_weight_property
         # was not, a new edge weight column must be added.
-        elif default_edge_weight:
+        elif default_edge_weight or create_with_edge_info:
```
rlratzel, line 1227: The comment above looks like it needs to be updated now.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -1611,6 +1650,13 @@ def is_multi_gpu(self):
         """
         return True
 
+    def is_remote(self):
```
rlratzel, line 1653: I thought we removed this, or does it need to be here after all? Same question for SG.

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -32,9 +36,16 @@ def gpubenchmark():
...
 
-    yield client
+    visible_devices = ",".join([str(i) for i in range(1, n_devices + 1)])
+
+    cluster = LocalCUDACluster(
+        protocol="ucx", rmm_pool_size="25GB", CUDA_VISIBLE_DEVICES=visible_devices
+    )
+    client = Client(cluster)
+    Comms.initialize(p2p=True)
+    rmm.reinitialize(pool_allocator=True)
```
rlratzel, line 49: Can this go in `mg_utils.start_dask_client()` instead so we don't have multiple implementations?

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -32,9 +36,16 @@ def gpubenchmark():
...
 
-    stop_dask_client(client)
-    print("\ndask_client fixture: client.close() called")
+    yield client
```
rlratzel, line 51: Don't we need the equivalent of `stop_dask_client()` after this line too to properly stop the cluster?

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -330,8 +330,12 @@ cdef class MGGraph(_GPUGraph):
         assert_CAI_type(weight_array, "weight_array", True)
         if edge_id_array is not None:
             assert_CAI_type(edge_id_array, "edge_id_array")
+            if len(edge_id_array) != len(src_array):
+                raise ValueError('Edge id array must be same length as edgelist')
         if edge_type_array is not None:
-            assert_CAI_type(edge_type_array, "edge_type_array")        
+            assert_CAI_type(edge_type_array, "edge_type_array")
```
rlratzel, line 336: For this and others, if you pass `True` it will allow None for this check, which has the advantage of checking that the value is either a CAI type or None, but nothing else. (see `weight_array` above)
```suggestion
        assert_CAI_type(edge_type_array, "edge_type_array", True)
```

python/pylibcugraph/pylibcugraph/random.pyx
```diff
@@ -0,0 +1,89 @@
...
+            )
+        )
+    
+    # reinterpret as unsigned
+    return h & (2**64 - 1)
+
+class global_random_instance_wrapper:
+    global_random_instance = None
+
+cdef class CuGraphRandomState:
```
rlratzel, line 54: I'm a little confused by how this is supposed to be used. The `__cinit__` implies that that I can have multiple instances if I wanted to, but all methods are classmethods which refer to a global.  Perhaps some docstrings would help?

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -32,9 +36,16 @@ def gpubenchmark():
 
 @pytest.fixture(scope="module")
 def dask_client():
-    client, cluster = start_dask_client(protocol="tcp")
+    n_devices = os.getenv("DASK_NUM_WORKERS", 4)
```
rlratzel, line 39: I think devs have 2 GPUs available more often than 4 (mainly when on their workstation), so can this default be changed to 2?

python/cugraph/cugraph/dask/structure/mg_property_graph.py
alexbarghi-nv: I did remove this.  It may have come back from a bad merge.  I'll remove it again.  🤦‍♀️

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -1032,6 +1048,7 @@ def extract_subgraph(
         check_multi_edges=True,
         renumber_graph=True,
         add_edge_data=True,
+        create_with_edge_info=False,
```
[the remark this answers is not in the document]
alexbarghi-nv: This param shouldn't even exist.  I thought I had reverted this change.  Will remove everything with "create_with_edge_info" again.

python/cugraph/cugraph/dask/structure/mg_property_graph.py
alexbarghi-nv: undid this change

python/cugraph/cugraph/tests/conftest.py
alexbarghi-nv: fixed

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
rlratzel: Sorry I'm late to this conversation, but the best UX in my opinion would be to be consistent with RAPIDS and accept a numpy or cupy `RandomState` instance, as both [cuDF](https://docs.rapids.ai/api/cudf/stable/api_docs/api/cudf.series.sample) and [cuML](https://docs.rapids.ai/api/cuml/stable/api.html#model-selection-and-data-splitting) do.

alexbarghi-nv: Wouldn't that have to be changed at the C++ level?

python/pylibcugraph/pylibcugraph/random.pyx
alexbarghi-nv: multiple instances are ok, see docstring

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
rlratzel: Possibly, I haven't looked into their implementations.  One option would be to rename the `rng_state` param `random_state` like the other libraries call it and accept an int (also like they do). Then eventually also accept a cupy/numpy `RandomState` instance at the python layers. This will give us a path to fully supporting it, consistent with the other libs, without breaking backwards compat.

For now, PLC could still pass a `CuGraphRandomState` that it creates with the int passed in, or the default global instance if not passed in.  It might be nice to avoid exposing `CuGraphRandomState` in the python API anyway since it's apparently non-standard.

python/cugraph/cugraph/tests/conftest.py
alexbarghi-nv: done

python/cugraph/cugraph/tests/conftest.py
alexbarghi-nv: fixed

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -145,22 +212,37 @@ def uniform_neighbor_sample(
...
+        start_list = input_graph.lookup_internal_vertex_id(start_list)
+
+    start_list = start_list.rename(start_col_name).to_frame()
+    if batch_id_list is not None:
+        ddf = start_list.join(batch_id_list.rename(batch_col_name))
+    else:
+        ddf = start_list
+
+    if isinstance(ddf, cudf.DataFrame):
+        splits = cp.array_split(cp.arange(len(ddf)), len(Comms.get_workers()))
```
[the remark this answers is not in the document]
alexbarghi-nv: `iloc[[]]` is valid but I will add another test to make sure sending empty arrays doesn't cause any issues in the C++ code

python/cugraph/cugraph/dask/structure/mg_property_graph.py
```diff
@@ -329,6 +329,22 @@ def vertices_ids(self):
...
+            .reset_index(drop=True)
+        )
+
+    def edge_types_from_numerals(self, nums):
+        return (
+            self.__edge_prop_dataframe[self.type_col_name]
+            .dtype.categories.to_series()
+            .iloc[nums]
+            .reset_index(drop=True)
+        )
```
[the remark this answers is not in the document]
alexbarghi-nv: this looks like code that was accidentally added from another PR.  I will remove it completely.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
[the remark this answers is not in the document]
alexbarghi-nv: @ChuckHastings can you confirm that the C++ sampling can handle empty arrays?  If not, we need to refactor this.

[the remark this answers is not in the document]
ChuckHastings: The C API should handle empty arrays.  If it does not, that's a bug we should fix in the C API.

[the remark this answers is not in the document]
alexbarghi-nv: I added a test that confirms that passing an empty start list results in an empty dataframe.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
alexbarghi-nv: @rlratzel @ChuckHastings based on our libs team discussion this morning, I am going to revert to accepting a seed argument, and not use a global random state.  At some point, we can support a cupy/numpy `RandomState` and properly transmute it at the PLC layer.

rlratzel: cc @jnke2016

rlratzel (approved):

python/pylibcugraph/pylibcugraph/_cugraph_c/random.pxd
```diff
@@ -0,0 +1,36 @@
...
+    cugraph_resource_handle_t,
+)
+
+cdef extern from "cugraph_c/random.h":
+    ctypedef struct cugraph_rng_state_t:
+        pass
+
+    cdef cugraph_error_code_t cugraph_rng_state_create(
+        const cugraph_resource_handle_t* handle,
+        size_t seed,
```
jnke2016, line 31: I had it `uint64_t` but I believe size_t is fine too

PR #3036, opened 2022-12-05
[REVIEW] Unpin `dask` and `distributed` for development

galipremsagar:
This PR unpins `dask` and `distributed` to `2022.12.0+` for `23.02` development.

xref: https://github.com/rapidsai/cudf/pull/12302

PR #3090, opened 2022-12-20
Fix tests for MG property graph

eriknw:
Small PR to fix MGPG tests.

Note that @galipremsagar fixed the equivalent tests for `test_property_graph.py` in #3036 by using `s._constructor`, which is pretty slick.

Also, `start_dask_client` returns a tuple of client and cluster now, so I updated the `dask_client` fixture so the cluster can be closed properly.


alexbarghi-nv (approved):
👍

rlratzel (approved):

PR #3091, opened 2022-12-20
Update `cugraph` recipes

ajschmidt8:
This PR includes the following changes:

- Uses `pin_subpackage` to specify the relationship between `cugraph-service-client` and `cugraph-service-server`
- Switches the `cugraph` dependency to use a regular version specifier since `pin_compatible` and `pin_subpackage` aren't applicable since `cugraph` is built in a different recipe
 
These changes are necessary because `pin_compatible` only works with dependencies that are listed in the `build`/`host` section of a given package ([relevant conda docs](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#extra-jinja2-functions)).

Additionally: 

- the `cudatoolkit` package has been removed from the `run` dependencies since it's not explicitly used in `cugraph-pyg` and comes in transitively through the other dependencies
- a duplicate `libraft-headers` entry has been removed from the `cugraph` recipe

### DIFF
--- a/conda/recipes/cugraph-pyg/meta.yaml
+++ b/conda/recipes/cugraph-pyg/meta.yaml
@@ -35,6 +35,9 @@ requirements:
     - numba>=0.56.2
     - numpy
     - pytorch
+    # TODO: Determine if `cudatoolkit` is a necessary runtime dependency for this package.
+    # If so, the version spec may need to be changed to `11.0>=,<12.0` to match the rest of our packages.
+    # If not, we should remove `cudatoolkit` and update the version string above accordingly.
     - cudatoolkit {{ cuda_version }}.*
     # - pyg (uncomment once pyg 2.2 is released)
 
--- a/conda/recipes/cugraph-pyg/meta.yaml
+++ b/conda/recipes/cugraph-pyg/meta.yaml
@@ -35,9 +35,6 @@ requirements:
     - numba>=0.56.2
     - numpy
     - pytorch
-    # TODO: Determine if `cudatoolkit` is a necessary runtime dependency for this package.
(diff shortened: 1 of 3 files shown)


conda/recipes/cugraph-pyg/meta.yaml
```diff
@@ -35,6 +35,9 @@ requirements:
     - numba>=0.56.2
     - numpy
     - pytorch
+    # TODO: Determine if `cudatoolkit` is a necessary runtime dependency for this package.
+    # If so, the version spec may need to be changed to `11.0>=,<12.0` to match the rest of our packages.
+    # If not, we should remove `cudatoolkit` and update the version string above accordingly.
```
ajschmidt8, line 40: cc: @alexbarghi-nv, do you have thoughts on this?

alexbarghi-nv: I believe it is a necessary runtime dependency, and should be >= 11.5

rlratzel: @ajschmidt8 based on @alexbarghi-nv's comment, this TODO should be removed and `cudatoolkit` should have a pin for `>11.5`

rlratzel (commented):
Thanks for the updates. Just one request below:

conda/recipes/cugraph-pyg/meta.yaml
ajschmidt8, line 40:
```suggestion
```

ajschmidt8: removed this in 3ff9bfa. thanks @alexbarghi-nv and @rlratzel.

rlratzel (approved):
LGTM

ajschmidt8:
@gpucibot merge

PR #3093, opened 2022-12-28
Fix C++ Bugs in Graph Creation with Edge Properties

alexbarghi-nv:
There are currently two bugs in graph creation with edge properties, which this PR resolves.

Closes https://github.com/rapidsai/graph_dl/issues/106


cpp/src/c_api/graph_mg.cpp
```diff
@@ -126,8 +126,8 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
           rmm::device_uvector<edge_t>(edge_ids_->size_, handle_.get_stream());
 
         raft::copy<edge_t>(edgelist_edge_ids.data(),
-                           edge_type_ids_->as_type<edge_t>(),
-                           edge_type_ids_->size_,
+                           edge_ids_->as_type<edge_t>(),
+                           edge_ids_->size_,
```
seunghwak, line 130: Yeah... I think `edge_type_id_t` and `edge_type_ids_` are a bit confusing names. They can be interpreted as edge types and IDs OR IDs (assigned to each edge type) of edge types and may have contributed to this bug.

Won't `edge_type_t` and `edge_types_` be clearer?

@ChuckHastings Based on search under the cugraph/cpp/src, we're mixing `edge_type_t`, `edge_type_id_t`, and even `edge_type_type_t`, should we better be consistent?

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -274,16 +274,12 @@ create_graph_from_edgelist_impl(
...
-  CUGRAPH_EXPECTS(!edgelist_id_type_pairs ||
-                    (edgelist_srcs.size() == std::get<1>((*edgelist_id_type_pairs)).size()),
-                  "Invalid input arguments: edgelist_srcs.size() != "
-                  "std::get<1>((*edgelist_id_type_pairs)).size().");
   CUGRAPH_EXPECTS(renumber,
                   "Invalid input arguments: renumber should be true if multi_gpu is true.");
+  // edgelist_id_pairs may be mismatched depending on how edges were distributed.
+  // This function is responsible for redistributing the tuple.
+  // The sizes of that tuple are checked at the end of this function to ensure it was
+  // redistributed correctly.
```
seunghwak, line 282: What is this? This function is assuming that edge lists (including edge types & IDs) are pre-shuffled.

cpp/src/c_api/graph_mg.cpp
ChuckHastings: Yes, we should be consistent.  I noticed this while working on #2997, but I was scrambling in trying to get something out that people could look at before going on vacation that I didn't address this.

I can try and make things more consistent in #2997.  I think for this PR just correcting the behavior is sufficient.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -29,7 +29,7 @@
 #include <cugraph/utilities/error.hpp>
 #include <cugraph/utilities/host_scalar_comm.hpp>
 
-#include <raft/core/handle.hpp>
+#include <raft/handle.hpp>
```
ChuckHastings, line 32: `raft/handle.hpp` is deprecated.  We should be using `raft/core/handle.hpp`.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: As @seunghwak mentioned, the C++ function here assumes the shuffling has been done before this call.

It appears to me (first glance at code I haven't looked at in 3 weeks) that the C API needs to be updated to shuffle this data properly.

The code that follows should fail if the edge type/ids are not shuffled with the edge src/dst pairs.  Moving the `CUGRAPH_EXPECTS` calls will mask this problem, not correct it.

ChuckHastings: I would suggest reverting all changes in this file.  I can look at fixing the shuffling in #2997.

ChuckHastings (changes requested):

alexbarghi-nv:
I'll revert the changes to `create_graph_from_edgelist_impl` and leave that to you to fix in #2997

alexbarghi-nv:
@ChuckHastings I reverted `create_graph_from_edgelist_impl`, this should be good to go now.

ChuckHastings (approved):

seunghwak (approved):

ChuckHastings:
rerun tests

PR #3096, opened 2022-12-30
Build pip wheels alongside conda CI

sevagh:
This PR adds pip wheel CI to the Conda CI, instead of having them work separately.


.github/workflows/pr.yaml
```diff
@@ -47,3 +51,50 @@ jobs:
...
+    needs: wheel-build-pylibcugraph
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-test.yml@upkeep/manylinux-improvements
+    with:
+      build_type: pull-request
+      package-name: pylibcugraph
+      # On arm also need to install cupy from the specific webpage.
+      test-before-arm64: "pip install cupy-cuda11x -f https://pip.cupy.dev/aarch64"
+      test-unittest: "RAPIDS_DATASET_ROOT_DIR=`pwd`/datasets pytest -v ./python/pylibcugraph/pylibcugraph/tests"
+      test-smoketest: "" #"python ci/wheel_smoke_test_pylibcugraph.py"
```
vyasr, line 75: @rlratzel do you have any recommendations on what we could use as a reasonable smoke test for pylibraft and raft-dask? Here are the current examples for [cudf](https://github.com/rapidsai/cudf/pull/12427/files#diff-f2d06f7697ad781f29c9c3322ac1f522feaba7d01d761d3fd086721d9ab872fb) and [rmm](https://github.com/rapidsai/rmm/pull/1182/files#diff-6f7e4916411f089af7bb37a031ef77d2a63b2d365a9df81d78116af7fb68c048). For context, for PR CI we will build packages on both arm and amd architectures, but to conserve GPU CI resources we will only be running the full test suite on amd and will instead only run a small smoke test on arm. The full test suite will be run for arm on nightly builds.

.github/workflows/build.yaml
```diff
@@ -55,3 +61,70 @@ jobs:
...
+      date: ${{ inputs.date }}
+      package-name: cugraph
+      package-dir: python/cugraph
+      versioneer-override: ${{ inputs.wheel_versioneer_override }}
+
+      # Note that this approach to cloning repos obviates any modification to
+      # the CMake variables in get_cumlprims_mg.cmake since CMake will just use
+      # the clone as is.
+      extra-repo: rapidsai/cugraph-ops
+      extra-repo-sha: branch-22.12
```
vyasr, line 115: As with cumlprims I think we we could probably use inputs.sha here instead, up to you.
```suggestion
      extra-repo-sha: branch-23.02
```

ci/release/update-version.sh
```diff
@@ -78,3 +78,6 @@ sed_runner "s/export UCX_PY_VERSION=.*/export UCX_PY_VERSION='${NEXT_UCX_PY_VERS
 sed_runner "/^ucx_py_version:$/ {n;s/.*/  - \"${NEXT_UCX_PY_VERSION}\"/}" conda/recipes/cugraph/conda_build_config.yaml
 sed_runner "/^ucx_py_version:$/ {n;s/.*/  - \"${NEXT_UCX_PY_VERSION}\"/}" conda/recipes/cugraph-service/conda_build_config.yaml
 sed_runner "/^ucx_py_version:$/ {n;s/.*/  - \"${NEXT_UCX_PY_VERSION}\"/}" conda/recipes/pylibcugraph/conda_build_config.yaml
+
+# Wheel builds clone cugraph-ops, update its branch
+sed_runner "s/extra-repo-sha: branch-.*/extra-repo-sha: branch-${NEXT_SHORT_TAG}/g" .github/workflows/*.yaml
```
vyasr, line 83: Also need to add dask-cuda.

vyasr (commented):
A couple minor comments then need smoke tests.

.github/workflows/build.yaml
```diff
@@ -52,3 +52,66 @@ jobs:
...
+      extra-repo-deploy-key: CUGRAPH_OPS_SSH_PRIVATE_DEPLOY_KEY
+
+      skbuild-configure-options: "-DDETECT_CONDA_ENV=OFF -DCUGRAPH_BUILD_WHEELS=ON -DCPM_cugraph-ops_SOURCE=/project/cugraph-ops/"
+  wheel-publish-pylibcugraph:
+    needs: wheel-build-pylibcugraph
+    secrets: inherit
+    uses: rapidsai/shared-action-workflows/.github/workflows/wheels-manylinux-publish.yml@main
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      repo: rapidsai/cugraph
```
ajschmidt8, line 81: we can remove this `repo` key from all of the `build.yaml` workflows (similarly to how I did for rmm [here](https://github.com/rapidsai/rmm/commit/7fb45d44c8ca4151581f5c57e413d1aea4b00f3f)).

This key was only ever used because we were using a `workflow_call` event to call this workflow from the https://github.com/rapidsai/actions/ repository for nightly builds.

Now that this file uses `workflow_dispatch` instead of `workflow_call`, this key isn't relevant anymore.

We should make this change for all jobs in this workflow and also all other wheel PRs.

ajschmidt8 (changes requested):

.github/workflows/pr.yaml
rlratzel: I just pushed commits to this PR that add `ci/wheel_smoke_test_pylibcugraph.py` and `ci/wheel_smoke_test_cugraph.py`, that I think will be adequate for what you described.  Thanks for starting this and let me know if you need anything else.

ajschmidt8 (approved):

rlratzel (approved):

ci/release/update-version.sh
```diff
@@ -85,8 +85,11 @@ sed_runner "s/extra-repo-sha: branch-.*/extra-repo-sha: branch-${NEXT_SHORT_TAG}
...
+
 # Wheel builds install intra-RAPIDS dependencies from same release
-sed_runner "s/${CURRENT_SHORT_TAG}/${NEXT_SHORT_TAG}/g" python/cugraph/setup.py
-sed_runner "s/${CURRENT_SHORT_TAG}/${NEXT_SHORT_TAG}/g" python/cugraph/_custom_build/backend.py
-sed_runner "s/${CURRENT_SHORT_TAG}/${NEXT_SHORT_TAG}/g" python/pylibcugraph/setup.py
-sed_runner "s/${CURRENT_SHORT_TAG}/${NEXT_SHORT_TAG}/g" python/pylibcugraph/_custom_build/backend.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/cugraph/setup.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/cugraph/_custom_build/backend.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/pylibcugraph/setup.py
+sed_runner "s/${CURRENT_SHORT_TAG_PEP440}/${NEXT_SHORT_TAG_PEP440}/g" python/pylibcugraph/_custom_build/backend.py
```
ajschmidt8, line 96: cross-posting the comment below since it's relevant here as well

- https://github.com/rapidsai/raft/pull/1200#discussion_r1089343751

BradReesWork (approved):

Issue #3079, opened 2022-12-14
[FEA] cuGraph-PyG Should Pass Canonical Edge Type to New Uniform Neighbor Sample Function


alexbarghi-nv:
Currently, cuGraph-PyG only passes the cuGraph edge type, but this adds unnecessary overhead.  Instead, the canonical edge type should be passed as an integer.  A lookup table may need to be added in `CuGraphStore` to handle this.

Partially resolved by #3101

PR #3101, opened 2023-01-03
[IMP] Speed Improvements for cuGraph-PyG (Short Circuit, Use Type Indices)

### LINKED ISSUE #3080: [FIX] cuGraph-PyG Repeats Dask Computations
Currently, cuGraph-PyG repeats dask computations due to not properly calling `persist()` when it should be called.  The `persist()` calls are missing from `_get_renumbered_edge_groups_from_sample()` and `_get_vertex_groups_from_sample()`.

### LINKED ISSUE #3099: cuGraph-PyG Performs Unnecessary Computations for Homogeneous Graphs
Currently, `cugraph-pyg` does type checking on homogeneous graphs, which is not necessary since all edge types are the same (due to there only being one edge type).  This should be bypassed for significantly faster performance.

### LINKED ISSUE #3100: cuGraph-PyG Can Use Indices to Store Edge Type Rather Than Edge ID
Currently, there is no reason to store edge ids using `indices` in `cugraph-pyg`.  As the new C++ API is not available yet, for a significant performance boost, the indices should be used to return edge type rather than edge id as is the current implementation.  This will save a large amount of calls to `get_edge_data` which is one of the worst perfoming `PropertyGraph` functions, especially in MG.

alexbarghi-nv:
This PR skips the type checking logic for homogeneous graphs, resulting in a significant performance improvement.
It also uses the existing sampling API to return types using "indices" instead of edge ids for another major performance gain.

- [x] Tests
- [x] Type annotations

Closes #3099 
Closes #3100 
Closes #3080 
Partially addresses #3079


VibhuJawa (approved):
LGTM

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -1760,7 +1766,11 @@ def edge_props_to_graph(
             # Ensure a valid edge_weight_property can be used for applying
             # weights to the subgraph, and if a default_edge_weight was
             # specified, apply it to all NAs in the weight column.
-            if edge_weight_property in edge_prop_df.columns:
+            if edge_weight_property == self.type_col_name:
```
rlratzel, line 1769: Can you update the comment above, since the code it's describing is different now? Same for the MG version.

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -1760,7 +1766,11 @@ def edge_props_to_graph(
...
+                prop_col = edge_prop_df[self.type_col_name].cat.codes.astype("float32")
+                edge_prop_df["temp_type_col"] = prop_col
+                edge_weight_property = "temp_type_col"
```
rlratzel, line 1772: probably unlikely, but can you change these new names to have leading underscores to minimize the chance of a collision with existing user columns? Same for the MG version.

python/cugraph/cugraph/structure/property_graph.py
```diff
@@ -501,6 +501,12 @@ def vertices_ids(self):
         """
         return self.get_vertices()
 
+    def vertex_types_from_numerals(self, nums):
+        return self.__vertex_prop_dataframe[self.type_col_name].dtype.categories[nums]
+
+    def edge_types_from_numerals(self, nums):
+        return self.__edge_prop_dataframe[self.type_col_name].dtype.categories[nums]
+
```
rlratzel, line 509: These should probably have docstrings since they're in the public API. Also, can you put `FIXME`s in the corresponding test files for these (or better yet, disabled empty tests).  Same for MG.

rlratzel (changes requested):
Some quick requests.

python/cugraph/cugraph/structure/property_graph.py
alexbarghi-nv: done

python/cugraph/cugraph/structure/property_graph.py
alexbarghi-nv: done

python/cugraph/cugraph/structure/property_graph.py
alexbarghi-nv: I did even better and wrote the full tests for those functions.

alexbarghi-nv: and docstrings, obviously

rlratzel (approved):

PR #3127, opened 2023-01-10
build.sh switch to use `RAPIDS` magic value

robertmaynard:
rapids-cmake 23.02 is deprecating the magic value of `ALL` since it doesn't cleanly map to the cmake magic value of `all`. Instead we use `RAPIDS` which better represents the architectures we are building for.


BradReesWork (approved):

rlratzel (approved):

BradReesWork:
rerun tests

BradReesWork:
rerun tests

PR #3148, opened 2023-01-17
Bulk Sampling

### LINKED ISSUE #3040: [FEA] Support Prefetching
cuGraph needs to support prefetching samples through `uniform_neighbor_sample` calls.  To start, this will use the current sampling function and multiple calls, but will update to process multiple batches in a single call.  There will also need to be changes to `cugraph-dgl` and `cugraph-pyg`, primarily a custom loader that sends batch information to the server.

The server will take in the batch information and prefetch up to a maximum prefetch size limit.  The buffer will be cleared and refilled as necessary.  There must be solid support for multiple DGL and PyG trainers, as well as PyTorch ddp.  Finally, although it is not specifically required by this story, #2876 is critical to accomplishing this with acceptable performance.

alexbarghi-nv:
Resolves #3040 
Merge after #3101 and #3082 

Also resolves two major bugs in MG sampling, one that results in an error when not passing a seed, and another that causes the sampling call to hang on MG.


python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,156 @@
...
+        self.__write(samples)
+
+    def __write(self, samples: Union[cudf.DataFrame, dask_cudf.DataFrame]) -> None:
+        # Ensure each rank writes to its own partition so there is no conflict
+        outer_partition = f"rank={self.__rank}"
+        if isinstance(samples, dask_cudf.DataFrame):
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_on=["batch_id", "hop_id"],
+            )
```
VibhuJawa, line 151: I suggest only partitioning on a single `batch_id` , I am afraid that we might end up with too small files which can  cause  bottlenecks on the IO layer.  

We might want to benchmark what is the fastest way to read each batch. I think we might have to have batches that are at least a few gigs to make this IO not be slow.  

We might have
```suggestion
            samples.to_parquet(
                os.path.join(self.__output_path, outer_partition),
                partition_on=["batch_id"],
            )
```

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,156 @@
...
+        else:
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_cols=["batch_id", "hop_id"],
```
VibhuJawa, line 155:
```suggestion
                partition_cols=["hop_id"],
```

VibhuJawa (changes requested):
Approach looks good but we have to benchmark the fastest scheme for IO reading.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
alexbarghi-nv: ok

VibhuJawa (approved):
The bulk loader part looks good to me  . 

Discussed off-line that we may require a second  optimization pass based on dataloader implementation.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,156 @@
...
+        outer_partition = f"rank={self.__rank}"
+        if isinstance(samples, dask_cudf.DataFrame):
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_on=["batch_id"],
+            )
+        else:
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_cols=["hop_id"],
```
VibhuJawa, line 155:
```suggestion
                partition_cols=["batch_id"],
```

VibhuJawa (changes requested):
Small bug in the writer

VibhuJawa (approved):

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,156 @@
...
+        if isinstance(samples, dask_cudf.DataFrame):
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_on=["batch_id"],
+            )
+        else:
+            samples.to_parquet(
+                os.path.join(self.__output_path, outer_partition),
+                partition_cols=["batch_id"],
+            )
```
VibhuJawa, line 156: No need to the check. 
```suggestion
        samples.to_parquet(
            os.path.join(self.__output_path, outer_partition),
            partition_cols=["batch_id"],
        )
```

alexbarghi-nv: This check is necessary.  For dask_cudf, the argument is `partition_on` but for cudf, the argument is `partition_cols`

VibhuJawa: My bad.Thanks for the clarification

alexbarghi-nv: We could raise this as an isssue to `cudf` since the inconsistency affects the user experimence

ChuckHastings (approved):
C++ changes look same as in #3082, let me know if anything is unique.  Otherwise I think it's fine since it should merge after 3082.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
VibhuJawa: Agreed, we should.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,156 @@
...
+    def flush(self) -> None:
+        """
+        Computes all uncomputed batches
+        """
+        end = min(self.__saturation_level, len(self.__batches))
+
+        sample_fn = (
+            cugraph.dask.uniform_neighbor_sample
+            if isinstance(self.__batches, dask_cudf.DataFrame)
+            else cugraph.uniform_neighbor_sample
```
VibhuJawa, line 125: We should do this based on what the Graph object is rather than the sampling list . 
```suggestion
            cugraph.dask.uniform_neighbor_sample
            if isinstance(self.__batches, dask_cudf.DataFrame)
            else cugraph.uniform_neighbor_sample
```
```suggestion
            cugraph.uniform_neighbor_sample
            if isinstance(self.__graph._plc_graph, pylibcugraph.graphs.SGGraph)
            else cugraph.dask.uniform_neighbor_sample
```

VibhuJawa (changes requested):
Minor change to do this based on graph object as we should support cudf batches for MGGraphs.

BradReesWork:
@alexbarghi-nv was did code coverage go down?

alexbarghi-nv:
> @alexbarghi-nv was did code coverage go down?

This PR is being overhauled right now.  An additional test will be added.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,204 @@
...
+            ix_partition_start_inclusive = (
+                min_batch_id + partition_k * self.batches_per_partition
+            )
+            ix_partition_end_inclusive = (
+                min_batch_id + (partition_k + 1) * self.batches_per_partition - 1
+            )
+
+            inner_path = os.path.join(
+                outer_partition_path,
+                f"batch={ix_partition_start_inclusive}-{ix_partition_end_inclusive}",
```
VibhuJawa, line 196:
```suggestion
            inner_path = os.path.join(
                outer_partition_path,
                f"batch={ix_partition_start_inclusive}-{ix_partition_end_inclusive}.parquet",
```

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,204 @@
...
+            )
+
+            f = (samples.batch_id >= ix_partition_start_inclusive) & (
+                samples.batch_id <= ix_partition_end_inclusive
+            )
+            samples[f].to_parquet(
+                inner_path,
```
VibhuJawa, line 203: So we also seem to be writing empty partitions here. I suggest having a check like 
```suggestion
           if len(samples[f])==0:
              break

            samples[f].to_parquet(
                inner_path,
```

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,204 @@
...
+        max_batch_id = min_batch_id + npartitions * self.batches_per_partition - 1
+        batch_id_filter = self.__batches[self.batch_col_name] <= max_batch_id
+
+        sample_fn = (
+            cugraph.uniform_neighbor_sample
+            if isinstance(self.__graph._plc_graph, pylibcugraph.graphs.SGGraph)
+            else cugraph.dask.uniform_neighbor_sample
+        )
+
+        # TODO semaphore check to prevent concurrent calls to uniform_neighbor_sample
```
VibhuJawa, line 160: Based on our discussion , we will use dask.lock in the sampling directly. 
```suggestion
```

VibhuJawa (changes requested):
Requested changes based on latest round of testing.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,204 @@
...
+                )
+
+        while self.size >= self.saturation_level:
+            self.flush()
+
+    def flush(self, skip_partition_size_check=False) -> None:
+        """
+        Computes all uncomputed batches
+        """
+        min_batch_id = self.__batches[self.batch_col_name].min()
```
VibhuJawa, line 138:
```suggestion
        if self.__batches is None:
                warnings.warn("Tried to flush with uncomputed batches"); 
                return 
        min_batch_id = self.__batches[self.batch_col_name].min()
```

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,204 @@
...
+            self.__graph,
+            **self.__sample_call_args,
+            start_list=self.__batches[self.start_col_name][batch_id_filter],
+            batch_id_list=self.__batches[self.batch_col_name][batch_id_filter],
+            with_edge_properties=True,
+        )
+
+        self.__batches = self.__batches[~batch_id_filter]
+        if len(self.__batches) == 0:
+            self.__batches = None
```
VibhuJawa, line 171: I think we will have to make this recursive. Based on testing we dont actually flush all the batches. 
```suggestion
        if len(self.__batches) == 0:
            self.__batches = None
        else:
           self.flush()
```

VibhuJawa (changes requested):

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
VibhuJawa, line 203:
```suggestion
                inner_path,
                index=False
```

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
```diff
@@ -0,0 +1,211 @@
...
+                )
+
+        if self.size >= self.saturation_level:
+            self.flush()
+
+    def flush(self, skip_partition_size_check=False) -> None:
+        """
+        Computes all uncomputed batches
+        """
+        if self.__batches is None:
```
VibhuJawa, line 139: I had to do something like below to make it work otherwise it sometimes gets stuck in recursive problems but dont have a MRE .  Do you have a test for this @alexbarghi-nv  ?  
```suggestion
        if self.__batches is None or len(self.__batches) == 0:
               self.__batches = None
```

VibhuJawa (changes requested):
A minor change to prevent recusions. Looks good otherwise.

python/cugraph/cugraph/gnn/data_loading/bulk_sampler.py
alexbarghi-nv: I have not encountered this problem while testing the loader.

alexbarghi-nv: but I have a slightly different fix for the bug

alexbarghi-nv: @VibhuJawa this should be resolved after the latest push.

VibhuJawa: So this [resolution](https://github.com/rapidsai/cugraph/pull/3148/commits/f51870e73e78131de32ff98a8003546f2045cc44) fails with 
```python
TypeError: object of type 'NoneType' has no len()
```
because  `self.__batches` can be None.

alexbarghi-nv: Ok, please try again now.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -30,6 +30,14 @@
...
 dst_n = "destinations"
 indices_n = "indices"
+weight_n = "weight"
+edge_id_n = "edge_id"
+edge_type_n = "edge_type"
+batch_id_n = "batch_id"
+hop_id_n = "hop_id"
+
+start_col_name = "_START_"
+batch_col_name = "_BATCH_"
```
VibhuJawa, line 40: I think this is a repetition. See line 42,43
```suggestion
```

VibhuJawa (approved):
This looks good to me based on testing.  Have requested a  minor changes but approving assuming that change is added.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -250,17 +273,21 @@ def uniform_neighbor_sample(
             weight_t=weight_t,
             with_edge_properties=with_edge_properties,
             # FIXME accept and properly transmute a numpy/cupy random state.
-            random_state=hash(random_state, i),
```
VibhuJawa, line 253: Just a comment that  catching bugs like this is why we need to get dask even on SG to run in CI ASAP. CC: @rlratzel

alexbarghi-nv:
pre-empted by #3170

PR #3156, opened 2023-01-18
Adding density algorithm and test

### LINKED ISSUE #3142: [FEA]: Add Density as a Graph function

New Feature

High

Add Graph Density as a function on the Graph

G.density()

none

### Additional context

N/A

### Code of Conduct

BradReesWork:
Added the density function

closes #3142 

Updating the docs will be a separate PR since there is a number of restructuring and clean-up in docs that needs to be done at the same time


python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -683,6 +683,27 @@ def add_nodes_from(self, nodes):
...
         self._Impl._nodes["all_nodes"] = cudf.Series(nodes)
 
+    def density(self):
+        """
+        Return the density of the graph.
+        Density is the measure of how many edges are in the graph versus
+        the max number of edges that could be present.
+        Note:  This function does not work on multigraph since there
+        is no theoretical max number of edges
+        """
```
rlratzel, line 693: This docstring will need to follow the format we use for public APIs, which includes an example, so it can be properly rendered by the API docs generating tools.  [Here's one](https://github.com/rapidsai/cugraph/blob/a529d762e18728223ac0f2f5c74eb3974ddcd7c0/python/cugraph/cugraph/structure/graph_classes.py#L618-L636) for reference.

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -683,6 +683,27 @@ def add_nodes_from(self, nodes):
         """
         self._Impl._nodes["all_nodes"] = cudf.Series(nodes)
 
+    def density(self):
```
rlratzel, line 686: We're slowly adding type annotations, starting with all new public APIs, so this will need an annotation too.

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -683,6 +683,27 @@ def add_nodes_from(self, nodes):
...
+        """
+        Return the density of the graph.
+        Density is the measure of how many edges are in the graph versus
+        the max number of edges that could be present.
+        Note:  This function does not work on multigraph since there
+        is no theoretical max number of edges
+        """
+        if self.is_multigraph():
+            raise TypeError("G cannot be a Multigraph")
```
rlratzel, line 695: This should be removed and the `MultiGraph` class should instead add a `density` method which raises this exception. That will also allow you to cleanup the docstring and mention the MultiGraph specifics there.

Also, the message shouldn't mention `G` since it's not used here and could be confusing (eg. a user has a variable named `G` in their code and they get this message).

python/cugraph/cugraph/tests/test_graph.py
```diff
@@ -739,3 +739,19 @@ def test_create_graph_with_edge_ids(graph_file):
     H = G.to_undirected()
     assert G.is_directed()
     assert not H.is_directed()
+
+
+# Test
+@pytest.mark.parametrize("graph_file", utils.DATASETS)
+def test_density(graph_file):
```
rlratzel, line 746: Might want to also add a check that `MultiGraph` instances raise `TypeError` if their `density()` method is called.

rlratzel (changes requested):

python/cugraph/cugraph/structure/graph_classes.py
BradReesWork: It will be updated to def density(self) -> float:

python/cugraph/cugraph/structure/graph_classes.py
BradReesWork: fixed

python/cugraph/cugraph/structure/graph_classes.py
BradReesWork: fixed

python/cugraph/cugraph/tests/test_graph.py
BradReesWork: added test

python/cugraph/cugraph/structure/graph_classes.py
BradReesWork: moved to MultiGraph section

rlratzel (approved):

Issue #2647, opened 2022-08-30
MNMG BC - Implementation


ChuckHastings:
Provide implementation for MNMG BC.

This Epic is focused on approximate BC, so the implementation should be based on MNMG BFS, as the current implementation is based on legacy BFS.

PR #3160, opened 2023-01-19
Implement Vertex betweenness centrality

ChuckHastings:
Implements Vertex Betweenness, part of #2647.  The core of this algorithm will also address Edge Betweenness, but the return values are different.  Edge betweenness will be finished in a separate PR.

We use a parallel variation of the Brandes Algorithm (2001) to compute exact or approximate betweenness, implemented using the graph primitives.


cpp/include/cugraph/graph_functions.hpp
```diff
@@ -875,4 +876,23 @@ weight_t compute_total_edge_weight(
...
+ * @param  select_count The number of vertices to select from the graph
+ * @param  with_replacement If true, select with replacement, if false select without replacement
+ * @return Device vector of selected vertices.
+ */
+template <typename vertex_t>
+rmm::device_uvector<vertex_t> select_random_vertices(raft::handle_t const& handle,
+                                                     raft::random::RngState& rng_state,
+                                                     vertex_t num_vertices,
+                                                     vertex_t select_count,
+                                                     bool with_replacement);
```
seunghwak, line 896: This is assuming that vertex IDs are consecutive integers starting from 0. This won't work once we start to support vertex masking. To be future proof, should we better take a graph_view object and call `number_of_vertices()` inside? Then, once we start to support vertex masking, we just need to update the function internals, and users don't need to make change in their side.

cpp/include/cugraph_c/centrality_algorithms.h
```diff
@@ -319,23 +314,18 @@ void cugraph_edge_centrality_result_free(cugraph_edge_centrality_result_t* resul
...
+ * Approximate betweenness can be computed specifying a list of vertices that should be
+ * used as seeds for the traversals.  Note that the function cugraph_select_random_vertices can be
+ * used to create a list of seeds.
  *
  * @param [in]  handle             Handle for accessing resources
  * @param [in]  graph              Pointer to graph
- * @param [in]  num_vertices       Number of vertices to randomly sample
  * @param [in]  vertex_list        Optionally specify a device array containing a list of vertices
- *                                 to use as seeds for BFS
+ *                                 to use as seeds for traversals
```
seunghwak, line 328: Better match the documentation in vertex betweenness centrality?

In the above, we're saying `to use as seeds for betweenness approximation`.

Users may be confused that `vertex_list` may have different roles in vertex & edge betweenness centrality.

cpp/src/structure/select_vertices.cpp
```diff
@@ -0,0 +1,74 @@
...
+                                                     bool with_replacement)
+{
+#ifndef NO_CUGRAPH_OPS
+  CUGRAPH_EXPECTS(with_replacement || (select_count <= num_vertices),
+                  "Requested more vertices than can be selected");
+
+  rmm::device_uvector<vertex_t> result(select_count, handle.get_stream());
+
+  if (select_count == num_vertices) {
+    detail::sequence_fill(handle.get_stream(), result.data(), result.size(), vertex_t{0});
```
seunghwak, line 44: I guess this is valid only when `with_replacement` = false.

cpp/src/structure/select_vertices.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#include <raft/core/handle.hpp>
+#include <rmm/device_scalar.hpp>
+#include <rmm/device_uvector.hpp>
+
+#ifndef NO_CUGRAPH_OPS
+#include <cugraph-ops/graph/sampling.hpp>
+#endif
+
+namespace cugraph {
```
seunghwak, line 28: May better rename this file to `select_random_vertices.cpp` unless you have a plan to add other non-random vertex selection functions?

seunghwak (commented):
Review part 1 (I haven't reviewed the actual betweenness centrality implementation yet).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+template <typename vertex_t>
+struct brandes_e_op_t {
+  const vertex_t max_vertex_t{std::numeric_limits<vertex_t>::max()};
+
+  template <typename value_t, typename ignore_t>
+  __device__ thrust::optional<value_t> operator()(
+    vertex_t, vertex_t, value_t src_property, vertex_t dst_property, ignore_t) const
+  {
+    thrust::optional<value_t> result{thrust::nullopt};
+    return (dst_property < max_vertex_t) ? result : thrust::make_optional(src_property);
```
seunghwak, line 52: I guess `return (dst_distance == invalid_distance)? thrust::make_optional(src_sigma) : thrust::nullopt` will be more readable.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+                      std::move(new_frontier),
+                      std::move(new_sigma),
+                      vertex_frontier,
+                      std::vector<size_t>{bucket_idx_next},
+                      thrust::make_zip_iterator(distance.begin(), sigma.begin()),
+                      thrust::make_zip_iterator(distance.begin(), sigma.begin()),
+                      [hop] __device__(auto v, auto old_values, auto v_sigma) {
+                        vertex_t next_distance = thrust::get<0>(old_values) == invalid_distance
+                                                   ? (hop + 1)
+                                                   : thrust::get<0>(old_values);
```
seunghwak, line 133: Is this check necessary? We're accumulating dst only when dst_distance == invalid_distance. Did you find any case thrust::get<0>(old_values) != invalid_distance?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+    distance.end(),
+    [] __device__(auto d) { return (d == invalid_distance) ? vertex_t{0} : d; },
+    vertex_t{0},
+    thrust::maximum<vertex_t>());
+
+  if constexpr (multi_gpu) {
+    diameter = host_scalar_allreduce(
+      handle.get_comms(), diameter, raft::comms::op_t::MAX, handle.get_stream());
+  }
+
```
seunghwak, line 175: You can use transform_reduce_v to merge the two (assuming that distance.size() == graph_view.local_vertex_partition_range_size()).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+      distance.begin(),
+      distance.end(),
+      [] __device__(auto d) { return (d == invalid_distance) ? vertex_t{0} : 1; },
+      vertex_t{0},
+      thrust::plus<vertex_t>());
+
+    if constexpr (multi_gpu) {
+      count = host_scalar_allreduce(
+        handle.get_comms(), count, raft::comms::op_t::SUM, handle.get_stream());
+    }
```
seunghwak, line 191: You can use count_if_v.

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: I'll work on this.  I hadn't thought about that particular use case, I had been thinking it might be better to pass in the vertex, but didn't really have a good use case to motivate it.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+//
+// The formula for BC(v) is the sum over all (s,t) where s != v != t of
+// sigma_st(v) / sigma_st.  Sigma_st(v) is the number of shortest paths
+// that pass through vertex v, whereas sigma_st is the total number of shortest
+// paths.
+namespace {
+
+template <typename vertex_t>
+struct brandes_e_op_t {
+  const vertex_t max_vertex_t{std::numeric_limits<vertex_t>::max()};
```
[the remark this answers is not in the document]
ChuckHastings: I will change this.  the `_t` was a typo, but I think `invalid_distance` is better.

cpp/src/c_api/random.cpp
```diff
@@ -13,13 +13,79 @@
...
+        local_vertices = cugraph::select_random_vertices(handle_,
+                                                         rng_state_->rng_state_,
+                                                         graph_view.number_of_vertices(),
+                                                         static_cast<vertex_t>(num_vertices_),
+                                                         false);
+      }
+
+      if constexpr (multi_gpu) {
+        local_vertices = cugraph::detail::shuffle_ext_vertices_to_local_gpu_by_vertex_partitioning(
+          handle_, std::move(local_vertices));
```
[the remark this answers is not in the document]
ChuckHastings: Yes.  I copied this from test code where the vertices generated were external.  I will consider the best way to fix this.

cpp/src/structure/select_vertices.cpp
ChuckHastings: Good point.  The only current use is `with_replacement == false`, I added the parameter for completeness.  I'll rework this.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+  // Do BFS with a multi-output.  If we're on hop k and multiple vertices arrive at vertex v,
+  // add all predecessors to the predecessor list, don't just arbitrarily pick one.
+  //
+  // Predecessors could be a CSR if that's helpful for doing the backwards tracing
+  constexpr vertex_t invalid_distance = std::numeric_limits<vertex_t>::max();
+  constexpr int bucket_idx_cur{0};
+  constexpr int bucket_idx_next{1};
+
+  rmm::device_uvector<vertex_t> sigma(graph_view.local_vertex_partition_range_size(),
+                                      handle.get_stream());
```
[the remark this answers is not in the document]
ChuckHastings: I will need to spend some time in the code to ponder the rest of your comments in this file.  This might need to be an `edge_t`, I'll work through the code and verify.

cpp/include/cugraph/graph_functions.hpp
jnke2016: The legacy implementation of betweenness has a similar python based [function](https://github.com/rapidsai/cugraph/blob/branch-23.02/python/cugraph/cugraph/centrality/betweenness_centrality.py#L282) which I am obviously going to substitute with the CAPI one from sampling_algorithms.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -335,9 +331,8 @@ void edge_betweenness_centrality(const raft::handle_t& handle,
  * @param graph_view Graph view object.
  * @param edge_weight_view Optional view object holding edge weights for @p graph_view. Currently,
  * edge_weight_view.has_value() should be false as we don't support weighted graphs, yet.
- * @param vertices Optional, if specified this provides either a vertex_t count of how many
- *         random seeds to select, or a device_span identifying a list of pre-selected vertices
- *         to use as seeds for the traversals for approximating betweenness.
+ * @param vertices Optional, if specified this provides a device_span identifying a list of
+ * pre-selected vertices to use as seeds for the traversals for approximating betweenness.
  * @param normalized         A flag indicating results should be normalized
```
jnke2016, line 336: It looks like this parameter is not working. I ran the CAPI tests with `normalized=True` and it is still passing

ChuckHastings: Added implementation of normalization in latest push

cpp/include/cugraph_c/centrality_algorithms.h
```diff
@@ -235,23 +236,18 @@ cugraph_error_code_t cugraph_katz_centrality(const cugraph_resource_handle_t* ha
 /**
  * @brief     Compute betweenness centrality
  *
- * Betweenness can be computed exactly by specifying num_vertices as 0 and
- * vertex_list as NULL.  This will compute betweenness centrality by doing a
- * traversal from every vertex and counting the frequency that a vertex appears on
- * a shortest path.
+ * Betweenness can be computed exactly by specifying This will compute betweenness centrality by
```
seunghwak, line 239: `specifying This`=>It seems like some words are missing between `specifying` and `This`.

cpp/src/centrality/betweenness_centrality_impl.cuh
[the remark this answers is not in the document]
seunghwak: ``` there will be at most number_of_vertices() paths```=>Is this true with multi-edges as well?

What happens if there are two vertices 0 and 1 and trillions of multi-edges between vertex 0 & 1?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -15,46 +15,368 @@
...
+  //
+  for (vertex_t d = diameter; d > 1; --d) {
+    per_v_transform_reduce_outgoing_e(
+      handle,
+      graph_view,
+      src_properties.view(),
+      dst_properties.view(),
+      cugraph::edge_dummy_property_t{}.view(),
+      [d] __device__(auto, auto, auto src_props, auto dst_props, auto) {
+        if ((thrust::get<0>(dst_props) == d) && (thrust::get<0>(src_props) == (d - 1))) {
```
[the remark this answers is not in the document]
seunghwak: Great, then could you update the FIXME comment above accordingly?

seunghwak (approved):
Few suggestions to consider, otherwise LGTM.

cpp/include/cugraph_c/centrality_algorithms.h
ChuckHastings: Cleaned up documentation

cpp/src/structure/select_random_vertices_impl.hpp
```diff
@@ -0,0 +1,76 @@
...
+      rmm::device_scalar<vertex_t> d_num_vertices(graph_view.number_of_vertices(),
+                                                  handle.get_stream());
+
+      cugraph::ops::gnn::graph::get_sampling_index(result.data(),
+                                                   rng_state,
+                                                   d_num_vertices.data(),
+                                                   1,
+                                                   select_count,
+                                                   with_replacement,
+                                                   handle.get_stream());
```
seunghwak, line 66: Just for a future reference, this will be executed using just a single warp, can be pretty slow if we're dealing with large graphs. Once we fix this problem, we can use this for seed generation in the sampling primitive & uniform sampling tests.

seunghwak (approved):

jnke2016 (approved):

Issue #3068, opened 2022-12-08
[FEA]: Address warnings and issues in notebooks and tests


BradReesWork:
### Is this a new feature, an improvement, or a change to existing functionality?

Change

### How would you describe the priority of this feature request

Low (would be nice)

### Please provide a clear description of problem this feature solves

Running notebooks, including in CI, is generating a number of warning messages that should be addressed

- [ ] cugraph/tests/test_louvain.py::test_louvain_csr_graph[False]

- [ ]   /opt/conda/envs/test/lib/python3.9/site-packages/cupy/sparse/__init__.py:17: DeprecationWarning: cupy.sparse is deprecated. Use cupyx.scipy.sparse instead.
    warnings.warn(msg, DeprecationWarning)

- [ ] cugraph/tests/test_betweenness_centrality.py: 144 warnings

- [ ]   /__w/cugraph/cugraph/python/cugraph/cugraph/tests/test_betweenness_centrality.py:163: DeprecationWarning: Sampling from a set deprecated
  since Python 3.9 and will be removed in a subsequent version.
    sources = random.sample(Gnx.nodes(), k)

- [ ] cugraph/tests/test_bfs.py: 26 warnings
  /__w/cugraph/cugraph/python/cugraph/cugraph/tests/test_bfs.py:280: DeprecationWarning: Sampling from a set deprecated
  since Python 3.9 and will be removed in a subsequent version.
    start_vertex = random.sample(Gnx.nodes(), 1)[0]

- [ ] cugraph/tests/test_bfs.py: 10 warnings

  /opt/conda/envs/test/lib/python3.9/site-packages/cugraph/traversal/bfs.py:108: DeprecationWarning: This function is deprecated in favor of cupy.from_dlpack
    distances = cp.fromDlpack(sorted_df["distance"].to_dlpack())

- [ ] cugraph/tests/test_bfs.py: 10 warnings

  /opt/conda/envs/test/lib/python3.9/site-packages/cugraph/traversal/bfs.py:109: DeprecationWarning: This function is deprecated in favor of cupy.from_dlpack
    preds = cp.fromDlpack(sorted_df["predecessor"].to_dlpack())

- [ ] cugraph/tests/test_compat_pr.py: 23 warnings

cugraph/tests/test_pagerank.py: 90 warnings
[...]

PR #3161, opened 2023-01-20
Update to support NetworkX 3.0 (and handle other deprecations)

### LINKED ISSUE #3084: Update NetworkX dependency meta-data to newer/latest version (3.0 as of 1/2023)
[This](https://github.com/rapidsai/cugraph/blob/de76e6436ee271f316c24b7546a6f480961170db/dependencies.yaml#L98) should be updated to the latest or at least a newer version of NetworkX

This requires re-running unit tests and notebooks to ensure compatibility.

(NetworkX is used for notebooks and testing, and (currently) an optional experimental module. It is also optionally imported in other parts of cugraph)

eriknw:
Closes #3084 and mostly takes care of #3068 (I didn't update `PendingDeprecationWarning`s that originate from `cugraph`).

Old versions of NetworkX should continue to work.

I ran the tests (and many--but not all--of the notebooks) with `-Werror::DeprecationWarning` to uncover issues.

`tornado` warns `DeprecationWarning: make_current is deprecated; start the event loop first`, which originate from their code. They deprecated it, but didn't update their code yet, so I updated the tests to ignore it.

### DIFF
--- a/conda/environments/all_cuda-115_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-115_arch-x86_64.yaml
@@ -30,7 +30,7 @@ dependencies:
 - librmm=23.02.*
 - nbsphinx
 - nccl>=2.9.9
-- networkx>=2.5.1,<3.0
+- networkx>=2.5.1
 - ninja
 - notebook>=0.5.0
 - numpydoc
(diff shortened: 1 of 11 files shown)


rlratzel (approved):
LGTM

ajschmidt8 (approved):

PR #3168, opened 2023-01-23
Data Generator

BradReesWork:
The generator mimics the OGB mag format. 

execution example:

python gnn_cite_graph_gen.py \
    -mg True\
    -outdir '/tmp/GNN2' \
     -papersScale 12  \
     -format "csv" \
     -papersLabeledPercent 0.10 \
     -papersFeatureNoise 0.0001 

these uses default of 10 features (-papersNumFeatures)


benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
```diff
@@ -0,0 +1,639 @@
...
+    parser.add_argument(
+        "-papersScale",
+        type=int,
+        default=None,
+        help="Scale factor of the Papers-Cite_Papers graph",
+        required=True,
+    )
+
+    parser.add_argument(
+        "-papersEgefactor",
```
alexbarghi-nv, line 511: typo in "EdgeFactor"

alexbarghi-nv:
@BradReesWork I keep getting this error: Size of output exceeds column size limit

Maybe we need to convert to host before writing CSV data?

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
```diff
@@ -0,0 +1,639 @@
...
+
+def save_data(args, dir, file_name, data, mg=False):
+
+    if mg:
+        f = os.path.join(dir, f"{file_name}")
+    else:
+        f = os.path.join(dir, f"{file_name}.{args.format}")
+    print(f"\tsaving to {f}")
+
+    if format == "parquet":
```
alexbarghi-nv, line 137:
```suggestion
    if args.format == "parquet":
```

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
```diff
@@ -0,0 +1,639 @@
...
+
+    # selected just the labeld nodes
+    labeled_df = (nodes["label"].loc[nodes["label"] == 1]).to_frame().reset_index()
+
+    # drop the "labeled" column since it is no longer needed
+    labeled_df.drop(columns=["label"], inplace=True)
+
+    # Save the list
+    dir = args.outdir + paper_dir
+    name = "node-label." + args.format
```
alexbarghi-nv, line 283:
```suggestion
    name = "node-label"
```

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
```diff
@@ -0,0 +1,639 @@
...
+    save_data(args, dir, name, labeled_df)
+
+    # some cleanup
+    del ran
+    del labeled_df
+    nodes.drop(columns=["rand"], inplace=True)
```
alexbarghi-nv, line 289:
```suggestion
    nodes.drop(columns=["rand", "label"], inplace=True)
```

alexbarghi-nv:
Could you please also output a JSON containing the number of nodes of each type and the starting offsets?

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
BradReesWork: oops.  I changed the input argument and missed this one

VibhuJawa (commented):
Looks good to me . Thanks for working on this Brad

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
BradReesWork: in next push

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
BradReesWork: in next push

benchmarks/shared/python/data_generators/gnn_cite_graph_gen.py
BradReesWork: oops, fixed

BradReesWork:
> 
You need to save a parquet in that case.  Seems like the CSV writer is arrow based, so no more than 2B rows

BradReesWork:
"Could you please also output a JSON containing the number of nodes of each type and the starting offsets?"

that can be done.  currently there is no starting offset since all vertex IDs are contiguous.   The function accept a starting offset as input, that could be 0 (zero) and an offset JSON file produced

alexbarghi-nv:
By offsets, I mean the starting vertex ID for each vertex type

alexbarghi-nv:
preempted by #3223

PR #3169, opened 2023-01-23
Performance tuning the sampling primitive for multi-node multi-GPU systems.

seunghwak:
* Update groupby code in multi-GPU communication to use atomics based partitioning instead of sort based partitioning (with atomics performance updates in recent NVIDIA GPUs, now the atomics based approach is significantly faster than the sorting based approach if the number of groups is not excessive).
* In random index generation, add an additional code to handle high-degree vertices with with_replacement = false.


ChuckHastings (approved):

rlratzel (approved):
(approving for python-codeowners to unblock merging, but didn't review any code here - assuming python-codeowners added from a file change no longer in this PR)

seunghwak:
> (approving for python-codeowners to unblock merging, but didn't review any code here - assuming python-codeowners added from a file change no longer in this PR)

Yes, this is due to branch re-targeting.

PR #3174, opened 2023-01-24
Pin gcc to 9.x.

vyasr:
Later versions of gcc are not yet supported.


ajschmidt8:
@vyasr, just wondering why this pin is needed for `cugraph` even though we don't have it in other repositories

vyasr:
We do have it in both [cudf](https://github.com/rapidsai/cuml/blob/branch-23.02/dependencies.yaml#L82) and [cuml](https://github.com/rapidsai/cuml/blob/branch-23.02/dependencies.yaml#L82). However, you're right that we don't have it consistently in every repository. My recollection is that the gcc 11 incompatibilities are only triggered by certain code paths, so we have probably only added the pins in repositories that hit those code paths.

ajschmidt8 (approved):

PR #3184, opened 2023-01-26
[FEA] Enable using cugraph uniform sampling in multi client environments

VibhuJawa:
In multi client environments our sampling algorithms tend to hang, this PR adds a lock to prevent that from happening.  This PR closes https://github.com/rapidsai/cugraph/issues/3186

See [notebooks](https://gist.github.com/VibhuJawa/1b705427f7a0c5a2a4f58e0a3e71ef21) here with MRE to reproduce. 

Before PR (HANG): 
Screen Shot 2023-01-25 at 9 01 39 PM
After PR (NO HANG): 
Screen Shot 2023-01-25 at 9 07 02 PM

I have not added tests for this here because mocking dask with multi client environment is not possible in CI yet so we will do this as future work. 

Also includes a bug fix in our sampling code path.

### DIFF
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -156,7 +156,8 @@ def uniform_neighbor_sample(
         Random seed to use when making sampling calls.
 
     _multiple_clients: bool, optional (default=False)
-        internal flag to ensure sampling works with multiple clients
+        internal flag to ensure sampling works with multiple dask clients
+        set to True to prevent hangs in multi-client environment
 
     Returns
     -------
@@ -189,6 +190,9 @@ def uniform_neighbor_sample(
                 Contains the hop ids from the sampling result
     """
     if _multiple_clients:
+        # Distributed centralized lock to allow
+        # two disconnected processes to coordinate a lock
+        # https://docs.dask.org/en/stable/futures.html?highlight=lock#distributed.Lock
         lock = Lock("sampling_mg_lock")
         lock.acquire(timeout=1)
 
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import numpy
-from dask.distributed import wait
+from dask.distributed import wait, Lock
 from cugraph.dask.common.input_utils import get_distributed_data
 
 
@@ -119,6 +119,7 @@ def uniform_neighbor_sample(
     with_edge_properties=False,
     batch_id_list=None,
     random_state=None,
+    _multiple_clients=False,
 ):
     """
     Does neighborhood sampling, which samples nodes from a graph based on the
@@ -154,6 +155,9 @@ def uniform_neighbor_sample(
     random_state: int, optional
         Random seed to use when making sampling calls.
 
+    _multiple_clients: bool, optional (default=False)
+        internal flag to ensure sampling works with multiple clients
+
     Returns
     -------
     result : dask_cudf.DataFrame
@@ -184,6 +188,10 @@ def uniform_neighbor_sample(
             df['hop_id']: dask_cudf.Series
                 Contains the hop ids from the sampling result
     """
+    if _multiple_clients:
+        lock = Lock("sampling_mg_lock")
+        lock.acquire(timeout=1)
+
     if isinstance(start_list, int):
         start_list = [start_list]
 
@@ -238,7 +246,7 @@ def uniform_neighbor_sample(
     client = input_graph._client
 
     session_id = Comms.get_session_id()
-
+    random_state = hash(None)
     result = [
         client.submit(
             _call_plc_uniform_neighbor_sample,
@@ -250,7 +258,7 @@ def uniform_neighbor_sample(
             weight_t=weight_t,
             with_edge_properties=with_edge_properties,
             # FIXME accept and properly transmute a numpy/cupy random state.
-            random_state=hash(random_state, i),
+            random_state=hash((random_state, i)),
             workers=[w],
             allow_other_workers=False,
             pure=False,
@@ -267,5 +275,8 @@ def uniform_neighbor_sample(
     if input_graph.renumbered:
         ddf = input_graph.unrenumber(ddf, "sources", preserve_order=True)
         ddf = input_graph.unrenumber(ddf, "destinations", preserve_order=True)
+
+    if _multiple_clients:
+        lock.release()
 
     return ddf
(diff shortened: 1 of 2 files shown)


alexbarghi-nv:
That bug is also fixed in 3148

alexbarghi-nv:
Could you add some documentation?  Is setting `_multi_clients=True` sufficient or is there something else that needs to be done in addition?

alexbarghi-nv:
Just a code comment is probably sufficient, maybe link to dask documentation.

VibhuJawa:
> Just a code comment is probably sufficient, maybe link to dask documentation.

Done, added both.

alexbarghi-nv (approved):
👍

jnke2016 (approved):

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -268,4 +279,7 @@ def uniform_neighbor_sample(
         ddf = input_graph.unrenumber(ddf, "sources", preserve_order=True)
         ddf = input_graph.unrenumber(ddf, "destinations", preserve_order=True)
 
+    if _multiple_clients:
+        lock.release()
```
rlratzel, line 283: Could the lock be released earlier to improve parallelism slightly, like right before the unrenumbering above (esp. since I think unrenumbering could be expensive)?

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -184,6 +189,13 @@ def uniform_neighbor_sample(
             df['hop_id']: dask_cudf.Series
                 Contains the hop ids from the sampling result
     """
+    if _multiple_clients:
+        # Distributed centralized lock to allow
+        # two disconnected processes (clients) to coordinate a lock
+        # https://docs.dask.org/en/stable/futures.html?highlight=lock#distributed.Lock
+        lock = Lock("sampling_mg_lock")
+        lock.acquire(timeout=1)
```
rlratzel, line 197: If you don't get the lock and the timeout expires, should we do something? In this case, if the timeout expires it will run sampling anyway - is that okay?

rlratzel, line 197: The lock isn't actually needed until the `client.submit()` call that runs sampling below, is that correct?  I'm just wondering if the acquire should be moved closer to that call to maximize parallelism by holding the lock as little as possible.

rlratzel (commented):
This is great, thanks. Just had a few questions below.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: @jnke2016 , Do we use PLC primitives to unrenumbe, I believe those are not multi-client proof so i dont think that will safe. 

If it is a pure dask operation we can do it safely.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Again, I am not sure how safe our renumbering is . I can add it and release it across a bunch of places if that helps.

VibhuJawa: I think we can increase the timeout further , i think not running sampling will be problematic. How about we timeout at `50s` ? I think that should be enough .  The max i have seen it is 4s for a few million nodes at 1B scale. 

It is obviously a heuristic.

eriknw: I share Rick's sentiment here. As someone who isn't super-familiar with the sampling code, it's unclear to me exactly which resources are being protected from the lock and which operations depend on those resources.

Also, what about possible exceptions between `lock.acquire` and `lock.release`? I think it would be best practice to use it as a context manager like `with Lock("sampling_mg_lock"): ...`

rlratzel (changes requested):
marking as "request changes" to prevent accidental merging while past feedback is being addressed.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
rlratzel: I believe unrenumbering does not use PLC and is pure cudf/dask_cudf code, but @jnke2016 should also comment.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
eriknw: Alright, so if I understand correctly, this lock is used to ensure that a dask cluster will only perform sampling for one dataset at a time, right? Even if two clients want to sample from two different graphs, they won't be able to sample at the same time. To reiterate my above thoughts, I think _all_ locks should indicate what resources they are protecting, even if it's as broad as "cluster usage".

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
jnke2016: @rlratzel correct: `unrenumbering` is just pure cudf/dask_cudf

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: > Alright, so if I understand correctly, this lock is used to ensure that a dask cluster will only perform sampling for one dataset at a time, right? Even if two clients want to sample from two different graphs, they won't be able to sample at the same time. 

Yes, my current guess of the problem is that our PLC layer is not multi-client safe so if we submit tasks to that layer we seem to hang.   Tried to address those with the latest push . 

Thanks for the feeback @eriknw  and @rlratzel . All great points that i agree with. Thanks for catching those subtleties. Please take a look and see if the last push solves all the raised points.   Just for completeness, i tested it with new changes and it seems to be working.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Thanks pushed latests changes with this assumption.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Raised this with the last commit to 100.

eriknw: LGTM! 👍

VibhuJawa:
@rlratzel , All changes should be addressed now. Lets try to get this in sooner as https://github.com/rapidsai/cugraph/issues/3186 builds on this.

rlratzel (approved):
Thanks!

PR #3189, opened 2023-01-26
Fix off-by-one in start_dask_client

### LINKED ISSUE #3187: [QST]: multi-GPU dask cluster setup
### What is your question?

Since #3082, multi-GPU tests attempt to start workers on devices 1 and 2 by default (rather than 0 and 1), since `cugraph.dask.common.start_dask_client`

https://github.com/rapidsai/cugraph/blob/5c021fb21230f256abf723a977aaecc24a315bf7/python/cugraph/cugraph/dask/common/mg_utils.py#L104-L117

Has a (deliberate?) off-by-one in selecting the relevant devices.

In contrast, the previous usage (from `cugraph.testing.mgutils.start_dask_client`) would usually pass `dask_worker_devices=None` to `LocalCUDACluster`, which results in selecting all the devices on the system (indexed from zero).

https://github.com/rapidsai/cugraph/blob/5c021fb21230f256abf723a977aaecc24a315bf7/python/cugraph/cugraph/testing/mg_utils.py#L59-L64

This seems like it is just a bug, but might have been done deliberately?
### Code of Conduct

- [X] I have searched the [open issues](https://github.com/rapidsai/cugraph/issues) and have found no duplicates for this question

wence-:
PR #3082 introduced a new version of start_dask_client, but requests devices [1, N], rather than [0, N) as the previous (still extant in testing/mg_utils) version gave us.

Closes #3187.

### DIFF
--- a/python/cugraph/cugraph/dask/common/mg_utils.py
+++ b/python/cugraph/cugraph/dask/common/mg_utils.py
@@ -105,7 +105,7 @@ def start_dask_client():
     n_devices = os.getenv("DASK_NUM_WORKERS", 2)
     n_devices = int(n_devices)
 
-    visible_devices = ",".join([str(i) for i in range(1, n_devices + 1)])
+    visible_devices = ",".join(map(str, range(n_devices)))
 
     cluster = LocalCUDACluster(
         protocol="ucx", rmm_pool_size="25GB", CUDA_VISIBLE_DEVICES=visible_devices


wence-:
cc @alexbarghi-nv is the 1-based indexing in the new `start_dask_client` a deliberate choice?

alexbarghi-nv:
It is intentional - this setup is designed for a large cluster, like the ones we run our CI on.  GPU 0 is reserved for other operations, while the graph is stored on GPUs 1-x.  But I have been thinking about whether we should have an alternative to this setup script.

alexbarghi-nv:
I think we should swap this script and the one in `testing/mg_utils`, and make the one in mg utils the one that's used for running CI tests.

wence-:
> It is intentional - this setup is designed for a large cluster, like the ones we run our CI on. GPU 0 is reserved for other operations, while the graph is stored on GPUs 1-x. But I have been thinking about whether we should have an alternative to this setup script.

Ah thanks, make sense.

VibhuJawa:
@wence, Closing this based on https://github.com/rapidsai/cugraph/issues/3187#issuecomment-1405285182

wence-:
> @wence, Closing this based on [#3187 (comment)](https://github.com/rapidsai/cugraph/issues/3187#issuecomment-1405285182)

Thanks!

PR #3192, opened 2023-01-27
[WIP] Make graph objects serializable

VibhuJawa:
This PR makes `cugraph.structure.graph_classes` dask serializable and de-serializable.  This means that we can cleanly use dask to transmit graph objects b/w different clients without much work and using all ready supported API. 

See docs here: https://distributed.dask.org/en/stable/publish.html

**On client 0**
```python3
print(type(g))
client.publish_dataset(mg_graph_for_sampling=g)
```
```
<class 'cugraph.structure.graph_classes.MultiGraph'>
```

**On client 1** 
```python3
g = client_worker_1.get_dataset('mg_graph_for_sampling')
type(g)
```
```
cugraph.structure.graph_classes.MultiGraph
```

CC: @rlratzel

### DIFF
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -156,7 +156,8 @@ def uniform_neighbor_sample(
         Random seed to use when making sampling calls.
 
     _multiple_clients: bool, optional (default=False)
-        internal flag to ensure sampling works with multiple clients
+        internal flag to ensure sampling works with multiple dask clients
+        set to True to prevent hangs in multi-client environment
 
     Returns
     -------
@@ -189,6 +190,9 @@ def uniform_neighbor_sample(
                 Contains the hop ids from the sampling result
     """
     if _multiple_clients:
+        # Distributed centralized lock to allow
+        # two disconnected processes to coordinate a lock
+        # https://docs.dask.org/en/stable/futures.html?highlight=lock#distributed.Lock
         lock = Lock("sampling_mg_lock")
         lock.acquire(timeout=1)
 
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import numpy
-from dask.distributed import wait
+from dask.distributed import wait, Lock
 from cugraph.dask.common.input_utils import get_distributed_data
 
 
@@ -119,6 +119,7 @@ def uniform_neighbor_sample(
     with_edge_properties=False,
     batch_id_list=None,
     random_state=None,
+    _multiple_clients=False,
 ):
     """
     Does neighborhood sampling, which samples nodes from a graph based on the
@@ -154,6 +155,9 @@ def uniform_neighbor_sample(
     random_state: int, optional
         Random seed to use when making sampling calls.
 
+    _multiple_clients: bool, optional (default=False)
+        internal flag to ensure sampling works with multiple clients
+
     Returns
     -------
     result : dask_cudf.DataFrame
@@ -184,6 +188,10 @@ def uniform_neighbor_sample(
             df['hop_id']: dask_cudf.Series
                 Contains the hop ids from the sampling result
     """
+    if _multiple_clients:
+        lock = Lock("sampling_mg_lock")
+        lock.acquire(timeout=1)
+
     if isinstance(start_list, int):
         start_list = [start_list]
 
@@ -238,7 +246,7 @@ def uniform_neighbor_sample(
     client = input_graph._client
 
     session_id = Comms.get_session_id()
-
+    random_state = hash(None)
     result = [
         client.submit(
             _call_plc_uniform_neighbor_sample,
@@ -250,7 +258,7 @@ def uniform_neighbor_sample(
             weight_t=weight_t,
             with_edge_properties=with_edge_properties,
             # FIXME accept and properly transmute a numpy/cupy random state.
-            random_state=hash(random_state, i),
+            random_state=hash((random_state, i)),
             workers=[w],
             allow_other_workers=False,
             pure=False,
@@ -267,5 +275,8 @@ def uniform_neighbor_sample(
     if input_graph.renumbered:
         ddf = input_graph.unrenumber(ddf, "sources", preserve_order=True)
         ddf = input_graph.unrenumber(ddf, "destinations", preserve_order=True)
+
+    if _multiple_clients:
+        lock.release()
 
     return ddf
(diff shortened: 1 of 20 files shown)


python/cugraph/cugraph/tests/mg/test_mg_graph.py
```diff
@@ -218,12 +218,19 @@ def test_create_graph_with_edge_ids(dask_client, graph_file):
...
         )
 
     G = cugraph.Graph(directed=True)
     G.from_dask_cudf_edgelist(
-        el, source="0", destination="1", edge_attr=["2", "id", "etype"]
+        el,
+        source="0",
+        destination="1",
+        edge_attr=["2", "id", "etype"],
+        legacy_renum_only=True,
```
VibhuJawa, line 233: Another example of why we need CI in dask . CC: @rlratzel

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -81,7 +81,9 @@ def __init__(self, m_graph=None, directed=False):
                 )
 
     def __getattr__(self, name):
-        if self._Impl is None:
+        # accessing `_Impl` will lead to recursion
```
rlratzel, line 84: Ah, thanks, we've never tested this with `deepcopy` before so that's not too surprising.  Can you also add this as the docstring, since this code and using \_\_getattr\_\_() this way is not intuitive IMO:
```suggestion
        """
        __getattr__() is called automatically by python when an attribute does not
        exist. Since this class is attempting to hide the internal `_Impl` object,
        which is intended to contain many of the attributes needed by this class,
        __getattr__ is used to "pass through" attribute access to _Impl and make it
        appear as if the _Impl attributes are contained in this class.
        """
```

rlratzel (changes requested):
Thanks, this is great. Can you also make a couple of minor changes below?

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -184,6 +189,13 @@ def uniform_neighbor_sample(
             df['hop_id']: dask_cudf.Series
                 Contains the hop ids from the sampling result
     """
+    if _multiple_clients:
+        # Distributed centralized lock to allow
+        # two disconnected processes (clients) to coordinate a lock
+        # https://docs.dask.org/en/stable/futures.html?highlight=lock#distributed.Lock
+        lock = Lock("sampling_mg_lock")
+        lock.acquire(timeout=1)
```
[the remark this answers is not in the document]
VibhuJawa: Yup, this PR build on top of 3184 so will just push the changes there.

alexbarghi-nv:
@VibhuJawa what is the correct way to wait for the graph object to become available on ranks 1-n?  Is it acceptable to just keep checking if it exists and sleep for x time if it doesn't?  Or like were you saying, should there be another lock?

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -81,7 +81,13 @@ def __init__(self, m_graph=None, directed=False):
...
     def __getattr__(self, name):
-        # accessing `_Impl` will lead to recursion
+        """
+        __getattr__() is called automatically by python when an attribute does not
+        exist. Since this class is attempting to hide the internal `_Impl` object,
+        which is intended to contain many of the attributes needed by this class,
+        __getattr__ is used to "pass through" attribute access to _Impl and make it
+        appear as if the _Impl attributes are contained in this class.
+        """
         # so avoiding that
```
rlratzel, line 91: oops, probably want to delete this line now.

rlratzel (approved):
just noticed a stray comment which can be removed now, but I'm approving ahead of time.

VibhuJawa:
> @VibhuJawa what is the correct way to wait for the graph object to become available on ranks 1-n? Is it acceptable to just keep checking if it exists and sleep for x time if it doesn't? Or like were you saying, should there be another lock?

So `client.list_datasets` is the if you want to check the existence.  I have been thinking of using events on ranks `1-n` to just make sure that this case does not happen.  (https://docs.dask.org/en/stable/futures.html?highlight=semaphore#events  
)

On rank  0
```python3
from dask.distributed import Event

event = Event("my-event-1")

# create dataset
graph_store =  create_gs (...)
client. publish_dataset(graph_store)
event.set()
``` 

On rank 1:-n 
```python3
# One one client
from dask.distributed import Event

event = Event("my-event-1")
event.wait()
graph_store = client.get_dataset(graph_store)
```

alexbarghi-nv:
> > @VibhuJawa what is the correct way to wait for the graph object to become available on ranks 1-n? Is it acceptable to just keep checking if it exists and sleep for x time if it doesn't? Or like were you saying, should there be another lock?
> 
> So `client.list_datasets` is the if you want to check the existence. I have been thinking of using events on ranks `1-n` to just make sure that this case does not happen. (https://docs.dask.org/en/stable/futures.html?highlight=semaphore#events )

Ah, ok, this is a good solution that I can use without changing the basic workflow.  I just have to make sure the graph store object is publishable...

VibhuJawa:
> > > @VibhuJawa what is the correct way to wait for the graph object to become available on ranks 1-n? Is it acceptable to just keep checking if it exists and sleep for x time if it doesn't? Or like were you saying, should there be another lock?
> > 
> > 
> > So `client.list_datasets` is the if you want to check the existence. I have been thinking of using events on ranks `1-n` to just make sure that this case does not happen. (https://docs.dask.org/en/stable/futures.html?highlight=semaphore#events )
> 
> Ah, ok, this is a good solution that I can use without changing the basic workflow. I just have to make sure the graph store object is publishable...

Yup.  Let me know if you run into issues , happy to help.

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -81,7 +81,9 @@ def __init__(self, m_graph=None, directed=False):
                 )
 
     def __getattr__(self, name):
-        if self._Impl is None:
```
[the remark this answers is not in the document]
eriknw: Boy am I glad you explained this. Your implementation is fine/works. Another implementation would be:
```python
        if self.__getattribute__("_Impl") is None:
```

python/cugraph/cugraph/structure/graph_classes.py
VibhuJawa: Done

python/cugraph/cugraph/structure/graph_classes.py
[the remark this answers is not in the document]
VibhuJawa: I want to stick with `if name == "_Impl"`  just cause thinking of python internal differences and order of `__getattribute__ ` and `__getattr__ ` makes my head hurt .   I explored this option earlier but decided against it for clarity sakes. 

Please let me know if we want to change that.

VibhuJawa:
@eriknw , @rlratzel . Thanks a lot for the reviews.  All  the feedback has been addressed  . Lets get this in soon.

VibhuJawa:
@BradReesWork , I dont this is blocked as the underlying PR is merged . Can we please remove the blocked label.

BradReesWork:
rerun tests

eriknw (approved):
LGTM 🚀

PR #3212, opened 2023-02-01
[REVIEW]Multi-trainers cugraph-DGL examples

VibhuJawa:
This PR adds a working Multi-GPU Graph (on 2 dask workers)  being trained/loaded on multiple pytorch trainers.  (3)

Todo: 
- [x] Verify works on multiple trainers and multiple dask workers
- [x] Show scaling as you increase training GPUs 

At 1 second we become bottlenecked by sampling dask cluster, but we see perf improvement by going from `1 GPU`->`2GPU`.   
**On OBGN-Products**
```md
| Number of Training GPUs | Time per epoch |
|-------------------------|----------------|
| 1                       | 2.3 s          |
| 2                       | 0.582 s        |
| 4                       | 0.792 s        |
```

This PR depends upon:  https://github.com/rapidsai/cugraph/pull/3393
CC: @rlratzel , @alexbarghi-nv , @BradReesWork


alexbarghi-nv:
@VibhuJawa is this is still WIP?  I thought you said this was ready for review.

VibhuJawa:
> @VibhuJawa is this is still WIP? I thought you said this was ready for review.

Sorry for the confusion. This is still WIP. Will let you know when it is ready for review

VibhuJawa:
Please note that this example is failing because of a bug that i am trying to triage.  That said, the bug is independent of the PR so we can probably review it.

python/cugraph-dgl/examples/muti_trainer_MG_example/workflow.py
```diff
@@ -0,0 +1,233 @@
...
+    print(f"Warmup loading took = {et-st} s on worker = {rank}")
+    torch.distributed.barrier()
+
+    n_epochs = 30
+    total_st = time.time()
+    for i in range(0, n_epochs):
+        st = time.time()
+        for step, (input_nodes, seeds, blocks) in enumerate(dataloader):
+            pass
+            # print(len(input_nodes))
```
alexbarghi-nv, line 201: I notice you're not actually doing model training here, that's intentional, right?

python/cugraph-dgl/examples/muti_trainer_MG_example/workflow.py
```diff
@@ -0,0 +1,233 @@
...
+    # cleanup dask cluster
+    if rank == 0:
+        client.unpublish_dataset("cugraph_gs")
+        client.unpublish_dataset("train_idx")
+        client.unpublish_dataset("valid_idx")
+        client.unpublish_dataset("test_idx")
+        event.clear()
+
+
+if __name__ == "__main__":
```
alexbarghi-nv, line 222: Any way we could do this just with 2 GPUs, since that cover more user environments?  Like my PyG example, I think it's ok for the GPUs to overlap (although since you're using rmm pool that could make that more complicated).

alexbarghi-nv (approved):
Looks ok, just had a couple comments that don't need to hold up merging this PR.

python/cugraph-dgl/examples/muti_trainer_MG_example/workflow.py
alexbarghi-nv: ok

python/cugraph-dgl/examples/muti_trainer_MG_example/workflow.py
alexbarghi-nv: fine with me

PR #3250, opened 2023-02-07
Remove legacy mg bfs

ChuckHastings:
No longer used.


seunghwak (approved):

cpp/src/traversal/legacy/bfs.cu
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022, NVIDIA CORPORATION.  All rights reserved.
+ * Copyright (c) 2019-2023, NVIDIA CORPORATION.  All rights reserved.
```
seunghwak, line 2: Just for double checking, can't we delete this file as well?

PR #3258, opened 2023-02-09
Reduce error handling verbosity in CI tests scripts

AjayThorve:
This PR adds a less verbose [trap method](https://github.com/rapidsai/cugraph/blob/f2b081075704aabc789603e14ce552eac3fbe692/ci/test.sh#L19), for error handling to help ensure that we capture all potential error codes in our test scripts. It's already being used in the `ci/test.sh` and `ci/test_notebooks.sh` scripts, this PR extends the usage to rest of the test scripts.

cc @ajschmidt8


ci/test_cpp.sh
```diff
@@ -38,6 +37,8 @@ pushd "${RAPIDS_DATASET_ROOT_DIR}"
 ./get_test_data.sh
 popd
 
+EXITCODE=0
+trap "EXITCODE=1" ERR
```
ajschmidt8, line 41: Can you rearrange these lines a bit:

- Consolidate the `EXITCODE` lines
- Move the `set +e` below the `EXITCODE` lines for consistency with the other repositories
- Remove the `TODO` a few lines down
- Put the `rapids-logger` statement right before the subsequent `for` loop

ajschmidt8 (changes requested):

ci/test_cpp.sh
AjayThorve: done, thanks!

ajschmidt8 (approved):
:fire: :fire:

PR #3261, opened 2023-02-09
Strongly connected components Part 1 (trimming zero in|out degree vertices)

wu6u3:


cpp/src/cores/trim.cu
```diff
@@ -0,0 +1,39 @@
...
+  auto out_degree = graph_view.compute_out_degree();
+ 
+  rmm::device_uvector<vertex_t> remaining_vertices(graph_view.local_vertex_partition_range_size(),  handle.get_stream());
+
+  remaining_vertices.resize(
+    thrust::distance( remaining_vertices.begin(), thrust::copy_if(
+        handle.get_thrust_policy(), thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first()),
+        thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last()),  remaining_vertices.begin(),
+        [ in_degree ] __device__(
+          auto in_degree) { return in_degree <= 1; })),
```
seunghwak, line 34: `thrust::distance` takes `(first, last)` (instead of `(last, first)`), so this should be `thrust::distance(remaining_vertices.begin(), thrust::copy_if(...))`

You cannot directly pass an rmm::device_uvector object by value to the device functor from the host, so this should be like
```
[in_degree = raft::device_span<edge_t const>(in_degrees.data(), in_degrees.size()), out_degree = raft::device_span<edge_t const>(out_degrees.data(), out_degrees.size()), vertex_first = graph_view.local_vertex_partition_range_first()]__device__(vertex_t v) {
  auto offset = v - vertex_first;
  return in_degrees[offset] >= 1 && out_degrees[offset] >=1;
}
```

Note that we are encouraging to use raft::host_span and raft::device_span instead of raw pointers (see std::span https://en.cppreference.com/w/cpp/container/span if you are not familiar with the concept of span).

cpp/src/cores/trim.cu
```diff
@@ -0,0 +1,39 @@
...
+    handle.get_stream());
+
+  return remaining_vertices;
```
seunghwak, line 37: Finding vertices with non-zero in|out degrees and finding induced subgraph is one way to solve this problem.

Another more direct approach is to use extract_if_e.

```

  if (graph_view.count_self_loops(handle) > edge_t{0}) {
    auto [srcs, dsts] = extract_if_e(handle,
                                     graph_view,
                                     edge_src_dummy_property_t{}.view(),
                                     edge_dst_dummy_property_t{}.view(),
                                     is_not_self_loop_t<vertex_t>{});

    if constexpr (multi_gpu) {
      std::tie(srcs, dsts, std::ignore, std::ignore) =
        detail::shuffle_ext_vertex_pairs_to_local_gpu_by_edge_partitioning<vertex_t,
                                                                           edge_t,
                                                                           weight_t,
                                                                           int32_t>(
          handle, std::move(srcs), std::move(dsts), std::nullopt, std::nullopt);
    }

    std::tie(*modified_graph, std::ignore, std::ignore, renumber_map) =
      create_graph_from_edgelist<vertex_t, edge_t, weight_t, int32_t, false, multi_gpu>(
        handle,
        std::nullopt,
        std::move(srcs),
        std::move(dsts),
        std::nullopt,
        std::nullopt,
        cugraph::graph_properties_t{true, graph_view.is_multigraph()},
        true);

    modified_graph_view = (*modified_graph).view();
  }
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/community/triangle_count_impl.cuh#L201

Here, instead of passing dummy property views, you need to pass in_degrees and out_degrees.

See
```
    edge_src_property_t<decltype(cur_graph_view), uint8_t> edge_src_in_two_cores(handle,
                                                                                 cur_graph_view);
    edge_dst_property_t<decltype(cur_graph_view), uint8_t> edge_dst_in_two_cores(handle,
                                                                                 cur_graph_view);
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/community/triangle_count_impl.cuh#L246

and

```
    update_edge_src_property(
      handle, cur_graph_view, in_two_core_flags.begin(), edge_src_in_two_cores);
    update_edge_dst_property(
      handle, cur_graph_view, in_two_core_flags.begin(), edge_dst_in_two_cores);
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/community/triangle_count_impl.cuh#L257

and

```
                                     edge_src_in_two_cores.view(),
                                     edge_dst_in_two_cores.view(),
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/community/triangle_count_impl.cuh#L263

to figure out how to do this.

Let me know if you have additional questions.

seunghwak:
https://github.com/rapidsai/cugraph/pull/3325 is replacing this PR.

PR #3267, opened 2023-02-12
Download datasets in ARM wheel tests.

bdice:
Datasets are not being downloaded in ARM wheel tests. This PR is a small fix for CI tests that are currently disabled (thus this error does not show up). I caught this while looking over the wheel test logic as a reference.


ajschmidt8 (approved):

bdice:
@ajschmidt8 We might have merged this a little too quickly. I hadn’t fully considered that this change means that ARM _smoke_ tests spend extra time and bandwidth fetching datasets that they don’t use. I think I might revert this and add a comment explaining why only amd64 builds fetch the datasets.

PR #3269, opened 2023-02-14
API improvements for end-to-end MG sampling performance

ChuckHastings:
Recent discussion has identified some performance issues in end-to-end testing of the MG sampling code.

This PR implements changes to the API to improve the end-to-end performance.


cpp/include/cugraph/algorithms.hpp
```diff
@@ -1752,17 +1760,18 @@ std::tuple<rmm::device_uvector<vertex_t>,
...
-  std::optional<
-    edge_property_view_t<edge_t,
-                         thrust::zip_iterator<thrust::tuple<edge_t const*, edge_type_t const*>>>>
-    edge_id_type_view,
+  std::optional<edge_property_view_t<edge_t, edge_t const*>> edge_id_view,
+  std::optional<edge_property_view_t<edge_t, edge_type_t const*>> edge_type_view,
   rmm::device_uvector<vertex_t>&& starting_vertices,
   std::optional<rmm::device_uvector<int32_t>>&& starting_labels,
+  std::optional<rmm::device_uvector<edge_t>>&& starting_vertex_offsets,
+  std::optional<rmm::device_uvector<int32_t>>&& batch_id_to_output_gpu_mapping,
```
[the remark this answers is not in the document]
seunghwak: And can we reconstruct a tree per starting vertex with the current API?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1752,17 +1760,18 @@ std::tuple<rmm::device_uvector<vertex_t>,
            std::optional<rmm::device_uvector<edge_t>>,
            std::optional<rmm::device_uvector<edge_type_t>>,
            rmm::device_uvector<int32_t>,
-           std::optional<rmm::device_uvector<int32_t>>>
+           std::optional<rmm::device_uvector<int32_t>>,
+           std::optional<rmm::device_vector<edge_t>>>
```
seunghwak, line 1764: Won't `edge_t` overflow? `starting_vertices.size()` can possibly overflow 32 bit integers (this may not happen frequently, but we cannot compromise on correctness).

seunghwak (commented):
Few quick comments, let me think more about this and I will post additional questions.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1752,17 +1760,18 @@ std::tuple<rmm::device_uvector<vertex_t>,
...
 uniform_neighbor_sample(
   raft::handle_t const& handle,
```
seunghwak, line 1766: So, these are what this API should satisfy.

* If `starting_vertices` includes seeds from multiple batches, the caller should be able to separate sampling outputs from different batches and the caller should be able to place sampling outputs on different GPUs based on mapping between batch IDs and GPUs.
* For the output stored on a single GPU, we should be able to sort them based on batch IDs.
* Once we sort based on the batch ID, some users may want to sort based on hops as a primary key (if they don't need to create a separate tree per seed within a batch) or seed IDs as a primary key and hops as a secondary key (if they need to create a separate tree per seed).

seunghwak: I am thinking about the following,

This function optionally takes `rmm::device_uvector<label_t>&& labels ` and `rmm::device_uvector<int>&& dst_comm_ranks` (`labels.size()` == `dst_comm_ranks.size()` == `start_vertices.size()`).
Here labels don't necessarily coincide with batch IDs, but users may set labels to `batch_id * batch_size + seed index within a batch` (if the result need to be sorted within a batch using seed IDs as a primary key and hops as a secondary key) or just `batch_id` (if no need to create a tree per seed). The output results are shuffled based on `dst_comm_ranks` for each starting vertex (if `dst_comm_ranks` is valid) or all the results for local seeds will be stored locally.

We return optional labels and hops for each edge data (src, dst, optional (edge weight, ID, type)). Within each GPU, the output results will be sorted using labels as a primary key (if `labels` are provided in the input) and hops as a secondary key.

One concern is that computing global "seed index within a batch" might be challenging for the caller if they want to distribute seeds for a single batch to multiple GPUs, but I guess this is not a common use case. If there are 100 batches and 10 GPUs, users may assign batch [0-10) to GPU0 (so `start_vertices` for GPU0 holding all the seeds for batch [0-10)), batch [10-20) to GPU1, and so on.

We may be able to reduce the output size if we don't return label, hop per every edge data and create offset arrays per label & hop, but I am not sure the benefit out-weights the increase in complexity. If the memory footprint is a major concern, we may just reduce the number of batches we process per call (as the main reason for processing multiple batches in a single call is to saturate GPUs, but if we're hitting the memory limit, we might be going to far in this direction).

Anything am I missing or any other thoughts?

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -679,42 +681,43 @@ extract_induced_subgraphs(
...
+ * @return Tuple of the generated graph and optional edge_property_t objects storing the provided
+ * edge properties and a renumber map (if @p renumber is true).
  */
 template <typename vertex_t,
           typename edge_t,
-          typename weight_t,
-          typename edge_type_t,
+          typename property1_t,
+          typename property2_t,
+          typename property3_t,
```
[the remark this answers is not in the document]
ChuckHastings: Longer term I wonder if the solution is to type erase the edge properties, since they are infrequently used, and dispatch at runtime to the proper type.

I had moved to this solely to offer flexibility if we decide a different combination of up to 3 properties was desirable for some algorithm.  You are correct we would need to instantiate whatever combinations we need up front (unless we type erase).

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: We had explicitly tried to avoid "batch ID" in the previous API, largely because the previous implementation had some baggage associated with it.  Since we've deleted that code, I'd be happy to switch back to batch id if that feels more intuitive.

I'm not sure if we should keep the old technique or just migrate to the new technique (passing CSR-like structure).  I'd be OK changing the API to use `starting_vertex_offsets` and eliminating `starting_labels`.  I was thinking it might be helpful to transition (support both, drop the `starting_labels` later).

cpp/include/cugraph/algorithms.hpp
ChuckHastings: For the use cases we've been discussing, I don't think that would happen.  However, as you described in our slack discussions, we want to have the API support future cases as well.  `size_t` would make more sense, I can make that change.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: This is a good point, they're a lot to unpack here.  Let me create a few separate comments.

The original objective of processing multiple batches in a single call is to efficiently utilize the parallelism on the GPU.  If we do individual calls with a small batch size then we don't get enough parallelism and the overhead of calling the function is large relative to the work accomplished.  By processing multiple batches of seeds in the same call we increase the amount of parallel work to be done on the GPU and we reduce the number of times we face the overhead of calling the function, allocating memory, launching kernels, etc.  This improves our overall efficiency.

I don't know where the threshold is, but at some point there are enough vertices in a call to get the efficiencies we are looking for.  Doing larger numbers of batches in a single call should still drive down overhead, but the improvement should decrease dramatically once we reach saturation of the GPU parallelism.  It seems like (with the large graphs we are targeting) that this threshold is much less than "all seeds to be sampled in a training epoch".

In the current pipeline, the output from this processing is going to be moved out of GPU memory to be retrieved by the trainers as they work (currently written at the python layer to parquet files).  As long as we process in the first call enough batches to keep the trainers busy while we do a second call, there would be no drop in pipeline throughput by making multiple calls to the sampling code with smaller numbers of batches.  That would reduce the memory requirement, and we can overlap the sampling with the training a bit more.  It might actually improve the overall latency for training an epoch (although perhaps only marginally, I don't really know the performance numbers you're dealing with).

This does suggest that the extra memory required to drive this function doesn't need to be a driving factor in the design of the API.

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Perhaps the more detailed discussion below should inform this choice.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Regarding the labels discussion.  Part of the reason I went with the name "label" rather than "batch" was this notion that batches might not be the only reason for labeling the data.

There's absolutely no reason that the sampling code needs to understand anything about how the caller assigns labels and uses them.  If we keep with the label as a vector (`labels.size() == start_vertices.size()`) rather than switching to a CSR-style representation, that label can literally be any arbitrary 32-bit integer value (doesn't need to be contiguous values starting from 0).  Either route for organizing the result (either mapping label to GPU id or providing a destination vector (`dst_comm_ranks.size() == start_vertices.size()`) allows for us to shuffle the data to the correct GPU.  We could talk about sorting options (beyond sorting by label) if that would be helpful, although that seems like an easy enough feature to add later when we need it.

It seems like much of the use cases you describe can be addressed by the caller creating the labels in a more sophisticated way and then grouping the results once they are returned.  There's no reason the caller couldn't create labels such that the training batch is decomposed into `k` labels that all get mapped to the same GPU and the results will be combined by the caller.  That allows for construction of the trees.  That allows for the same seed to be repeated within a batch and being able to differentiate between which tree came from which seed.

The beauty there is that the responsibility for managing the sophistication lies in the caller rather than in the library function.

cpp/include/cugraph/graph_functions.hpp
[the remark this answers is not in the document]
seunghwak: Yeah, I agree that type erasing might be a good longer term solution, but we are not there yet (and we're not processing three arrays separately but building a tuple out of these three types and passing the tuple to a functor, so we actually still need to explicitly instantiate in the lower level if I am not mistaken).

If we're not planning to move to the type erasing direction in short term, I am a bit concerned about the inconsistency this will bring to our API.

cpp/include/cugraph/algorithms.hpp
seunghwak: Yeah, I believe this should be general purpose, and I hope we have more users in the future.

cpp/include/cugraph/algorithms.hpp
seunghwak: >I don't know where the threshold is, but at some point there are enough vertices in a call to get the efficiencies we are looking for. 
I guess the memory footprint for the sampling is a significant percentage (say more than 10%) of the GPU's total memory, we may not be too far from saturating the GPU. And I agree that "This does suggest that the extra memory required to drive this function doesn't need to be a driving factor in the design of the API." We still need to be frugal with memory but it should not be the #1 priority when we need to make trade-offs.

seunghwak: >Regarding the labels discussion. Part of the reason I went with the name "label" rather than "batch" was this notion that batches might not be the only reason for labeling the data.
>There's absolutely no reason that the sampling code needs to understand anything about how the caller assigns labels and uses them.

Yes, now I agree with this. And I just want to emphasize that this API should allow creating one tree per seed within a batch, so we need to provide a mechanism to distinguishes sampled edges from different seeds, and labels can serve this purpose (32 bit might be sufficient for the foreseeable future, but in few GPU generations in the future, especially with grace-hopper like systems, we may exceed the 32 bit boundary, so we may keep it as label_t than int32_t).

cpp/include/cugraph/graph_functions.hpp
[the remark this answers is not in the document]
ChuckHastings: We can't go to type erasing until we've built up some infrastructure for that inside the C++ code base.  I wasn't think we'd try and go there yet.  We haven't even really talked about putting that on our road map yet.

I guess I was trying to be forward thinking about other potential uses.  Today the only properties we support are weight, edge id, edge type, where weight is of type `weight_t`, edge id is of type `edge_t` and edge type is of type `int32_t`.  This more generic naming would allow us to support, for example, creating a graph with multiple weights and picking which one to apply to different algorithms (a feature supported by networkx).  We do nothing inside this function, or inside any of our algorithms, that really uses anything but a weight.  The sampling algorithm blindly returns whatever values are in the properties.  It doesn't matter that they are an edge id and an edge type.

I'm OK changing them back if we think that's better until we have a type erased solution.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I will add a `label_t` and only instantiate `int32_t` for now.

cpp/include/cugraph/graph_functions.hpp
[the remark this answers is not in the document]
seunghwak: I have thought more about this, and there are few things to consider.

1) Can users currently use this without knowing that property1_t is weight_t, property2_t is edge_t, and property3_t is edge_type_t? They can't if I am not mistaken. I guess property?_t makes this API more difficult to use.
2) Should we really go to the type erasing path and what do we really get? We need to explicitly instantiate in lower level if we're willing to pass a thrust::tuple to a functor. Otherwise, individual functors need to deal with type erased variables, and this adds overhead in both coding complexity and performance.
3) If we're passing many properties in a type erased fashion, we may better take a different approach. If we create an array per property and if we access 100 properties, this will involve 100 4-8 byte random accesses. If we store 100 properties in a packed fashion, this will involve 1 access of size 100 * (4-8) bytes. In this case, we may store device_span to the packed data in a view object and better pass this device_span object to individual functors. Assuming that individual functors know how to unpack the array, the API does not explicitly specify the types.

I guess using property?_t is little pre-mature at this point as we are not sure this choice will really be future proof or not at this point.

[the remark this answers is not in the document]
alexbarghi-nv: I agree this makes the API confusing, since there are algorithms that use weight, for instance, and it isn't clear whether the weight is property1, property2, or property3.  If users could name the properties and weren't restricted in the types of each property, then the algorithms would just require the user specify which property to treat as weight.  But I wouldn't do that without type erasure.  I concur with Seungwha on leaving them as they are.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1728,41 +1728,51 @@ k_core(raft::handle_t const& handle,
...
  * level
+ * @param rng_state A pre-initialized raft::RngState object for generating random numbers
  * @param with_replacement boolean flag specifying if random sampling is done with replacement
  * (true); or, without replacement (false); default = true;
- * @param rng_state A pre-initialized raft::RngState object for generating random numbers
  * @return tuple device vectors (vertex_t source_vertex, vertex_t destination_vertex,
- * optional weight_t weight, optional edge_t edge id, optional edge_type_t edge type, int32_t hop,
- * optional int32_t label)
+ * optional weight_t weight, optional edge_t edge id, optional edge_type_t edge type,
+ * optional int32_t hop, optional int32_t label, optional edge_t offsets)
```
seunghwak, line 1749: "optional edge_t offsets" => "optional size_t offsets", and I think we need to re-read this function's documentation and make this up to date.

cpp/include/cugraph/graph_functions.hpp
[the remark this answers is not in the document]
ChuckHastings: I will revert this.

Type erasing is, admittedly, a big design conversation we should have independent of this activity.  If we go the type erasing route it really needs to be for more than just this special case, and it would likely need to have effects that ripple through all of the primitives to truly be effective.  Cudf uses the type erased route all the way to cuda kernel calls and has had good success with it.  But we would need to evaluate and decide to go that route (if it makes sense) and lay out a comprehensive plan, not do it piecemeal.

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
ChuckHastings: Switching back to "label" in the next push.

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Fixed in next push

seunghwak (approved):
LGTM

alexbarghi-nv (approved):
👍

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -663,7 +663,9 @@ extract_induced_subgraphs(
  *
  * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
  * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
- * @tparam edge_type_t Type of edge type identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weight.  Needs to be floating point type
+ * @tparam edge_id_t Type of edge id.  Needs to be an integral type.
+ * @tparam edge_type_t Type of edge type.  Needs to be an integral type.
```
seunghwak, line 668: Aren't we currently limiting edge type to be int32_t? Better add something like, `currently, we support only int32_t for edge_type_t`?

cpp/src/detail/shuffle_vertex_pairs.cu
```diff
@@ -80,58 +82,93 @@ shuffle_vertex_pairs_by_gpu_id_impl(
   std::optional<rmm::device_uvector<edge_t>> rx_edge_ids{std::nullopt};
   std::optional<rmm::device_uvector<edge_type_t>> rx_edge_types{std::nullopt};
 
+  // FIXME: Consider a generic function that takes a value tuple of optionals
+  //        to eliminate this complexity
```
seunghwak, line 86: Yeah... I have thought about this in the past, but not sure this is possible as long as we want to explicitly instantiate for different type combinations.

Updating `groupby_and_count` or creating a wrapper function for `groupby_and_count` to take optional objects and dynamically evaluate them sounds more like pushing the complexity to one lower level, but this sounds like an overkill at this moment unless this logic gets reused in many different places (and this function is already a pretty low level detail space function).

cpp/src/detail/shuffle_vertex_pairs.cu
```diff
@@ -160,96 +197,159 @@ shuffle_vertex_pairs_by_gpu_id_impl(
...
+          majors.resize(0, handle.get_stream());
+          majors.shrink_to_fit(handle.get_stream());
+          minors.resize(0, handle.get_stream());
+          minors.shrink_to_fit(handle.get_stream());
+          (*weights).resize(0, handle.get_stream());
+          (*weights).shrink_to_fit(handle.get_stream());
+          (*edge_ids).resize(0, handle.get_stream());
+          (*edge_ids).shrink_to_fit(handle.get_stream());
+          (*edge_types).resize(0, handle.get_stream());
+          (*edge_types).shrink_to_fit(handle.get_stream());
```
seunghwak, line 236: I think we can move this logic to the end of this function (right before return) to avoid repetition.

```
          majors.resize(0, handle.get_stream());
          majors.shrink_to_fit(handle.get_stream());
          minors.resize(0, handle.get_stream());
          minors.shrink_to_fit(handle.get_stream());
          if (weights) { ... }
          if (edge_ids) { ... }
          if (edge_types) { ... }
```

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -530,59 +619,110 @@ create_graph_from_edgelist_impl(
...
     }
 
     edge_partition_offsets.push_back(std::move(offsets));
     edge_partition_indices.push_back(std::move(indices));
     if (edge_partition_weights) { (*edge_partition_weights).push_back(std::move(*weights)); }
-    if (edge_partition_id_type_pairs) {
-      (*edge_partition_id_type_pairs).push_back(std::move(*id_type_pairs));
-    }
-    if (edge_partition_dcs_nzd_vertices) {
-      (*edge_partition_dcs_nzd_vertices).push_back(std::move(*dcs_nzd_vertices));
```
seunghwak, line 547: Why did you delete this?

seunghwak (commented):
Review part 1 (mostly minor comments about documentation & names).

I will review the actual implementation of the uniform neighbor sampling updates in a separate PR.

cpp/src/sampling/detail/graph_functions.hpp
```diff
@@ -30,48 +30,67 @@ namespace detail {
...
  * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
  * handles to various CUDA libraries) to run graph algorithms.
- * @param graph_view Non-owning graph object.
+ * @param graph_view Graph View object to generate NBR Sampling on.
+ * @param edge_weight_view Optional view object holding edge weights for @p graph_view.
+ * @param edge_id_view Optional view object holding edge ids for @p graph_view.
+ * @param edge_type_view Optional view object holding edge types for @p graph_view.
  * @param active_majors Device vector containing all the vertex id that are processed by
  * gpus in the column communicator
+ * @param active_major_labels Optional device vector containing labels for each device vector
```
seunghwak, line 48: Do we ever use this without valid `active_major_labels`? (If it is invalid, isn't this just decompressing a graph to edge list?).

cpp/src/sampling/detail/graph_functions.hpp
```diff
@@ -82,26 +101,103 @@ template <typename vertex_t,
...
+ *
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param label_offsets The offsets array for labels
+ *
+ * @returns device vector containing labels for each seed vertex
+ */
+template <typename label_t>
+rmm::device_uvector<label_t> expand_label_offsets(raft::handle_t const& handle,
+                                                  rmm::device_uvector<size_t> const& label_offsets);
```
seunghwak, line 184: This wont' be necessary if we update the sampling API to take a label for each seed vertex.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -249,77 +376,88 @@ template <typename vertex_t,
...
+                                   label_t,
+                                   multi_gpu>(handle,
+                                              graph_view,
+                                              edge_weight_view,
+                                              edge_id_view,
+                                              edge_type_view,
+                                              active_majors,
+                                              vertex_label_frontier,
+                                              do_expensive_check);
   } else {
```
seunghwak, line 418: If we always take `active_major_labels`, this path can be deleted.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +631,320 @@ sample_edges(
...
+  // calling groupby_gpu_id_and_shuffle_values there is no benefit in reducing peak memory as we
+  // need to allocate a receive buffer anyways) but this reduces the maximum memory allocation size
+  // by half or more (thrust::sort used inside the groupby_and_count allocates the entire temporary
+  // buffer in a single chunk, and the pool allocator  often cannot handle a large single allocation
+  // (due to fragmentation) even when the remaining free memory in aggregate is significantly larger
+  // than the requested size).
+  auto d_tx_value_counts = cugraph::groupby_and_count(
+    labels->begin(),
+    labels->end(),
+    [label_to_output_gpu_mapping] __device__(auto val) { return label_to_output_gpu_mapping[val]; },
```
seunghwak, line 691: We can use the pre-defined device functor.

```
struct indirection_t {
  Iterator first{};

  __device__ typename thrust::iterator_traits<Iterator>::value_type operator()(size_t i) const
  {
    return *(first + i);
  }
};
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/include/cugraph/utilities/device_functors.cuh#L41

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +631,320 @@ sample_edges(
...
+    comm_size,
+    mem_frugal_threshold,
+    handle.get_stream());
```
seunghwak, line 694: You should shuffle (majors, minors, ...) with labels. I guess this is a bug here.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +631,320 @@ sample_edges(
...
+
+  std::vector<size_t> h_tx_value_counts(d_tx_value_counts.size());
+  raft::update_host(h_tx_value_counts.data(),
+                    d_tx_value_counts.data(),
+                    d_tx_value_counts.size(),
+                    handle.get_stream());
+  handle.sync_stream();
+
+  if (mem_frugal_flag) {  // trade-off potential parallelism to lower peak memory
```
seunghwak, line 703: I am not sure this additional complication here is really worth the effort. If the memory footprint really becomes a bottleneck, should we better reduce the number of `start_vertices`?

seunghwak: Or even in this shuffle function, we can shuffle in chunks. And shuffles here won't save the peak memory if we call `groupby_and_count` above with (majors, minors, ...).

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +631,320 @@ sample_edges(
...
+    }
+  }
+
+  return std::make_tuple(std::move(majors),
+                         std::move(minors),
+                         std::move(weights),
+                         std::move(edge_ids),
+                         std::move(edge_types),
+                         std::move(hops),
+                         std::move(labels));
```
seunghwak, line 927: And this function does not sort the output based on labels (and I guess this is the right design), so we need to sort the output based on labels before constructing an offset array for the labels.

cpp/src/sampling/detail/sampling_utils_impl.cuh
seunghwak: Have you tested this in multi-GPU? If yes, we should update our test suites to detect this.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1720,50 +1720,63 @@ k_core(raft::handle_t const& handle,
...
+ * @param rng_state A pre-initialized raft::RngState object for generating random numbers
+ * @param return_hops boolean flag specifying if the hop information should be returned
  * @param with_replacement boolean flag specifying if random sampling is done with replacement
  * (true); or, without replacement (false); default = true;
- * @param rng_state A pre-initialized raft::RngState object for generating random numbers
  * @return tuple device vectors (vertex_t source_vertex, vertex_t destination_vertex,
- * optional weight_t weight, optional edge_t edge id, optional edge_type_t edge type, int32_t hop,
- * optional int32_t label)
+ * optional weight_t weight, optional edge_t edge id, optional edge_type_t edge type,
+ * optional int32_t hop, optional int32_t label, optional size_t offsets)
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1720,50 +1720,63 @@ k_core(raft::handle_t const& handle,
...
  * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
  * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
  * handles to various CUDA libraries) to run graph algorithms.
  * @param graph_view Graph View object to generate NBR Sampling on.
  * @param edge_weight_view Optional view object holding edge weights for @p graph_view.
- * @param edge_id_type_view Optional view object holding edge ids and types for @p graph_view.
+ * @param edge_id_view Optional view object holding edge ids for @p graph_view.
+ * @param edge_type_view Optional view object holding edge types for @p graph_view.
  * @param starting_vertices Device vector of starting vertex IDs for the sampling.
  * @param starting_labels Optional device vector of starting vertex labels for the sampling.
```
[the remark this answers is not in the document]
ChuckHastings: We don't require these to be local vertices in multi-GPU, as implemented.  This is different from the usual paradigm.  The reason is that to support multiple levels of fanout, we have to shuffle.  As implemented we shuffle prior to sampling, which means if the `starting_vertices` passed in are not shuffled they get shuffled in that first step.  We could require the vertices to be shuffled and skip the shuffling step for level 0.  But I'm not sure there's a real benefit to that.  If the data is properly shuffled there is little extra cost in doing the shuffling.

I'll update the documentation.  If you think there would be a performance benefit let me know and I'll look into skipping the first shuffle.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1720,50 +1720,63 @@ k_core(raft::handle_t const& handle,
...
+ * @param starting_vertex_offsets Optional device vector organizing @p starting_vertices into
+ * batches of vertices, used in place of @p starting_labels.  If @p starting_vertex_offsets is
+ * specified then output will include offsets element (defining the offsets that correspond to a
+ * particular input batch)
```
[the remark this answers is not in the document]
ChuckHastings: See the above discussion.  I did not intend this parameter to be what you perhaps are imagining.  I think we should put together some example input/output samples of what we think we want in different scenarios and review these (including @alexbarghi-nv and @VibhuJawa in that review) so that we can all agree on what is desired here.  I'll try to put together some examples for review and schedule something for next week.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1720,50 +1720,63 @@ k_core(raft::handle_t const& handle,
...
+ * @param batch_id_to_output_gpu_mapping Optional device vector maping batch id to a particular
+ * output GPU.  Index is the batch id (either from @p starting_labels, or an intger <
+ * starting_vertex_offsets->size() - 1), value should be in the range 0 to num_gpus - 1
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: I didn't delete this.  It's up higher (line 630 on the right).  Not sure why I needed up moving it before the others instead of leaving it here.

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -40,23 +40,25 @@ namespace detail {
...
-  rmm::device_uvector<vertex_t>,
-  rmm::device_uvector<vertex_t>,
-  std::optional<rmm::device_uvector<weight_t>>,
-  std::optional<std::tuple<rmm::device_uvector<edge_t>, rmm::device_uvector<edge_type_id_t>>>>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<weight_t>>,
+           std::optional<rmm::device_uvector<edge_t>>,
+           std::optional<rmm::device_uvector<edge_type_id_t>>>
 shuffle_ext_vertex_pairs_to_local_gpu_by_edge_partitioning(
```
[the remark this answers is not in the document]
ChuckHastings: Trying new name: `shuffle_ext_vertex_pairs_with_values_to_logcal_gpu_by_vertex_partitioning` in next push.

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -75,25 +77,28 @@ shuffle_ext_vertex_pairs_to_local_gpu_by_edge_partitioning(
...
-  rmm::device_uvector<vertex_t>,
-  rmm::device_uvector<vertex_t>,
-  std::optional<rmm::device_uvector<weight_t>>,
-  std::optional<std::tuple<rmm::device_uvector<edge_t>, rmm::device_uvector<edge_type_id_t>>>>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<weight_t>>,
+           std::optional<rmm::device_uvector<edge_t>>,
+           std::optional<rmm::device_uvector<edge_type_id_t>>>
 shuffle_int_vertex_pairs_to_local_gpu_by_edge_partitioning(
```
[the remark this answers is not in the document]
ChuckHastings: Trying new name: `shuffle_int_vertex_pairs_with_values_to_local_gpu_by_edge_partitioning` in next push.

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: Fixed in next push

cpp/src/detail/shuffle_vertex_pairs.cu
ChuckHastings: Perhaps overkill at the moment.  My concern is that as properties proliferate codes like this will become much more complex.  But I agree for now it's fine to defer this notion (that's why I made it a comment).

cpp/src/detail/shuffle_vertex_pairs.cu
ChuckHastings: Is there a reason to prefer this strategy (storing results in `rx_majors`, shrinking `majors` and later doing a `std::move` over the top) rather than just overwriting `majors` in the first place?

That is, I could do:
`std::tie(majors, minors, *weights, *edge_ids, *edge_types)` as the return for `shuffle_values` and it will just do an `std::move` over the top of those values.

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -33,67 +33,82 @@ template <typename vertex_t,
...
-                         thrust::zip_iterator<thrust::tuple<edge_t const*, edge_type_t const*>>>>
-    edge_id_type_view,
+  std::optional<edge_property_view_t<edge_t, edge_t const*>> edge_id_view,
+  std::optional<edge_property_view_t<edge_t, edge_type_t const*>> edge_type_view,
   rmm::device_uvector<vertex_t>&& seed_vertices,
-  std::optional<rmm::device_uvector<int32_t>>&& seed_vertex_labels,
-  raft::host_span<int32_t const> h_fan_out,
+  std::optional<rmm::device_uvector<label_t>>&& seed_vertex_labels,
+  std::optional<rmm::device_uvector<size_t>>&& seed_vertex_offsets,
+  std::optional<rmm::device_uvector<int32_t>>&& label_to_output_gpu_mapping,
```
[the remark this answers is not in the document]
ChuckHastings: Will change `seed_vertices` to `start_vertices`.

Agree we should identify the criteria for when to choose `raft::device_span` and when to pass an R-value.

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -33,67 +33,82 @@ template <typename vertex_t,
...
                                                                               // to store hops
 
+  CUGRAPH_EXPECTS(!(seed_vertex_labels && seed_vertex_offsets),
+                  "Cannot specify both seed_vertex_labels and seed_vertex_offsets");
+
+  if constexpr (!multi_gpu) {
+    CUGRAPH_EXPECTS(!label_to_output_gpu_mapping,
+                    "cannot specify output GPU mapping in SG implementation");
+
+    CUGRAPH_EXPECTS(!label_to_output_gpu_mapping || seed_vertex_labels || seed_vertex_offsets,
```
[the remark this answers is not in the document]
ChuckHastings: Fixed in next push

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -33,67 +33,82 @@ template <typename vertex_t,
...
+                    "cannot specify output GPU mapping without also specifying either "
+                    "seed_vertex_labels or seed_vertex_offsets");
+  }
+
+  if (seed_vertex_offsets) {
+    seed_vertex_labels = detail::expand_label_offsets<label_t>(handle, *seed_vertex_offsets);
```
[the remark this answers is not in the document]
ChuckHastings: We should address this as part of the earlier comment/discussion.

cpp/src/sampling/detail/graph_functions.hpp
ChuckHastings: Yes.  This was added specifically to support the specifying of offsets in the API.  If we revert that we would eliminate this function.

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
seunghwak: >But I'm not sure there's a real benefit to that.

I assume the key benefit is the API consistency. Consistent API is easier to use, and we can use the same implementation in the PLC/C layer, so this makes code validation & maintenance easier.

If we don't take `std::optional<rmm::device_uvector<int32_t>>&& label_to_output_gpu_mapping`, we may store the sampling results based on which GPU the seeds are stored, but as we have `label_to_output_gpu_mapping`, no need for this.

And shuffling initial seeds is much cheaper than shuffling the final results (assuming multiple levels, and fanout values >> 1), so I am more inclined for API consistency.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1720,50 +1720,63 @@ k_core(raft::handle_t const& handle,
...
  * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
  * handles to various CUDA libraries) to run graph algorithms.
  * @param graph_view Graph View object to generate NBR Sampling on.
  * @param edge_weight_view Optional view object holding edge weights for @p graph_view.
- * @param edge_id_type_view Optional view object holding edge ids and types for @p graph_view.
+ * @param edge_id_view Optional view object holding edge ids for @p graph_view.
+ * @param edge_type_view Optional view object holding edge types for @p graph_view.
  * @param starting_vertices Device vector of starting vertex IDs for the sampling.
  * @param starting_labels Optional device vector of starting vertex labels for the sampling.
+ * @param starting_vertex_offsets Optional device vector organizing @p starting_vertices into
```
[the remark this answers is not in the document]
seunghwak: > So in implementation we instantiate the labels column internally before shuffling.

Yes, so we aren't saving memory, but I feel like just making the API little more cryptic.

> We could, instead require passing starting_labels

+1 for this.

cpp/include/cugraph/algorithms.hpp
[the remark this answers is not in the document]
seunghwak: I assume this discussion is no longer relevant if we use `starting_vertex_labels` (size == `starting_vertices.size()`).

cpp/include/cugraph/detail/shuffle_wrappers.hpp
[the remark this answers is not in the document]
seunghwak: Yes, I like this better!

cpp/src/detail/shuffle_vertex_pairs.cu
seunghwak: > Is there a reason to prefer this strategy

No, and I think what you described is better.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
seunghwak: Could you move the code back here? It make the code easy to follow.

cpp/src/sampling/uniform_neighbor_sampling_impl.hpp
```diff
@@ -33,67 +33,82 @@ template <typename vertex_t,
...
+  level_result_dst_vectors.reserve(fan_out.size());
+  if (level_result_weight_vectors) { (*level_result_weight_vectors).reserve(fan_out.size()); }
+  if (level_result_edge_id_vectors) { (*level_result_edge_id_vectors).reserve(fan_out.size()); }
+  if (level_result_edge_type_vectors) { (*level_result_edge_type_vectors).reserve(fan_out.size()); }
+  if (level_result_label_vectors) { (*level_result_label_vectors).reserve(fan_out.size()); }
 
   std::vector<size_t> level_sizes{};
   int32_t hop{0};
-  for (auto&& k_level : h_fan_out) {
+  for (auto&& k_level : fan_out) {
```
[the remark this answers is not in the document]
seunghwak: > but it does make things consistent.

Yeah, I think we'd better be consistent if comparable.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +626,455 @@ sample_edges(
...
+      static_cast<size_t>(static_cast<double>(total_global_mem / element_size) * mem_frugal_ratio);
+
+    // FIXME:  Need some MG tests that actually exercise all of these variations...
+
+    auto d_tx_value_counts = cugraph::groupby_and_count(
+      labels->begin(),
+      labels->end(),
+      [output_label = std::get<0>(*label_to_output_comm_rank),
+       output_rank  = std::get<1>(*label_to_output_comm_rank)] __device__(auto val) {
+        auto pos = thrust::find(thrust::seq, output_label.begin(), output_label.end(), val);
```
seunghwak, line 683: `thrust::find` is linear scan, this can be expensive if output_label.size() is large. Better use binary search (if output_labels can be sorted) or a hash table. May look for prims/kv_store.cuh as well.

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +626,455 @@ sample_edges(
...
+          }
+        }
+      }
+    }
+
+    size_t num_unique_labels =
+      thrust::count_if(handle.get_thrust_policy(),
+                       thrust::make_counting_iterator<size_t>(0),
+                       thrust::make_counting_iterator<size_t>(labels->size()),
+                       [d_labels = labels->data()] __device__(size_t idx) {
```
seunghwak, line 1036: You can use predefined `is_first_in_run_t`.
```
        auto num_unique_majors =
          thrust::count_if(rmm::exec_policy(loop_stream),
                           thrust::make_counting_iterator(size_t{0}),
                           thrust::make_counting_iterator(this_chunk_size),
                           is_first_in_run_t<vertex_t const*>{tmp_majors.data()});
```
https://github.com/rapidsai/cugraph/blob/branch-23.04/cpp/src/structure/renumber_edgelist_impl.cuh#L449

cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -448,5 +626,455 @@ sample_edges(
...
+                      [d_unique_labels = raft::device_span<label_t const>{
+                         unique_labels.data(), unique_labels.size()}] __device__(label_t label) {
+                        auto pos = thrust::lower_bound(
+                          thrust::seq, d_unique_labels.begin(), d_unique_labels.end(), label);
+                        return static_cast<label_t>(thrust::distance(d_unique_labels.begin(), pos));
+                      });
+
+    offsets = detail::compute_sparse_offsets<size_t>(
+      labels->begin(), labels->end(), size_t{0}, unique_labels.size(), handle.get_stream());
+    labels = std::move(unique_labels);
```
seunghwak, line 1066: Can't you just use `reduce_by_key`? `labels` are already sorted and you want to compute the counts of each labels (you can compute offsets from counts by running scan).

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -898,6 +902,7 @@ weight_t compute_total_edge_weight(
 template <typename vertex_t, typename edge_t, bool store_transposed, bool multi_gpu>
 rmm::device_uvector<vertex_t> select_random_vertices(
   raft::handle_t const& handle,
+
```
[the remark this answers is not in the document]
ChuckHastings: Removed in next push

cpp/src/sampling/detail/sampling_utils_impl.cuh
ChuckHastings: I meant to go back and look at that, I will fix in next push.

cpp/src/sampling/detail/sampling_utils_impl.cuh
ChuckHastings: Thanks for the link. I knew we did something like this in the renumbering code, but I wasn't able to find it.

Refactored in the next push to look like the linked renumbering code.

cpp/src/sampling/detail/sampling_utils_impl.cuh
ChuckHastings: Refactored in the next push to look like the linked renumbering code.

seunghwak (approved):
LGTM

alexbarghi-nv (approved):
👍

cpp/tests/sampling/mg_uniform_neighbor_sampling.cu
```diff
@@ -143,11 +143,31 @@ class Tests_MGUniform_Neighbor_Sampling
...
     thrust::tabulate(handle_->get_thrust_policy(),
-                     batch_number.begin(),
-                     batch_number.end(),
-                     [batch_size = uniform_neighbor_sampling_usecase.batch_size] __device__(
-                       int32_t index) { return index / batch_size; });
+                     comm_ranks.begin(),
+                     comm_ranks.end(),
+                     [num_gpus = handle_->get_comms().get_size()] __device__(auto index) {
+                       return index % num_gpus;
+                     });
```
seunghwak, line 170: Is this logic correct?

After
```
    thrust::tabulate(
      handle_->get_thrust_policy(),
      batch_number.begin(),
      batch_number.end(),
      [rank       = handle_->get_comms().get_rank(),
       batch_size = uniform_neighbor_sampling_usecase.batch_size] __device__(int32_t index) {
        return rank + (index / batch_size);
      });
```
Batch IDs in each GPU will be [rank, (local_number_of_sources + (batch_size - 1)) / batch_size).

Say local_number_of_sources = 100 and batch_size = 10.

GPU 0, batch IDs will be [0, 10)
GPU 1, Batch IDs will be [1, 11).

Isn't `num_batches` the number of unique batch IDs? In the case above, this should be 11 (0, 1, 2, ... , 10), right?

Not sure how
```
    size_t num_batches = 1 + handle_->get_comms().get_rank() +
                         (batch_number.size() / uniform_neighbor_sampling_usecase.batch_size);
    num_batches = cugraph::host_scalar_allreduce(
      handle_->get_comms(), num_batches, raft::comms::op_t::MAX, handle_->get_stream());
```
computes the number of unique batches.

And shouldn't the batch_size be globally enforced for each batch ID? Based on this logic, batch_size will be a local batch_size, and global batch_size for each batch ID will grow with # GPUs. Am I missing something?

ChuckHastings: Yes, this logic is faulty.  I will reconsider.

ChuckHastings: Restructured this code in latest push.  Batch sizes should now be correct except for the last batch which might be small.

cpp/tests/sampling/mg_uniform_neighbor_sampling.cu
```diff
@@ -143,11 +143,33 @@ class Tests_MGUniform_Neighbor_Sampling
...
+
+    thrust::tabulate(
+      handle_->get_thrust_policy(),
+      batch_number.begin(),
+      batch_number.end(),
+      [seed_offset = seed_offsets[handle_->get_comms().get_rank()],
+       num_batches] __device__(int32_t index) { return (seed_offset + index) % num_batches; });
+
+    rmm::device_uvector<int32_t> unique_batches(num_batches, handle_->get_stream());
+    rmm::device_uvector<int32_t> comm_ranks(num_batches, handle_->get_stream());
```
seunghwak, line 163: So, are we replicating the same `unique_batches` and `comm_ranks` in every GPU?

I might be thinking too far, but this can be a problem if `# labels` is very large. May not need to address in this PR, but something we need to at least start thinking about.

cpp/tests/sampling/mg_uniform_neighbor_sampling.cu
seunghwak: The new update looks good to me.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -27,13 +29,20 @@
 
 from cugraph.dask.comms import comms as Comms
 
+from typing import Sequence, List, Union, Tuple
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from cugraph import Graph
+
 src_n = "sources"
```
BradReesWork, line 38: no code change

We really need a utility class with all the standard column names.

BradReesWork (approved):

PR #3270, opened 2023-02-14
Add missing cugraph-ops conditional

vyasr:
This change is necessary for compilation with `USE_CUGRAPH_OPS=OFF`.


vyasr:
After this change I'm still seeing an error `/home/vyasr/local/rapids/cugraph/cpp/tests/sampling/sg_uniform_neighbor_sampling.cu (144): error: no instance of function template "cugraph::uniform_neighbor_sample" matches the argument list` locally. That doesn't seem related to cugraph-ops at all, but perhaps some test needs to be updated in some way for another reason.

ChuckHastings:
> After this change I'm still seeing an error `/home/vyasr/local/rapids/cugraph/cpp/tests/sampling/sg_uniform_neighbor_sampling.cu (144): error: no instance of function template "cugraph::uniform_neighbor_sample" matches the argument list` locally. That doesn't seem related to cugraph-ops at all, but perhaps some test needs to be updated in some way for another reason.

That is likely unrelated.  Are you sure this is a clean build from source?  The `cugraph::uniform_neighbor_sample` function signature was recently changed.  If you are somehow picking up an older version of the header file you might see this error.  The compilation in CI seems to be succeeding.

vyasr:
Hmm I'm not sure what is happening locally that's causing this issue. I have cleaned the build but I'm still seeing the error. In any case, since this works in CI and I'm not doing active cugraph work other than verifying that builds are working after the spdlog/fmt update in rmm, I'm fine pushing this change out now.

ChuckHastings (approved):

PR #3285, opened 2023-02-14
Define and implement C API for RMAT generators

ChuckHastings:
To get rid of `cython.cu` we need to be able add access to the RMAT generators through the C API.

This PR will define the C API, C unit tests and stub out the implementations so they can be reviewed.


seunghwak (approved):
LGTM

seunghwak (approved):
LGTM

jnke2016 (approved):

cpp/include/cugraph_c/graph_generators.h
```diff
@@ -0,0 +1,258 @@
...
+ * @param [out]    error              Pointer to an error object storing details of any error.  Will
+ *                                    be populated if error code is not CUGRAPH_SUCCESS
+ */
+cugraph_error_code_t cugraph_generate_edge_ids(const cugraph_resource_handle_t* handle,
+                                               cugraph_coo_t* coo,
+                                               bool_t multi_gpu,
+                                               cugraph_error_t** error);
+
+/**
+ * @brief      Generate random edge types, add them to an RMAT COO
```
rlratzel, line 236: Same question here: does it have to be an RMAT COO?

cpp/include/cugraph/graph_generators.hpp
```diff
@@ -76,10 +80,59 @@ std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>> generat
...
+ * affect cuGraph's graph construction performance, so this is generally unnecessary).
+ * If `edge_factor` is given (e.g. Graph 500), set @p num_edges to
+ * (size_t{1} << @p scale) * `edge_factor`. To generate an undirected graph, set @p b == @p c and @p
+ * clip_and_flip = true. All the resulting edges will be placed in the lower triangular part
+ * (including the diagonal) of the graph adjacency matrix.
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param rng_state RAFT RNG state, updated with each call
```
rlratzel, line 100: It looks like this is being passed a NULL ptr in tests. Is there a special behavior if this is NULL vs. a ptr to a RngState?

rlratzel (changes requested):
Minor change requests and some questions.

cpp/include/cugraph_c/graph_generators.h
ChuckHastings: See above answer

cpp/include/cugraph/graph_generators.hpp
rlratzel: My mistake, I did indeed comment in the wrong header file, and I see where the rng_state param is created/initialized before the call.

cpp/include/cugraph_c/graph_generators.h
```diff
@@ -0,0 +1,258 @@
...
+cugraph_error_code_t cugraph_generate_edge_weights(const cugraph_resource_handle_t* handle,
+                                                   cugraph_rng_state_t* rng_state,
+                                                   cugraph_coo_t* coo,
+                                                   cugraph_data_type_id_t dtype,
+                                                   double minimum_weight,
+                                                   double maximum_weight,
+                                                   cugraph_error_t** error);
+
+/**
+ * @brief      Add edge ids to an RMAT COO
```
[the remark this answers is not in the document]
rlratzel: > does this function need to have an RMAT COO, or could it operate on other inputs,

This is closer to what I was wondering.  IOW, the COO is a `cugraph_coo_t` and adding edge IDs and types to a graph seems generally useful outside of an RMAT call, so it wasn't obvious why this was tied to RMAT.  I was expecting the type to be named `cugraph_rmat_coo_t` if it was intended to be used only for RMAT, but I suppose that's more of a comment on the type name.

rlratzel (approved):

cpp/include/cugraph_c/graph_generators.h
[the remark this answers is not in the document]
ChuckHastings: It doesn't strictly have to be a `cugraph_coo_t` generated from the RMAT generator.  When we get around to exposing other graph generators, they will each generate a `cugraph_coo_t` and these functions will be equally applicable.

At the moment, the only `cugraph_coo_t` is from the RMAT generator, and I don't have other generators on the road map yet.  So the comment is correct for now.  We probably will want to update the documentation when we add generators that aren't RMAT based.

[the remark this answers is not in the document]
ChuckHastings: Removed `RMAT` from the documentation block (since I had to merge the latest code and rerun through CI anyway).

PR #3288, opened 2023-02-16
Update cugraph-pyg Recipe and CI Script

### LINKED ISSUE #3083: [FIX] Update Dependencies for cuGraph-PyG to Include PyTorch and PyG
Version 22.12 of `cugraph-pyg` did not include PyTorch and PyG as hard dependencies due to the unavailability of PyG 2.2 and limited time for testing.  Updating the dependency list is critical to ensure CI runs correctly and we maintain compatibility with PyG and PyTorch.  It may also be required to make `cupy` a hard dependency, at least until the issues with `CuGraphStore` using `cupy` functions where it shouldn't be are resolved.

### LINKED ISSUE #3225: Update cuGraph-PyG Recipe to Include Dependencies
The current recipe for cugraph-pyg currently does not include its dependencies, which prevents proper builds.  This should be resolved in a CI-friendly manner.

alexbarghi-nv:
Properly installs torch in test script; skips PyG tests on CUDA 11.8.  Removes OGB from dependencies since it is no longer needed and created a version conflict.  Properly adds PyG and PyTorch as dependencies of cugraph-pyg.

Closes #3083 
Closes #3225


VibhuJawa (approved):
LGTM . Thanks for working on this Alex. This resolves a important P0 for this release.

ajschmidt8 (changes requested):
We might need to brainstorm on how to get this properly working in CI.

Although the tests are currently passing, I'm seeing both the `cudatoolkit` and `cuda-toolkit` packages being installed in our CI test environments now, which is not something that we want.

Additionally, the `cuda-toolkit` package version that is being installed doesn't match the `cudatoolkit` version. See the links below for an example on the `11.2.2` test job:

- `cudatoolkit` `11.2.2` is installed (which is what we want) - https://github.com/rapidsai/cugraph/actions/runs/4200038972/jobs/7286563545#step:6:87
- `cuda-toolkit` `11.7.1` is installed shortly after - https://github.com/rapidsai/cugraph/actions/runs/4200038972/jobs/7286563545#step:6:513

I'm not very familiar with the `cugraph` packages that are being changed here.

If they have dependencies that conflict with the other RAPIDS libraries, we might want to consider creating an entirely new conda environment in this test job to accommodate testing these packages.

Happy to hop on a call to chat about this further if needed.

ajschmidt8:
I want to mention that we could also move the `cugraph-pyg` tests to a new job that doesn't run against a full matrix if that makes more sense.

Instead, it would only test against a single CUDA version (e.g. `11.8`).

To make this work, you could adapt the `conda-notebook-tests` job below.

- https://github.com/rapidsai/cugraph/blob/2cf487f88a968b2344f83b0b875645c9b2f2234f/.github/workflows/pr.yaml#L55-L64

The only changes you would need to make are:

- Adjusting the container to run in a container with an appropriate CUDA version (all of our CI image tags are [here](https://hub.docker.com/r/rapidsai/ci/tags) for reference)
- Adjusting the script to something like `ci/test_python_pyg.sh`

alexbarghi-nv:
@rlratzel and @ajschmidt8 could you please re-review this PR?

VibhuJawa (approved):
LGTM

ci/test_python.sh
```diff
@@ -123,5 +107,57 @@ pytest \
...
+    rapids-print-env
+
+    rapids-mamba-retry install \
+      --channel "${CPP_CHANNEL}" \
+      --channel "${PYTHON_CHANNEL}" \
+      libcugraph \
+      pylibcugraph \
+      cugraph \
+      cugraph-pyg
+
```
ajschmidt8, line 140: Two quick notes:

- `rapids-print-env` does a `conda list` ([src](https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-print-env)). so it might be worth placing that command after all the mamba install commands.
- I notice that CI is installing the wrong `cugraph-pyg` version (see [here](https://github.com/rapidsai/cugraph/actions/runs/4486137081/jobs/7899136258?pr=3288#step:7:2410)). This is probably from a dependency issue somewhere. It might take a bit of iterating to figure out why this is happening. I wrote some docs about reproducing CI locally [here](https://docs.rapids.ai/resources/reproducing-ci/) that might help. Happy to hop on a call to help or discuss this more if needed.

ajschmidt8 (commented):
Look good!

Looks like the wrong `cugraph-pyg` version is getting installed though. We'll want to get that sorted out.

conda/recipes/cugraph-pyg/meta.yaml
```diff
@@ -29,13 +29,14 @@ requirements:
     - distributed >=2023.1.1
     - numba >=0.56.2
     - numpy
-    - pytorch <=1.12.1
-    # - pyg (uncomment once pyg 2.2 is released)
+    - pytorch >=2.0
+    - cupy >=9.5.0,<12.0.0a0
+    - cugraph ={{ version }}
+    - pyg-nightly
```
rlratzel, line 35: just want to confirm that we're putting a nightly dependency in our release packages, or are we planning on changing this before code freeze?  If not, can we put an upper-bound pin on this so users installing this package some time later don't pull in a possibly incompatible newer nightly version?

rlratzel (changes requested):
looks good, just possibly one change request.

conda/recipes/cugraph-pyg/meta.yaml
alexbarghi-nv: I just changed this now to resolve the issue of pulling the wrong version of cugraph-pyg.  We will need to bind to pyg 2.3 once it is released, which is supposed to be next week.

rlratzel (approved):

ajschmidt8 (approved):

rlratzel (approved):

PR #3289, opened 2023-02-16
Update cuGraph-PyG Tests and Support Loading Saved Bulk Samples

### LINKED ISSUE #3176: [IMP] CuGraphStore Should Accept GPU Arrays for Input Graph Structure
Currently, cugraph-PyG's `CuGraphStore` only accepts numpy arrays for the graph structure.  The option for the structure to already be on the GPU should be added for a better user experience, and possibly a perfomance improvement in some circumstances.

### LINKED ISSUE #3287: Support loading bulk samples that were already processed in cugraph-pyg
Need to update the loaders and sampling logic in cugraph-pyg to support the case where sampling is done in a separate process.

alexbarghi-nv:
Updates tests to support presence of PyTorch and PyG in the environment.

Updates cugraph-pyg to include support for loading from samples on disk. Makes CuGraphStore serializable and allows creating CuGraphStore and BulkSampleLoader instances without the graph structure.

Merge after #3288 

Closes #3287 
Closes #3176


alexbarghi-nv:
The CI failures on this PR should be resolved when #3288 is merged.

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -188,34 +189,54 @@ class EXPERIMENTAL__CuGraphStore:
...
+                    should match that of the graph.
+
+                G : dict[str, tuple[tensor]] or dict[str, int] (Required)
+                    Dictionary of edge indices.
+                    Option 1 (graph in memory):
+                        Pass the edge indices
+                        i.e. {
+                            ('author', 'writes', 'paper'): [[0,1,2],[2,0,1]],
+                            ('author', 'affiliated', 'institution'): [[0,1],[0,1]]
+                        }
```
VibhuJawa, line 216: Not a change request but i wonder if we should allow users to also add `dask_cudf` frames here .

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -250,10 +271,19 @@ def __init__(
...
-        # Infer number of edges from the edge index dict
-        num_edges_dict = {
-            pyg_can_edge_type: len(ei[0]) for pyg_can_edge_type, ei in G.items()
-        }
+        construct_graph = True
+        if isinstance(next(iter(G.values())), int):
+            # User has passed in the number of edges 
+            # (not the actual edge index), so the number of edges
+            # does not need to be counted.
+            num_edges_dict = dict(G) # make sure the cugraph store owns this dict
```
VibhuJawa, line 279: What does `make sure the cugraph store owns this dict` ensure.

python/cugraph-pyg/cugraph_pyg/loader/cugraph_node_loader.py
```diff
@@ -132,26 +162,42 @@ def __init__(
...
         # Load the next set of sampling results if necessary
         if self.__next_batch >= self.__end_exclusive:
             # Read the next parquet file into memory
-            rank_path = os.path.join(self.__directory.name, f"rank={self.__rank}")
+            dir_path = (
+                self.__directory
+                if isinstance(self.__directory, str)
+                else self.__directory.name
+            )
+            rank_path = os.path.join(dir_path, f"rank={self.__rank}")
```
VibhuJawa, line 178: I have been thinking of removing this `rank` behavior for bulk sampler in the next release. I dont think it is adding any value which cant be derived by just changing the dir path.

VibhuJawa (changes requested):
General Questions, looks good mostly.

python/cugraph-pyg/cugraph_pyg/loader/cugraph_node_loader.py
alexbarghi-nv: I agree we should remove it in 23.06.  I'll create an issue

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
alexbarghi-nv: It's copying the dictionary so that if it is changed later, the version saved in the `CuGraphStore` object is not affected.

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -28,19 +28,22 @@
 
 from cugraph.utilities.utils import import_optional, MissingModule
 
-import dask.dataframe as dd
-from dask.distributed import get_client
-
+dd = import_optional("dask.dataframe")
+distributed = import_optional("dask.distributed")
+dask_cudf = import_optional("dask_cudf")
```
VibhuJawa, line 33: Are these not always imported by cugraph, CC: @rlratzel  ?

alexbarghi-nv: I believe these are always imported, but in anticipation of making those optional in cugraph in the future, I used `import_optional` here

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
alexbarghi-nv: fixed

python/cugraph-pyg/cugraph_pyg/loader/cugraph_node_loader.py
alexbarghi-nv: #3385

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
rlratzel: We have indeed been talking about the possibility of making the MG dependencies optional in cugraph.

python/cugraph-pyg/cugraph_pyg/tests/test_cugraph_store.py
```diff
@@ -383,16 +371,22 @@ def test_get_tensor_from_tensor_attrs(graph):
...
-        assert cugraph_store.get_tensor_size(tensor_attr) == sz
+        assert cugraph_store.get_tensor_size(tensor_attr) == torch.Size((sz,))
+
+
+def test_serialize(multi_edge_multi_vertex_no_graph_1):
+    import pickle
+
+    F, G, N = multi_edge_multi_vertex_no_graph_1
+    cugraph_store = CuGraphStore(F, G, N)
+    pickle.dumps(cugraph_store)
```
rlratzel, line 392: should this at least have an assert, maybe something like this (assuming `==` works here):
```suggestion
    assert pickle.loads(pickle.dumps(cugraph_store)) == cugraph_store
```

rlratzel (approved):
Left an optional suggestion, otherwise I'm just assuming the new test is more of a smoketest to make sure it doesn't throw an exception?

PR #3293, opened 2023-02-17
[IMP] Add CI to cugraph-dgl

VibhuJawa:
This PR adds CI to `cugraph-dgl`.  

This work builds on top of all the work that @alexbarghi-nv did  for enabling `cugraph-pyg` with PR: https://github.com/rapidsai/cugraph/pull/3288


VibhuJawa:
@ajschmidt8 , based on your feedback on https://github.com/rapidsai/cugraph/pull/3288, I tried this approach.  Does something like this seem reasonable to you ?

ajschmidt8:
> @ajschmidt8 , based on your feedback on #3288, I tried this approach for `cugraph-dgl`. Does something like this seem reasonable to you ?
> 
> Note, I do not want to commit on this right now, without discussions with team (compared to earlier PR's approach) . Just adding this option to the table.

Yes, this is what I had in mind.

There are a few changes I'd suggest we make if you guys do decide that you want to pursue this option:

- The `container_image` field is currently invalid, so we'd need to update that. Probably to `rapidsai/ci:cuda11.8.0-ubuntu22.04-py3.10`. Hardcoding this version would mean that you guys would be responsible for maintaining it though.
- There is some duplication between the existing `ci/test_python.sh` file and your new `ci/test_python_cugraph_dgl.sh` file. We can probably DRY this up a bit. You can see how `cudf` did something similar by examing the files below. Notice how `ci/test_python_cudf.sh` and `ci/test_python_other.sh` both make use of `ci/test_python_common.sh`
  - https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_common.sh
  - https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_cudf.sh
  - https://github.com/rapidsai/cudf/blob/branch-23.04/ci/test_python_other.sh
- We created [rapids-dependency-file-generator](https://github.com/rapidsai/dependency-file-generator) / [dependencies.yaml](https://github.com/rapidsai/cugraph/blob/branch-23.04/dependencies.yaml) to eliminate one-off dependency lists [like this one](https://github.com/rapidsai/cugraph/blob/3b6ee7794f16a07e45472bfc1de089ce56726c76/ci/test_python_cugraph_dgl.sh#L39-L46). Ideally we'd have those dependencies captured in the `dependencies.yaml` file. However, it looks like these particular dependencies require additional channels which may not be needed by the other dependencies in that file. I'd need to make some changes to the `dependency-file-generator` tool to support a setup like this. This wouldn't take long, but I'd want to get this resolved to eliminate this single-use dependency list before these changes are merged.

ajschmidt8:
@VibhuJawa, hope you don't mind me pushing a few commits to your PR.

I noticed a few syntax errors in `test_python_cugraph_dgl.sh` script.

I explained the issues in the messages of my commits.

ajschmidt8:
Even though these tests are passing, they're still installing `pytorch` from `conda-forge`, which I don't think is correct based on your `pytorch::pytorch` specification in the install step.

I have some more changes that I'm working on locally to resolve this.

ajschmidt8:
Pushed some more changes that I think should get things passing/installing correctly.

I made a few discoveries when I was making these changes.


- I realize now that [this dependency list](https://github.com/rapidsai/cugraph/blob/3b6ee7794f16a07e45472bfc1de089ce56726c76/ci/test_python_cugraph_dgl.sh#L39-L46) that I originally pointed out as needing to go in `dependencies.yaml` is already specified in the recipe's `run` section [here](https://github.com/rapidsai/cugraph/blob/2cf487f88a968b2344f83b0b875645c9b2f2234f/conda/recipes/cugraph-dgl/meta.yaml#L24-L30). I incorrectly assumed that these were additional test dependencies that needed to be installed in the environment. We can just leave them in the test script for now. I removed all but `pytorch` in my latest changes since we need to specify a CUDA version for that package.
- I'm having second thoughts about suggesting that we move this test script into its own GHAs job. From observing the timestamps for the last successful run of the job, it seems that the job spends ~7.5 minutes creating the conda environments and only 45 seconds actually running the Python tests themselves. Given that GPU resources are finite in CI, this doesn't seem like a good way to spend them. We should probably move this `cugraph-dgl` test script back into the main test scripts.

Let's sync during the work week to make sure all these changes look correct and go over the next steps.

VibhuJawa:
> I made a few discoveries when I was making these changes.
> 
> * I realize now that [this dependency list](https://github.com/rapidsai/cugraph/blob/3b6ee7794f16a07e45472bfc1de089ce56726c76/ci/test_python_cugraph_dgl.sh#L39-L46) that I originally pointed out as needing to go in `dependencies.yaml` is already specified in the recipe's `run` section [here](https://github.com/rapidsai/cugraph/blob/2cf487f88a968b2344f83b0b875645c9b2f2234f/conda/recipes/cugraph-dgl/meta.yaml#L24-L30). I incorrectly assumed that these were additional test dependencies that needed to be installed in the environment. We can just leave them in the test script for now. I removed all but `pytorch` in my latest changes since we need to specify a CUDA version for that package.


So the pytorch version in that script will require `pytorch-cuda` to be specified along with `pytorch>=2.0` for CI testing. 

> * I'm having second thoughts about suggesting that we move this test script into its own GHAs job. From observing the timestamps for the last successful run of the job, it seems that the job spends ~7.5 minutes creating the conda environments and only 45 seconds actually running the Python tests themselves. Given that GPU resources are finite in CI, this doesn't seem like a good way to spend them. We should probably move this `cugraph-dgl` test script back into the main test scripts (guarded by a conditional to ensure that it only runs on certain CUDA versions).


I think we can the conda solving time with provide pytorch and pytorch-cuda  . 

Something like: 
```bash
mamba install pytorch>=2.0  pytorch-cuda=11.7 -c pytorch-nightly -c nvidia
```
Let me update the PR and see how much it reduces. 

> Let's sync during the work week to make sure all these changes look correct and go over the next steps.

@alexbarghi-nv  is organizing a sync b/w all of us to decide which approach to take.

ci/test_python_cugraph_dgl.sh
```diff
@@ -0,0 +1,69 @@
...
+  --channel "${CPP_CHANNEL}" \
+  --channel "${PYTHON_CHANNEL}" \
+  --channel pytorch-nightly \
+  --channel dglteam/label/cu117 \
+  --channel pytorch \
+  libcugraph \
+  pylibcugraph \
+  cugraph \
+  cugraph-dgl \
+  "pytorch::pytorch>=2.0" \
```
alexbarghi-nv, line 41: try changing this to "pytorch>=2.0"

PR #3298, opened 2023-02-23
Stop using versioneer to manage versions

vyasr:
This PR replaces usage of versioneer with hard-coded version numbers in setup.py and init.py. Since cugraph needs to manage versions across a wide range of file types (CMake, C++, Sphinx and doxygen docs, etc), versioneer cannot be relied on as a single source of truth and therefore does not allow us to single-source our versioning to the Git repo as is intended. Additionally, since the primary means of installing raft is via conda packages (or now, pip packages), information from the package manager tends to be far more informative than the version strings for troubleshooting and debugging purposes. Conversely, the nonstandard version strings that it produces tend to be problematic for other tools, which at best will ignore such versions but at worst will simply fail.

This PR also replaces usage of an environment variable to set the package name for wheels in setup.py, instead moving the renaming logic into the same sed script used to update package versions. This change makes setup.py essentially static, paving the way for migration to pyproject.toml.


ci/release/apply_wheel_modifications.sh
```diff
@@ -0,0 +1,46 @@
...
+VERSION=${1}
+CUDA_SUFFIX=${2}
+
+# __init__.py updates
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/cugraph/cugraph/__init__.py
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/cugraph-dgl/cugraph-dgl/__init__.py
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/cugraph-pyg/cugraph-pyg/__init__.py
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/cugraph-service/client/cugraph_service_client/__init__.py
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/cugraph-service/server/cugraph_service_server/__init__.py
+sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" python/pylibcugraph/pylibcugraph/__init__.py
```
ajschmidt8, line 15:
```suggestion
sed -i "s/__version__ = .*/__version__ = \"${VERSION}\"/g" \
  python/cugraph/cugraph/__init__.py \
  python/cugraph-dgl/cugraph-dgl/__init__.py \
  python/cugraph-pyg/cugraph-pyg/__init__.py \
  python/cugraph-service/client/cugraph_service_client/__init__.py \
  python/cugraph-service/server/cugraph_service_server/__init__.py \
  python/pylibcugraph/pylibcugraph/__init__.py
```

`sed` can be run against multiple files.

Should we DRY up these commands like this?

vyasr: Yeah we can, I just hadn't initially because it doesn't work in update-version.sh due to the sed_runner.

rlratzel (approved):

ajschmidt8 (approved):

PR #3305, opened 2023-02-28
Update datasets download URL

jjacobelli:
Some URLs were not updated by https://github.com/rapidsai/cugraph/pull/3299. This PR is fixing this and removing the `datasets/get_test_data.sh` from the ignored files


rlratzel (approved):
Thanks for this update!

PR #3309, opened 2023-03-01
[REVIEW] Fix docs build to be `pydata-sphinx-theme=0.13.0` compatible

galipremsagar:
This PR fixes cugraph docs build to be compatible with `pydata-sphinx-theme=0.13.0` by adding an empty `icon_links` entry.

### DIFF
--- a/docs/cugraph/source/conf.py
+++ b/docs/cugraph/source/conf.py
@@ -116,6 +116,8 @@ todo_include_todos = False
 #
 html_theme_options = {
     "external_links": [],
+    # https://github.com/pydata/pydata-sphinx-theme/issues/1220
+    "icon_links": [],
     "github_url": "https://github.com/rapidsai/cugraph",
     "twitter_url": "https://twitter.com/rapidsai",
     "show_toc_level": 1,


ajschmidt8 (approved):

rlratzel (approved):

rlratzel:
pydata-sphinx-theme v0.13.1 was recently released, making this change no longer necessary.

galipremsagar:
> pydata-sphinx-theme v0.13.1 was recently released, making this change no longer necessary.

@rlratzel Wouldn't it be better to have this fix in place if at all the CI pipelines pickup `0.13.0` because of outside factors?(Implicit dependencies)

ajschmidt8:
> Wouldn't it be better to have this fix in place if at all the CI pipelines pickup 0.13.0 because of any implicit dependencies?

I don't think that's necessary. Our docs building environments are pretty slim. I think it's safe to omit this since a patch has been released.

@galipremsagar, feel free to close any other PRs like this that haven't been merged if you want. If you'd rather merge them, that's fine too. But they're no longer necessary to unblock CI.

PR #3347, opened 2023-03-18
Random vertex sampling utility function for C++ tests

seunghwak:
Code clean-up and use a common vertex sampling function throughout C++ tests.

Additional bug fixes
* C++ induced subgraph test (the reference CPU code assumes that vertex sets are sorted within a set, but this is not required in the cuGraph induced subgraph function).
https://github.com/rapidsai/cugraph/pull/3347/commits/aa25afba18f7c5e7884b8edfd05f7f8a33bb9605
* divide by 0 error when creating a graph with 0 edges (or # edges / # vertices is close to 0) in multi-GPU. Note that static_cast<size_t>(inf) is (surprisingly) 0.
https://github.com/rapidsai/cugraph/pull/3347/commits/290751f928650a890256ab9b07ef388f28428ad8


cpp/tests/utilities/test_utilities_impl.cuh
```diff
@@ -351,5 +355,143 @@ rmm::device_uvector<value_t> mg_vertex_property_values_to_sg_vertex_property_val
...
 
+template <typename vertex_t, bool multi_gpu>
+rmm::device_uvector<vertex_t> randomly_sample_vertices(
+  raft::handle_t const& handle,
+  raft::random::RngState& rng_state,
+  std::vector<vertex_t> const& vertex_partition_range_lasts,
+  size_t sample_size,
+  bool with_replacement,
+  bool sort_samples)
+{
```
ChuckHastings, line 366: Should we perhaps use https://github.com/rapidsai/cugraph/blob/60bd0eb0a5860deda1c440a105be0781c0afec31/cpp/include/cugraph/graph_functions.hpp#L903, or combine these by adding the `sort_samples` parameter to that function and perhaps using this implementation?  Seems like the functionality should be the same.

seunghwak: Oh, yes, you're right. I forgot about the function.

ChuckHastings (approved):

cpp/src/structure/select_random_vertices_impl.hpp
```diff
@@ -27,50 +30,132 @@
...
+      auto min_buffer_size = *std::min_element(buffer_sizes.begin(), buffer_sizes.end());
+      if (min_buffer_size <= select_count / comm_size) {
+        auto new_sizes    = std::vector<size_t>(comm_size, min_buffer_size);
+        auto num_deficits = select_count - min_buffer_size * comm_size;
+        for (int i = 0; i < comm_size; ++i) {
+          auto delta = std::min(num_deficits, buffer_sizes[i] - min_buffer_size);
+          new_sizes[i] += delta;
+          num_deficits -= delta;
+        }
+        this_gpu_select_count = new_sizes[comm_rank];
```
naimnv, line 141: I wonder when would this be necessary?

PR #3354, opened 2023-03-22
Missing indentation leading to an UnboundedLocalError

AmroAlJundi:
In line 105 of the file `python/cugraph/cugraph/sampling/random_walks.py`, the `warnings.warn` call uses the `warning_msg` variable which is only defined inside the preceding if statement (which this statement isn't indented under). This leads to the error:
```
UnboundLocalError: local variable 'warning_msg' referenced before assignment
```
When calling the `random_walks` function with `legacy_result_type == False.`


alexbarghi-nv (approved):
👍

PR #3361, opened 2023-03-24
Properly handle dask change

jnke2016:
A `dask` breaking change updates the way a worker is passed into the distributed function causing all the MG tests to fail.
This PR properly handles the change and also removes the temporary raft PR pinned in the `ci` shell script.


alexbarghi-nv (approved):
Looks fine to me, will also need ops approval.

PR #3371, opened 2023-03-24
MG C++ test updates

seunghwak:
* Generate a SG test graph from the MG graph (instead of re-creating assuming that the underlying random number generator output is deterministic for the same seed).
* Add a utility function to collect MG results to compare with SG results.

Additional bug fix
* Recent PR (#3285) mistakenly updated the code to use the same seed for every GPU. Fix this.
https://github.com/rapidsai/cugraph/pull/3371/commits/f86c2c57ea3757f0825a3fd39df3358b08303375


cpp/src/c_api/random.cpp
```diff
@@ -66,10 +66,8 @@ struct select_random_vertices_functor : public cugraph::c_api::abstract_functor
...
       rmm::device_uvector<vertex_t> local_vertices(0, handle_.get_stream());
 
-      if (!multi_gpu || (handle_.get_comms().get_rank() == 0)) {
-        local_vertices = cugraph::select_random_vertices(
-          handle_, graph_view, rng_state_->rng_state_, num_vertices_, false, false);
-      }
+      local_vertices = cugraph::select_random_vertices(
+        handle_, graph_view, rng_state_->rng_state_, num_vertices_, false, false);
 
       if constexpr (multi_gpu) {
```
ChuckHastings, line 72: Is this block still necessary?  I think your updates to `cugraph::select_random_vertices` already guarantees that the vertices are shuffled.

naimnv (approved):

cpp/src/c_api/random.cpp
seunghwak: You're right and deleted the unnecessary shuffle code.

ChuckHastings (approved):

PR #3393, opened 2023-03-29
[BUG] Reindex Start Vertices and Batch Ids Prior to Sampling Call

alexbarghi-nv:
This fixes a bug where output sample batch ids do not match those expected when using the bulk sampler, causing subgraphs that are larger than expected and incorrect.  Without reindexing, the wrong batch ids are assigned to the start vertices.  Reindexing ensures that the same order is preserved for batch ids and start vertices.


VibhuJawa (changes requested):
PR looks good. Thanks for debugging but we should add a test to catch it please.

rlratzel (approved):

VibhuJawa (approved):

jnke2016 (approved):

PR #3410, opened 2023-03-31
Update raft dependency to 23.06

ChuckHastings:
Looks like we failed to update the `update-version.sh` script to handle `libraft` instead of `libraft-distance` when we made the raft changes in 23.04.

This PR fixes that and updates the raft version to 23.06.


msadang (approved):
LGTM

PR #3416, opened 2023-04-04
[BUG] Critical: Force cudf.concat when passing in a cudf Series to MG Uniform Neighbor Sample

alexbarghi-nv:
Currently, cudf does not merge series properly when they already share an index.  I'm not sure if this is a bug in cudf, or intentional behavior.  This issue does not occur with dask_cudf.  The resolution is to use `cudf.concat` when passing a `cudf.Series` for start vertices and batch ids, and `df.to_frame().merge` when passing in a `dask_cudf.Series` for start vertices and batch ids.

This PR also adds an additional test which tests both cudf and dask_cudf inputs to catch these sort of problems in the future.

### DIFF
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -396,15 +396,25 @@ def uniform_neighbor_sample(
     else:
         indices_t = numpy.int32
 
-    start_list = start_list.rename(start_col_name).to_frame()
+    start_list = start_list.rename(start_col_name)
     if batch_id_list is not None:
-        batch_id_list = batch_id_list.rename(batch_col_name).to_frame()
-        ddf = start_list.merge(
-            batch_id_list,
-            how='left',
-            left_index=True,
-            right_index=True,
-        )
+        batch_id_list = batch_id_list.rename(batch_col_name)
+        if hasattr(start_list, 'compute'):
+            # mg input
+            start_list = start_list.to_frame()
+            batch_id_list = batch_id_list.to_frame()
+            ddf = start_list.merge(
+                batch_id_list,
+                how='left',
+                left_index=True,
+                right_index=True,
+            )
+        else:
+            # sg input
+            ddf = cudf.concat([
+                start_list,
+                batch_id_list,
+            ], axis=1)
     else:
         ddf = start_list.to_frame()
     
@@ -416,8 +426,6 @@ def uniform_neighbor_sample(
             ddf,
             npartitions=len(Comms.get_workers())
         )
-    
-    print(ddf.columns)
 
     ddf = get_distributed_data(ddf)
     wait(ddf)
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -396,17 +396,15 @@ def uniform_neighbor_sample(
     else:
         indices_t = numpy.int32
 
-    start_list = start_list.rename(start_col_name)
+    start_list = start_list.rename(start_col_name).to_frame()
     if batch_id_list is not None:
-        batch_id_list = batch_id_list.rename(batch_col_name)
-        if isinstance(start_list, cudf.Series):
-            concat_fn = cudf.concat
-        else:
-            concat_fn = dask_cudf.concat
-        ddf = concat_fn([
-            start_list,
-            batch_id_list   
-        ], axis=1, names=[start_col_name, batch_col_name])
+        batch_id_list = batch_id_list.rename(batch_col_name).to_frame()
+        ddf = start_list.merge(
+            batch_id_list,
+            how='left',
+            left_index=True,
+            right_index=True,
+        )
     else:
         ddf = start_list.to_frame()
     
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -396,23 +396,34 @@ def uniform_neighbor_sample(
     else:
         indices_t = numpy.int32
 
-    if input_graph.renumbered:
-        start_list = input_graph.lookup_internal_vertex_id(start_list)
-
-    start_list = start_list.rename(start_col_name).to_frame()
+    start_list = start_list.rename(start_col_name)
     if batch_id_list is not None:
-        ddf = start_list.join(batch_id_list.rename(batch_col_name))
+        batch_id_list = batch_id_list.rename(batch_col_name)
+        if isinstance(start_list, cudf.Series):
+            concat_fn = cudf.concat
+        else:
+            concat_fn = dask_cudf.concat
+        ddf = concat_fn([
+            start_list,
+            batch_id_list   
+        ], axis=1, names=[start_col_name, batch_col_name])
     else:
-        ddf = start_list
+        ddf = start_list.to_frame()
+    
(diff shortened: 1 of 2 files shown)


python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -396,23 +396,41 @@ def uniform_neighbor_sample(
...
 
-    if input_graph.renumbered:
-        start_list = input_graph.lookup_internal_vertex_id(start_list)
-
-    start_list = start_list.rename(start_col_name).to_frame()
+    start_list = start_list.rename(start_col_name)
     if batch_id_list is not None:
-        ddf = start_list.join(batch_id_list.rename(batch_col_name))
+        batch_id_list = batch_id_list.rename(batch_col_name)
+        if hasattr(start_list, "compute"):
```
rlratzel, line 402: It looks like the docstring for `start_list` might be incomplete, since it looks like `start_list` can be an int, list, cudf.Series, or dask_cudf.Series?

rlratzel (changes requested):
Thanks for adding a test.  I have one minor change request related to the docstring, otherwise it LGTM.

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: I'll update it

alexbarghi-nv: done

rlratzel (approved):

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -396,23 +396,41 @@ def uniform_neighbor_sample(
...
+            ddf = cudf.concat(
+                [
+                    start_list,
+                    batch_id_list,
+                ],
+                axis=1,
+            )
     else:
-        ddf = start_list
+        ddf = start_list.to_frame()
```
VibhuJawa, line 422: Do we really care about the index here ? I think not . 

```python3
start_list = start_list.reset_index(drop=True)
batch_id_list = batch_id_list.reset_index(drop=True)

if isinstance(start_list, dask_cudf.Series):
   ddf = dd.concat([start_list, batch_id_list], ignore_unknown_divisions=True, axis=1) 
else:
   ddf = cudf.concat([start_list, batch_id_list], axis =1) 
```

VibhuJawa (changes requested):

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: If we reset index can we join batch id and start list correctly?

alexbarghi-nv: And also, I ran into an issue with `dask_cudf.concat` where the name of the series was dropped in one of my first attempts at a solution.  `dask_cudf.merge` doesn't have that problem.

VibhuJawa: I think we should be able to, from the logic you shared , you are merging on index  `( left_index=True, right_index=True)` in dask . 

Edit: Also added `ingore_index=True` to make it more concrete in `cuDF`.

alexbarghi-nv: ok, let me try this

VibhuJawa: Another thing to try is below: 

```python3
if isinstance(start_list, dask_cudf.Series): 
    start_list.divisions = start_list.compute_current_divisions()
     batch_id_list.divisions = batch_id_list.compute_current_divisions()
    ddf = dd.concat([start_list, batch_id_list], axis=1) 
else:
    ddf = cudf.concat([start_list, batch_id_list], axis =1, ignore_index=True) 
```

alexbarghi-nv: First solution isn't working, but the second makes sense, I think that should work.  I'm trying it now.

alexbarghi-nv: 2nd solution works when `ignore_index=False`, I'm going to go ahead with that

VibhuJawa: Cool.  Thanks

VibhuJawa: I am confused why it is failing though , any chance you could post a MRE to cuDF (After merging this fix in), getting this fix in is a higher priority.

alexbarghi-nv: I'll try to get a MRE.  That has been a challenge, though.  What I've been doing is running both the new pytests and my code for processing the MAG dataset.

alexbarghi-nv: At some point I'd like to turn the MAG dataset examples into tests that get run in the MNMG nightlies.

alexbarghi-nv: sorry for the delay, I see some tests failing where the column names aren't propagating correctly

alexbarghi-nv: @VibhuJawa dask_cudf.concat renames the columns to 0 and 1, and when I try to rename them, I get a "cannot rename index" error.

VibhuJawa: Lets file an issue with that detail and just go with your earlier fix. Sorry for the back and forth .


It is unfortunate that we potentially have bugs in such a basic operation in `dask-cudf` and `cuDF`.  Can you add a comment explaining the bugs here to make the code parsable for future.

alexbarghi-nv:
```
ddf = dask_cudf.concat([series1, series2], axis=1)
print(ddf.columns)
```
Expected:
```
Index(['series1_name', 'series2_name'])
```
Actual:
```
Int64Index([0, 1], dtype='int64')
```

alexbarghi-nv: I'll file the issue with cudf tonight.  I know my fix isn't ideal, but at least for now it's not a huge deal since most of our use cases are passing a cudf (not dask cudf) dataframe.

VibhuJawa (approved):
LGTM

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
alexbarghi-nv: I'm not sure why calling `merge` instead of `concat` before `get_distributed_data` works, but for some reason the bug completely disappears with merge.

jnke2016: I can take a look too

jnke2016 (approved):

python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
VibhuJawa: Thanks for creating an issue .

PR #3422, opened 2023-04-05
[BUG] Support Minor Releases of PyG

alexbarghi-nv:
Required to support the upcoming minor release of PyG, which will allow full compatibility with pylibcugraphops.


VibhuJawa (approved):
LGTM

PR #3459, opened 2023-04-07
Addressing Spinx doc build issues

BradReesWork:
Spinx is very picky and looks for indentations and black lines.   Also during the build, if there is an error, Spinx just ignores the function that caused the error rather to terminating the build.  

Some docstrings also used reserved character {* and _} which are used for formating

This PR:
* addresses Spinx Warnings and Errors
* addresses issues where rst is md is not added to the output
* any any other issues

You will see a lot of additional blank lines added and some text clean-up for readability.

Seems like the PR also picked up all the the 23.06 updates - so half the files are not really for this PR


BradReesWork:
Dropping this PR and recreating

PR #3468, opened 2023-04-11
Sphinx updates

BradReesWork:
Spinx is very picky and looks for indentations and black lines. Also during the build, if there is an error, Spinx just ignores the function that caused the error rather to terminating the build.

Some docstrings also used reserved character {* and _} which are used for formating

This PR:

    addresses Spinx Warnings and Errors
    addresses issues where rst is md is not added to the output
    any any other issues

You will see a lot of additional blank lines added and some text clean-up for readability.

Seems like the PR also picked up all the the 23.06 updates - so half the files are not really for this PR


ajschmidt8:
@BradReesWork, in other repositories that didn't have a lot of sphinx warnings, we used the `-W` flag to treat warnings as errors:

- https://github.com/rapidsai/cuml/blob/fb0038757eef379542c557a649c6792c833825fe/ci/build_docs.sh#L35-L39

It's probably worth adding that to `cugraph`'s `ci/build_docs.sh` file as well.

BradReesWork:
@ajschmidt8 I do use the "-W" flag when I'm debugging.  There are still a large number of improvements that need to be finished before the code is ready for that flag - but by the end of this release it should be turned on

docs/cugraph/source/tutorials/cugraph_notebooks.md
```diff
@@ -1,6 +1,6 @@
 # cuGraph Notebooks
 
-![GraphAnalyticsFigure](img/GraphAnalyticsFigure.jpg)
+![GraphAnalyticsFigure](https://github.com/rapidsai/cugraph/tree/main/img/GraphAnalyticsFigure.jpg)
```
rlratzel, line 3: Do we need to use the URL like this now, or can we continue to use the relative path like we were doing before this change?  Seems like using the relative path might be better for working on the docs locally and/or offline.

rlratzel (approved):
Looks good.  I'm approving and this can be merged, but I do have a question which could result in a change based on what you think.

python/pylibcugraph/pylibcugraph/bfs.pyx
```diff
@@ -102,47 +102,45 @@ def bfs(ResourceHandle handle, _GPUGraph graph,
...
+    >>>         cudf.Series([0], dtype='int32'),
+    >>>         False,
+    >>>         10,
+    >>>         True,
+    >>>         False
+    >>> )
+    >>>
+    >>> distances, predecessors, vertices = res
+    >>>
+    f>>> inal_results = cudf.DataFrame({
```
galipremsagar, line 139: Typo?
```
>>> final_results = cudf.DataFrame({
```

PR #3482, opened 2023-04-11
packed bool specialization to store edge endpoint|edge properties

### LINKED ISSUE #3472: Implement efficient mechanism to store Boolean (bit) property values

seunghwak:
This PR replaces https://github.com/rapidsai/cugraph/pull/2829

Pack 32 bools in uint32_t (1 bit per bool) instead of using 1 byte per bool (uint8_t).

This can benefit the current BFS implementation and also can be used to implement vertex/edge masking.

closes #3472


ChuckHastings (approved):
I like that this is both more efficient and more readable.

naimnv (approved):

jnke2016 (approved):

PR #3488, opened 2023-04-14
adding cugraph-ops

BradReesWork:
The goal is to have a unified docs area for all graph work.  This PR is the first step in getting cugraph-ops included in the main graph doc area

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -251,6 +251,7 @@ dependencies:
           - sphinx-markdown-tables
           - sphinx<6
           - sphinxcontrib-websupport
+          - pylibcugraphops=23.6.*
   py_version:
     specific:
       - output_types: [conda]
@@ -368,6 +369,7 @@ dependencies:
           - sphinxcontrib-websupport
           - sphinx-markdown-tables
           - sphinx-copybutton
+          - pylibcugraphops=23.6.*
   test_notebook:
     common:
       - output_types: [conda, requirements]
--- a/docs/cugraph/source/api_docs/cugraph-ops/index.rst
+++ b/docs/cugraph/source/api_docs/cugraph-ops/index.rst
@@ -1,0 +1,16 @@
+=========================
+cugraph-ops API reference
+=========================
+
+This page provides a list of all publicly accessible modules, methods and classes through `pylibcugraphops.*` namespace.
+
+.. toctree::
+    :maxdepth: 2
+    :caption: API Documentation
+
+    graph_types
+    mfg_operators
+    bipartite_operators
+    static_operators
+    fg_operators
+    dimenet
--- a/docs/cugraph/source/api_docs/cugraph-ops/index.rst
+++ b/docs/cugraph/source/api_docs/cugraph-ops/index.rst
@@ -9,8 +9,10 @@ This page provides a list of all publicly accessible modules, methods and classe
     :caption: API Documentation
 
     graph_types
+    pytorch
     mfg_operators
     bipartite_operators
     static_operators
     fg_operators
     dimenet
+    pytorch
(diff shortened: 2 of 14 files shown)


dependencies.yaml
```diff
@@ -251,6 +251,7 @@ dependencies:
           - sphinx-markdown-tables
           - sphinx<6
           - sphinxcontrib-websupport
+          - pylibcugraphops=23.6.*
```
ajschmidt8, line 254: can you add an entry for this file in `update-version.sh`?

BradReesWork: update and in latest push

ajschmidt8 (approved):

docs/cugraph/source/api_docs/cugraph-ops/index.rst
```diff
@@ -0,0 +1,16 @@
...
+cugraph-ops API reference
+=========================
+
+This page provides a list of all publicly accessible modules, methods and classes through `pylibcugraphops.*` namespace.
+
+.. toctree::
+    :maxdepth: 2
+    :caption: API Documentation
+
+    graph_types
```
MatthiasKohl, line 11: should also add pytorch here at the top IMO (just noticed that our own docs have them at the end, but pytorch is the main API for end-users so this should be the first thing to see)

```suggestion
    pytorch
    graph_types
```

MatthiasKohl (changes requested):
Like @tingyu66 pointed out, the pytorch file naming is incorrect.

Additionally, I'm not sure what the plan is, but the C++ API docs are not found here: is the plan to just not add them at all, add them in a later PR, or should they be added immediately here?

docs/cugraph/source/api_docs/cugraph-ops/index.rst
BradReesWork: Adding the C++ API is an upcoming step.   Working on covering from Doxogen to Sphinx.  Once I have the process setup for the other cugraph C code, then I'll be rolling in cugraph-ops.

tingyu66 (approved):
👍

docs/cugraph/source/api_docs/cugraph-ops/index.rst
MatthiasKohl: I believe you now added the pytorch entry twice, as well as not adding it at the top as suggested above.
Let's just remove one of these instances, or completely remove this if it doesn't pass CI

docs/cugraph/source/index.rst
```diff
@@ -1,13 +1,27 @@
...
-NetworkX Graphs, or even CuPy or SciPy sparse Matrices.
+Welcome to RAPIDS Graph documentation
+=====================================
+RAPIDS Graph covers a range of libraries and packages, that includes:
+
+__cugraph__: focuses on GPU accelerated graph algorithms.
+
+__cugraph-ops__: focuses on GNN aggregators and operators.
+
+__cugraph-pg__: focuses on providing support for Property Graph (and future Knowledge Graph).
```
rlratzel, line 9: I don't think this exists yet, is that okay?

rlratzel (commented):
Overall looks good to me. I do think it's a little odd to add docs to components that live in another repo here, but I can't think of a better solution now.

docs/cugraph/source/index.rst
BradReesWork: yep.   That will be discussed in the section on PG and I will be updating this intro section

BradReesWork:
> a little odd to add docs to components that live in another repo

I have been pondering how to create a unified docs page that brings in all the various sections.  Most examples from other products use a central repo that clones all the other repos into one so that docs can be created.  I'm trying to avoid that.

Yes it is odd to have stuff from another repo - but it all falls under Graph

acostadon (approved):
Looks good to me

ajschmidt8 (approved):

PR #3495, opened 2023-04-17
Remove Dummy Edge Weights, Support Specifying Edge Ids/Edge Types/Weights Separately

### LINKED ISSUE #3486: Update Graph creation API (add_cudf_edgelist) to allow for individual edge ID/type/weight specifications
`Graph.add_cudf_edgelist()` should add the ability for users to specify only one or several of edge types/IDs/weights. 

At the moment, users can specify only weights, or weights and type and ID, but they cannot, for example, specify only ID and types.

If a user wants only IDs and types, they must specify weight too, which wastes memory since weights would be allocated but unused.

alexbarghi-nv:
Closes #3486 
Merge after #3513 - Merged


cpp/src/c_api/graph_mg.cpp
```diff
@@ -338,19 +338,19 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
     weight_type = cugraph_data_type_id_t::FLOAT32;
   }
 
-  CAPI_EXPECTS((edge_type_ids == nullptr) || (p_edge_ids->type_ == edge_type),
+  CAPI_EXPECTS((edge_ids == nullptr) || (p_edge_ids->type_ == edge_type),
                CUGRAPH_INVALID_INPUT,
                "Invalid input arguments: Edge id type must match edge (src/dst) type",
```
seunghwak, line 343: Is this comment accurate?

Edge ID should match with edge_t, not vertex_t.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -513,26 +513,19 @@ extern "C" cugraph_error_code_t cugraph_sg_graph_create(
...
-  //         that is that is specified
-  CAPI_EXPECTS(
-    (edge_type_ids == nullptr && edge_ids == nullptr) ||
-      (edge_type_ids != nullptr && edge_ids != nullptr),
-    CUGRAPH_INVALID_INPUT,
-    "Invalid input arguments: either none or both of edge ids and edge types must be provided.",
-    *error);
+  CAPI_EXPECTS((edge_ids == nullptr) || (p_edge_ids->type_ == edge_type),
+               CUGRAPH_INVALID_INPUT,
+               "Invalid input arguments: Edge id type must match edge (src/dst) type",
```
seunghwak, line 518: Same here. edge ID type should match with edge_t, not vertex_t.

alexbarghi-nv: I changed the message, but it's still confusing to the user why they may get an error if their input graph has int64 vertex ids but only int32 edge ids.  It's not that crazy for that to happen (i.e. vertex ids are a hash).  I'll consider adding something at the Python layer to make that more clear.

alexbarghi-nv: I just added the warning at the Python layer.

rlratzel (approved):
I believe I reviewed these changes in #3524

seunghwak (approved):

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -23,9 +23,9 @@ dependencies:
 - doxygen
 - fsspec[http]>=0.6.0
 - gcc_linux-64=11.*
-- gmock=1.10.0
+- gmock>=1.13.0.*
```
raydouglass, line 26: You don't need the `.*` if using relative (`>=`) spec.

raydouglass (approved):

PR #3501, opened 2023-04-19
[cugraph-dgl] Add TransformerConv

tingyu66:
Fixes https://github.com/rapidsai/graph_dl/issues/181


MatthiasKohl (commented):
From my side, everything looks good. I think it would be good if @stadlmax also had a quick look since he's a bit more familiar with the MHA variants than me I believe.

rlratzel (approved):

VibhuJawa (approved):
LGTM

PR #3503, opened 2023-04-20
[cugraph-dgl] Add support for bipartite node features and optional edge features in GATConv

tingyu66:
See title. Fixes https://github.com/rapidsai/graph_dl/issues/183

### DIFF
--- a/python/cugraph-dgl/cugraph_dgl/nn/conv/gatconv.py
+++ b/python/cugraph-dgl/cugraph_dgl/nn/conv/gatconv.py
@@ -14,15 +14,17 @@
 primitives in cugraph-ops"""
 # pylint: disable=no-member, arguments-differ, invalid-name, too-many-arguments
 from __future__ import annotations
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 from cugraph_dgl.nn.conv.base import BaseConv
 from cugraph.utilities.utils import import_optional
 
+from pylibcugraphops.pytorch import BipartiteCSC, SampledCSC, StaticCSC
+from pylibcugraphops.pytorch.operators import mha_gat_n2n, mha_gat_n2n_bipartite
+
 dgl = import_optional("dgl")
 torch = import_optional("torch")
 nn = import_optional("torch.nn")
-ops_torch = import_optional("pylibcugraphops.pytorch")
 
 
 class GATConv(BaseConv):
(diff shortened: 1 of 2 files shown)


python/cugraph-dgl/tests/nn/test_gatconv.py
```diff
@@ -25,52 +25,126 @@
...
     "max_in_degree": [None, 8],
-    "num_heads": [1, 2, 3, 7],
+    "num_heads": [1, 2, 7],
     "to_block": [False, True],
 }
 
 
-@pytest.mark.parametrize(",".join(options.keys()), product(*options.values()))
-def test_gatconv_equality(idtype_int, max_in_degree, num_heads, to_block):
+@pytest.mark.parametrize(",".join(options1.keys()), product(*options1.values()))
```
rlratzel, line 37: Curious why you did it this way instead of letting pytest produce the cartesian product for you.  That has the advantage of easily adding marks, ids, etc. to individual parameters ([example](https://github.com/rapidsai/cugraph/blob/2c43e3051a65f7047a7e3709410337e83839d119/python/cugraph/cugraph/tests/link_analysis/test_pagerank.py#L161-L168)).  I don't have a strong preference, just wondering if there's an advantage for doing it this way.

python/cugraph-dgl/cugraph_dgl/nn/conv/gatconv.py
```diff
@@ -145,32 +172,65 @@ def forward(
...
-                _graph = ops_torch.StaticCSC(offsets_fg, indices)
+        if efeat is not None and self.fc_edge is not None:
+            efeat = self.fc_edge(efeat)
+
+        if bipartite:
+            _graph = BipartiteCSC(
+                offsets=offsets, indices=indices, num_src_nodes=g.num_src_nodes()
+            )
+            nfeat_src = self.fc_src(nfeat[0])
+            nfeat_dst = self.fc_dst(nfeat[1])
```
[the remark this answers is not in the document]
tingyu66: I agree and have added an assertion. AssertionError should be more informative to users than the AttributeError in these cases.

python/cugraph-dgl/cugraph_dgl/nn/conv/gatconv.py
```diff
@@ -145,32 +172,65 @@ def forward(
...
-        out = ops_torch.operators.mha_gat_n2n(
-            feat_transformed,
-            self.attn_weights,
-            _graph,
-            self.num_heads,
-            "LeakyReLU",
-            self.negative_slope,
-            concat_heads=True,
-        ).view(-1, self.num_heads, self.out_feats)[: g.num_dst_nodes()]
+            nfeat = self.fc(nfeat)
```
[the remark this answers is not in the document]
tingyu66: Assertions added.

python/cugraph-dgl/tests/nn/test_gatconv.py
tingyu66: It is just for convenience (at some cost of readability). I would only need to modify one line of code when adding / removing / renaming a parameter.

tingyu66: I changed the code to use pytest cartesian product directly, so that the style is aligned with other cugraph tests.

PR #3509, opened 2023-04-24
Correct MG Leiden and SCC tests

ChuckHastings:
Fix return value for MG implementation of Leiden and SCC, which aren't implemented currently.

The MG tests are failing, although the test code is checking for the proper return value.


seunghwak (approved):
LGTM

Issue #2075, opened 2022-02-15
[ENH] Add Bipartite graph generation option to R-MAT


mughetto:
Sec 3.5 of the R-MAT paper (https://www.cs.cmu.edu/~christos/PUBLICATIONS/siam04.pdf) documents an option for generating a bipartite graph by specifying the number of node on each dimension. 

As far as I can tell this is not doable with the current implementation (that assumes square adjacency).

Is it something that could be easily added?

BradReesWork:
@mughetto we can do that.  We plan to be expanding the number of generators and options.  I will work this into our roadmap

mughetto:
Great to hear! Looking forward to it!

BradReesWork:
@rlratzel time to get Python wrappers

ChuckHastings:
@rlratzel @BradReesWork - Created an epic to track this along with some companion stories.  Please review.

BradReesWork:
Being tracked under this new Epic:  https://github.com/rapidsai/cugraph/issues/2075

Issue #3480, opened 2023-04-11
Refactor cugraph RMAT to use PLC RMAT


rlratzel:
Who: cugraph developers
What: update cugraph RMAT to remove the legacy cython wrapper and call PLC RMAT directly
Why: This allows the cugraph team to remove tech debt by removing the cython.cu file completely. This also makes RMAT consistent with the other APIs that call PLC APIs directly instead of binding to libcugraph C++ directly (PLC is intended to be the layer which binds to libcugraph directly)

```[tasklist]
### Tasks
- [ ] #2817
- [ ] #2818
- [ ] #2820
- [ ] https://github.com/rapidsai/cugraph/issues/3670
```

PR #3512, opened 2023-04-25
Bipartite R-mat graph generation.

seunghwak:
Addresses https://github.com/rapidsai/cugraph/issues/2075

This function will generate (source, destination) vertex ID pairs. Source vertex IDs will have values in `[0, 2^src_scale)` and destination vertex IDs will have values in `[0, 2^dst_scale)`.

Additionally,

* `scramble_vertex_ids` function had unused input parameters and it was internally erroneously setting scale. Fixed this bug.

* Rmat_Usecase was ignoring scramble_vertex_ids flag, fixed this bug.

* Added `scramble_vertex_ids` that take a just single vertex list (instead of src, dst pair)

* Update `scramble_vertex_ids` to take input vectors as R-values and return scrambled vectors (instead taking in/out parameters)


seunghwak:
@BradReesWork Will this API be sufficient to address the issue https://github.com/rapidsai/cugraph/issues/2075?

BradReesWork:
@seunghwak the num_edges  will be the number of edges between the sets?   Also how are you enforcing that edges are only between sets and no within a set?

seunghwak:
> the num_edges will be the number of edges between the sets?
Yes.

> Also how are you enforcing that edges are only between sets and no within a set?

Here, expectation is that source vertices belong to one set and destination vertices belong to the other set.

Say there are two sets U and V.

Say U has 2^u_scale vertices and V has 2^v_scale vertices.

To generate edges from U to V, you call this function with src_scale = u_scale and dst_scale = v_scale.

To generate edges from V to U, you call this function with src_scale = v_scale and dst_scale = u_scale.

And the caller wants vertex IDs for U to be in [0, 2^u_scale) and V to be in [2^u_scale, 2^u_scale + 2^v_scale), they need to add 2^u_scale for either source or vertex IDs after calling this function.

naimnv (approved):

ajschmidt8:
@seunghwak, just an FYI. adding `[skip-ci]` to the PR title doesn't do anything for GitHub Actions.

please review the page below to figure out how to skip CI for GitHub Actions.

- https://docs.rapids.ai/resources/github-actions/#skipping-ci-for-commits

seunghwak:
> Can the caller have any other options than adding |U| to destination vertexIDs (or adding |V| to source vertexIds) if the output edge-lists are used to create graph using `cugraph::create_graph_from_edgelist`?

What do you mean? To explicitly create a bi-partite graph? We currently do not store bi-partitie graphs in a different format, but in the future and if it turns out to be hugely beneficial, we may consider taking (vertex ID within a set, set number) pairs instead of simple vertex IDs for k-partitie graphs. Not considering this right at this moment.

seunghwak:
@ChuckHastings Can we delete the deprecated R-mat generators (which takes a seed) or do we still need to keep them?

ChuckHastings:
> @ChuckHastings Can we delete the deprecated R-mat generators (which takes a seed) or do we still need to keep them?

I believe we need to keep them until the python code is modified to use the new C API approach.  This work is tracked in #3480

cpp/src/generators/generate_rmat_edgelist.cu
```diff
@@ -212,6 +213,83 @@ generate_rmat_edgelists(raft::handle_t const& handle,
...
+            dst += dst_bit_set ? static_cast<vertex_t>(vertex_t{1} << (dst_scale - (level + 1))) : 0;
+          }
+        }
+        return thrust::make_tuple(src, dst);
+      });
+    num_edges_generated += num_edges_to_generate;
+  }
+
+  return std::make_tuple(std::move(srcs), std::move(dsts));
+}
```
naimnv, line 291: Just wondering, is it the algorithm used here?
https://www.cambridge.org/core/journals/network-science/article/linear-work-generation-of-rmat-graphs/68A0DDA58A7B84E9B3ACA2DBB123A16C

naimnv (approved):

naimnv:
> What do you mean?

Yeah right, the vertex IDs in U+V has be to be unique.

cpp/src/generators/generate_rmat_edgelist.cu
seunghwak: Possibly, but this code basically follows the algorithm used in graph 500 reference code.

seunghwak: Maybe not... I just skimmed the abstract, and it seems like the paper claims that edge generation time is just function of the number of edges to generate and irrelevant to scale.

Here, it is dependent on scale. Not sure the algorithm in the paper might actually be faster, but here, scale is pretty much limited, and R-mat graph generation is fast enough for our use cases.

ChuckHastings (approved):

naimnv (approved):

BradReesWork (approved):

PR #3513, opened 2023-04-25
Add support in C API for handling unweighted graphs in algorithms that expect weights

### LINKED ISSUE #3499: Handle dummy weights in C API
The python layer currently passes dummy weights to several algorithms if there are no weights for the graph and the graph requires weights.  This can be an expensive cost, since the weights get added to the graph which consumes GPU memory for its life (which may be longer than the algorithm that requires weights).

Modify the C API so that the following algorithms create a local dummy weight of 1.0 for all edges if the weights aren't specified and removes that weight when the algorithm completes:
* Louvain
* Leiden
* Balanced Cut Clustering
* K-core
* Ego Graphs
* SSSP

This issue is current blocking https://github.com/rapidsai/cugraph/pull/3495

ChuckHastings:
Closes #3499 

Some of our C++ algorithms require a graph to be weighted.  The python layer will add "fake" edge weights of 1 to every edge in order to call these algorithms.  This adds a persisted constant value for every edge that will continue to exist for the life span of the python graph object, even though it is only necessary for a handful of algorithms.

This PR addresses these concerns.

Many of the C++ algorithms already actually support the graph not having weights.  A few should support it and had bugs that were corrected (induced subgraph, k-core, egonet).  A few can easily support it but did not (legacy spectral algorithms).

The Louvain and Leiden algorithms need to have a weighted graph because they will coarsen the weights as the graph is coarsened, and the weights will eventually not be one while the algorithm is running.  These require a slightly different approach.

This PR also added C API tests for unweighted graphs for these algorithms.

### DIFF
--- a/cpp/include/cugraph/edge_property.hpp
+++ b/cpp/include/cugraph/edge_property.hpp
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cugraph/detail/utility_wrappers.hpp>
 #include <cugraph/utilities/dataframe_buffer.hpp>
 #include <cugraph/utilities/thrust_tuple_utils.hpp>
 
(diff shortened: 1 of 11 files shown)


cpp/include/cugraph/edge_property.hpp
```diff
@@ -126,6 +127,26 @@ class edge_dummy_property_t {
...
+  std::for_each(
+    thrust::make_zip_iterator(mutable_view.value_firsts().begin(),
+                              mutable_view.edge_counts().begin()),
+    thrust::make_zip_iterator(mutable_view.value_firsts().end(), mutable_view.edge_counts().end()),
+    [&handle, constant_value](auto tuple) {
+      detail::scalar_fill(handle, thrust::get<0>(tuple), thrust::get<1>(tuple), constant_value);
+    });
+
+  return edge_property;
+}
```
seunghwak, line 148: We have `fill_edge_src|dst_property()` (https://github.com/rapidsai/cugraph/blob/branch-23.06/cpp/src/prims/fill_edge_src_dst_property.cuh).

So, it will be more consistent to create `fill_edge_property()` and call this instead of adding this function. I can quickly write `fill_edge_property()` if you agree.

ChuckHastings: Merged your PR and then my change to use it.

ChuckHastings: Hang on... pushed to wrong branch.

cpp/include/cugraph/edge_property.hpp
```diff
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <cugraph/detail/utility_wrappers.hpp>
```
seunghwak, line 19: Is this include still necessary?

seunghwak (approved):
Better delete the no longer relevant include statement.

LGTM otherwise.

PR #3521, opened 2023-04-27
New mtmg API for integration

ChuckHastings:
Creating a new API for integrating multi-threaded multi-GPU programs into the cugraph library.

This API will extend our OPG (one [process] per GPU) model to support a single process handling multiple GPUs, and will also ultimately support a multi-node configuration where some compute nodes might not have GPUs.

closes https://github.com/rapidsai/graph_dl/issues/241


cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+    size_t device_buffer_size{64 * 1024 * 1024};
+    size_t thread_buffer_size{4 * 1024 * 1024};
+
+    int num_gpus    = gpu_list.size();
+    int num_threads = num_gpus * 4;
+
+    cugraph::mtmg::resource_manager_t resource_manager;
+
+    std::for_each(gpu_list.begin(), gpu_list.end(), [&resource_manager](int gpu_id) {
+      resource_manager.register_local_gpu(gpu_id, gpu_id);
```
Manoj-red-hat, line 114: How we are planning to add remote GPUs ?

cpp/include/cugraph/mtmg/sample.cpp
```diff
@@ -0,0 +1,170 @@
...
+      1. Create a resource_manager_t instance for each process
+      2. Add each GPU to its local resource_manager as a local resource and each remote GPU to all
+    other resource managers as a remote resource
+      3. When we run a GSQL query:
+        a. Determine how many GPUs will be required
+        b. Allocate an instance manager with that many GPUs
+        c. Each thread can request a handle from the instance manager.  The handle contains the
+    necessary information to properly coordinate sending data to GPU memory and launching any
+    required GPU kernels.  Threads are assigned to GPUs in a round-robin fashion.  We will need to
+    have at least one thread for each local GPU in order for the cuGraph code to execute properly.
```
Manoj-red-hat, line 75: So each GPU assigned 1 or more threads, right ?

cpp/include/cugraph/mtmg/sample.cpp
```diff
@@ -0,0 +1,170 @@
...
+   * @return const reference to a raft handle
+   */
+  raft::handle_t const& raft_handle() const { return *raft_handle_; }
+
+  /**
+   * @brief Get thread rank
+   *
+   * @return thread rank
+   */
+  int get_thread_rank() const { return thread_rank_; }
```
Manoj-red-hat, line 38: What is the concept of rank ? thread / GPU

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+    auto h_weights_v = cugraph::test::to_host(handle, d_weights_v);
+
+    result_t constexpr alpha{0.85};
+    result_t constexpr epsilon{1e-6};
+
+    size_t device_buffer_size{64 * 1024 * 1024};
+    size_t thread_buffer_size{4 * 1024 * 1024};
+
+    int num_gpus    = gpu_list.size();
+    int num_threads = num_gpus * 4;
```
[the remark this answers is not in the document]
ChuckHastings: This is simply an artifact of the test.  I wanted to test some number of threads that is greater than the number of GPUs, I picked 4x.

In actual implementation I expect that the calling code will have some means of launching parallel threads, and controlling how many CPU threads are running.  The only real requirement for the code is that the number of CPU threads on each node with GPUs is greater than or equal to the number of GPUs.

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+
+    cugraph::mtmg::resource_manager_t resource_manager;
```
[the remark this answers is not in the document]
ChuckHastings: When we add multi-node support to the resource manager (the method is there, the implementation is not), we will need to provide the information necessary for us to set up NCCL communications between the GPUs.  This will be done via the  `register_remote_gpu` method.

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+    for (int i = 0; i < num_threads; ++i) {
+      running_threads.emplace_back([&instance_manager,
+                                    thread_buffer_size,
+                                    &edgelist,
+                                    &h_src_v,
+                                    &h_dst_v,
+                                    &h_weights_v,
+                                    i,
+                                    num_threads]() {
+        auto thread_handle = instance_manager->get_handle();
```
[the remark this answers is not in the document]
ChuckHastings: Correct.  The `create_instance_manager`, in a multi-node configuration, will coordinate with the remote resource managers to ensure that everything is configured correctly.

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+                                    &graph_view,
+                                    &edgelist,
+                                    &renumber_map,
+                                    &pageranks,
+                                    is_symmetric,
+                                    renumber,
+                                    do_expensive_check]() {
+        auto thread_handle = instance_manager->get_handle();
+
+        if (thread_handle.get_thread_rank() > 0) return;
```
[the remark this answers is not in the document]
ChuckHastings: I mentioned above that we need a CPU thread for each GPU to launch kernels and do host-side computation associated with the GPU kernels.

This if statement is demonstrating a check for the case where you have more CPU threads than GPU threads, but you're performing GPU work.  In this case we don't want the extra CPU threads trying to launch kernels also.  So this aborts early.

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+        raft::update_host(my_pageranks.data(),
+                          d_my_pageranks.data(),
+                          d_my_pageranks.size(),
+                          thread_handle.raft_handle().get_stream());
+
+        {
+          std::lock_guard<std::mutex> lock(computed_pageranks_lock);
+          computed_pageranks_v.push_back(
+            std::make_tuple(std::move(my_vertex_list), std::move(my_pageranks)));
+        }
```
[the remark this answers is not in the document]
ChuckHastings: The `gather` call above will handle these in the multi-node case.

When we create the instance manager, we will configure a NCCL communications object that allows us to do point-to-point messaging between the participating GPUs (entirely in device memory).  The `gather` call above will perform the following steps:
* Shuffle the vertex ids from `d_my_vertex_list` (in device memory) to the GPU where the data for that information resides
* On each GPU, do a local gather of the vertices on that GPU
* Shuffle the vertex ids/pagerank pairs back to the original GPU

This will all be buried in that call, no extra work is required by the caller.

cpp/tests/mtmg/threaded_test.cpp
```diff
@@ -0,0 +1,469 @@
...
+    std::for_each(running_threads.begin(), running_threads.end(), [](auto& t) { t.join(); });
+    running_threads.resize(0);
+    instance_manager->reset_threads();
+
+    std::map<vertex_t, result_t> computed_pageranks;
+    std::for_each(
+      computed_pageranks_v.begin(), computed_pageranks_v.end(), [&computed_pageranks](auto tuple) {
+        auto [vertex, pagerank] = tuple;
+        std::for_each(thrust::make_zip_iterator(vertex.begin(), pagerank.begin()),
+                      thrust::make_zip_iterator(vertex.end(), pagerank.end()),
```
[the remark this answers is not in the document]
ChuckHastings: Yes, the `thrust::zip_iterator` works both in GPU and CPU memory.  I use it fairly regularly in host code (mostly for testing purposes), since STL doesn't offer one yet, and bringing in boost is heavier than we need (since we're already using thrust).

cpp/tests/mtmg/threaded_test.cpp
ChuckHastings: From an API perspective this will look similar.  The `register_remote_gpu` method will have an extra parameter that identifies the communications information necessary to set up NCCL.  That information is still TBD.

cpp/include/cugraph/mtmg/sample.cpp
```diff
@@ -0,0 +1,170 @@
...
+  // to the threads... not worrying about that mechanism at this point.
+  //
+  device_shared_wrapper_t<per_device_edgelist_t<vertex_t, weight_t, edge_t, edge_type_t>> edgelist;
+
+  //
+  //  Within each thread, do something like this
+  //
+  auto handle = instance_manager.get_handle();
+
+  if (handle.get_thread_rank() == 0) {
```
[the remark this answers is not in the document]
ChuckHastings: Yes.  The call on line 119 will get a specific handle instance for this thread.  The threads race to get handles (it's a thread-safe call).  The first one on GPU `k` will get thread id 0, the second one on GPU `k` will get thread id 1, etc.  The `get_thread_rank()` method will allow us to see the order in which they run the race.  By convention (simplest way, used in many MPI programs) is to compare with 0 and allow the thread that one that initial race to be the special thread that will call compute kernels, or do anything that you only want to happen in one CPU thread associated with a GPU.

cpp/include/cugraph/mtmg/sample.cpp
ChuckHastings: Correct.

cpp/include/cugraph/mtmg/sample.cpp
ChuckHastings: I think I described this above, so you can read that for a detailed explanation.  The idea is that each CPU thread associated with a particular GPU is numbered (from `[0..m)`).  Mostly we do that to allow for identifying rank 0 as a special thread.  There are (in MPI world, where this concept was borrowed from) some things that you can do with the rank beyond this (you might use the thread rank to break ties, for example).  But I haven't seen anything so far that we would need beyond thread rank 0 and thread rank > 0.

cpp/tests/mtmg/threaded_test.cpp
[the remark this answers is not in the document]
Manoj-red-hat: So ideally each GPU is attached to 1 or more threads and all threads must be daemon threads, that are going to run till the process doesn't release  its resource manager

cpp/tests/mtmg/threaded_test.cpp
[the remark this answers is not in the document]
Manoj-red-hat: but the resource manager will only accessible to the master node, which is going to create that insance, rest of the workers can talk to resource manager. And if the master is launced dynamically than each and every time we need to create resource manager

cpp/tests/mtmg/threaded_test.cpp
Manoj-red-hat: Understood

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,208 @@
...
+  std::unique_ptr<instance_manager_t> create_instance_manager(
+    std::vector<int> ranks_to_include, ncclUniqueId instance_manager_id) const
+  {
+    std::for_each(
+      ranks_to_include.begin(), ranks_to_include.end(), [local_ranks = local_rank_map_](int rank) {
+        CUGRAPH_EXPECTS(local_ranks.find(rank) != local_ranks.end(),
+                        "requesting inclusion of an invalid rank");
+      });
+
+    ncclGroupStart();
```
seunghwak, line 114: I guess this is unnecessary if ncclCommIinitRank is called in different threads.

cpp/tests/mtmg/bug_reproducer.cu
```diff
@@ -0,0 +1,496 @@
...
+
+  // Wait for CPU threads to complete
+  std::for_each(running_threads.begin(), running_threads.end(), [](auto& t) { t.join(); });
+  running_threads.resize(0);
+  instance_manager->reset_threads();
+}
+
+TEST_F(Tests_MTMG, CheckInt32Int32FloatFloat) { run_current_test(); }
+
+CUGRAPH_TEST_PROGRAM_MAIN()
```
seunghwak, line 496: This initializes RMM only in GPU0.

cpp/include/cugraph/mtmg/sample.cpp
```diff
@@ -0,0 +1,170 @@
...
+ */
+
+#pragma once
+
+#include <raft/core/handle.hpp>
+
+namespace cugraph {
+namespace mtmg {
+
+class handle_t {
```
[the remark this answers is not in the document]
ChuckHastings: This file was an early attempt at demonstrating how things would work.  It will not be part of the final PR.

ChuckHastings (commented):
Thanks, I'll look at these suggestions!

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Removed

cpp/include/cugraph/mtmg/instance_manager.hpp
```diff
@@ -0,0 +1,87 @@
...
+   * the request.  Threads will be assigned to GPUs in a round-robin fashion to
+   * spread requesting threads around the GPU resources.
+   *
+   * This function will be CPU thread-safe.
+   *
+   * @return a handle for this thread.
+   */
+  handle_t get_handle()
+  {
+    int local_id = thread_counter_++;
```
seunghwak, line 58: Sorry for nitpicking but `auto local_gpu_id = thread_counter_++ % raft_handle_.size()` might be cleaner? (avoid duplicating local_id % raft_handle_.size()?)

cpp/include/cugraph/mtmg/instance_manager.hpp
```diff
@@ -0,0 +1,87 @@
...
+ public:
+  /**
+   * @brief Constructor
+   *
+   * @param handles   Vector of RAFT handles, one for each device on this node
+   */
+  instance_manager_t(std::vector<std::shared_ptr<raft::handle_t>>&& handles,
+                     std::vector<std::shared_ptr<ncclComm_t>>&& nccl_comms,
+                     std::vector<rmm::cuda_device_id>&& device_ids)
+    : thread_counter_{0}, raft_handle_{handles}, nccl_comms_{nccl_comms}, device_ids_{device_ids}
```
seunghwak, line 41: And do we really needs `device_ids_`? Isn't it [0, num_gpus)? Or are you considering the case we are using only a subset of the available GPUs?

cpp/include/cugraph/mtmg/instance_manager.hpp
```diff
@@ -0,0 +1,87 @@
...
+   *
+   * This function will be CPU thread-safe.
+   *
+   * @return a handle for this thread.
+   */
+  handle_t get_handle()
+  {
+    int local_id = thread_counter_++;
+
+    cudaSetDevice(device_ids_[local_id % raft_handle_.size()].value());
```
seunghwak, line 60: Guard this in RAFT_CUDA_TRY.

cpp/include/cugraph/mtmg/instance_manager.hpp
```diff
@@ -0,0 +1,87 @@
...
+   */
+  void reset_threads() { thread_counter_.store(0); }
+
+ private:
+  std::atomic<int> thread_counter_{0};
+
+  // FIXME: Should this be an std::map<> where the key is the rank?
+  //        On a multi-node system we might have nodes with fewer
+  //        (or no) GPUs, so mapping rank to a handle might be a challenge
+  //
```
seunghwak, line 80: Can we just compute a local GPU ID by global_rank - min_global_rank_this_node?

seunghwak: We are using map in some places and we are sort of mixing global rank, local rank, local GPU ID, and so on. Can't we just use global rank and local rank (== local GPU ID, or are we assuming that we may not use all the available GPUs)?

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+    int num_gpus_this_node;
+    RAFT_CUDA_TRY(cudaGetDeviceCount(&num_gpus_this_node));
+
+    CUGRAPH_EXPECTS((device_id.value() >= 0) && (device_id.value() < num_gpus_this_node),
+                    "device id out of range");
+
+    local_rank_map_.insert(std::pair(rank, device_id));
+
+    cudaSetDevice(device_id.value());
```
seunghwak, line 81: RAFT_CUDA_TRY

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+    // FIXME: There is a bug in the cuda_memory_resource that results in a Hang.
+    //   using the pool resource as a work-around.
+    //
+    // Ultimately there should be some RMM parameters passed into this function
+    // (or the constructor of the object) to configure this behavior
```
seunghwak, line 87: We can set `NCCL_LAUNCH_MODE=GROUP` to workaround this problem.

seunghwak: And we may need to use this with the pool allocator as well assuming the pool allocator will use cudaMalloc/cudaFree for bulk allocation. Just the odd for hang is much lower with the pool allocator but it could be harder to reproduce and debug once it happens.

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+#if 0
+  /**
+   * @brief add a remote GPU to the resource manager.
+   *
+   * FIXME: Need some sort of comms information for the remote GPU here...
+   *
+   * @param rank    The rank to assign to the remote GPU
+   */
+  void register_remote_gpu(int rank, TBD const& remote_info);
+#endif
```
seunghwak, line 114: Delete dead code.

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+    std::vector<int> ranks_to_include, ncclUniqueId instance_manager_id) const
+  {
+    std::for_each(
+      ranks_to_include.begin(), ranks_to_include.end(), [local_ranks = local_rank_map_](int rank) {
+        CUGRAPH_EXPECTS(local_ranks.find(rank) != local_ranks.end(),
+                        "requesting inclusion of an invalid rank");
+      });
+
+    std::vector<std::shared_ptr<ncclComm_t>> nccl_comms{};
+    std::vector<std::shared_ptr<raft::handle_t>> handles{};
```
seunghwak, line 142: What's the rationale behind using shared_ptr here? Do we really expect multiple owner? Or instance manager will be the sole owner and all the other callers will just use it without managing the memory allocation/deallocation?

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+    auto gpu_row_comm_size = static_cast<int>(sqrt(static_cast<double>(ranks_to_include.size())));
+    while (ranks_to_include.size() % gpu_row_comm_size != 0) {
+      --gpu_row_comm_size;
+    }
+
+    // FIXME: not quite right for multi-node
+    for (size_t i = 0; i < ranks_to_include.size(); ++i) {
+      int rank = ranks_to_include[i];
+      auto pos = local_rank_map_.find(rank);
+      cudaSetDevice(pos->second.value());
```
seunghwak, line 159: RAFT_CUDA_TRY

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+                                    idx = i,
+                                    gpu_row_comm_size,
+                                    comm_size = ranks_to_include.size(),
+                                    &ranks_to_include,
+                                    &local_rank_map = local_rank_map_,
+                                    &nccl_comms,
+                                    &handles]() {
+        int rank = ranks_to_include[idx];
+        auto pos = local_rank_map.find(rank);
+        cudaSetDevice(pos->second.value());
```
seunghwak, line 182: RAFT_CUDA_TRY

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+        ncclCommInitRank(nccl_comms[idx].get(), comm_size, instance_manager_id, rank);
```
seunghwak, line 184: NCCL_TRY

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+
+#if 0
+  //
+  // TBD: Probably a map with rank used as the key and
+  //    some sort of comms class as the value.  Needs
+  //    to be something we can make a raft handle and
+  //    initialize comms
+  //
+  std::map<int, TBD> remote_rank_map_{};
+#endif
```
seunghwak, line 237: Delete dead code.

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -0,0 +1,241 @@
...
+        return pair.first;
+      });
+
+    return registered_ranks;
+  }
+
+ private:
+  mutable std::mutex lock_{};
+  std::map<int, rmm::cuda_device_id> local_rank_map_{};
+  // Look at making this immutable (normal)
```
seunghwak, line 225: Yeah... I guess marking this as mutable doesn't make much sense.

cpp/include/cugraph/mtmg/thread_edgelist.hpp
```diff
@@ -0,0 +1,174 @@
...
+ * stored in GPU memory (distributed across multiple GPUs in a multi-GPU configuration).
+ *
+ * This class provides a mechanism for populating the edgelist object from independent CPU threads.
+ *
+ * Calls to the append() method will take edges (in CPU host memory) and append them to a local
+ * buffer.  As the local buffer fills, the buffer will be sent to GPU memory using the flush()
+ * method.  This allows the CPU to GPU transfers to be larger (and consequently more efficient).
+ */
+template <typename vertex_t, typename weight_t, typename edge_t, typename edge_type_t>
+class thread_edgelist_t {
```
seunghwak, line 39: `per_thread_edgelist_t` to be consistent with `per_device_edgelist_t`?

cpp/src/mtmg/vertex_result.cu
```diff
@@ -0,0 +1,167 @@
...
+namespace cugraph {
+namespace mtmg {
+
+template <typename result_t>
+template <typename vertex_t, typename edge_t, bool store_transposed, bool multi_gpu>
+rmm::device_uvector<result_t> vertex_result_t<result_t>::gather(
+  handle_t const& handle,
+  raft::device_span<vertex_t const> vertices,
+  cugraph::mtmg::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu> const& graph_view,
+  std::optional<cugraph::mtmg::renumber_map_t<vertex_t>>& renumber_map)
```
seunghwak, line 35: Can't this be const?

seunghwak: And we may consider defining a MTMG span or view object instead of passing a constant reference of owning objects to be consistent with the rest of cugraph.

cpp/src/mtmg/vertex_result.cu
```diff
@@ -0,0 +1,167 @@
...
+{
+  auto my_graph_view = graph_view.get_pointer(handle);
```
seunghwak, line 37: `my_graph_view` here sounds like a lazy name. Shouldn't this better be this_gpu_graph_view or this_thread_graph_view or something else that tells that we are extracting a graph view for a single GPU from an object that holds graph_view objects for the entire set of GPUs in this node?

cpp/src/mtmg/vertex_result.cu
```diff
@@ -0,0 +1,167 @@
...
+
+  rmm::device_uvector<vertex_t> local_vertices(vertices.size(), handle.get_stream());
+  rmm::device_uvector<int> vertex_gpu(vertices.size(), handle.get_stream());
```
seunghwak, line 40: `vertex_gpu_ids`?

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Done.  Will probably add again in next phase, but the signature will be clarified.

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Deleted, will add back in next phase.

cpp/src/mtmg/vertex_result.cu
ChuckHastings: Added view objects in the latest push.  The view objects are const.

cpp/src/mtmg/vertex_result.cu
ChuckHastings: Definitely lazy :-). Renamed as suggested.

cpp/src/mtmg/vertex_result.cu
ChuckHastings: Renamed.

cpp/include/cugraph/mtmg/graph.hpp
```diff
@@ -0,0 +1,136 @@
...
+  std::optional<cugraph::mtmg::edge_property_t<
+    cugraph::mtmg::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>,
+    weight_t>>& edge_weights,
+  std::optional<cugraph::mtmg::edge_property_t<
+    cugraph::mtmg::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>,
+    edge_id_t>>& edge_ids,
+  std::optional<cugraph::mtmg::edge_property_t<
+    cugraph::mtmg::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>,
+    edge_type_t>>& edge_types,
+  std::optional<cugraph::mtmg::renumber_map_t<vertex_t>>& renumber_map,
```
seunghwak, line 92: Why are you taking references instead of returning?

cpp/include/cugraph/mtmg/graph.hpp
```diff
@@ -0,0 +1,136 @@
...
+  bool do_expensive_check = false)
+{
+  if (handle.get_thread_rank() > 0) return;
```
seunghwak, line 95: I am not following this code, so you are creating a graph object only for GPU 0?

cpp/include/cugraph/mtmg/per_thread_edgelist.hpp
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+ private:
+  detail::per_device_edgelist_t<vertex_t, weight_t, edge_t, edge_type_t>& edgelist_;
+  size_t current_pos_{0};
+  std::vector<vertex_t> src_;
+  std::vector<vertex_t> dst_;
+  std::optional<std::vector<weight_t>> wgt_;
+  std::optional<std::vector<edge_t>> edge_id_;
+  std::optional<std::vector<edge_type_t>> edge_type_;
```
seunghwak, line 170: Adding {} at the end might be a good habit.

cpp/include/cugraph/mtmg/resource_manager.hpp
seunghwak: I didn't know this is deprecated, better ask to the NCCL folks about a proper longer term solution to work around this problem.

cpp/include/cugraph/mtmg/detail/device_shared_wrapper.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+    auto pos = pointers_.find(handle.get_rank());
+
+    return std::const_pointer_cast<wrapped_t const>(pos->second);
+  }
+
+ private:
+  mutable std::mutex lock_{};
+
+  std::map<int, std::shared_ptr<wrapped_t>> pointers_{};
```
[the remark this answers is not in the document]
seunghwak: I think `std::vector<std::unique_ptr<T>>` may work, but we may not need to address this right away.

cpp/include/cugraph/mtmg/graph.hpp
ChuckHastings: This function is designed to be called by multiple threads concurrently.  The implementation will insert results into the references passed in.  The only way I can think to return would be to push the host threading logic inside this function, but I was trying to allow the calling code to manage the threads.

cpp/include/cugraph/mtmg/graph.hpp
ChuckHastings: `get_rank()` returns the GPU id.  `get_thread_rank()` returns the host thread id for this GPU.  This check is to ensure that you don't get multiple threads calling the logic for any given GPU.

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -0,0 +1,274 @@
...
+    if (use_edge_type) {
+      edge_type_ = std::make_optional(std::vector<rmm::device_uvector<edge_type_t>>());
+    }
+
+    create_new_buffers(handle);
+  }
+
+  per_device_edgelist_t(per_device_edgelist_t&& other)
+    : device_buffer_size_{other.device_buffer_size_},
+      current_pos_{0},
```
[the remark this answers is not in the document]
ChuckHastings: Fixed this to set `current_pos_` from `other.current_pos_`.  The original use case for adding this only called it while `current_pos_` was 0, so I didn't catch that.

Changed `per_device_edgelist_t& operator=(per+device_edgelist_t&&)` to be deleted (don't need this).

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -0,0 +1,274 @@
...
+  mutable std::mutex lock_{};
+
+  size_t current_pos_{0};
+  size_t device_buffer_size_;
+
+  std::vector<rmm::device_uvector<vertex_t>> src_;
+  std::vector<rmm::device_uvector<vertex_t>> dst_;
+  std::optional<std::vector<rmm::device_uvector<weight_t>>> wgt_;
+  std::optional<std::vector<rmm::device_uvector<edge_t>>> edge_id_;
+  std::optional<std::vector<rmm::device_uvector<edge_type_t>>> edge_type_;
```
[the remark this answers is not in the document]
ChuckHastings: Done

cpp/include/cugraph/mtmg/per_thread_edgelist.hpp
ChuckHastings: Done

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Sent a message to NCCL folks

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  per_device_edgelist_t(per_device_edgelist_t&& other)
+    : device_buffer_size_{other.device_buffer_size_},
+      current_pos_{other.current_pos_},
+      src_{std::move(other.src_)},
+      dst_{std::move(other.dst_)},
+      wgt_{std::move(other.wgt_)},
+      edge_id_{std::move(other.edge_id_)},
+      edge_type_{std::move(other.edge_type_)}
```
seunghwak, line 92: I guess `= default` will do the same.

seunghwak (approved):
LGTM

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
ChuckHastings: `std::mutex` (the lock) is not moveable.  This move constructor skips that and creates a new mutex.  That does affect thread safety (if you move the object while you're using it).  I'll add documentation about that.

seunghwak: Gotcha, yeah, my bad, I thought we already have the move operator implemented with `= default` and thought why `= default` for the move operator but a custom code for the move constructor. Just find out that the move operator is actually deleted.

PR #3525, opened 2023-04-28
Build wheels using new single image workflow

vyasr:
Up until now wheel builds have been using a two-pass, Docker-in-Docker approach that involves having one container run cibuildwheel, which in turn launches a manylinux container in which the build occurs. This approach is cumbersome to work with and makes it difficult to create local builds that match CI exactly. This PR updates to using the new single-image workflow in https://github.com/rapidsai/shared-action-workflows/pull/71 that uses a new set of Docker images built just for RAPIDS and eschews cibuildwheel in favor of launching `pip wheel` directly. This approach should be far easier to make maintain and make work locally. It also removes our dependence on modifying upstream pypa manylinux images, which is fairly difficult to maintain.


raydouglass (approved):

rlratzel:
@vyasr thank you for these updates. Can you also update the description to include at least an executive summary (or link to it elsewhere) to what the single image workflow is?  I didn't feel like I had enough background to properly review.

vyasr:
@rlratzel updated. But agreed, I think there's a lot to unpack here and not enough context for someone without background. I am happy to walk you through more of the workflow if you like so that you can see how things are working under the hood now.

rlratzel (approved):
Thanks for adding the background info, @vyasr

PR #3539, opened 2023-05-03
Update cupy dependency

vyasr:
Switch to requiring cupy 12


rlratzel (commented):
I know we have other dependencies with open upper version pins (`nccl`, `numpy`, etc.), but should we start getting into the habit of limiting to at least the next major version, or is there a reason we want it open?  eg `cupy>=12.0.0,<13.0.0a0`

rlratzel (approved):
After some offline discussion, not having an upper bound on the version pin for this and others (unless there's a known reason to have one) seems okay to me.  [Here's](https://iscinumpy.dev/post/bound-version-constraints/#tldr) some additional reading on the topic.

PR #3540, opened 2023-05-04
Updates pytest benchmarks to use synthetic data and multi-GPUs

### LINKED ISSUE #2810: `cugraph.tests.test_dataset.test_load_all` often fails with a timeout error in CI
`cugraph.tests.test_dataset.test_load_all` has caused CI to time-out in at least 3 PRs in the past few days lately.  A `rerun tests` <s>usually</s> sometimes seems to work, but either the test needs to be re-written somehow or the implementation changed.

[Link to a log showing the failure](https://gpuci.gpuopenanalytics.com/job/rapidsai/job/gpuci/job/cugraph/job/prb/job/cugraph-gpu-test/CUDA=11.2,GPU_LABEL=driver-450,LINUX_VER=centos7,PYTHON=3.8/2374/testReport/cugraph.tests/test_dataset/test_load_all/)

### LINKED ISSUE #3282: [FEA]:  MG Testing should use datasets that requier more than 1 GPU

Improvement

Critical (currently preventing usage)

The current set of MG test use a dataset that is too small,  it is even small for SG testing.  
The datasets need to be over 200 million edges, 1 billion edges and larger preferred

Nothing smaller than the cyber.csv , Twitter dataset should be mandatory.

The issues is that we are not uncovering DASK and other issues when using trivial datasets.  Testing MG on a dataset with 2K edges is pointless

Create a MG dataset that every MG test uses.
Update all test to use the new dataset 

_No response_

### Additional context

_No response_

### Code of Conduct

rlratzel:
closes #2810 
closes #3282

* Adds the ability to use datasets read from files on disk and/or RMAT-generated synthetic datasets.
* Adds markers for "file_data" and "rmat_data" for use by benchmark scripts, based on cluster size.
* Adds CLI options for specifying the RMAT scale and edgefactor in order to generate datasets large enough for MNMG runs.
* Adds fixtures for use by `bench_algos.py` benchmarks which will instantiate graph objs based on dataset type and SG or MG markers.
* Updated `Dataset` class to allow instances to be used as test params and properly provide human-readable/deterministic test IDs.
* Added ability for `Dataset` ctor to take a .csf file as input, useful when a metadata.yaml file for a dataset has not been created yet.
* Added options to `get_test_data.sh` in the CI scripts to download a subset of datasets for C++ (to save time/space since most datasets aren't needed), and to only download the benchmark data for python (for use when running benchmarks as tests).


benchmarks/cugraph/pytest-based/bench_algos.py
```diff
@@ -43,41 +47,104 @@ def setFixtureParamNames(*args, **kwargs):
...
-        create_using=graphType,
-        renumber=True)
-
+dataset_fixture_params = gen_fixture_params_product(
+    (directed_datasets +
+     undirected_datasets +
+     [rmat_sg_dataset, rmat_mg_dataset], "ds"))
 
 # Record the current RMM settings so reinitialize() will be called only when a
 # change is needed (RMM defaults both values to False). The --allow-rmm-reinit
```
alexbarghi-nv, line 150: Is this where we should be setting the `cupy` and `torch` allocators as well?

alexbarghi-nv (approved):
👍

benchmarks/cugraph/pytest-based/bench_algos.py
VibhuJawa: `cupy` is set to use rmm pool when we import `cuDF`. 

 We can decide to add `pytorch` based on wether we decide to run `dgl/pyg ` benchmarks here.

VibhuJawa: Two questions:

1. Is this config for SG or for MG too ? 
2. If only SG, it might be worth exploring setting `pool_alloc=True` to speed up CI (if that is a concern).

benchmarks/cugraph/pytest-based/bench_algos.py
```diff
@@ -111,179 +178,228 @@ def reinitRMM(managed_mem, pool_alloc):
...
     # manually call the helper to do so. Ensure the order of the name list
     # passed to it matches if there are >1 params.
     # If the request only contains n params, only the first n names are set.
-    setFixtureParamNames(request, ["dataset", "managed_mem", "pool_allocator"])
-
-    csvFileName = request.param[0]
-    reinitRMM(request.param[1], request.param[2])
-    return utils.read_csv_file(csvFileName)
+    setFixtureParamNames(request, ["managed_mem", "pool_allocator"])
+    reinitRMM(request.param[0], request.param[1])
```
VibhuJawa, line 190: I think `reinitializing` RMM pool will fail if you have anything lying around in RMM memory.

python/cugraph/cugraph/testing/mg_utils.py
```diff
@@ -35,6 +35,9 @@ def start_dask_client(
     device_memory_limit=0.8,
 ):
     dask_scheduler_file = os.environ.get("SCHEDULER_FILE")
+    dask_num_workers = os.getenv("DASK_NUM_WORKERS")
```
VibhuJawa, line 38: Place holder for suggestion for swapping this out to `dask_worker_devices `

VibhuJawa (changes requested):
Did a initial first glance review

python/cugraph/cugraph/testing/mg_utils.py
rlratzel: Thanks, I changed to `dask_worker_devices` in commit [f4433f4](https://github.com/rapidsai/cugraph/pull/3540/commits/f4433f499403aa04e0393e9cef21306cbb5497ea)

benchmarks/cugraph/pytest-based/bench_algos.py
rlratzel: Thanks for looking into this, here's some background:

`bench_algos.py` was originally SG-only, and this PR adds MG coverage. The parameterization of managed mem and pool alloc adds (potentially a lot of) benchmark run time and we've been leaving the defaults in place for automated runs. The use case of actually changing the RMM config multiple times in the same process also did not seem to match a typical user's workflow and sometimes led to problems if we didn't clean up properly (as you pointed out elsewhere), so having pytest run through all combinations is only done using a special CLI option `--allow-rmm-reinit` (see [here](https://github.com/rapidsai/cugraph/blob/f4433f499403aa04e0393e9cef21306cbb5497ea/benchmarks/cugraph/pytest-based/conftest.py#L18)). Otherwise, if a user wants to change the default they just pick the marker they want to use.  The current defaults are pool_alloc `True`, managed_mem `False` (see [here](https://github.com/rapidsai/cugraph/blob/f4433f499403aa04e0393e9cef21306cbb5497ea/benchmarks/cugraph/pytest-based/conftest.py#L58)).  

To answer @VibhuJawa 's questions more directly:
1. it's for SG, I should add a FIXME to figure out how to apply these settings to the dask config when using MG. This is actually a big gap IMO which I'll address either in this PR or a follow-up.
2. pool_alloc defaults to True in conftest.py, but I should add a comment here mentioning that.

To answer @alexbarghi-nv 's question:
> Is this where we should be setting the cupy and torch allocators as well?

I like @VibhuJawa 's suggestion ("We can decide to add pytorch based on wether we decide to run dgl/pyg benchmarks here.")

benchmarks/cugraph/pytest-based/bench_algos.py
rlratzel: Agreed. In fact, based on the observations noted [here](https://github.com/rapidsai/cugraph/pull/3540#discussion_r1198366436) and the suggestion [here](https://github.com/rapidsai/cugraph/blob/f4433f499403aa04e0393e9cef21306cbb5497ea/benchmarks/cugraph/pytest-based/conftest.py#L25), I'm wondering if I should change this from being marker-based (which implies it could run multiple combinations of RMM configurations within a single test suite run) to something set only once using a custom CLI option in `conftest.py`.

I'm going to add a FIXME to say that unless you object.

benchmarks/cugraph/pytest-based/bench_algos.py
VibhuJawa: Thanks for the detailed answer. It makes sense to me.

VibhuJawa (approved):
LGTM

raydouglass (approved):

PR #3559, opened 2023-05-15
add unit test for checking is_symmetric is valid, update documentatio…

### LINKED ISSUE #3149: `is_symmetric` is still ignored in pylibcugraph when `do_expensive_check` is enabled
The `is_symmetric` property allows PLC algos to assume a graph is symmetric/undirected and relies on the user accurately setting it based on the graph contents.  The `do_expensive_check` option causes the code to not make that assumption and instead perform a check at algo runtime to verify the graph is symmetric.

However, the `do_expensive_check` check doesn't seem to be verifying symmetry when `is_symmetric` is set.  Below is a reproducer of that. In the script below, the assertions made on the pagerank results are valid for a symmetric graph, and the expensive check should fail since the graph is not symmetric.  Instead, the assertions fail.

```python
import cupy
from pylibcugraph import ResourceHandle, GraphProperties, SGGraph, pagerank

src_array = cupy.asarray([100, 201, 302], dtype="int32")
dst_array = cupy.asarray([201, 302, 403], dtype="int32")
wgt_array = cupy.asarray([1.0, 1.0, 1.0], dtype="float32")

resource_handle = ResourceHandle()

G = SGGraph(resource_handle,
            GraphProperties(is_symmetric=True, is_multigraph=False),
            src_array,
            dst_array,
            wgt_array,
            store_transposed=True,
            renumber=True,
            do_expensive_check=True,
            )

(vertices, pageranks) = pagerank(resource_handle=resource_handle,
                                 graph=G,
                                 precomputed_vertex_out_weight_vertices=None,
                                 precomputed_vertex_out_weight_sums=None,
                                 initial_guess_vertices=None,
                                 initial_guess_values=None,
                                 alpha=0.85,
                                 epsilon=1.0e-6,
                                 max_iterations=500,
                                 do_expensive_check=True,
                                 )

assert(pageranks.sum() == 1.0)
results = dict(zip(vertices.tolist(),pageranks.tolist()))
assert(results[100] == results[403])
assert(results[201] == results[302])
```

output of running this script:
```
  File "/Projects/repro.py", line 35, in <module>
    assert(results[100] == results[403])
AssertionError
```

ChuckHastings:
Closes #3149 

Seunghwa updated the graph creation in #3533.  This PR adds a unit test which validates that enabling `do_expensive_check` does verify that the input is symmetric.

### DIFF
--- a/cpp/include/cugraph_c/graph.h
+++ b/cpp/include/cugraph_c/graph.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2022, NVIDIA CORPORATION.
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -72,7 +72,7 @@ cugraph_error_code_t cugraph_sg_graph_create(
   const cugraph_type_erased_device_array_view_t* edge_type_ids,
   bool_t store_transposed,
   bool_t renumber,
-  bool_t check,
+  bool_t do_expensive_check,
   cugraph_graph_t** graph,
   cugraph_error_t** error);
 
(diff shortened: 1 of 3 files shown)


seunghwak (approved):
LGTM

PR #3565, opened 2023-05-16
[IMP] Match Default PyG Hop ID Behavior in cuGraph-PyG

### LINKED ISSUE #3563: Update cuGraph-PyG to Label Nodes by Hop ID and Return Sample Sizes

### LINKED ISSUE #3564: Update cuGraph-PyG to Renumber Edge Index using Hash Table

alexbarghi-nv:
Closes #3563 
Closes #3564 

Merge after #3567


python/cugraph-pyg/cugraph_pyg/tests/mg/test_mg_cugraph_store.py
```diff
@@ -184,7 +182,7 @@ def test_renumber_vertices_basic(single_vertex_graph, dask_client):
     )
 
     index = cugraph_store._get_vertex_groups_from_sample(nodes_of_interest)
-    assert index["vt1"].tolist() == sorted(nodes_of_interest.tolist())
+    assert index["vt1"].tolist() == nodes_of_interest.tolist()
```
rlratzel, line 185: No suggestion here, but I'm curious why `sorted()` was removed (or why it wasn't added to both lists being compared), since wouldn't that be a more reliable compare? Or, does order matter?

rlratzel (approved):
LGTM

python/cugraph-pyg/cugraph_pyg/tests/mg/test_mg_cugraph_store.py
alexbarghi-nv: Order does matter now, that is one of the changes in this PR.  Before, the nodes of interest were sorted, but that is not correct - they should be returned in order of appearance.

PR #3567, opened 2023-05-17
include hop as part of the sort criteria for sampling results

### LINKED ISSUE #3552: Update C++ Sampling Code to Sort by Hop ID
Right now, the samples are sorted by batch (which is good) but we need each batch to be sorted by hop in order to avoid making unnecessary copies of the dataframe and resorting during the training phase.

```
Output would look something like this:
batch    src    dst    hop
0           2       0       0
0           0       1       0
0           3       1       0
0           1       0       1
0           0       5       1
0           1       9       1
1           3       6       0
...
```

ChuckHastings:
Closes #3552 

If hop is specified, we'll organize the output by label and hop.


naimnv (approved):

seunghwak (approved):
LGTM

PR #3582, opened 2023-05-18
MG Leiden and MG MIS

### LINKED ISSUE #3469: Perform MG testing of C++ Leiden to validate functionality

naimnv:
- [x] Update MIS to work in MNMG using distributed sampling primitives

- [x] Update Leiden to work in MNMG

closes #3469


cpp/src/community/detail/mis_impl.cuh
```diff
@@ -102,10 +92,7 @@ rmm::device_uvector<vertex_t> compute_mis(
   out_degrees.resize(0, handle.get_stream());
   out_degrees.shrink_to_fit(handle.get_stream());
 
-  thrust::default_random_engine g;
-  size_t seed = 0;
-  if constexpr (multi_gpu) { seed = handle.get_comms().get_rank(); }
-  g.seed(seed);
+  raft::random::RngState rng_state(multi_gpu ? handle.get_comms().get_rank() : 0);
```
ChuckHastings, line 95: This should be a parameter to the `compute_mis` function rather than being constructed here.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -117,22 +104,47 @@ rmm::device_uvector<vertex_t> compute_mis(
...
-    // and for MG extension, select from disributed array remaining vertices
-    thrust::shuffle(
-      handle.get_thrust_policy(), remaining_vertices.begin(), remaining_vertices.end(), g);
 
-    vertex_t nr_candidates =
-      (remaining_vertices.size() < 1024)
-        ? remaining_vertices.size()
-        : std::min(static_cast<vertex_t>((0.50 + 0.25 * loop_counter) * remaining_vertices.size()),
-                   static_cast<vertex_t>(remaining_vertices.size()));
+    vertex_t nr_remaining_vertices_to_check = remaining_vertices.size();
```
ChuckHastings, line 108: I think this is functional.  I wonder if this could be done more efficiently.  Perhaps a FIXME to review the performance here.

My thought on a more efficient implementation would be this:

1. Outside of the loop, on each GPU, randomly shuffle the original remaining_vertices array
2. Outside of the loop, do the `host_scalar_allreduce` call (or just ask the graph for the total number of vertices)
3. Inside the loop, decide on the next batch of 1024 elements to process.  Does it matter that this is truly random?  If not, you could just round-robin grab 1024 vertices to process.  If it does, you could determine which GPUs to select the next 1024 vertices from.
4. For GPU X, you now know (from step 3) how many elements (k) you are using from its remaining_vertices set, take the next k elements from the list to operate on.  Update a local pointer to keep track of where you are so the next iteration continues working down the randomly ordered list.
5. Decrement the global vertex counter by 1024 - you know that's how many vertices you need

I feel like this would be more efficient (reduce the number of comms calls, reduce the number of sorts and other function calls required by the `select_random_vertices` call).

But again, this appears functional, I'm inclined to use this approach and just add a FIXME to improve performance later.

cpp/src/community/detail/refine_impl.cuh
```diff
@@ -540,23 +601,41 @@ refine_clustering(
     //
     // Create decision graph from edgelist
     //
-    constexpr bool storage_transposed = false;
-    constexpr bool multi_gpu          = GraphViewType::is_multi_gpu;
-    using DecisionGraphViewType       = cugraph::graph_view_t<vertex_t, edge_t, false, multi_gpu>;
+    constexpr bool store_transposed = false;
```
ChuckHastings, line 604: Should this come from the `GraphViewType` as well?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -760,7 +760,8 @@ std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> leiden(
   graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
   std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
   size_t max_level    = 100,
-  weight_t resolution = weight_t{1});
+  weight_t resolution = weight_t{1},
+  weight_t theta      = weight_t{1});
```
seunghwak, line 764: Documentation missing for theta.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -804,7 +805,8 @@ std::pair<size_t, weight_t> leiden(
   std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
   vertex_t* clustering,  // FIXME: Use (device_)span instead
   size_t max_level    = 100,
-  weight_t resolution = weight_t{1});
+  weight_t resolution = weight_t{1},
+  weight_t theta      = weight_t{1});
```
seunghwak, line 809: Same here, documentation missing for theta.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2033,6 +2035,13 @@ std::tuple<rmm::device_uvector<size_t>, rmm::device_uvector<vertex_t>> k_hop_nbr
   size_t k,
   bool do_expensive_check = false);
 
+// FIXME: Write doc string
```
seunghwak, line 2038: Yes, please write documentation!!!

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2033,6 +2035,13 @@ std::tuple<rmm::device_uvector<size_t>, rmm::device_uvector<vertex_t>> k_hop_nbr
...
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+rmm::device_uvector<vertex_t> compute_mis(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view);
```
seunghwak, line 2043: Do we need edge weights to find MIS?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2033,6 +2035,13 @@ std::tuple<rmm::device_uvector<size_t>, rmm::device_uvector<vertex_t>> k_hop_nbr
   size_t k,
   bool do_expensive_check = false);
 
+// FIXME: Write doc string
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+rmm::device_uvector<vertex_t> compute_mis(
```
seunghwak, line 2040: And should we better rename this to `maximal_independent_set` to be consistent? We have `pagerank` instead of `compute_pagerank`.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -102,10 +92,7 @@ rmm::device_uvector<vertex_t> compute_mis(
   out_degrees.resize(0, handle.get_stream());
```
seunghwak, line 92:
```
  thrust::for_each(
    handle.get_thrust_policy(),
    vertex_begin,
    vertex_end,
    [out_degrees = raft::device_span<edge_t const>(out_degrees.data(), out_degrees.size()),
     ranks       = raft::device_span<vertex_t>(ranks.data(), ranks.size()),
     v_first     = graph_view.local_vertex_partition_range_first()] __device__(auto v) {
      auto v_offset = v - v_first;
      if (out_degrees[v_offset] == 0) { ranks[v_offset] = std::numeric_limits<vertex_t>::lowest(); }
    });
```

You can use `thrust::transform_if` for this. It will be more specific than for_each (one of the RAPIDS reviewer mentioned `for_each` as glorified for loop, better avoid it if there exists a better matching thrust algorithm).

seunghwak: `transform_if` much better documents the intention (transform to std::numeric_limits<vertex_t>::lowest() if the degree is 0) without comments than `for_each`.

cpp/src/community/detail/mis_impl.cuh
seunghwak: Yes, and if run Leiden multiple times with the same `rng_state`, we will get different results in each run and this will be helpful in implementing ECG as well.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -117,22 +104,47 @@ rmm::device_uvector<vertex_t> compute_mis(
...
+                                                             handle.get_stream());
+    }
+
+    vertex_t nr_candidates = (nr_remaining_vertices_to_check < 1024)
+                               ? nr_remaining_vertices_to_check
+                               : std::min(static_cast<vertex_t>((0.50 + 0.25 * loop_counter) *
+                                                                nr_remaining_vertices_to_check),
+                                          nr_remaining_vertices_to_check);
+
+    // FIXME: (nr_remaining_vertices_to_check < 1024)? avoid calling select_random_vertices
```
seunghwak, line 122: This FIXME is out-dated, right?

cpp/src/community/detail/mis_impl.cuh
seunghwak: I guess here, we are selecting at least 1024, not just 1024. As the loop count grows, we may select all the remaining vertices.

And I agree with Chuck that "inclined to use this approach and just add a FIXME to improve performance later." if this turns out to be a performance bottleneck later.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -117,22 +104,47 @@ rmm::device_uvector<vertex_t> compute_mis(
...
+
+    thrust::set_difference(handle.get_thrust_policy(),
+                           remaining_vertices.begin(),
+                           remaining_vertices.end(),
+                           d_sampled_vertices.begin(),
+                           d_sampled_vertices.end(),
+                           non_candidate_vertices.begin());
 
     // Set temporary ranks of non-candidate vertices to std::numeric_limits<vertex_t>::lowest()
     thrust::for_each(
```
seunghwak, line 144: You can implement this using `thrust::transform_if` if you use `thrust::permutation_iterator` for the output. But permutation iterator is a less known trick, I won't argue that it is a better option, just FYI.

cpp/src/community/detail/mis_impl.cuh
```diff
@@ -117,22 +104,47 @@ rmm::device_uvector<vertex_t> compute_mis(
...
       handle.get_thrust_policy(),
-      remaining_vertices.begin(),
-      remaining_vertices.end() - nr_candidates,
+      non_candidate_vertices.begin(),
+      non_candidate_vertices.end(),
       [temporary_ranks =
          raft::device_span<vertex_t>(temporary_ranks.data(), temporary_ranks.size()),
        v_first = graph_view.local_vertex_partition_range_first()] __device__(auto v) {
```
seunghwak, line 150: Regarding the comment below

```

    //
    // Find maximum rank outgoing neighbor for each vertex
    // (In case of Leiden decision graph, each vertex has at most one outgoing edge)
    //
```

If this code is for general MIS and you plan to expose this in the public API, better remove any Leiden specific comments.

seunghwak (commented):
Review part 1/2

cpp/tests/CMakeLists.txt
```diff
@@ -476,6 +476,16 @@ if(BUILD_CUGRAPH_MG_TESTS)
     # - MG LOUVAIN tests --------------------------------------------------------------------------
     ConfigureTestMG(MG_LOUVAIN_TEST community/mg_louvain_test.cpp)
 
+    ###############################################################################################
+    # - MG LEIDEN tests --------------------------------------------------------------------------
+    ConfigureTestMG(MG_LEIDEN_TEST community/mg_leiden_test.cpp)
+
+
+
```
seunghwak, line 484: Why 3 empty lines? I think we use just 1.

cpp/tests/community/mg_mis_test.cu
```diff
@@ -0,0 +1,255 @@
...
+}
+
+TEST_P(Tests_MGMaximalIndependentSet_Rmat, CheckInt32Int64FloatFloat)
+{
+  run_current_test<int32_t, int64_t, float, int>(
+    override_Rmat_Usecase_with_cmd_line_arguments(GetParam()));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+  file_test_pass,
```
seunghwak, line 232: Why did you name this to `file_test_pass`? We typically name this to `file_test`.

cpp/tests/community/mg_mis_test.cu
```diff
@@ -0,0 +1,255 @@
...
+{
+  run_current_test<int32_t, int32_t, float, int>(
+    override_Rmat_Usecase_with_cmd_line_arguments(GetParam()));
+}
+
+TEST_P(Tests_MGMaximalIndependentSet_Rmat, CheckInt32Int64FloatFloat)
+{
+  run_current_test<int32_t, int64_t, float, int>(
+    override_Rmat_Usecase_with_cmd_line_arguments(GetParam()));
+}
```
seunghwak, line 229: No int64, int64 tests?

seunghwak: This is necessary for large scale testing.

cpp/src/community/detail/mis_impl.cuh
naimnv: Resolved.

ChuckHastings (approved):

cpp/include/cugraph/algorithms.hpp
seunghwak: Not really a comment for this PR but `rng_state` appear right after handle in `sample_neighbors_adjacency_list` and `sample_neighbors_edgelist` but appears right before the last 3 Boolean input parameters in `uniform_neighbor_sample`.

@ChuckHastings This can be confusing and we need some consistency in input parameter ordering.

I think `rng_state` appearing after graph data makes little more logical sense (graph data is more essential than rng_state in computing Leiden).

cpp/include/cugraph_c/community_algorithms.h
```diff
@@ -130,6 +131,7 @@ cugraph_error_code_t cugraph_louvain(const cugraph_resource_handle_t* handle,
  * @return error code
  */
 cugraph_error_code_t cugraph_leiden(const cugraph_resource_handle_t* handle,
+                                    cugraph_rng_state_t* rng_state,
                                     cugraph_graph_t* graph,
                                     size_t max_level,
                                     double resolution,
```
seunghwak, line 137: Not exposing theta here?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1992,6 +2009,24 @@ std::tuple<rmm::device_uvector<size_t>, rmm::device_uvector<vertex_t>> k_hop_nbr
...
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param graph_view Graph view object.
+ * @return A device vector containing vertices found in the maximal independent set
+ */
+
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+rmm::device_uvector<vertex_t> maximal_independent_set(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::random::RngState& rng_state);
```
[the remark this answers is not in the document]
naimnv: Added missing doc for rng_state.

seunghwak (approved):

rlratzel (approved):

Issue #1347, opened 2021-01-21
Python tests should have a shared set of fixtures


rlratzel:
Many of the tests do very similar setup steps, including iterating over a set of datasets, creating a specific graph type, creating a specif Nx graph type, running an algo and checking/comparing results.  The steps prior to running the algo are almost identical across most tests, yet each test has its own set of fixtures.

The shared util file defines the dataset params (which is good), but the actual fixtures could be shared as well.  This should allow for a great deal of cleanup when applied to all python tests.

* Add/update a fixture for reading in datasets and returning a graph object
  * For use with all datasets, subset of datasets, etc.
  * Have pytest options for which datasets to run
    * This also includes datasets which are applied to either a SG graph object or a MG graph object, and will be used when determining if a test will be providing SG or MG coverage.
    * Datasets/Graph provided by fixtures will have markers assigned to them. This is needed primarily for SG inputs or MG inputs for testing SG or MG algos, or both.
  * Add this to a shared “fixtures” module, configured to scope “session”
* Add/update a fixture for returning a Nx graph object
  * Calls fixture above to read in dataset and return cugraph obj
  * Add this to a shared “fixtures” module, configured to scope “session”
* Add/update a fixture for per-algo Nx runs with params for comparisons
  * Unique to a particular algo comparison (scope=”module”)

rlratzel:
This is on-hold but should be addressed in upcoming refactorings.

rlratzel:
Some of the proposed changes mentioned here have either been done for a subset of tests or are still being planned.  Since this issue is relatively stale, we'll close it and re-open as one or more smaller issues when we're closer to continuing work on these changes.

PR #3596, opened 2023-05-22
Enables MG python tests using a single-GPU LocalCUDACluster in CI

### LINKED ISSUE #3413: Run MG tests in CI using a single-GPU `LocalCUDACluster`
This adds coverage for the dask/MG code paths in CI
* Ensure new sg/mg labels are compatible
* Update CI scripts to run MG
* Ensure MG tests that are run oin Ci are appropriate

```[tasklist]
### Tasks
- [ ] #2714 
```

rlratzel:
closes #3413 

This PR enables MG python tests using a single-GPU `LocalCUDACluster` in CI by setting the `DASK_WORKER_DEVICES` to `0`.  This does not affect SG tests, which continue to run in CI as they previously did.

PR #3540 added support for `DASK_WORKER_DEVICES` to the pytest fixture used in python MG tests, allowing CI scripts (and developers) to restrict workers to specific devices, which should now allow single-GPU CI runs to cover the MG/dask code paths in python.

_Note: despite the changes here to now run python MG tests, it should be noted that this isn't actual multi-GPU test coverage in `libcugraph` since multiple GPUs are not communicating and the code to set up comms, distribute graph data, etc. is not being exercised using >1 GPU._


ajschmidt8 (approved):

VibhuJawa (approved):
Approving but we should wait before merging to see amount of time added , CC: @rlratzel

rlratzel:
I ran the MG tests using `DASK_WORKER_DEVICES=0` locally and observed they took ~20 minutes, meaning our CI runs will be _at least_ 20 minutes longer now assuming the CI system is equal to or slower than my system.

This isn't ideal but it's a decent tradeoff for getting the additional coverage.

A few notes:
* This PR does not remove any of the conditional skip markers that existed that check for >1 GPU, so there's still some missing coverage. But even with those in place, we're still adding significant coverage.  Those will be removed in a future PR.
* We still have a lot of testing tech debt to clean up, so making the behavior of using the `mg` marker be "uses >1 GPU" and `sg` be "uses only 1 GPU" _without setting `DASK_WORKER_DEVICES`_ will be done as part of closing #1347

BradReesWork (approved):

PR #3599, opened 2023-05-23
[BUG] Fix Incorrect File Selection in cuGraph-PyG Loader

alexbarghi-nv:
The current version of the cuGraph-PyG loader does not properly handle the case where samples were previously computed and saved to disk.  It also compares the batch ids to determine when it is done processing, which is also incorrect, since batches can be loaded in any order.

This PR updates the cuGraph-PyG loader to use only the list of input files to determine when it is done.  It also allows specifying the input file list to support cases where previously-computed batches are loaded from disk.  The loader tests are updated to verify the new behavior and ensure that loading previously-computed batches works as expected.


VibhuJawa (approved):
LGTM

BradReesWork (approved):

PR #3600, opened 2023-05-23
[BUG] Skip Empty Partitions in Bulk Sample Writing

alexbarghi-nv:
Currently, when the bulk sampler writes its samples, it does not check whether the partition is empty.  This causes all of the empty partitions to collide and attempt to write to the same (empty) file at once.  This writing should not occur at all.

This PR adds a check that skips writing if the partition is empty.

This should resolve the nightly test failures.


VibhuJawa (approved):
LGTM. Thanks for triaging this @alexbarghi-nv

BradReesWork (approved):

PR #3604, opened 2023-05-24
[BUG] Critical: Add the Required Hop ID Change to CuGraphStore

alexbarghi-nv:
I forgot to add some of the changes I made to #3565; these changes update `CuGraphStore` to properly handle renumbering vertices and edges when they are not in ascending order.  They also contain the test changes that verify that hop id is being handled properly.


python/cugraph-pyg/cugraph_pyg/sampler/cugraph_sampler.py
```diff
@@ -38,24 +38,27 @@ def _count_unique_nodes(
...
+        node_position: str ('src' or 'dst')
+            Whether to examine source or destination nodes.
+    <<<<<<< HEAD
+    =======
+
+        Returns
+        -------
+        int
+            The number of unique nodes of the given node type.
+    >>>>>>> 64690fe39e5f2ae5559ebfac7c9d3ee49e96afe6
```
rlratzel, line 61: Looks like an unresolved merge?

rlratzel (changes requested):
I only see an update to a docstring and an additional test, was there supposed to be a code change to address a bug here?

PR #3605, opened 2023-05-24
[BUG] Critical: Fix cuGraph-PyG Edge Index Renumbering for Single-Edge Graphs

alexbarghi-nv:
The changes for multi-edge graphs to accommodate the new hop id behavior were not properly applied to single-edge graphs.  This PR resolves that.


rlratzel (approved):
I'm not familiar with this code so I'm approving based on not seeing any really obvious bugs.

PR #3614, opened 2023-05-25
Add new cugraph-nx package (networkx backend using pylibcugraph)

eriknw:
`cugraph-nx` is working with two algorithm (`betweenness_centrality`, `edge_betweenness_centrality`).

It can efficiently (i.e., faster than `cugraph` in my benchmarks) convert from (and to) networkx graphs, and runs (and passes) networkx tests using networkx dispatching machinery.

It renumbers vertex labels, which we handle with python dicts.

It only depends on `networkx`, `cupy`, and `pylibcugraph`. `cupy` is a transitive dependency from `pylibcugraph`, so we don't actually need to list it as a dependency.

I think it's nice the way we depend on `networkx`, which allows us to use its exceptions and utilities.

There is still plenty to do (and I have a few questions), but I think it's a good start and gets cugraph in on the NetworkX dispatching party.

I was planning on including `cugraph-pg` with this PR, but now I think it's probably better to have that in a new PR.


python/cugraph-nx/cugraph_nx/algorithms/components/strongly_connected.py
```diff
@@ -0,0 +1,36 @@
...
+from cugraph_nx.utils.decorators import not_implemented_for
+
+__all__ = ["is_strongly_connected"]
+
+
+@not_implemented_for("undirected")
+def is_strongly_connected(G):
+    G = to_directed_graph(G)
+    N = len(G)
+    if N == 0:
```
BradReesWork, line 28: why not check graph size on the Nx graph before converting it to GPU?

python/cugraph-nx/cugraph_nx/classes/graph.py
```diff
@@ -0,0 +1,133 @@
...
+if TYPE_CHECKING:
+    import cupy as cp
+
+NodeType = Hashable
+
+
+__all__ = ["Graph"]
+
+
+class Graph:
```
BradReesWork, line 29: who might want a comment that Graph saves data in COO format

python/cugraph-nx/cugraph_nx/classes/graph.py
```diff
@@ -0,0 +1,133 @@
...
+    def __contains__(self, n) -> bool:
+        if self.key_to_id is not None:
+            container = self.key_to_id
+        else:
+            container = range(self._N)
+        try:
+            return n in container
+        except TypeError:
+            return False
+
```
BradReesWork, line 87: how many of these function are truly used within an algorithms?  There is value to those functions, but it seems like Nx could just leverage their internal representation first for those calls.   

Now if the thought that Nx would totally off load all functions and just be a broker?

eriknw: I think methods on graphs are "nice to have", but most of them aren't strictly necessary for the example workloads we're after. I only added the necessary ones and the "easy to implement" ones. Long term, having methods on graphs will let users copy/paste examples from documentation, docstrings, etc. and update to use cugraph-nx. It gives users a familiar experience that may be useful when interacting with graphs in a notebook. I don't expect to expend significant effort in the near term to implement the missing methods.

python/cugraph-nx/cugraph_nx/algorithms/components/strongly_connected.py
eriknw: Being more intellegent when to convert graphs is a good idea, but this logic likely belongs elsewhere (TBD). On the other hand, small graphs should be quick regardless.

eriknw:
@rlratzel should we add anything to `dependencies.yaml`?

Also, @betochimas and I last week discovered we need `pytest-mpl` installed to run the networkx tests, so we should probably add this to the conda environment files.

python/cugraph-nx/run_nx_tests.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2023, NVIDIA CORPORATION.
+# NETWORKX_GRAPH_CONVERT=cugraph NETWORKX_BACKEND_TEST_EXHAUSTIVE=True pytest --pyargs networkx "$@"
+NETWORKX_TEST_BACKEND=cugraph NETWORKX_TEST_FALLBACK_TO_NX=True pytest --pyargs networkx "$@"
```
rlratzel, line 5: What's the behavior of `NETWORKX_TEST_FALLBACK_TO_NX`? I'm just wondering if this will silently revert to Nx and pass the test at times when it should be using cugraph and failing because there's a bug we need to fix.

python/cugraph-nx/cugraph_nx/tests/test_convert.py
```diff
@@ -0,0 +1,202 @@
...
+    G = graph_class()
+    cG = cnx.from_networkx(G, **kwargs)
+    H = cnx.to_networkx(cG)
+    assert G.number_of_nodes() == cG.number_of_nodes() == H.number_of_nodes() == 0
+    assert G.number_of_edges() == cG.number_of_edges() == H.number_of_edges() == 0
+    assert cG.edge_values == cG.edge_masks == cG.node_values == cG.node_masks == {}
+    assert G.graph == cG.graph == H.graph == {}
+
+
+def test_convert():
```
rlratzel, line 51: Wow, this test is testing a lot.  Would this be better as separate tests more focused on specific aspects of converting? OTOH, with `pytest --pdb` it may not matter much in practice, so I'm not too concerned about diagnosing a failing test that's this big.

rlratzel (commented):
This is great, thanks for the attention to detail.

python/cugraph-nx/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+# limitations under the License.
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--bench",
+        action="store_true",
+        default=False,
+        help="Run benchmarks (sugar for --benchmark-enable) and skip other tests",
+    )
```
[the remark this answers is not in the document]
rlratzel: This is somewhat outdated now given that the `--all` option was added, but I do think that using a personalized `pytest.ini` file and the `pytest -c` option could still be an alternative to these options.

python/cugraph-nx/cugraph_nx/convert.py
```diff
@@ -0,0 +1,496 @@
...
+    node_dtypes: Dtype | dict[AttrKey, Dtype | None] | None = None,
+    preserve_all_attrs: bool = False,
+    preserve_edge_attrs: bool = False,
+    preserve_node_attrs: bool = False,
+    preserve_graph_attrs: bool = False,
+    as_directed: bool = False,
+    name: str | None = None,
+    graph_name: str | None = None,
+) -> cnx.Graph:
+    """Convert a networkx graph to cugraph_nx graph; can convert all attributes.
```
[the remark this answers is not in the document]
eriknw: We can probably remove `can convert all attributes` from the first line. It is meant to reinforce that we are flexible enough to handle networkx graphs with multiple attributes. I did mean `"can"`.

python/cugraph-nx/cugraph_nx/classes/graph.py
```diff
@@ -0,0 +1,586 @@
...
+    edge_values: dict[AttrKey, cp.ndarray[EdgeValue]]
+    edge_masks: dict[AttrKey, cp.ndarray[bool]]
+    node_values: dict[AttrKey, cp.ndarray[NodeValue]]
+    node_masks: dict[AttrKey, cp.ndarray[bool]]
+    key_to_id: dict[NodeKey, IndexValue] | None
+    _id_to_key: dict[IndexValue, NodeKey] | None
+    _N: int
+    graph: dict
+
+    __networkx_plugin__: ClassVar[str] = "cugraph"
```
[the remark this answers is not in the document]
eriknw: Done. It's how networkx knows to dispatch our objects to us.

python/cugraph-nx/cugraph_nx/classes/graph.py
```diff
@@ -0,0 +1,586 @@
...
+        edge_values: dict[AttrKey, cp.ndarray[EdgeValue]] | None = None,
+        edge_masks: dict[AttrKey, cp.ndarray[bool]] | None = None,
+        node_values: dict[AttrKey, cp.ndarray[NodeValue]] | None = None,
+        node_masks: dict[AttrKey, cp.ndarray[bool]] | None = None,
+        *,
+        key_to_id: dict[NodeKey, IndexValue] | None = None,
+        id_to_key: dict[IndexValue, NodeKey] | None = None,
+        **attr,
+    ) -> Graph:
+        self = object.__new__(cls)
```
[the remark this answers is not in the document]
eriknw: Heh, I'm not sure if using `object.__new__` is common enough for it to be idiomatic. I think reasonable choices are `self`, `obj`, `rv`, `instance`, or `result`. 

Here are example in the wild using `self`:
- https://github.com/matplotlib/matplotlib/blob/3ab6e1f409796543155d79f09e9e61d26181f5dd/lib/matplotlib/dviread.py#L882
- https://github.com/python/cpython/blob/4890bfe1f906202ef521ffd327cae36e1afa0873/Lib/_pydatetime.py#L699
- https://github.com/python/cpython/blob/4890bfe1f906202ef521ffd327cae36e1afa0873/Lib/_pydecimal.py#L555

python/cugraph-nx/cugraph_nx/classes/graph.py
```diff
@@ -0,0 +1,586 @@
...
+        self.col_indices = col_indices
+        self.edge_values = {} if edge_values is None else dict(edge_values)
+        self.edge_masks = {} if edge_masks is None else dict(edge_masks)
+        self.node_values = {} if node_values is None else dict(node_values)
+        self.node_masks = {} if node_masks is None else dict(node_masks)
+        self.key_to_id = None if key_to_id is None else dict(key_to_id)
+        self._id_to_key = None if id_to_key is None else dict(id_to_key)
+        self._N = op.index(N)  # Ensure N is integral
+        self.graph = self.graph_attr_dict_factory()
+        self.graph.update(attr)
```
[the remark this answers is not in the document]
eriknw: `graph` is a networkx attribute. Comment added.

python/cugraph-nx/cugraph_nx/convert.py
```diff
@@ -0,0 +1,496 @@
...
+    if node_attrs is not None:
+        if isinstance(node_attrs, Mapping):
+            # Copy so we don't mutate the original
+            node_attrs = dict(node_attrs)
+        else:
+            node_attrs = {node_attrs: None}
+
+    if graph.__class__ in {nx.Graph, nx.DiGraph}:
+        # This is a NetworkX private attribute, but is much faster to use
+        adj = graph._adj
```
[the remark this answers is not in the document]
eriknw: I don't think so. If the object is not _exactly_ a networkx graph, then we use the more generic `.adj`, which is slower.

As the code says, we use this to be (much) faster, but it does add the burden that we may need to change if networkx changes this. I'm happy to take on this burden for the increased performance.

python/cugraph-nx/cugraph_nx/convert.py
```diff
@@ -0,0 +1,496 @@
...
+        adj = {k: dict(v) for k, v in adj.items()}
+
+    N = len(adj)
+    if (
+        not preserve_edge_attrs
+        and not edge_attrs
+        # Faster than graph.number_of_edges() == 0
+        or next(concat(rowdata.values() for rowdata in adj.values()), None) is None
+    ):
+        edge_attrs = None
```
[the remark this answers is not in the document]
eriknw: Done.

python/cugraph-nx/run_nx_tests.sh
eriknw: Comment added.

Good question. It reverts to networkx if we don't implement a function, or if `can_run` returns False or our algorithm raises `NotImplementedError`. This is helpful to get increased testing and coverage. Other exceptions will raise, so I don't think this will fail silently when we don't want it to. But, we may want to toggle this option during development. Also, it would be nice to ensure our algorithm files have decent coverage when running networkx tests to make sure they are being exercised.

python/cugraph-nx/cugraph_nx/tests/test_convert.py
eriknw: Heh, yeah, I know. It's probably better to break it up, but then we need to figure out how to name all the tests ;)

python/cugraph-nx/cugraph_nx/convert.py
```diff
@@ -0,0 +1,496 @@
...
+    rv.graph.update(G.graph)
+    return rv
+
+
+def to_graph(
+    G,
+    edge_attr: AttrKey | None = None,
+    edge_default: EdgeValue | None = 1,
+    edge_dtype: Dtype | None = None,
+) -> cnx.Graph | cnx.DiGraph:
```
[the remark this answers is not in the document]
eriknw: Ah, but there are _three_ versions: `to_graph`, `to_directed_graph` and `to_undirected_graph`, so `create_using` won't work for how we (plan on) using them. Speaking of which, these need documented to explain how they're used.

Right now, using these helper function is completely optional. When calling algorithms not through networkx, these are used to ensure we have a valid graph and will convert if needed. I think it's a useful, _simple_ pattern for functions that only need graphs with up to a single attribute. `to_graph` is useful if the input is allowed to be directed or undirected. The other two specify the directedness.

python/cugraph-nx/cugraph_nx/classes/graph.py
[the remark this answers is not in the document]
rlratzel: I had to do a double-take at first but it's not a big deal.  I personally do like `instance` as you proposed since it makes it very obvious.

python/cugraph-nx/cugraph_nx/convert.py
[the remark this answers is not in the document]
rlratzel: ok thanks, just checking.

python/cugraph-nx/cugraph_nx/convert.py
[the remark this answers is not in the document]
rlratzel: yeah, I was thinking the default `create_using=None` (the value when not specified) would be equivalent to `to_graph`, then `create_using=cnx.Graph` would be `to_undirected_graph` and `create_using=cnx.DiGraph` would be `to_directed_graph`.  I don't feel strongly though if you think it's better to keep them as-is.

I do agree that docstrings would be nice.

python/cugraph-nx/run_nx_tests.sh
rlratzel: ok, my fear (maybe I'm being paranoid) is a bug in our plugin that prevents it from being loaded somehow, and everything falls back to native Nx and silently passes.

If we run without this in dev then that's probably adequate.

python/cugraph-nx/cugraph_nx/tests/test_convert.py
rlratzel: ok yeah, coverage is the priority and it looks good from that perspective, so maybe a FIXME would suffice.

python/cugraph-nx/README.md
```diff
@@ -0,0 +1,15 @@
+# cugraph_nx
+
+## Description
+[RAPIDS](https://rapids.ai) cugraph-nx is an experimental [backend to NetworkX](https://networkx.org/documentation/stable/reference/classes/index.html#backends)
+with minimal dependencies (`networkx`, `cupy`, and `pylibcugraph`).
+It is a work in progress, and cugraph should typically be used instead.
```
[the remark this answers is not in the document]
rlratzel: on second thought...
I think we should do whatever a NetworkX user would expect, since this is after all a plugin to NetworkX.  I'm assuming that means do NOT raise an experimental warning, but maybe it's worth further discussion.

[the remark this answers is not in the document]
rlratzel: I still think we should drop the last sentence though :)

python/cugraph-nx/run_nx_tests.sh
eriknw: I'll add `--cov=cugraph_nx/algorithms --cov-report term-missing` to this pytest command to show coverage of algorithms, which will let us know if they run. I agree this is an important sanity check.

python/cugraph-nx/cugraph_nx/convert.py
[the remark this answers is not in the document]
eriknw: I see. These aren't meant to be user-facing functions (yet), so I'd like to keep them as is for now, but I'll rename them with `_` prefix to make them private. Let's see how we use them and like them going forward, and I think it should be easy to change them if, say, we prefer your proposal and want to make the functions public b/c they're convenient in practice.

python/cugraph-nx/cugraph_nx/classes/graph.py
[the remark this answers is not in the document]
eriknw: Changing to `new_graph` in graph.py and `instance` in decorators.py (yes, I used this trick twice).

python/cugraph-nx/cugraph_nx/tests/test_convert.py
eriknw: Adding FIXME comment.

At a glance, breaking this up looks kind of annoying, because we sometimes mutate the current graph to change its characteristics. And you're exactly right: this was done to increase coverage by testing more cases. Having unit tests be, well, a single unit is nice to have, but having tests is better than no tests, so I'm fine with adding the FIXME for now.

eriknw:
@rlratzel, the last commit hopefully addresses all current comments. Shall we aim to merge soon?

python/cugraph-nx/README.md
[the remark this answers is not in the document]
eriknw: I'm so glad you looked at the README! I had completely forgotten about it. Last sentence deleted, and generally improved the README (but still lots to do).

rlratzel (approved):
Looks great, eager to get this merged, thanks!

raydouglass:

python/cugraph-nx/cugraph_nx/tests/__init__.py
BradReesWork, line 1: This should have the license header

BradReesWork (changes requested):

python/cugraph-nx/cugraph_nx/tests/__init__.py
eriknw: Thanks for the careful review. I'm not sure an empty file needs the license header. If the license header belongs here, then it also belongs in these files:
```
./python/cugraph-service/server/cugraph_service_server/testing/__init__.py
./python/cugraph-service/tests/__init__.py
./python/cugraph/cugraph/dask/link_analysis/__init__.py
./python/cugraph/cugraph/dask/link_prediction/__init__.py
./python/cugraph/cugraph/dask/components/__init__.py
./python/cugraph/cugraph/dask/comms/__init__.py
./python/cugraph/cugraph/dask/structure/__init__.py
./python/cugraph/cugraph/dask/common/__init__.py
./python/cugraph/cugraph/dask/traversal/__init__.py
./python/cugraph/cugraph/experimental/compat/__init__.py
./python/cugraph/cugraph/experimental/link_prediction/__init__.py
./python/cugraph/cugraph/experimental/components/__init__.py
./python/cugraph/cugraph/experimental/datasets/metadata/__init__.py
./python/cugraph/cugraph/experimental/structure/__init__.py
./python/cugraph/cugraph/datasets/metadata/__init__.py
./python/cugraph-dgl/cugraph_dgl/utils/__init__.py
./python/pylibcugraph/pylibcugraph/components/__init__.py
./python/pylibcugraph/pylibcugraph/utilities/__init__.py
./python/pylibcugraph/pylibcugraph/internal_types/__init__.py
./python/pylibcugraph/pylibcugraph/_cugraph_c/__init__.py
./python/pylibcugraph/pylibcugraph/structure/__init__.py
./benchmarks/shared/python/cugraph_benchmarking/__init__.py
```
I was following established practices.

rlratzel: @BradReesWork if it's okay, I'll open a separate PR to address the `__init__.py` file in this PR as well as all other empty ones in the entire repo. That will make us more consistent with other RAPIDS repos (I just checked `cudf`) which include at least a copyright line.
I'll follow up here with the PR once it's up.

eriknw: These files identified via:
```bash
find . -name '__init__.py' -print -type f | xargs wc | grep -v skbuild | grep -v cpp.build | grep ' 0  *0  *0'
```

rlratzel: Here's the followup PR: https://github.com/rapidsai/cugraph/pull/3799

BradReesWork (approved):

PR #3618, opened 2023-05-26
Add `python/cugraph-pg`: new experimental PropertyGraph package

eriknw:
This is the first task for https://github.com/rapidsai/cugraph-pg/issues/10.

Much more functionality coming soon!


python/cugraph-pg/pyproject.toml
```diff
@@ -0,0 +1,188 @@
...
+include = [
+    "cugraph_pg*",
+    "cugraph_pg.*",
+]
+
+[tool.black]
+line-length = 88
+target-version = ["py39", "py310", "py311"]
+
+[tool.isort]
```
rlratzel, line 58: It would be nice to have `isort` added to our pre-commit checks

rlratzel (commented):
Thanks, some suggestions and general comments below.
I may come back and review `core.py` more thoroughly before approving since I don't think I mentally captured all the class hierarchy. This also seems like a lot of code to commit without accompanying test coverage - will there be some tests added in this PR?

rlratzel:
@eriknw and I discussed this offline and decided to close it.  If cugraph ever decides to break out `PropertyGraph` into a separate package, we'll revisit the changes done in this PR.

PR #3619, opened 2023-05-26
fix logic for shuffling results

ChuckHastings:
@alexbarghi-nv found a bug in sampling during python testing.  The process of shuffling sampled edges for output formatting was corrupting the label associated with each sampled edge.  This PR fixes this bug.


cpp/src/sampling/detail/sampling_utils_impl.cuh
```diff
@@ -827,30 +840,33 @@ shuffle_and_organize_output(
...
+                mem_frugal_threshold,
+                handle.get_stream());
+
+              std::vector<size_t> h_tx_value_counts(d_tx_value_counts.size());
+              raft::update_host(h_tx_value_counts.data(),
+                                d_tx_value_counts.data(),
+                                d_tx_value_counts.size(),
+                                handle.get_stream());
+
+              handle.sync_stream();
```
naimnv, line 868: We can call `cugraph::groupby_and_count` only once before exponential if/else checks if we use an index array and then use ` `thrust::gather`  or thrust::permutation_iterator`  to rearrange 

```
majors
minors
weights
edge_ids
edge_types
hops
```
using  a zip iterator.

naimnv (commented):
Looks good to me, but we can shirk the code volume I think.

cpp/src/sampling/detail/sampling_utils_impl.cuh
seunghwak: Or isn't it possible to make the body of each if-else cases as a template function. If I am not mistaken, the only differences between the cases is the iterators we are zipping using `thrust::make_zip_iterator(...)`

seunghwak (approved):
We may consider refactoring this code to reduce code duplication in the next release.

cpp/src/sampling/detail/sampling_utils_impl.cuh
naimnv: the only differences between the cases is the iterators we are zipping using thrust::make_zip_iterator(...) - that's correct. 
It can be merged as it and can be changed at later point as Chuck might be away.

naimnv (approved):

PR #3628, opened 2023-06-02
Add Benchmark for Bulk Sampling

alexbarghi-nv:
Adds the benchmark for bulk sampling.


alexbarghi-nv:
Dependency file generator reverted to 23.06; don't merge this until that issue is resolved

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,618 @@
...
+        label_df = pd.DataFrame({
+            'node': np.arange(num_labeled_nodes),
+            # 'label': generator.integers(0, num_labels - 1, num_labeled_nodes).astype('float32')
+        })
+        
+        dask_label_df = ddf.from_pandas(label_df)
+        del label_df
+        gc.collect()
+
+        dask_label_df = dask_cudf.from_dask_dataframe(dask_label_df)
```
VibhuJawa, line 348: Minor nitpick of moving this to a function so that we can use it at other places.

python/cugraph-service/server/pyproject.toml
```diff
@@ -31,7 +31,7 @@ dependencies = [
     "numpy>=1.21",
     "rmm==23.6.*",
     "thriftpy2",
-    "ucx-py==0.32.*",
+    "ucx-py==0.33.*",
```
VibhuJawa, line 34: Is this supposed to be part of the PR ?

VibhuJawa (changes requested):
Minor nitpick about adding a function, bulk sampling changes look good. 

Also lets remove what i think are un related changes in `.toml` from the PR.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: resolved

python/cugraph-service/server/pyproject.toml
alexbarghi-nv: fixed the bad merge

VibhuJawa:
@alexbarghi-nv , Can we try to merge this in ?

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+        required=False,
+        default='10_25',
+    )
+
+    parser.add_argument(
+        '--batch_sizes',
+        type=str,
+        help='Comma separated list of batch sizes (i.e. 500,1000)',
+        required=False,
+        default='500,1000'
```
VibhuJawa, line 524:
```suggestion
        default='512,1024'
```

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+    logging.basicConfig()
+
+    args = get_args()
+    fanouts = [[int(f) for f in fanout.split('_')] for fanout in args.fanouts.split(',')]
+    datasets = args.datasets.split(',')
+    batch_sizes = [int(b) for b in args.batch_sizes.split(',')]
+    seeds_per_call_opts = [int(s) for s in args.seeds_per_call_opts.split(',')]
+    dask_worker_devices = [int(d) for d in args.dask_worker_devices.split(',')]
+
+    client, cluster = start_dask_client(dask_worker_devices=dask_worker_devices, jit_unspill=False, rmm_pool_size=28e9, rmm_async=True)
```
VibhuJawa, line 573: Below does not work as `rmm_async` is not available in `start_dask_client`
```suggestion
    client, cluster = start_dask_client(dask_worker_devices=dask_worker_devices, jit_unspill=False, rmm_pool_size=28e9, rmm_async=True)
```

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+
+    return dask_edgelist_df, dask_label_df, node_offsets, edge_offsets, total_num_nodes
+
+
+def load_disk_dataset(dataset, dataset_dir='.', reverse_edges=True, replication_factor=1):
+    path = os.path.join(dataset_dir, dataset)
+    parquet_path = os.path.join(path, 'parquet')
+
+    with open(os.path.join(path, 'meta.json')) as meta_file:
+        meta = json.load(meta_file)
```
VibhuJawa, line 262: This does not download datasets. Can we add instructions for this dataset creation as this is for external consumption.

VibhuJawa (changes requested):

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: oh, that's part of another PR that isn't merged yet

alexbarghi-nv: it should not be part of thisone

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: Sure, I can do that

VibhuJawa: Thanks, I am blocked  for `cugraph-dgl` benchmarks as i need `meta.json` and `parquet` which are not available .

VibhuJawa (changes requested):

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+
+    if reverse_edges:
+        dask_edgelist_df = dask_edgelist_df.rename(columns={'src':'dst', 'dst':'src'})
+
+    return dask_edgelist_df, dask_label_df, node_offsets, edge_offsets, total_num_nodes
+
+
+def load_disk_dataset(dataset, dataset_dir='.', reverse_edges=True, replication_factor=1):
+    path = os.path.join(dataset_dir, dataset)
+    parquet_path = os.path.join(path, 'parquet')
```
rlratzel, line 259: If you're interested in cleaning up all the `os.path.join(...)` calls, you could make these `pathlib.Path` objects then use the slash operator
```suggestion
    from pathlib import Path
    path = Path(dataset_dir) / dataset
    parquet_path = path / 'parquet'
```

rlratzel (commented):
LGTM, I like how the dataframe is used at the end to collect results and output a CSV.  I had a couple minor general python comments.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+    sampler = BulkSampler(
+        batch_size=batch_size,
+        output_path=output_path,
+        graph=G,
+        fanout_vals=fanout,
+        with_replacement=False,
+        random_state=seed,
+        seeds_per_call=seeds_per_call,
+        batches_per_partition=batches_per_partition,
+        log_level = 'INFO'
```
VibhuJawa, line 167:
```suggestion
       This fails, i think this should be part of a different PR. 
```

VibhuJawa (changes requested):

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+
+    dask_edgelist_df = generate_edgelist_rmat(
+        scale=scale, edgefactor=edgefactor, seed=seed, unweighted=True, mg=True,
+    )
+    dask_edgelist_df = dask_edgelist_df.astype("int64")
+    dask_edgelist_df = dask_edgelist_df.reset_index(drop=True)
+
+
+    dask_edgelist_df = renumber_ddf(dask_edgelist_df).persist()
+    dask_edgelist_df = symmetrize_ddf(dask_edgelist_df).persist()
```
VibhuJawa, line 230: Persist causes a big memory overhead, please avoid it. 
```suggestion
```

VibhuJawa, line 230: Persist causes a big memory overhead, please avoid it. 
```suggestion
    dask_edgelist_df = renumber_ddf(dask_edgelist_df)
    dask_edgelist_df = symmetrize_ddf(dask_edgelist_df)
```

VibhuJawa (commented):
I cant seem to be able to run this on 8 GPUs, can you comment on if this works on 8 GPUs or not.

alexbarghi-nv:
> I cant seem to be able to run this on 8 GPUs, can you comment on if this works on 8 GPUs or not.

I've been reliably running this script on 8 GPUs on 2x papers100M without issues.  I'll try taking out the persist calls, but the ETL would take 3x as long without those persists.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: Persistence is now disabled unless enabled by choice.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,647 @@
...
+from cugraph.testing.mg_utils import (
+    generate_edgelist_rmat,
+    # get_allocation_counts_dask_persist,
+    get_allocation_counts_dask_lazy,
+    sizeof_fmt,
+    get_peak_output_ratio_across_workers,
+    restart_client,
+)
+
+from cugraph.testing.mg_utils import (
```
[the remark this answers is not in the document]
alexbarghi-nv: yes, yes it should.  fixed.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: Readme added.  Will add a script that automatically does the conversion from ogb in another PR.

VibhuJawa:
> I  've been reliably running this script on 8 GPUs on 2x papers100M without issues. I'll try taking out the persist calls, but the ETL would take 3x as long without those persists.

We triaged this to lack of rmm async , switching on rmm_async fixes this

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,685 @@
...
+    """
+    assert dask_dataframe['src'].dtype == 'int64'
+    assert dask_dataframe['dst'].dtype == 'int64'
+    assert dask_dataframe['etp'].dtype == 'int32'
+
+    G = cugraph.MultiGraph(directed=True)
+    G.from_dask_cudf_edgelist(
+        dask_dataframe, source="src", destination="dst", edge_type='etp', renumber=False
+    )
+    return G
```
VibhuJawa, line 78: So `cugraph_dgl` requires edge ids for `homogenous` multi graphs and `heterogenous`  multi graphs .   I am curious on how cugraph_pyg is handing this.

VibhuJawa (approved):

VibhuJawa (changes requested):
Still need changes around eid . Sorry approved by mistake.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: edge ids aren't required

alexbarghi-nv: But I can add them if DGL needs them

VibhuJawa: How does PYG differentiate b/w 2 edges between the same src and destination but with different eids ? Or does it just not handle MultiGraphs ?

alexbarghi-nv: It doesn't need to differentiate unless there are edge properties, all it cares about is the vertex ids.

VibhuJawa: I guess we will have to make adding edge types and edge ids optional, i think if you make adding edge type optional , I can just chime in a follow up PR that addes edge ids  and not block this PR.

alexbarghi-nv: Same edge can appear multiple times, that is legal

VibhuJawa: But how do you know which edge id it came from ?  Like if i have `edge id-1`  and `edge id-2` with different features , how do you know which edge id those samples came from ?

alexbarghi-nv: Well, edge type is not optional because that's used to determine vertex type.  It's just the edge ids that are optional.  And also, PyG's GraphStore interface technically doesn't support edge ids or edge properties right now (they're only supported in the native `Data` and `HeteroData` classes).

alexbarghi-nv: The answer is we don't care on the PyG end which edge it came from

alexbarghi-nv: papers100M, MAG, etc. don't have edge properties.

alexbarghi-nv: Ok, I'm going to add the option to output edge type.  Give me a few minutes...

VibhuJawa: > the remote graph (GraphStore) interface

I am not sure where or why we will use ` remote graph `, i always thought that it was a cugraph service thing that we deprecated for GNN use cases . 

Anyways, I guess the idea now is to just update the script by making adding edge types optional. I can do a follow up to add edge ids.

alexbarghi-nv: By "remote graph", PyG is referring to any backend that is not native, basically.

alexbarghi-nv: It doesn't literally mean a remote graph in this case.

alexbarghi-nv: https://pytorch-geometric.readthedocs.io/en/latest/modules/data.html?highlight=remote#remote-backend-interfaces

alexbarghi-nv: ok, @VibhuJawa, I added adding edge properties as an option that is off by default.  I tested it and it seems to be working correctly.  Please verify that it works with your DGL workflow.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,736 @@
...
+            for batch_size in batch_sizes:
+                for seeds_per_call in seeds_per_call_opts:
+                    print(f'dataset: {dataset}')
+                    print(f'batch size: {batch_size}')
+                    print(f'fanout: {fanout}')
+                    print(f'seeds_per_call: {seeds_per_call}')
+
+                    if re.match(r'([A-z]|[0-9])+\[[0-9]+\]', dataset):
+                        replication_factor = int(dataset[-2])
+                        dataset = dataset[:-3]
```
VibhuJawa, line 675: We seem to trim `replication` factor here for the second iteration which means that the numbers are not correct from the second point onwards

benchmarks/cugraph/standalone/README.bulk_sampling.md
```diff
@@ -0,0 +1,116 @@
+# cuGraph Bulk Sampling
```
VibhuJawa, line 1: Recommend moving this to a different folder like `bulk_sampling/` for readme to be rendered.

VibhuJawa (changes requested):

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,736 @@
...
+            'start': cudf.Series(dtype='int64')
+        })
+    
+    if replication_factor > 1:
+        for r in range(1, replication_factor):
+            df_replicated = df
+            for col, offset in col_offsets.items():
+                df_replicated[col] += offset * r
+        
+            df = cudf.concat([df, df_replicated], ignore_index=True)
```
VibhuJawa, line 157: For a factor of 4 we get 8x edges is this  expected ? I would have imagined  we would only have 4x edges. 
 
 ```python3
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({"src":[0],"dst":[1]})

In [3]: replication_factor = 4

In [4]: col_offsets = {"src":0, "dst":0}

In [5]: for r in range(1, replication_factor):
   ...:     df_replicated = df
   ...:     for col, offset in col_offsets.items():
   ...:         df_replicated[col] += offset * r
   ...:     df = pd.concat([df, df_replicated], ignore_index=True)
   ...: 

In [6]: print(df)
   src  dst
0    0    1
1    0    1
2    0    1
3    0    1
4    0    1
5    0    1
6    0    1
7    0    1

 ```
```suggestion
        df_replicated = df
        for r in range(1, replication_factor):
            for col, offset in col_offsets.items():
                df_replicated[col] += offset * r
            df = cudf.concat([df, df_replicated], ignore_index=True)
```

VibhuJawa (changes requested):

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
VibhuJawa: Requested some more changes based on more testing

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,736 @@
...
+
+    with open(os.path.join(path, 'meta.json')) as meta_file:
+        meta = json.load(meta_file)
+    
+    node_offsets, node_offsets_replicated, total_num_nodes = \
+        assign_offsets_pyg(meta['num_nodes'], replication_factor=replication_factor)
+
+    edge_index_dict = {}
+    for edge_type in os.listdir(parquet_path):
+        if re.match(r'[a-z]+__[a-z]+__[a-z]+', edge_type):
```
VibhuJawa, line 287: Minor nitpick: I would suggest using meta.json and ['num_edges'] for this rather than relying on regex matching.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,736 @@
...
+                "num_input_edges",
+                "directed",
+                "renumber",
+                "input_memory_per_worker",
+                "peak_allocation_across_workers",
+                "input_to_peak_ratio",
+                "output_to_peak_ratio",
+            ],
+        )
+        stats_df.to_csv("cugraph_graph_creation_stats.csv")
```
VibhuJawa, line 732:
```suggestion
        stats_df.to_csv("cugraph_sampling_stats.csv")
```

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,736 @@
...
+        dask_edgelist_df
+    )
+    print('constructed graph')
+
+    input_memory = G.edgelist.edgelist_df.memory_usage().sum().compute()
+    print(f'input memory: {input_memory}')
+
+    now = datetime.now()
+    timestring = datetime.strftime(now, '%Y-%m-%d-%H:%M:%S')
+    output_subdir = os.path.join(output_path, f'{dataset}[{replication_factor}]_b{batch_size}_f{fanout}_{timestring}')
```
VibhuJawa, line 466: Having a timing string attached here is making scripting non straightforward for downstream consumption of this in bulk sampling script .  Would it be possible for us to remove this ? 

```suggestion
    output_subdir = os.path.join(output_path, f'{dataset}[{replication_factor}]_b{batch_size}_f{fanout})
```

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: Should be fixed now

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: done

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: fixed

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,734 @@
...
+
+    meta = cudf.DataFrame({
+        'node': cudf.Series(dtype='int64'),
+        'batch': cudf.Series(dtype='int32')
+    })
+
+    
+    batch_df = label_df.map_partitions(_make_batch_ids, batch_size, n_workers, meta=meta)
+    if persist:
+        batch_df = batch_df.persist()
```
VibhuJawa, line 188: This should always be persisted

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,734 @@
...
+            if persist:
+                edge_index_dict[can_edge_type] = edge_index_dict[can_edge_type].persist()
+        
+        gc.collect()
+
+        if reverse_edges:
+            edge_index_dict[can_edge_type] = edge_index_dict[can_edge_type].rename(columns={'src':'dst','dst':'src'})
+            
+            if persist:
+                edge_index_dict[can_edge_type] = edge_index_dict[can_edge_type].persist()
```
VibhuJawa, line 320: This should be outside the if  `reverse_edges` block ?

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
```diff
@@ -0,0 +1,734 @@
...
+    num_input_edges = len(dask_edgelist_df)
+    print(
+        f"Number of input edges = {num_input_edges:,}"
+    )
+
+    G = construct_graph(
+        dask_edgelist_df
+    )
+    print('constructed graph')
+
```
VibhuJawa, line 460: Suggest deleting `dask_edgelist_df `

del dask_edgelist_df

VibhuJawa (changes requested):
Requested some more minor changes but i think we are all most there.

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
VibhuJawa: This does  not look to be fixed , Can you please check

VibhuJawa: Ahh, Sorry. I see that you now loop on `datasets`

VibhuJawa (approved):
LGTM

benchmarks/cugraph/standalone/cugraph_bulk_sampling.py
alexbarghi-nv: Yeah, you were looking at an outdated version.  No worries.

rlratzel (approved):

PR #3641, opened 2023-06-06
Disables memory usage assertions

rlratzel:
This PR disables the assertions intended to check if memory is being allocated and free'd as expected.  [Recent failures](https://github.com/rapidsai/cugraph/actions/runs/5188532065/jobs/9352311316) occurring seemingly infrequently have caused suspicion as to whether these are accurately ensuring if memory is free'd or not. Plus, if the tests change to use memory pools, or pytest-xdist which result in GPU sharing, these assertions will also fail.

Instead, this PR adds a variable which can be easily set to enable or disable the assertions.  This will make it convenient for testing for memory leaks locally when the developer knows the environment is appropriate for these assertions.  We will likely add memory leak checking back in on a case-by-case basis, possibly using different tools, in a future PR.

### DIFF
--- a/python/pylibcugraph/pylibcugraph/tests/test_uniform_neighbor_sample.py
+++ b/python/pylibcugraph/pylibcugraph/tests/test_uniform_neighbor_sample.py
@@ -25,6 +25,11 @@ from pylibcugraph import (
 )
 from pylibcugraph import uniform_neighbor_sample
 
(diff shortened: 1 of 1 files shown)


BradReesWork (approved):

alexbarghi-nv (approved):
👍

PR #3654, opened 2023-06-10
Fix broadcast root during the replication call

jnke2016:
A Raft [PR](https://github.com/rapidsai/raft/pull/1573) assigning deterministic ranks to dask workers was merged, breaking batch algorithms like `batch_edge_betweenness_centrality` by picking the wrong worker as the root for the `broadcast` operation.
This PR ensures that the worker with `rank = 0` is the root of the broadcast operation


jnke2016:
Pushed the  PR to the wrong branch. This is targeted for `23.06` as a hotfix

PR #3669, opened 2023-06-23
[FIX] Fix the hang in cuGraph Python Uniform Neighbor Sample, Add Logging to Bulk Sampler

alexbarghi-nv:
Some dask operations were not being done correctly, and time was being lost in broadcasting the rank and label arrays to all workers.  This PR resolves those issues.

Also pulls in the previously-experimental changes that add logging to the bulk sampler.

Credit to @VibhuJawa for isolating and fixing the issues with the column merge in `uniform_neighbor_sample` and the new sampling notebook and shell script.

This PR does modify the sampling APIs so it is breaking.  The API changes are necessary to avoid unnecessary shuffling, and eventually, to improve batch id assignment.

**Dataset:** `ogbn_papers100M x 2`; **Fanout:** `[25, 25]`; **Batch Size:** `512`; **Seeds Per Call:** `524288`
Current runtime: 2.69 s ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
Previous runtime: 4.51 s ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
Speedup: 1.7x

**Dataset:** `ogbn_papers100M x 4`; **Fanout:** `[25, 25]`; **Batch Size:** `512`; **Seeds Per Call:** `524288`
Current runtime: 6.32 s ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
Previous runtime: 10.7 s ± 0 ns per loop (mean ± std. dev. of 1 run, 10 loops each)
Speedup: 1.7x

PR #3671, opened 2023-06-27
[WIP]Python sampling fix hang

VibhuJawa:
Do not merge, builds on top of https://github.com/rapidsai/cugraph/pull/3669


python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
```diff
@@ -418,8 +418,9 @@ def uniform_neighbor_sample(
                 input_graph.renumber_map.renumbered_src_col_name
             ].dtype,
         )
-    elif with_edge_properties and batch_id_list is None:
-        batch_id_list = cudf.Series(cp.zeros(len(start_list), dtype="int32"))
+    elif with_edge_properties and with_batch_ids is False:
+        start_list = start_list.to_frame()
+        start_list[batch_id_n] = cudf.Series(cp.zeros(len(start_list), dtype="int32"))
```
alexbarghi-nv, line 423: I'll update this with a function that generates the list of zeros locally instead of copying it.

PR #3680, opened 2023-06-29
Promote `datasets` from experimental to stable

### LINKED ISSUE #3675: [ENH] Move `datasets` API from experimental to stable
- Promote `datasets` API by copying `cugraph.experimental.datasets` to `cugraph.datasets`
- Add deprecation warning to `cugraph.experimental.datasets`

Follow-up: update docs, notebooks, and tests to use the stable version of the API.

nv-rliu:
This PR moves the `datasets` API from experimental to stable. Users can now do:

```python
from cugraph.datasets import karate

G = karate.get_graph()
```

The `promoted_experimental_warning_wrapper()` has been added to the existing `experimental.datasets` package.

closes #3675

### DIFF
--- a/python/cugraph/cugraph/datasets/__init__.py
+++ b/python/cugraph/cugraph/datasets/__init__.py
@@ -1,0 +1,71 @@
+# Copyright (c) 2022-2023, NVIDIA CORPORATION.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from pathlib import Path
+
+# datasets module
+from cugraph.datasets.dataset import (
+    Dataset,
+    load_all,
+    set_download_dir,
+    get_download_dir,
+    default_download_dir,
+)
+from cugraph.datasets import metadata
+
+# metadata path for .yaml files
+meta_path = Path(__file__).parent / "metadata"
+
+# invidual datasets
+karate = Dataset(meta_path / "karate.yaml")
+karate_data = Dataset(meta_path / "karate_data.yaml")
+karate_undirected = Dataset(meta_path / "karate_undirected.yaml")
+karate_asymmetric = Dataset(meta_path / "karate_asymmetric.yaml")
+karate_disjoint = Dataset(meta_path / "karate-disjoint.yaml")
+dolphins = Dataset(meta_path / "dolphins.yaml")
+polbooks = Dataset(meta_path / "polbooks.yaml")
+netscience = Dataset(meta_path / "netscience.yaml")
+cyber = Dataset(meta_path / "cyber.yaml")
+small_line = Dataset(meta_path / "small_line.yaml")
+small_tree = Dataset(meta_path / "small_tree.yaml")
+toy_graph = Dataset(meta_path / "toy_graph.yaml")
+toy_graph_undirected = Dataset(meta_path / "toy_graph_undirected.yaml")
+email_Eu_core = Dataset(meta_path / "email-Eu-core.yaml")
+ktruss_polbooks = Dataset(meta_path / "ktruss_polbooks.yaml")
+
+# batches
+DATASETS_UNDIRECTED = [karate, dolphins]
+
+DATASETS_UNDIRECTED_WEIGHTS = [netscience]
+
+DATASETS_UNRENUMBERED = [karate_disjoint]
+
+DATASETS = [dolphins, netscience, karate_disjoint]
+
+DATASETS_SMALL = [karate, dolphins, polbooks]
+
+STRONGDATASETS = [dolphins, netscience, email_Eu_core]
+
+DATASETS_KTRUSS = [(polbooks, ktruss_polbooks)]
+
+MEDIUM_DATASETS = [polbooks]
+
+SMALL_DATASETS = [karate, dolphins, netscience]
+
+RLY_SMALL_DATASETS = [small_line, small_tree]
+
+ALL_DATASETS = [karate, dolphins, netscience, polbooks, small_line, small_tree]
+
+ALL_DATASETS_WGT = [karate, dolphins, netscience, polbooks, small_line, small_tree]
+
+TEST_GROUP = [dolphins, netscience]
--- a/python/cugraph/cugraph/datasets/metadata/karate.yaml
+++ b/python/cugraph/cugraph/datasets/metadata/karate.yaml
@@ -1,0 +1,24 @@
+name: karate
+file_type: .csv
+author: Zachary W.
+url: https://raw.githubusercontent.com/rapidsai/cugraph/branch-22.08/datasets/karate.csv
+refs:
+  W. W. Zachary, An information flow model for conflict and fission in small groups,
+  Journal of Anthropological Research 33, 452-473 (1977).
+delim: " "
+header: None
+col_names:
+  - src
+  - dst
+  - wgt
+col_types:
+  - int32
+  - int32
+  - float32
+has_loop: true
+is_directed: true
+is_multigraph: false
+is_symmetric: true
+number_of_edges: 156
+number_of_nodes: 34
+number_of_lines: 156
--- a/python/cugraph/cugraph/__init__.py
+++ b/python/cugraph/cugraph/__init__.py
@@ -120,4 +120,6 @@ from cugraph import gnn
 
 from cugraph import exceptions
 
+from cugraph import datasets
+
 __version__ = "23.08.00"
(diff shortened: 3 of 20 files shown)


rlratzel (changes requested):
Thanks for working on this!  Here's a partial review, I'll add more feedback later.

Some other general feedback is to add unit tests.  It would be nice if the meta-data was tested/verified too (eg. the resulting graphs have the same number of vertices and edges as reported in the yaml files) as well as a test to ensure the new warning exceptions are being raised.

python/cugraph/cugraph/datasets/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+
+SMALL_DATASETS = [karate, dolphins, netscience]
+
+RLY_SMALL_DATASETS = [small_line, small_tree]
+
+ALL_DATASETS = [karate, dolphins, netscience, polbooks, small_line, small_tree]
+
+ALL_DATASETS_WGT = [karate, dolphins, netscience, polbooks, small_line, small_tree]
+
+TEST_GROUP = [dolphins, netscience]
```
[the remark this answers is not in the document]
nv-rliu: You're right. Most of these were were created because many of the pre-existing unit tests defined their own _sets_ of files. Now that we're moving `datasets` out of experimental, it would be better to eliminate these from the `__init__` file and define groups elsewhere (perhaps in the `test` directory)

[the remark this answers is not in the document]
nv-rliu: These have been moved to `cugraph/testing/__init__.py`. While the batches are mainly used for testing purposes, they were not limited to being a test fixture as of yet. Users can still import dataset batches in Python if they wish to do so.

nv-rliu:
Update on unit test coverage suggestions by @rlratzel 

- `test_dataset.py::test_node_and_edge_count` creates a Graph using `cugraph.datasets` and checks that the number of nodes and edges in the resulting object is the same as the metadata files.
- `test_dataset.py::test_is_directed` creates a Graph and checks if was correctly constructed with directed edges by comparing with the metadata file.

python/cugraph/cugraph/testing/__init__.py
```diff
@@ -11,6 +11,52 @@
...
+    email_Eu_core,
+    toy_graph,
+    toy_graph_undirected,
+]
+ALL_DATASETS_WGT = [karate, dolphins, netscience, polbooks, small_line, small_tree]
+TEST_GROUP = [dolphins, netscience]
+
+# FIXME: removed karate variant. check if unit tests are breaking
+DATASETS_UNRENUMBERED = []
+DATASETS = [dolphins, netscience, karate]
```
rlratzel, line 62: This is a lot better, but we should still clean these up if possible: eg. `polbook` is in both the "small" and "medium" lists, we use underscores most of the time but `STRONGDATASETS` does not, `RLY` is an abbreviation but we don't abbreviate anything else, `DATASETS_KTRUSS` should be defined in the ktruss tests if possible instead of here (plus it's using a tuple and no others are), etc.  I'm also still thinking we may not want the ALL_CAPS naming convention here, unless anyone else thinks it's helpful in some way.

rlratzel (changes requested):
Looking better, thanks!  We should also figure out how to refactor the tests that still need the various karate derivatives, or do something else so we can stop importing from experimental.

python/cugraph/cugraph/testing/__init__.py
nv-rliu: I've taken a closer look at the unit tests and figuring out just how many of these batches actually need to be supported. It looks like a handful of them can be removed, since they were leftover namespaces from previous iterations of our tests. The batches that are not being used more than once can also be declared within the scope of the test file and be removed from `testing/__init__.py`

It looks like the following can be safely removed:
- `TEST_GROUP`: no longer used
- `DATASETS_UNRENUMBERED`: can be replaced with `karate_disjoint`
- `DATASETS_UNDIRECTED_WEIGHTS`: can be replaced with `netscience`
- `STRONGDATASETS`: no longer used
- `RLY_SMALL_DATASETS`: no longer used
- `MEDIUM_DATASETS`: can be replaced with `polbooks`
- `DATASETS_KTRUSS`: can be declared inside `test_k_truss.py`

nv-rliu:
Merged branch that updated notebooks, docstrings, and unit tests into this PR because there was a lot of overlap in work and the changes are non-breaking.

python/cugraph/cugraph/testing/__init__.py
```diff
@@ -54,9 +55,4 @@
     toy_graph,
     toy_graph_undirected,
 ]
-ALL_DATASETS_WGT = [karate, dolphins, netscience, polbooks, small_line, small_tree]
-TEST_GROUP = [dolphins, netscience]
-
-# FIXME: removed karate variant. check if unit tests are breaking
-DATASETS_UNRENUMBERED = []
-DATASETS = [dolphins, netscience, karate]
+DATASETS_TESTING = [dolphins, netscience, karate_disjoint]
```
rlratzel, line 58: Hmmm, this name seems a little awkward IMO.  Since we're treating these as the default set of datasets to use for tests, how about `DEFAULT_DATASETS`?  Then to be consistent (if it's not too much trouble) maybe change the naming convention of the others to be `SMALL_DATASETS`, `WEIGHTED_DATASETS`, etc.?

python/cugraph/cugraph/tests/components/test_connectivity.py
```diff
@@ -28,14 +29,17 @@
...
 
 # Temporarily suppress warnings till networkX fixes deprecation warnings
 # (Using or importing the ABCs from 'collections' instead of from
 # 'collections.abc' is deprecated, and in 3.8 it will stop working) for
 # python 3.7.  Also, this import networkx needs to be relocated in the
 # third-party group once this gets fixed.
-import warnings
 
 with warnings.catch_warnings():
     warnings.filterwarnings("ignore", category=DeprecationWarning)
```
rlratzel, line 45: I think we should get rid of this now.

rlratzel (changes requested):
Looks good, thanks!  I just have a couple small requests.

python/cugraph/cugraph/testing/__init__.py
nv-rliu: Sounds good. I created a table that includes all the datasets that are associated with each unit test, so that should be a simple fix.

python/cugraph/cugraph/tests/components/test_connectivity.py
nv-rliu: Issue has been raised to address this concern. Since this isn't within the scope of the promotion of the `datasets` API, it will be resolved in a separate PR.

notebooks/algorithms/centrality/Eigenvector.ipynb
```diff
@@ -169,25 +178,7 @@
    ]
   },
   {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "### Read in the data - GPU\n",
-    "cuGraph depends on cuDF for data loading and the initial Dataframe creation\n",
-    "\n",
-    "The data file contains an edge list, which represents the connection of a vertex to another.  The `source` to `destination` pairs is in what is known as Coordinate Format (COO).  In this test case, the data is just two columns.  However a third, `weight`, column is also possible"
```
acostadon, line 178: not a big deal at all, but explaining COO form is useful even though using datasets makes it more transparent

notebooks/algorithms/structure/Renumber.ipynb
acostadon, line 1: No header change?

python/cugraph/cugraph/datasets/metadata/karate.yaml
```diff
@@ -0,0 +1,24 @@
+name: karate
+file_type: .csv
+author: Zachary W.
+url: https://raw.githubusercontent.com/rapidsai/cugraph/branch-22.08/datasets/karate.csv
```
acostadon, line 4: Should the url be pointing at the 22.08 version of the data? What happens if the data is updated?

notebooks/algorithms/centrality/Eigenvector.ipynb
nv-rliu: I'll go back and include a sentence or two about COO form even though the notebook was converted to use Datasets

notebooks/algorithms/structure/Renumber.ipynb
nv-rliu: In this file I only updated the Copyright statement to display the correct year. The contents of the notebook were left untouched.

python/cugraph/cugraph/datasets/metadata/karate.yaml
nv-rliu: You're right. This should point to the new 23.08 version of the data. I'm not sure if the data has changed at all but it is a misleading URL. Perhaps we should revisit hosting a folder of reliable datasets in our S3 bucket. I'll raise and issue for this and get this updated to the current branch for now.

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -239,3 +240,58 @@ def test_unload():
     assert ds._edgelist is not None
     ds.unload()
     assert ds._edgelist is None
+
+
```
acostadon, line 244: Nice to see the metadata testing.

python/cugraph/cugraph/datasets/metadata/karate.yaml
nv-rliu: [Here](https://github.com/rapidsai/ops/issues/2737) is the ops issue that will upload dataset files to s3

nv-rliu:
The changes in this PR branch have been merged into #3712

PR #3684, opened 2023-07-03
Update conv layers in `cugraph-dgl` and `-pyg` based on `cugraph-ops` refactoring

tingyu66:
This PR address the breaking changes in `pylibcugraphops` pytorch APIs. It should fix the [pytest error](https://github.com/rapidsai/cugraph/actions/runs/5425211493/jobs/9890754410?pr=3683) in the CI for `cugraph-dgl` and `-pyg` in https://github.com/rapidsai/cugraph/pull/3683.

@MatthiasKohl We need both your PR and this one to pass CI. Please feel free to consolidate them.

PR #3683, opened 2023-06-30
[FIX] renaming of cugraph-ops symbols (refactoring)

MatthiasKohl:
This PR will fix naming of symbols from cugraph-ops once https://github.com/rapidsai/cugraph-ops/pull/516 is merged.

CI may not run through right away, but opening this PR to merge as soon as the cugraph-ops PR is in nightly.

PR #3688, opened 2023-07-05
[FIX] Rename `cugraph-ops` symbols (refactoring) and update GHA workflows to call pytest via `python -m pytest`

naimnv:
This PR:
- renames `cugraph-ops` symbols and updates tests in `cugraph-dgl` and `-pyg` based on cugraph-ops refactoring
- updates GHA workflows to call pytest via `python -m pytest`. This is to fix the `pytest not found error` in [log](https://github.com/rapidsai/cugraph/actions/runs/5420960384/jobs/9855784044#step:9:260).

### DIFF
--- a/cpp/src/sampling/neighborhood.cu
+++ b/cpp/src/sampling/neighborhood.cu
@@ -22,6 +22,8 @@
 
 #include <raft/random/rng_state.hpp>
 
+#include <type_traits>
+
 namespace cugraph {
 
 template <typename vertex_t, typename edge_t>
@@ -34,14 +36,19 @@ sample_neighbors_adjacency_list(raft::handle_t const& handle,
                                 size_t sampling_size,
                                 ops::graph::SamplingAlgoT sampling_algo)
 {
-  const auto [ops_graph, max_degree] = detail::get_graph_and_max_degree(graph_view);
-  return ops::graph::uniform_sample_csr(rng_state,
+  using base_vertex_t = std::decay_t<vertex_t>;
+  using base_edge_t   = std::decay_t<edge_t>;
+  static_assert(std::is_same_v<base_vertex_t, base_edge_t>,
+                "cugraph-ops sampling not yet implemented for different node and edge types");
+
+  const auto ops_graph = detail::get_graph(graph_view);
+  return ops::graph::uniform_sample_csc(rng_state,
                                         ops_graph,
                                         ptr_d_start,
                                         num_start_vertices,
                                         sampling_size,
                                         sampling_algo,
-                                        max_degree,
+                                        ops_graph.dst_max_in_degree,
                                         handle.get_stream());
 }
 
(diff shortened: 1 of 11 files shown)


tingyu66 (commented):
@naimnv I merged in my fixes originally in https://github.com/rapidsai/cugraph/pull/3684 so that python tests will pass. 

This PR should supercede #3683 #3684 #3685 upon passing CI checks.

MatthiasKohl (approved):
LGTM, thanks for consolidating the PRs

ChuckHastings (approved):

cpp/src/sampling/neighborhood.cu
```diff
@@ -34,14 +36,19 @@ sample_neighbors_adjacency_list(raft::handle_t const& handle,
                                 size_t sampling_size,
                                 ops::graph::SamplingAlgoT sampling_algo)
 {
-  const auto [ops_graph, max_degree] = detail::get_graph_and_max_degree(graph_view);
-  return ops::graph::uniform_sample_csr(rng_state,
+  using base_vertex_t = std::decay_t<vertex_t>;
+  using base_edge_t   = std::decay_t<edge_t>;
```
seunghwak, line 40: If we require `decay` here, should we better apply decay at the caller site? `vertex_t` implies it is a non-const non-volatile value type in our code base, and I think we should better maintain that convention.

cpp/src/sampling/neighborhood.cu
```diff
@@ -34,14 +36,19 @@ sample_neighbors_adjacency_list(raft::handle_t const& handle,
...
+  static_assert(std::is_same_v<base_vertex_t, base_edge_t>,
+                "cugraph-ops sampling not yet implemented for different node and edge types");
+
+  const auto ops_graph = detail::get_graph(graph_view);
+  return ops::graph::uniform_sample_csc(rng_state,
```
seunghwak, line 45: It was `csr` in the old code, now it is `csr`, is this intentional?

cpp/src/sampling/neighborhood.cu
```diff
@@ -55,14 +62,19 @@ std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>> sample_
   size_t sampling_size,
   ops::graph::SamplingAlgoT sampling_algo)
 {
-  const auto [ops_graph, max_degree] = detail::get_graph_and_max_degree(graph_view);
+  using base_vertex_t = std::decay_t<vertex_t>;
+  using base_edge_t   = std::decay_t<edge_t>;
```
seunghwak, line 66: Same here, if decay is necessary, I think we should apply this in the caller site.

cpp/src/utilities/cugraph_ops_utils.hpp
```diff
@@ -20,34 +20,26 @@
...
 
 template <typename NodeTypeT, typename EdgeTypeT>
-ops::graph::fg_csr<EdgeTypeT> get_graph(
+ops::graph::csc<EdgeTypeT, NodeTypeT> get_graph(
   graph_view_t<NodeTypeT, EdgeTypeT, false, false> const& gview)
 {
-  ops::graph::fg_csr<EdgeTypeT> graph;
-  graph.n_nodes   = gview.number_of_vertices();
-  graph.n_indices = gview.number_of_edges();
+  ops::graph::csc<EdgeTypeT, NodeTypeT> graph;
```
seunghwak, line 30: Just to double check. `csr` to `csc` here is intentional, right?

cpp/src/utilities/cugraph_ops_utils.hpp
```diff
@@ -20,34 +20,26 @@
...
+  graph.n_src_nodes = gview.number_of_vertices();
+  graph.n_dst_nodes = gview.number_of_vertices();
+  graph.n_indices   = gview.number_of_edges();
+  // FIXME this is sufficient for now, but if there is a fast (cached) way
+  // of getting max degree, use that instead
```
seunghwak, line 35: Just for curiosity, is there any performance benefit in providing a tighter bound for max_in_degree?

MatthiasKohl: For neighborhood sampling, no. For other operations in cugraph-ops, yes.

cpp/src/sampling/neighborhood.cu
seunghwak: I see, yeah, then, I don't care whether we apply decay here or not (I initially thought the caller is also our code, but if the caller is a user, then, I am not sure removing decay here is good or bad).

seunghwak (approved):

PR #3697, opened 2023-07-06
Use rapids-cmake testing to run tests in parallel

robertmaynard:
By switching to rapids-cmake testing infrastructure we can both run all the tests in parallel from the build environment and the test/install env.

Parallel testing performance.
 - serial baseline: 820sec
 - serial + improved memory allocation: 798sec
 - 1 gpu j8: 220sec

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -14,6 +14,10 @@
 # limitations under the License.
 #
 #=============================================================================
+enable_testing()
+
+include(rapids-test)
+rapids_test_init()
 
 ###################################################################################################
 # - common test utils -----------------------------------------------------------------------------
@@ -28,7 +32,7 @@ add_library(cugraphtestutil STATIC
             centrality/betweenness_centrality_validate.cu
             community/egonet_validate.cu
             cores/k_core_validate.cu
-	    structure/induced_subgraph_validate.cu
+	        structure/induced_subgraph_validate.cu
             sampling/random_walks_check_sg.cu
             ../../thirdparty/mmio/mmio.c)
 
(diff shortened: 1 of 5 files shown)


cpp/tests/CMakeLists.txt
```diff
@@ -28,7 +32,7 @@ add_library(cugraphtestutil STATIC
             centrality/betweenness_centrality_validate.cu
             community/egonet_validate.cu
             cores/k_core_validate.cu
-	    structure/induced_subgraph_validate.cu
+	        structure/induced_subgraph_validate.cu
```
vyasr, line 35:
```suggestion
	    structure/induced_subgraph_validate.cu
```

cpp/tests/CMakeLists.txt
```diff
@@ -69,18 +88,18 @@ function(ConfigureTest CMAKE_TEST_NAME)
...
-        TARGETS ${CMAKE_TEST_NAME}
-        COMPONENT testing
-        DESTINATION bin/gtests/libcugraph
-        EXCLUDE_FROM_ALL)
+    rapids_test_add(
+        NAME ${CMAKE_TEST_NAME}
+        COMMAND ${CMAKE_TEST_NAME}
+        GPUS ${_CUGRAPH_TEST_GPUS}
+        PERCENT ${_CUGRAPH_TEST_PERCENT}
+        INSTALL_COMPONENT_SET testing #bin/gtests/libcugraph
```
vyasr, line 100:
```suggestion
        INSTALL_COMPONENT_SET testing
```

cpp/tests/CMakeLists.txt
```diff
@@ -96,29 +115,44 @@ function(ConfigureTestMG CMAKE_TEST_NAME)
...
              ${CMAKE_TEST_NAME}
-             ${MPIEXEC_POSTFLAGS})
-
-    set_target_properties(
-        ${CMAKE_TEST_NAME}
-            PROPERTIES INSTALL_RPATH "\$ORIGIN/../../../lib")
+             ${MPIEXEC_POSTFLAGS}
+        GPUS ${GPU_COUNT}
+        PERCENT 100
+        INSTALL_COMPONENT_SET testing_mg #bin/gtests/libcugraph_mg
```
vyasr, line 133:
```suggestion
        INSTALL_COMPONENT_SET testing_mg
```

cpp/tests/CMakeLists.txt
vyasr: Unless you intended to keep that to document where the file was going? That's not obvious to me, but if that's the intent feel free to leave it as is. That information is otherwise provided by the install_relocatable calls at the bottom.

cpp/tests/utilities/base_fixture.hpp
```diff
@@ -69,7 +69,12 @@ inline auto make_managed() { return std::make_shared<rmm::mr::managed_memory_res
 
 inline auto make_pool()
 {
-  return rmm::mr::make_owning_wrapper<rmm::mr::pool_memory_resource>(make_cuda());
+  auto const [free, total] = rmm::detail::available_device_memory();
+  auto min_alloc =
```
vyasr, line 73: Could you document why you want this amount? It's not obvious how this value was selected, or that it's potentially tied to the maximum parallelism used in our test suite.

Also:
```suggestion
  auto const min_alloc =
```

vyasr (approved):
A couple of suggestions for improvement, but generally looks good. I don't need to review again.

cpp/tests/CMakeLists.txt
robertmaynard: This was leftover as I was porting to `install_relocatable` will remove

cpp/tests/utilities/base_fixture.hpp
robertmaynard: Updated the `make_pool` to have a comment on why we have this magic value, and the impact of changing it.

BradReesWork (approved):

BradReesWork:
@robertmaynard Looks good but a minor conflict 
I'll ask OPS to review

raydouglass (approved):

PR #3699, opened 2023-07-06
Integrate C++ Sampling Source Behavior Updates

alexbarghi-nv:
Updates pylibcugraph and cuGraph-Python to support the new flags at the C++/C API layer (`unique_sources`, `carry_over_sources`, `deduplicate_sources`.  Adds appropriate tests.  Fixes a couple cuGraph-Python bugs that the new tests uncovered as well, and properly exposes `return_hops`.

Merge after #3696


cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -242,6 +309,57 @@ cugraph_error_code_t cugraph_uniform_neighbor_sample_with_edge_properties(
...
+ *                           needs to be transposed
+ * @param [in]  start_vertices Device array of start vertices for the sampling
+ * @param [in]  start_vertex_labels  Device array of start vertex labels for the sampling.  The
+ * labels associated with each start vertex will be included in the output associated with results
+ * that were derived from that start vertex.  We only support label of type INT32. If label is
+ * NULL, the return data will not be labeled.
+ * @param [in]  label_list Device array of the labels included in @p start_vertex_labels.  If
+ * @p label_to_comm_rank is not specified this parameter is ignored.  If specified, label_list
+ * must be sorted in ascending order.
+ * @param [in]  label_to_comm_rank Device array identifying which comm rank the output for a
```
jnke2016, line 333: Are `label_list` and `label_to_comm_rank` device arrays of same size()? And if yes can there be repeated ranks in `label_to_comm_rank`

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -242,6 +309,57 @@ cugraph_error_code_t cugraph_uniform_neighbor_sample_with_edge_properties(
...
+ * @param [in]  handle       Handle for accessing resources
+ * @param [in]  graph        Pointer to graph.  NOTE: Graph might be modified if the storage
+ *                           needs to be transposed
+ * @param [in]  start_vertices Device array of start vertices for the sampling
+ * @param [in]  start_vertex_labels  Device array of start vertex labels for the sampling.  The
+ * labels associated with each start vertex will be included in the output associated with results
+ * that were derived from that start vertex.  We only support label of type INT32. If label is
+ * NULL, the return data will not be labeled.
+ * @param [in]  label_list Device array of the labels included in @p start_vertex_labels.  If
+ * @p label_to_comm_rank is not specified this parameter is ignored.  If specified, label_list
```
jnke2016, line 331: Can you also add that `label_to_comm_rank` cannot be specified  unless `start_vertex_labels` is also specified like it is checked in the test?

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -187,8 +187,75 @@ typedef struct {
...
+cugraph_error_code_t cugraph_sampling_options_create(cugraph_sampling_options_t** options,
+                                                     cugraph_error_t** error);
+
+/**
+ * @brief   Set flag to sample with_replacement
+ *
+ * @param options - opaque pointer to the sampling options
+ * @param value - Boolean value to assign to the option
+ */
+void cugraph_sampling_set_with_replacement(cugraph_sampling_options_t* options, bool_t value);
```
jnke2016, line 215: This is a general question but couldn't we have a single function like below instead of having multiple function calls?
```
cugraph_sampling_set_options(
       cugraph_sampling_options_t* options,
       bool_t with_replacement,
       bool_t with_return_hops,
       bool_t with_unique_sources,
       bool_t with_carry_over_sources,
       bool_t with_dedupe_sources)
```

alexbarghi-nv: @ChuckHastings

cpp/include/cugraph_c/sampling_algorithms.h
alexbarghi-nv: They're the same size and labels should not be repeated.  But ranks can be repeated any number of times.

cpp/include/cugraph_c/sampling_algorithms.h
alexbarghi-nv: @ChuckHastings

cpp/include/cugraph_c/sampling_algorithms.h
jnke2016: Got it. That was my understanding. Thanks

cpp/tests/c_api/uniform_neighbor_sample_test.c
```diff
@@ -20,11 +20,351 @@
 #include <cugraph_c/graph.h>
```
jnke2016, line 20: I don't see any tests for updates to the tests in `mg_uniform_neighbor_sample_test.c`

python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample.py
```diff
@@ -528,6 +528,206 @@ def test_uniform_neighbor_sample_empty_start_list():
...
+
+
+@pytest.mark.sg
+def test_uniform_neighbor_sample_carry_over_sources_email_eu_core():
+    el = email_Eu_core.get_edgelist()
+
+    G = cugraph.Graph(directed=True)
+    G.from_cudf_edgelist(el, source="src", destination="dst")
+
+    seeds = el.src.sample(frac=0.001).unique().sample(frac=1)
```
jnke2016, line 674: I believe you can also use [select_random_vertices](https://github.com/rapidsai/cugraph/blob/branch-23.08/python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py#L674)

python/pylibcugraph/pylibcugraph/uniform_neighbor_sample.pyx
```diff
@@ -212,7 +249,18 @@ def uniform_neighbor_sample(ResourceHandle resource_handle,
...
     cdef cugraph_rng_state_t* rng_state_ptr = \
         cg_rng_state.rng_state_ptr
-    error_code = cugraph_uniform_neighbor_sample_with_edge_properties(
+
+    cdef cugraph_sampling_options_t* sampling_options
+    error_code = cugraph_sampling_options_create(&sampling_options, &error_ptr)
+    assert_success(error_code, error_ptr, "cugraph_sampling_options_create")
+
+    cugraph_sampling_set_with_replacement(sampling_options, with_replacement)
+    cugraph_sampling_set_return_hops(sampling_options, <bool_t>True)
```
jnke2016, line 258: Why is this one hardcoded to True and not part of the python and PLC API?

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: First comment... these changes are actually from PR #3696, so it might be a better question there.  But I'll address it here.

We could.  I considered that.  Interestingly, I ended up at this because some of the parameters don't make sense to set together.  In #3696 we are combining these into an enum, so a single call might make more sense.  I will consider that while refactoring that PR.

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Yes, I'll make that change in #3696

cpp/tests/c_api/uniform_neighbor_sample_test.c
ChuckHastings: Again, that's a shortcoming of PR #3696.  I'll take a look at this in that PR.

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Updated in #3696

cpp/tests/c_api/uniform_neighbor_sample_test.c
ChuckHastings: Added tests to PR #3696

python/pylibcugraph/pylibcugraph/uniform_neighbor_sample.pyx
alexbarghi-nv: fixed

python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample.py
alexbarghi-nv: fixed

rlratzel (approved):
LGTM, thanks for adding the tests.  
Re: the tests, I wonder if in the future we could reduce so much duplication (eg. figure out how to parametrize email_eu_core and the other dataset instead of separate tests).  Also, we should probably have some coverage at the PLC level, then have the cugraph tests verify the additions unique to cugraph (which is mainly the dask code but possibly more).

ChuckHastings (approved):

PR #3704, opened 2023-07-12
Stop using setup.py in build.sh

vyasr:
This PR changes replaces direct invocations of setup.py in build.sh with `pip install` commands since setup.py usage is deprecated. This change will facilitate upcoming changes to our build systems.


build.sh
```diff
@@ -60,7 +60,7 @@ HELP="$0 [<target> ...] [<flag> ...]
  and <flag> is:
    -v                         - verbose build mode
    -g                         - build for debug
-   -n                         - do not install after a successful build
+   -n                         - do not install after a successful build (does not affect Python packages)
```
rlratzel, line 63: I think there was an internal discussion and we determined that we didn't know of anyone using this flag. Since this is now doing less, I wonder if it should just be EOL'd instead - ie. allow it to be understood by the script, print a message that it's no longer valid, exit 1.  Or we could skip the message and just let it error as unrecognized if we think no one in the community is using it either.

rlratzel (approved):

Issue #2416, opened 2022-07-15
Store Nx results for cuGraph using a `Resultset` object containing expected test results


oorliu:
Currently, a large number of cuGraph unit tests will construct NetworkX objects when comparing algorithm outputs. For example:
```
assert nx_n_components == cg_n_components
assert nx_vertices == cg_vertices
```

This issue proposes that the new method of comparison could be done by storing a set of "golden results" in a `Resultset` object (similar in design to a [`Dataset`](https://github.com/rapidsai/cugraph/issues/1348) object). This is feasible since the tests are run using a finite set of static datasets.

By doing so, it would reduce cuGraph's dependency on NetworkX, and also eliminate the amount of testing overhead needed when repeatedly creating graph objects. 

Because much of the functionality overlaps with [`Dataset`](https://github.com/rapidsai/cugraph/issues/1348), work on the `Resultset` class should include factoring out common code into a module shared by both classes.

PR #3705, opened 2023-07-12
Testing util improvements and refactoring

betochimas:
closes https://github.com/rapidsai/cugraph/issues/2416

Removes a dependency when running python tests, which will speed up test runs. Also introduces a new testing component (`Resultset`) to clean up unit tests by generating golden results and storing them locally/on the cloud. This PR is focused on tests from the `traversal` category.

NOTE: Devs will download specific golden results from the cloud to avoid local overhead.


python/cugraph/cugraph/tests/traversal/test_paths.py
```diff
@@ -37,13 +38,29 @@
...
+    "shortest_path_length_-1_1": ValueError,
+    "shortest_path_length_1_10": ValueError,
+    "shortest_path_length_0_42": ValueError,
+    "shortest_path_length_1_8": 3.4028235e38,
+}
+
+
+@pytest.fixture(scope="module")
+def load_traversal_results():
+    return load_resultset("traversal", None)
```
rlratzel, line 55: same comment as above, I don't think this needs to return anything, same for all others like this.

python/cugraph/cugraph/tests/traversal/test_bfs.py
```diff
@@ -313,22 +322,27 @@ def get_cu_graph_nx_results_and_params(seed, depth_limit, G, dataset, directed,
 )
 
 
+@pytest.fixture(scope="module")
+def load_traversal_results():
+    return load_resultset("traversal", None)
```
rlratzel, line 327: What is this returning?  I don't think this fixture needs to return anything.

python/cugraph/cugraph/tests/traversal/test_bfs.py
```diff
@@ -337,21 +351,28 @@ def single_small_dataset_nx_graph(request):
 # results, the starting vertex for BFS, and flag if shortes path counting was
 # used.
 @pytest.fixture(scope="module", params=algo_test_fixture_params)
-def dataset_nxresults_startvertex_spc(dataset_nx_graph, request):
-    return get_cu_graph_nx_results_and_params(*request.param, *dataset_nx_graph)
+def dataset_goldenresults_startvertex_spc(
+    dataset_golden_results, load_traversal_results, request
+):
+    return get_cu_graph_golden_results_and_params(
+        *request.param, *dataset_golden_results, load_traversal_results
```
rlratzel, line 358: I don't think you need to pass `load_traversal_results`.

python/cugraph/cugraph/tests/traversal/test_bfs.py
```diff
@@ -245,34 +249,38 @@ def _compare_bfs(cugraph_df, nx_distances, source):
...
-        G,
-        dataset_path,
-        directed,
-        utils.generate_nx_graph_from_file(dataset_path, directed),
-    )
 
-
-def get_cu_graph_nx_results_and_params(seed, depth_limit, G, dataset, directed, Gnx):
+def get_cu_graph_golden_results_and_params(
+    seed, depth_limit, G, dataset_path, dataset_name, directed, load_results
```
rlratzel, line 264: is `load_results` unused here?

python/cugraph/cugraph/tests/traversal/test_paths.py
```diff
@@ -37,13 +38,29 @@
...
+paths_test_data = {
+    "shortest_path_length_1_1": 0,
+    "shortest_path_length_1_5": 2.0,
+    "shortest_path_length_1_3": 2.0,
+    "shortest_path_length_1_6": 2.0,
+    "shortest_path_length_-1_1": ValueError,
+    "shortest_path_length_1_10": ValueError,
+    "shortest_path_length_0_42": ValueError,
+    "shortest_path_length_1_8": 3.4028235e38,
+}
```
rlratzel, line 50: Is this a placeholder for calls to `get_resultset()`, or are you thinking this is permanent due to the resutls being just single values?  If it's a placeholder, now is probably the time to replace it with `get_resultset()`, but either way a comment would be helpful.

python/cugraph/cugraph/tests/traversal/test_sssp.py
```diff
@@ -258,22 +265,62 @@ def test_sssp_invalid_start(gpubenchmark, dataset_source_nxresults, cugraph_inpu
...
-    gpubenchmark, single_dataset_source_nxresults, cugraph_input_type
+@pytest.mark.parametrize("cugraph_input_type", utils.MATRIX_INPUT_TYPES)
+def test_sssp_nonnative_inputs_matrix(
+    gpubenchmark, single_dataset_source_goldenresults, cugraph_input_type
 ):
-    test_sssp(gpubenchmark, single_dataset_source_nxresults, cugraph_input_type)
+    test_sssp(gpubenchmark, single_dataset_source_goldenresults, cugraph_input_type)
+
+
+# MARK
```
rlratzel, line 275: This comment can be removed.

rlratzel (changes requested):
Starting to look good for a first pass.  We'll eventually refactor quite a lot of the fixtures out on future PRs, but just replacing Nx calls initially is a good start.

We need to get the resultset data uploaded and the rest of the download/unpack code written to fully test end-to-end and get this passing CI.

python/cugraph/cugraph/testing/generate_resultsets.py
```diff
@@ -0,0 +1,284 @@
...
+
+from tempfile import NamedTemporaryFile
+import random
+
+import numpy as np
+import networkx as nx
+
+import cudf
+import cugraph
+from cugraph.experimental.datasets import (
```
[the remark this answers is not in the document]
betochimas: Fixed.

python/cugraph/cugraph/testing/resultset.py
```diff
@@ -0,0 +1,119 @@
...
+            for val in arg_dict_pair
+        ]
+    )
+    uuid = _resultsets.get(resultset_key)
+    if uuid is None:
+        raise KeyError(f"results for {arg_dict} not found")
+
+    # results_filename = RAPIDS_RESULTSET_ROOT_DIR_PATH / (uuid + ".csv")
+    results_filename = (
+        RAPIDS_DATASET_ROOT_DIR_PATH / "tests" / "resultsets" / (uuid + ".csv")
```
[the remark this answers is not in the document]
betochimas: Yes, in both cases they are `RAPIDS_RESULTSET_ROOT_DIR_PATH / (uuid + ".csv")` for each results file

python/cugraph/cugraph/tests/traversal/test_bfs.py
betochimas: Changed to only call `load_resultset`, rather than returning it, which itself returns nothing.

python/cugraph/cugraph/tests/traversal/test_paths.py
betochimas: It's not a placeholder, but it's been renamed to `paths_golden_results`. I added a comment to explain why it's not a placeholder/added as a Resultset

python/cugraph/cugraph/testing/__init__.py
```diff
@@ -11,7 +11,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from cugraph.testing.utils import RAPIDS_DATASET_ROOT_DIR_PATH, RAPIDS_DATASET_ROOT_DIR
+from cugraph.testing.utils import (
+    RAPIDS_DATASET_ROOT_DIR_PATH,
+    RAPIDS_DATASET_ROOT_DIR,
+    RAPIDS_RESULTSET_ROOT_DIR_PATH,
```
rlratzel, line 17: I think it might be better to start off by just using `RAPIDS_DATASET_ROOT_DIR`, then adding another env var just for Resultset only if we realize it's needed.

betochimas: Sure, I removed `RAPIDS_RESULTSET_ROOT_DIR_PATH`.

python/cugraph/cugraph/testing/resultset.py
```diff
@@ -43,6 +43,7 @@ def get_cudf_dataframe(self):
 
 
 _resultsets = {}
+_results_dir_path = RAPIDS_DATASET_ROOT_DIR_PATH / "tests" / "resultsets"
```
rlratzel, line 46: I noticed this constant is also defined in `generate_resultsets.py`.  Can we define this path in just one place and use it in both modules?  I think having it defined here in `resultset.py' might be better.

Also, I noticed this imports `RAPIDS_DATASET_ROOT_DIR` from `cugraph.testing`, and `generate_resultsets.py` imports it from `cugraph.testing.utils`, so I think we should be using the same import for that too.

rlratzel (changes requested):
Looks good, just one small request.

rlratzel (approved):
Thanks!

PR #3707, opened 2023-07-13
Renumber utility function for sampling output

### LINKED ISSUE #3718: C++ Create MFG Capability for sampling output
PyG and DGL have requirements on the numbering of vertices in the sampling output.  Add a capability to renumber the sampling output to satisfy these requirements.

seunghwak:
Implements a utility function to renumber sampling function (e.g. `uniform_neighbor_sample`) outputs based on DGL and PyG requirements.

closes #3718


seunghwak:
Define the function API first. Several questions to be resolved.

1. Does multi-GPU implementation make sense? If we are processing only one batch, sampled output will be small and it will easily fit within a single GPU. If we are processing many batches, different batches will be independently processed in different GPUs. I don't see a practical multi-GPU use case. Any thoughts?
2. Say `edgelist_srcs` is [0, 5, 5] and `edgelist_hops` is [1, 0, 2]. Then, should we associate hop # 0 to vertex 5 or hop # 2 to vertex 5? (so [0, 5] should be renumbered to [1, 0] or [0,1]?). If vertex 5 is associated with hop # 0 & 2, this API assumed that vertex 5 is associated with the minimum hop # (i.e. 0). Any concerns with this?

Besides the questions above, do you have any other concerns?

@ChuckHastings @tingyu66 @VibhuJawa @alexbarghi-nv

alexbarghi-nv:
> Define the function API first. Several questions to be resolved.
> 
> 1. Does multi-GPU implementation make sense? If we are processing only one batch, sampled output will be small and it will easily fit within a single GPU. If we are processing many batches, different batches will be independently processed in different GPUs. I don't see a practical multi-GPU use case. Any thoughts?
> 2. Say `edgelist_srcs` is [0, 5, 5] and `edgelist_hops` is [1, 0, 2]. Then, should we associate hop # 0 to vertex 5 or hop # 2 to vertex 5? (so [0, 5] should be renumbered to [1, 0] or [0,1]?). If vertex 5 is associated with hop # 0 & 2, this API assumed that vertex 5 is associated with the minimum hop # (i.e. 0). Any concerns with this?
> 
> Besides the questions above, do you have any other concerns?
> 
> @ChuckHastings @tingyu66 @VibhuJawa @alexbarghi-nv

For (1): We are already partitioning, sending all samples for a single batch to a single GPU.  So single GPU implementation is ok.  There should be no comms/transfer between GPUs.  Each worker should just run this function independent of all other workers.

For (2): I think we always want the minimum hop.  So this behavior is fine.

alexbarghi-nv:
@seunghwak this API looks good to me 👍

tingyu66:
> Define the function API first. Several questions to be resolved.
> 
> 1. Does multi-GPU implementation make sense? If we are processing only one batch, sampled output will be small and it will easily fit within a single GPU. If we are processing many batches, different batches will be independently processed in different GPUs. I don't see a practical multi-GPU use case. Any thoughts?
> 2. Say `edgelist_srcs` is [0, 5, 5] and `edgelist_hops` is [1, 0, 2]. Then, should we associate hop # 0 to vertex 5 or hop # 2 to vertex 5? (so [0, 5] should be renumbered to [1, 0] or [0,1]?). If vertex 5 is associated with hop # 0 & 2, this API assumed that vertex 5 is associated with the minimum hop # (i.e. 0). Any concerns with this?
> 
> Besides the questions above, do you have any other concerns?
> 
> @ChuckHastings @tingyu66 @VibhuJawa @alexbarghi-nv

2). Yes, a src vertex should be associated with its minimum hop

VibhuJawa:
> 2\. Say `edgelist_srcs` is [0, 5, 5] and `edgelist_hops` is [1, 0, 2]. Then, should we associate hop # 0 to vertex 5 or hop # 2 to vertex 5? (so [0, 5] should be renumbered to [1, 0] or [0,1]?). If vertex 5 is associated with hop # 0 & 2, this API assumed that vertex 5 is associated with the minimum hop # (i.e. 0). Any concerns with this?

Yes, Agreed it should be associated with the minimum hop. 

API looks good to me.

seunghwak:
OK, I will drop the `multi_gpu` flag and implement this.

alexbarghi-nv (changes requested):
This looks good, just needs a C API.  I assume @ChuckHastings is working on that?

cpp/src/sampling/renumber_sampled_edgelist_sg.cu
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Just 2023

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
```diff
@@ -0,0 +1,723 @@
...
+                     raft::device_span<vertex_t const> edgelist_dsts,
+                     std::optional<raft::device_span<size_t const>> label_offsets)
+{
+  auto approx_edges_to_sort_per_iteration =
+    static_cast<size_t>(handle.get_device_properties().multiProcessorCount) *
+    (1 << 20) /* tuning parameter */;  // for segmented sort
+
+  std::optional<rmm::device_uvector<label_index_t>> edgelist_label_indices{std::nullopt};
+  if (label_offsets) {
+    edgelist_label_indices =
```
ChuckHastings, line 62: Is this the same logical function as https://github.com/rapidsai/cugraph/blob/cab836896aeba2c5549c57ebfd4818e883c17b5c/cpp/src/structure/detail/structure_utils.cuh#L87?

If so, perhaps this should call that function.  If this implementation is better then perhaps we should swap the implementation to the better one.

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
```diff
@@ -0,0 +1,723 @@
...
+renumber_sampled_edgelist(
+  raft::handle_t const& handle,
+  rmm::device_uvector<vertex_t>&& edgelist_srcs,
+  std::optional<raft::device_span<int32_t const>> edgelist_hops,
+  rmm::device_uvector<vertex_t>&& edgelist_dsts,
+  std::optional<std::tuple<raft::device_span<label_t const>, raft::device_span<size_t const>>>
+    label_offsets,
+  bool do_expensive_check)
+{
+  using label_index_t = uint32_t;
```
alexbarghi-nv, line 471:
```suggestion
  using label_index_t = int32_t;
```
Shouldn't this be int32_t?

alexbarghi-nv (changes requested):

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
seunghwak: This can either be uint32_t or int32_t. Note that `label_t` and `label_index_t` are two different things.

Say, if we can have labels

`10, 25, 30, 7`

and label_indices

`0, 1, 2, 3`.

uint32_t here is basically 4 byte version of size_t (as we may not have more than 2^32-1 labels per GPU in the foreseeable future).

alexbarghi-nv: Ok, sorry for the confusion with `label_t`.  This makes sense.

alexbarghi-nv: I am getting an error though:
```
RuntimeError: non-success value returned from cugraph_uniform_neighbor_sample: CUGRAPH_UNKNOWN_ERROR cuGraph failure at file=/root/cugraph/cpp/src/sampling/renumber_sampled_edgelist_impl.cuh line=475: Invalid input arguments: current implementation assumes that the number of unique labels is no larger than std::numeric_limits<uint32_t>::max().
```

alexbarghi-nv: Oh, I know what's going on.  We need to use a flag to turn renumbering on/off.

alexbarghi-nv: @seunghwak can you add a flag to this function, and then either Chuck or I can properly set it from the value provided to the C API.

alexbarghi-nv: And we should also enforce that renumbering can only be done if the label list is provided, which I think is the direct cause of the error here.

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
seunghwak: What about moving expand_sparse_offsets (https://github.com/rapidsai/cugraph/blob/cab836896aeba2c5549c57ebfd4818e883c17b5c/cpp/src/structure/detail/structure_utils.cuh#L87) to misc_utils.cuh (https://github.com/rapidsai/cugraph/blob/branch-23.08/cpp/include/cugraph/utilities/misc_utils.cuh).

We are using this function multiple places outside `src/structure` (including few files under the `test` directory).

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
seunghwak: > Oh, I know what's going on. We need to use a flag to turn renumbering on/off.

This function is a renumbering function, so if the flag is turned off, this function should not be called, am I missing something?

alexbarghi-nv: Right now, the uniform neighbor sampling functor is always attempting to renumber.  We want to specify whether or not the functor should do that.

seunghwak: So shouldn't this be handled above this PR? `renumber_sampled_edgelist` doesn't control whether to renumber or not. I assume this should be handled in the sampling function (sampling function should either call this function or not based on the flag).

alexbarghi-nv: That's right, I was asking if you'd update the sampling function as well

alexbarghi-nv: But I can just do that, it's only an extra boolean and an additional check

alexbarghi-nv: The issue is that the way it's written in this PR, the sampling python tests will fail

seunghwak: Yeah... this PR does not touch any code related to sampling. Better be addressed in the sampling PR calling this function.

alexbarghi-nv: Ok, I'll take care of it

cpp/src/sampling/renumber_sampled_edgelist_impl.cuh
ChuckHastings: That sounds like a great idea.  It's very reasonable to be part of the public API.

seunghwak: OK, I addressed your review comments.

ChuckHastings (approved):

alexbarghi-nv (approved):
Tested, working.

PR #3710, opened 2023-07-15
[BUG] Unsupported graph for similiarity algos

jnke2016:
This PR update the docstrings raises an error when running any similarity algos with vertices from a graph that are unrenumbered.


BradReesWork (approved):

jnke2016:
Apart from the known `WCC`, `Louvain` and `Leiden` failure, I can't reproduce any of the `similarity` algos python tests failure locally. Some of them being OOM. I am still looking into it.

jnke2016:
Now we are left with the C++ tests known failures

PR #3711, opened 2023-07-17
[BUG]   Fix namesapce to default_hash and hash_functions

naimnv:
Fix namesapce to default_hash and  hash_functions

### DIFF
--- a/cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
+++ b/cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
@@ -27,6 +27,7 @@
 
 #include <cudf/detail/nvtx/ranges.hpp>
 #include <cudf/detail/utilities/device_atomics.cuh>
+#include <cudf/hashing/detail/default_hash.cuh>
 #include <cudf/hashing/detail/hash_functions.cuh>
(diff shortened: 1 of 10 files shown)


cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
```diff
@@ -118,7 +119,7 @@ union pair_packer<pair_type, std::enable_if_t<is_packable<pair_type>()>> {
  */
 template <typename Key,
           typename Element,
-          typename Hasher    = cudf::detail::default_hash<Key>,
+          typename Hasher    = cudf::hashing::detail::default_hash<Key>,
```
seunghwak, line 122: Should we use cudf's hash function? In other places, we are using cuco's hash function (and you may check whether cudf's hash function is actually an alias of cuco's hash function).

seunghwak: Oh, this isn't cugraph main code, this is libcugraph_etl, yeah... just disregard the comment.

seunghwak (approved):

ChuckHastings (approved):

PR #3716, opened 2023-07-18
Update to Cython 3.0.0

vyasr:
This PR contains the minimal set of changes to compile using Cython 3 without warnings. Future PRs can be made to take advantage of new or improved features.

The main change is that a the graph_primtypes module runs into https://github.com/cython/cython/issues/5554. I was able to work around that issue by removing all the fused type logic and instead just duplicate one function. There were multiple functions present that were not used at all corresponding to fused type specializations that were never instantiated, so the changes still ended up being a net reduction in code.

### DIFF
--- a/python/cugraph/cugraph/structure/graph_primtypes.pxd
+++ b/python/cugraph/cugraph/structure/graph_primtypes.pxd
@@ -154,22 +154,9 @@ ctypedef GraphCOOView[int,int,double] GraphCOOViewDouble
 ctypedef GraphCSRView[int,int,float] GraphCSRViewFloat
 ctypedef GraphCSRView[int,int,double] GraphCSRViewDouble
 
-ctypedef fused GraphCOOViewType:
-    GraphCOOViewFloat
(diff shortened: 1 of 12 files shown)


raydouglass (approved):

python/cugraph/cugraph/structure/graph_primtypes.pxd
rlratzel, line 1: Looks like all the changes to this file are related to removing the various GraphView fused types. That's fine by me, but I'm curious if that's just for code cleanup reasons or if Cython 3 requires it.  I looked through release notes and current docs but nothing jumped out.

rlratzel (approved):
LGTM. Thank you very much for doing this.
I have one question below, mostly just curiosity.

python/cugraph/cugraph/structure/graph_primtypes.pxd
rlratzel: Ah, sorry!  I just re-read the PR description, which answers my question.  Please disregard :)

PR #3693, opened 2023-07-05
Make `pylibcugraphops` optional imports in `cugraph-dgl` and `-pyg`

tingyu66:
Fixes #3691 .

PR #3456, opened 2023-04-07
cugraph: Build CUDA 12 packages

vyasr:
First attempt at building libcugraph with CUDA 12 and nvidia channel packages

Closes https://github.com/rapidsai/cugraph/issues/3271

PR #3721, opened 2023-07-18
Reorder dependencies.yaml channels

raydouglass:
This PR reorders the channels to be compatible with other RAPIDS libs.

#3693 added two new channels, but they were ordered differently than [other libraries](https://github.com/rapidsai/cudf/blob/branch-23.08/dependencies.yaml#L163-L169). This caused [an error](https://github.com/rapidsai/docker/actions/runs/5591817313/jobs/10223438072?pr=545#step:10:349) in the `conda-merge` script in https://github.com/rapidsai/docker/pull/545.

It's also worth noting that docker overhaul requiring the channel ordering being consistent across libraries' `dependencies.yaml` seems very fragile and will undoubtedly cause more issues down the line. Curious if anyone has thoughts on making this more reliable.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -181,10 +181,10 @@ channels:
   - rapidsai
   - rapidsai-nightly
   - dask/label/dev
-  - conda-forge
-  - nvidia
   - pytorch
   - dglteam/label/cu118
+  - conda-forge
+  - nvidia
 dependencies:
   checks:
     common:
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -4,10 +4,10 @@ channels:
 - rapidsai
 - rapidsai-nightly
 - dask/label/dev
-- conda-forge
-- nvidia
 - pytorch
 - dglteam/label/cu118
+- conda-forge
+- nvidia
 dependencies:
 - aiohttp
 - c-compiler
--- a/python/cugraph-dgl/conda/cugraph_dgl_dev_cuda-118.yaml
+++ b/python/cugraph-dgl/conda/cugraph_dgl_dev_cuda-118.yaml
@@ -4,10 +4,10 @@ channels:
 - rapidsai
 - rapidsai-nightly
 - dask/label/dev
-- conda-forge
-- nvidia
 - pytorch
 - dglteam/label/cu118
+- conda-forge
+- nvidia
 dependencies:
 - cugraph==23.8.*
 - dgl>=1.1.0.cu*


ajschmidt8 (approved):

dependencies.yaml
```diff
@@ -181,10 +181,10 @@ channels:
   - rapidsai
   - rapidsai-nightly
   - dask/label/dev
-  - conda-forge
-  - nvidia
   - pytorch
   - dglteam/label/cu118
```
bdice, line 185: It concerns me that we have a channel pointing to CUDA 11.8. This will need to be considered in #3456. cc: @rlratzel

bdice (approved):
This change is probably valid, though it is hard to know for sure because CI was skipped.

dependencies.yaml
tingyu66: Just found out this PR as #3693 was mentioned. 

`cugraph-dgl` package requires `dgl` from `dglteam/label/cu118` channel for testing purposes. That channel/label only consists of the `dgl` package, whose dependencies are surprisingly simple:
```
$ conda search dgl=1.1.1 -c dglteam/label/cu118 --info

Loading channels: done
dgl 1.1.1.cu118 py310_0
-----------------------
file name   : dgl-1.1.1.cu118-py310_0.tar.bz2
name        : dgl
version     : 1.1.1.cu118
build       : py310_0
build number: 0
size        : 71.1 MB
license     : Apache
subdir      : linux-64
url         : https://conda.anaconda.org/dglteam/label/cu118/linux-64/dgl-1.1.1.cu118-py310_0.tar.bz2
md5         : 9f4c8315366dc3091e5e92aade891388
timestamp   : 2023-06-26 05:45:51 UTC
dependencies: 
  - networkx
  - numpy
  - psutil
  - python >=3.10,<3.11.0a0
  - requests
  - scipy
  - tqdm
```

Hopefully it won't collide with cuda packages from `conda-forge` channel.

tingyu66:
> This PR reorders the channels to be compatible with other RAPIDS libs.
> 
> #3693 added two new channels, but they were ordered differently than [other libraries](https://github.com/rapidsai/cudf/blob/branch-23.08/dependencies.yaml#L163-L169). This caused [an error](https://github.com/rapidsai/docker/actions/runs/5591817313/jobs/10223438072?pr=545#step:10:349) in the `conda-merge` script in [rapidsai/docker#545](https://github.com/rapidsai/docker/pull/545).
> 
> It's also worth noting that docker overhaul requiring the channel ordering being consistent across libraries' `dependencies.yaml` seems very fragile and will undoubtedly cause more issues down the line. Curious if anyone has thoughts on making this more reliable.

Thanks for the fix. Probably need to rerun `rapids-dependency-file-generator` locally to pass CI.

PR #3723, opened 2023-07-19
Fix primitive bug discovered in MG edge betweenness centrality testing

ChuckHastings:
@jnke2016 found a bug in MG edge betweenness centrality.  Upon investigation, the `major_offset` in this block of code wasn't being set correctly if `major` is in the hypersparse region.

### DIFF
--- a/cpp/src/prims/transform_e.cuh
+++ b/cpp/src/prims/transform_e.cuh
@@ -295,7 +295,7 @@ void transform_e(raft::handle_t const& handle,
         auto major_offset            = edge_partition.major_offset_from_major_nocheck(major);
         vertex_t major_idx{major_offset};
 
-        if ((major_hypersparse_first) && (*major_hypersparse_first < major)) {
+        if ((major_hypersparse_first) && (major >= *major_hypersparse_first)) {
           auto major_hypersparse_idx =
             edge_partition.major_hypersparse_idx_from_major_nocheck(major);
           assert(major_hypersparse_idx);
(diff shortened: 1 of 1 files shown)


cpp/src/prims/transform_e.cuh
```diff
@@ -291,23 +291,16 @@ void transform_e(raft::handle_t const& handle,
...
-          major_idx    = major_offset;
+        auto major_offset            = edge_partition.major_offset_from_major_nocheck(major);
+        vertex_t major_idx{major_offset};
+
+        if ((major_hypersparse_first) && (*major_hypersparse_first < major)) {
+          auto major_hypersparse_idx =
+            edge_partition.major_hypersparse_idx_from_major_nocheck(major);
+          assert(major_hypersparse_idx);
+          major_idx = edge_partition.major_offset_from_major_nocheck(*major_hypersparse_first) +
+                      *major_hypersparse_idx;
```
seunghwak, line 303: What's the difference between these two?

First, `if ((major_hypersparse_first) && (*major_hypersparse_first < major))` should be `if ((major_hypersparse_first) && (*major_hypersparse_first <= major))`

And if I am not mistaken, if we're using DCSR/DCSC and major is in the hypersparse range, we are setting major_idx to major_hypersparse_idx + major_idx of the beginning of the hypersparse range.

I am not sure how these two are different.

seunghwak: Oh, I see, this isn't about major_idx, but this is about major_offset, let me take a closer look.

seunghwak: ` if ((major_hypersparse_first) && (*major_hypersparse_first < major))`
=>
` if ((major_hypersparse_first) && (major >= *major_hypersparse_first))`

Otherwise, LGTM.

ChuckHastings: Fixed

seunghwak (approved):

jnke2016 (approved):
Reviewed and tested. Looks good to me.

PR #3731, opened 2023-07-21
Switch to new wheel building pipeline

vyasr:
Moves the wheel build and test logic out of the workflow into the repo. This matches conda tests more closely and allows each repo to manage its own wheels more easily.

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+package_name=$1
+package_dir=$2
+
+source rapids-configure-sccache
+source rapids-date-string
+
+# Use gha-tools rapids-pip-wheel-version to generate wheel version then
+# update the necessary files
+version_override="$(rapids-pip-wheel-version ${RAPIDS_DATE_STRING})"
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+bash ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
+echo "The package name and/or version was modified in the package source. The git diff is:"
+git diff
+
+cd "${package_dir}"
+
+python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
+
+mkdir -p final_dist
+python -m auditwheel repair -w final_dist dist/*
+
+RAPIDS_PY_WHEEL_NAME="${package_name}_${RAPIDS_PY_CUDA_SUFFIX}" rapids-upload-wheels-to-s3 final_dist
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eoxu pipefail
+
+package_name=$1
+package_dir=$2
+
+mkdir -p ./dist
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+# echo to expand wildcard before adding `[extra]` requires for pip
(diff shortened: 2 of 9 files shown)


raydouglass (approved):

ci/test_wheel_pylibcugraph.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eoxu pipefail
```
ajschmidt8, line 4:
```suggestion
set -euo pipefail
```

Do we need the `-x` here? Or can we keep it consistent with the other scripts?

ci/test_wheel_cugraph.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eoxu pipefail
```
ajschmidt8, line 4:
```suggestion
set -euo pipefail
```

Do we need the `-x` here? Or can we keep it consistent with the other scripts?

ci/test_wheel.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# Copyright (c) 2023, NVIDIA CORPORATION.
+
+set -eoxu pipefail
```
ajschmidt8, line 4:
```suggestion
set -euo pipefail
```

Do we need the `-x` here? Or can we keep it consistent with the other scripts?

ci/build_wheel.sh
```diff
@@ -0,0 +1,29 @@
...
+source rapids-configure-sccache
+source rapids-date-string
+
+# Use gha-tools rapids-pip-wheel-version to generate wheel version then
+# update the necessary files
+version_override="$(rapids-pip-wheel-version ${RAPIDS_DATE_STRING})"
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+bash ci/release/apply_wheel_modifications.sh ${version_override} "-${RAPIDS_PY_CUDA_SUFFIX}"
```
ajschmidt8, line 18: Can we remove the `bash` keyword here?

This will require that the called script has a proper shebang (`#!/bin/bash`) and is executable (`chmod +x <file>`).

ci/test_wheel_cugraph.sh
```diff
@@ -0,0 +1,22 @@
...
+python -m pip install --no-deps ./local-pylibcugraph-dep/pylibcugraph*.whl
+
+# Always install latest dask for testing
+python -m pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.08
+
+# Only download test data for x86
+arch=$(uname -m)
+if [[ "${arch}" == "x86_64" ]]; then
+    pushd ./datasets
+    bash ./get_test_data.sh
```
ajschmidt8, line 18: Can we remove the `bash` keyword here?

This will require that the called script has a proper shebang (`#!/bin/bash`) and is executable (`chmod +x <file>`).

ci/test_wheel_cugraph.sh
```diff
@@ -0,0 +1,22 @@
...
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+RAPIDS_PY_WHEEL_NAME="pylibcugraph_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./local-pylibcugraph-dep
+python -m pip install --no-deps ./local-pylibcugraph-dep/pylibcugraph*.whl
+
+# Always install latest dask for testing
+python -m pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main git+https://github.com/rapidsai/dask-cuda.git@branch-23.08
+
+# Only download test data for x86
+arch=$(uname -m)
+if [[ "${arch}" == "x86_64" ]]; then
```
ajschmidt8, line 16:
```suggestion
if [[ "$(arch)" == "aarch64" && ${RAPIDS_BUILD_TYPE} == "pull-request" ]]; then
```
`arch` and `uname -m` return the same value, so this can be shortened.

ajschmidt8 (commented):
Leaving a post-merge review

PR #3741, opened 2023-07-25
Compute shortest distances between given sets of origins and destinations for large diameter graphs

seunghwak:
This code computes shortest distances between given origins and destinations and produces an OD matrix.

This code runs many SSSPs in parallel to extract sufficient parallelism considering that an SSSP run on a large diameter graph has only limited amount of parallelism.

This code uses hash table (cuco::static_map) to reduce memory footprint (and this enable running many SSSPs by cutting memory requirement to store distances from O(V) to the number of vertices around the active frontier).

For small diameter graphs (where one can extract sufficient parallelism with only a single SSSP run), sequentially running multiple SSSPs may be more performant (as hash table access is more expensive than linear array access + additional overheads to run many SSSPs in parallel).


BradReesWork:
LGTM from cmake perspective

BradReesWork (approved):

cpp/src/prims/vertex_frontier.cuh
```diff
@@ -204,8 +203,6 @@ class key_bucket_t {
                                           merged_pair_first + merged_vertices.size())),
           handle_ptr_->get_stream());
         merged_tags.resize(merged_vertices.size(), handle_ptr_->get_stream());
-        merged_vertices.shrink_to_fit(handle_ptr_->get_stream());
-        merged_tags.shrink_to_fit(handle_ptr_->get_stream());
```
ChuckHastings, line 208: Is there a reason we're not shrinking these here?

ChuckHastings (commented):
Couple of questions.  Otherwise looks good.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1179,6 +1179,40 @@ void sssp(raft::handle_t const& handle,
...
 
+/*
+ * @brief Compute the shortest distances from the given origins to all the given destinations.
+ *
+ * @throws cugraph::logic_error on erroneous input arguments.
+ *
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam weight_t Type of edge weights. Needs to be a floating point type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
```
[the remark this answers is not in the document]
seunghwak: Yes.

cpp/src/prims/vertex_frontier.cuh
seunghwak: `shrink_to_fit` basically allocates another array; copies data; and destroy the original array.

The copying part is expensive. If the expected number of duplicates is low, this really does not reduce memory footprint much, increase the peak memory usage (as we temporarily maintain two arrays), and increases execution time. Here two merged arrays typically don't have that many duplicates. I guess paying this cost for few % or less cut in memory is not a good trade-off.

ChuckHastings (approved):

seunghwak:
@jnke2016 Take a look for this PR especially the code below.

https://github.com/rapidsai/cugraph/pull/3741/files#diff-1d11ed707cec17cd1996e73fe04a22243e3698d052deed45a73244331cbc5286R344

I think addressing this FIXME can be a good first task to work on primitives.

naimnv (approved):

PR #3747, opened 2023-07-27
Fixes options for `--pydevelop` to remove unneeded CWD path ("."), restores use of `setup.py` temporarily for develop builds

rlratzel:
This PR adds the following:
* Removes the now unnecessary "." path used in the install options when `--pydevelop` is used, which resulted in an error when called.
* Temporarily adds a call to `setup.py bdist_ext --inplace` only when `--pydevelop` is used to workaround a [bug in scikit-build](https://github.com/scikit-build/scikit-build/issues/981) when `setuptools>=64` is installed.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -175,7 +175,7 @@ if hasArg --cmake_default_generator; then
     CMAKE_GENERATOR_OPTION=""
 fi
 if hasArg --pydevelop; then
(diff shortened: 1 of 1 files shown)


BradReesWork (approved):

vyasr (approved):
Looks reasonable to me. I'm glad we're only changing this for the develop case, that way we'll get some testing of the non-develop case using `pip install` for a while before we switch to requiring `pip install` when setup.py goes away.

PR #3767, opened 2023-08-04
Calling `dataset.get_edgelist()` returns a copy of an edge list instead of global

### LINKED ISSUE #3421: Edgelist Returned from dataset.get_edgelist() is global
Currently, when calling `get_edgelist()`, a global edgelist is returned.  This causes issues when performing operations such as:
```
df_unweighted = datasets.karate.get_edgelist()
df.drop('wgt', axis=1, inplace=True)
...
df_weighted = datasets.karate.get_edgelist()
print(df_weighted)
```
Expected:
```
src  dst  wgt
...
```

Actual:
```
src  dst
...
```

To fix this issue, datasets should return a copy of the global edgelist so the edgelist can be used in multiple places without affecting the original object.

nv-rliu:
Closes #3421

This PR fixes a bug where `get_edgelist()` would return a edge list, which causes any changes made to the first returned edge list to be reflected in subsequent calls to `get_edgelist()`.

Now, `get_edgelist()` returns a copy of the dataframe stored by the `Dataset` object.


nv-rliu:
Opening another PR after #3774 is merged to avoid merge conflicts.

PR #3776, opened 2023-08-09
Fix metadata mismatch

jnke2016:
This PR fixes the metadata mismatch resulting from a modification of the input dataframe by making an initial copy.

### DIFF
--- a/python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
+++ b/python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
@@ -150,6 +150,9 @@ class simpleDistributedGraphImpl:
         legacy_renum_only=False,
     ):
 
+        # Make a copy of the input dask dataframe before doing any modification.
+        input_ddf = input_ddf.map_partitions(lambda df: df.copy())
+
         if not isinstance(input_ddf, dask_cudf.DataFrame):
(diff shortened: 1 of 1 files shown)


rlratzel:
Converted to Draft since @jnke2016 is still investigating the root cause.

rlratzel (approved):

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -1192,7 +1192,5 @@ def _get_column_from_ls_dfs(lst_df, col_name):
     if len_df == 0:
         return lst_df[0][col_name]
     output_col = cudf.concat([df[col_name] for df in lst_df], ignore_index=True)
-    for df in lst_df:
-        df.drop(columns=[col_name], inplace=True)
```
VibhuJawa, line 1196: This will introduce a memory overhead. I dont think this is the best fix.

VibhuJawa (changes requested):
I think we should try creating a copy

VibhuJawa (approved):
Looks good to me, still request filing an issue to cugraph so that we can raise it to cuDF once we triage it down

alexbarghi-nv (approved):
👍

rlratzel:
Followup issue [here](https://github.com/rapidsai/cugraph/issues/3790).

PR #3779, opened 2023-08-10
Clean-up old testing conventions in `test_ecg.py` 

### LINKED ISSUE #3235: `test_ecg_clustering` is too sensitive to different `RAPIDS_DATA_SET_ROOT_DIR` formats
Users should be able to set `RAPIDS_DATASET_ROOT_DIR` to an absolute path, or a path relative to their CWD, a path with sym links, etc.  However, `test_ecg_clustering` only allows `RAPIDS_DATASET_ROOT_DIR` to be set in a particular way, likely due to the code [here](https://github.com/rapidsai/cugraph/blob/branch-23.02/python/cugraph/cugraph/tests/test_ecg.py#L36).  

As an alternative to just fixing the code in `test_ecg.py` could be to migrate this test to use the `datasets` package.

nv-rliu:
Closes #3235 

This PR cleans up the `test_ecg.py` unit test. It removes the old way of comparing networkx results to cugraph, which used `PurePath` to compare the location of the stored dataset when getting networkx results to compare with cugraph.

Since the cugraph tests use the `Datasets` API, the golden results can be fetched based on the name of the dataset being used.


alexbarghi-nv (approved):
👍

PR #3780, opened 2023-08-11
MFG creation optimzation

VibhuJawa:


VibhuJawa:
This was covered in https://github.com/rapidsai/cugraph/pull/3810

PR #3782, opened 2023-08-11
Allow models to use a lightweight sparse structure

tingyu66:
This PR introduces `SparseGraph` class to allow `SAGEConv` to use a more lightweight graph structure. The goal is to provide an option to bypass `to_block` which is the bottleneck in the sampling workflow.

I will submit another PR to extend the pattern to other models.


VibhuJawa:
I think this works for us and we should priortize getting this done properly soon

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
```diff
@@ -48,3 +49,85 @@ def pad_offsets(self, offsets: torch.Tensor, size: int) -> torch.Tensor:
...
+        Source indices of the edges.
+
+    dst_ids: torch.Tensor, optional
+        Destination indices of the edges.
+
+    csrc_ids: torch.Tensor, optional
+        Compressed source indices. It is a monotonically increasing array of
+        size (num_src_nodes + 1,). For the k-th source node, its neighborhood
+        consists of the destinations between `dst_indices[csrc_indices[k]]` and
+        `dst_indices[csrc_indices[k+1]]`.
```
VibhuJawa, line 73: I have a question regarding when `num_dst_nodes` > `len(cdst_ids)-1` for this case. 

Lets look at below case: 

```python3
cdst_ids (Compressed Destinations Indices):0,2,5,7
```
```python3
src_indices: 1,2,2,3,4,4,5
```
I believe following will work (please correct me):

```python3
num_src_nodes = 6
num_dst_nodes = 3 
```

And i guess below will fail ((please correct me):

```python3
num_src_nodes = 6
num_dst_nodes = 5 # Modified it to a higher value to ensure alignment for output nodes that are missing
```

**Question:** 
So this will have to handled by ensuring correct creation because we want to handle alignment problem b/w blocks.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
```diff
@@ -48,3 +49,85 @@ def pad_offsets(self, offsets: torch.Tensor, size: int) -> torch.Tensor:
...
+        if self._cdst_ids is None:
+            if not dst_ids_is_sorted:
+                self._dst_ids, self._perm = torch.sort(self._dst_ids)
+                self._src_ids = self._src_ids[self._perm]
+            self._cdst_ids = torch._convert_indices_from_coo_to_csr(
+                self._dst_ids,
+                self.num_dst_nodes,
+                out_int32=self._dst_ids.dtype == torch.int32,
+            )
+
```
VibhuJawa, line 130: Do you think we should remove `dst_ids`  if we are forcing `CSC` conversion because that will mean memory overhead of maintaining it always ?

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
```diff
@@ -48,3 +49,85 @@ def pad_offsets(self, offsets: torch.Tensor, size: int) -> torch.Tensor:
...
+        `src_indices[cdst_indices[k+1]]`.
+
+    dst_ids_is_sorted: bool
+        Whether `dst_ids` has been sorted in an ascending order.
+
+    Notes
+    -----
+    COO-format requires `src_ids` and `dst_ids`.
+    CSC-format requires `cdst_ids` and `src_ids`.
+    CSR-format requires `csrc_ids` and `dst_ids`.
```
VibhuJawa, line 88: I think we should force **user** to provide the format requirement  to prevent confusion.  Like add a  format variable something like  `input_format` which will take values `coo`, `csc` and `csr`. 

Then we can raise errors according to the input what the user provided. 

Also, i dont like `input_format` variable name but you get the idea.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
VibhuJawa: I think if we are forcing `csc` conversions  for now but in future we may want to expand to other formats right, I think we might want to either have this configurable via   a class variable. 

We can probably borrow the convention from `formats`.  

We wont follow their default of `'coo' -> 'csr' -> 'csc'`, but have our own version. 

See [formats docs](https://docs.dgl.ai/en/1.1.x/generated/dgl.DGLGraph.formats.html) .

python/cugraph-dgl/cugraph_dgl/nn/conv/sageconv.py
```diff
@@ -98,50 +98,58 @@ def reset_parameters(self):
...
         max_in_degree : int
-            Maximum in-degree of destination nodes. It is only effective when
-            :attr:`g` is a :class:`DGLBlock`, i.e., bipartite graph. When
-            :attr:`g` is generated from a neighbor sampler, the value should be
-            set to the corresponding :attr:`fanout`. If not given,
-            :attr:`max_in_degree` will be calculated on-the-fly.
+            Maximum in-degree of destination nodes. When :attr:`g` is generated
+            from a neighbor sampler, the value should be set to the corresponding
+            :attr:`fanout`. This option is used to invoke the MFG-variant of
+            cugraph-ops kernel.
```
VibhuJawa, line 117: Why have stopped handing this ?  I am kind of worried about making this change . Can we discuss this more .

python/cugraph-dgl/cugraph_dgl/nn/conv/sageconv.py
```diff
@@ -98,50 +98,58 @@ def reset_parameters(self):
...
+            num_dst_nodes = g.num_dst_nodes
+        elif isinstance(g, dgl.DGLHeteroGraph):
+            offsets, indices, _ = g.adj_tensors("csc")
+            _graph = ops_torch.CSC(
+                offsets=offsets,
+                indices=indices,
+                num_src_nodes=g.num_src_nodes(),
+                dst_max_in_degree=max_in_degree,
+            )
+            num_dst_nodes = g.num_dst_nodes()
```
VibhuJawa, line 144: I think we should have functions like `num_dst_nodes` and `num_src_nodes` for our `SparseGraph` too, to prevent duplication

VibhuJawa (changes requested):
The overall implementation looks great. Thanks for prioritizing work on this.  

Have asked a couple of followups and minor code changes.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
```diff
@@ -11,14 +11,15 @@
...
+
 from cugraph.utilities.utils import import_optional
 
 torch = import_optional("torch")
-nn = import_optional("torch.nn")
 ops_torch = import_optional("pylibcugraphops.pytorch")
 
 
-class BaseConv(nn.Module):
+class BaseConv(torch.nn.Module):
```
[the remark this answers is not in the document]
tingyu66: This is not a user-facing class. It is only used to handle the case where we fall back to full graph variant. In addition, with the recent cugraph-ops refactoring disabling the MFG-variant, we might totally remove this class.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
tingyu66: It should be illegal when `num_dst_nodes` != `len(cdst_ids)-1`. I will improve the error handling in this case. For example, pyg does lots of assertions to check the size. We should throw proper exceptions.

```
cdst_ids (Compressed Destinations Indices):0,2,5,7
src_indices: 1,2,2,3,4,4,5
```
In your example with num_src_nodes = 6, num_dst_nodes = 3, this translates to a COO of
(1,2,2,3,4,4,5)
(0,0,1,1,1,2,2)

With num_src_nodes = 6, num_dst_nodes = 5, the constructor should have failed, unless `cdst_ids` is augmented (cdst_ids = 0,2,5,7,7,7).

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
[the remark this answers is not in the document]
VibhuJawa: Okay.  I think it is fine as is then.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
VibhuJawa: Thanks. Yup, this is what i was expecting. We will just make sure that the changes @seunghwak  from cugraph sampling ensures that all the MFGs line up.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
tingyu66: As we discussed via slack, we will provide `input_format` and `output_format` to help specify which tensor is needed.

python/cugraph-dgl/cugraph_dgl/nn/conv/sageconv.py
tingyu66: As we discussed, this is due to the recent cugraph-ops refactoring and has nothing to do with this PR.

tingyu66:
@VibhuJawa Please review it again. [Here](https://gist.github.com/tingyu66/c34a6f5b8775e7ce2234c8a7419878fb) are some example usages. The `formats` option denotes the layouts users want to create. I chose not to add an "`input_format`" because it is redundant. The kwargs already carries the information, and `__init__()` gets  unnecessarily complicated with it. The `reduce_memory` flag does what you suggested -- removing unused tensors from class.

python/cugraph-dgl/cugraph_dgl/nn/conv/base.py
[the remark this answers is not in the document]
VibhuJawa: Resolved

VibhuJawa (approved):
LGTM

alexbarghi-nv:
@tingyu66 should this still be in draft?  It seems to be ready for review.

tingyu66:
> @tingyu66 should this still be in draft? It seems to be ready for review.

Yes, it's ready. Forgot to click the button.

BradReesWork (approved):

PR #3789, opened 2023-08-14
Fix ValueError Caused By Batches With No Samples

### LINKED ISSUE #3785: [BUG] Bulk Sampling fails at batch_size<=64
Bulk Sampling fails at batch_size<=64. 
This is because at that batch size, we encountered batches where all input seeds had an out-degree of 0, and the maximum batch id exceeded what was expected.  This is because the C++ code will drop empty samples when it produces the offsets array (which is the correct behavior) but the Python code checks the length of that array based on the maximum batch id.

alexbarghi-nv:
There is currently a bug in the bulk sampler where empty samples (which result from batches where all input seeds have 0 out-degree) cause a `ValueError` to be raised since the batch ids do not match with what is expected based on the number of returned offset values.

This PR resolves this issue by checking whether the maximum batch id for a partition exceeds what is expected, and if so, renumbers the batch ids to be contiguous starting from 0.  This has the effect of throwing out the empty batches, which is acceptable behavior in both `cuGraph-DGL` and `cuGraph-PyG`.

Closes #3785


python/cugraph/cugraph/gnn/data_loading/bulk_sampler_io.py
```diff
@@ -54,35 +54,46 @@ def _write_samples_to_parquet(
...
 
-        results_p["batch_id"] = offsets_p.batch_id.repeat(
+        if end_batch_id - start_batch_id + 1 > len(offsets_p):
+            # This occurs when some batches returned 0 samples.
+            # To properly account this, the remaining batches are
+            # renumbered to have contiguous batch ids and the empty
+            # samples are dropped.
+            offsets_p.drop("batch_id", axis=1, inplace=True)
+            batch_id_range = cudf.Series(cupy.arange(len(offsets_p)))
+            end_batch_id = start_batch_id + len(offsets_p) - 1
```
VibhuJawa, line 85: I think because we are renumbering `batch_ids`  here , they would no longer correspond to the  input batch_ids, this is not a problem downstream but can cause problems in the future while trying to debug failures. 

I think we should make an explicit note of this somewhere in our code because this will non intutive once we forget the implementation details.

VibhuJawa: I think we should actually file a tech debt issue for this too , linking it in the note.

VibhuJawa (changes requested):
Thanks a lot for the crazy fast turn around time in fixing this @alexbarghi-nv .   Super impressive. 

While the implementation looks great i request adding a note about the implementation subtlety of changing batch ids (Which i honestly hope we should have the core team fix but we can not given the time constraints). 


The other ask i have is adding a test for this .

python/cugraph/cugraph/gnn/data_loading/bulk_sampler_io.py
alexbarghi-nv: @VibhuJawa I think this should be resolved now, please re-review

alexbarghi-nv:
@VibhuJawa tests have been added.

VibhuJawa (approved):
LGTM

BradReesWork (approved):

PR #3792, opened 2023-08-15
Expose threshold in louvain

### LINKED ISSUE #3791: Expose threshold parameter in Louvain for nx compatibility
The threshold parameter is exposed in the nx Louvain implementation.  We should expose this parameter through the C++ and C API.

This will provide the C API and C++ work to support #3700

ChuckHastings:
The threshold parameter (referred to as `epsilon` in most of the centrality measures) is used to define when to stop the iterative steps of Louvain.  Once the modularity increase for an iteration of Louvain is smaller than the threshold we will stop that iteration and start coarsening the graph.

This parameter was hard-coded in the initial C++ implementation of Louvain.  This PR exposes this parameter through the C++, C API, PLC and Python layers.

The PR also renames the python parameter `max_iter` to be `max_level`, which is more appropriate semantically.

Closes #3791


python/cugraph/cugraph/community/louvain.py
```diff
@@ -21,7 +21,8 @@
 from pylibcugraph import ResourceHandle
 
 
-def louvain(G, max_iter=100, resolution=1.0):
+# FIXME: max_iter should be renamed to max_level
```
ChuckHastings, line 24: I'm not sure how to do this cleanly in python.  We want to rename the parameter, but in a safe way (deprecating the old name and continuing to support it for a release.

Same FIXME is in place for the task version.

@rlratzel

python/cugraph/cugraph/community/louvain.py
```diff
@@ -41,17 +42,24 @@ def louvain(G, max_iter=100, resolution=1.0):
...
     max_iter : integer, optional (default=100)
-        This controls the maximum number of levels/iterations of the Louvain
+        This controls the maximum number of levels of the Louvain
         algorithm. When specified the algorithm will terminate after no more
-        than the specified number of iterations. No error occurs when the
+        than the specified number of levels. No error occurs when the
         algorithm terminates early in this manner.
 
-    resolution: float/double, optional (default=1.0)
+    resolution: float, optional (default=1.0)
```
ChuckHastings, line 50: Dropped double from the documentation.  Couldn't find a way to properly support both float and double through the cython code.

python/cugraph/cugraph/dask/community/louvain.py
```diff
@@ -45,19 +45,29 @@ def convert_to_cudf(result: cp.ndarray) -> Tuple[cudf.DataFrame, float]:
 
 
 def _call_plc_louvain(
-    sID: bytes, mg_graph_x, max_iter: int, resolution: int, do_expensive_check: bool
+    sID: bytes,
+    mg_graph_x,
+    max_iter: int,
+    threshold: float,
+    resolution: float,
```
ChuckHastings, line 52: Resolution was called type int.  Changed it to float, as that seems incorrect.  Or is that an undocumented trick to get around the fact that double wasn't supported?

python/pylibcugraph/pylibcugraph/_cugraph_c/community_algorithms.pxd
```diff
@@ -104,6 +104,18 @@ cdef extern from "cugraph_c/community_algorithms.h":
             cugraph_error_t** error
         )
     
+    cdef cugraph_error_code_t \
```
ChuckHastings, line 107: I should mark the above function (the old `cugraph_louvain`) as deprecated.  Not sure how to annotate that correctly.

python/pylibcugraph/pylibcugraph/louvain.pyx
```diff
@@ -51,6 +52,7 @@ from pylibcugraph.utils cimport (
 def louvain(ResourceHandle resource_handle,
             _GPUGraph graph,
             size_t max_level,
+            double threshold,
```
ChuckHastings, line 55: Interestingly, I could support double here.  Probably should make these consistent, left it in this odd state for discussion.

python/pylibcugraph/pylibcugraph/_cugraph_c/community_algorithms.pxd
rlratzel: I believe you can raise `DeprecationWarning` from the cython .pyx code for this.  Although, don't we want the cleaner `cugraph_louvain` function name longer term? `cugraph_louvain_with_threshold` seems clunkier.

rlratzel (commented):
Thanks for also updating the doc strings.  I left some comments below, and didn't approve yet due to the int/float/double cython questions.

python/cugraph/cugraph/community/louvain.py
ChuckHastings: Can't see a reason why a double would be necessary.

python/cugraph/cugraph/community/louvain.py
ChuckHastings: Updated as suggested

python/pylibcugraph/pylibcugraph/louvain.pyx
ChuckHastings: Changed these to float

naimnv (commented):
Reviewed the C++ parts and LGTM.

python/pylibcugraph/pylibcugraph/_cugraph_c/community_algorithms.pxd
seunghwak: +1 for making a breaking change if we are not aware of any users.

seunghwak (approved):

naimnv (approved):

jnke2016 (approved):

rlratzel (approved):
Thanks!

python/pylibcugraph/pylibcugraph/_cugraph_c/community_algorithms.pxd
rlratzel: I'm also a +1 for a breaking change to keep the better API name without the process required above, esp. since we're not aware of any users.

ChuckHastings: Made this change.

PR #3793, opened 2023-08-15
Adds updates to build wheel and conda packages for `cugraph-nx`

### LINKED ISSUE #3770: Update CI scripts to run native NetworkX unit tests using various `cugraph-nx` algos
The goal is to ensure the `cugraph-nx` implementations work as part of the NetworkX API via the dispatching mechanism and are drop-in compatible.

Update the CI scripts to install the native NetworkX unit tests (ie. the actual unit tests included in the NetworkX repo, written for NetworkX) and make them available to run as part of CI to verify the `cugraph-nx` implementations are drop-in compatible.

This task need not be specific to CI, and it may actually be better if developers could run the native NetworkX unit tests in their development environment too.

### LINKED ISSUE #3771: Add conda recipe for `cugraph-nx`
Add conda recipe for new `cugraph-nx` package.  
This task includes:

```[tasklist]
### Tasks
- [x] Add the `meta.yaml` file
- [x] Update `dependencies.yaml` if necessary
- [x] Notify Ops
```

### LINKED ISSUE #3772: Add setup.py/pyproject.toml for pip package for `cugraph-nx`
Add setup.py/pyproject.toml for pip package for `cugraph-nx`.

rlratzel:
closes #3771 
closes #3772 
closes #3770 

This PR adds updates to build new `cugraph-nx` wheels and conda packages, updates the CI scripts to test them, and also updates the auto-versioning script used during releases to also update `cugraph-nx`.

This PR also cleans up some tech debt in `dependencies.yaml` and various conda recipes (added missing meta-data).


conda/recipes/cugraph-nx/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  run:
+    - pylibcugraph ={{ version }}
+    - networkx >=3.1
+    - cupy >=12.0.0
+    - python
+
+tests:
+  imports:
+    - cugraph_nx
+
```
eriknw, line 33: It's becoming more common to run `pip check` as a sanity check in conda recipes. Dunno if any RAPIDS libraries do this:
```yaml
  commands:
    - pip check
  requires:
    - pip 
```

conda/recipes/cugraph-nx/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  build:
+      number: {{ GIT_DESCRIBE_NUMBER }}
+      string: py{{ py_version }}_{{ date_string }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+
+requirements:
+  host:
+    - python
+  run:
+    - pylibcugraph ={{ version }}
+    - networkx >=3.1
```
eriknw, line 26: Missed one:
```suggestion
    - networkx >=3.0
```

python/cugraph-nx/pyproject.toml
```diff
@@ -0,0 +1,219 @@
...
+    "networkx>=3.0",
+    "numpy>=1.21",
+    "pandas",
+    "pytest",
+    "pytest-benchmark",
+    "pytest-cov",
+    "pytest-mpl",
+    "pytest-xdist",
+    "scipy",
+] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
```
eriknw, line 49: Thanks for filling this out.

I did add coverage to `run_nx_tests.sh`, so we do need `pytest-cov`. `pytest-xdist` is _sometimes_ useful (and sometimes it uses all GPU memory).

We don't yet use `pandas` anywhere, and `scipy` is only used in a benchmark. But, these are in the "default" networkx dependencies, which we probably want to include if we plan to run networkx tests: https://github.com/networkx/networkx/blob/main/requirements/default.txt

btw, `pytest-mpl` will install `matplotlib`, which was included in "default.txt".

python/cugraph-nx/pyproject.toml
```diff
@@ -0,0 +1,219 @@
...
+    "Intended Audience :: Developers",
+    "Topic :: Software Development :: Libraries :: Python Modules",
+]
+dependencies = [
+    "cupy-cuda11x>=12.0.0",
+    "networkx>=3.0",
+    "pylibcugraph==23.10.*",
+] # This list was generated by `rapids-dependency-file-generator`. To make changes, edit ../../dependencies.yaml and run `rapids-dependency-file-generator`.
+
+[project.optional-dependencies]
```
eriknw, line 38: btw, I added `packaging` test dependency in #3808

python/cugraph-service/server/pyproject.toml
```diff
@@ -45,7 +45,7 @@ dynamic = ["entry-points"]
 test = [
     "aiohttp",
     "fsspec[http]>=0.6.0",
-    "networkx>=2.5.1",
+    "networkx>=3.0",
```
eriknw, line 48: Is it really necessary to update nx versions in other packages? It's probably simplest to keep dependency versions between packages in-sync, but cugraph-nx may at times want to depend on pretty new versions of networkx.

conda/recipes/cugraph-nx/meta.yaml
rlratzel: ok, I thought this would have been updated automatically via `dependencies.yaml` and `rapids-dependency-file-generator`, but I guess this needs to be manually set.  Will do in the next commit.

python/cugraph-nx/pyproject.toml
rlratzel: I think this is coming from the addition of `test_python_common` to cugraph-nx in `dependencies.yaml`, but perhaps I should redefine that for a new definition of "common" now.  I'll take care of this in the next commit.

python/cugraph-service/server/pyproject.toml
rlratzel: > Is it really necessary to update nx versions in other packages?

possibly not, `dependencies.yaml` makes it easy to align everything so I ended up doing that, but I'd be in favor of having all cugraph components always work with/depend on the latest stable Nx release though anyway.  Especially since (I think) we'll be migrating use cases that require cugraph to import networkx to `cugraph-nx` hopefully soon anyway.

Do you think aligning all cugraph components (including cugraph-nx) on the latest stable networkx will be a problem?  If so, I can make it so `cugraph-nx` depends on a different version of networkx than all other cugraph packages.

rlratzel: well...now I'm thinking about it more and perhaps we don't want to force Nx upgrades on people that don't want them, as long as we're compatible with the older versions.  `cugraph-nx` is a different case though in that it may require newer versions.  For that reason, we could have the other cugraph packages specify a wider range, eg. `>=2.5.1`.

However, keeping them all aligned could simply things such as development envs (which will probably just have to have a newer version to accommodate both cugraph* and cugraph-nx development), and CI runs which will benefit from scripts that only need to worry about a single networkx install.

I think I'm leaning towards aligning everything on the latest stable networkx, but could be convinced otherwise.  Thoughts?

eriknw: I think it's a good idea for all cugraph packages to be _compatible_ with the latest stable release of networkx.

For now, I expect cugraph-nx to be compatible with networkx versions `>=3.0`. However, I don't think it's worth much development effort for cugraph-nx to remain backwards compatible, and it's possible--even likely--that we'll want to depend on very recent versions in the future.

So, from my perspective, I want the freedom to be able to choose the networkx versions supported by cugraph-nx, and I don't think this should dictate what other cugraph packages support. What is the cost of doing it this way?

rlratzel: I think from the perspective of allowing different recipes to have different networkx version specs, we can do that pretty easily (we can have everything use the `networkx` anchor in `dependencies.yaml` except `cugraph-nx` which will specify its own version spec).  

However, it gets harder for our generated conda dev envs and CI scripts, which will likely need to settle on a single networkx version to keep from getting too complicated.  From that perspective, the `cugraph-nx` networkx minimum version will dictate what other cugraph packages have to support.  I'm fine with that though since we want cugraph* to work with the latest anyway, plus the amount of cugraph functionality that uses networkx is decreasing over time so my guess is we won't hit many (or any) incompatibilities anyway (famous last words, I know).

So I'm interpreting this as let's allow the `cugraph-nx` recipe to have a higher minimum version than the rest of cugraph, but test and develop all of cugraph with the version required by `cugraph-nx`.  If we run into incompatibilities, we update the cugraph code, remove the dependency, or look into different dev/test setups at that time.  Does that sound reasonable?

eriknw: Yeah, that sounds great to me. Let's plan to keep `cugraph-nx` working with the normal dev environment. This may limit how "bleeding edge" `cugraph-nx` can be, but the alternative seems worse.

conda/environments/all_cuda-118_arch-x86_64.yaml
```diff
@@ -38,6 +38,7 @@ dependencies:
 - nbsphinx
 - nccl>=2.9.9
 - networkx>=2.5.1
+- networkx>=3.0
```
rlratzel, line 41: `rapids-dependency-file-generator` is including the networkx version used by `cugraph-nx`, and the less-restrictive version independently used by other cugraph packages.  This did not seem to be a problem when testing the env files (`mamba` installed networkx version 3.1).

python/pylibcugraph/pyproject.toml
```diff
@@ -41,14 +41,12 @@ classifiers = [
...
 test = [
     "cudf==23.10.*",
-    "networkx>=2.5.1",
     "numpy>=1.21",
     "pandas",
     "pytest",
     "pytest-benchmark",
     "pytest-cov",
     "pytest-xdist",
-    "scipy",
```
rlratzel, line 51: I need to double check if it's safe to drop nx and scipy from the PLC test deps.

python/cugraph-nx/pyproject.toml
rlratzel: This still isn't completely resolved in the latest commit.  I still need to add the [networkx test deps](https://github.com/networkx/networkx/blob/main/requirements/default.txt) either as a test dep in depedencies.yaml, or in our CI script when we run against networkx tests.

python/cugraph-service/server/pyproject.toml
rlratzel: This might be resolved in the latest commit, but I want to see if it's okay that `rapids-dependency-file-generator` is [adding both networkx versions in the dev env file](https://github.com/rapidsai/cugraph/pull/3793#discussion_r1300290379).

python/cugraph-nx/pyproject.toml
eriknw: We'll probably want to add `scipy` and `pandas` to our test dependencies soon, so I'm okay if we have them listed here (or we could wait until we actually need them). Having them will help run some networkx tests.

ci/test_python.sh
```diff
@@ -88,6 +89,23 @@ pytest \
...
+  --junitxml="${RAPIDS_TESTS_DIR}/junit-cugraph-service.xml" \
+  --cov-config=../../.coveragerc \
+  --cov=cugraph_nx \
+  --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cugraph-nx-coverage.xml" \
+  --cov-report=term \
+  --benchmark-disable \
+  tests
+popd
+
+# FIXME: add call to test cugraph-nx against the networkx test suite
```
eriknw, line 107: Hopefully this FIXME is easily resolved once #3808 is merged by running `run_nx_tests.sh` with the necessary extra pytest arguments.

python/pylibcugraph/pyproject.toml
rlratzel: `scipy` is needed, it was added back.

conda/recipes/cugraph-nx/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://rapids.ai/
+  license: Apache-2.0
+  license_file: ../../../LICENSE
+  summary: cuGraph backend for GPU-accelerated NetworkX
```
eriknw, line 42: Why don't any of the conda recipes have
```yaml
  dev_url: https://github.com/rapidsai/cugraph
```
?

eriknw:
Does this PR actually add nightly wheels and conda packages? If so, the README for `cugraph-nx` should probably be updated, which currently says _`Nightly conda packages and pip wheels coming soon.`_

conda/recipes/cugraph-nx/meta.yaml
rlratzel: Good question!  I added those to all recipes in the most recent commit.

ci/test_python.sh
rlratzel: A call to `run_nx_tests.sh` has been added now.

raydouglass (approved):

python/cugraph-service/server/pyproject.toml
vyasr: This won't cause any issues for dfg itself. dfg does deduplicate dependencies, but it takes the entire dependency spec as a unit so the same package with different constraints will be treated as two different dependencies. Having two compatible constraints shouldn't cause any problems for pip or conda solvers though, so I don't see any issues with that.

eriknw (approved):
LGTM 🚀

alexbarghi-nv (approved):
👍

Issue #3773, opened 2023-08-07
Add support for Louvain community detection to `cugraph-nx`


rlratzel:
Add the code to allow NetworkX to dispatch to `cugraph-nx` for Louvain community detection.
This requires #3700 to be complete.
This also includes cugraph-nx unit tests, benchmarks, and docstrings.

PR #3803, opened 2023-08-16
Add `louvain_communities` to cugraph-nx

eriknw:
See: #3773 

Possible follow-up tasks:
- Update to use threshold parameter exposed from C++ (#3792)
- Add `max_level` argument to networkx implementation
  - ~Or, add `max_level` as extra`cugraph_nx`-specific argument~ (**done**)
- Update PLC to handle empty graphs gracefully (#3804)
- Update PLC to handle directed graphs
- Add `louvain_partitions` (needs added to PLC)
  - https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.community.louvain.louvain_partitions.html

This is passing many networkx tests. I don't have this as draft, b/c it's usable (and I would argue) mergable as is.


python/cugraph-nx/cugraph_nx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pylibcugraph as plc
+
+from cugraph_nx.convert import _to_undirected_graph
+from cugraph_nx.utils import _groupby, networkx_algorithm, not_implemented_for
+
+__all__ = ["louvain_communities"]
+
+
+@not_implemented_for("directed")
```
rlratzel, line 23: curious about this: is this better than allowing `_can_run` to return False, or is this something separate?  I'm mainly wondering if the UX is better.

python/cugraph-nx/cugraph_nx/algorithms/community/louvain.py
```diff
@@ -0,0 +1,52 @@
...
+    return [set(G._nodearray_to_list(node_ids)) for node_ids in groups.values()]
+
+
+@louvain_communities._can_run
+def _(G, weight="weight", resolution=1, threshold=0.0000001, seed=None):
+    # NetworkX allows both directed and undirected, but cugraph only allows undirected.
+    try:
+        return not G.is_directed()
+    except Exception:  # pragma: no cover (safety)
+        return False
```
rlratzel, line 52: This seems like it could mask a bug. Thoughts? I'm not as familiar with Nx conventions, but would a warning be appropriate if this handler end up being used?

python/cugraph-nx/cugraph_nx/algorithms/community/louvain.py
eriknw: `can_run` is only called when dispatching from networkx. Having `not_implemented_for` here will raise when the function is called directly with a directed graph.

python/cugraph-nx/cugraph_nx/algorithms/community/louvain.py
eriknw: Yeah, I was going back and forth on this one. We can drop it for now: `Errors should never pass silently`. On the other hand, if this exception is caught and we return False, then False is also valid since we're saying we can't run.

eriknw: Exception handling removed.

If desired, networkx could also use a diaper pattern around `can_run` and warn.

rlratzel (approved):

eriknw:
`cugraph-nx` tests are now passing CI :tada:

But, now `cugraph` tests are failing in CI.

PR #3809, opened 2023-08-17
Simplify wheel build scripts and allow alphas of RAPIDS dependencies

vyasr:
This PR makes a handful of changes aimed at simplifying the CI pipeline for building wheels as a precursor to switching RAPIDS nightlies to using proper alpha versions:
- Inlines apply_wheel_modifications.sh in build_wheel.sh. Now that the build doesn't rely excessively on logic in shared workflows, there's no real benefit to having a separate script (previously apply_wheel_modification.sh was a special script that the shared workflow knew to execute i.e. it was a hook into an externally controlled workflow).
- Consolidates the textual replacements using for loops and makes the replacements more targeted by only modifying the Python package being built in a given script. For instance, python/cugraph/pyproject.toml is no longer overwritten when building pylibcugraph.
- Modifies dependency specs for RAPIDS packages to include a `>=0.0.0a0` component. This is the key change that will allow alpha dependencies to be discovered. dask-cuda is the canary here because we already upload alphas of it, so the installation of cugraph in the test job should pull the latest dask-cuda alpha now without requiring direct installation from git.


vyasr:
A good sign: [the dask-cuda nightly is correctly being installed](https://github.com/rapidsai/cugraph/actions/runs/5965920597/job/16185946606?pr=3809#step:9:157).

ci/build_wheel_cugraph.sh
```diff
@@ -5,8 +5,12 @@ set -euo pipefail
 
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 
+# Download the pylibcugraph wheel built in the previous step and make it
+# available for pip to find. We must use PIP_FIND_LINKS because the package
+# must be made available to the isolated build step, and there is no way to
+# manually install it into that environment.
 RAPIDS_PY_WHEEL_NAME=pylibcugraph_${RAPIDS_PY_CUDA_SUFFIX} rapids-download-wheels-from-s3 ./local-pylibcugraph
-python -m pip install --no-deps ./local-pylibcugraph/pylibcugraph*.whl
+export PIP_FIND_LINKS=`pwd`/local-pylibcugraph
```
ajschmidt8, line 13:
```suggestion
export PIP_FIND_LINKS=$(pwd)/local-pylibcugraph
```

I know we don't run `shellcheck` on our scripts, but if we did, it would complain about the use of backticks due to the rule below

- https://www.shellcheck.net/wiki/SC2006

ci/build_wheel.sh
```diff
@@ -15,9 +15,35 @@ version_override="$(rapids-pip-wheel-version ${RAPIDS_DATE_STRING})"
...
+# For nightlies we want to ensure that we're pulling in alphas as well. The
+# easiest way to do so is to augment the spec with a constraint containing a
+# min alpha version that doesn't affect the version bounds but does allow usage
+# of alpha versions for that dependency without --pre
+alpha_spec=''
+if ! rapids-is-release-build; then
+    alpha_spec=',>=0.0.0a0'
+fi
+
+for dep in rmm cudf raft-dask pylibcugraph pylibraft ucx-py; do
```
ajschmidt8, line 37: I like the approach of the alpha spec.

I wish there were a way where we didn't have to re-specify cugraph's dependencies here.

though I don't think it's the end of the world.

ajschmidt8 (approved):

ci/build_wheel_cugraph.sh
vyasr: Good to know, thanks for the info!

ci/build_wheel.sh
vyasr: Agreed, in the future we can investigate if there's a less duplicative dfg-based solution.

PR #3825, opened 2023-08-25
Updating docs and build script

BradReesWork:
Updated Doxygen to produce xml for Sphinx

Updating the build script to:
* have a build all option
* build doxygen and then move XML 

This will eventually drop the need to publish the C docs


acostadon (changes requested):
Found a bad link in basics/cugraph_intro.html
the link to the project board doesn't work.
https://github.com/rapidsai/cugraph/projects/28
that wasn't changed in your pr.

acostadon:
installation/source_build.html
2 typos
Set the localtion to cuGraph in an environment variable CUGRAPH_HOME     
location

Download the cuGraph repo - if you have a folked version, use that path here instead
forked

Where XXX is the CUDA 11 version
conda env update --name cugraph_dev --file conda/environments/all_cuda-1[XXX]_arch-x86_64.yml
cuda 11 or 12 now

BradReesWork:
To be broken into multiple smaller PRs

PR #3830, opened 2023-08-28
Add nx-cugraph notebook for showing accelerated networkX APIs

betochimas:
This PR adds a notebook showing how to use the `nx-cugraph` package to accelerate your NetworkX workflow.

closes rapidsai/graph_dl#277


cugraph_nx_demo.ipynb
```diff
@@ -0,0 +1,504 @@
...
+   "source": [
+    "%%timeit\n",
+    "bc = nx.betweenness_centrality(Gnx, k=50, backend='cugraph')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Method 2: Perform backend conversion on a separate line"
```
rlratzel, line 402: I'm thinking method 2 should be less common, but we're actually going over these use cases so there's a good chance these may change.  Regardless of that, this needs to explain why someone would use one method over the other, the tradeoffs of one over the other, etc.

cugraph_nx_demo.ipynb
rlratzel, line 1: This is a good start, but I'm thinking we'll also want to show the reader how this compares to a native NX workflow, using the same algos and dataset, to really showcase why they'd want to use the cugraph backend.

cugraph_nx_demo.py
rlratzel, line 1: Just wondering if this is here only to get reviewed for understanding purposes, or if it's going to be committed to our repo somewhere.  If the latter, where would this live?

rlratzel (commented):
Thanks, Dylan, looking good so far. I added a few items below.

cugraph_nx_demo.py
betochimas: I used this script before migrating the code blocks to the notebook. The benchmarks are outdated and you have more comprehensive benchmarks, so this will be removed in the next commit.

cugraph_nx_demo.ipynb
betochimas: When you mean 'a native NX workflow', does that include pandas and pd.from_edgelist? I assumed the user doesn't have the email_Eu_core dataset, but should I not assume that? If not, then I would include a cell downloading the email_Eu_core within the current working directory (I wouldn't include the download cell in the benchmarking and treat it as pre-processing).

cugraph_nx_demo.ipynb
```diff
@@ -0,0 +1,504 @@
...
+    "You may need to restart the kernel to have the development version of networkX be visible."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
```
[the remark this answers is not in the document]
betochimas: I included the output of the notebook cells so you, or any other reviewer, could reproduce it. The entire ## FOR REPRODUCIBILITY / DEV WORK / TEMP chunk would be removed, since the cells are highly specific to my own workflow, anyway.

I will include the benchmark cells' outputs and clear the output for the rest. If you can provide me your benchmarks from the larger datasets, I can include them on the notebook as a graphic. It would help highlight performance differences, and it would also be useful for other things too.

cugraph_nx_demo.ipynb
```diff
@@ -0,0 +1,504 @@
...
+   "metadata": {},
+   "source": [
+    "# How to use `cugraph-nx` to accelerate graph analysis while using networkX "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Using the cugraph plugin for NetworkX requires Python 3.9, 3.10, or 3.11, in addition to the latest development version of NetworkX. Follow the instructions from `INSTALL.rst` on the NetworkX repository for instructions on installing the development version of `networkx`"
```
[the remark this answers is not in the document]
betochimas: Thanks, I'll remove that.

cugraph_nx_demo.ipynb
```diff
@@ -0,0 +1,504 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df = pd.read_csv(\"datasets/email-Eu-core.csv\", header=None, names=['src', 'dst', 'wgt'], sep=' ')\n",
+    "Gnx = nx.from_pandas_edgelist(df, source='src', target='dst', create_using=nx.Graph())"
```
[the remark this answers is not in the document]
betochimas: Got it, thanks

cugraph_nx_demo.ipynb
```diff
@@ -0,0 +1,504 @@
...
+   "metadata": {},
+   "source": [
+    "## Customization"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Depending on your graph needs, `networkx` supports the ability to choose your dispatcher. A dispatchable networkX algorithm will use the chosen dispatcher to convert networkx Graph to the backend equivalent, dispatch the work, then convert back to the required nx form. More information here: https://networkx.org/documentation/stable/reference/classes/index.html#backends"
```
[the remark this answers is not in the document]
betochimas: Sorry, I thought there was a way to create a dispatcher along with a way to create a backend. I wrote this cell using https://networkx.org/documentation/stable/reference/classes/index.html#backends  as a reference (also why I included that as a link!), which is, I suppose, outdated reference material.
I'll correct this info on the next commit.

cugraph_nx_demo.ipynb
betochimas: I think this will be rewritten substantially, since the API is going many changes and quite a bit isn't set in stone. However, just looking at these two methods, I agree that method 1 is easier to use over method 2. Method 1 is conducive to devs refactoring their code to introduce using one of the backends, but even then setting a dispatching order in a configs file / location makes more sense. I suppose Method 2 is helpful for explicitly showing the conversion and also if the same Graph object is going to be used across multiple calls. I'll make a thread here as I compare the methods/use cases more intensely.

betochimas:
Currently this notebook
- Shows which versions of Python, NetworkX, and cuGraph are compatible, where and how to install dev version of nX
- A brief example of native NetworkX and accelerated NetworkX, with benchmarks
- A section on customization tools to configure backend behavior
- Lists which algs are supported

rlratzel (approved):

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+    "## Background\n",
+    "Networkx version 3.0 introduced a dispatching mechanism that allows users to configure NetworkX to dispatch various algorithms to third-party backends. Backends can provide different implementations of graph algorithms, allowing users to take advantage of capabilities not available in NetworkX. `nx-cugraph` is a NetworkX backend provided by the [RAPIDS](https://rapids.ai) cuGraph project that adds GPU acceleration to greatly improve performance.\n",
+    "\n",
+    "## System Requirements\n",
+    "Using `nx-cugraph` with this notebook requires the following: \n",
+    "- NVIDIA GPU, Pascal architecture or later\n",
+    "- CUDA 11.2, 11.4, 11.5, 11.8, or 12.0\n",
+    "- Python versions 3.9, 3.10, or 3.11\n",
+    "- NetworkX >= version 3.2\n",
+    "  - _NetworkX 3.0 supports dispatching and is compatible with `nx-cugraph`, but this notebook will demostrate features added in 3.2_\n",
```
eriknw, line 25:
```suggestion
    "  - _NetworkX 3.0 supports dispatching and is compatible with `nx-cugraph`, but this notebook will demonstrate features added in 3.2_\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### Automatic GPU acceleration\n",
+    "When `nx-cugraph` is installed, `networkx` will detect it on import and make it available as a backend for APIs supported by that backend.  However, `networkx` does not assume the user always wants to use a particular backend, and instead looks at various configuration mechanisms in place for users to specify how `networkx` should use installed backends. Since `networkx` was not configured to use a backend for the above `betweenness_centrality` call, it used the default implementation provided by NetworkX.\n",
+    "\n",
+    "The first configuration mechanism to be demonstrated below is the `NETWORKX_AUTOMATIC_BACKENDS` environment variable.  This environment variable directs `networkx` to use the backend specified everywhere it's supported, and does not require the user to modify any of their existing NetworkX code.\n",
```
eriknw, line 208:
```suggestion
    "The first configuration mechanism to be demonstrated below is the `NETWORKX_AUTOMATIC_BACKENDS` environment variable.  This environment variable directs `networkx` to use the backend specified everywhere it's supported and does not require the user to modify any of their existing NetworkX code.\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+    "bash> export NETWORKX_AUTOMATIC_BACKENDS=cugraph\n",
+    "bash> python my_nx_app.py  # uses nx-cugraph wherever possible, then falls back to default implementation where it's not.\n",
+    "```\n",
+    "or in the case of multiple backends installed\n",
+    "```\n",
+    "bash> export NETWORKX_AUTOMATIC_BACKENDS=cugraph,graphblas\n",
+    "bash> python my_nx_app.py  # uses nx-cugraph if possible, then nx-graphblas if possible, then default implementation.\n",
+    "```\n",
+    "\n",
+    "`networkx` looks at the environment variable and the installed backends at import time, and will not re-examine the environment after that.  Because `networkx` was already imported in this notebook, the `reimport_nx()` utility will be called after the `os.environ` disctionary is updated to simulate an environment variable being set in the shell.\n",
```
eriknw, line 221:
```suggestion
    "`networkx` looks at the environment variable and the installed backends at import time and will not re-examine the environment after that.  Because `networkx` was already imported in this notebook, the `reimport_nx()` utility will be called after the `os.environ` dictionary is updated to simulate an environment variable being set in the shell.\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+    "karate_cg_bc_results = nx.betweenness_centrality(karate_club_graph)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "We may see that the same computation actually took *longer* using `nx-cugraph`. This is not too surprising given how small the graph is, since there's a small amount of overhead to copy data to and from the GPU which becomes more obvious on very small graphs.  We'll see with a larger graph how this overhead becomes negligable."
```
eriknw, line 273:
```suggestion
    "We may see that the same computation actually took *longer* using `nx-cugraph`. This is not too surprising given how small the graph is, since there's a small amount of overhead to copy data to and from the GPU which becomes more obvious on very small graphs.  We'll see with a larger graph how this overhead becomes negligible."
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### Results Comparison\n",
+    "\n",
+    "Let's examine the results of each run to see how they compare.  \n",
+    "The `betweenness_centrality` results are a dictionary mapping vertex IDs to betweenness_centrality scores.  The score itself is usually not as important as the relative rank of each vertex ID (eg. vertex A is ranked higher than vertex B in both sets of results)."
```
eriknw, line 285:
```suggestion
    "The `betweenness_centrality` results are a dictionary mapping vertex IDs to betweenness_centrality scores.  The score itself is usually not as important as the relative rank of each vertex ID (e.g. vertex A is ranked higher than vertex B in both sets of results)."
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "### `betweenness_centrality` on larger graphs - The U.S. Patent Citation Network<sup>1</sup>\n",
+    "\n",
+    "The U.S. Patent Citation Network dataset is much larger with over 3.7M nodes and over 16.5M edges, and demonstrates how `nx-cugraph` enables NetworkX to run `betweenness_centrality` on graphs this large (and larger) in seconds instead of minutes.\n",
```
eriknw, line 363:
```suggestion
    "The U.S. Patent Citation Network dataset is much larger with over 3.7M nodes and over 16.5M edges and demonstrates how `nx-cugraph` enables NetworkX to run `betweenness_centrality` on graphs this large (and larger) in seconds instead of minutes.\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+    "    return G"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "## Running `betweeness_centrality`\n",
```
eriknw, line 147:
```suggestion
    "## Running `betweenness_centrality`\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os\n",
+    "os.environ[\"NETWORKX_AUTOMATIC_BACKENDS\"] = \"cugraph\"\n",
+    "nx = reimport_networkx()\n",
+    "# reimporting nx requires reinstantiating Graphs since python consideres\n",
```
eriknw, line 235:
```suggestion
    "# reimporting nx requires reinstantiating Graphs since python considers\n",
```

notebooks/demo/nx_cugraph_demo.ipynb
```diff
@@ -0,0 +1,672 @@
...
+    "nx.betweenness_centrality(nxcg_cit_patents_graph, k=k)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Conclusion\n",
+    "\n",
+    "This notebook demostarated `nx-cugraph`'s support for `betweenness_centrality`.  At the time of this writing, `nx-cugraph` also provides support for `edge_netweenness_centrality` and `louvain_communities`.  Other algorithms are scheduled to be supported based on their availability in the cuGraph [pylibcugraph](https://github.com/rapidsai/cugraph/tree/branch-23.10/python/pylibcugraph/pylibcugraph) package and demand by the NetworkX community.\n",
```
eriknw, line 626:
```suggestion
    "This notebook demonstrated `nx-cugraph`'s support for `betweenness_centrality`.  At the time of this writing, `nx-cugraph` also provides support for `edge_betweenness_centrality` and `louvain_communities`.  Other algorithms are scheduled to be supported based on their availability in the cuGraph [pylibcugraph](https://github.com/rapidsai/cugraph/tree/branch-23.10/python/pylibcugraph/pylibcugraph) package and demand by the NetworkX community.\n",
```

eriknw (commented):
Pretty good overall! I left a few suggestions for grammar and spelling mistakes.

Another thing to consider is usage of "networkx" and "NetworkX".

notebooks/demo/nx_cugraph_demo.ipynb
rlratzel: Thanks for finding this and the other typos below. I think I should also be using `networkx` and NetworkX more consistently in the latest commits.

Issue #3790, opened 2023-08-14
Graph creation fails when copying the dask input dataframe


jnke2016:
### Describe the bug.

The `cuGraph` graph creation from a dask cudf dataframe fails when copying the input dask dataframe with dask `map_partitions` right before equally persisting the dask dataframe, for cases where some partitions have empty dataframe.

### Minimum reproducible example

```shell
    # The datasets must have less edges than there are workers
    ddf = dask_cudf.read_csv(
        input_data_path,
        chunksize=chunksize,
        delimiter=" ",
        names=["src", "dst", "value"],
        dtype=["int32", "int32", "float32"],
    )

    dg1 = cugraph.Graph(directed=True)
    dg1.from_dask_cudf_edgelist(ddf, source="src", destination="dst", edge_attr="value")

    dg2 = cugraph.Graph(directed=True)
    dg2.from_dask_cudf_edgelist(ddf, source="src", destination="dst", edge_attr="value")
```

### Relevant log output

```shell
def set_by_label(self, key: Any, value: Any, validate: bool = True):
            """
            Add (or modify) column by name.
    
            Parameters
            ----------
            key
                name of the column
            value : column-like
                The value to insert into the column.
            validate : bool
                If True, the provided value will be coerced to a column and
                validated before setting (Default value = True).
            """
            key = self._pad_key(key)
            if validate:
                value = column.as_column(value)
                if len(self._data) > 0:
                    if len(value) != self._column_length:
                        raise ValueError("All columns must be of equal length")
                else:
                    self._column_length = len(value)
    
            self._data[key] = value
            self._clear_cache()
    
        def _select_by_names(self, names: abc.Sequence) -> Self:
            return self.__class__(
                {key: self[key] for key in names},
                multiindex=self.multiindex,
                level_names=self.level_names,
            )
    
        def _select_by_label_list_like(self, key: Any) -> ColumnAccessor:
            # Might be a generator
            key = tuple(key)
            # Special-casing for boolean mask
            if (bn := len(key)) > 0 and all(map(is_bool, key)):
                if bn != (n := len(self.names)):
                    raise IndexError(
                        f"Boolean mask has wrong length: {bn} not {n}"
                    )
                data = dict(
                    item
                    for item, keep in zip(self._grouped_data.items(), key)
                    if keep
                )
            else:
                data = {k: self._grouped_data[k] for k in key}
            if self.multiindex:
                data = _to_flat_dict(data)
            return self.__class__(
                data,
[...]

PR #3837, opened 2023-08-30
Enable temporarily disabled MG tests

naimnv:
Enable TEMPORARILY disable single-GPU "MG" tests
And Skip deleting copied Dataframe while creating distributed graph from cudf edge-lists.
Ideally we would like to merger this PR once the [issue 3790](https://github.com/rapidsai/cugraph/issues/3790) is closed, but me  might need to merger it if the issue is not resolved before the next release.


rlratzel (approved):

ajschmidt8 (approved):

rlratzel (approved):

PR #3841, opened 2023-09-01
Integrate C++ Renumbering and Compression

alexbarghi-nv:
- [x] C API
- [x] PLC
- [x] Python API
- [x] Bulk Sampling API
- [x] Documentation for Python SG
- [x] Documentation for Python MG
- [x] Documentation for Bulk Sampler
- [x] Resolve the C++ empty batch issue with new check
- [x] Add FutureWarnings for all temporary flags
- [x] Remove all print statements and pytest tags
- [x] Verify cuGraph-PyG and cuGraph-DGL tests


ChuckHastings (approved):
Reviewed C++ and C API modifications.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1894,6 +1894,21 @@ k_core(raft::handle_t const& handle,
...
+ *            is compressed into a row pointer.
+ * @param CSC Compresses in CSC format.  This means the col (dst) column
+ *            is compressed into a column pointer.
+ * @param DCSR Compresses in DCSR format.  This outputs an additional index
+ *             that avoids empty entries in the row pointer.
+ * @param DCSC Compresses in DCSC format.  This outputs an additional index
+ *             that avoid empty entries in the row pointer.
+ */
+enum class compression_type_t { COO = 0, CSR, CSC, DCSR, DCSC };
+
```
seunghwak, line 1911: This is used only in the C API, right? (no use in the C++ code) Better move this to a C-API file.

cpp/src/sampling/sampling_post_processing_impl.cuh
```diff
@@ -166,9 +166,14 @@ void check_input_edges(
                                               std::numeric_limits<label_index_t>::max()),
                   "Invalid input arguments: current implementation assumes that the number of "
                   "unique labels is no larger than std::numeric_limits<uint32_t>::max().");
+
+  // FIXME figure out a version of this condition that still allows empty batches
+  /*
   CUGRAPH_EXPECTS(!edgelist_label_offsets || std::get<1>(*edgelist_label_offsets) > 0,
                   "Invlaid input arguments: there should be 1 or more labels if "
                   "edgelist_label_offsets.has_value() is true.");
+  */
```
seunghwak, line 175: You can just delete this check if you want this function to handle empty batches (or you can just skip calling this function for empty batches). Just make sure that the returned value still makes sense if you call this with an empty batch.

cpp/include/cugraph/algorithms.hpp
alexbarghi-nv: fixed

cpp/src/sampling/sampling_post_processing_impl.cuh
alexbarghi-nv: There is a Python test specifically for empty batches, so I think we are ok here.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
```diff
@@ -58,15 +60,20 @@ def uniform_neighbor_sample(
...
     random_state: int = None,
     return_offsets: bool = False,
     return_hops: bool = True,
+    include_hop_column: bool = True,  # deprecated
     prior_sources_behavior: str = None,
     deduplicate_sources: bool = False,
     renumber: bool = False,
+    use_legacy_names=True,  # deprecated
+    compress_per_hop=False,
+    compression="COO",
```
tingyu66, line 76:
```suggestion
    use_legacy_names: bool = True,  # deprecated
    compress_per_hop: bool = False,
    compression: str = "COO",
```

tingyu66 (approved):
LGTM. Left some minor comments.
On a side note, I think it would be helpful to briefly document the parameters needed in `uniform_neighbor_sampler()` for DGL and PyG integration respectively, as there are already a few kwargs to keep in mind.

seunghwak (approved):
Approving the C++ part

jnke2016 (approved):
Looks good to me

BradReesWork (approved):

PR #3842, opened 2023-09-02
Use new `raft::compiled_static` targets

divyegala:
https://github.com/rapidsai/raft/pull/1746 added static targets for RAFT that can be used directly now instead of building RAFT static explicitly

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -364,7 +364,7 @@ if (USE_CUGRAPH_OPS)
             $<$<NOT:$<BOOL:${CUDA_STATIC_RUNTIME}>>:raft::compiled>
         PRIVATE
             $<$<BOOL:${CUDA_STATIC_RUNTIME}>:raft::raft>
-            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:raft::compiled>
+            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:raft::compiled_static>
             cuco::cuco
             cugraph::cuHornet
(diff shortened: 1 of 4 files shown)


cpp/CMakeLists.txt
```diff
@@ -346,11 +346,20 @@ target_include_directories(cugraph
         "$<INSTALL_INTERFACE:include>"
 )
 
+set(COMPILED_RAFT_LIB_STATIC "")
```
robertmaynard, line 349:
```suggestion
```

cpp/CMakeLists.txt
```diff
@@ -346,11 +346,20 @@ target_include_directories(cugraph
...
+set(COMPILED_RAFT_LIB "")
 if(CUDA_STATIC_RUNTIME)
   get_target_property(_includes raft::raft INTERFACE_INCLUDE_DIRECTORIES)
   target_include_directories(cugraph PUBLIC ${_includes})
   # Add CTK include paths because we make our CTK library links private below
   target_include_directories(cugraph SYSTEM PUBLIC ${CUDAToolkit_INCLUDE_DIRS})
+  if(CUGRAPH_COMPILE_RAFT_LIB)
+    set(COMPILED_RAFT_LIB_STATIC raft::compiled_static)
```
robertmaynard, line 357:
```suggestion
    set(COMPILED_RAFT_LIB raft::compiled_static)
```

robertmaynard (changes requested):

ChuckHastings (approved):

PR #3844, opened 2023-09-05
Add benchmark datasets yamls files to use with Dataset

betochimas:
This PR adds the soc-LiveJournal, cit-Patents, europe_osm, hollywood, and twitter Dataset objects so that Dataset can be used for benchmarking purposes.
When https://github.com/rapidsai/ops/issues/2840 is closed, users will be able to use Dataset to easily download these datasets and load them into Dataframes and Graph instances.
Eventually, all benchmarking components using these datasets will be refactored to save overhead when running CI, creating demo notebooks, and obtaining benchmark results.
Since twitter is larger than the rest (by at least 4x), adding the twitter metadata yaml is optional. 

Also features some changes to Dataset.get_edgelist and adds support for different column names other than 'src', 'dst', 'wgt'


betochimas:
- Added the 'create_using' arg in Dataset.get_edgelist to add support for cuDF-like (pandas) read_csv. Would it be better to move this to a separate PR?
- Found that the has_loop('livejournal' dataset) and has_loop('hollywood' dataset) both return True when they should not have self-loops. Check 'has_loop' implementation or datasets to see why this is the case and fix it (MAIN blocking point for this PR)
- Added hollywood into Datasets namespace (also on s3), but is commented-out because such a large dataset requires more memory (wait until dask_cudf is supported on Dataset??)
- Added support for different column names other than 'src', 'dst', 'wgt'

PR #3848, opened 2023-09-07
Add entry point to tell NetworkX about nx-cugraph without importing it.

eriknw:
This allows NetworkX docstrings to be updated (among other things).

This will have a companion PR in NetworkX. We still need to determine (and agree) on the dict returned by this entry point, and NetworkX doesn't need to use everything I have here. We should probably add a string for `"description"` that gives a very short description of the backend, and maybe `"url"` or `"homepage"` or whatever so online docs can have links.

Here's how to use the entry point (Python >= 3.10) after installing it:
```python
In [1]: from importlib.metadata import entry_points

In [2]: items = entry_points(group="networkx.plugin_info")

In [3]: [plugin] = items

In [4]: plugin.load()()
Out[4]:
{'backend_name': 'cugraph',
 'project': 'nx-cugraph',
 'package': 'nx_cugraph',
 'functions': {'betweenness_centrality',
  'edge_betweenness_centrality',
  'louvain_communities'},
 'extra_docstrings': {'betweenness_centrality': '`weight` parameter is not yet supported.',
  'edge_betweenness_centrality': '`weight` parameter is not yet supported.',
  'louvain_communities': '`threshold` and `seed` parameters are currently ignored.'},
 'extra_parameters': {'louvain_communities': {'max_level': 'Upper limit of the number of macro-iterations.'}}}
```

CC @rlratzel @betochimas


python/nx-cugraph/_nx_cugraph/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+            "louvain_communities": "`threshold` and `seed` parameters are currently ignored.",
+            # END: extra_docstrings
+        },
+        "extra_parameters": {
+            # BEGIN: extra_parameters
+            "louvain_communities": {
+                "max_level": "Upper limit of the number of macro-iterations.",
+            },
+            # END: extra_parameters
+        },
```
eriknw, line 53: I think it might be nice to break out extra parameters like this. However, this does not yet have the parameter _type_, which is often used in docstrings such as:
```
    Parameters
    ----------
    max_level : int, optional
      Upper limit of the number of macro-iterations.
```
Should we try to include the types?

eriknw: Added. Answered my own question--sorry for the noise!

eriknw:
Companion networkx PR is up: https://github.com/networkx/networkx/pull/6911

python/nx-cugraph/_nx_cugraph/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+            "louvain_communities": "`threshold` and `seed` parameters are currently ignored.",
+            # END: extra_docstrings
+        },
+        "extra_parameters": {
+            # BEGIN: extra_parameters
+            "louvain_communities": {
+                "max_level : int, optional": "Upper limit of the number of macro-iterations.",
+            },
+            # END: extra_parameters
+        },
```
eriknw, line 56: Alternatively, we could organize info like:
```python
        "functions": {
            # BEGIN: functions
            "betweenness_centrality": {
                "extra_docstring": "`weight` parameter is not yet supported.",
            },
            "edge_betweenness_centrality" : {
                "extra_docstring": "`weight` parameter is not yet supported.",
            },
            "louvain_communities": {
                "extra_docstring": "louvain_communities": "`threshold` and `seed` parameters are currently ignored.",
                "extra_parameters": {
                    "max_level : int, optional": "Upper limit of the number of macro-iterations.",
                },
            },
            # END: functions
        },
```

python/nx-cugraph/nx_cugraph/__init__.py
```diff
@@ -12,9 +12,21 @@
...
 from .classes import *
+
+from . import convert
 from .convert import *
 
+# from . import convert_matrix
+# from .convert_matrix import *
+
+# from . import generators
+# from .generators import *
```
rlratzel, line 27: Is this dead code we can remove?

python/nx-cugraph/_nx_cugraph/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+    d["functions"] = {
+        func: {key: vals[func] for key in func_keys if func in (vals := d[key])}
+        for func in d["functions"]
+    }
+    for key in func_keys:
+        del d[key]
+    return d
+
+
+__version__ = "23.10.00"
```
rlratzel, line 71: Should [update-version.sh](https://github.com/rapidsai/cugraph/blob/branch-23.10/ci/release/update-version.sh) be updated for this?

rlratzel (commented):
Looked good, nothing jumped out.  I had a couple of questions though.

python/nx-cugraph/nx_cugraph/__init__.py
eriknw: It is future code, yet to be born

python/nx-cugraph/_nx_cugraph/__init__.py
eriknw: Good catch! Done.

rlratzel (approved):

raydouglass (approved):

eriknw:
@rlratzel mentioned we should update louvain to handle isolated nodes as done in #3886. I'll go ahead and make this change in this PR.

PR #3852, opened 2023-09-08
Updates to build and test `nx-cugraph` wheel as part of CI and nightly workflows

rlratzel:
closes rapidsai/graph_dl#302

* Updates GHA yaml files to build and test a `nx-cugraph` wheel
* Adds CI scripts for building and testing the `nx-cugraph` wheel
* Adds a smoketest script for `nx-cugraph`
* Relevant code cleanup: removes unnecessary dataset download from cugraph wheel testing


ci/wheel_smoke_test_nx-cugraph.py
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import math
+
+import networkx as nx
+import nx_cugraph
```
eriknw, line 17: Aw, this doesn't use the import convention we discussed:
```suggestion
import nx_cugraph as nxcg
```
Usage of `nx_cugraph` below would also need updated.

ci/test_wheel.sh
```diff
@@ -9,9 +9,9 @@ package_dir=$2
 mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 
-# echo to expand wildcard before adding `[extra]` requires for pip
```
rlratzel, line 12: There was a case where the wildcard did not match the .whl files on disk and 'echo' simply echo'd the wildcard pattern, which made `pip` generate a misleading error message.  Using 'ls' will result in the script erroring out with a clear message about the missing/mis-named files.

raydouglass (approved):

ci/check_style.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright (c) 2020-2022, NVIDIA CORPORATION.
+# Copyright (c) 2020-2023, NVIDIA CORPORATION.
```
vyasr, line 2: Probably a holdover from an earlier iteration? There are no changes in this file other than the copyright.

ci/test_wheel.sh
vyasr: Odd, but I'm fine with the change if it works.

ci/test_wheel_cugraph.sh
```diff
@@ -11,12 +11,4 @@ python -m pip install --no-deps ./local-pylibcugraph-dep/pylibcugraph*.whl
 # Always install latest dask for testing
 python -m pip install git+https://github.com/dask/dask.git@main git+https://github.com/dask/distributed.git@main
 
-# Only download test data for x86
-arch=$(uname -m)
-if [[ "${arch}" == "x86_64" ]]; then
-    pushd ./datasets
-    bash ./get_test_data.sh
-    popd
-fi
```
vyasr, line 20: I guess this change is necessary for nightlies? Were nightly tests for arm not running the tests that required datasets?

vyasr (approved):
Couple of notes, but looks fine otherwise.

ci/check_style.sh
rlratzel: Thanks, I reverted that in the latest commit.

ci/test_wheel.sh
bdice: I agree with this change, I see the problem with shell wildcard expansion.

ci/test_wheel_cugraph.sh
bdice: We shouldn't remove this, but might need to modify it to only be skipped in ARM PR jobs. In ARM PR jobs, we can only run "smoke tests" due to limited resource capacity.

https://github.com/rapidsai/cugraph/blob/5c34d3dd340c76678d8f2667057c6b0ce2f1f480/ci/test_wheel.sh#L16

Downloading the datasets is time-intensive and should not be performed for "smoke tests" in ARM PR jobs because it wastes a lot of time on the ARM GPU node.

bdice (changes requested):
Requesting changes. We must skip dataset downloads for ARM smoke tests (ARM PRs).

ci/test_wheel_cugraph.sh
bdice: I'm terribly sorry. I read this backwards. This removed downloading datasets for ALL jobs (all arches, nightly/PR). Are we not dependent on datasets for tests to pass anymore?

rlratzel: The `get_test_data.sh` script will download supplemental datasets that are too big to commit to our github repo.  Those are currently only used by C++ tests.  The python tests use the smaller .csv datasets committed to the repo [here](https://github.com/rapidsai/cugraph/tree/branch-23.10/datasets), so it's safe to skip downloading when only testing python code.

I only removed it here since it was in the proximity of the changes for the wheel builds, but I think an audit of what runs require downloading the supplemental datasets would be good for a separate PR.

Prior to the recent commits to the PR which triggered CI running now, we had everything passing with the download step removed.

bdice: Thanks @rlratzel, I'll approve now.

bdice (approved):

PR #3857, opened 2023-09-12
Add tests for `Resultset`, some refactoring of `Resultset` and `Dataset`

betochimas:
This PR primarily adds testing for the `Resultset` class, introduced earlier in 23.10. The tests take a similar approach to test_dataset, creating a temporary directory to test downloading all result files. To align `Resultset` and `Dataset`, the setter and getter for each download directory is moved into `DefaultDownloadDir`, so that each class shares an instance of `DefaultDownloadDir` and can be configured independently, although their default locations are still both dependent on the RAPIDS_DATASET_ROOT_DIR_PATH environment variable. The old patterns are present but commented-out, so this change would be breaking.


betochimas:
- Moves set_download_dir and get_download_dir into DefaultDownloadDir class
- Both default_download_dir and default_resultset_download_dir are instances of DefaultDownloadDir
- Adds test_resultset.py which is compatible, created with the above refactoring in mind
- Old Dataset and Resultset helpers are commented out in case the above refactoring is scrapped / moved to another PR
- Passes CI without issue as it stands

PR #3860, opened 2023-09-18
Link prediction notebook and dining dataset added

acostadon:
Adding notebook to run all link predictions and a new small dataset with string names for nodes


datasets/dining_prefs.mtx
```diff
@@ -0,0 +1,54 @@
+%%MatrixMarket matrix coordinate pattern symmetric
```
BradReesWork, line 1: Why do we need MTX file

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
+name: dining_prefs
+file_type: .csv
+author: J.L. Moreno
```
nv-rliu, line 3: Between `file_type` and `author`, we should add a `description` field that briefly explains what this graph represents

datasets/dining_prefs.csv
```diff
@@ -0,0 +1,52 @@
+Ada Cora 1
```
BradReesWork, line 1: I thought that dataset was being saved in the cloud and accessed via:  https://data.rapids.ai/cugraph/datasets/dining_prefs.csv

so why do we needed that datafile here?

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
...
+url:  https://data.rapids.ai/cugraph/datasets/dining_prefs.csv
```
nv-rliu, line 4: All the metadata fields are standardized now so the `url` field is at the very end.

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - string
+  - string
+  - int
+has_loop: true
+is_directed: true
+is_multigraph: false
+is_symmetric: true
+number_of_edges: 52
+number_of_nodes: 26
+number_of_lines: 52
```
nv-rliu, line 23: The other YAML files no longer include `number_of_lines` in their metadata. Just # nodes and # edges

BradReesWork (changes requested):

PR #3866, opened 2023-09-20
adding dining preference dataset 

acostadon:
This dataset is very small, and uses strings as node names. It will be used to test force atlas, and in a new link prediction/similarity notebook.

the licensing is contained here.
http://networkdata.ics.uci.edu/netdata/html/Dining-table_partners.html


python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
+name: dining_prefs
+file_type: .csv
+author: J.L. Moreno
```
nv-rliu, line 3: Between `file_type` and `author`, we should add a `description` field that briefly explains what this graph represents

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
...
+url:  https://data.rapids.ai/cugraph/datasets/dining_prefs.csv
```
nv-rliu, line 4: Move `url` field to the end of the file

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - string
+  - string
+  - int
+has_loop: false
+is_directed: false
+is_multigraph: false
+is_symmetric: true
+number_of_edges: 42
+number_of_nodes: 26
+number_of_lines: 52
```
nv-rliu, line 23: We got rid of `number_of_lines` when promoting the experimental datasets to stable. I'm not entirely sure if we would keep this for certain datasets. Why is it necessary?

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
BradReesWork: +1

acostadon: Done

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
acostadon: Didn't know it wasn't necessary, noticed that the tests verify that the metadata is correct...which is great.
I removed the lines item.

python/cugraph/cugraph/datasets/metadata/dining_prefs.yaml
acostadon: moved to the end.

acostadon:
rerun tests

BradReesWork (approved):

nv-rliu (approved):

PR #3869, opened 2023-09-21
Fix torch seed in `cugraph-dgl` and `-pyg` tests for conv layers

tingyu66:
Fixes https://github.com/rapidsai/graph_dl/issues/325

Recently, a few CI runs (ex. [1](https://github.com/rapidsai/cugraph/actions/runs/6254253684/job/16983164330?pr=3828#step:7:5078), [2](https://github.com/rapidsai/cugraph/actions/runs/6224345348/job/16896416094?pr=3843)) failed when comparing results from cugraph-ops-based conv layers against results from upstream frameworks. The tests pass most of the time, but occasionally fail due to a combination of using a strict tolerance and bad numerics (floating point error). This PR fixes the seed used for generating random feature tensors so that CI behaves consistently across different runs.


alexbarghi-nv (approved):
👍

PR #3874, opened 2023-09-23
WholeGraph Feature Store for cuGraph-PyG and cuGraph-DGL

alexbarghi-nv:
Created based on code from @dongxuy04 

Adds support for `WholeGraph` `WholeMemory` in the cuGraph `FeatureStore` class.  This enables both DGL and PyG to take advantage of distributed feature store functionality.

Adds `pylibwholegraph` as a testing dependency so the feature store can be tested.  Adds appropriate SG and MG tests.


raydouglass (changes requested):
`update-version.sh` needs to be updated for this new dependency. Probably just adding it to the list [here](https://github.com/rapidsai/cugraph/blob/branch-23.10/ci/release/update-version.sh#L81).

raydouglass (approved):

BradReesWork (approved):

python/cugraph/cugraph/gnn/feature_storage/feat_storage.py
```diff
@@ -104,6 +133,31 @@ def _cast_feat_obj_to_backend(feat_obj, backend: str):
                 return _cast_to_torch_tensor(feat_obj.values)
             else:
                 return _cast_to_torch_tensor(feat_obj)
+        elif backend == "wholegraph":
+            wg_comm_obj = kwargs.get("wg_comm", wgth.get_local_node_communicator())
+            wg_type_str = kwargs.get("wg_type", "distributed")
+            wg_location_str = kwargs.get("wg_location", "cuda")
```
VibhuJawa, line 139: I think we should  store these while initiating the FeatureStore object as I don't expect these to change b/w function calls.

VibhuJawa (changes requested):
Requested a minot change around API

python/cugraph/cugraph/gnn/feature_storage/feat_storage.py
alexbarghi-nv: @VibhuJawa is it better now?

VibhuJawa (approved):
LGTM

alexbarghi-nv:
This PR is going to be stuck until we have 23.10 pip packages for WholeGraph

PR #3875, opened 2023-09-25
Enable parallel mode

jnke2016:
This PR enables parallel mode

Closes https://github.com/rapidsai/graph_dl/issues/328


cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(dst);
+  auto p_weights =
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(weights);
+
+  /*
+  reinterpret_cast<cugraph::c_api::cugraph_induced_subgraph_result_t**>(result),
+  reinterpret_cast<cugraph::c_api::cugraph_error_t**>(error)
+  */
+
+  // FIXME: Support int64_t and float as well
```
ChuckHastings, line 139: You should be able to call `cugraph::c_api::vertex_dispatcher` to dispatch to a functor like the other algorithms.  You can pass `p_src->type_` as the vertex type, INT32 as the edge type (we don't do anything with edges here), `p_weights->type_` as the weight type (default to `FLOAT32` if no weight) and default store transposed and multi gpu to `FALSE` (also not used).

That would let you handle any arbitrary combination of types in a simple call.

You should probably add a `CAPI_EXPECTS` check to verify that src and dst are the same type.

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+    return CUGRAPH_UNKNOWN_ERROR;
+  }
+
+  return CUGRAPH_SUCCESS;
+}
+
+}  // namespace
+
+// template <typename vertex_t, typename weight_t>
+//  why taking out the 'extern "C"' worked.
```
ChuckHastings, line 117: Unsure of the context of this comment.

`extern "C"` is necessary for this to be callable from C code.  Without the `extern "C"` here, the C++ compiler will create the entry point for this function using C++ semantics, so it will mangle the name to include all of the type information.  A regular C compiler doesn't know how to do that.

The `template` construct can't be used on an `extern "C"` call, since templates are not a supported capability in C.

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+  cugraph::c_api::cugraph_type_erased_device_array_view_t const* src,
+  cugraph::c_api::cugraph_type_erased_device_array_view_t const* dst,
+  cugraph::c_api::cugraph_type_erased_device_array_view_t const* weights,
+  cugraph::c_api::cugraph_induced_subgraph_result_t** result,
+  cugraph::c_api::cugraph_error_t** error)
+{
+  return CUGRAPH_SUCCESS;
+}
+
+template <typename vertex_t, typename weight_t>
```
ChuckHastings, line 46: With the change proposed below, this can become a functor like the other C API implementations.  The templating can occur on the `()` operator.

cpp/include/cugraph_c/graph_functions.h
```diff
@@ -184,6 +184,14 @@ cugraph_error_code_t cugraph_extract_induced_subgraph(
   cugraph_induced_subgraph_result_t** result,
   cugraph_error_t** error);
 
+cugraph_error_code_t cugraph_replicate_edgelist(
```
ChuckHastings, line 187: Need a documentation block here.

We probably should add support for edge type and edge id as well, to be consistent with the rest of the graph creation API.  API changes in the C API are harder (there are no optional parameters, so deprecating a change is complex).

ChuckHastings: Also, we're technically doing an `allgather` in the implementation, so it might be worth calling this function `cugraph_allgather_edgelist`.  If the user were to call this with a distributed edge list then `allgather` is the right name.  We can describe in the documentation that a user can *broadcast* from a single node by passing in empty lists from every rank that doesn't have data.

cpp/src/c_api/capi_helper.cu
```diff
@@ -74,6 +75,28 @@ template void sort_by_key(raft::handle_t const& handle,
                           raft::device_span<int64_t> keys,
                           raft::device_span<int64_t> values);
 
+template <typename T>
```
ChuckHastings, line 78: I would suggest moving this to `src/detail`.  It has applicability outside of the C API, this is exposing the device comms methods to an `.cpp` file.

cpp/src/c_api/capi_helper.cu
```diff
@@ -74,6 +75,28 @@ template void sort_by_key(raft::handle_t const& handle,
...
+rmm::device_uvector<T> device_allgatherv(raft::handle_t const& handle,
+                                         raft::device_span<T const> d_input)
+{
+  auto& comm      = handle.get_comms();
```
ChuckHastings, line 82: I'd be inclined to either:
 1. Consolidate this to one line (`return cugraph::device_allgatherv(handle, handle.get_comms(), ...)`), or
 2. Add the comm as a parameter to match the called method.

Leaning toward the latter (match the function signatures match).  Thoughts @seunghwak ?

cpp/src/c_api/capi_helper.hpp
```diff
@@ -36,6 +36,10 @@ void sort_by_key(raft::handle_t const& handle,
                  raft::device_span<key_t> keys,
                  raft::device_span<value_t> values);
 
+template <typename T>
```
ChuckHastings, line 39: I would suggest moving this to `include/cugraph/detail`.  Perhaps a new file called `collect_comm_wrapper.hpp` within that directory to mirror the name of the functions we're wrapping.  We could ultimately wrap other methods from the `collect_comm.cuh` file as well.

cpp/src/c_api/capi_helper.cu
ChuckHastings: File would need a new name, perhaps `collect_comm_wrapper.cu`.

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+#include <c_api/utils.hpp>
+
+#include <cugraph/algorithms.hpp>
+#include <cugraph/detail/shuffle_wrappers.hpp>
+#include <cugraph/detail/utility_wrappers.hpp>
+#include <cugraph/graph_functions.hpp>
+
+namespace {
+
+template <typename vertex_t, typename weight_t>
```
ChuckHastings, line 34: I assume this is leftover from debugging and should be deleted.

ChuckHastings (commented):
Some C API comments to start with.

python/pylibcugraph/pylibcugraph/replicate_edgelist.pyx
```diff
@@ -0,0 +1,163 @@
...
+        c_resource_handle_ptr, subgraph_offsets_ptr)
+
+    # Free pointer
+    cugraph_induced_subgraph_result_free(result_ptr)
+
+    #print("Done doing the replication")
+    #print("src in plc = ", cupy_sources)
+    #print("dst in plc = ", cupy_destinations)
+    #print("weights in plc = ", cupy_edge_weights)
+
```
naimnv, line 161: Remove these lines

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
```diff
@@ -272,3 +274,108 @@ def test_pagerank_non_convergence(dask_client):
...
+            max_iter=max_iter,
+            tol=1e-6,
+            nstart=nstart,
+        )
+
+        result_pr = batch_pr.get_partition(i).compute().reset_index(drop=True)
+
+        assert_frame_equal(result_pr, expected_pr, check_dtype=False, check_like=True)
+
+        print("pr = \n", expected_pr)
```
naimnv, line 381: Should be removed.

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
```diff
@@ -272,3 +274,108 @@ def test_pagerank_non_convergence(dask_client):
...
+        personalization=personalization_ddf,
+        precomputed_vertex_out_weight=pre_vtx_o_wgt,
+        max_iter=max_iter,
+        tol=1e-6,
+        nstart=nstart,
+    )
+
+    for i in range(batch_pr.npartitions):
+
+        # print("batch_pr = \n", batch_pr.get_partition(i).compute())
```
naimnv, line 363: Need to remove this line

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
```diff
@@ -272,3 +274,108 @@ def test_pagerank_non_convergence(dask_client):
...
+        pre_vtx_o_wgt = (
+            df.groupby(["src"], as_index=False)
+            .sum()
+            .rename(columns={"src": "vertex", "value": "sums"})
+        )
+
+    if has_guess == 1:
+        nstart = cugraph.pagerank(g, personalization=personalization, tol=1e-6).rename(
+            columns={"pagerank": "values"}
+        )
```
naimnv, line 344: Can we use as much global variable for column names and numpy const for data types?

python/cugraph/cugraph/structure/replicate_edgelist.py
```diff
@@ -0,0 +1,110 @@
...
+                edata,
+                workers=[w],
+                allow_other_workers=False,
+                pure=False,
+            )
+            for w, edata in edgelist_ddf.items()
+        ]
+        # print("result before = \n", result)
+        # wait(result)
+        # print("result after = \n", result)
```
naimnv, line 98: Need to remove these lines

python/cugraph/cugraph/structure/replicate_edgelist.py
```diff
@@ -0,0 +1,110 @@
...
+        gathered_edgelist_df["dst"] = dst
+        gathered_edgelist_df["wgt"] = wgt
+        # print("the gathered edgelist = \n", gathered_edgelist_df)
+
+        return gathered_edgelist_df
+
+    def _call_plc_replicate_edgelist(
+        sID: bytes, edgelist_df: cudf.DataFrame
+    ) -> cudf.Series:
+        # print("edgelist_df = \n", edgelist_df)
```
naimnv, line 69: Need to remove

python/cugraph/cugraph/structure/replicate_edgelist.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        Creates a cudf Dataframe from cupy arrays
+        """
+        # vertices = cudf.Series(cp_arrays)
+        src, dst, wgt, _ = cp_arrays
+        gathered_edgelist_df = cudf.DataFrame()
+        gathered_edgelist_df["src"] = src
+        gathered_edgelist_df["dst"] = dst
+        gathered_edgelist_df["wgt"] = wgt
+        # print("the gathered edgelist = \n", gathered_edgelist_df)
```
naimnv, line 62: these lines

python/cugraph/cugraph/structure/replicate_edgelist.py
```diff
@@ -0,0 +1,110 @@
...
+    _client = default_client()
+
+    edgelist_ddf = persist_dask_df_equal_parts_per_worker(edgelist_ddf, _client)
+    edgelist_ddf = get_persisted_df_worker_map(edgelist_ddf, _client)
+
+    def convert_to_cudf(cp_arrays: cp.ndarray) -> cudf.DataFrame:
+        """
+        Creates a cudf Dataframe from cupy arrays
+        """
+        # vertices = cudf.Series(cp_arrays)
```
naimnv, line 56: Need to remove

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
+            )
+
+        """
         if self.batch_enabled:
             self._replicate_edgelist()
-
-        self._make_plc_graph(
-            value_col=value_col, store_transposed=store_transposed, renumber=renumber
-        )
+        print("making the plc graph")
```
naimnv, line 370: Need to remove

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
+            }
+            self._plc_graph = {
+                w: client.compute(delayed_task, workers=w, allow_other_workers=False)
+                for w, delayed_task in delayed_tasks_d.items()
+            }
+            wait(list(self._plc_graph.values()))
+            del delayed_tasks_d
+            client.run(gc.collect)
+            print("Done computing the plc graph")
+            print("the plc graph is = \n", self._plc_graph)
```
naimnv, line 357: Need to remove

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
+            # FIXME: Making another copy here which is expensive. Better to update
+            # 'self.edgelist'
+            # and if the user querry the edges, return only one partition?
+            print("the edgelist type = ", type(self.edgelist))
+            if isinstance(self.edgelist.edgelist_df, cudf.DataFrame):
+                input_df = dask_cudf.from_cudf(
+                    self.edgelist.edgelist_df, npartitions=len(Comms.get_workers())
+                )
+            for i in range(input_df.npartitions):
+                print("before ", input_df.get_partition(i).compute())
```
naimnv, line 327: Need to remove?

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
+            input_df = replicate_edgelist(input_df)
+
+            for i in range(input_df.npartitions):
+                print("after ", input_df.get_partition(i).compute())
```
naimnv, line 331: Need to remove?

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
+
+            print("input_df = \n", input_df)
+            # Create a dask_cudf dataframe
```
naimnv, line 334: Need to remove

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -294,12 +310,65 @@ def __from_edgelist(
...
         self.edgelist = simpleGraphImpl.EdgeList(source_col, dest_col, value_col)
 
+        if self.properties.parallel:
+            # FIXME: Ensure a dask cluster was created
+            # FIXME: If the user call edgelist, compute the result of a single
+            # partition?
+            # len(Comms.get_workers())
+            # FIXME: Making another copy here which is expensive. Better to update
+            # 'self.edgelist'
+            # and if the user querry the edges, return only one partition?
```
naimnv, line 320: Reformat this comment block?

python/cugraph/cugraph/link_analysis/pagerank.py
```diff
@@ -248,31 +304,105 @@ def pagerank(
...
+                    )
+                else:
+                    # FIXME: Assumed to be a dask_cudf instead. But ensure that
+                    personalization_ddf = personalization
+
+                data_prsztn = get_distributed_data(personalization_ddf)
+
+                client = default_client()
+                # print("nstart_vertices = ", initial_guess_vertices)
+                # print("nstart_values = ", initial_guess_values)
```
naimnv, line 346: Need to remove

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+  /*
+  return dummy_function<int32_t, float>(
+   *p_handle->handle_,
+   p_src,
+   p_dst,
+   p_weights,
+   reinterpret_cast<cugraph::c_api::cugraph_induced_subgraph_result_t**>(result),
+   reinterpret_cast<cugraph::c_api::cugraph_error_t**>(error));
+   */
```
naimnv, line 156: Need to remove

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -0,0 +1,157 @@
...
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(src);
+  auto p_dst =
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(dst);
+  auto p_weights =
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(weights);
+
+  /*
+  reinterpret_cast<cugraph::c_api::cugraph_induced_subgraph_result_t**>(result),
+  reinterpret_cast<cugraph::c_api::cugraph_error_t**>(error)
+  */
```
naimnv, line 137: Need to remove

cpp/src/c_api/replicate_edgelist.cpp
jnke2016: > Unsure of the context of this comment.
I was annotating the code with comments (for myself) when I had questions and learning on the process. I figured out "C" doesn't have template and removing the extern "C" will not have C linkage. I forgot to remove this comment  once I figured that out

cpp/src/c_api/replicate_edgelist.cpp
jnke2016: Right. I had it hardcoded to certain type just to be able to move on. That's why I added a FIXME and I eventually came back to fix it with the `vertex_dispatcher`

> INT32 as the edge type (we don't do anything with edges here)

Actually this should not be hardcoded to INT32 because if we do so, we won't be able to support INT64 `src` and `dst`.

cpp/src/c_api/replicate_edgelist.cpp
jnke2016: Right. I created a functor instead that matches other C API implementations

cpp/include/cugraph_c/graph_functions.h
jnke2016: I added documentation and updated the function name to `cugraph_allgather_edgelist`. I also added support for edge_ids and edge_type_ids however, I think we should rename `cugraph_induced_subgraph_result_t` later because it is quite odd to have it as a result type when doing an `allgatherv`.

cpp/src/c_api/capi_helper.cu
jnke2016: I moved this functionality to `src/detail` per your recommendation.

cpp/src/c_api/capi_helper.cu
jnke2016: Got it I chose the the latter.

cpp/src/c_api/capi_helper.hpp
jnke2016: resolved

cpp/src/c_api/replicate_edgelist.cpp
jnke2016: resolved

python/pylibcugraph/pylibcugraph/replicate_edgelist.pyx
```diff
@@ -0,0 +1,163 @@
...
+            of the ith edge of the graph.
+
+        Returns
+        -------
+        return a cupy arrays of 'first' and 'second' or a 'cugraph_vertex_pairs_t'
+        which can be directly passed to the similarity algorithm?
+    """
+    #print("src array before all gather= ", src_array.to_cupy())
+    #print("dst array before all gather= ", dst_array.to_cupy())
+    #print("wgt array before all gather= ", weight_array.to_cupy())
```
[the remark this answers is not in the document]
jnke2016: resolved.

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
jnke2016: resolved as I reverted the changes I made to this file.

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
jnke2016: resolved as I reverted the changes I made to this file.

python/cugraph/cugraph/tests/link_analysis/test_pagerank_mg.py
jnke2016: resolved as I reverted the changes I made to this file.

python/cugraph/cugraph/structure/replicate_edgelist.py
jnke2016: resolved

python/cugraph/cugraph/structure/replicate_edgelist.py
jnke2016: resolved

python/cugraph/cugraph/structure/replicate_edgelist.py
jnke2016: resolved

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
jnke2016: resolved

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -18,111 +18,167 @@
...
-        new cugraph::c_api::cugraph_type_erased_device_array_t(edge_offsets,
-                                                               cugraph_data_type_id_t::SIZE_T)};
-    }
+    cugraph::c_api::cugraph_induced_subgraph_result_t* result = NULL;
 
-  } catch (std::exception const& ex) {
-    *error = new cugraph::c_api::cugraph_error_t{ex.what()};
-    return CUGRAPH_UNKNOWN_ERROR;
-  }
+    std::optional<rmm::device_uvector<weight_t>> gathered_weights =
```
ChuckHastings, line 124: If you overwrite this can become something like:

```
if (edgelist_weights) {
   edgelist_weights = cugraph::detail::device_allgatherv(handle_, comm, raft::device_span<weight_t const>(edgelist_weights->data(), edgelist_weights->size()));
}
```

Which is a bit easier to read.

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -18,111 +18,167 @@
...
   const cugraph_type_erased_device_array_view_t* weights,
-  cugraph_induced_subgraph_result_t** result,
+  const cugraph_type_erased_device_array_view_t* edge_ids,
+  const cugraph_type_erased_device_array_view_t* edge_type_ids,
+  cugraph_induced_subgraph_result_t** edgelist,
   cugraph_error_t** error)
 {
+  *edgelist = nullptr;
+  *error    = nullptr;
+  constexpr size_t int32_threshold{std::numeric_limits<int32_t>::max()};
```
ChuckHastings, line 180: I don't think we need this (more below in the logic that uses it)

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -131,27 +187,84 @@ extern "C" cugraph_error_code_t cugraph_replicate_edgelist(
...
+
+  CAPI_EXPECTS((weights == nullptr) || (p_weights->size_ == p_src->size_),
+               CUGRAPH_INVALID_INPUT,
+               "Invalid input arguments: src size != weights size.",
+               *error);
+
+  cugraph_data_type_id_t edge_type;
+  cugraph_data_type_id_t weight_type;
+
+  if (p_src->size_ < int32_threshold) {
```
ChuckHastings, line 213: This is copied from the graph construction (I may have suggested looking there).  But we don't actually create anything of type `edge_type` within this code.

I would suggest the following:
* If `edge_ids` is specified, use p_edge_ids->type_ as the edge type
* Otherwise just use `cugraph_data_type_id_t::INT32`

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -131,27 +187,84 @@ extern "C" cugraph_error_code_t cugraph_replicate_edgelist(
...
+  if (weights != nullptr) {
+    weight_type = p_weights->type_;
+  } else {
+    weight_type = cugraph_data_type_id_t::FLOAT32;
+  }
+
+  cugraph_data_type_id_t edge_type_id_type = cugraph_data_type_id_t::INT32;
+  if (edge_type_ids != nullptr) { edge_type_id_type = p_edge_type_ids->type_; }
+
+  CAPI_EXPECTS((edge_ids == nullptr) || (p_edge_ids->type_ == edge_type),
```
ChuckHastings, line 228: If you make the above recommended change this is unnecessary.

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -18,111 +18,167 @@
...
+    raft::copy<vertex_t>(
+      edgelist_srcs.data(), src_->as_type<vertex_t>(), src_->size_, handle_.get_stream());
+    raft::copy<vertex_t>(
+      edgelist_dsts.data(), dst_->as_type<vertex_t>(), dst_->size_, handle_.get_stream());
 
-  if (edgelist_weights) {
-    raft::copy<weight_t>(
-      edgelist_weights->data(), weights->as_type<weight_t>(), weights->size_, handle.get_stream());
-  }
+    std::optional<rmm::device_uvector<weight_t>> edgelist_weights =
```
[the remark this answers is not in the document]
jnke2016: Right. You are reordering the instructions and this makes it cleaner

cpp/src/c_api/replicate_edgelist.cpp
```diff
@@ -18,111 +18,167 @@
...
+            bool multi_gpu>
+  void operator()()
+  {
+    rmm::device_uvector<vertex_t> edgelist_srcs(src_->size_, handle_.get_stream());
+    rmm::device_uvector<vertex_t> edgelist_dsts(dst_->size_, handle_.get_stream());
 
-  std::optional<rmm::device_uvector<weight_t>> edgelist_weights =
-    weights ? std::make_optional(rmm::device_uvector<weight_t>(weights->size_, handle.get_stream()))
-            : std::nullopt;
+    raft::copy<vertex_t>(
```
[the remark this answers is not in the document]
jnke2016: I believe it is also unnecessary when copying vectors of `edge_t`, `weight_t` and `edge_type_id_t`?

cpp/src/c_api/replicate_edgelist.cpp
jnke2016: Isn't this code doing what you recommended above but for `edge_type_ids`?

cpp/include/cugraph_c/graph_functions.h
ChuckHastings: Agree this needs a new name.  I don't want to create different objects, that would be overkill.  Finding a better name that can be used by this logic, the induced subgraph logic, egonet, ktruss and anything else returning a list of edges would be ideal.  I'm inclined to do that in a separate PR as it affects other algorithms and is orthogonal to this.  Perhaps add a FIXME for now.

jnke2016: I added already added a `FIXME` in L205

cpp/src/c_api/replicate_edgelist.cpp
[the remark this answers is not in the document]
ChuckHastings: Yes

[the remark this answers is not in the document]
ChuckHastings: There's no harm in specifying, but leaving it out generally makes the code easier to read... assuming the compiler can correctly resolve it.

cpp/src/c_api/replicate_edgelist.cpp
ChuckHastings: I think what you've created is correct.  This comment is kind of obsolete at this point.  This was targeted at the error check `p_edge_ids->type_ == edge_type`, which is obsolete... but not in your code anymore.

ChuckHastings (approved):

python/cugraph/cugraph/structure/replicate_edgelist.py
```diff
@@ -0,0 +1,150 @@
...
+    get_persisted_df_worker_map,
+    persist_dask_df_equal_parts_per_worker,
+)
+
+import dask
+import cupy as cp
+import cugraph.dask.comms.comms as Comms
+from typing import Union, Tuple
+
+edgeWeightCol = "weights"
```
BradReesWork, line 32: Is there a global structure that captures the standard column names so that each files does not need to define “src” each time and we have no conflict with “weight” vs “weights”?

BradReesWork (changes requested):

cpp/include/cugraph/detail/collect_comm_wrapper.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#include <raft/core/handle.hpp>
+#include <rmm/device_uvector.hpp>
+
+#include <optional>
+
+namespace cugraph {
+namespace detail {
+/*
+Add some description here
```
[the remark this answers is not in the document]
jnke2016: This was resolved

python/cugraph/cugraph/structure/replicate_edgelist.py
jnke2016: I just pushed a commit that removed the hardcoded column names and allows the user to provide any column names.

ChuckHastings (approved):

naimnv (approved):

python/cugraph/cugraph/__init__.py
```diff
@@ -53,6 +53,9 @@
     is_multigraph,
     is_bipartite,
     is_multipartite,
+    replicate_edgelist,
+    replicate_cudf_dataframe,
+    replicate_cudf_series,
```
rlratzel, line 58: If these aren't intended to be part of the user-visible API then we shouldn't put them in this file.  ( didn't think they were supposed to be user-visible, but now I'm not sure)

python/cugraph/cugraph/structure/replication.py
```diff
@@ -0,0 +1,350 @@
...
+    persist_dask_df_equal_parts_per_worker,
+)
+
+import dask
+import cupy as cp
+import cugraph.dask.comms.comms as Comms
+from typing import Union, Tuple
+
+
+def convert_to_cudf(cp_arrays: Tuple[cp.ndarray], col_names: list) -> cudf.DataFrame:
```
rlratzel, line 34: If this is only used in this file then it might be better to name it `_convert_to_cudf`, since otherwise it seems like it's supposed to be a general-purpose utility, and maybe we'd put it in a common place for utils.

python/cugraph/cugraph/tests/internals/test_replicate_edgelist_mg.py
```diff
@@ -0,0 +1,126 @@
...
+            (
+                "graph_file",
+                "use_weights",
+                "use_edge_ids",
+                "use_edge_type_ids",
+                "distributed",
+            ),
+            request.param,
+        )
+    )
```
rlratzel, line 76: This pattern isn't as useful here since there's only one test.  It might be better for future maintainers if the test in this file just used regular pytest parameters.

python/pylibcugraph/pylibcugraph/replication.pyx
```diff
@@ -0,0 +1,210 @@
...
+)
+from pylibcugraph.utils cimport (
+    assert_success,
+    assert_CAI_type,
+    copy_to_cupy_array,
+    create_cugraph_type_erased_device_array_view_from_py_obj
+)
+
+
+def replication(ResourceHandle resource_handle,
```
rlratzel, line 51: I think a better name would be `replicate_edgelist`, same for the file itself.

python/pylibcugraph/pylibcugraph/replication.pyx
```diff
@@ -0,0 +1,210 @@
...
+
+        Returns
+        -------
+        return a cupy arrays of 'src' and 'dst', 'weight', 'edge_id' and 'edge_type_id'
+        (if the were passed)
+    """
+
+
+    #print("src array = \n", src_array)
+    print("wgt array = \n", weight_array)
```
rlratzel, line 105: Can you remove this and all other debug prints?

python/pylibcugraph/pylibcugraph/replication.pyx
```diff
@@ -0,0 +1,210 @@
...
+
+def replication(ResourceHandle resource_handle,
+                src_array,
+                dst_array,
+                weight_array,
+                edge_id_array,
+                edge_type_id_array):
+    """
+        Compute vertex pairs that are two hops apart. The resulting pairs are
+        sorted before returning.
```
rlratzel, line 59: Looks like a copy/paste error?

python/pylibcugraph/pylibcugraph/replication.pyx
```diff
@@ -0,0 +1,210 @@
...
+        edge_type_id_array : device array type, optional
+            Device array containing the edge type id values of each directed edge. The
+            order of the array corresponds to the ordering of the src_array and
+            dst_array arrays, where the ith item in edge_type_id_array is the type id
+            value of the ith edge of the graph.
+
+        Returns
+        -------
+        return a cupy arrays of 'src' and 'dst', 'weight', 'edge_id' and 'edge_type_id'
+        (if the were passed)
```
rlratzel, line 100: There's some typos in here.

rlratzel (changes requested):
Thanks, Joseph, I'm looking forward to having this capability.  I found a few issues below.

python/cugraph/cugraph/structure/replication.py
```diff
@@ -0,0 +1,350 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import dask_cudf
+import cudf
+from dask.distributed import wait, default_client
+import numpy as np
```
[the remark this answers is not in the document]
jnke2016: Ya. Few algos like `sssp` and `graph_sage_mg`. I think we should add it in `dependencies.yaml` like you suggested

python/cugraph/cugraph/__init__.py
jnke2016: I will remove it from the user-visible API for now. If in the future we change our mind we can't always bring it back

python/cugraph/cugraph/structure/replication.py
jnke2016: Right! several MG algos use this pattern and it should definitely be a general-purpose function. I used the `_` for now and added a FIXME as this will require refactoring a lot of our algos

python/pylibcugraph/pylibcugraph/replication.pyx
jnke2016: Done

python/pylibcugraph/pylibcugraph/replication.pyx
jnke2016: Done.

python/pylibcugraph/pylibcugraph/replication.pyx
jnke2016: Done

python/pylibcugraph/pylibcugraph/replication.pyx
jnke2016: Fixed

python/cugraph/cugraph/structure/replication.py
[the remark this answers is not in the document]
jnke2016: Actually I get a failure when adding `numpy` to `dependencies.yaml` in `python_run_cugraph:`

jnke2016:
rerun tests

BradReesWork (approved):

python/cugraph/cugraph/tests/internals/test_replicate_edgelist_mg.py
```diff
@@ -0,0 +1,129 @@
...
+
+edgeWeightCol = "weights"
+edgeIdCol = "edge_id"
+edgeTypeCol = "edge_type"
+srcCol = "src"
+dstCol = "dst"
+
+
+input_data = UNDIRECTED_DATASETS
+input_data.append(karate_disjoint)
```
rlratzel, line 42: You're modifying the `UNDIRECTED_DATASETS` list for every test in the pytest session here.  I think you really want to do this?
```suggestion
input_data = UNDIRECTED_DATASETS + [karate_disjoint]
```

rlratzel (changes requested):
I think I found out why CI is failing for this PR.

rlratzel (approved):

AyodeAwe (approved):

PR #3828, opened 2023-08-26
Refactor python code for similarity algos to use latest CAPI

naimnv:
This PR
-  refactors python code for similarity algorithms (Jaccard, Sorensen, Overlap) to use latest CAPI
-  removes legacy cuda c/c++ code and python wrapper around legacy code
-  update CAPI tests
-  remove and update python tests

Closes #2546 
Closes #2547 
Closes #2548
Closes #2549 
Closes #2749

PR #3879, opened 2023-09-26
Enable weights for MG similarity algorithms 

jnke2016:
This is a follow up PR to #3828 which enabled weighted for the python SG similarity algorithms. 
This PR also updates the tests, docstrings and remove experimental calls


python/cugraph/cugraph/tests/link_prediction/test_overlap_mg.py
```diff
@@ -72,7 +76,10 @@ def input_expected_output(input_combo):
     input_data_path = input_combo["graph_file"]
     directed = input_combo["directed"]
     has_vertex_pair = input_combo["has_vertex_pair"]
-    G = utils.generate_cugraph_graph_from_file(input_data_path, directed=directed)
+    is_weighted = input_combo["is_weighted"]
```
naimnv, line 79: I wonder, wouldn't it be cleaner to have these column names in global variable declared at the beginning of the files?

alexbarghi-nv (approved):
👍

PR #3886, opened 2023-09-26
Adds logic to handle isolated vertices at python layer

### LINKED ISSUE #3804: [BUG]: `cugraph.louvain` does not handle empty graphs well
### Version

22.10 (dev)

Source

### Describe the bug.

`cugraph.louvain` is very, very slow for empty input graphs. I would expect this condition to be detected and handled quickly.

I handled this in `cugraph-nx` here: https://github.com/rapidsai/cugraph/pull/3803/files#diff-506c93446829d8794348fa8b6773e7326ede87b095ac120c9c47fed46e56c699R31-R33

```python
import sys
import networkx as nx
import cugraph

G = nx.Graph()
G.add_nodes_from(range(5))
# G = cugraph.utilities.ensure_cugraph_obj_for_nx(G)  # Optional
cugraph.louvain(G, sys.maxsize)  # Runs for a very long time
```

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

### Code of Conduct

naimnv:
- Adds logic to handle isolated vertices
- Clamp downs max level to a sane number

closes #3804

### DIFF
--- a/python/cugraph/cugraph/community/louvain.py
+++ b/python/cugraph/cugraph/community/louvain.py
@@ -150,7 +150,8 @@ def louvain(
         max_level = 100
 
     if max_level > 500:
-        warnings.warn("clamping down max_level to 500.")
+        w_msg = "max_level is set too high, clamping it down to 500."
+        warnings.warn(w_msg)
         max_level = 500
 
     vertex, partition, modularity_score = pylibcugraph_louvain(
@@ -179,7 +180,7 @@ def louvain(
             [result, isolated_vtx_and_cids], ignore_index=True, sort=False
         )
 
-    if G.renumbered and G.number_of_nodes() > len(isolated_vertices):
+    if G.renumbered and len(G.input_df) > 0:
         result = G.unrenumber(result, VERTEX_COL_NAME)
 
     if isNx is True:
--- a/python/cugraph/cugraph/community/louvain.py
+++ b/python/cugraph/cugraph/community/louvain.py
@@ -76,9 +76,9 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
         GPU data frame of size V containing two columns the vertex id and the
         partition id it is assigned to.
 
-        result_df['vertex'] : cudf.Series
+        result_df[VERTEX_COL_NAME] : cudf.Series
             Contains the vertex identifiers
-        result_df['partition'] : cudf.Series
+        result_df[CLUSTER_ID_COL_NAME] : cudf.Series
             Contains the partition assigned to the vertices
 
     modularity_score : float
@@ -92,6 +92,10 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
     >>> parts = cugraph.louvain(G)
 
     """
+
+    # FIXME: Onece the graph construction calls support isolated vertices through
+    #  the C API (the C++ interface already supports this) then there will be
+    # no need to compute isolated vertices here.
 
     isolated_vertices = list()
     if is_nx_graph_type(type(G)):
--- a/python/cugraph/cugraph/community/louvain.py
+++ b/python/cugraph/cugraph/community/louvain.py
@@ -101,6 +101,9 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
     isolated_vertices = list()
     if is_nx_graph_type(type(G)):
         isolated_vertices = list(nx.isolates(G))
+    else:
+        # FIXME: Gather isolated vertices of G
+        pass
 
     G, isNx = ensure_cugraph_obj_for_nx(G)
 
--- a/python/cugraph/cugraph/community/louvain.py
+++ b/python/cugraph/cugraph/community/louvain.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 
 from cugraph.utilities import (
+    is_nx_graph_type,
     ensure_cugraph_obj_for_nx,
     df_score_to_dictionary,
 )
@@ -20,6 +21,9 @@ import cudf
 import warnings
 from pylibcugraph import louvain as pylibcugraph_louvain
 from pylibcugraph import ResourceHandle
+
+VERTEX_COL_NAME = "vertex"
+CLUSTER_ID_COL_NAME = "partition"
 
 
 # FIXME: max_level should default to 100 once max_iter is removed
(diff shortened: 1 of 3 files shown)


python/cugraph/cugraph/community/louvain.py
```diff
@@ -89,6 +93,14 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
 
     """
 
+    isolated_vertices = list()
```
ChuckHastings, line 96: Perhaps a FIXME indicating this is a temporary solution.

If we add a mechanism in the graph construction calls to support isolated vertices through the C API (the C++ interface already supports this) then there will be no need to compute isolated vertices here.

python/cugraph/cugraph/community/louvain.py
```diff
@@ -72,9 +76,9 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
         GPU data frame of size V containing two columns the vertex id and the
         partition id it is assigned to.
```
BradReesWork, line 78: You should use the defined column names below

python/cugraph/cugraph/community/louvain.py
```diff
@@ -89,6 +93,14 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
 
     """
 
+    isolated_vertices = list()
+    if is_nx_graph_type(type(G)):
+        num_of_vertices = len(G.nodes)
+        isolated_vertices = [v for v in range(num_of_vertices) if G.degree[v] == 0]
```
BradReesWork, line 99: this seems like a slow process vs putting this code later

G, isNx = ensure_cugraph_obj_for_nx(Gin)

if (isNx)
   if (G.number_of_nodes != Gin.number_of_nodes())
   then do the slow look over every vertex degree

BradReesWork (changes requested):

python/cugraph/cugraph/community/louvain.py
naimnv: Added

python/cugraph/cugraph/community/louvain.py
naimnv: Then we would be keeping two copies of the graph, G and Gin

We might need to handle isolated vertices for cugraph Graph as well (line 105)

Though we hope we would add support at C layers to handle isolated vertices before 23.12

naimnv: And `ensure_cugraph_obj_for_nx` only checks type, even if we have G and Gin,

G.number_of_nodes() and Gin.number_of_nodes() would always be the same number.

python/cugraph/cugraph/community/louvain.py
naimnv: Changed to use defined column names

python/cugraph/cugraph/community/louvain.py
jnke2016: > We might need to handle isolated vertices for cugraph Graph as well (line 104), that's more expensive though.

I don't think we will handle isolated vertices this way once we start supporting them. Ideally we should be able to just create a graph by specifying isolated vertices at the graph creation and shouldn't have to rely on this trick. I think the
```
else:
    pass
```
is unnecessary here because there is no feature allowing a user to pass isolated vertices when using cuGraph only API

python/cugraph/cugraph/community/louvain.py
```diff
@@ -89,6 +93,17 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
...
 
+    # FIXME: Onece the graph construction calls support isolated vertices through
+    #  the C API (the C++ interface already supports this) then there will be
+    # no need to compute isolated vertices here.
+
+    isolated_vertices = list()
+    if is_nx_graph_type(type(G)):
+        isolated_vertices = [v for v in range(G.number_of_nodes()) if G.degree[v] == 0]
+    else:
+        # FIXME: Gather list of isolated vertices of G (of cugraph.Graph type)
```
jnke2016, line 104: I don't understand this `FIXME`. THere is currently only one way to pass isolated vertices to our API and it is only through networkx so I don't see why the else statement is necessary here. Later once we start supporting isolated vertices it wouldn't be through this else statement. Unless I am missing something

jnke2016: The point I am trying to make is that when we start supporting isolated vertices with a cuGraph Graph, I don't think it will be through this `else `statement therefore I find it unnecessary.

jnke2016: >There is currently only one way to pass isolated vertices to our API

Actually a user can pass isolated vertices to our API by creating a cuGraph Graph from an adjacency list. In this case the `else` statement will be useful if you decide to add support for that particular case in this PR.

python/cugraph/cugraph/community/louvain.py
```diff
@@ -89,6 +93,17 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
 
     """
 
+    # FIXME: Onece the graph construction calls support isolated vertices through
+    #  the C API (the C++ interface already supports this) then there will be
+    # no need to compute isolated vertices here.
+
+    isolated_vertices = list()
+    if is_nx_graph_type(type(G)):
+        isolated_vertices = [v for v in range(G.number_of_nodes()) if G.degree[v] == 0]
```
jnke2016, line 102: Like Brad probably mentioned. This may hurt our performance any time a user passes a networkX graph and this check will always be True whether we have isolated vertices or not. But it is ok for now since we don't have any alternative.

python/cugraph/cugraph/community/louvain.py
naimnv: @jnke2016 Handling isolated vertices at python layer is a temporary fix.
C API will be changed soon to support isolated vertices, as we already have support at C++ layer. I'm planning to make a separate PR for that. 
These are all temporary for now.

jnke2016 (approved):

BradReesWork (approved):

ChuckHastings (approved):

rlratzel (changes requested):
Thanks! One request: can you add a test that has a small inline graph with at least one isolated vertex to verify that these changes work?

python/cugraph/cugraph/community/louvain.py
```diff
@@ -112,6 +127,9 @@ def louvain(G, max_level=None, max_iter=None, resolution=1.0, threshold=1e-7):
     if max_level is None:
         max_level = 100
 
+    if max_level > 1000:
+        max_level = 1000
```
rlratzel, line 131: You need to at least update the docstring to say that you're changing the user's option.  A better change would be to *also* omit a warning, but the docstring must be updated at a minimum.

rlratzel (changes requested):

python/cugraph/cugraph/community/louvain.py
naimnv: Updated

PR #3887, opened 2023-09-26
Integrate renumbering and compression to `cugraph-dgl` to accelerate MFG creation

tingyu66:
Allow cugraph-dgl dataloader to consume sampled outputs from BulkSampler in CSC format.


seunghwak:
I think all the C/C++ updates in this PR is from #3841, this won't require C/C++ review once #3841 gets merged to the main branch.

seunghwak (commented):
This has no C/C++ updates and approving without review.

tingyu66:
Using `ogbn-products` dataset, the time spent on MFG creation (__excluding__ IO) for each epoch is reduced from 4.34 s to 2.95 s. The script used for the comparison is [here](https://gist.github.com/tingyu66/9ff32d5146b475f119c3309fed4e81f7). CC: @VibhuJawa 

#### COO
image


#### CSC
image

python/cugraph-dgl/cugraph_dgl/dataloading/dataloader.py
```diff
@@ -54,12 +54,13 @@ def __init__(
         batch_size: int = 1024,
         drop_last: bool = False,
```
VibhuJawa, line 55: @tingyu66 , Can we also change `seeds_per_call` to `100_000` to make a better default based on your testing ?

python/cugraph-dgl/cugraph_dgl/dataloading/utils/sampling_helpers.py
```diff
@@ -401,3 +403,154 @@ def create_heterogenous_dgl_block_from_tensors_dict(
...
+        List of the number of nodes in each message passing layer. For the
+        k-th hop, mfg_sizes[k] and mfg_sizes[k+1] is the number of sources and
+        destinations, respectively.
+    """
+    # dropna
+    major_offsets = df.major_offsets.dropna().values
+    label_hop_offsets = df.label_hop_offsets.dropna().values
+    renumber_map_offsets = df.renumber_map_offsets.dropna().values
+    renumber_map = df.map.dropna().values
+    minors = df.minors.dropna().values
```
VibhuJawa, line 451: I think this assumes that the length of the renumber_map is smaller than major_offsets. I will check this again if possible.

python/cugraph-dgl/cugraph_dgl/dataloading/dataloader.py
tingyu66: Should we change it to the default value of BulkSampler: `200_000`? After our call the other day, I tested a wide range of `seeds_per_call` values and none of the runs threw a OOM error.

VibhuJawa: Interesting,  if it just works we can probably set the default  to None and let upstream handle it  ? What do you think, any default which is reasonable and just works is fine by me.

python/cugraph-dgl/cugraph_dgl/dataloading/utils/sampling_helpers.py
tingyu66: I don't think we are making any assumptions here. `renumber_map` and `major_offsets` are simply two different tensors that happen to be stored in a single df.

alexbarghi-nv (approved):
👍

python/cugraph-dgl/cugraph_dgl/dataloading/utils/sampling_helpers.py
VibhuJawa: Ahh, My bad I just re-read the code below and I think we should be fine.

tingyu66: For each batch:
- The length of renumber_map = number of distinct nodes (i.e., the number of sources in hop 0)
- The length of major_offsets = sum of number of destination nodes in each hop

VibhuJawa:
>Using ogbn-products dataset, the time spent on MFG creation (excluding IO) for each epoch is reduced from 4.34 s to 2.95 s. The script used for the comparison is [here](https://gist.github.com/tingyu66/9ff32d5146b475f119c3309fed4e81f7). CC: @VibhuJawa

Curious: What is the batch size being tested here ,  @tingyu66  ?

tingyu66:
> > Using ogbn-products dataset, the time spent on MFG creation (excluding IO) for each epoch is reduced from 4.34 s to 2.95 s. The script used for the comparison is [here](https://gist.github.com/tingyu66/9ff32d5146b475f119c3309fed4e81f7). CC: @VibhuJawa
> 
> Curious: What is the batch size being tested here , @tingyu66 ?

batch_size: 128, batches_per_partition: 50.

python/cugraph-dgl/cugraph_dgl/dataloading/dataloader.py
tingyu66: I updated the value to `200_000` in [45f93f2](https://github.com/rapidsai/cugraph/pull/3887/commits/45f93f27611721d3fbc3fa8d845958c71500350a) to align with BulkSampler. Did not set to None to avoid the extra step of handling None case.

VibhuJawa (approved):
LGTM

PR #3894, opened 2023-09-27
Some MTMG code cleanup and small optimizations

ChuckHastings:
Added some missing documentation.

A couple of optimizations:
 * Modified the `append` logic to keep the mutex lock only long enough to compute what needs to be copied and where.
 * Modified the handle created by the resource manager for each GPU to have a stream pool to enable different threads to operate on different streams.


cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -110,45 +125,67 @@ class per_device_edgelist_t {
...
-    //   would require more careful coding.  The raft::update_device
-    //   calls could be done without the lock if we made a local
-    //   of the values of *.back() and did an increment of current_pos_
-    //   while we hold the lock.
-    std::lock_guard<std::mutex> lock(lock_);
+    std::vector<std::tuple<vertex_t*, vertex_t const*, size_t>> src_copies;
+    std::vector<std::tuple<vertex_t*, vertex_t const*, size_t>> dst_copies;
+    std::vector<std::tuple<weight_t*, weight_t const*, size_t>> wgt_copies;
+    std::vector<std::tuple<edge_t*, edge_t const*, size_t>> edge_id_copies;
+    std::vector<std::tuple<edge_type_t*, edge_type_t const*, size_t>> edge_type_copies;
```
seunghwak, line 132: Should we maintain 5 variables or one variable storing (input_start_offset, output_start_offset, size) triplets will be sufficient?

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -110,45 +125,67 @@ class per_device_edgelist_t {
...
+          edge_id_copies.push_back(std::make_tuple(
+            edge_id_->back().begin() + current_pos_, edge_id->begin() + pos, copy_count));
+        if (edge_type)
+          edge_type_copies.push_back(std::make_tuple(
+            edge_type_->back().begin() + current_pos_, edge_type->begin() + pos, copy_count));
+
+        count -= copy_count;
+        pos += copy_count;
+        current_pos_ += copy_count;
+      }
```
seunghwak, line 160: What happens if `count` = 1000, `src_.back().size()` = 100, and `current_pos_` = 0?

At the end of the first loop, `copy_count` = 100, `count` = 900, `pos`=100, `current_pos_`=100. From the second loop, `copy_count`=0 and this loop won't finish or am I missing something?

Shouldn't we allocate additional buffers and reset `current_pos_` to 0 for this loop to finish?

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -110,45 +125,67 @@ class per_device_edgelist_t {
...
-      pos += copy_count;
-      current_pos_ += copy_count;
-
-      if (current_pos_ == src_.back().size()) { create_new_buffers(handle); }
-    }
+        std::get<0>(tuple), std::get<1>(tuple), std::get<2>(tuple), handle.get_stream());
+    });
 
-    handle.raft_handle().sync_stream();
+    handle.raft_handle().sync_stream(handle.get_stream());
```
seunghwak, line 188: If we add get_stream() to mtmg::handle, what about adding sync_stream to mtmg::handle as well?

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -153,11 +154,12 @@ class resource_manager_t {
       auto pos = local_rank_map_.find(rank);
       RAFT_CUDA_TRY(cudaSetDevice(pos->second.value()));
 
-      raft::handle_t tmp_handle;
-
+      size_t n_streams{16};
```
seunghwak, line 157: Why 16?

cpp/src/mtmg/vertex_result.cu
```diff
@@ -97,7 +97,7 @@ rmm::device_uvector<result_t> vertex_result_view_t<result_t>::gather(
       return vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(v);
     });
 
-  thrust::gather(handle.raft_handle().get_thrust_policy(),
+  thrust::gather(rmm::exec_policy(handle.get_stream()),
```
seunghwak, line 100: What about adding `(mtmg::)handle.get_thrust_policy()`?

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
ChuckHastings: Good suggestion, I'll look into that for next push.

cpp/include/cugraph/mtmg/resource_manager.hpp
seunghwak: Maybe # of GPUs? (assuming that 1 stream per thread and # threads == # GPUs)

ChuckHastings: Each GPU will have its own pool of streams.  The pool so far is used by different thread ranks copying data to the GPU independently.

ChuckHastings: I've added it as a parameter.

cpp/tests/mtmg/threaded_test.cu
```diff
@@ -107,7 +107,7 @@ class Tests_Multithreaded
     ncclGetUniqueId(&instance_manager_id);
 
     auto instance_manager = resource_manager.create_instance_manager(
-      resource_manager.registered_ranks(), instance_manager_id);
+      resource_manager.registered_ranks(), instance_manager_id, 4);
```
seunghwak, line 110: What is 4 here?

seunghwak (approved):
Looks good to me (besides the reason behind 4 in the code, some documentation will be helpful).

cpp/tests/mtmg/threaded_test.cu
ChuckHastings: Made this a constant, added a comment describing why it's 4 in the latest push.

PR #3897, opened 2023-09-28
nx-cugraph: handle louvain with isolated nodes

eriknw:
This handles isolated nodes in `louvain_communities` similar to what is done in #3886. This is expected to be a temporary fix until pylibcugraph can handle isolated nodes.

As a bonus, I added `isolates` algorithm :tada:

CC @naimnv @rlratzel


python/nx-cugraph/_nx_cugraph/__init__.py
```diff
@@ -31,14 +31,15 @@
...
+        "isolates",
         "louvain_communities",
         # END: functions
     },
     "extra_docstrings": {
         # BEGIN: extra_docstrings
         "betweenness_centrality": "`weight` parameter is not yet supported.",
         "edge_betweenness_centrality": "`weight` parameter is not yet supported.",
-        "louvain_communities": "`threshold` and `seed` parameters are currently ignored.",
+        "louvain_communities": "`seed` parameter is currently ignored.",
```
rlratzel, line 42: Do we need to add something in the docs about `max_level` being capped at 500?

eriknw: Added `(max: 500)` to the max_level parameter doc.

rlratzel (approved):

PR #3914, opened 2023-10-05
Add `HeteroGATConv` to `cugraph-pyg`

tingyu66:
Fixes https://github.com/rapidsai/graph_dl/issues/311
Adding @stadlmax POC code to cugraph-pyg


BradReesWork (approved):

stadlmax:
Looks good to me, but it would also be great if we could add a test against PyG like for other operators. The initialization should match PyG, so both this and the general primitive ideally should be tested for correctness. :)

tingyu66:
> Looks good to me, but it would also be great if we could add a test against PyG like for other operators. The initialization should match PyG, so both this and the general primitive ideally should be tested for correctness. :)

Sure, will test against PyG's HeteroConv + PyG's GATConv. What does "general primitive" here refer to?

stadlmax:
> Sure, will test against PyG's HeteroConv + PyG's GATConv. What does "general primitive" here refer to?

just the conv layer vs. the initialization of it

tingyu66:
@stadlmax I have added a test to validate against PyG. Could you review it again?

python/cugraph-pyg/cugraph_pyg/tests/nn/utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+HAS_PYG_24 = None
+if not isinstance(torch_geometric, MissingModule):
+    major, minor, patch = torch_geometric.__version__.split(".")[:3]
+    pyg_version = tuple(map(int, [major, minor, patch]))
+    HAS_PYG_24 = pyg_version >= (2, 4, 0)
+
+
+# TODO: Remove this function when dropping support to pyg 2.3
+def convert_edge_type_key(edge_type_str):
```
stadlmax, line 26: since this eventually goes back to something like
https://github.com/pyg-team/pytorch_geometric/pull/8163 fixing
https://github.com/pytorch/pytorch/issues/71203,
we actually should look at whether we want to update our indexing inside our own layer as well

tingyu66: PyG 2.4.0 fixed the issue of using reserved keywords as keys. `HeteroGATConv` now uses on `torch_geometric.nn.ParameterDict` to store parameters.

alexbarghi-nv (approved):
👍

PR #3916, opened 2023-10-05
new build all option

BradReesWork:
option to build all

Renamed the old buildAll function to buildDefault 
Added a check for the "all" argument


jnke2016 (approved):

build.sh
```diff
@@ -270,7 +272,7 @@ if buildAll || hasArg libcugraph; then
 fi
 
 # Configure, build, and install libcugraph_etl
-if buildAll || hasArg libcugraph_etl; then
+if buildDefault || hasArg libcugraph_etl || hasArg all; then
```
rlratzel, line 275: Do we want to build `libcugraph_etl` by default?  I could go either way.

build.sh
```diff
@@ -71,7 +73,7 @@ HELP="$0 [<target> ...] [<flag> ...]
    --clean                    - clean an individual target (note: to do a complete rebuild, use the clean target described above)
    -h                         - print this text
 
- default action (no args) is to build and install 'libcugraph' then 'libcugraph_etl' then 'pylibcugraph' then 'cugraph' targets
+ default action (no args) is to build and install 'libcugraph' then 'pylibcugraph' and then 'cugraph' targets
```
rlratzel, line 76: Looks like you removed `libcugraph_etl` here, but it's still built by default below.  I don't have a preference for building it by default or not, but just would like the help to be accurate.

rlratzel (changes requested):

rlratzel (approved):

PR #3922, opened 2023-10-09
Add Benchmark files to Datasets API

nv-rliu:
This PR replaces and is a continuation of #3844 (credit to @betochimas)

This PR adds the soc-LiveJournal, cit-Patents, europe_osm, and hollywood Dataset objects so that Dataset can be used for benchmarking purposes.

Eventually, all benchmarking components using these datasets will be refactored to save overhead when running CI, creating demo notebooks, and obtaining benchmark results.

Some changes to Dataset.get_edgelist. Now, get_edgelist has a `reader` option, where users can specify 'cudf' or 'pandas'. Also adds support for different column names other than 'src', 'dst', 'wgt'


python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -168,6 +168,10 @@ def get_edgelist(self, download=False):
         download : Boolean (default=False)
             Automatically download the dataset from the 'url' location within
             the YAML file.
+
+        create_using : module (default=cudf)
```
BradReesWork, line 172: should this list the accepted "create_using" option?

BradReesWork: you state "cudf" or "pandas" later, so that should be listed in the docs

BradReesWork (changes requested):

acostadon (approved):
Looks good to me

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -168,6 +168,9 @@ def get_edgelist(self, download=False):
...
+            Specify between cudf and pandas to use when reading the dataset.
```
rlratzel, line 173: This could be made more flexible if it accepted the function to call itself rather than a module.
```suggestion
        create_using : callable compatible with pandas.read_csv (default=cudf.read_csv)
            Object to call to read a CSV and return a pandas-compatible edgelist DataFrame.
```

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -183,7 +186,15 @@ def get_edgelist(self, download=False):
             header = None
             if isinstance(self.metadata["header"], int):
                 header = self.metadata["header"]
-            self._edgelist = cudf.read_csv(
+            if create_using is None:
+                reader = cudf
```
rlratzel, line 190: If applying the above suggestion:
```suggestion
                reader = cudf.read_csv
```

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -183,7 +186,15 @@ def get_edgelist(self, download=False):
...
+            elif str(type(create_using)) != "<class 'module'>":
+                raise RuntimeError("create_using must be a module.")
+            elif create_using.__name__ == "cudf" or "pandas":
```
rlratzel, line 193: if applying the above suggestion:
```suggestion
            import types
            elif type(create_using) == types.FunctionType:
```

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -183,7 +186,15 @@ def get_edgelist(self, download=False):
...
+                reader = create_using
+            else:
+                raise NotImplementedError()
```
rlratzel, line 196: if applying the above suggestion:
```suggestion
                raise ValueError("create_using must be a read_csv function compatible with pandas.read_csv")
```

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -183,7 +186,15 @@ def get_edgelist(self, download=False):
...
+            self._edgelist = reader.read_csv(
```
rlratzel, line 197: if applying the above suggestion:
```suggestion
            self._edgelist = create_using(
```

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -94,7 +95,10 @@ def setup_deprecation_warning_tests():
 # Helpers
 
 # check if there is a row where src == dst
-def has_loop(df):
+def has_selfloop(dataset):
+    if not dataset.metadata["is_directed"]:
+        return False
+    df = dataset.get_edgelist(download=True)
```
rlratzel, line 101: Some of the newer datasets added here a quite large. Can we still get decent coverage of the code added without testing every one?

python/cugraph/cugraph/datasets/dataset.py
rlratzel, line 193: if the suggested changes to `create_using` are not used, then this bug needs to be fixed.  This will always evaluate to True, but I think you meant this?
```suggestion
            elif create_using.__name__ in {"cudf", "pandas"}:
```

rlratzel (changes requested):

python/cugraph/cugraph/tests/utils/test_dataset.py
nv-rliu: Yes. In order to avoid downloading the large datasets along with our old ones every time for testing, the benchmark datasets are all tested separately in a single unit test.

acostadon (approved):
Looks good to me.

BradReesWork (approved):

rlratzel:
Closing this because it appears that #3957 contains all the commits from this PR, plus additional updates.

PR #3923, opened 2023-10-09
Updating Docs

BradReesWork:
starting with all the packages in a table


acostadon (approved):
I like the content. This is just a preference, but I would have put the detail about each section into an extra column in the table rather than below it.

As for an image above, I was thinking maybe a visual to divide cuGraph into its functionality.
Like a pie with sections for  

- data scientists/python
- C/C++ application integrators
- GNN solution spaces
Not sure that is the right approach just thinking

docs/cugraph/source/index.rst
```diff
@@ -1,15 +1,40 @@
...
+
+
+.. list-table:: RAPIDS Graph covers a range of graph libraries and packages, that includes:
+   :widths: 25 25 25
+   :header-rows: 1
+
+   * - Core
+     - GNNß
+     - Extension
+   * - :abbr:`cugraph (Heavy-weight Python wrapper with lot of guard rails)`
```
rlratzel, line 16: minor typo, but the bigger feedback is that IMO "heavy-weight" has bad connotations, do we want to use that?
```suggestion
   * - :abbr:`cugraph (Python wrapper with lots of guard rails)`
```

docs/cugraph/source/index.rst
```diff
@@ -1,15 +1,40 @@
...
+     - :abbr:`cugraph-ops (GPU-accelerated GNN aggregators and operators)`
+     - :abbr:`cugraph-service (Graph-as-a-service provides both Client and Server packages)`
+   * - :abbr:`pylibcugraph (light-weight Python wrapper no guard rails)`
```
rlratzel, line 19:
```suggestion
   * - :abbr:`pylibcugraph (light-weight Python wrapper with no guard rails)`
```

rlratzel (changes requested):
a couple very minor things, and a question

docs/cugraph/source/index.rst
BradReesWork: fixed

docs/cugraph/source/index.rst
BradReesWork: fixed

rlratzel (approved):

PR #3929, opened 2023-10-11
Temporarily suppress the timeout error

jnke2016:


jnke2016:
This PR is no longer needed because this [PR](https://github.com/rapidsai/dask-cuda/pull/1260) from `dask-cuda` increases the closing timeout of `Nanny` in LocalCUDACluster

PR #3932, opened 2023-10-12
MTMG multi node

ChuckHastings:
This PR extends the MTMG framework to support a multi-node configuration.

Multi-node configuration of MTMG assumes some externally available mechanism for communicating configuration parameters and the NCCL unique id between the processes that will participate.  For testing purposes we assume a shared file system where we can create a directory and read/write configuration information rather than constructing a more complex communication mechanism (NFS is sufficient for testing).


cpp/CMakeLists.txt
```diff
@@ -181,7 +181,9 @@ endif()
 
 set(CUGRAPH_SOURCES
     src/detail/shuffle_vertices.cu
-    src/detail/shuffle_vertex_pairs.cu
+    src/detail/shuffle_vertex_pairs_int32_int32.cu
+    src/detail/shuffle_vertex_pairs_int32_int64.cu
+    src/detail/shuffle_vertex_pairs_int64_int64.cu
```
seunghwak, line 186: What's the rationale behind this? Is this to cut compile time?

seunghwak: We should be more consistent on this to keep our codebase maintainable.

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -70,6 +70,8 @@ class resource_manager_t {
   {
     std::lock_guard<std::mutex> lock(lock_);
 
+    CUGRAPH_EXPECTS(remote_rank_set_.find(rank) == remote_rank_set_.end(),
```
seunghwak, line 73: Just to double check, is `rank` still local? I assume `remote_rank_set_` stores global ranks, and can we query `remote_rank_set_` with a local rank?

seunghwak: Or from now on, `rank` is global and `device_id` might serve as local GPU rank?

cpp/include/cugraph/mtmg/resource_manager.hpp
```diff
@@ -130,29 +152,40 @@ class resource_manager_t {
...
+                  [&local_ranks  = local_rank_map_,
+                   &remote_ranks = remote_rank_set_,
+                   &local_ranks_to_include](int rank) {
+                    if (local_ranks.find(rank) == local_ranks.end()) {
+                      CUGRAPH_EXPECTS(remote_ranks.find(rank) != remote_ranks.end(),
+                                      "requesting inclusion of an invalid rank");
+                    } else {
+                      local_ranks_to_include.push_back(rank);
+                    }
+                  });
```
seunghwak, line 168: I think using `std::copy_if` with `std::back_inserter` better documents the intention.

cpp/tests/mtmg/multi_node_threaded_test.cu
```diff
@@ -0,0 +1,603 @@
...
+
+    int num_local_gpus = gpu_list.size();
+    int num_threads    = num_local_gpus * 4;
+
+    //
+    //  This is intended to mimic a multi-node host application (non-MPI) integrating
+    //  with MTMG library.  This is a simple implementation using a shared file system
+    //  to pass configuration messages.  Terribly inefficient, but should mimic
+    //  expected behavior.
+    //
```
seunghwak, line 138: What do you think about this vs using mpi but creating just one process per node? We are already using MPI for testing, so the latter won't require a new execution environment. The former assumes availability of a sort of NFS and users need to setup the directory path correctly. This is not necessary if we just re-use the existing MPI test environment but just create one process per node.

cpp/CMakeLists.txt
ChuckHastings: Yes, this was to cut compile time.  It was the one file that was being constantly recompiled while I was testing this that took a long time.

I have been picking these up when I see them to split them.  I think ultimately we need to re-evaluate how we are doing type dispatching so that we can generally improve compile time.

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: In my implementation here (trying to somewhat mimic MPI) `rank` is global.  `device_id` is local.

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: I will look at this change.

cpp/CMakeLists.txt
seunghwak: Yeah... we should re-evaluate this, but AFAIK, this isn't the file with the longest compile time, and just splitting this file looks a bit odd. No problem in doing this for local development, but I guess pushing this change to the main repo (without really agreeing on how to tackle compile time issue) sounds a bit premature.

cpp/include/cugraph/mtmg/resource_manager.hpp
seunghwak: Yeah.. in that case, we need to review the documentation.
```
  /**
   * @brief add a local GPU to the resource manager.
   *
   * @param rank       The rank to assign to the local GPU
   * @param device_id  The device_id corresponding to this rank
   */
```
It is not clear to figure out that `rank` is global and `device_id` is local just seeing the documentation.

cpp/tests/mtmg/multi_node_threaded_test.cu
seunghwak: I was think more like updating

```
#define CUGRAPH_MG_TEST_PROGRAM_MAIN()                                                  \
  int main(int argc, char** argv)                                                       \
  {                                                                                     \
    cugraph::test::initialize_mpi(argc, argv);                                          \
    auto comm_rank = cugraph::test::query_mpi_comm_world_rank();                        \
    auto comm_size = cugraph::test::query_mpi_comm_world_size();                        \
    int num_gpus_per_node{};                                                            \
    RAFT_CUDA_TRY(cudaGetDeviceCount(&num_gpus_per_node));                              \
    RAFT_CUDA_TRY(cudaSetDevice(comm_rank % num_gpus_per_node));                        \
    ::testing::InitGoogleTest(&argc, argv);                                             \
    auto const cmd_opts = parse_test_options(argc, argv);                               \
    auto const rmm_mode = cmd_opts["rmm_mode"].as<std::string>();                       \
    auto resource       = cugraph::test::create_memory_resource(rmm_mode);              \
    rmm::mr::set_current_device_resource(resource.get());                               \
    cugraph::test::g_perf = cmd_opts["perf"].as<bool>();                                \
    cugraph::test::g_rmat_scale =                                                       \
      (cmd_opts.count("rmat_scale") > 0)                                                \
        ? std::make_optional<size_t>(cmd_opts["rmat_scale"].as<size_t>())               \
        : std::nullopt;                                                                 \
    cugraph::test::g_rmat_edge_factor =                                                 \
      (cmd_opts.count("rmat_edge_factor") > 0)                                          \
        ? std::make_optional<size_t>(cmd_opts["rmat_edge_factor"].as<size_t>())         \
        : std::nullopt;                                                                 \
    cugraph::test::g_test_file_name =                                                   \
      (cmd_opts.count("test_file_name") > 0)                                            \
        ? std::make_optional<std::string>(cmd_opts["test_file_name"].as<std::string>()) \
        : std::nullopt;                                                                 \
                                                                                        \
    auto ret = RUN_ALL_TESTS();                                                         \
    cugraph::test::finalize_mpi();                                                      \
    return ret;                                                                         \
  }
```

and create  `CUGRAPH_MTMG_TEST_PROGRAM_MAIN()` We can just broadcast NCCL unique ID here and we won't touch anything related to MPI after this (and we can even call cugraph::test::finalize_mpi() right after broadcasting NCCL unique ID).

jnke2016 (approved):

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Updated the parameter names and documentation to be `global_rank` and `local_device_id`.  Added a small improvement in the class documentation to identify the global rank concept.

cpp/include/cugraph/mtmg/resource_manager.hpp
ChuckHastings: Updated in next push.

cpp/tests/mtmg/multi_node_threaded_test.cu
ChuckHastings: Got the MPI configuration working in the next push.  I did not abstract it into a reusable function/MACRO, I can look at that when we added a second test (perhaps in a PR for adding Louvain or Jaccard).

cpp/tests/mtmg/multi_node_threaded_test.cu
```diff
@@ -0,0 +1,518 @@
...
+
+    int num_local_gpus = gpu_list.size();
+    int num_threads    = num_local_gpus * 4;
+
+    ncclUniqueId instance_manager_id{};
+
+    if (g_node_rank == 0) RAFT_NCCL_TRY(ncclGetUniqueId(&instance_manager_id));
+
+    RAFT_MPI_TRY(
+      MPI_Bcast(&instance_manager_id, sizeof(instance_manager_id), MPI_CHAR, 0, MPI_COMM_WORLD));
```
seunghwak, line 110: If we want to hide the existence of  MPI in individual tests, we may add

// valid only for MTMG tests, updated by CUGRAPH_MTMG_TEST_PROGRAM_MAIN()  
std::optional<ncclUniqueId> g_nccl_unique_id{std::nullopt};

after
```
// these variables are updated by command line arguments
static bool g_perf{false};
static std::optional<size_t> g_rmat_scale{std::nullopt};
static std::optional<size_t> g_rmat_edge_factor{std::nullopt};
static std::optional<std::string> g_test_file_name{std::nullopt};
```
https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/tests/utilities/base_fixture.hpp#L119

A bit annoying to add a global variable (only valid for MTMG tests) to all C++ tests, but I think this is OK as this is only for testing and the variable is std::optional (so having invalid values for non-MTMG tests is somewhat acceptable).

Then, we can expose our dependency on MPI only within `CUGRAPH_MTMG_TEST_PROGRAM_MAIN()`  and we can finalize MPI before invoking individual tests.

cpp/tests/mtmg/multi_node_threaded_test.cu
```diff
@@ -0,0 +1,518 @@
...
+#include <cugraph/mtmg/vertex_result.hpp>
+
+#include <raft/comms/mpi_comms.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+#include <rmm/device_uvector.hpp>
+
+#include <gtest/gtest.h>
+
+#include <nccl.h>
```
seunghwak, line 39: And I am not sure exposing NCCL in cuGraph is a good thing. Up to this point, we have been largely hiding the existence of NCCL under raft. Now we are explicitly dependent on NCCL. Hiding NCCL for MTMG may require an additional raft function and this can be a FIXME thing, but I feel like hiding explicit NCCL dependency is a good design choice.

seunghwak (approved):
LGTM in general.

I added few comments for potential code improvements (better hiding NCCL & MPI dependencies).

cpp/tests/mtmg/multi_node_threaded_test.cu
ChuckHastings: I was planning on visiting this issue in depth when I add Louvain.  I was also thinking of adding some test methods to hide some of these things.

cpp/include/cugraph/mtmg/instance_manager.hpp
```diff
@@ -89,7 +100,6 @@ class instance_manager_t {
   std::vector<std::unique_ptr<raft::handle_t>> raft_handle_{};
   std::vector<std::unique_ptr<ncclComm_t>> nccl_comms_{};
```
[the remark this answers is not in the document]
ChuckHastings: We can look at updating raft for this.  I'll talk with Corey.

cpp/tests/mtmg/multi_node_threaded_test.cu
ChuckHastings: I will address this also in the Louvain PR.  I can actually pick this dependency up through a raft header.  So if we address your other comment (about `ncclComm_t` and migrating some functionality in raft) that should clean this up as well.

cpp/tests/mtmg/multi_node_threaded_test.cu
ChuckHastings: I tried creating `g_nccl_unique_id` as a global and initializing it in the main program.  As currently constructed, each call to `create_instance_manager` needs a unique `g_nccl_unique_id`, I couldn't reuse them.  Each test could potentially use a different GPU configuration (something I eventually want to add).  So I needed to move the creation of the nccl_unique_id back into the test.

What I'm imagining is creating an MTMG test function that will create and broadcast the nccl unique id so that the use of MPI is restricted there.  It does require that we not finalize MPI before we run the tests.  But if we capture this in helper methods then we should be able to easily identify if we are using MPI calls that could potentially be a problem within one of the MTMG tests.

PR #3933, opened 2023-10-12
Setup Consistent Nightly Versions for Pip and Conda

divyegala:
See https://github.com/rapidsai/rmm/pull/1347


ci/build_python.sh
```diff
@@ -11,15 +11,24 @@ rapids-print-env
...
+version=$(rapids-generate-version)
+git_commit=$(git rev-parse HEAD)
+export RAPIDS_PACKAGE_VERSION=${version}
+echo "${version}" | tr -d '"' > VERSION
+
 rapids-logger "Begin py build"
 
 # TODO: Remove `--no-test` flags once importing on a CPU
 # node works correctly
+version_file_pylibcugraph="python/pylibcugraph/pylibcugraph/_version.py"
```
ajschmidt8, line 23: same comments as here: https://github.com/rapidsai/raft/pull/1880#discussion_r1375068582

ajschmidt8 (approved):

python/cugraph-dgl/cugraph_dgl/_version.py
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+import importlib.resources
+
+__version__ = (
+    importlib.resources.files("cugraph_dgl").joinpath("VERSION").read_text().strip()
```
rlratzel, line 19: I think this was mentioned in another PR for this in another repo, but I'm also not crazy about reading another file on import, but I think I understand the reasoning.  Can there be a comment above these lines explaining why this is happening, and summarize what's it's doing?
Also, it looks like `VERSION` redirects to `../../../VERSION`.  Does this mean when the package is installed (without something like `pip -e`) there's a file named `VERSION` in `site-packages` (or somewhere similar)?

rlratzel (changes requested):
Thanks for updating us. Since this change adds new infrastructure that we'll want to be able to maintain, I requested a few comments be added in places. Looks good otherwise.

python/cugraph-dgl/cugraph_dgl/_version.py
rlratzel: I'm curious if there's an advantage to having this file read `VERSION` in its package which then redirects to the repo `../../../VERSION` as opposed to just having this file read `../../../VERSION`? Or maybe I'm missing something.

divyegala: The symlink is necessary because importlib reads resources at the module level. For the multitude of ways that we can install a python package, we can only guarantee that the module level VERSION file will be available.

ChuckHastings (approved):

python/cugraph-dgl/cugraph_dgl/_version.py
vyasr: It's also worth noting that when the package is actually built, setuptools will copy the original files for symlinks pointing outside the package, so if you download a conda package or a wheel this `VERSION` file won't be a symlink anymore, it'll be a copy of the file.

rlratzel: > It's also worth noting that when the package is actually built, setuptools will copy the original files for symlinks pointing outside the package, so if you download a conda package or a wheel this `VERSION` file won't be a symlink anymore, it'll be a copy of the file.

Ah, I suspected something like that but I think this comment was the missing piece. Thanks!

rlratzel (approved):
Thank you very much, I think that is very helpful to future maintainers.

PR #3934, opened 2023-10-13
Add multigraph support to nx-cugraph

eriknw:


python/nx-cugraph/nx_cugraph/classes/graph.py
```diff
@@ -58,7 +59,7 @@ class Graph:
     node_values: dict[AttrKey, cp.ndarray[NodeValue]]
     node_masks: dict[AttrKey, cp.ndarray[bool]]
     key_to_id: dict[NodeKey, IndexValue] | None
-    _id_to_key: dict[IndexValue, NodeKey] | None
+    _id_to_key: list[NodeKey] | None
```
rlratzel, line 62: Is this a breaking change? If so, I'll add the `breaking` label

python/nx-cugraph/nx_cugraph/convert.py
```diff
@@ -254,7 +265,23 @@ def from_networkx(
         key_to_id = None
     else:
         col_iter = map(key_to_id.__getitem__, col_iter)
-    col_indices = cp.fromiter(col_iter, np.int32)
+    if graph.is_multigraph():
+        col_indices = np.fromiter(col_iter, np.int32)
+        num_multiedges = np.fromiter(
+            map(len, concat(map(dict.values, adj.values()))), np.int32
+        )
+        # cp.repeat is slow to use here, so use numpy instead
```
rlratzel, line 273: I didn't catch this before but we should probably add a dependency on `numpy` for this (instead of assuming we'll get it transitively through `cupy` or something else)

python/nx-cugraph/nx_cugraph/tests/test_convert.py
```diff
@@ -88,7 +91,10 @@ def test_convert():
     cp.testing.assert_array_equal(Gcg.node_values["foo"], [10, 20])
     assert Gcg.edge_values == Gcg.edge_masks == {}
     H = nxcg.to_networkx(Gcg)
-    assert set(G.edges) == set(H.edges) == {(0, 1)}
+    if G.is_multigraph():
+        assert set(G.edges) == set(H.edges) == {(0, 1, 0)}
```
rlratzel, line 95: I don't know if this is worth a comment or not, but can you explain why `G.edges` is used instead of `G.edges()`? (I'm not familiar with what the 3-tuple represents)

python/nx-cugraph/nx_cugraph/convert.py
```diff
@@ -144,7 +142,7 @@ def from_networkx(
         else:
             node_attrs = {node_attrs: None}
 
-    if graph.__class__ in {nx.Graph, nx.DiGraph}:
+    if graph.__class__ in {nx.Graph, nx.DiGraph, nx.MultiGraph, nx.MultiDiGraph}:
```
rlratzel, line 145: I'm assuming we can't use `isinstance()` here because this is accessing a private attr below?

python/nx-cugraph/nx_cugraph/tests/test_convert.py
rlratzel: nevermind, I see that the third value is an edge key from the docstring in MultiGraph.

python/nx-cugraph/nx_cugraph/classes/graph.py
eriknw: Technically, yes, practically, no. Nearly all internal uses of it was as `G.id_to_key[index]`, which works whether a dict or a list. It is breaking if one did other things such as `G.id_to_key.items()`.

python/nx-cugraph/nx_cugraph/convert.py
eriknw: Correct, we want to be more conservative with subclasses.

python/nx-cugraph/nx_cugraph/convert.py
eriknw: Will do!

python/nx-cugraph/nx_cugraph/classes/graph.py
rlratzel: ok I'll leave off the "breaking" label then.

python/nx-cugraph/nx_cugraph/convert.py
eriknw: Done.

rlratzel (approved):

PR #3905, opened 2023-10-02
Updates to 23.12

raydouglass:
A couple PRs were merged after `branch-23.12` was created and contained RAPIDS versions that need to be updated in `branch-23.12`.

Ref:
- https://github.com/rapidsai/cugraph/pull/3838
- https://github.com/rapidsai/cugraph/pull/3852

PR #3935, opened 2023-10-13
[DO NOT MERGE] Debug devcontainers runs

AyodeAwe:
Debugging failures in the devcontainer runs:

https://github.com/rapidsai/cugraph/actions/runs/6511511984/job/17687358314?pr=3919
https://github.com/rapidsai/cugraph/actions/runs/6511511984/job/17687358551?pr=3919


rlratzel:
@AyodeAwe @raydouglass - can we close this PR?  I made the relevant changes here to [Ray's PR](https://github.com/rapidsai/cugraph/pull/3905) yesterday and merged it, and the `devcontainers` jobs seem to be passing for other PRs now.  All 23.10 forward merges are also in 23.12, so the other non-devcontainer changes here are either already applied or are something else that maybe shouldn't be merged?

PR #3941, opened 2023-10-17
Moves more MG graph ETL to libcugraph and re-enables MG tests in CI

jnke2016:
This PR includes changes that moves some of the MG graph etl steps (such as computing number of edges) to libcugraph to reduce the amount of dask overhead involved in graph creation. Those ETL steps were also responsible for various dask-related transient errors that caused us to temporarily disable MG testing in CI.  These changes allow us to re-enable MG testing in CI, so this PR includes that update too.


python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -323,10 +329,7 @@ def __from_edgelist(
             ddf, _client, return_type="dict"
         )
         del ddf
```
VibhuJawa, line 331: We should not `persist` these dataframes anymore.  We used to `persist` to prevent `edata` computation.

VibhuJawa (changes requested):
Left comment for `dask` part. That should also allow us not `cache` anything in `b/w` which should allow us to get rid of the `caching` MG problem we ran into.

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -125,13 +146,16 @@ cdef class SGGraph(_GPUGraph):
                   GraphProperties graph_properties,
                   src_or_offset_array,
                   dst_or_index_array,
+                  vertices_array=None,
```
rlratzel, line 149: Can you move this to a position after `input_array_format` so calling this with position-only params still works?  Otherwise, this requires users to use kwarg params which would be a breaking change.

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -286,93 +322,141 @@ cdef class MGGraph(_GPUGraph):
...
                   store_transposed=False,
-                  num_edges=-1,
-                  do_expensive_check=False,
+                  size_t num_arrays=1, # default value to not break users
+                  do_expensive_check=True,
                   edge_id_array=None,
-                  edge_type_array=None):
+                  edge_type_array=None,
+                  drop_self_loops=False,
+                  drop_multi_edges=False):
```
rlratzel, line 345: same with these: can these be arranged so the new params are last so users with position-only args aren't broken?

rlratzel (changes requested):
looks good, just some feedback on docstrings and one request to change the order of params so this doesn't become a breaking change.

python/pylibcugraph/pylibcugraph/graphs.pyx
jnke2016: Right! I had some tests failing because of this so I passed argument by keyword. I will also preserve the order (position) as you suggested.

VibhuJawa:
I somehow cant add a PR to your fork @jnke2016 . 

Anyways below should give you the required changes to apply. 

https://github.com/rapidsai/cugraph/compare/branch-23.12...VibhuJawa:cugraph:branch-23.12_re-enable-mg-testing

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -89,51 +88,58 @@ def _make_plc_graph(
...
     ):
 
         weights = None
         edge_ids = None
         edge_types = None
+        num_arrays = len(edata_x)
 
         if simpleDistributedGraphImpl.edgeWeightCol in edata_x[0]:
-            weights = _get_column_from_ls_dfs(
-                edata_x, simpleDistributedGraphImpl.edgeWeightCol
```
VibhuJawa, line 101: Can you also remove the now dead function `_get_column_from_ls_dfs ` please.

VibhuJawa (changes requested):

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
jnke2016: Right,  I removed that locally. I will push my changes shortly

cpp/include/cugraph_c/resource_handle.h
```diff
@@ -57,6 +57,18 @@ typedef struct cugraph_resource_handle_ {
  */
 cugraph_resource_handle_t* cugraph_create_resource_handle(void* raft_handle);
 
+/**
+ * @brief get comm_size from resource handle
+ *
+ * If the resource handle has been configured for multi-gpu, this will return
+ * the comm_size for this cluster.  If the resource handle has not been configured for
```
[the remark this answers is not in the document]
jnke2016: It is described in `@brief ` Line 61

rlratzel:
Added a `DO NOT MERGE` label because Joseph has at least one more commit to push.

AyodeAwe (approved):

python/pylibcugraph/pylibcugraph/graphs.pyx
```diff
@@ -146,14 +141,14 @@ cdef class SGGraph(_GPUGraph):
...
                   dst_or_index_array,
-                  vertices_array=None,
                   weight_array=None,
                   store_transposed=False,
                   renumber=False,
                   do_expensive_check=False,
                   edge_id_array=None,
                   edge_type_array=None,
                   input_array_format="COO",
+                  vertices_array=None,
```
rlratzel, line 151: This (and also the updated MGGraph ctor) still does not have all new args at the end, but maybe since only `drop_self_loops` and `drop_multi_edges` are affected, it's not that likely to break users using only positional params?  I'll approve it but I'll also label this PR as a breaking change.

rlratzel (approved):
Approved, but I'll label the PR as a breaking change since users constructing PLC graph objects that include `drop_self_loops` and `drop_multi_edges` as positional args will be affected.

PR #3943, opened 2023-10-19
nx-cugraph: handle seed argument in edge_betweenness_centrality

eriknw:
CC @rlratzel who brought this to my attention and can test this fix


rlratzel (approved):
I tested this change and it passed.

rlratzel:
Followup: do we know why our CI coverage didn't initially catch this?  I'm wondering if this PR should also have an updated test.

eriknw:
A test would be nice... and it sounds like you have a case that exercises this ;) . Will your code be added to the repo at some point?

PR #3944, opened 2023-10-19
nx-cugraph: xfail test_louvain.py:test_threshold in Python 3.9

eriknw:
This test is flaky b/c Louvain in PLC does not have a seed or random state parameter. This test holds the seed fixed and changes the threshold, but clearly this may not work if we can't set the seed.

### DIFF
--- a/python/nx-cugraph/nx_cugraph/interface.py
+++ b/python/nx-cugraph/nx_cugraph/interface.py
@@ -176,6 +176,10 @@ class BackendInterface:
                     ): louvain_different,
                     key("test_louvain.py:test_none_weight_param"): louvain_different,
                     key("test_louvain.py:test_multigraph"): louvain_different,
+                    # See networkx#6630
+                    key(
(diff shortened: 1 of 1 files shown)


rlratzel:
> _This test is flaky b/c Louvain in PLC does not have a seed or random state parameter._

Should we also file an issue to add a seed/random state param to PLC Louvain?  cc @ChuckHastings

rlratzel (approved):

ChuckHastings:
> > _This test is flaky b/c Louvain in PLC does not have a seed or random state parameter._
> 
> Should we also file an issue to add a seed/random state param to PLC Louvain? cc @ChuckHastings

There is no randomness in our C++ Louvain implementation, it is deterministic based on the vertex numbering.  We are exploring some updates related to Leiden that might add some randomness, but the current implementation uses none.

naimnv (approved):

PR #3945, opened 2023-10-19
nx-cugraph: add k_truss and degree centralities

eriknw:
New algorithms:
  - `degree_centrality`
  - `in_degree_centrality`
  - `k_truss`
  - `number_of_selfloops`
  - `out_degree_centrality`

Also, rename `row_indices, col_indices` to `src_indices, dst_indices`

### DIFF
--- a/python/nx-cugraph/nx_cugraph/algorithms/core.py
+++ b/python/nx-cugraph/nx_cugraph/algorithms/core.py
@@ -63,7 +63,7 @@ def k_truss(G, k):
         edge_masks = {key: val[edge_indices] for key, val in G.edge_masks.items()}
     # Renumber step 2: edge indices
     mapper = cp.zeros(len(G), src_indices.dtype)
-    mapper[node_indices] = cp.arange(node_indices.size, dtype=np.int64)
+    mapper[node_indices] = cp.arange(node_indices.size, dtype=mapper.dtype)
     src_indices = mapper[src_indices]
     dst_indices = mapper[dst_indices]
     # Renumber step 3: node values
--- a/python/nx-cugraph/nx_cugraph/algorithms/core.py
+++ b/python/nx-cugraph/nx_cugraph/algorithms/core.py
@@ -1,0 +1,91 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import cupy as cp
+import networkx as nx
+import numpy as np
+import pylibcugraph as plc
+
+import nx_cugraph as nxcg
+from nx_cugraph.utils import networkx_algorithm, not_implemented_for
+
+__all__ = ["k_truss"]
+
+
+@not_implemented_for("directed")
+@not_implemented_for("multigraph")
+@networkx_algorithm
+def k_truss(G, k):
+    if is_nx := isinstance(G, nx.Graph):
+        G = nxcg.from_networkx(G, preserve_all_attrs=True)
+    if nxcg.number_of_selfloops(G) > 0:
+        raise nx.NetworkXError(
+            "Input graph has self loops which is not permitted; "
+            "Consider using G.remove_edges_from(nx.selfloop_edges(G))."
+        )
+    # TODO: create renumbering helper function(s)
+    if k < 3:
+        # Drop nodes with zero degree
+        degrees = G._degrees_array()
+        # Renumber step 0: node indices
+        node_indices = degrees.nonzero()[0]
+        if degrees.size == node_indices.size:
+            # No change
+            return G if is_nx else G.copy()
+        src_indices = G.src_indices
+        dst_indices = G.dst_indices
+        # Renumber step 1: edge values (no changes needed)
+        edge_values = {key: val.copy() for key, val in G.edge_values.items()}
+        edge_masks = {key: val.copy() for key, val in G.edge_masks.items()}
+    else:
+        # int dtype for edge_indices would be preferred
+        edge_indices = cp.arange(G.src_indices.size, dtype=np.float64)
+        src_indices, dst_indices, edge_indices, _ = plc.k_truss_subgraph(
+            resource_handle=plc.ResourceHandle(),
+            graph=G._get_plc_graph(edge_array=edge_indices),
+            k=k,
+            do_expensive_check=False,
+        )
+        # Renumber step 0: node indices
+        node_indices = cp.unique(cp.concatenate([src_indices, dst_indices]))
+        # Renumber step 1: edge values
+        edge_indices = edge_indices.astype(np.int64)
+        edge_values = {key: val[edge_indices] for key, val in G.edge_values.items()}
+        edge_masks = {key: val[edge_indices] for key, val in G.edge_masks.items()}
+    # Renumber step 2: edge indices
+    mapper = cp.zeros(len(G), src_indices.dtype)
+    mapper[node_indices] = cp.arange(node_indices.size, dtype=np.int64)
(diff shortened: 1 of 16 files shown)


python/nx-cugraph/nx_cugraph/algorithms/core.py
```diff
@@ -0,0 +1,91 @@
...
+            # No change
+            return G if is_nx else G.copy()
+        src_indices = G.src_indices
+        dst_indices = G.dst_indices
+        # Renumber step 1: edge values (no changes needed)
+        edge_values = {key: val.copy() for key, val in G.edge_values.items()}
+        edge_masks = {key: val.copy() for key, val in G.edge_masks.items()}
+    else:
+        # int dtype for edge_indices would be preferred
+        edge_indices = cp.arange(G.src_indices.size, dtype=np.float64)
```
eriknw, line 51: Does PLC only handle edge weights with float dtypes?

eriknw: Yes. Only float32 and float64. This is handled more generally in #3954:
https://github.com/rapidsai/cugraph/pull/3954/files#diff-b87ebafc61a8ff368beff74819475091903f5c7590ae4a7d56363fb5ed40c1a9

python/nx-cugraph/nx_cugraph/algorithms/core.py
```diff
@@ -0,0 +1,91 @@
...
+def k_truss(G, k):
+    if is_nx := isinstance(G, nx.Graph):
+        G = nxcg.from_networkx(G, preserve_all_attrs=True)
+    if nxcg.number_of_selfloops(G) > 0:
+        raise nx.NetworkXError(
+            "Input graph has self loops which is not permitted; "
+            "Consider using G.remove_edges_from(nx.selfloop_edges(G))."
+        )
+    # TODO: create renumbering helper function(s)
+    if k < 3:
```
rlratzel, line 36: Can you add a comment as an overview of what's being done differently here vs. the else block and why? (eg. PLC isn't even being called, etc.)

python/nx-cugraph/nx_cugraph/algorithms/core.py
```diff
@@ -0,0 +1,91 @@
...
+    node_masks = {key: val[node_indices] for key, val in G.node_masks.items()}
+    # Renumber step 4: key_to_id
+    if (id_to_key := G.id_to_key) is not None:
+        key_to_id = {
+            id_to_key[old_index]: new_index
+            for new_index, old_index in enumerate(node_indices.tolist())
+        }
+    else:
+        key_to_id = None
+    new_graph = G.__class__.from_coo(
```
rlratzel, line 80: I'm curious why `from_coo` needs to be called this way.

rlratzel (commented):
Thanks Erik, LGTM, I just had two requests/questions.

Also, are we getting good coverage via the NX test suite? We were bit by this for edge BC so I want to make sure we're covered.

python/nx-cugraph/nx_cugraph/algorithms/core.py
eriknw: I thought I did leave a comment: we drop nodes with zero degree. No need to call into pylibcugraph just for that. What comment would you find helpful here?

python/nx-cugraph/nx_cugraph/algorithms/core.py
eriknw: I didn't give this much thought or have a strong preference. `from_coo` is a classmethod, so there are three reasonable way to call it:
```python
G.from_coo(...)
G.__class__.from_coo(...)
type(G).from_coo(...)
```
Do you have a preference? Sometimes I think it's weird to read constructors being used off of an instance of a class, hence why I went with `G.__class__`, which is clear it's a class method and is also a little more flexible than `type(G)`. Perhaps this is a "me" thing and I should just use `G.from_coo`. Thoughts?

eriknw:
> Also, are we getting good coverage via the NX test suite? We were bit by this for edge BC so I want to make sure we're covered.

Great question. Yeah, it's pretty good. I think lines 29, 31, 43, 79 aren't covered, which doesn't concern me atm.

After #3954, I think it will be a lot easier to do comparison testing with a variety of input graphs without too much effort.

fwiw, I do like to have solid coverage, but our shared tooling today isn't great. It would help if I could quickly find "nx-cugraph" tests in CI. As is, I test for major coverage during development and occasionally look at coverage reports locally.

python/nx-cugraph/nx_cugraph/algorithms/core.py
rlratzel: sorry I meant a comment describing the overall goal of the entire `k < 3` block vs. the `else` block. I admittedly don't know much about k_truss (maybe that's the problem?), but I was hoping to get at least a rough idea why they're so different.  For instance, I'm curious why `plc.k_truss_subgraph()` couldn't just be used unconditionally even when k < 3, or why we wouldn't remove zero degree nodes in both cases. However, if this would require an explanation of the k_truss algo itself, then feel free to ignore this.

python/nx-cugraph/nx_cugraph/algorithms/core.py
rlratzel: I see, I've always used your option 1 for classmethods (`G.from_coo(...)`), but the pattern you're using is fine by me if you don't think others will get tripped up by it.  I thought it was called that way to somehow get a specific behavior I wasn't familiar with, but your explanation is reasonable to me.

rlratzel:
> It would help if I could quickly find "nx-cugraph" tests in CI.

I added `rapids_logger` calls to the [CI script](https://github.com/rapidsai/cugraph/blob/9b28458dae82e85270bad0f8abf806f318fc84c8/ci/test_python.sh#L111) to announce in the log when it's running the nx-cugraph tests.  You should be able to search for "pytest nx-cugraph" or "pytest networkx" to find them.  Would that help, or did I misunderstand?

eriknw:
>> It would help if I could quickly find "nx-cugraph" tests in CI.
> 
> I added rapids_logger calls to the [CI script](https://github.com/rapidsai/cugraph/blob/9b28458dae82e85270bad0f8abf806f318fc84c8/ci/test_python.sh#L111) to announce in the log when it's running the nx-cugraph tests. You should be able to search for "pytest nx-cugraph" or "pytest networkx" to find them. Would that help, or did I misunderstand?

I suppose that helps somewhat. Browsing and searching CI is pretty slow and cumbersome, so my preferred UX would to have an exandable/collapsable `nx-cugraph` section.

rlratzel:
> my preferred UX would to have an exandable/collapsable `nx-cugraph` section.

ah, that would be nice.  I'll keep that in mind.  I think it would require editing the yaml in the .github folder, which I'm not that familiar with, but I like that idea in general.

python/nx-cugraph/nx_cugraph/algorithms/core.py
eriknw: Gotcha. Take a quick look at k_truss documentation:
https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.core.k_truss.html

See: _"[...] where every edge is incident to at least `k-2` triangles."_. So, `k < 3` is an edge case.

> I'm curious why plc.k_truss_subgraph() couldn't just be used unconditionally even when k < 3

It can--we can delete this entire branch if we wanted. PLC behaves the same as networkx for `0 <= k < 3`. We do the branch b/c it is faster, and it lets us check the "stop early" condition if there are no nodes of degree zero. I like fast, and I don't think this branch is excessively complicated.

> why we wouldn't remove zero degree nodes in both cases

k_truss always removes zero-degree nodes. PLC's k_truss does this for us when we call it.

I suppose I can add a couple comments to the code like I just explained here ;) --thanks for the discussion.

eriknw: More detailed code comment added.

python/nx-cugraph/nx_cugraph/algorithms/core.py
eriknw: I added a code comment about this too :)

rlratzel (approved):

python/nx-cugraph/nx_cugraph/algorithms/core.py
eriknw: Oh, I remember the primary reason I made this code branch in the first place and why I think it should exist for now: I was exploring renumbering and wanted to try multiple approaches. I'm sure we'll be doing more renumbering soon.

PR #3952, opened 2023-10-23
Errors compiling for DLFW on CUDA 12.3

ChuckHastings:
DLFW compilation found a few issues building cugraph 23.10.

This PR addresses these issues.  Two main issues:
* We were brace initializing `raft::device_span` objects with only `nullptr`.  The device span already defaults to `nullptr`, and the class hierarchy doesn't support just passing `nullptr`.
* We were brace initializing the `raft::random::DeviceState`, but there's no default constructor.  Changed it to not brace initialize and require a reference to make it clear that you must initialize the DeviceState.


seunghwak (approved):
LGTM

robertmaynard (approved):

PR #3956, opened 2023-10-24
Python Build Dependency Issues

nv-rliu:
Attempting to resolve [ucx version error](https://github.com/rapidsai/cugraph/actions/runs/6621722770/job/18010826156?pr=3550#step:7:1206) to unblock CI python builds


nv-rliu:
I believe the `ucx` version issue is no longer blocking CI. With that, I will be closing this PR

PR #3957, opened 2023-10-26
`Resultset` and `Dataset` Refactors

nv-rliu:
This PR replaces and is a continuation of #3857 (by @betochimas)

> This PR primarily adds testing for the `Resultset` class, introduced earlier in 23.10. The tests take a similar approach to test_dataset, creating a temporary directory to test downloading all result files. To align `Resultset` and `Dataset`, the setter and getter for each download directory is moved into `DefaultDownloadDir`, so that each class shares an instance of `DefaultDownloadDir` and can be configured independently, although their default locations are still both dependent on the RAPIDS_DATASET_ROOT_DIR_PATH environment variable. The old patterns are present but commented-out, so this change would be breaking. 

This PR also removes the deprecated `experimental.datasets` package due to it being promoted to stable for >=1 release.


rlratzel (approved):

raydouglass:

nv-rliu:
Closes https://github.com/rapidsai/graph_dl/issues/297

PR #3961, opened 2023-10-30
[BUG] Check if Dask has quit to avoid throwing an exception and triggering a segfault on ddp exit

alexbarghi-nv:
Currently, when training with ddp, if dask exits before the `CuGraphStore` is cleaned up, an exception is thrown, which causes ddp to quit with an error, which then causes a segfault, making users think that the workflow has failed when it has actually succeeded.  This bug gracefully displays a warning if the dask dataset can't be deleted, which resolves this issue.


VibhuJawa (approved):
LGTM

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -320,7 +320,13 @@ def __init__(
     def __del__(self):
         if self.__is_graph_owner:
             if isinstance(self.__graph._plc_graph, dict):
-                distributed.get_client().unpublish_dataset("cugraph_graph")
+                try:
+                    distributed.get_client().unpublish_dataset("cugraph_graph")
+                except Exception:
```
rlratzel, line 325: I'm not familiar with the error, but is it possible to catch a more specific exception?  I'm just worried this has the potential to also catch typos and other things we should know about, and we forever end up just assuming it's not able to unpublish.
If catching everything is really better though for this, then feel free to ignore.

tingyu66 (approved):
:+1:

rlratzel (approved):

PR #3965, opened 2023-10-31
nx-cugraph: add CC for undirected graphs to fix k-truss

### LINKED ISSUE #3963: Add limitation of one connected component within input Graph from `nx_cugraph.k_truss`
The current cugraph k_truss algo is not supported for input graphs with >1 connected component (separate issue needed).

In order to prevent nx-cugraph users from receiving bad results/crashes, `nx_cugraph.k_truss()` should raise an informative exception if the input graph has >1 connected component.

One proposal is to simply call the PLC WCC algo and check the number of components, then raise `NotImplementedError` if >1.  This check and exception can be removed when a future version of PLC k_truss that supports >1 connected component is available (24.02 at the earliest?).

cc @ChuckHastings

eriknw:
Fixes #3963 and add `connected_components`, `is_connected`, `node_connected_component`, and `number_connected_components`.

Also updated `_groupby` to handle groups that are not consecutive integers starting with 0.

Also, `plc.weakly_connected_components` does not handle isolated nodes well, and I needed to handle this at the Python layer as was done in #3897


python/nx-cugraph/nx_cugraph/algorithms/components/connected.py
```diff
@@ -0,0 +1,130 @@
...
+    # return cp.unique(labels).size
+
+
+@number_connected_components._can_run
+def _(G):
+    # NetworkX <= 3.2.1 does not check directedness for us
+    try:
+        return not G.is_directed()
+    except Exception:
+        return False
```
eriknw, line 56: I fixed this upstream here: https://github.com/networkx/networkx/pull/7074

rlratzel (commented):
changes LGTM. Should there be a test for k_truss that passes 
```
G=nx.Graph([(0,1), (1,2), (2,3), (3,0), (4,5), (5,6), (6,4)])
```
and asserts that `NotImplementedError` is raised, which we then `xfail` when k_truss is updated?
OTOH, maybe that's overkill for (hopefully) temporary code that I'm assuming was at least manually tested? I'm fine either way.

eriknw:
> changes LGTM. Should there be a test for k_truss that passes

Thanks for providing a test case, but now we actually have the opposite problem: now we don't test the main branches of k_truss! All test graphs apparently have more than 1 connected components.

rlratzel:
> Thanks for providing a test case, but now we actually have the opposite problem: now we don't test the main branches of k_truss! All test graphs apparently have more than 1 connected components.

Karate club has only 1 CC; maybe add a quick test for that too to make sure we at least exercise `nx_cugraph`->`PLC`->`libcugraph` and back?

eriknw:
> Karate club has only 1 CC; maybe add a quick test for that too to make sure we at least exercise nx_cugraph->PLC->libcugraph and back?

Excellent suggestion, although we'd need #3954 to test with karate club (b/c it has string node values). Test added, and code is exercised 👍

I'm looking forward to having more graph generators to enable more comparison testing of algorithms.

python/nx-cugraph/nx_cugraph/utils/misc.py
```diff
@@ -12,43 +12,60 @@
...
+                prev = cur
+
+
 __all__ = ["_groupby", "_seed_to_int"]
 
 
-def _groupby(groups: cp.ndarray, values: cp.ndarray) -> dict[int, cp.ndarray]:
+def _groupby(
+    groups: cp.ndarray, values: cp.ndarray, groups_are_canonical: bool = False
+) -> dict[int, cp.ndarray]:
```
eriknw, line 39: I'm really happy with this groupby implementation btw. Using `prepend` with `cp.diff` is a little slow compared to `cp.diff` w/o it, but this isn't where the bulk of the time is spent.

PR #3968, opened 2023-11-01
nx-cugraph: add `eigenvector_centrality`, `katz_centrality`, `hits`, `pagerank`

eriknw:
Add `eigenvector_centrality`, `katz_centrality`, and `hits`. I may add pagerank next.


python/nx-cugraph/nx_cugraph/interface.py
```diff
@@ -67,7 +67,11 @@ def key(testpath):
         louvain_different = "Louvain may be different due to RNG"
         no_string_dtype = "string edge values not currently supported"
 
-        xfail = {}
+        xfail = {
+            key(
+                "test_eigenvector_centrality.py:TestEigenvectorCentrality.test_P3"
+            ): "Power iteration failed (don't know why; networkx converges in 8 steps)"
```
eriknw, line 73: I have no idea why PLC's eigenvector centrality would fail to converge for this graph. NetworkX converges in 8-ish steps.
The graph has three nodes with the following indices:
- `src_indices = [0, 1, 1, 2]`
- `dst_indices = [1, 0, 2, 1]`

In matrix form, this looks like this:
```
[[0 1 0]
 [1 0 1]
 [0 1 0]]
```

eriknw: I will raise a separate issue for this.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/pagerank_alg.py
```diff
@@ -0,0 +1,90 @@
...
+        "resource_handle": plc.ResourceHandle(),
+        "graph": G._get_plc_graph(weight, 1, np.float32, store_transposed=True),
+        "precomputed_vertex_out_weight_vertices": None,
+        "precomputed_vertex_out_weight_sums": None,
+        "initial_guess_vertices": None
+        if nstart is None
+        else cp.arange(N, dtype=index_dtype),  # Why is this necessary?
+        "initial_guess_values": nstart,
+        "alpha": alpha,
+        "epsilon": tol**0.5 / N,  # TODO: see if tol is handled the same by PLC and NX
```
eriknw, line 61: I did this to `tol` to pass tests... I still need to compare how PLC and networkx use tolerance.

eriknw:
This is ready for review. There is nothing more I plan to add.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/pagerank_alg.py
```diff
@@ -0,0 +1,99 @@
...
+    max_iter=100,
+    tol=1.0e-6,
+    nstart=None,
+    weight="weight",
+    dangling=None,
+):
+    """`dangling` parameter is not supported."""
+    # Use np.float32 dtype everywhere for performance. In the future, it may
+    # be nice to use whatever dtype the user provides (such as np.float64) and
+    # let PLC "do the right thing" to make sure dtypes of inputs are compatible.
```
eriknw, line 38: pagerank is the only algorithm that gave me errors regarding mismatched dtypes, but I went ahead and handled all the power iteration algorithms like this for now.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/pagerank_alg.py
```diff
@@ -0,0 +1,99 @@
...
+    if max_iter <= 0:
+        raise nx.PowerIterationFailedConvergence(max_iter)
+    kwargs = {
+        "resource_handle": plc.ResourceHandle(),
+        "graph": G._get_plc_graph(weight, 1, np.float32, store_transposed=True),
+        "precomputed_vertex_out_weight_vertices": None,
+        "precomputed_vertex_out_weight_sums": None,
+        "initial_guess_vertices": None
+        if nstart is None
+        else cp.arange(N, dtype=index_dtype),  # Why is this necessary?
```
eriknw, line 67: I find it strange that we need to provide node indices here (and 12 lines below). I think other places in PLC assume `0, 1, ..., N-1` if indices aren't given.

eriknw:
Idea: perhaps we could add `dtype=` keyword argument specific to the `nx-cugraph` backend for algorithms such as in this PR. Would users want this? There are performance and memory implications. Other backends with native data structures could have the same pattern.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/pagerank_alg.py
eriknw: Better dtype handling added to algorithms in this PR. Resolving.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/hits_alg.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@networkx_algorithm(
+    extra_params={
+        'weight : string or None, optional (default="weight")': (
+            "The edge attribute to use as the edge weight."
+        )
+    }
+)
+def hits(G, max_iter=100, tol=1.0e-8, nstart=None, normalized=True, *, weight="weight"):
```
eriknw, line 30: XREF: https://github.com/networkx/networkx/pull/7081

`weight=` argument may be added to networkx soon.

python/nx-cugraph/nx_cugraph/algorithms/centrality/eigenvector.py
```diff
@@ -0,0 +1,64 @@
...
+    else:
+        dtype = np.float32
+    if nstart is not None:
+        # Check if given nstart is valid even though we don't use it
+        nstart = G._dict_to_nodearray(nstart, dtype=dtype)
+        if (nstart == 0).all():
+            raise nx.NetworkXError("initial vector cannot have all zero values")
+        if (total := nstart.sum()) == 0:
+            raise ZeroDivisionError
+        nstart /= total
```
rlratzel, line 52: I'm not following why we're processing `nstart` like this. As a user, this makes me think it's being used if I see these exceptions. 

Why isn't this raising `NotImplementedError`, or even a warning that it's not being used if set?

python/nx-cugraph/nx_cugraph/algorithms/centrality/katz.py
```diff
@@ -0,0 +1,98 @@
...
+        return {}
+    if dtype is not None:
+        dtype = _get_float_dtype(dtype)
+    elif weight in G.edge_values:
+        dtype = _get_float_dtype(G.edge_values[weight].dtype)
+    else:
+        dtype = np.float32
+    if nstart is not None:
+        # Check if given nstart is valid even though we don't use it
+        nstart = G._dict_to_nodearray(nstart, 0, dtype)
```
rlratzel, line 56: Same question as above.

python/nx-cugraph/nx_cugraph/algorithms/centrality/katz.py
```diff
@@ -0,0 +1,98 @@
...
+    beta=1.0,
+    max_iter=1000,
+    tol=1.0e-6,
+    nstart=None,
+    normalized=True,
+    weight=None,
+    *,
+    dtype=None,
+):
+    return normalized
```
rlratzel, line 98: Can you remind me again how `_can_run` is used? I'm mainly wondering if the `NotImplementedError` being raised above is redundant or if it's needed for something `_can_run` doesn't handle?

Also, is there a reason this isn't checking `nstart` and returning False if it's set?  Same question for eigenvector_centrality.  I think I'm not understanding the difference between "unused" and "unsupported" from the user's perspective.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/hits_alg.py
```diff
@@ -0,0 +1,73 @@
...
+        if nx.__version__[:3] <= "3.2":
+            raise ValueError("`maxiter` must be a positive integer.")
+        raise nx.PowerIterationFailedConvergence(max_iter)
+    try:
+        node_ids, hubs, authorities = plc.hits(
+            resource_handle=plc.ResourceHandle(),
+            graph=G._get_plc_graph(weight, 1, dtype, store_transposed=True),
+            tol=tol,
+            initial_hubs_guess_vertices=None,
+            initial_hubs_guess_values=nstart,
```
rlratzel, line 62: Have you tested this with `nstart` passed in?  I'm thinking you need to also pass a vertices array (as you mentioned in pagerank below) based on [this](https://github.com/rapidsai/cugraph/blob/f4bcdc2667a15e2e5031987550322a9d29d8f713/python/pylibcugraph/pylibcugraph/hits.pyx#L121-L123).

python/nx-cugraph/nx_cugraph/algorithms/centrality/eigenvector.py
eriknw: Oh, right, I guess I don't need to normalize `nstart`. I can comment this line out (but not delete it), since it's doing unnecessary work.

We handle `nstart` as if we are going to use it to prepare for when we will actually use it. If we ignore it completely, then users can write code that works today, but will fail for newer versions or when running with a different backend. Simply put: it's better to raise when raising is the correct behavior. I don't think we should raise `NotImplementedError`, b/c then  e.g. networkx will raise, so why convert to another backend or nx if we know it's going to raise an error?

I expect operating on node vectors like this to be super fast. This check is cheap.

python/nx-cugraph/nx_cugraph/algorithms/centrality/katz.py
eriknw: It's run when going through the networkx dispatch _before_ input data is converted. We shouldn't make assumptions about the input arguments, and we shouldn't convert the graphs in this function.

Raising `NotImplementedError` when `can_run` will return False is indeed redundant. It is included in case the function is called directly and not via dispatching.

> Also, is there a reason this isn't checking nstart and returning False if it's set? Same question for eigenvector_centrality. I think I'm not understanding the difference between "unused" and "unsupported" from the user's perspective.

I consider `nstart` to be an optimization. Generally speaking, it shouldn't change the answer, but it may change the number of iterations required. So, these algorithms run just fine w/o `nstart`, hence why `can_run` still returns True if nstart is given.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/hits_alg.py
eriknw: Yes, this functions appears to work fine with only `initial_hubs_guess_values` provided.

python/nx-cugraph/nx_cugraph/algorithms/centrality/katz.py
eriknw: Contrast this with `dangling` in pagerank, which _does_ generally change the result.

python/nx-cugraph/nx_cugraph/algorithms/link_analysis/hits_alg.py
jnke2016: So I just took a look at the HITS CAPI and the reason it is not failing is simply because both the initial_hubs_guess vertices and values are ignored when the former is [None](https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/src/c_api/hits.cpp#L116)

python/nx-cugraph/nx_cugraph/interface.py
```diff
@@ -67,7 +67,14 @@ def key(testpath):
...
         no_string_dtype = "string edge values not currently supported"
 
-        xfail = {}
+        xfail = {
+            key(
+                "test_eigenvector_centrality.py:TestEigenvectorCentrality.test_P3"
+            ): "Power iteration failed (don't know why; networkx converges in 8 steps)",
+            key(
+                "test_hits.py:TestHITS.test_hits_not_convergent"
+            ): "HITS doesn't raise after failing to converge after max_iter iterations",
```
eriknw, line 76: @rlratzel, I may not be able to test this PR with https://github.com/rapidsai/cugraph/pull/3979 before my PTO. Hopefully, #3979 will allow us to delete these xfail cases. CI will guide you, and delete things here as necessary.

rlratzel: I was able to remove the `eigenvector_centrality` xfail, but we still need the `hits` xfail for NX versions <= 3.2.1 that don't have [your NX update](https://github.com/networkx/networkx/pull/7084).

python/nx-cugraph/nx_cugraph/algorithms/centrality/eigenvector.py
rlratzel: Right, I'm not worried about if the check is expensive as much as I am about meeting expectations. I see you updated the docstring on this arg so IMO that's sufficient for now.

> then users can write code that works today, but will fail for newer versions or when running with a different backend

I think this is the real problem we'll have to deal with: how do we support drop-in compatibility when NX and different backends have varying levels of support for different args?  I suspect the approach we're taking here won't scale (e.g. if NX adds an arg or changes what it considers valid for an arg, will backends be expected to track these changes and update their code?), but that's outside the scope of this PR so we may have to have a bigger discussion later. cc @mriduls

python/nx-cugraph/nx_cugraph/algorithms/centrality/katz.py
rlratzel: I see what you mean re: optimizations vs. args that change results, so for this check ("can this run at all or not?") this makes sense.

rlratzel (approved):

PR #3976, opened 2023-11-03
nx-cugraph: add SSSP (unweighted)

eriknw:
There are many more traversal algorithms to implement, but these get us started!


eriknw:
The errors are interesting:
> RuntimeError: non-success value returned from cugraph_bfs: CUGRAPH_UNKNOWN_ERROR std::bad_alloc: out_of_memory: CUDA error at:

We are calling plc.bfs repeatedly as part of `all_pairs_shortest_path_length`, but shouldn't this work still? Shouldn't memory be cleaned up? All results are turned into Python dicts and they shouldn't be on device anymore.

Heh, so it looks like `single_source_shortest_path_length` is a good example of a networkx algorithm that other networkx algorithms call in a loop. It's used by (at least) these:
- `all_pairs_shortest_path_length`
- `closeness_centrality`
- `bipartite.closeness_centrality`
- `incremental_closeness_centrality`
- `average_shortest_path_length`

BradReesWork (approved):

rlratzel (approved):

PR #3978, opened 2023-11-06
[BUG] Fix Incorrect Edge Index, Directory Selection in cuGraph-PyG Loader

### LINKED ISSUE #3959: [BUG] cuGraph-PyG BulkSampleLoader Doesn't Handle Directories Correctly
Currently, the `BulkSampleLoader` incorrectly creates a new temporary directory instead of directly writing the files to the path given.  This makes it nearly impossible to use it within multi-GPU offline workflows since they require that all files be written to the same top-level directory.

alexbarghi-nv:
Fixes three major bugs:
1. Edge index is set to [dst, dst] instead of [dst, src] in some cases
2. The sample directory is always set to a new temporary directory rather than the path given
3. The version of `pylibcugraphops` in `meta.yaml` is wrong and causes the wrong packages to be resolved

This PR also simplifies `ci/test_python.sh` by doing only a single conda install when creating the `test_cugraph_pyg` environment.

Closes #3959


python/cugraph-pyg/cugraph_pyg/loader/cugraph_node_loader.py
```diff
@@ -437,11 +455,10 @@ def __next__(self):
...
         if self.__coo and self.__graph_store.order == "CSC":
-            for node_type in out.edge_index_dict:
-                out[node_type].edge_index[0], out[node_type].edge_index[1] = (
-                    out[node_type].edge_index[1],
-                    out[node_type].edge_index[0],
-                )
+            for edge_type in out.edge_index_dict:
+                src = out[edge_type].edge_index[0]
+                dst = out[edge_type].edge_index[1]
+                out[edge_type].edge_index = torch.stack([dst, src])
```
tingyu66, line 461:
```suggestion
                out[edge_type].edge_index = out[edge_type].edge_index.flip(dims=[0])
```

python/cugraph-pyg/cugraph_pyg/tests/test_cugraph_loader.py
```diff
@@ -40,7 +43,9 @@
 
 
 @pytest.mark.skipif(isinstance(torch, MissingModule), reason="torch not available")
-def test_cugraph_loader_basic(karate_gnn):
+def test_cugraph_loader_basic(
+    karate_gnn: Tuple[FeatureStore, Dict[str, np.ndarray], Dict[str, int]]
```
tingyu66, line 47: The type of G (the 2nd return value) should be `Dict[Tuple[str, str, str], ......] ` to be exact. I think the CuGraphStore class has the same typing issue.

tingyu66 (commented):
Left some comments. LGTM otherwise.

python/cugraph-pyg/cugraph_pyg/tests/test_cugraph_loader.py
alexbarghi-nv: fixed

BradReesWork (approved):

raydouglass (approved):

PR #3980, opened 2023-11-06
[BUG] Restore the original default order of CSR, which does not reverse edges in cuGraph-PyG

### LINKED ISSUE #3969: [BUG] CSC Order Causes Issues with OGB Datasets
While CSC order is technically correct, it does not match the order of the ogb datasets, which specify CSR order.  This has caused numerous issues with the examples based on OGB datasets, which are also the templates for most user datasets.  To avoid user confusion, the default order should be reverted to CSR.

alexbarghi-nv:
OGB and many other datasets express the edge index in CSR order, but in release 23.10, the default order was changed to CSC, which broke downstream workflows.  This PR restores the original default, which should be acceptable for the most common datasets, and familiar to most PyG users.

Closes #3969

### DIFF
--- a/python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
+++ b/python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
@@ -214,7 +214,7 @@ class EXPERIMENTAL__CuGraphStore:
         num_nodes_dict: Dict[str, int],
         *,
         multi_gpu: bool = False,
-        order: str = "CSC",
+        order: str = "CSR",
     ):
         """
         Constructs a new CuGraphStore from the provided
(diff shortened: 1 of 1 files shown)


tingyu66 (approved):
LGTM.

Just to clarify, "CSR" in this context translates into: using destination nodes (in the COO) as seeds during sampling, right?

alexbarghi-nv:
> LGTM.
> 
> Just to clarify, "CSR" in this context translates into: using destination nodes (in the COO) as seeds during sampling, right?

Yes

rlratzel (approved):
The change looks reasonable to me (disclaimer: I'm not familiar with this code), but I'm surprised there aren't any tests that need to be changed too.

PR #3982, opened 2023-11-07
Update C API graph creation function signatures

### LINKED ISSUE #3947: Update PLC, C, (and C++?) APIs to better accommodate python MG graph creation use cases
There are three separate requests here. I can break these up into three separate issues if that's better:

### Do not require `num_edges`

The python/dask code currently has to compute/persist the dask_cudf dataframe in order to compute the total number of edges the PLC API requires for creating a graph.  This should not be necessary since each array already has the size known internally.  Not requiring `num_edges` would eliminate the need to do the somewhat expensive dask compute/persist call.

_NOTE: adding this feature is related to the fix done for [this issue](https://github.com/rapidsai/cugraph/issues/3790)_

### Allow multiple src/dst arrays

In order to ease the burden on the cugraph dask/python code for handling dask_cudf input with multiple partitions per worker, the PLC, C, and possibly C++ APIs could accept multiple `src` and `dst` vertex arrays.  This would allow the dask/python layer to pass the src/dst arrays from each partition as-is, instead of combining each partition's arrays in python in order to pass them as a single src and single dst array to PLC/C.

### Implement a "move" option to transfer array ownership

Another improvement could be to allow the PLC/C/C++ layers to own the src/dst arrays currently maintained in python.  This would allow PLC/C/C++ modify and delete the incoming arrays as needed, instead of requiring a copy step to preserve the arrays owned by the user/python layer.  This could be a new option, possibly called "move" which would default to `False` (the current behavior).

### LINKED ISSUE #3974: [BUG]: PLC inconsistent handling of isolated nodes; example using WCC
### Version

23.12 dev

Source

### Describe the bug.

PLC sometimes handles isolated nodes inconsistently, and we are forced to handle the edge cases in Python (if possible).

I will illustrate this with `plc.weakly_connected_components`, but the same behavior has been seen elsewhere such as with Louvain.

Let's consider an undirected graph with edge indices `[1, 2]` and `[2, 1]` in COO format. There is currently no way to tell PLC how many nodes are in this graph, and it _assumes_ the number of nodes is `max(indices) + 1`. Here's the first example:
```python
import cupy as cp
import numpy as np
import pylibcugraph as plc

src_indices = cp.array([1, 2], np.int32)
dst_indices = cp.array([2, 1], np.int32)

G = plc.SGGraph(
    graph_properties=plc.GraphProperties(
        is_multigraph=False,
        is_symmetric=True,
    ),
    src_or_offset_array=src_indices,
    dst_or_index_array=dst_indices,
    input_array_format="COO",
    weight_array=None,
    store_transposed=False,
    renumber=False,
    do_expensive_check=False,
)
node_ids, labels = plc.weakly_connected_components(
    graph=G,
    offsets=None,
    indices=None,
    weights=None,
    labels=None,
    do_expensive_check=False,
)
print(node_ids)
print(labels)
```
which outputs
```
[0 1 2]
[0 2 2]
```
This is almost perfect, right? Because we gave `renumber=False` when creating `G`, it assumes nodes are 0, 1, and 2 (b/c of max indices). But, what if we had four (or more) nodes in the graph and node 3 (and above) have no edges? Then we have no results for them, and we have no way to tell PLC to create results for them.

What if instead of COO we used CSR? One might expect CSR to work since the number of nodes is implied by the length of the offsets array. With WCC, we have two ways to use CSR: when creating the graph, or when calling WCC.
```python
    # Change creation of G to use this
    src_or_offset_array=cp.array([0, 0, 1, 2, 2], np.int32),
    dst_or_index_array=cp.array([2, 1], np.int32),
    input_array_format="CSR",
```
```python
    # Or call plc.weakly_connected_components with this
    graph=None,
    offsets=cp.array([0, 0, 1, 2, 2], np.int32),
    indices=cp.array([2, 1], np.int32),
```
Both approaches give the same result as before (only nodes 0, 1, 2).

It seems the fundamental issue is that COO contains incomplete information about the graph. If we are not renumbering, then I want to tell PLC the number of nodes in the graph. If we _are_ renumbering, then I want to give it an array of nodes. Then I want results for each node.

Is this something that should be handled in the PLC/C layers?

```shell
See above
```

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

### Code of Conduct

ChuckHastings:
Updating the C API graph creation functions to support the following:
* Add support for isolated vertices
* Add MG optimization to support multiple device arrays per rank as input and concatenate them internally
* Add MG optimization to internally compute the number of edges via allreduce rather than requiring it as an input parameter (this can be expensive to compute in python)

This PR implements these features.  Some simple tests exist to check for isolate vertices (by running pagerank which generates a different result if the graph has isolated vertices).  A simple test for multiple input arrays exists for the MG case.

Closes #3947 
Closes #3974


cpp/include/cugraph_c/graph.h
```diff
@@ -76,9 +77,55 @@ cugraph_error_code_t cugraph_sg_graph_create(
...
   cugraph_error_t** error);
 
+/**
+ * @brief     Construct an SG graph
+ *
+ * @param [in]  handle         Handle for accessing resources
+ * @param [in]  properties     Properties of the constructed graph
+ * @param [in]  vertices       Optional device array containing a list of vertex ids
+ *                             (specify NULL if we should create vertex ids from the
+ *                             unique contents of @p src and @p dst)
```
eriknw, line 87: How does NULL vertices behave with and w/o renumbering? What if edge data only contain unique ids of e.g. 1, 3, and 5? Are 0, 2, and 4 implied if renumber is False?

cpp/include/cugraph_c/graph.h
```diff
@@ -76,9 +77,55 @@ cugraph_error_code_t cugraph_sg_graph_create(
...
+                               argument that can be NULL if edge ids are not used.
+ * @param [in]  edge_type_ids  Device array containing the edge types for each edge.  Optional
+                               argument that can be NULL if edge types are not used.
+ * @param [in]  store_transposed If true create the graph initially in transposed format
+ * @param [in]  renumber       If true, renumber vertices to make an efficient data structure.
+ *    If false, do not renumber.  Renumbering is required if the vertices are not sequential
+ *    integer values from 0 to num_vertices.
+ * @param [in]  do_expensive_check    If true, do expensive checks to validate the input data
+ *    is consistent with software assumptions.  If false bypass these checks.
+ * @param [in]  properties     Properties of the graph
```
eriknw, line 102: properties is defined twice for this function (and also twice for the function below).

cpp/include/cugraph_c/graph.h
```diff
@@ -116,19 +163,67 @@ cugraph_error_code_t cugraph_sg_graph_create_from_csr(
...
+ * @param [in]  offsets        Device array containing the CSR offsets array
+ * @param [in]  indices        Device array containing the destination vertex ids
+ * @param [in]  weights        Device array containing the edge weights.  Note that an unweighted
+ *                             graph can be created by passing weights == NULL.
+ * @param [in]  edge_ids       Device array containing the edge ids for each edge.  Optional
+                               argument that can be NULL if edge ids are not used.
+ * @param [in]  edge_type_ids  Device array containing the edge types for each edge.  Optional
+                               argument that can be NULL if edge types are not used.
+ * @param [in]  store_transposed If true create the graph initially in transposed format
+ * @param [in]  renumber       If true, renumber vertices to make an efficient data structure.
```
eriknw, line 180: How does `renumber=True` work for CSR? Wouldn't `offsets` already imply 0..N-1 vertex ids?

cpp/include/cugraph_c/graph.h
ChuckHastings: If `vertices` is NULL, this will provide the same behavior as today.  That is:
* Renumber enabled will ignore vertices 0, 2, 4... they will not exist in the graph.  We will create a graph with 3 vertices which will be internally numbered 0, 1, 2.  Any function that is called will translate any vertex id inputs to the mapping 0, 1, 2 (e.g. if you call BFS and specify a seed of 1, it will be mapped to whichever id that vertex 1 was mapped to), and any vertex ids outputs will be renumbered from the internal mapping (0, 1, 2) to the external mapping (1, 3, 5)
* Renumber disabled will use the vertex ids exactly as specified, so vertices 0, 2 and 4 will be isolated vertices in the graph.  However, if your graph also had a vertex 6... that vertex would be lost even with renumber disabled, since there is no mechanism to discover that vertex 6 exists in your graph with the current (and proposed API).

cpp/include/cugraph_c/graph.h
ChuckHastings: `renumber=True` would still work for CSR input.  Yes, `offsets` does imply vertex ids in the range `[0, N)`.  But our renumbering logic adds some performance optimizations by putting the vertex ids in an order that makes the CUDA kernels more efficient.  So there is value in specifying `renumber=True` even if the vertex ids are densely packed.

I will be fixing the C API implementation of this function so that it will properly reflect the isolated vertices that can be implied in CSR format.

cpp/include/cugraph_c/graph.h
```diff
@@ -165,13 +256,66 @@ cugraph_error_code_t cugraph_mg_graph_create(
...
+/**
+ * @brief     Construct an MG graph
+ *
+ * @param [in]  handle          Handle for accessing resources
+ * @param [in]  properties      Properties of the constructed graph
+ * @param [in]  vertices        List of device arrays containing the unique vertex ids.
+ *                              If NULL we will construct this internally using the unique
+ *                              entries specified in src and dst
+ *                              All entries in this list will be concatenated on this GPU
+ *                              into a single array.
```
seunghwak, line 268: What happens if vertices are explicitly provided but for this GPU, there is no vertex ID? This caused a bug in our C++ API (this is no longer a problem as we use std::optional). Can we handle this in the C API?

cpp/include/cugraph_c/graph.h
```diff
@@ -116,19 +160,66 @@ cugraph_error_code_t cugraph_sg_graph_create_from_csr(
...
+  bool_t do_expensive_check,
+  cugraph_graph_t** graph,
+  cugraph_error_t** error);
+
+/**
+ * @brief     Destroy an graph
+ *
+ * @param [in]  graph  A pointer to the graph object to destroy
+ */
+void cugraph_graph_free(cugraph_graph_t* graph);
```
seunghwak, line 207: So, this function works for both SG&MG graphs, right? Then, should we place this function after SG & MG graph creation functions? I was searching for a new graph deletion function for MG (replacing the deprecated function) and it took sometime to find that this function works for both SG & MG.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -96,49 +135,27 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
         edge_type_id_t>>
         new_edge_types{std::nullopt};
 
-      rmm::device_uvector<vertex_t> edgelist_srcs(src_->size_, handle_.get_stream());
-      rmm::device_uvector<vertex_t> edgelist_dsts(dst_->size_, handle_.get_stream());
```
seunghwak, line 100: Maybe something outside the scope of this PR, but are we planning to shuffle edges in the C-API in the future? If we perform shuffle in the C-API, should we still need to take multiple arrays from the python layer?

cpp/src/c_api/graph_mg.cpp
```diff
@@ -153,6 +170,11 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
           std::move(edgelist_edge_ids),
           std::move(edgelist_edge_types));
 
+      if (vertex_list) {
+        vertex_list = cugraph::detail::shuffle_ext_vertices_to_local_gpu_by_vertex_partitioning(
+          handle_, std::move(*vertex_list));
+      }
```
seunghwak, line 176: If we are assuming that edges are pre-shuffled, shouldn't we also assume that vertices are also pre-shuffled? I guess we'd better be consistent.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -298,87 +267,176 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
...
+  }
+
+  size_t num_edges = cugraph::host_scalar_allreduce(p_handle->handle_->get_comms(),
+                                                    local_num_edges,
+                                                    raft::comms::op_t::SUM,
+                                                    p_handle->handle_->get_stream());
+
+  // FIXME:  Need to handle the case where a GPU gets a NULL pointer but other GPUs
+  //   Get values.  Override vertex_type/weight_type/edge_id_type but don't
+  //   fail.
```
seunghwak, line 339: I think we should handle this instead of leaving this as FIXME. This indeed happened in Louvain in the past.

cpp/src/c_api/graph_mg.cpp
ChuckHastings: We do shuffle the edges in the C API currently.

Perhaps @VibhuJawa can offer some insight into the need for multiple arrays, but I believe this is an artifact of dask and how the arrays are distributed on the workers.  It is possible for there to be multiple device arrays on the same worker.  This new mechanism allows the concatenation of these multiple arrays to occur in the C API rather than in dask/cudf.

cpp/src/c_api/graph_mg.cpp
ChuckHastings: We are not assuming things are reshuffled.

cpp/src/c_api/graph_mg.cpp
ChuckHastings: I will address this as an update to this PR.  I wanted to get the PR up for review rather than waiting for a little more validation.

cpp/include/cugraph_c/graph.h
ChuckHastings: I can move this.

cpp/src/c_api/graph_mg.cpp
seunghwak: Oh I see, I missed the cugraph::detail::shuffle_ext_vertex_pairs_with_values_to_local_gpu_by_edge_partitioning() call.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -298,87 +297,178 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
...
+  auto vertex_types = cugraph::host_scalar_allgather(
+    p_handle->handle_->get_comms(), static_cast<int>(vertex_type), p_handle->handle_->get_stream());
+
+  auto weight_types = cugraph::host_scalar_allgather(
+    p_handle->handle_->get_comms(), static_cast<int>(weight_type), p_handle->handle_->get_stream());
+
+  CAPI_EXPECTS(
+    std::count_if(vertex_types.begin(),
+                  vertex_types.end(),
+                  [vertex_type](auto t) { return vertex_type != static_cast<int>(t); }) == 0,
```
seunghwak, line 380: A minor point, but is `std::all_of` more specific for this? (https://en.cppreference.com/w/cpp/algorithm/all_any_none_of). I guess all_of may better document the intention.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -298,87 +297,178 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
...
+    "different vertex type used on different GPUs",
+    *error);
+
+  CAPI_EXPECTS(
+    std::count_if(weight_types.begin(),
+                  weight_types.end(),
+                  [weight_type](auto t) { return weight_type != static_cast<int>(t); }) == 0,
+    CUGRAPH_INVALID_INPUT,
+    "different weight type used on different GPUs",
+    *error);
```
seunghwak, line 391: A minor point, but is `std::all_of` more specific for this? (https://en.cppreference.com/w/cpp/algorithm/all_any_none_of). I guess all_of may better document the intention.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -160,6 +181,40 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+      if (drop_self_loops_) {
+        std::tie(
+          edgelist_srcs, edgelist_dsts, edgelist_weights, edgelist_edge_ids, edgelist_edge_types) =
+          cugraph::remove_self_loops(handle_,
+                                     std::move(edgelist_srcs),
+                                     std::move(edgelist_dsts),
+                                     std::move(edgelist_weights),
+                                     std::move(edgelist_edge_ids),
+                                     std::move(edgelist_edge_types));
+      }
```
seunghwak, line 216: Sorry for nitpicking but if `drop_self_loops` appear before `drop_multi_edges` in the input parameters, should we better perform drop_self_loops() before drop_multi_edges()? This consistency may make the code easier to read and slightly easier to maintain.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
```diff
@@ -0,0 +1,196 @@
...
+namespace detail {
+
+template <typename vertex_t>
+void sort_and_remove_multi_edges(raft::handle_t const& handle,
+                                 rmm::device_uvector<vertex_t>& edgelist_srcs /* [INOUT] */,
+                                 rmm::device_uvector<vertex_t>& edgelist_dsts /* [INOUT] */)
+{
+  auto edge_first =
+    thrust::make_zip_iterator(thrust::make_tuple(edgelist_srcs.begin(), edgelist_dsts.begin()));
+  thrust::sort(handle.get_thrust_policy(), edge_first, edge_first + edgelist_srcs.size());
```
seunghwak, line 41: This will require 2x the size of edge list and will increase the peak memory requirement. Our goal is to keep the peak memory footprint in graph creation to 1.5x edge list + per vertex data.

See the code in

https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/src/structure/detail/structure_utils.cuh#L164

We can call mem_frugal_partitoin (which requires 0.5x the input array) and process the resulting two partitions one by one if the input array size is larger than the certain threshold.

cpp/src/structure/remove_self_loops_impl.cuh
```diff
@@ -0,0 +1,180 @@
...
+{
+  auto edge_first =
+    thrust::make_zip_iterator(thrust::make_tuple(edgelist_srcs.begin(), edgelist_dsts.begin()));
+  edgelist_srcs.resize(
+    thrust::distance(
+      edge_first,
+      thrust::remove_if(handle.get_thrust_policy(),
+                        edge_first,
+                        edge_first + edgelist_srcs.size(),
+                        [] __device__(auto e) { return thrust::get<0>(e) == thrust::get<1>(e); })),
```
seunghwak, line 46: I am 70% sure thrust::remove_if requires 2x the input array size memory. If this becomes a memory bottleneck (e.g. the input edge list size is large compared to the device memory size, see https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/src/structure/create_graph_from_edgelist_impl.cuh#L523C23-L523C23), we may create a bitmap storing whether to remove an edge or not. And based on this bitmap, we can call remove_if on srcs, dsts, weights, edge_ids, and edge_types, separately. Then, we can keep the peak memory requirement under our goal.

cpp/src/structure/remove_self_loops_impl.cuh
```diff
@@ -0,0 +1,180 @@
...
+    handle.get_stream());
+  edgelist_dsts.resize(edgelist_srcs.size(), handle.get_stream());
+}
+
+template <typename vertex_t, typename A>
+void remove_self_loops(raft::handle_t const& handle,
+                       rmm::device_uvector<vertex_t>& edgelist_srcs /* [INOUT] */,
+                       rmm::device_uvector<vertex_t>& edgelist_dsts /* [INOUT] */,
+                       rmm::device_uvector<A>& edgelist_a /* [INOUT] */)
```
seunghwak, line 55: I am not sure we need three separate functions that takes 1, 2, or 3 additional variables in addition to edgelist_srcs & edgelist_dsts.

See https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/src/structure/detail/structure_utils.cuh#L148

Then, we can set

edge_first = thrust::make_zip_iterator(edgelist_srcs.begin(), edgelist_dsts.begin(), get_dataframe_buffer_begin());

And the remaining code will be same. Same for the sort_and_remove_multi_edges function.

This will be more important if we bring additional code to limit peak memory footprint.

cpp/src/c_api/graph_mg.cpp
VibhuJawa: We can do it in `dask-cudf` too once we remove the  `num_edges`  signature (which this PR does). 

https://github.com/rapidsai/cugraph/blob/8c104a52bbece783f91001c64cc6bf0732f10594/python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py#L134

If we don't have to get  `num_edges` , we can just do this in `cuDF/cuPY` one array at a time and the resulting overhead will be  `2x(array_memory)` instead of `2x(dataframe_memory)` .

cpp/src/structure/remove_self_loops_impl.cuh
ChuckHastings: That would also eliminate the need for the ugliness you mention below.  I'll refactor.

cpp/src/c_api/graph_sg.cpp
ChuckHastings: Will address in next push

cpp/src/c_api/graph_mg.cpp
ChuckHastings: Will address when I get to the FIXME, which might change this logic anyway.

ChuckHastings: Fixed in next push.

cpp/src/c_api/graph_mg.cpp
ChuckHastings: Fixed in next push

cpp/include/cugraph_c/graph.h
```diff
@@ -117,18 +172,50 @@ cugraph_error_code_t cugraph_sg_graph_create_from_csr(
...
+ * @param [out] graph          A pointer to the graph object
+ * @param [out] error          Pointer to an error object storing details of any error.  Will
+ *                             be populated if error code is not CUGRAPH_SUCCESS
+ *
+ * @return error code
  */
-// FIXME:  This should probably just be cugraph_graph_free
-//         but didn't want to confuse with original cugraph_free_graph
-void cugraph_sg_graph_free(cugraph_graph_t* graph);
+cugraph_error_code_t cugraph_graph_create_sg_from_csr(
```
jnke2016, line 200: Since `cugraph_graph_create_sg` supports flags to drop both self loops and multi edges through the methods `cugraph::remove_self_loops` and `cugraph::sort_and_remove_multi_edges`, shouldn't `cugraph_graph_create_sg_from_csr` also support those flags? In fact, `cugraph_graph_create_sg_from_csr` extracts the `COO` from the `CSR` before calling `cugraph::create_graph_from_edgelist` therefore, we can drop self loops and multi edges (if set to True) right before the graph question.

ChuckHastings: We could add that feature.  I was assuming that if you already have a CSR as input then you have already done whatever ETL steps are desired.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -298,87 +297,198 @@ extern "C" cugraph_error_code_t cugraph_mg_graph_create(
...
-               CUGRAPH_INVALID_INPUT,
-               "Invalid input arguments: src size != dst size.",
-               *error);
-  CAPI_EXPECTS(p_src->type_ == p_dst->type_,
+    CAPI_EXPECTS(p_src[i]->type_ == vertex_type,
+                 CUGRAPH_INVALID_INPUT,
+                 "Invalid input arguments: all vertex types must match",
+                 *error);
+
+    CAPI_EXPECTS(p_weights[i]->type_ == weight_type,
```
jnke2016, line 362: You forgot to check if `weights` was not a `nullptr` otherwise we get a `segfault`. I changed it locally to this

```
CAPI_EXPECTS((weights == nullptr) || (p_weights[i]->type_ == weight_type),
                 CUGRAPH_INVALID_INPUT,
                 "Invalid input arguments: all weight types must match",
                 *error);
```

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -973,4 +973,69 @@ renumber_sampled_edgelist(
...
+ * @return Tuple of vectors storing edge sources, destinations, optional weights,
+ *    optional edge ids, optional edge types.
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, typename edge_type_t>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<weight_t>>,
+           std::optional<rmm::device_uvector<edge_t>>,
+           std::optional<rmm::device_uvector<edge_type_t>>>
+sort_and_remove_multi_edges(raft::handle_t const& handle,
```
seunghwak, line 1034: I just realized this but now it seems like "sort" here is a misnomer. We are not really globally sorting anymore, so I think we'd better rename this function to `remove_multi_edges`.

seunghwak: And may better document that this function is not stable (which means this does not preserve the order in contrast to std::remove or thrust::remove).

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -973,4 +973,69 @@ renumber_sampled_edgelist(
...
+           std::optional<rmm::device_uvector<edge_type_t>>>
+remove_self_loops(raft::handle_t const& handle,
+                  rmm::device_uvector<vertex_t>&& edgelist_srcs,
+                  rmm::device_uvector<vertex_t>&& edgelist_dsts,
+                  std::optional<rmm::device_uvector<weight_t>>&& edgelist_weights,
+                  std::optional<rmm::device_uvector<edge_t>>&& edgelist_edge_ids,
+                  std::optional<rmm::device_uvector<edge_type_t>>&& edgelist_edge_types);
+
+/**
+ * @brief Sort the edges and remove all but one edge when a multi-edge exists
```
seunghwak, line 1008: I think this comment is no longer correct.

cpp/src/structure/detail/structure_utils.cuh
```diff
@@ -496,6 +498,62 @@ void sort_adjacency_list(raft::handle_t const& handle,
   }
 }
 
+template <typename T>
+struct indirect_array_reference {
+  T const* array_;
+
+  T operator() __host__ __device__(size_t index) { return array_[index]; }
+};
```
seunghwak, line 506: Can't we use 
```
template <typename index_t, typename Iterator>
struct indirection_t {
  Iterator first{};

  __device__ typename thrust::iterator_traits<Iterator>::value_type operator()(index_t i) const
  {
    return *(first + i);
  }
};
```
instead?

https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/include/cugraph/utilities/device_functors.cuh#L70

cpp/src/structure/detail/structure_utils.cuh
```diff
@@ -496,6 +498,62 @@ void sort_adjacency_list(raft::handle_t const& handle,
...
+    handle.get_thrust_policy(),
+    thrust::make_counting_iterator(size_t{0}),
+    thrust::make_counting_iterator(src.size()),
+    [comparison, d_remove_flags = remove_flags.data()] __device__(size_t i) {
+      if (comparison(i)) {
+        atomicOr(d_remove_flags + cugraph::packed_bool_offset(i), cugraph::packed_bool_mask(i));
+        return true;
+      }
+      return false;
+    });
```
seunghwak, line 531: Should we better call this `remove_flags.size()`? We can avoid atomics and no need to call thrust::fill.

Something like
```
thrust::tabulate(
  handle.get_thrust_policy(),
  remove_flags.begin(),
  remove_flags.end(),
  [edge_first, comparison]__device__(auto i) {
    auto word = packed_bool_empty_mask();
    ...
    return word;
  });
```

cpp/src/structure/detail/structure_utils.cuh
```diff
@@ -496,6 +498,62 @@ void sort_adjacency_list(raft::handle_t const& handle,
...
+
+  return std::make_tuple(remove_count, std::move(remove_flags));
+}
+
+template <typename T>
+rmm::device_uvector<T> remove_flagged_elements(raft::handle_t const& handle,
+                                               rmm::device_uvector<T>&& vector,
+                                               rmm::device_uvector<uint32_t> const& remove_flags,
```
seunghwak, line 539: Better use raft::device_span for consistency?

cpp/src/structure/sort_and_remove_multi_edges.cu
seunghwak, line 1: Better remove `sort_` from the file name.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
seunghwak, line 1: Better remove `sort_` from the file name.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
```diff
@@ -0,0 +1,317 @@
...
+                                            thrust::lower_bound(handle.get_thrust_policy(),
+                                                                hash_counts.begin(),
+                                                                hash_counts.end(),
+                                                                edgelist_srcs.size() / 2)));
+
+    auto second_first = detail::mem_frugal_partition(pair_first,
+                                                     pair_first + edgelist_srcs.size(),
+                                                     hash_src_dst_pair<vertex_t>{},
+                                                     pivot,
+                                                     handle.get_stream());
```
seunghwak, line 109: You can use https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/include/cugraph/utilities/shuffle_comm.cuh#L782 with num_groups = 2. Then you just need to provide a `key_to_group_id` functor.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
```diff
@@ -0,0 +1,317 @@
...
+                                                                hash_counts.begin(),
+                                                                hash_counts.end(),
+                                                                edgelist_srcs.size() / 2)));
+
+    auto second_first = detail::mem_frugal_partition(pair_first,
+                                                     pair_first + edgelist_srcs.size(),
+                                                     get_dataframe_buffer_begin(edgelist_values),
+                                                     hash_src_dst_pair<vertex_t>{},
+                                                     pivot,
+                                                     handle.get_stream());
```
seunghwak, line 167: Same here, you can call https://github.com/rapidsai/cugraph/blob/branch-23.12/cpp/include/cugraph/utilities/shuffle_comm.cuh#L782

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
```diff
@@ -0,0 +1,317 @@
...
+  edgelist_srcs.resize(
+    thrust::distance(pair_first,
+                     thrust::unique(handle.get_thrust_policy(),
+                                    pair_first,
+                                    pair_first + edgelist_srcs.size(),
+                                    [] __device__(auto lhs, auto rhs) {
+                                      return (thrust::get<0>(lhs) == thrust::get<0>(rhs)) &&
+                                             (thrust::get<1>(lhs) == thrust::get<1>(rhs));
+                                    })),
+    handle.get_stream());
```
seunghwak, line 125: Have you checked thrust::unique does not use a temporary memory buffer (size = original data size). We can test this by creating an array with random data close to 70% of the total memory size and this unique call succeeds.

naimnv (approved):

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
ChuckHastings: My concern with setting num_groups = 2 was that we might get a pretty bad skew in the multi-edges.  Consider an rmat graph.  We're going to get certain edges that occur with great frequency, and other edges that occur with lesser frequency.  I thought that by first grouping by some larger number and then finding the median break point we'll be closer to a 50/50 split at the pivot.

I certainly could use `groupby_and_count` instead of recreating it here (my `compute_hash_sizes` function is essentially the same.

cpp/src/structure/detail/structure_utils.cuh
ChuckHastings: I can rework this.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
ChuckHastings: Good catch.  I'll run that test.  I have an idea to work around that.  I'll work on it over the weekend.

cpp/src/structure/sort_and_remove_multi_edges_impl.cuh
seunghwak: You mean there can be millions (or billions) of edges between a unique (src, dst) pair? Yeah... this might be possible if there are two really big institutes and there are a huge number of transactions between the two institutes.

But I don't think this will happen in synthetic R-mat graphs (we are not partitioning based on just src or dst but src, dst pair, so high degree vertices won't have much impact on partitioning quality).

Yeah... and calling with num_groups = 2 vs some large number can be a compute vs solution robustness trade-offs. num_goups = 2 will be cheaper than than num_groups = 128 (2^1 vs 2^7). And in the theoretical worst case (i.e. all the edges are between the same src, dst pair), we will need 2x memory of the edge list vs 1.5x. And if there are very high multiplicity edge pairs, we may be sometimes able to combine those edges and create a weighted graph. Just a thought material. Not sure what's the right trade-off point.

jnke2016 (approved):
Looks good to me

seunghwak (approved):
LGTM

PR #3985, opened 2023-11-07
[BUG] Fix Graph Construction From Pandas in cuGraph-PyG

alexbarghi-nv:
The current graph construction creates a single pandas dataframe, which for larger datasets (i.e. ogbn-papers100M) cannot be serialized.  This PR resolves this by breaking up the dataframe into scattered numpy arrays that are then reassembled.

Merge after #3978


VibhuJawa (approved):
Reviewed the dask changes and they look good to me, will review other bits after the 

https://github.com/rapidsai/cugraph/pull/3978 is merged in

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -367,6 +368,15 @@ def __infer_offsets(
             }
         )
 
+    def __dask_array_from_numpy(
+        self, array: np.ndarray, client: distributed.client.Client, npartitions: int
```
tingyu66, line 372: Do we need `client` argument here as it's not used explicitly?

alexbarghi-nv: no, I forgot to take that out after updating the function.

alexbarghi-nv: should be fixed now

BradReesWork (approved):

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
```diff
@@ -464,32 +472,42 @@ def __construct_graph(
...
+            df = dd.from_dask_array(etp_dar, columns=["etp"])
+            df["src"] = dst_dar if order == "CSC" else src_dar
+            df["dst"] = src_dar if order == "CSC" else dst_dar
+
+            del src_dar
+            del dst_dar
+            del etp_dar
+
+            if df.etp.dtype != "int32":
+                raise ValueError("Edge type must be int32!")
```
tingyu66, line 500: Do we need the same check in the `else` branch?

tingyu66 (approved):
LGTM. Only have a minor question.

python/cugraph-pyg/cugraph_pyg/data/cugraph_store.py
alexbarghi-nv: No, we set the datatype for edge type to int32 on line 464.  That check is to catch the error before it starts a long dask job.

PR #3987, opened 2023-11-07
Remove CMake dependencies to UCP

pentschev:
Apparently UCX/UCP isn't directly used anymore in cuGraph's C++ implementation but the UCP getter is still there which shouldn't be required.


pentschev:
Alright, I see now that cuGraph has to satisfy RAFT's dependency to UCP via its CMake, so this doesn't apply. Closing.

PR #3992, opened 2023-11-09
Add ucx to libcugraph recipe

naimnv:
Add ucx to libcugraph recipe

### DIFF
--- a/conda/recipes/libcugraph/meta.yaml
+++ b/conda/recipes/libcugraph/meta.yaml
@@ -77,7 +77,7 @@ requirements:
     - librmm ={{ minor_version }}
     - nccl {{ nccl_version }}
     - ucx-proc=*=gpu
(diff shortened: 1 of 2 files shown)


conda/recipes/libcugraph/meta.yaml
```diff
@@ -77,6 +77,7 @@ requirements:
     - librmm ={{ minor_version }}
     - nccl {{ nccl_version }}
     - ucx-proc=*=gpu
+    - ucx >=1.14.0
```
raydouglass, line 80: Should this match the pin in `ucx-py`: https://github.com/rapidsai/ucx-py/blob/branch-0.35/conda/recipes/ucx-py/meta.yaml#L38
```suggestion
    - ucx >=1.14.1,<1.16.0
```

PR #4000, opened 2023-11-15
Try upgrading pip to fix devcontainer builds

vyasr:
If this works we probably want to upgrade pip in the underlying containers, not here.


vyasr:
The problem diagnosis in this PR is correct, but the solution doesn't work because the pip upgrade is happening after the problematic installation has already been attempted. We're going to get the underlying containers updated so that this works out of the box.

PR #4001, opened 2023-11-15
Update transform_e to support edge masking

seunghwak:
Update transform_e to support graphs with edge masking. This is necessary for K-truss.

### DIFF
--- a/cpp/src/prims/transform_e.cuh
+++ b/cpp/src/prims/transform_e.cuh
@@ -20,6 +20,7 @@
 #include <cugraph/edge_src_dst_property.hpp>
 #include <cugraph/graph_view.hpp>
 #include <cugraph/utilities/error.hpp>
+#include <cugraph/utilities/mask_utils.cuh>
 #include <cugraph/utilities/packed_bool_utils.hpp>
 
 #include <raft/core/handle.hpp>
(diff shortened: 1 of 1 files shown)


seunghwak:
@jnke2016 FYI

naimnv (approved):

cpp/include/cugraph/graph_view.hpp
```diff
@@ -285,6 +289,20 @@ class graph_base_t {
   bool is_symmetric() const { return properties_.is_symmetric; }
   bool is_multigraph() const { return properties_.is_multigraph; }
 
+  void attach_edge_mask(edge_property_view_t<edge_t, uint32_t const*, bool> edge_mask_view)
+  {
+    edge_mask_view_ = edge_mask_view;
+  }
+
+  void clear_edge_mask() { edge_mask_view_ = std::nullopt; }
```
ChuckHastings, line 297: I have no preference.  Is there a reason to prefer this syntax over `edge_mask_view_.reset()`?

ChuckHastings (approved):

cpp/include/cugraph/graph_view.hpp
seunghwak: But consistency matters, so if we are using `.reset()` elsewhere in our codebase, I am OK to make a change.

ChuckHastings: That was my assessment.  A quick check in the code, I see a few calls to `.reset()` and a bunch more assignments to `std::nullopt`.

I'm going to merge this PR.  We should probably pick a way and create a PR to update things to be consistent.

PR #4011, opened 2023-11-17
Find rmm before cuco

vyasr:
RAPIDS currently relies on copies of CCCL headers bundled into rmm. This dependency is centralized by virtue of rmm installing these into the package and everything else finding those installed packages. To do this, however, rmm must be loaded first so that the libcudacxx install location is patched into CMake's search paths. cugraph also uses cuco, which requires libcudacxx but does not bundle its own, so rmm must be found first so that cuco can find libcudacxx where rmm installed it.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -155,6 +155,7 @@ rapids_cpm_init()
 
 # Need to make sure rmm is found before cuco so that rmm patches the libcudacxx
 # directory to be found by cuco.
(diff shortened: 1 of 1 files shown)


bdice (commented):
Yes, this is probably the problem. I am a little upset I didn't think of this earlier. I checked the order of CPM packages, but I didn't consider that RMM's libcudacxx is put into a different path and thus needs to be handled first. 😢

bdice (approved):

rlratzel (approved):

PR #4015, opened 2023-11-20
update the version number in the build.sh script

BradReesWork:
quick update to the version number based on it now being needed in the build.sh for docs


BradReesWork:
merged into 3938

PR #4026, opened 2023-11-29
Replace graph_view.hpp::number_of_edges with compute_number_of_edges

seunghwak:
Replace graph_view.hpp::number_of_edges (deprecated, throws an exception if an edge mask is attached to the graph view object) with compute_number_of_edges (this function works with or without edge mask)


jnke2016 (approved):
Looks good to me

ChuckHastings (approved):

cpp/src/structure/graph_view_impl.cuh
```diff
@@ -532,19 +487,62 @@ graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu, std::enable_if_t<!mu
...
+    auto value_firsts = (*(this->edge_mask_view())).value_firsts();
+    auto edge_counts  = (*(this->edge_mask_view())).edge_counts();
+    assert(value_firsts.size() == 0);
+    assert(edge_counts.size() == 0);
+    return static_cast<edge_t>(detail::count_set_bits(handle, value_firsts[0], edge_counts[0]));
+  } else {
+    return this->number_of_edges_;
+  }
+}
+
```
naimnv, line 523: Do we need two specialization based on `multi_gpu`?
 `value_firsts.size() == 1` and isn't first one with for loop is good enough?

seunghwak: I am not sure, let me experiment.

We have two specializations of graph_view_t based on multi_gpu. I am not sure we have two specializations of graph_view_t but have just one common member function definition.

This function no longer can't be in graph_base_t which is a parent class of graph_t & graph_view_t as masking is applied to graph_view_t but not on graph_t.

naimnv (approved):
LGTM

cpp/src/structure/graph_view_impl.cuh
naimnv: That makes a lot of sense.

naimnv (approved):

PR #4029, opened 2023-11-30
nx-cugraph: add `ancestors` and `descendants`

eriknw:


python/nx-cugraph/nx_cugraph/algorithms/traversal/breadth_first_search.py
```diff
@@ -0,0 +1,256 @@
...
+            cp.array([], dtype=index_dtype),
+            id_to_key=[source],
+        )
+    # TODO: create renumbering helper function(s)
+    unique_node_ids = cp.unique(cp.hstack((predecessors, node_ids)))
+    # Renumber edges
+    # Option 1
+    src_indices = cp.searchsorted(unique_node_ids, predecessors)
+    dst_indices = cp.searchsorted(unique_node_ids, node_ids)
+    # Option 2
```
rlratzel, line 129: Why are we keeping the commented-out option 2 code?

python/nx-cugraph/nx_cugraph/classes/graph.py
```diff
@@ -458,6 +458,24 @@ def has_edge(self, u: NodeKey, v: NodeKey) -> bool:
...
+        if self.key_to_id is not None:
+            n = self.key_to_id[n]
+        nbrs = self.dst_indices[self.src_indices == n]
+        if self.is_multigraph():
+            nbrs = cp.unique(nbrs)
+        return nbrs
+
+    @networkx_api
+    def neighbors(self, n: NodeKey) -> Iterator[NodeKey]:
+        nbrs = self._neighbors(n)
```
rlratzel, line 476: I'm curious about this pattern here; why is there a separate `_neighbors` method instead of just including it inline here in `neighbors`?

python/nx-cugraph/nx_cugraph/interface.py
```diff
@@ -246,6 +246,39 @@ def key(testpath):
             key("test_tree_isomorphism.py:test_positive"): too_slow,
             key("test_tree_isomorphism.py:test_negative"): too_slow,
             key("test_efficiency.py:TestEfficiency.test_using_ego_graph"): maybe_oom,
+            # These repeatedly call `bfs_layers`, which converts the graph every call
```
rlratzel, line 249: Is this reason enough to move this PR to Draft until there's some kind of caching mechanism added?

python/nx-cugraph/nx_cugraph/classes/graph.py
eriknw: Different return type. Sometimes we want the cupy array. `G.neighbors` is part of the networkx API and is an iterator.

python/nx-cugraph/nx_cugraph/interface.py
eriknw: I don't think that's necessary. I think this PR is reviewable and mergable. We already have this of repeated conversions elsewhere. Perhaps we could create an issue to track this.

python/nx-cugraph/nx_cugraph/algorithms/traversal/breadth_first_search.py
```diff
@@ -0,0 +1,256 @@
...
+        compute_predecessors=True,
+        do_expensive_check=False,
+    )
+    mask = predecessors >= 0
+    return distances[mask], predecessors[mask], node_ids[mask]
+
+
+@networkx_algorithm
+def generic_bfs_edges(G, source, neighbors=None, depth_limit=None, sort_neighbors=None):
+    """`neighbors` parameter is not yet supported."""
```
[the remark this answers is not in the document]
eriknw: No preference. I can mention it.

python/nx-cugraph/nx_cugraph/algorithms/traversal/breadth_first_search.py
eriknw: Chose option 1 and removed option 2. Perhaps we can revisit option 2 once we make renumbering helper functions.

eriknw:
@rlratzel what do you think about the three tests that fail due to running out of GPU memory? I don't think the graphs are particularly large. What can we try? Can we investigate? Could it help to call `gc.collect()` regularly during tests (since maybe device memory is being help by released Python objects, but Python's garbage collection doesn't trigger b/c host memory usage is low)? Or, should we just skip the tests (and maybe run them one-by-one)?

python/nx-cugraph/nx_cugraph/interface.py
rlratzel: ok, I created this issue: https://github.com/rapidsai/cugraph/issues/4055

python/nx-cugraph/nx_cugraph/algorithms/dag.py
```diff
@@ -0,0 +1,55 @@
...
+__all__ = [
+    "descendants",
+    "ancestors",
+]
+
+
+def _ancestors_and_descendants(G, source, *, is_ancestors):
+    G = _to_graph(G)
+    if source not in G:
+        hash(source)  # To raise TypeError if appropriate
```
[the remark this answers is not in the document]
rlratzel: I was thinking about this more and realized it might be better to have a function each NX algo defined then called to verify it's inputs, then that same function could be called by backends too - no code duplication and guaranteed consistency.  This would also make it possible to support applications that call `nx-cugraph` directly.  The pattern could be something like this:
```python
######## NX dispatcher
def _dispatch(func):
    def _set_assert_args(aa_func):
        func._assert_args = aa_func
    func._arg_assertion = _set_assert_args
    ...
    return func

######## NX algo
@_dispatch
def foo(a, b):
  foo._assert_args(a, b)
  return a + b

@foo._arg_assertion
def _(a, b):
    if not(isinstance(a, int)):
        raise TypeError(f"expected int for a, got {type(a).__name__}")
    if not(isinstance(b, int)):
        raise TypeError(f"expected int for b, got {type(b).__name__}")

######## backend
def backend_foo(a, b):
    foo._assert_args(a, b)
    return a + b
```

but I agree that this would be a lot of work to refactor every dispatchable NX algo to apply this pattern everywhere.  We could do this incrementally though...
I don't think this should block this PR though, but maybe we should file an issue to NX (cc @MridulS)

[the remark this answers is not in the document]
rlratzel: I should say, I don't think the lack of reusable algo arg assertion funcs should block the PR, but I think the update to check if `source is not None` and raise `ValueError` to match NX should be done.

rlratzel:
> @rlratzel what do you think about the three tests that fail due to running out of GPU memory? I don't think the graphs are particularly large. What can we try? Can we investigate? 

I was wondering if we have a memory leak, possibly related to PLC graph creation during the conversion from the NX Graph to the PLC Graph happening more often now that `nx-cugraph` has implemented more algos.  Do you see this when you run tests locally, or see anything that resembles a leak when polling `nvidia-smi` during testing?  

> Could it help to call `gc.collect()` regularly during tests (since maybe device memory is being help by released Python objects, but Python's garbage collection doesn't trigger b/c host memory usage is low)? 

Maybe? If this is happening during tests, I'd be concerned that it could happen in a large user application too, and they may not know or would prefer not to manually invoke the GC.  I'll try recreating this on my workstation using your branch.  I can then add a test setup hook that calls `gc.collect()` before each test to see if that makes a difference.

rlratzel:
@eriknw I just ran your branch on my workstation and saw the same results - the same three tests failed with the same OOM error.  I was polling `nvidia-smi` and never saw it exceed ~800MB, and my workstation GPUs have (I think) more memory than the V100s in the CI runners (I have 50GB cards).

I don't think this is a leak or anything `gc.collect()` will fix; I think there's a bug somewhere in the algos (or combination of algos) called from the tests.  Are you seeing the same result when you run these tests locally?

I suspect something is causing an abrupt spike in memory which fails the test and the memory is reclaimed faster than my `nvidia-smi` polling interval, or there's a bug where we're computing an amount of memory to allocate that's much too big.  I'm just guessing at this point though.  I'll keep looking...

image

rlratzel:
@eriknw - I have a minimal repro script.  Can you use this to debug further?  I'm going to stop for the night but I can help next week.
```python
import networkx as nx
from networkx.algorithms import flow

G = nx.fast_gnp_random_graph(100, 0.0575, seed=42)
cut = nx.minimum_node_cut(G, flow_func=flow.boykov_kolmogorov)

G = nx.DiGraph()
G.add_nodes_from([0, 1, 2])
list(nx.algorithms.dag.antichains(G))
```

```
(base) root@98e76e671240:/Projects/eriknw/cugraph# NETWORKX_AUTOMATIC_BACKENDS=cugraph python oom_repro.py
  File "/Projects/eriknw/cugraph/oom_repro.py", line 9, in <module>
    list(nx.algorithms.dag.antichains(G))
  File "/opt/conda/lib/python3.10/site-packages/networkx/algorithms/dag.py", line 940, in antichains
    TC = nx.transitive_closure_dag(G, topo_order)
  File "/opt/conda/lib/python3.10/site-packages/networkx/utils/decorators.py", line 770, in func
    return argmap._lazy_compile(__wrapper)(*args, **kwargs)
  File "<class 'networkx.utils.decorators.argmap'> compilation 28", line 4, in argmap_transitive_closure_dag_25
    import inspect
  File "/opt/conda/lib/python3.10/site-packages/networkx/utils/backends.py", line 576, in __call__
    return self.orig_func(*args, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/networkx/algorithms/dag.py", line 811, in transitive_closure_dag
    TC.add_edges_from((v, u) for u in nx.descendants_at_distance(TC, v, 2))
  File "/opt/conda/lib/python3.10/site-packages/networkx/utils/backends.py", line 569, in __call__
    return self._convert_and_call(
  File "/opt/conda/lib/python3.10/site-packages/networkx/utils/backends.py", line 817, in _convert_and_call
    result = getattr(backend, self.name)(*converted_args, **converted_kwargs)
  File "/Projects/eriknw/cugraph/python/nx-cugraph/nx_cugraph/utils/decorators.py", line 86, in __call__
    return self.__wrapped__(*args, **kwargs)
  File "/Projects/eriknw/cugraph/python/nx-cugraph/nx_cugraph/algorithms/traversal/breadth_first_search.py", line 239, in descendants_at_distance
    distances, predecessors, node_ids = plc.bfs(
  File "bfs.pyx", line 181, in pylibcugraph.bfs.bfs
  File "utils.pyx", line 53, in pylibcugraph.utils.assert_success
RuntimeError: non-success value returned from cugraph_bfs: CUGRAPH_UNKNOWN_ERROR std::bad_alloc: out_of_memory: CUDA error at: /opt/conda/include/rmm/mr/device/cuda_memory_resource.hpp
```

eriknw:
> Do you see this when you run tests locally, or see anything that resembles a leak when polling nvidia-smi during testing?

Yeah, I see it (just like you) when I run the tests locally. However, if I run each failing test individually, they pass.

Thanks for investigating and making a reproducer!

ajschmidt8 (approved):

python/nx-cugraph/nx_cugraph/algorithms/dag.py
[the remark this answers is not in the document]
eriknw: NetworkX raises `NetworkXError` for these algorithms if source is `None`, so we are correct how we match in this case. I think we may need to be more careful about `None` as a node (which isn't allowed) elsewhere--you're right, `ValueError: None cannot be a node` is somewhat common.

ci/test_python.sh
```diff
@@ -111,6 +111,11 @@ popd
 rapids-logger "pytest networkx using nx-cugraph backend"
 pushd python/nx-cugraph
 ./run_nx_tests.sh
+# Individually run tests that are skipped above b/c they may run out of memory
+PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestDAG and test_antichains"
+PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestMultiDiGraph_DAGLCA and test_all_pairs_lca_pairs_without_lca"
+PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestDAGLCA and test_all_pairs_lca_pairs_without_lca"
+PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestEfficiency and test_using_ego_graph"
```
rlratzel, line 118: We can leave this in place in order to get this PR merged, but I think this is masking a bug (likely in pylibcugraph/libcugraph, not in nx-cugraph).

I filed [this issue](https://github.com/rapidsai/cugraph/issues/4067) and I'll open a PR to remove these lines and the skipped tests separately, which will be merged when the bug is fixed.

rlratzel (approved):

ci/test_python.sh
rlratzel: The PR to eventually fix the bug is #4068

PR #4030, opened 2023-12-01
MNMG ECG

naimnv:
ECG based on Louvain.


cpp/src/community/louvain_impl.cuh
```diff
@@ -82,11 +84,108 @@ std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> louvain(
...
+              nr_samples_ith_gpu -= delta;
+            }
+          }
+        }
+        // shuffle as many vertices as local vertex partition size to each GPU.
+        std::tie(random_cluster_assignments, std::ignore) =
+          cugraph::shuffle_values(handle.get_comms(),
+                                  random_cluster_assignments.begin(),
+                                  tx_value_counts,
+                                  handle.get_stream());
```
seunghwak, line 173: You don't need to re-process the entire values.

Say, you count the deficits and extras.

You can run `device_allgatherv` on extras.
You can run exclusive_scan on deficits.
Then, in a GPU with a non-zero deficit, you can determine the subrange of the device_allgatherv output to copy.

This will involve much less traffic. If you want to make this algorithm to scale to many thousands of GPUs, you can replace device_allgatherv with P2P. You can compute which GPUs contain the data for the subrange. Then, you can set-up P2P based on this calculation. This will be necessary if the device_allgatherv output array is comparable or larger than # vertices / # GPUs.

cpp/tests/community/mg_ecg_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+// will be instantiated as the parameter to the tests defined below using
+// INSTANTIATE_TEST_SUITE_P()
+//
+struct Ecg_Usecase {
+  double min_weight_{0.1};
+  size_t ensemble_size_{10};
+  size_t max_level_{100};
+  double threshold_{1e-7};
+  double resolution_{1.0};
+  bool check_correctness_{false};
```
seunghwak, line 49: check_correctness should better be true by default (that is our convention).

cpp/tests/community/mg_ecg_test.cpp
```diff
@@ -0,0 +1,270 @@
...
+    override_Rmat_Usecase_with_cmd_line_arguments(GetParam()));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+  file_tests,
+  Tests_MGEcg_File,
+  ::testing::Combine(
+    // enable correctness checks for small graphs
+    ::testing::Values(Ecg_Usecase{0.1, 2, 100, 1e-7, 1.0, true}),
+    ::testing::Values(cugraph::test::File_Usecase("test/datasets/karate.mtx"))));
```
seunghwak, line 237: Have you tested with other larger graphs?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -779,6 +781,51 @@ void ecg(raft::handle_t const& handle,
...
+ * @brief Computes the ecg clustering of the given graph.
+ *
+ * ECG runs truncated Louvain on an ensemble of permutations of the input graph,
+ * then uses the ensemble partitions to determine weights for the input graph.
+ * The final result is found by running full Louvain on the input graph using
+ * the determined weights. See https://arxiv.org/abs/1809.05578 for further
+ * information.
+ *
+ * @throws     cugraph::logic_error when an error occurs.
+ *
```
seunghwak, line 794: Missing documentation for template parameters (and could you fix the documentation error for Louvain as well? Louvain is using the same template parameters but it's tparam documentation is listing graph_view_t).
``` 
 * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
 * @tparam weight_t Type of edge weights. Needs to be a floating point type.
 * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
 ```

File: cpp/src/community/ecg_impl.cuh [hunk 5]
```diff
@@ -0,0 +1,176 @@
...
+{
+  using graph_view_t = cugraph::graph_view_t<vertex_t, edge_t, false, multi_gpu>;
+
+  CUGRAPH_EXPECTS(min_weight >= weight_t{0.0},
+                  "Invalid input arguments: min_weight must be positive");
+  CUGRAPH_EXPECTS(ensemble_size >= 1,
+                  "Invalid input arguments: ensemble_size must be a non-zero integer");
+  CUGRAPH_EXPECTS(
+    threshold > 0.0 && threshold <= 1.0,
+    "Invalid input arguments: threshold must be a positive number in range [0.0, 1.0]");
```
seunghwak, line 57: `[0.0, 1.0]` means >= 0.0 and <= 1.0. shouldn't this be `(0.0, 1.0]`?

cpp/src/community/ecg_impl.cuh
```diff
@@ -0,0 +1,176 @@
...
+  CUGRAPH_EXPECTS(
+    resolution > 0.0 && resolution <= 1.0,
+    "Invalid input arguments: resolution must be a positive number in range [0.0, 1.0]");
```
seunghwak, line 60: Same here, `(0.0, 1.0]`?

cpp/src/community/louvain_impl.cuh
```diff
@@ -82,11 +84,130 @@ std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> louvain(
...
+                                             *rng_state);
+        thrust::sort_by_key(handle.get_thrust_policy(),
+                            random_numbers.begin(),
+                            random_numbers.end(),
+                            random_cluster_assignments.begin());
+
+        // take care of deficits and extras numbers
+
+        vertex_t nr_extras = static_cast<vertex_t>(random_cluster_assignments.size()) -
+                             current_graph_view.local_vertex_partition_range_size();
```
seunghwak, line 169: This won't work if vertex_t is unsigned.

cpp/src/community/louvain_impl.cuh
```diff
@@ -82,11 +84,130 @@ std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> louvain(
...
+
+          h_random_gpu_ranks.resize(d_random_numbers.size());
+
+          raft::update_host(h_random_gpu_ranks.data(),
+                            d_random_numbers.data(),
+                            d_random_numbers.size(),
+                            handle.get_stream());
+        }
+
+        for (int i = 0; i < static_cast<int>(random_cluster_assignments.size() % comm_size); i++) {
```
ChuckHastings, line 142: I'm not sure what this loop (perhaps this block of code) is accomplishing.  Suppose we have 100 vertices spread across 2 GPUs where GPU 0 has 52 vertices and GPU 1 has 48 vertices.  comm_size will be 2.

d_random_numbers in this code block is going to be (52 % 2) long (which is 0 elements) on GPU 0, and (48 % 2) long (which is 0 elements) on GPU 1.  

`h_random_gpu_ranks` will also have no elements, so this code block will AFAICT, do nothing.  I haven't thought through what the shuffle_values call below will do... probably overwrite random_cluster_assignments with empty arrays on each node.

seunghwak: This code is relevant only when # local vertices % comm_size != 0.

We initially assign # local vertices / comm_size to each rank, then, for the remaining # local vertices % comm_size, this logic decides how to distribute the remaining.

My concern is that assigning # local vertices / comm_size to each rank does not achieve uniform random distribution.

I think we should do something like the following.
```
generate a sequence of vertex IDs [local_vertex_partition_range_first, local_vertex_partition_range_last).
if (multi_gpu) {
  Generate a random number in [0, comm_size) per vertex ID and assign a rank based on this random number.
  shuffle.
}
locally shuffle.
if (multi_gpu) {
  move based on extras/deficits.
}
```

seunghwak: Umm... wait a second. Let me think more about this...

seunghwak: Yeah... I think the logic above will work.

naimnv: @seunghwak Thank for address @ChuckHastings's question. I was about to do it after refactoring that part of the code into a new utility function `permute_range`.
Your suggestion on generation a random rank for each vertex_id is also implement in `permute_range` function.

cpp/src/detail/permute_range.cu
```diff
@@ -0,0 +1,191 @@
...
+      sub_range_sizes[comm_rank] == local_range_start,
+      "Invalid input arguments: a rage must have contiguous and non-overlapping values");
+  }
+  rmm::device_uvector<vertex_t> permuted_intergers(local_range_size, handle.get_stream());
+
+  // generate as many number as #local_vertices on each GPU
+  detail::sequence_fill(
+    handle.get_stream(), permuted_intergers.begin(), permuted_intergers.size(), local_range_start);
+
+  // shuffle/permute locally
```
ChuckHastings, line 66: This step is not necessary for for multi gpu, the first thing you do in the multi-gpu case is to sort by a different random ordering.

I would suggest structuring this in the else block of the multi-gpu block.  So, if (MG) do the code currently in the MG, else do this block of code.

cpp/tests/community/mg_ecg_test.cpp
```diff
@@ -0,0 +1,271 @@
...
+                                                                       sg_graph_view,
+                                                                       sg_edge_weight_view,
+                                                                       min_weight,
+                                                                       ensemble_size,
+                                                                       max_level,
+                                                                       threshold,
+                                                                       resolution);
+    }
+    if (comm_rank == 0) {
+      // FIXME: SG and MG modularity values can differ by a few percent,
```
ChuckHastings, line 115: Do you have a strategy for comparing results?  We should also test SG, which I don't see yet in the PR.

cpp/src/detail/permute_range.cu
```diff
@@ -0,0 +1,191 @@
...
+{
+  if (multi_gpu) {
+    auto& comm           = handle.get_comms();
+    auto const comm_size = comm.get_size();
+    auto const comm_rank = comm.get_rank();
+
+    auto global_start =
+      cugraph::host_scalar_bcast(handle.get_comms(), local_range_start, 0, handle.get_stream());
+    auto sub_range_sizes =
+      cugraph::host_scalar_allgather(handle.get_comms(), local_range_size, handle.get_stream());
```
seunghwak, line 53: Global communication is not cheap. As this is a detail function, we may better assume that it is the caller's responsibility to provide valid input arguments.

cpp/src/detail/permute_range.cu
```diff
@@ -0,0 +1,191 @@
...
+                      permuted_intergers.begin());
+
+  if (multi_gpu) {
+    // distribute shuffled/permuted numbers to other GPUs
+    auto& comm           = handle.get_comms();
+    auto const comm_size = comm.get_size();
+    auto const comm_rank = comm.get_rank();
+
+    std::vector<size_t> tx_value_counts(comm_size);
+    std::fill(tx_value_counts.begin(), tx_value_counts.end(), 0);
```
seunghwak, line 88: `std::vector<size_t> tx_value_counts(comm_size, 0)` is simpler.

cpp/src/detail/permute_range.cu
seunghwak: Or you can separate the above multi GPU path to three blocks: 1) multi-GPU shuffling (if MG), 2) local shuffle (for both SG and MG), 3) taking care of extras/deficits (if MG). This will minimize code repetition.

cpp/src/detail/permute_range.cu
```diff
@@ -0,0 +1,191 @@
...
+    thrust::sort_by_key(handle.get_thrust_policy(),
+                        fractional_random_numbers.begin(),
+                        fractional_random_numbers.end(),
+                        permuted_intergers.begin());
+
+    // take care of deficits and extras numbers
+
+    int nr_extras =
+      static_cast<int>(permuted_intergers.size()) - static_cast<int>(local_range_size);
+    int nr_deficits = nr_extras >= 0 ? 0 : -nr_extras;
```
seunghwak, line 154: `int` can possibly overflow for very large graphs in the future (local range size can possibly exceed 2^31-1).

```
size_t nr_extras{0};
size_t nr_deficits{0};
if (permuted_integers.size() > static_cast<size_t>(local_range_size) {
  nr_extras = permuted_integers.size() - static_cast<size_t>(local_range_size);
}
else {
  nr_deficits = static_cast<size_t>(local_range_size) - permuted_integers.size();
}
```

seunghwak (approved):

ChuckHastings (approved):

PR #4032, opened 2023-12-01
Clean up self-loop and multi-edge removal logic

ChuckHastings:
There are mask utilities that perform some of the functions that were implemented to do this cleanup.

Use the mask utilities instead of replicating functionality.


seunghwak (approved):
LGTM

PR #4033, opened 2023-12-01
Add support for Louvain to MTMG

ChuckHastings:
Added vertex result instantiation for `vertex_t` which is necessary to handle Louvain results.

Added an MTMG test for Louvain to demonstrate how to use Louvain in MTMG.

Closes https://github.com/rapidsai/graph_dl/issues/330

### DIFF
--- a/cpp/src/mtmg/vertex_result.cu
+++ b/cpp/src/mtmg/vertex_result.cu
@@ -163,5 +163,113 @@ template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
   cugraph::mtmg::graph_view_t<int64_t, int64_t, true, true> const& graph_view,
   std::optional<cugraph::mtmg::renumber_map_view_t<int64_t>>& renumber_map_view);
 
+template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
+  handle_t const& handle,
+  raft::device_span<int32_t const> vertices,
+  cugraph::mtmg::graph_view_t<int32_t, int32_t, false, false> const& graph_view,
+  std::optional<cugraph::mtmg::renumber_map_view_t<int32_t>>& renumber_map_view);
+
+template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
+  handle_t const& handle,
+  raft::device_span<int32_t const> vertices,
+  cugraph::mtmg::graph_view_t<int32_t, int64_t, false, false> const& graph_view,
+  std::optional<cugraph::mtmg::renumber_map_view_t<int32_t>>& renumber_map_view);
+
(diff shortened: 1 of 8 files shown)


cpp/src/mtmg/vertex_result.cu
```diff
@@ -163,5 +163,113 @@ template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
   cugraph::mtmg::graph_view_t<int64_t, int64_t, true, true> const& graph_view,
   std::optional<cugraph::mtmg::renumber_map_view_t<int64_t>>& renumber_map_view);
 
+template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
+  handle_t const& handle,
+  raft::device_span<int32_t const> vertices,
+  cugraph::mtmg::graph_view_t<int32_t, int32_t, false, false> const& graph_view,
```
seunghwak, line 169: Should we really take `graph_view` here or we can just take vertex partitioning ranges? Many of these explicit instantiations won't be necessary if we just take something like vertex_partition_range_lasts (https://github.com/rapidsai/cugraph/blob/branch-24.02/cpp/include/cugraph/graph_functions.hpp#L237).

LGTM otherwise.

ChuckHastings: Updated in latest push.

cpp/src/mtmg/vertex_result.cu
```diff
@@ -130,37 +129,85 @@ rmm::device_uvector<result_t> vertex_result_view_t<result_t>::gather(
...
+  vertex_partition_view_t<int64_t, false> vertex_partition_view,
+  std::optional<cugraph::mtmg::renumber_map_view_t<int64_t>>& renumber_map_view);
+
 template rmm::device_uvector<float> vertex_result_view_t<float>::gather(
   handle_t const& handle,
   raft::device_span<int32_t const> vertices,
-  cugraph::mtmg::graph_view_t<int32_t, int64_t, true, false> const& graph_view,
+  std::vector<int32_t> const& vertex_partition_range_lasts,
+  vertex_partition_view_t<int32_t, true> vertex_partition_view,
   std::optional<cugraph::mtmg::renumber_map_view_t<int32_t>>& renumber_map_view);
```
seunghwak, line 148: What's the difference between this one and the one in line 129 (https://github.com/rapidsai/cugraph/pull/4033/files#diff-9e67e4600f1b06cb0c48b6303edf4571f5366dbee772c45a38d9c64cb5def518R129)?

seunghwak (approved):
LGTM

jnke2016 (approved):
Looks good to me

naimnv (approved):
Thanks, It would be good start point to implement MTMG for other algorithms.

PR #4036, opened 2023-12-04
Create `cugraph-equivariant` package

tingyu66:
Bring up `cugraph-equivariant` package and add TensorProduct conv layers.


tingyu66:
@stadlmax I've updated the wrapper based on the new fused_tp kernels. Can you take a look? Tests won't pass until we have the new kernels reflected in pylibcugraphops nightly.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+                w_edge, w_src, w_dst = torch.chunk(self.mlp[0].weight, chunks=3, dim=-1)
+                tp_weights = (
+                    edge_emb @ w_edge.T
+                    + (src_scalars @ w_src.T)[src]
+                    + (dst_scalars @ w_dst.T)[dst]
+                    + self.mlp[0].bias
+                )
+                tp_weights = self.mlp[1:](tp_weights)
+            else:
+                tp_weights = self.mlp(edge_emb)
```
mariogeiger, line 232: Here I would either (i) raise an error if `src_scalars` or `dst_scalars` are provided but `self.mlp_fast_first_layer` is False or (ii) concatenate `src_scalars[src]` and `dst_scalars[dst]` into `edge_emb` in the `else:` block

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+        self.use_e3nn_tp = use_e3nn_tp
+
+        self.batch_norm = BatchNorm(out_irreps) if batch_norm else None
+
+        if mlp_fast_first_layer:
+            assert mlp_channels is not None
+            assert mlp_channels[0] % 3 == 0
+            self.num_scalars = int(mlp_channels[0] / 3)
+        else:
+            self.num_scalars = None
```
mariogeiger, line 136: Why `edge_emb`, `src_scalars` and `dst_scalars` are assumed to be of the same size?

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def __init__(
+        self,
+        in_irreps: o3.Irreps,
+        sh_irreps: o3.Irreps,
+        out_irreps: o3.Irreps,
+        batch_norm: bool = True,
+        mlp_channels: Optional[Sequence[int]] = None,
+        mlp_activation: Optional[Callable[..., nn.Module]] = nn.GELU,
+        mlp_fast_first_layer: bool = False,
```
mariogeiger, line 113: Why does this argument need to be here? Can't the special just happen when `src_scalars` and `dst_scalars` are provided?

mariogeiger (changes requested):
There is a problem if `mlp_fast_first_layer=False` but `src_scalars` and `dst_scalars` are provided.

(optional) It would be nice if we can remove that `mlp_fast_first_layer` and simply do the thing when `src_scalars` and `dst_scalars` are provided.

python/cugraph-equivariant/cugraph_equivariant/nn/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .tensor_product_conv import FullyConnectedTensorProductConv
+
+DiffDockTensorProductConv = FullyConnectedTensorProductConv
```
DejunL, line 16: Why do we need this alias?

DejunL (changes requested):

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
DejunL: I agree with Mario here. There are a few branches when it comes to computing the tp weights:
1. user direct input the precomputed weights to forward
2. feed only the edge_emb to the mlp to compute the weight. This could include the case that the node embeddings are already concatenated into the edge_emb, which we can't distinguish
3. As in point 2 but additionally have one node embedding tensor, i.e., the graph is non-directed
4. As in point 2 but additionally have separate src and dst node embedding tensors

tingyu66:
@mariogeiger @DejunL 
I referred to the diffdock repo when implementing it and thought `edge`, `src` and `dst` sharing the same `num_scalers` would the only use case . If we do need support for various `num_scalers` for different component, I can make the change accordingly.

I will also remove the `mlp_fast_first_layer` flag, since it is only used to validate the dimensionality.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
tingyu66: Will remove `mlp_fast_first_layer` option here, and trigger the faster GEMM based on inputs from forward.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+    >>> #         features from edges, sources and destinations have to be passed
+    >>> #         in separately.
+    >>>
+    >>> conv3 = FullyConnectedTensorProductConv(in_irreps, sh_irreps, out_irreps,
+    >>>     mlp_channels=[6, 16, 16], mlp_fast_first_layer=True).cuda()
+    >>> out = conv3(src_features, edge_sh, edge_scalars, graph,
+    >>>     src_scalars=src_scalars, dst_scalars=dst_scalars)
+    """
+
+    def __init__(
```
DejunL, line 105: I just confirmed with Guoqing, who is working on the Diffdock training, that we need to use dropout. We previously concluded that we can omit dropout in the API based on the Diffdock inference code. Can you add an option to allow the user to specify the dropout rate here and add `torch.nn.Dropout` to the mlp definition?

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+            self.num_scalars = None
+        self.mlp_fast_first_layer = mlp_fast_first_layer
+
+        if mlp_channels is not None:
+            dims = list(mlp_channels) + [self.tp.weight_numel]
+            mlp = []
+            for i in range(len(dims) - 1):
+                mlp.append(nn.Linear(dims[i], dims[i + 1]))
+                if mlp_activation is not None and i != len(dims) - 2:
+                    mlp.append(mlp_activation())
```
DejunL, line 145: add something like `torch.nn.Dropout(self.prob_dropout)` here

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+    >>> out = conv3(src_features, edge_sh, edge_scalars, graph,
+    >>>     src_scalars=src_scalars, dst_scalars=dst_scalars)
+    """
+
+    def __init__(
+        self,
+        in_irreps: o3.Irreps,
+        sh_irreps: o3.Irreps,
+        out_irreps: o3.Irreps,
+        batch_norm: bool = True,
```
DejunL, line 110: In the Diffdock training code, we have customized batch_norm function. Instead of being a boolean option, can this take a callable that defaults to `e3nn.nn.BatchNorm`?

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
tingyu66: The dropout option can be absorbed into `mlp_activation` argument, if same dropout rate applies to all hidden layers. I can add additional documentation to show the usage.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
tingyu66: Sure. Will make the change.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
DejunL: I see. Do you mean the user will do something like `mlp_activation = lambda: x : nn.Dropout(p)(nn.GELU(x))`?

tingyu66: > I see. Do you mean the user will do something like `mlp_activation = lambda x : nn.Dropout(p)(nn.GELU()(x))`?

Yeah. Or simply `nn.Sequential(nn.ReLU(), nn.Dropout(p))`

python/cugraph-equivariant/cugraph_equivariant/nn/__init__.py
tingyu66: This is from one of our discussions on Monday, but it's totally optional. @mariogeiger Do you think we need the alias?

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+class FullyConnectedTensorProductConv(nn.Module):
+    r"""Message passing layer for tensor products in DiffDock-like architectures.
+    The left operand of tensor product is the spherical harmonic representation
+    of edge vector; the right operand consists of node features in irreps.
+
+    .. math::
+        \sum_{b \in \mathcal{N}_a} Y\left(\hat{r}_{a b}\right)
+        \otimes_{\psi_{a b}} \mathbf{h}_b
+
+    where the path weights :math:`\psi_{a b}` are from user input to the forward()
```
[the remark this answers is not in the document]
tingyu66: \psi denotes the weights while \Psi is the MLP. I have added a brief note to show the two options here

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
tingyu66: I've updated the code to support arbitrary number of scalar features in [src, dst]_scalars. They can also be `None` if needed.

Regarding use case 3, users should input `src_scalars=node_attr, dst_scalars=node_attr`.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -0,0 +1,252 @@
...
+            Shape: (num_src_nodes, num_scalars)
+
+        dst_scalars: torch.Tensor, optional
+            Scalar features of destination nodes.
+            Shape: (num_dst_nodes, num_scalars)
+
+        reduce : str, optional (default="mean")
+            Reduction operator. Choose between "mean" and "sum".
+
+        edge_envelope: torch.Tensor, optional
```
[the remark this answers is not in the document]
tingyu66: This comes from our Monday's discussion as well. To me, it's just a function to tweak the amplitude of the output and thus add some degree of expresstivity. @mariogeiger Can you help with the description here? Thanks.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
tingyu66: I haven't came up with a neat way to support this because if we change the argument to be a callable (with the default value `=Batchnorm`), the customized batch_norm function must have the same function signature as e3nn.nn.Batchnorm. With that, I would suggest to apply customized function outside of the conv layer.

ci/test_python.sh
```diff
@@ -247,5 +247,46 @@ else
...
+    set +u
+    conda activate test_cugraph_dgl
+    set -u
+    rapids-mamba-retry install \
+      --channel "${CPP_CHANNEL}" \
+      --channel "${PYTHON_CHANNEL}" \
+      --channel pytorch \
+      --channel nvidia \
+      cugraph-equivariant
+    pip install e3nn==0.5.1
```
rlratzel, line 263: I'm not sure, but would this be better off as a dependency in the `py_test_cugraph_equivariant` section of the `dependencies.yaml` file?

python/cugraph-equivariant/cugraph_equivariant/tests/test_tensor_product_conv.py
```diff
@@ -0,0 +1,115 @@
...
+        tp_conv(
+            src_features=src_features,
+            edge_sh=edge_sh,
+            edge_emb=edge_emb,
+            graph=(edge_index, (num_src_nodes, num_dst_nodes)),
+            src_scalars=src_scalars,
+            dst_scalars=dst_scalars,
+        )
+        @ D_out
+    )
```
mariogeiger, line 113:
```suggestion
    out_before = tp_conv(
        src_features=src_features @ D_in.T,
        edge_sh=edge_sh @ D_sh.T,
        edge_emb=edge_emb,
        graph=(edge_index, (num_src_nodes, num_dst_nodes)),
        src_scalars=src_scalars,
        dst_scalars=dst_scalars,
    )

    # rotate after
    out_after = (
        tp_conv(
            src_features=src_features,
            edge_sh=edge_sh,
            edge_emb=edge_emb,
            graph=(edge_index, (num_src_nodes, num_dst_nodes)),
            src_scalars=src_scalars,
            dst_scalars=dst_scalars,
        )
        @ D_out.T
    )
```

Without the transpose we are relying on the fact that these representations are orthonormal and that the inverse rotation is in the group of rotations. With the `.T` it's closer to the definition of equivariance.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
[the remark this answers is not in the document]
mariogeiger:
```suggestion
        edge_envelope: torch.Tensor, optional
            Optional, typically used as attenuation factors to fade out messages coming from nodes close to the cutoff distance used to create the graph. This is important to make the model smooth to the changes in node's coordinates.
```

ci/test_python.sh
tingyu66: Ideally yes, but `e3nn` depends on PyTorch. Having that in pyproject.toml might pull wrong versions of pytorch for users. cugraph-dgl and -pyg's pyproject.toml does not have pytorch either (I guess for the same reason).

AyodeAwe (approved):

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
DejunL: Both the e3nn.nn.BatchNorm and the modified version we used are nn.module so I should have suggested change the type of the `batch_norm` argument to `nn.module`. But then it becomes a moot point since we can apply the BatchNorm outside of the conv layer like you said. Let me double check with Guoqing, who authored the modified version of BatchNorm

DejunL: Just confirmed with Guoqing that the current API is OK. So it's up to you if you want to keep it as it is

DejunL (approved):

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
stadlmax: Wouldn't it make more sense to just not have BN in the API then?

stadlmax (approved):
besides the minor comment w.r.t. having BN in the API, no further comments

mariogeiger (approved):
Same, looks good except that BN that I'm not super super happy about

python/cugraph-equivariant/cugraph_equivariant/nn/__init__.py
mariogeiger: No need for me, the name `FullyConnectedTensorProductConv` is perfect by its own

ci/test_python.sh
mariogeiger: @tingyu66 do you want me to remove pytorch from the dependencies of e3nn?

rlratzel (approved):

PR #4044, opened 2023-12-05
Remove checks for Pascal, no longer supported

### LINKED ISSUE #4022: Drop support for PASCAL
RAPIDS is dropping support for PASCAL in release 24.02.  There are a collection of places in the C++ code where we check at runtime to see if we're executing on PASCAL and change the behavior (e.g. raise an exception, run a different code path).

We should review the code (I think it's all Louvain related) and drop all of the extra code for supporting PASCAL.

ChuckHastings:
The only remaining "support" for Pascal was a check that if we're on a Pascal GPU we expect the function to throw an exception.  Not sure it does anymore (don't see evidence of it, we haven't tested on Pascal for a long time), but if we're not supporting Pascal it doesn't matter.

Closes #4022


seunghwak (approved):
LGTM

seunghwak:
We should update this as well.

set(supported_archs "60" "62" "70" "72" "75" "80" "86" "89" "90")

Pascal is no longer supported

ChuckHastings:
> We should update this as well.
> 
> set(supported_archs "60" "62" "70" "72" "75" "80" "86" "89" "90")
> 
> Pascal is no longer supported

Dropped 60 and 62 from supported architecture list

ChuckHastings:
> Should this PR also remove the various Pascal checks from the CI scripts, like [here](https://github.com/rapidsai/cugraph/blob/branch-24.02/ci/notebook_list.py) and [here](https://github.com/rapidsai/cugraph/blob/branch-24.02/ci/utils/is_pascal.py)?

I believe so.  Fixed in latest push.

ajschmidt8 (approved):

rlratzel (approved):

PR #4065, opened 2023-12-19
Adds `nx-cugraph` benchmarks for 23.12 algos (SSSP, pagerank, hits, katz_centrality, degree_centrality, eigenvector_centrality)

rlratzel:
closes rapidsai/graph_dl#404

* Adds benchmarks for algos added in the 23.12 release
  * SSSP
  * pagerank
  * hits
  * katz_centrality
  * degree_centrality
  * eigenvector_centrality
* Refactors fixtures for easier usage and mainentance
* uses `benchmark.pedantic` instead of `benchmark` to provide complete control over how to benchmark the algos, since `benchmark` will result in no fewer than 3 runs (calibrate timer, timed run, run to generate func return value) which can be too time consuming for slower runs.
* Removes code to create `Dataset` objects for larger datasets and replaces with the equivalent objects that are now part of `cugraph.datasets`


benchmarks/nx-cugraph/pytest-based/bench_algos.py
```diff
@@ -123,90 +149,308 @@ def get_legacy_backend_selector(backend_name):
...
+    degrees = graph_obj.degree()  # list of tuples of (node, degree)
+    node = max(degrees, key=lambda t: t[1])[0]
+    G = get_graph_obj_for_benchmark(graph_obj, backend_wrapper)
+
+    # Ensure the benchmark time includes computation of each result from the
+    # returned generator
+    def run_generator(*args, **kwargs):
+        func = backend_wrapper(nx.single_target_shortest_path_length)
+        results = func(*args, **kwargs)
+        return list(results)
```
eriknw, line 444: There will be more functions that return generators. Is there a good way to generalize this pattern? Perhaps add e.g. `is_generator=True` to `backend_wrapper`?

eriknw (approved):

benchmarks/nx-cugraph/pytest-based/bench_algos.py
rlratzel: Good suggestion, done in https://github.com/rapidsai/cugraph/pull/4065/commits/7ac6e314ad58d5980b32195134d252cd4d51ba6b

PR #4069, opened 2023-12-21
nx-cugraph: indicate which plc algorithms are used and version_added

eriknw:
Pretty simple PR. I would like for us to use this metadata when creating tables of supported algorithms.


eriknw:
I'm thinking something like this:
```python
import re

import networkx as nx
from networkx.utils.backends import _registered_algorithms as algos

from _nx_cugraph import get_info
from nx_cugraph.interface import BackendInterface


def get_funcpath(func):
    return f"{func.__module__}.{func.__name__}"


def add_branch(G, funcpath, plc):
    branch = funcpath.split(".")
    prev = branch[0]
    for i in range(2, len(branch)):
        cur = ".".join(branch[:i])
        G.add_edge(prev, cur)
        prev = cur
    if plc is not None:
        funcpath += " (" + ", ".join(sorted(plc)) + ")"
    G.add_edge(prev, funcpath)


path_to_name = {
    get_funcpath(algos[funcname]): funcname
    for funcname in get_info()["functions"].keys() & algos.keys()
}
G = nx.DiGraph()
for funcpath in sorted(path_to_name):
    funcname = path_to_name[funcpath]
    add_branch(G, funcpath, getattr(BackendInterface, funcname)._plc_names)

print(re.sub(r"[A-Za-z_\.]*\.", "", ("\n".join(nx.generate_network_text(G)))))
```
which creates
```
╙── networkx
    ├─╼ algorithms
    │   ├─╼ bipartite
    │   │   └─╼ generators
    │   │       └─╼ complete_bipartite_graph
    │   ├─╼ centrality
    │   │   ├─╼ betweenness
    │   │   │   ├─╼ betweenness_centrality (betweenness_centrality)
    │   │   │   └─╼ edge_betweenness_centrality (edge_betweenness_centrality)
    │   │   ├─╼ degree_alg
    │   │   │   ├─╼ degree_centrality
    │   │   │   ├─╼ in_degree_centrality
    │   │   │   └─╼ out_degree_centrality
    │   │   ├─╼ eigenvector
    │   │   │   └─╼ eigenvector_centrality (eigenvector_centrality)
    │   │   └─╼ katz
    │   │       └─╼ katz_centrality (katz_centrality)
    │   ├─╼ community
    │   │   └─╼ louvain
    │   │       └─╼ louvain_communities (louvain)
    │   ├─╼ components
    │   │   └─╼ connected
    │   │       ├─╼ connected_components (weakly_connected_components)
    │   │       ├─╼ is_connected (weakly_connected_components)
    │   │       ├─╼ node_connected_component (weakly_connected_components)
    │   │       └─╼ number_connected_components (weakly_connected_components)
    │   ├─╼ core
    │   │   └─╼ k_truss (k_truss_subgraph)
    │   ├─╼ dag
    │   │   ├─╼ ancestors (bfs)
    │   │   └─╼ descendants (bfs)
    │   ├─╼ isolate
    │   │   ├─╼ is_isolate
    │   │   ├─╼ isolates
    │   │   └─╼ number_of_isolates
    │   ├─╼ link_analysis
    │   │   ├─╼ hits_alg
    │   │   │   └─╼ hits (hits)
    │   │   └─╼ pagerank_alg
    │   │       └─╼ pagerank (pagerank, personalized_pagerank)
    │   ├─╼ shortest_paths
    │   │   └─╼ unweighted
    │   │       ├─╼ single_source_shortest_path_length (bfs)
    │   │       └─╼ single_target_shortest_path_length (bfs)
    │   └─╼ traversal
    │       └─╼ breadth_first_search
    │           ├─╼ bfs_edges (bfs)
    │           ├─╼ bfs_layers (bfs)
    │           ├─╼ bfs_predecessors (bfs)
    │           ├─╼ bfs_successors (bfs)
    │           ├─╼ bfs_tree (bfs)
    │           ├─╼ descendants_at_distance (bfs)
    │           └─╼ generic_bfs_edges (bfs)
    ├─╼ convert_matrix
    │   ├─╼ from_pandas_edgelist
    │   └─╼ from_scipy_sparse_array
    └─╼ generators
        ├─╼ classic
        │   ├─╼ barbell_graph
        │   ├─╼ circular_ladder_graph
        │   ├─╼ complete_graph
        │   ├─╼ complete_multipartite_graph
        │   ├─╼ cycle_graph
        │   ├─╼ empty_graph
        │   ├─╼ ladder_graph
        │   ├─╼ lollipop_graph
        │   ├─╼ null_graph
        │   ├─╼ path_graph
        │   ├─╼ star_graph
        │   ├─╼ tadpole_graph
        │   ├─╼ trivial_graph
        │   ├─╼ turan_graph
        │   └─╼ wheel_graph
        ├─╼ community
        │   └─╼ caveman_graph
        ├─╼ small
        │   ├─╼ bull_graph
        │   ├─╼ chvatal_graph
        │   ├─╼ cubical_graph
        │   ├─╼ desargues_graph
        │   ├─╼ diamond_graph
        │   ├─╼ dodecahedral_graph
        │   ├─╼ frucht_graph
        │   ├─╼ heawood_graph
        │   ├─╼ house_graph
        │   ├─╼ house_x_graph
        │   ├─╼ icosahedral_graph
        │   ├─╼ krackhardt_kite_graph
        │   ├─╼ moebius_kantor_graph
        │   ├─╼ octahedral_graph
        │   ├─╼ pappus_graph
        │   ├─╼ petersen_graph
        │   ├─╼ sedgewick_maze_graph
        │   ├─╼ tetrahedral_graph
        │   ├─╼ truncated_cube_graph
        │   ├─╼ truncated_tetrahedron_graph
        │   └─╼ tutte_graph
        └─╼ social
            ├─╼ davis_southern_women_graph
            ├─╼ florentine_families_graph
            ├─╼ karate_club_graph
            └─╼ les_miserables_graph
```
@rlratzel what do you think and what would you find helpful? I'm not sure if showing PLC usage is helpful to users, but I think it is to us.

It may be better to show the dispatch name in parentheses if it's different from the networkx name, and then maybe show PLC usage in a different diagram.

eriknw:
Here's an example that shows how PLC functions are used:
```
╟── plc.betweenness_centrality
╎   └─╼ algorithms
╎       └─╼ centrality
╎           └─╼ betweenness
╎               └─╼ betweenness_centrality
╟── plc.bfs
╎   └─╼ algorithms
╎       ├─╼ dag
╎       │   ├─╼ ancestors
╎       │   └─╼ descendants
╎       ├─╼ shortest_paths
╎       │   └─╼ unweighted
╎       │       ├─╼ single_source_shortest_path_length
╎       │       └─╼ single_target_shortest_path_length
╎       └─╼ traversal
╎           └─╼ breadth_first_search
╎               ├─╼ bfs_edges
╎               ├─╼ bfs_layers
╎               ├─╼ bfs_predecessors
╎               ├─╼ bfs_successors
╎               ├─╼ bfs_tree
╎               ├─╼ descendants_at_distance
╎               └─╼ generic_bfs_edges
╟── plc.edge_betweenness_centrality
╎   └─╼ algorithms
╎       └─╼ centrality
╎           └─╼ betweenness
╎               └─╼ edge_betweenness_centrality
╟── plc.eigenvector_centrality
╎   └─╼ algorithms
╎       └─╼ centrality
╎           └─╼ eigenvector
╎               └─╼ eigenvector_centrality
╟── plc.hits
╎   └─╼ algorithms
╎       └─╼ link_analysis
╎           └─╼ hits_alg
╎               └─╼ hits
╟── plc.k_truss_subgraph
╎   └─╼ algorithms
╎       └─╼ core
╎           └─╼ k_truss
╟── plc.katz_centrality
╎   └─╼ algorithms
╎       └─╼ centrality
╎           └─╼ katz
╎               └─╼ katz_centrality
╟── plc.louvain
╎   └─╼ algorithms
╎       └─╼ community
╎           └─╼ louvain
╎               └─╼ louvain_communities
╟── plc.pagerank
╎   └─╼ algorithms
╎       └─╼ link_analysis
╎           └─╼ pagerank_alg
╎               └─╼ pagerank
╟── plc.personalized_pagerank
╎   └─╼ algorithms
╎       └─╼ link_analysis
╎           └─╼ pagerank_alg
╎               └─╼ pagerank
╙── plc.weakly_connected_components
    └─╼ algorithms
        └─╼ components
            └─╼ connected
                ├─╼ connected_components
                ├─╼ is_connected
                ├─╼ node_connected_component
                └─╼ number_connected_components
```

eriknw:
Other metadata that might be nice to add to `@networkx_algorithm` functions is whether or not they fully implement the NetworkX function (maybe use `incomplete=True` or `is_complete=False`) and whether the results are the same within floating point tolerance (maybe use `results_different=True` or `is_identical=False`). Any other ideas for metadata or better names for these?

eriknw:
I just added the ability to see what functions are "incomplete" (such as unsupported API) and "different" (such as behavior of RNG). It's a little sobering, but can be a good guide for identifying potential cugraph-core work.

It might be nice to be able to filter by complete/incomplete/etc.

Also, the regex used by `print_tree.py` is a little fragile, so if we want to extend this to do much more, we may want to consider reimplementing instead of adding hack onto hack.

python/nx-cugraph/pyproject.toml
```diff
@@ -52,12 +52,20 @@ test = [
...
 [project.entry-points."networkx.plugin_info"]
 cugraph = "_nx_cugraph:get_info"
 
+# "backend" used in nx version >= 3.2
+[project.entry-points."networkx.backends"]
+cugraph = "nx_cugraph.interface:BackendInterface"
+
+[project.entry-points."networkx.backend_info"]
+cugraph = "_nx_cugraph:get_info"
+
```
eriknw, line 68: btw, this is needed to work with dev version of NetworkX.

python/nx-cugraph/nx_cugraph/algorithms/centrality/betweenness.py
```diff
@@ -18,7 +18,12 @@
 __all__ = ["betweenness_centrality", "edge_betweenness_centrality"]
 
 
-@networkx_algorithm
+@networkx_algorithm(
+    is_incomplete=True,  # weight not supported
+    is_different=True,  # RNG with seed is different
+    plc="betweenness_centrality",
+    version_added="23.10",
```
rlratzel, line 25: My initial reaction was to think it's kind of odd to be sprinkling metadata throughout the source code like this (I was initially envisioning a single yaml file or some other kind of text file that served as both the input to the tree/table dumping tools, and a doc itself that humans could easily refer to in a pinch)...however, the locality of the metadata to the implementation seems like it could be handy for nx-cugraph devs.

One thing I don't like is how the metadata mixes with the more critical args to the decorator (`extra_params`, `nodes_or_number`), making me think the implementation is using these new params at runtime somehow (or is/will it?).  IOW, it's not obvious that it's just metadata.  Perhaps this could instead be done with a single new param named `metadata` which is a dictionary?  That would also allow for easy expansion later and support for algo-specific metadata.
```suggestion
    metadata=dict(
        is_incomplete=True,  # weight not supported
        is_different=True,  # RNG with seed is different
        plc="betweenness_centrality",
        version_added="23.10",
    ),
```
As for the metadata itself, here's my thoughts:
* `is_incomplete`: This seems useful.  Although, it's really helpful to have the comment you added too ("# weight not supported").  What if it was a string or list of strings instead, where empty string or None indicated completeness?  Maybe then it could be renamed something like "missing_features", "not_supported", "not_supported_features", "lacuna", ...?
* `is_different`: This also seems useful, but again, even more so if it was a string/list of strings. Seems like it could also be named "quirks".
* `plc`: I'm starting to think this isn't worthy of the space and maintenance burden. It's only useful to a couple of cugraph devs, and  I'm guessing even that's debatable. PLC algo was initially added to the README table simply because it was an easy way to get a list (just dump all the PLC algos and make a table), but I think readers mainly care about what NX algo is supported, not what PLC algos get called.
* `version_added`: This one seems useful as-is.

rlratzel (approved):
Very neat, thanks!  I'm approving but I also had some suggestions I hope you consider.  Of all of them, the single `metadata` dictionary is my biggest preference.

python/nx-cugraph/nx_cugraph/algorithms/centrality/betweenness.py
eriknw: Thanks for the suggestions! I added clarifying code comments that emphasize that these are metadata only. I like them as keyword arguments--seems cleaner, easier to check, etc. Dict could be nice if we expected to have lots of different metadata fields.

Also, I like `is_*`, which are clearly boolean, and, if True, indicate you should look at the extra docstring to see details.

rlratzel (approved):

PR #4077, opened 2024-01-04
Fix OOB error, BFS C API should validate that the source vertex is a valid vertex

### LINKED ISSUE #4067: OOM error raised after subsequent calls to `plc.bfs` with different `SGGraph`s
The following code produces the error shown below.  The graph is too small to result in a legitimate OOM error, so a bug is likely causing something in `libcugraph` to allocate or attempt to allocate more memory than is available.

```python
import pylibcugraph as plc
import cupy as cp

edges = [
    (0,2), (0,4), (0,5), (0,6), (0,7), (0,8), (0,10), (0,11), (0,12), (0,13),
    (0,17), (0,19), (0,21), (0,31), (2,7), (2,8), (2,9), (2,13), (2,27), (2,28),
    (2,32), (4,6), (4,10), (5,6), (5,10), (5,16), (6,16), (8,30), (8,32),
    (8,33), (9,33), (13,33), (14,32), (14,33), (15,32), (15,33), (18,32),
    (18,33), (19,33), (20,32), (20,33), (22,32), (22,33), (23,25), (23,27),
    (23,29), (23,32), (23,33), (24,25), (24,27), (24,31), (25,31), (26,29),
    (26,33), (27,33), (28,31), (28,33), (29,32), (29,33), (30,32), (30,33),
    (31,32), (31,33), (32,33),
]
edges += [(target, source) for (source, target) in edges]

source=0
src_indices=cp.array([e[0] for e in edges], dtype="int32")
dst_indices=cp.array([e[1] for e in edges], dtype="int32")
G=plc.SGGraph(
    graph_properties=plc.GraphProperties(
        is_multigraph=False,
        is_symmetric=False,
    ),
    src_or_offset_array=src_indices,
    dst_or_index_array=dst_indices,
    weight_array=None,
    store_transposed=False,
    renumber=False,
    do_expensive_check=False,
)

distances, predecessors, node_ids = plc.bfs(
    handle=plc.ResourceHandle(),
    graph=G,
    sources=cp.array([source], "int32"),
    direction_optimizing=False,
    depth_limit=-1,
    compute_predecessors=False,
    do_expensive_check=False,
)
MAXINT=2147483647
mask = distances != MAXINT
it = zip(node_ids[mask].tolist(), distances[mask].tolist())

del G

source=1
src_indices=cp.array([], dtype="int32")
dst_indices=cp.array([], dtype="int32")

G=plc.SGGraph(
    graph_properties=plc.GraphProperties(
        is_multigraph=False,
        is_symmetric=False,
    ),
    src_or_offset_array=src_indices,
    dst_or_index_array=dst_indices,
    weight_array=None,
    store_transposed=False,
    renumber=False,
    do_expensive_check=False,
)

distances, predecessors, node_ids = plc.bfs(
    handle=plc.ResourceHandle(),
    graph=G,
    sources=cp.array([source], "int32"),
    direction_optimizing=False, 
    depth_limit=1,
    compute_predecessors=False,
    do_expensive_check=False,
)
```

```
(base) root@98e76e671240:/Projects/cugraph# python oom_repro.py
  File "/Projects/cugraph/oom_repro.py", line 66, in <module>
    distances, predecessors, node_ids = plc.bfs(
  File "bfs.pyx", line 181, in pylibcugraph.bfs.bfs
  File "utils.pyx", line 53, in pylibcugraph.utils.assert_success
RuntimeError: non-success value returned from cugraph_bfs: CUGRAPH_UNKNOWN_ERROR std::bad_alloc: out_of_memory: CUDA error at: /opt/conda/include/rmm/mr/device/cuda_memory_resource.hpp
```

This was initially discovered in #4029

ChuckHastings:
* Added error check to be sure that the source vertex is a valid vertex in the graph.
* Updated `nx_cugraph.Graph` class to create PLC graphs using `vertices_array` in order to include isolated vertices. This is now needed since the error check added in this PR prevents NetworkX tests from passing if isolated vertices are treated as invalid, so this change prevents that.
* This resolves the problem that required the test workarounds done [here](https://github.com/rapidsai/cugraph/pull/4029#discussion_r1433332010) in [4029](https://github.com/rapidsai/cugraph/pull/4029), so those workarounds have been removed in this PR.

Closes #4067


cpp/src/detail/utility_wrappers.cu
```diff
@@ -227,5 +229,36 @@ template bool is_equal(raft::handle_t const& handle,
...
 
+template <typename data_t, bool multi_gpu>
+size_t count_values(raft::handle_t const& handle,
+                    raft::device_span<data_t const> span,
+                    data_t value)
+{
+  size_t local_count = thrust::count_if(
+    handle.get_thrust_policy(), span.begin(), span.end(), [value] __device__(data_t d) {
+      return d == value;
+    });
```
seunghwak, line 240: You can use thrust::count (https://thrust.github.io/doc/group__counting_ga4f2955d87ada73fc89f3442a19ca1bfa.html#ga4f2955d87ada73fc89f3442a19ca1bfa) and no need to provide a predicate.

cpp/src/detail/utility_wrappers.cu
```diff
@@ -227,5 +229,36 @@ template bool is_equal(raft::handle_t const& handle,
...
+
+  if constexpr (multi_gpu) {
+    local_count = cugraph::host_scalar_allreduce(
+      handle.get_comms(), local_count, raft::comms::op_t::SUM, handle.get_stream());
+  }
```
seunghwak, line 245: All the other functions in this file are SG functions, and this function is an MG function. This might be confusing...

Should we better resolve this?

1. No.
2. Create another file for MG utility wrapper functions
3. Just call thrust::count() and host_scalar_allreduce() instead of creating this function.
4. Rename functions to better hint that this function performs allerduce if MG
5. Something else?

What do you think?

ChuckHastings: I'll look at option 2.

While I like option 3, there are many cases where we don't need to create a `.cu` file except for common utilities like this.  This allows us to leave many files as `.cpp` files instead.

ChuckHastings: Actually, I think what I'll do is move the all reduce (which doesn't use thrust, so can be called in a `.cpp` file) out of the function, making it an SG function.  So I'll leave it here but remove the `multi_gpu` flag.

cpp/src/detail/utility_wrappers.cu
ChuckHastings: Fixed.

seunghwak (approved):
LGTM

ci/test_python.sh
```diff
@@ -111,11 +111,6 @@ popd
 rapids-logger "pytest networkx using nx-cugraph backend"
 pushd python/nx-cugraph
 ./run_nx_tests.sh
-# Individually run tests that are skipped above b/c they may run out of memory
-PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestDAG and test_antichains"
-PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestMultiDiGraph_DAGLCA and test_all_pairs_lca_pairs_without_lca"
-PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestDAGLCA and test_all_pairs_lca_pairs_without_lca"
-PYTEST_NO_SKIP=True ./run_nx_tests.sh --cov-append -k "TestEfficiency and test_using_ego_graph"
```
rlratzel, line 118: ops-codeowners: these lines are removed because the tests are all running as part of the same invocation again, so we're not losing any tests, we're just running them the standard way again (ie. removing a special case no longer needed)

raydouglass (approved):

python/nx-cugraph/nx_cugraph/classes/graph.py
```diff
@@ -157,6 +159,19 @@ def from_coo(
...
+
+        # If the graph contains isolates, plc.SGGraph() must be passed a value
+        # for vertices_array that contains every vertex ID, since the
+        # src/dst_indices arrays will not contain IDs for isolates. Create this
+        # only if needed. Like src/dst_indices, the _node_ids array must be
+        # maintained for the lifetime of the plc.SGGraph
+        all_node_ids = cp.arange(new_graph._N, dtype=index_dtype)
+        isolates = cp.setdiff1d(
+            cp.setdiff1d(all_node_ids, new_graph.src_indices), new_graph.dst_indices
+        )
```
eriknw, line 171: Thanks for the code comments.

I think we could also use `isolates = nxcg.algorithms.isolate._isolates(new_graph)` here (not sure which way is faster--they're both probably pretty fast).

python/nx-cugraph/nx_cugraph/classes/graph.py
```diff
@@ -649,6 +666,7 @@ def _get_plc_graph(
             store_transposed=store_transposed,
             renumber=False,
             do_expensive_check=False,
+            vertices_array=self._node_ids,
```
eriknw, line 669: Cool! I didn't realized `vertices_array` made it in. Also, the recent `drop_self_loops` and `drop_multi_edges` may be useful too 👍

python/nx-cugraph/nx_cugraph/classes/graph.py
rlratzel: I'm using `isolates = nxcg.algorithms.isolate._isolates(new_graph)` to reduce duplication and since it's a bit faster in my experiments.

eriknw (approved):

PR #4083, opened 2024-01-10
Correct `cugraph-pyg` package name used in wheels and fix test script

tingyu66:
This is an oversight from #4075. With this fix, the generated wheels should have proper a CUDA suffix (e.g. `cugraph-pyg-cu12`).

Edit: This PR also fixes https://github.com/rapidsai/graph_dl/issues/421 
The mg tests were not skipped as expected due to wrong relative path in `test_wheel_cugraph-pyg.sh`.

PR #4084, opened 2024-01-10
refactor CUDA versions in dependencies.yaml

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/7.

Proposes splitting the `cuda-version` dependency in `dependencies.yaml` out to its own thing, separate from the bits of the CUDA Toolkit this project needs.

### Benefits of this change

* prevents accidental inclusion of multiple `cuda-version` version in environments
* reduces update effort (via enabling more use of globs like `"12.*"`)
* improves the chance that errors like "`conda` recipe is missing a dependency" are caught in CI

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,7 +52,7 @@ repos:
         pass_filenames: false
         additional_dependencies: [gitpython]
   - repo: https://github.com/rapidsai/dependency-file-generator
-    rev: v1.5.1
+    rev: v1.8.0
     hooks:
         - id: rapids-dependency-file-generator
           args: ["--clean"]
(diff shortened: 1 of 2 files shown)


vyasr (approved):

vyasr (changes requested):
At first I thought the error here was just a timeout, but it looks like it's caused by a real underlying issue. Are we tracking that anywhere in cugraph? If not, should probably open an issue/ping people on Slack.

bdice (approved):

bdice:
@vyasr It looks like https://github.com/rapidsai/cugraph/pull/4083/ touched some relevant code here. I merged in the upstream to see if that helps.

jameslamb:
> I merged in the upstream to see if that helps.

Thanks for that! Seems like it did help.

PR #4086, opened 2024-01-11
Provide explicit pool sizes and avoid RMM detail APIs

### LINKED ISSUE #4066: [BUG]: Update cuGraph to always explicitly specify RMM pool size and avoid using rmm::detail APIs
https://github.com/rapidsai/rmm/pull/1417 and followon PRs make a number of improvements and changes to help consumers of RMM avoid accessing RMM's detail namespace.

    initial pool size in pool_memory_resource is required to be provided (no longer optional)
    rmm::detail::available_device_memory() (moving to `rmm::available_device_memory())
    Move alignment utility functions out of detail namespace.
    Add utility to calculate an aligned percentage of free device memory in bytes. This way, existing instantiations of pool_memory_resource(upstream) can be replaced by pool_memory_resource(upstream, rmm::percent_of_free_device_memory(50), which matches previous behavior.

cuGraph should be updated for these changes once https://github.com/rapidsai/rmm/pull/1417 is merged. I have a PR ready to open.

### Version

24.02

Source

### Describe the bug.

cuGraph calls RMM detail APIs, namely this one, which is going away (it's becoming a public API, but still this is a great example of what not to do).

https://github.com/rapidsai/cugraph/blob/d7c88d1a676015af1def62500d1f7255bc25aa7f/cpp/include/cugraph/mtmg/resource_manager.hpp#L109

_No response_

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

### Code of Conduct

harrism:
This PR fixes up cuGraph to avoid usage that will soon be deprecated in RMM.

Depends on rapidsai/rmm#1417

Fixes #4066


ChuckHastings (approved):
Thanks for the fix.  Yes, careless on my part.  I copied some code from cudf without thinking about the fact I was referencing a detail method outside of cugraph.

harrism:
> Thanks for the fix. Yes, careless on my part. I copied some code from cudf without thinking about the fact I was referencing a detail method outside of cugraph.

Not your fault, cuDF did it first. :) And anyway, these alignment utilities should probably have been made public. The real thing I'm fixing here is the change in behavior of `pool_memory_resource` (requires specifying initial pool size now)

harrism:
@ChuckHastings do you know if the wheel test failures here are a known issue?

ChuckHastings:
> @ChuckHastings do you know if the wheel test failures here are a known issue?

Every day around this time I tend to see failures in local builds and in CI builds like this.  They appear to be transient.  I suspect some sort of race condition in building/installing artifacts into the conda repositories, and since we depend on many RAPIDS artifacts that we are more susceptible to this than other packages.  They generally resolve themselves by rerunning a few hours later.

harrism:
I don't have permission to rerun actions on this repo. But I can merge latest which will require rerunning all.

PR #4092, opened 2024-01-14
nx-cugraph: PLC now handles isolated nodes; clean up our workarounds

eriknw:
Hooray for removing and cleaning code! Tests also added (we already tested isolated nodes for Louvain).

nx-cugraph was updated to handle isolated nodes by passing the node set to PLC in #4077


eriknw:
I forgot that #4071 was still open. @rlratzel let's merge this PR first then I can update #4071.

rlratzel (approved):
Thanks! And thanks for adding the test too.

PR #4093, opened 2024-01-15
nx-cugraph: add triangles and clustering algorithms

eriknw:
NetworkX tests are somewhat underspecified regarding how to handle self-loops for these algorithms. Also, I'm not sure if transitivity is supposed to work on directed graphs.

Once #4071 is merged, it should be easy to add `is_bipartite` function (and maybe others?).


eriknw:
Note that `transitivity` and `is_bipartite` could use e.g. `total_triangles` if PLC had a faster way to implement it. `is_bipartite` only needs to know whether there are _any_ triangles, which in GraphBLAS could be done as `any_pair(L @ U.T).new(mask=L.S).reduce_scalar(any)`

python/nx-cugraph/nx_cugraph/algorithms/cluster.py
```diff
@@ -0,0 +1,139 @@
...
+    node_ids, triangles, is_single_node = _triangles(G, nodes)
+    if len(G) == 0:
+        return {}
+    if is_single_node:
+        numer = int(triangles[0])
+        if numer == 0:
+            return 0
+        degree = int((G.src_indices == nodes).sum())
+        return 2 * numer / (degree * (degree - 1))
+    # What about self-edges?
```
rlratzel, line 73: This leaves me curious; what happens with self-edges? I'm guessing the degree value for a node with a self-edge might be artificially high, but I hate to guess. Is this a limitation we need to address (perhaps as a FIXME) and/or something we need to include in `extra_docstrings`? Maybe at a minimum can you update the comment to elaborate?

rlratzel (commented):
LGTM, just a couple of questions/feedback.

python/nx-cugraph/nx_cugraph/algorithms/bipartite/basic.py
```diff
@@ -0,0 +1,31 @@
...
+from nx_cugraph.algorithms.cluster import _triangles
+from nx_cugraph.convert import _to_graph
+from nx_cugraph.utils import networkx_algorithm
+
+__all__ = [
+    "is_bipartite",
+]
+
+
+@networkx_algorithm(plc="triangle_count", version_added="24.02")
```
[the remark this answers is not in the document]
eriknw: would it help to rename `plc` to `_plc` to signal that _this parameter doesn't matter if you don't know what it's for_?

python/nx-cugraph/nx_cugraph/algorithms/cluster.py
eriknw: NetworkX is underspecified. There are no tests with self-edges. Perhaps we could add a test to match NetworkX, _or_ we could try to upstream tests to NetworkX. The latter will require the NetworkX team being intentional about what is supposed to happen in this case. Should we ignore self-edges or not?

python/nx-cugraph/nx_cugraph/algorithms/bipartite/basic.py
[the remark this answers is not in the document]
eriknw: We may want to use `version_added` to update the docstring, but I haven't fully baked that yet.

We could also make `plc` a _code comment_, which is a less useful, but also potentially less confusing.

python/nx-cugraph/nx_cugraph/algorithms/cluster.py
rlratzel: @eriknw and I discussed offline and both think self-edges probably should not be included.

However, there's a discussion about this started [here](https://github.com/networkx/networkx/discussions/7238) which could change this (probably in a future PR if not answered soon).

python/nx-cugraph/nx_cugraph/algorithms/bipartite/basic.py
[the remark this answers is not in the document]
rlratzel: yeah, a leading underscore could help emphasize these are "internal" params.  I was mainly bringing it up for awareness, and this can (and probably should, unless you'd actually prefer to address it here) be done in a separate PR.

python/nx-cugraph/nx_cugraph/algorithms/cluster.py
eriknw: updated to ignore selfloops and comparison tests added.

python/nx-cugraph/nx_cugraph/algorithms/bipartite/basic.py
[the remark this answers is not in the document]
eriknw: ooh, so we have a reason to merge all open nx-cugraph PRs ;)

python/nx-cugraph/nx_cugraph/algorithms/bipartite/basic.py
```diff
@@ -26,6 +26,6 @@ def is_bipartite(G):
     G = _to_graph(G)
     # Counting triangles may not be the fastest way to do this, but it is simple.
     node_ids, triangles, is_single_node = _triangles(
-        G, None, symmetrize="union" if G.is_directed else None
+        G, None, symmetrize="union" if G.is_directed() else None
```
rlratzel, line 29: Was this discovered by the new tests? :)

rlratzel (approved):
LGTM, thanks for the tests.

PR #4095, opened 2024-01-16
Optimize the drop-duplicate functionality

jnke2016:
Our current python API leverages dask to implement the `drop-duplicate` functionality but it carries a lot of overhead as it draws a significant amount of host memory and results into a crash when processing large graphs (4+ billion edges).

This PR
1. Leverages the CAPI to internally drop multi edges when creating the PLC graph.
2. Deprecates the parameter `multi` which, when set to False, triggers the dask based `drop-duplicate` functionality
3. Add flag `do_expensive_check` to check for `NULL` values in the edgelist


python/cugraph/cugraph/structure/symmetrize.py
```diff
@@ -84,6 +86,11 @@ def symmetrize_df(
     if multi:
         return result
     else:
+        warnings.warn(
+            "'multi' is deprecated as the removal of multi edges is "
+            "only supported when creating a multi-graph",
+            FutureWarning,
+        )
```
rlratzel, line 93: I think I'm missing something: does this mean a user who wants a MultiGraph could also want to remove the multiple edges? Why would they want a MultiGraph without multiple edges?

jnke2016: No, It means that the only way to drop multi edges is by creating a multi-graph. In fact prior to that, we were using the flag `multi` in the `symmetrize` function  to drop multi-edges (if creating a directed or undirected Graph) through the `_memory_efficient_drop_duplicates` function which is dask based. But we can't just get rid of that flag because some users might still have it in their code base if they were directly calling the `symmetrize` function directly

python/cugraph/cugraph/structure/symmetrize.py
```diff
@@ -208,6 +224,7 @@ def symmetrize(
         weights column name.
 
     multi : bool, optional (default=False)
+        Deprecated.
```
VibhuJawa, line 227: Same comment here.

VibhuJawa (changes requested):
Looks good mostly apart from some changes around `docstrings`.  


Can we also quickly also run the graph creation benchmarks here too see how perf may change. 

https://github.com/rapidsai/cugraph/blob/branch-24.02/benchmarks/cugraph/standalone/cugraph_graph_creation.py

python/cugraph/cugraph/structure/symmetrize.py
rlratzel: Okay, thanks.  I think my concerns will go away when Vibhu's feedback is addressed.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -1306,6 +1335,14 @@ def neighbors(self, n):
             n = node[0]
 
         df = self.edgelist.edgelist_df
+
+        vertex_col_name = [simpleGraphImpl.srcCol, simpleGraphImpl.dstCol]
+
+        if not self.properties.multi_edge:
+            # Drop parallel edges for non MultiGraph
+            # FIXME: Drop multi edges with the CAPI instead.
+            df = df.groupby(by=[*vertex_col_name], as_index=False).min()
```
VibhuJawa, line 1344: Same question here , assuming this a cuDF dataframe.

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -1176,10 +1176,12 @@ def test_extract_subgraph_vertex_prop_condition_only(
         )
     # Should result in two edges, one a "relationship", the other a "referral"
     expected_edgelist = cudf.DataFrame(
-        {"src": [89216, 78634], "dst": [78634, 89216], "weights": [99, 8]}
+        {"src": [78634, 89216], "dst": [89216, 78634], "weights": [8, 99]}
```
VibhuJawa, line 1179: Wonder why we flipped it around ?

VibhuJawa (approved):
Mostly looks good, just some clarification around if we are dropping duplicates for  single GPU's only.

rlratzel (approved):
Approving assuming all of @VibhuJawa 's feedback gets addressed.

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -506,6 +511,18 @@ def view_edge_list(self):
...
+                vertex_col_name = srcCol + dstCol
+
+            else:
+                vertex_col_name = [srcCol, dstCol]
+
+            # Drop parallel edges for non MultiGraph
+            # FIXME: Drop multi edges with the CAPI instead.
+            edgelist_df = edgelist_df.groupby(
+                by=[*vertex_col_name], as_index=False
+            ).min()
```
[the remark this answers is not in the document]
jnke2016: Yes it is a cuDF dataframe

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
jnke2016: Yes it is a cuDF dataframe

python/cugraph/cugraph/tests/data_store/test_property_graph.py
jnke2016: I initially removed the `drop_duplicate` call which was swapping the order of some columns. This expected result was set to match the actual_result. But this is no longer relevant since I am no longer updating this test file

jnke2016:
> just some clarification around if we are dropping duplicates for single GPU's only.
 
@VibhuJawa that's correct

python/cugraph/cugraph/tests/data_store/test_property_graph.py
VibhuJawa: Thanks for fixing this.

VibhuJawa (approved):
LGTM

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
VibhuJawa: Thanks

python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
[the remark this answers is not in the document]
VibhuJawa: Thanks

PR #4098, opened 2024-01-17
Mark kernels as internal

robertmaynard:
Downstream consumers of static built versions of RAPIDS C++ projects have encountered runtime issues due to multiple instances of the same kernel existing in different DSOs.

To resolve this issue we need to ensure that all CUDA kernels in all RAPIDS libraries need to be have internal linkage ( static for projects using whole compilation, __attribute__((visibility("hidden"))) for header libraries / separable compilation ).

This updates all cugraph kernels to have internal linkage, and adds a CI job to verify that no new kernels are added with external linkage.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -56,7 +56,7 @@ jobs:
   conda-cpp-checks:
     needs: conda-cpp-build
     secrets: inherit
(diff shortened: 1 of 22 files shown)


BradReesWork (approved):

naimnv (approved):
LGTM

seunghwak (approved):

AyodeAwe (approved):

PR #4081, opened 2024-01-08
cuGraph-DGL Performance Testing

alexbarghi-nv:
Large-scale cuGraph-DGL performance testing scripts.

PR #4099, opened 2024-01-18
WholeGraph Performance Testing

alexbarghi-nv:
Adds option to use WholeGraph Feature Store backend for the GNN training benchmarks.


alexbarghi-nv:
Combined with #4081

PR #4108, opened 2024-01-23
nx-cugraph: use coverage to ensure all algorithms were run

eriknw:
Heh, there is probably a "better" way to do this, but this way was fast and easy enough to do, and I hope is "good enough".


rlratzel (approved):
Nice!  Thanks for the quick response for this.  I was actually poking around with something similar but mine was based on examining the output of `coverage annotate` and I wasn't feeling too good about it, so let's go with this approach :)

raydouglass (approved):

PR #4127, opened 2024-01-30
Re-run `ci/release/update-version.sh 24.04.00`

jakirkham:
Looks like `ci/release/update-version.sh` may not have been run for `24.04.00`. This does that

Also noticed the following errors on macOS. Not sure if these are important. Could someone please double check?

```
sed: 1: "/^ucx_py_version:$/ {n; ...": bad flag in substitute command: '}'
sed: 1: "/^ucx_py_version:$/ {n; ...": bad flag in substitute command: '}'
sed: 1: "/^ucx_py_version:$/ {n; ...": bad flag in substitute command: '}'
```

Edit: Should add do see the right `ucx_py_version` values in these locations. So it does already appear to be correctly updated already

https://github.com/rapidsai/cugraph/blob/00f906f6e97949fd33c85143af4fdf069c791558/conda/recipes/cugraph/conda_build_config.yaml#L19-L20

https://github.com/rapidsai/cugraph/blob/00f906f6e97949fd33c85143af4fdf069c791558/conda/recipes/pylibcugraph/conda_build_config.yaml#L19-L20

https://github.com/rapidsai/cugraph/blob/00f906f6e97949fd33c85143af4fdf069c791558/conda/recipes/cugraph-service/conda_build_config.yaml#L1-L2


BradReesWork (approved):

raydouglass:
> may not have been run for 24.04.00

Looks like these were `24.02` changes merged (https://github.com/rapidsai/cugraph/pull/4123) after `branch-24.04` was created.

> Also noticed the following errors on macOS.

This is because MacOS ships with BSD sed. BSD sed doesn't parse the commands the same as GNU sed. We try to write things so that the commands will work with either, but sometimes it may not be possible.

raydouglass (approved):

PR #4134, opened 2024-02-01
Implement new All Pairs similarity algorithm

ChuckHastings:
Added a new entry point for similarity functionality that combines the functionality of `k_hop_nbrs` and similarity.

This entry point allows us to compute similarity for all pairs of vertices in the graph in a single call.  We also add the optional parameter `topk` which, if specified, will only return the vertices that have the highest scores.  If `topk` is specified on an all pairs call, we compute the scores for pairs in batches and extract the topk as we go along to keep the memory footprint low.

This PR also updates a FIXME in the C++ similarity test.  The C++ similarity test had been written before we had a `k_hop_nbrs` call, so there was some inefficient test code to compute that.  Now that we have a `k_hop_nbrs` call, the test code was refactored to use that call.


cpp/tests/mtmg/threaded_test_louvain.cu
```diff
@@ -277,6 +277,7 @@ class Tests_Multithreaded
 
         std::tie(std::ignore, modularity) = cugraph::louvain<vertex_t, edge_t, weight_t, true>(
           thread_handle.raft_handle(),
+          std::nullopt,
```
ChuckHastings, line 280: We somehow missed this in the ECG update... we don't regularly compile the MTMG code in CI.

ChuckHastings:
This branch got corrupted and pushed.  Closing with a new PR from a cleaner branch.

PR #4135, opened 2024-02-01
Updates nx-cugraph README.md with latest algos

### LINKED ISSUE #4079: Update nx-cugraph README/docs for 24.02
The README.md file indicating the state of nx-cugraph support for NetworkX needs to be updated to reflect
* new algos added for 24.02
* planned algos for future releases

It should also be improved to show NetworkX APIs, and put less of an emphasis (perhaps just remove altogether) the PLC APIs.

PR rapidsai/cugraph#4069 will help here.
```[tasklist]
### Tasks
- [x] merge rapidsai/cugraph#4069
- [ ] update README with output from scripts in rapidsai/cugraph#4069
- [ ] Discuss and add updates for 24.04+ to README (to serve as a crude roadmap for users)
```

rlratzel:
closes #4079 

* Updates nx-cugraph `README.md` with latest algos
* Adds script to auto-update `README.md` from `nx_cugraph` metadata directly


python/nx-cugraph/nx_cugraph/scripts/print_table.py
```diff
@@ -59,7 +59,7 @@ def main(path_to_info=None, *, file=sys.stdout):
     if path_to_info is None:
         path_to_info = get_path_to_info(version_added_sep=".")
     lines = ["networkx_path,dispatch_name,version_added,plc,is_incomplete,is_different"]
-    lines.extend(",".join(info) for info in path_to_info.values())
+    lines.extend(",".join([str(i) for i in info]) for info in path_to_info.values())
```
eriknw, line 62: This is indeed the fix I mentioned, although I prefer the solution here: https://github.com/rapidsai/cugraph/pull/4108/files#diff-c0145493e321fbd8f77cf9de0c1b0889eeceaebe06c780511cc5b13646785489R62

```suggestion
    lines.extend(",".join(map(str, info)) for info in path_to_info.values())
```

python/nx-cugraph/README.md
```diff
@@ -89,48 +89,106 @@ interface to its CUDA-based graph analytics library) and
...
+| is_strongly_connected                | 24.02                    |
+| is_tree                              | 24.02                    |
+| is_weakly_connected                  | 24.02                    |
+| isolates                             | 23.10                    |
+| k_truss                              | 23.12                    |
+| karate_club_graph                    | 23.12                    |
+| katz_centrality                      | 23.12                    |
+| krackhardt_kite_graph                | 23.12                    |
+| ladder_graph                         | 23.12                    |
+| leiden                               | ?                        |
```
eriknw, line 155: Is it worth listing functions we don't yet implement? I would rather we only include ones we currently have.

python/nx-cugraph/README.md
```diff
@@ -89,48 +89,106 @@ interface to its CUDA-based graph analytics library) and
 [CuPy](https://cupy.dev/) (a GPU-accelerated array library) to NetworkX's
 familiar and easy-to-use API.
 
-Below is the list of algorithms (many listed using pylibcugraph names),
-available today in pylibcugraph or implemented using CuPy, that are or will be
+Below is the list of algorithms that are currently supported or planned to be
```
eriknw, line 92: I think it's more accurate to say _functions_, not algorithms here (and in the section header).

eriknw:
I'm not sure I like the table view as is. It feels and looks like a raw data dump and is awkward to navigate and read. Ideally, if we want a table, I would want it to be interactive--sortable, filterable, etc. (but I don't know how to do this). Multiple "simple" tables would be better than a single table, such as three tables: 1. algorithms, 2. generators, 3. other/misc. I don't like mixing these if there's no way to unmix them.

Did you try the tree view? What do you think of it?

eriknw:
`update_readme.py` feels misplaced. It probably doesn't belong in `nx_cugraph/scripts` like other scripts that could be useful for users, so maybe it ought to go in a new directory such as `scripts/update_readme.py`.

eriknw:
I don't know if we ever create and distribute Python source distributions (sdists), but I think we ought to _allow_ sdists to be created correctly. To me, this means including more files such as `update_readme.py`.

This can be done by creating `MANIFEST.in` such as:
```
include Makefile
include conftest.py
include lint.yaml
include run_nx_tests.sh
include update_readme.py
```
This does not change the contents of wheels.

acostadon (approved):
Great idea to automate this.

python/nx-cugraph/README.md
rlratzel: There was feedback that having the future functions included was good for generating interest, but maybe it's better to do that when we have a known future version they'll be a part of, or have them in a separate table.  I removed them in the latest commit.

rlratzel:
@eriknw I pushed changes based on your suggestions here and resolved a merge conflict.

We discussed bigger changes you wanted to make, which all seem good to me, but if you're not able to get them done in time we can now merge this as-is so the README is at least accurate for 24.02.

eriknw (approved):

PR #4145, opened 2024-02-05
Fix export name

vyasr:
The build export for cugraph-etl was specified as cugraph, so it wouldn't be possible to find.


vyasr:
I'm going to lump this fix in with #4144. I was hoping to keep that PR as requiring only a Python review to speed things along, but there are multiple issues that required expanding its scope anyway.

PR #4152, opened 2024-02-07
Add environment-agnostic scripts for running ctests and pytests

trxcllnt:
This PR adds environment-agnostic `run_*_{ctests,pytests}.sh` scripts, and updates `test_*_{cpp,python}.sh` to call them.

The `test_*_{cpp,python}.sh` scripts assume they're running in our CI environment, and they do more than just run the tests.

This PR allows devs and downstream consumers to only run the tests, and skip the unrelated logic in `test_*_{cpp,python}.sh`.


trxcllnt:
Seems like the C++ tests are [silently failing](https://github.com/rapidsai/cugraph/actions/runs/7812581905/job/21311057193?pr=4152#step:7:754), but I don't think it's related to the changes in this PR?

bdice:
The silent failure of C++ tests should be fixed by https://github.com/rapidsai/rapids-cmake/pull/533 which was merged two hours ago. You'll have to restart the whole workflow (not just rerun tests).

AyodeAwe (approved):

PR #4156, opened 2024-02-08
nx-cugraph: automatically generate tables in README.md

eriknw:
This updates how we create tables. Also, CI now tests that auto-generated files are up-to-date (not updating these has gotten me a couple of times).


.gitignore
```diff
@@ -78,6 +78,8 @@ datasets/*
 !datasets/karate-disjoint.csv
 !datasets/netscience.csv
 
+# nx-python side effects
```
rlratzel, line 81: Was this supposed to be `# nx-cugraph side effects`?

.pre-commit-config.yaml
```diff
@@ -17,6 +17,7 @@ repos:
         language_version: python3
         args: [--target-version=py38]
         files: ^(python/.*|benchmarks/.*)$
+        exclude: ^python/nx-cugraph/
```
rlratzel, line 20: Am I correct in assuming this means we won't have `black` formatting for all of nx-cugraph?  Was this intentional?

ci/test_python.sh
```diff
@@ -149,6 +149,13 @@ python -m nx_cugraph.scripts.print_tree --dispatch-name --plc --incomplete --dif
 python -m nx_cugraph.scripts.print_table
 popd
 
+rapids-logger "ensure nx-cugraph autogenerated files are up to date"
+pushd python/nx-cugraph
+make || true
+git diff --exit-code .
+git checkout .
+popd
+
```
rlratzel, line 158: I like this check but we should do this as part of pre-commit like we're doing for the `rapids-dependency-file-generator` code gen. This will not only ensure the README is up-to-date on commit, but the same check you're doing here will be done as part of the [CI style checks](https://github.com/rapidsai/cugraph/blob/branch-24.04/ci/check_style.sh) automatically.  The CI style check is preferred because it fails much faster and has less log output to scroll through.

python/nx-cugraph/Makefile
```diff
@@ -15,3 +15,10 @@ lint-update:
 .PHONY: plugin-info
 plugin-info:
 	python _nx_cugraph/__init__.py
+
+objects.inv:
+	wget https://networkx.org/documentation/stable/objects.inv
```
rlratzel, line 20: Rather than make this a target, should it just be done unconditionally as part of the `readme` target (and possibly removed each time)?  I'm wondering if it can go stale if left on disk unnoticed.

python/nx-cugraph/scripts/update_readme.py
```diff
@@ -0,0 +1,203 @@
...
+    text = readme_file.read()
+    tree = create_tree(get_payload=get_payload)
+    # Algorithms
+    match = re.search(
+        r"### .Algorithms(?P<preamble>.*?)<pre>\n(?P<body>.*?)\n</pre>",
+        text,
+        re.DOTALL,
+    )
+    if not match:
+        raise RuntimeError("Algorithms section not found!")
```
rlratzel, line 139: Personal preference, feel free to ignore: I like being less sensitive to the current README contents and just dumping all the generated content in a large predefined section vs. surgically inserting smaller sections. This feels more fragile for no obvious gain, but I don't feel strongly either way.

rlratzel (commented):
Thanks for the update! I checked out the new README here and it looks nicer with links for some reason `¯\_(ツ)_/¯` :)

.gitignore
eriknw: haha, yes.  oops.  fixed.

.pre-commit-config.yaml
eriknw: yes intentional, no it doesn't mean we won't have black. It means we won't use black 22.10.0 as in this file. Instead, we use black 24.1.1 as in python/nx-cugraph/lint.yaml. I prefer the linting of the more recent black.

python/nx-cugraph/Makefile
eriknw: Going stale is possible, but it probably changes very slowly the vast majority of the time. If it _is_ stale, CI will catch it.

python/nx-cugraph/scripts/update_readme.py
eriknw: The current method allows us to add more notes between the section header and the table, which I think we'll probably want to do. The current surgical regexes doesn't seem particularly fragile to me and I think will be easy to evolve.

ci/test_python.sh
eriknw: No strong feelings from me, but adding this as a pre-commit slows down pre-commit for everybody. Could this be switched to a pre-commit hook in a different PR?

.pre-commit-config.yaml
eriknw: An alternative would be to update the versions here in `.pre-commit-config.yaml`

python/nx-cugraph/Makefile
eriknw: also, since I updated this to `stable` (instead of `latest`), it should only change a couple times a year.

ci/test_python.sh
rlratzel: Yeah that's also an option that seems better than performing the check as part of a test step.  I can do this in an upcoming PR I still need to do to fix the supported python version and CUDA version.

rlratzel (approved):

AyodeAwe (approved):

PR #4158, opened 2024-02-08
Add new all-pairs similarity algorithm

ChuckHastings:
Added a new entry point for similarity functionality that combines the functionality of k_hop_nbrs and similarity.

This entry point allows us to compute similarity for all pairs of vertices in the graph in a single call. We also add the optional parameter topk which, if specified, will only return the vertices that have the highest scores. If topk is specified on an all pairs call, we compute the scores for pairs in batches and extract the topk as we go along to keep the memory footprint low.

This PR also updates a FIXME in the C++ similarity test. The C++ similarity test had been written before we had a k_hop_nbrs call, so there was some inefficient test code to compute that. Now that we have a k_hop_nbrs call, the test code was refactored to use that call.

Supersedes PR #4134


cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,302 @@ rmm::device_uvector<weight_t> similarity(
...
+
+    thrust::inclusive_scan(handle.get_thrust_policy(),
+                           two_hop_degrees.begin(),
+                           two_hop_degrees.end(),
+                           two_hop_degrees.begin());
+
+    size_t current_pos{0};
+    size_t next_pos{0};
+
+    while (true) {
```
seunghwak, line 308: We may use https://github.com/rapidsai/cugraph/blob/branch-24.04/cpp/include/cugraph/utilities/misc_utils.cuh#L40

I think vertex_t and edge_t in (template <typename vertex_t, typename edge_t>) should better be renamed, but we can get all the chunk boundaries at once with this function.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,302 @@ rmm::device_uvector<weight_t> similarity(
...
+  if (topk) {
+    //  We can reduce memory footprint by doing work in batches and
+    //  computing/updating topk with each batch
+    rmm::device_uvector<vertex_t> top_v1(0, handle.get_stream());
+    rmm::device_uvector<vertex_t> top_v2(0, handle.get_stream());
+    rmm::device_uvector<weight_t> top_score(0, handle.get_stream());
+
+    top_v1.reserve(*topk, handle.get_stream());
+    top_v2.reserve(*topk, handle.get_stream());
+    top_score.reserve(*topk, handle.get_stream());
```
seunghwak, line 252: We can defer defining/allocating these variables till right before the beginning of the while loop to minimize the variable scopes.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,302 @@ rmm::device_uvector<weight_t> similarity(
...
+                   std::make_tuple(raft::device_span<vertex_t const>{v1.data(), v1.size()},
+                                   raft::device_span<vertex_t const>{v2.data(), v2.size()}),
+                   functor,
+                   do_expensive_check);
+
+      thrust::sort_by_key(handle.get_thrust_policy(),
+                          score.begin(),
+                          score.end(),
+                          thrust::make_zip_iterator(v1.begin(), v2.begin()),
+                          thrust::greater<weight_t>{});
```
seunghwak, line 378: In case top_v1.size() >= topk (in case of multi-gpu, aggregate top_v1.size() >= topk), we know the minimum threshold to be included in topk, and we can run thrust::remove_if first to avoid sorting a large array.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,302 @@ rmm::device_uvector<weight_t> similarity(
...
+      }
+
+      thrust::copy(
+        handle.get_thrust_policy(), v1.begin(), v1.begin() + top_v1.size(), top_v1.begin());
+      thrust::copy(
+        handle.get_thrust_policy(), v2.begin(), v2.begin() + top_v1.size(), top_v2.begin());
+      thrust::copy(handle.get_thrust_policy(),
+                   score.begin(),
+                   score.begin() + top_v1.size(),
+                   top_score.begin());
```
seunghwak, line 412: In multi-GPU, shouldn't we return topk pairs in aggregate?

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,302 @@ rmm::device_uvector<weight_t> similarity(
...
+                     std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
+                     std::optional<raft::device_span<vertex_t const>> source_vertices,
+                     std::optional<size_t> topk,
+                     functor_t functor,
+                     bool do_expensive_check = false)
+{
+  using GraphViewType = graph_view_t<vertex_t, edge_t, false, multi_gpu>;
+
+  CUGRAPH_EXPECTS(graph_view.is_symmetric(),
+                  "similarity algorithms require an undirected(symmetric) graph");
```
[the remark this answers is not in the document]
ChuckHastings: I can add that.  My hope is to fix that limitation, but we know the results are somehow incorrect so rejecting it for the moment seems best.

cpp/src/link_prediction/similarity_impl.cuh
ChuckHastings: Yes.  I thought I had written that logic.  It was probably in my first attempt that I abandoned and not here.  I will add it here.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2137,6 +2137,172 @@ rmm::device_uvector<weight_t> overlap_coefficients(
...
+ * @brief     Compute Jaccard all pairs similarity coefficient
+ *
+ * Similarity is computed for all pairs of vertices.  Note that in a sparse
+ * graph, many of the vertex pairs will have a score of zero.  We actually
+ * compute similarity only for vertices that are two hop neighbors within
+ * the graph, since vertices that are not two hop neighbors will have
+ * a score of 0.
+ *
+ * If @p vertices is specified we will compute similarity on two hop
+ * neighbors the @p vertices.  If @p vertices is not specified it will
```
seunghwak, line 2150: neighbors the @p vertices=>neighbors "of" @p vertices?

cpp/include/cugraph/utilities/misc_utils.cuh
```diff
@@ -36,40 +36,40 @@ namespace cugraph {
...
-std::tuple<std::vector<vertex_t>, std::vector<edge_t>> compute_offset_aligned_edge_chunks(
+template <typename data_t, typename offset_t>
+std::tuple<std::vector<data_t>, std::vector<offset_t>> compute_offset_aligned_edge_chunks(
   raft::handle_t const& handle,
-  edge_t const* offsets,
-  vertex_t num_vertices,
-  edge_t num_edges,
+  offset_t const* offsets,
+  data_t num_offsets,
+  offset_t num_elements,
```
seunghwak, line 44: Should we use raw pointers or better switch to device_span?

And we need to double check whether the offset array size is num_vertices or num_vertices + 1 (I assume it should be num_vertices + 1). In this case, offsets array size is now num_offsets + 1. I guess this is a bit misleading.

seunghwak: What about something like
```
template <typename major_idx_t, typename minor_idx_t>
std::tuple<std::vector<major_idx_t>, std::vector<minor_idx_t>> compute_offset_aligned_compressed_sparse_array_chunks(
  raft::handle_t const& handle,
  raft::device_span<minor_idx_t> offsets,  // num_majors == offsets.size() - 1
  minor_idx_t num_minors,
  size_t approx_minor_chunk_size)
```

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -162,5 +163,390 @@ rmm::device_uvector<weight_t> similarity(
...
+    thrust::sort_by_key(handle.get_thrust_policy(),
+                        two_hop_degrees.begin(),
+                        two_hop_degrees.end() - 1,
+                        tmp_vertices.begin(),
+                        thrust::greater<size_t>{});
+
+    thrust::exclusive_scan(handle.get_thrust_policy(),
+                           two_hop_degrees.begin(),
+                           two_hop_degrees.end(),
+                           two_hop_degrees.begin());
```
seunghwak, line 304: auto two_hop_degree_offsets = std::move(two_hop_degrees)? `two_hop_degrees` is a misnomer from this point.

seunghwak (approved):
Looks good beside the below comments about code cosmetics.

cpp/include/cugraph/utilities/misc_utils.cuh
ChuckHastings: Pushing an update that makes the input a device_span.  I named it a bit differently than you suggested.  My use of the function seems (at least to me) less like major/minor, so I left things as `data_t` and `offset_t`, but I tweaked the names to get rid of the notion of vertex and edge.  Let me know what you think.

cpp/src/link_prediction/similarity_impl.cuh
ChuckHastings: OK.  I added this.  Don't use it very much from here, but it's essentially free to move and perhaps a bit clearer.

cpp/include/cugraph/utilities/misc_utils.cuh
seunghwak: So, here we are partitioning/chunking a compressed sparse array, right?

In case of the all-pairs similarity algorithm, we have seeds and their two hop degree offsets (to index an array to store each seed's two-hop neighbors).

`offset_t` sounds right, but my concern with data_t or element_t is that it is not clear whether this is to index the seeds or the two-hop neighbors.

Not sure what should be the right name for seeds or vertices in chunking compressed sparse array.

Maybe just use vertex_t as seeds are still vertices?

cpp/src/prims/per_v_transform_reduce_dst_key_aggregated_outgoing_e.cuh
```diff
@@ -355,13 +355,15 @@ void per_v_transform_reduce_dst_key_aggregated_outgoing_e(
...
-          ? (*segment_offsets)[detail::num_sparse_segments_per_vertex_partition] +
-              *(edge_partition.dcs_nzd_vertex_count())
-          : edge_partition.major_range_size(),
+        raft::device_span<edge_t const>{
+          edge_partition.offsets(),
+          static_cast<size_t>(
+            edge_partition.dcs_nzd_vertices()
+              ? (*segment_offsets)[detail::num_sparse_segments_per_vertex_partition] +
+                  *(edge_partition.dcs_nzd_vertex_count())
+              : edge_partition.major_range_size())},
```
seunghwak, line 366: Shouldn't we add + 1 to the size of the device_span?

cpp/src/sampling/sampling_post_processing_impl.cuh
```diff
@@ -741,10 +737,10 @@ renumber_sampled_edgelist(
       static_cast<size_t>(handle.get_device_properties().multiProcessorCount) *
       (1 << 20) /* tuning parameter */;  // for segmented sort
 
-    auto [h_label_offsets, h_edge_offsets] = detail::compute_offset_aligned_edge_chunks(
+    auto [h_label_offsets, h_edge_offsets] = detail::compute_offset_aligned_element_chunks(
       handle,
-      (*renumber_map_label_offsets).data(),
-      static_cast<size_t>((*renumber_map_label_offsets).size() - 1),
+      raft::device_span<size_t const>{(*renumber_map_label_offsets).data(),
+                                      (*renumber_map_label_offsets).size() - 1},
```
seunghwak, line 743: Shouldn't we remove -1 here?

ChuckHastings: Fixed

cpp/src/prims/per_v_transform_reduce_dst_key_aggregated_outgoing_e.cuh
ChuckHastings: fixed

naimnv (commented):
Sorry that I missed reviewing this interesting PR.

PR #4162, opened 2024-02-09
Add cuda-nvtx-dev, add missing CUDA library dependencies.

bdice:
This PR fixes issues with devcontainer builds where `cuda-nvtx-dev` was missing when building `libcugraph_etl`.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -327,10 +327,18 @@ dependencies:
           - matrix:
               cuda: "12.*"
             packages:
+              - cuda-cudart-dev
+              - cuda-nvtx-dev
+              - cuda-profiler-api
+              - libcublas-dev
+              - libcurand-dev
+              - libcusolver-dev
+              - libcusparse-dev
           - matrix:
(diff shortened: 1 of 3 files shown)


dependencies.yaml
```diff
@@ -327,10 +327,18 @@ dependencies:
           - matrix:
               cuda: "12.*"
             packages:
+              - cuda-cudart-dev
+              - cuda-nvtx-dev
+              - cuda-profiler-api
+              - libcublas-dev
+              - libcurand-dev
+              - libcusolver-dev
+              - libcusparse-dev
```
naimnv, line 336: Just curious, why  packages other than  `cuda-nvtx-dev` needed here?

bdice: They're build dependencies of cugraph. These are listed in the conda recipes. However, they've been coming transitively as dependencies of libraft. We should be explicit about the libraries that are used rather than relying on them coming from raft.

AyodeAwe (approved):

PR #4176, opened 2024-02-16
Removes APIs that have been deprecated and have exceeded the grace period, improves *_warning_wrapper performance 

rlratzel:
* Removes APIs that have been deprecated >1 releases ago.
* Updates tests that assumed the deprecated APIs were in place.
* Changes the `*_warning_wrapper` utilities to accept a namespace name to use in the warning messages instead of making an expensive `inspect.stack()` call.  `inspect.stack()` is still used if a namespace name was not provided.  This reduces overall `import` time, especially when used with `cudf.pandas` (attn: @shwina).
* Removes the redundant `cugraph.utilities.api_tools` module in favor of just using the equivalent `pylibcugraph` module directly.


python/pylibcugraph/pylibcugraph/utilities/api_tools.py
```diff
@@ -40,16 +47,17 @@ def experimental_warning_wrapper(obj):
...
-    calling_frame = call_stack[1].frame
-    ns_name = calling_frame.f_locals.get("__name__")
-    dot = "." if ns_name is not None else ""
+    if obj_namespace_name is None:
+        # Assume the caller of this function is the module containing the
+        # experimental obj and try to get its namespace name. Default to no
+        # namespace name if it could not be found.
+        call_stack = inspect.stack()
+        calling_frame = call_stack[1].frame
+        obj_namespace_name = calling_frame.f_locals.get("__name__")
```
naimnv, line 56: That's really cool.

naimnv (approved):
LGTM

notebooks/img/karate_similarity.png
acostadon, line 1: The original image showed two nodes that are being compared. Why is it replaced with the smaller one?

acostadon (changes requested):
The explanations helped explain the algorithm. 
Also, we have been keeping an overall history in the notebooks. 
Do you think this should be changed?

notebooks/algorithms/link_prediction/Jaccard-Similarity.ipynb
```diff
@@ -8,47 +8,16 @@
     "# Jaccard Similarity\n",
     "----\n",
     "\n",
-    "In this notebook we will explore the Jaccard vertex similarity metrics available in cuGraph. cuGraph supports:\n",
```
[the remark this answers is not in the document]
rlratzel: The "cugraph supports:" section in this notebook was referring to weighted Jaccard, which isn't being shown in the notebook anymore, so I just took it out to keep it simple.  I could mention that weighted Jaccard is supported though and that it's currently not being demonstrated here.

FYI: weighted Jaccard was removed because we no longer support the old vertex weight approach, which the notebook was still trying to show (i.e. the notebook was broken).  We instead support edge weights now, but I don't have an example of a graph with unique edge weights to properly demonstrate this.

Re: history - I was wondering if we could discuss this.  Are these tables needed?  They seem like a maintenance burden that doesn't add value (git/github tracks revision history much better), and I don't see any other notebooks in the community or even RAPIDS doing this.  For example, [here's a cudf notebook](https://github.com/rapidsai/cudf/blob/ac60656bc929c08f522d23328a515caa52cee989/docs/cudf/source/user_guide/10min.ipynb) that just looks so much easier to read.  I do think if we document timing numbers we should document the hardware used, but this notebook isn't doing that either.

notebooks/img/karate_similarity.png
rlratzel: The changes to the notebook - inspired by the changes to jaccard itself - made it harder to compare those two nodes specifically.  We could still compare them, but in the current set of nodes compared by default with the current jaccard, they won't show up as the most similar.

I changed the image since the highlighted nodes were now misleading (we're not comparing them anymore), plus I wanted to make sure the image used wasn't copyrighted.  The new image was generated.  The file is smaller, but the image is rendered in the page to be about the same size as the old one.

acostadon (approved):
thanks for explaining the changes.

BradReesWork (approved):

PR #4189, opened 2024-02-23
Adds additional detail to Jaccard notebook

rlratzel:
Followup to PR #4176 to add additional detail to the Jaccard notebook.

* Adds revision history to bottom
* Adds more detail to Jaccard description
* Adds cell output
* Adds example of using the `vertex_pair` arg


acostadon (approved):
My only comment is that usually we don't include output in the notebooks that we check in.
I see arguments for including it or not. 
Personally, as long as it doesn't make it so long as to make it difficult to read, I am fine with leaving it in.

Changes look good.

BradReesWork (approved):

Issue #4187, opened 2024-02-22
[FEA]: Cleanup Existing MG Tests


nv-rliu:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

Medium

### Please provide a clear description of problem this feature solves

A lot of our existing Python MG algorithm tests are cluttered with old ways of creating edge lists and graphs. We also use a lot of nested pytest fixtures in-place of using the @pytest.parametrize feature, which could improve clarity and reduce the amount of code needed for the same behavior.

Example:
```
# Outdated code
datasets = utils.DATASETS_UNDIRECTED + [
    utils.RAPIDS_DATASET_ROOT_DIR_PATH / "email-Eu-core.csv"
]

G = utils.generate_cugraph_graph_from_file(
        input_data_path, directed=directed, edgevals=True
    )

ddf = dask_cudf.read_csv(
        input_data_path,
        chunksize=chunksize,
        delimiter=" ",
        names=["src", "dst", "value"],
        dtype=["int32", "int32", "float32"],
    )

dg = cugraph.Graph(directed=directed)
dg.from_dask_cudf_edgelist(
        ddf,
        source="src",
        destination="dst",
        edge_attr="value",
        renumber=True,
        store_transposed=True,
    )

@pytest.fixture(scope="module")
def input_expected_output(input_combo):
   ...
```

And as a result, the tests also depend on helper functions that could be cleaned up and simplified to pytest fixtures. 

### Describe your ideal solution

The MG tests can be cleaned up using the `cugraph.datasets` API, which also now supports creation of `dask_cudf` edge lists. 
We can also use `@pytest.mark.parametrize` in place of `input_expected_output`

### Code of Conduct

- [X] I agree to follow cuGraph's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/rapidsai/cugraph/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

PR #4197, opened 2024-02-27
Refactor MG Centrality Tests

nv-rliu:
Addresses #4187 

This PR makes improvements to old testing conventions used in `cugraph.centrality` MG tests

Instead of using nested fixtures. eg. `input_expected_output -> input_combo -> fixture_params` which can be confusing, the `@pytest.mark.parametrize` marker is used to iterate through combinations of parameters used for testing. The fixtures are also replaced by functions used to create SG and MG graphs.


nv-rliu:
For anyone who happens to see this PR, some unrelated tests are failing and I'm currently investigating/reproducing this locally.

python/cugraph/cugraph/tests/centrality/test_betweenness_centrality_mg.py
```diff
@@ -33,158 +31,118 @@ def setup_function():
...
-                "subset_seed",
-                "subset_size",
-                "directed",
-                "vertex_list_type",
-            ),
-            request.param,
-        )
+
+def get_sg_graph(dataset, directed):
+    dataset.unload()
```
rlratzel, line 52: I'm curious why we need to do this here and other places where we call it before getting a graph obj.

rlratzel (commented):
Looks good but I was wondering about some of the `dataset.unload()` calls and if we need them. Do these tests have problems if the `unload()` calls aren't there, or were they added just in case?

python/cugraph/cugraph/tests/centrality/test_betweenness_centrality_mg.py
nv-rliu: These are needed because dataset objects persist their `self.edgelist` at the first time `get_edgelist` is called. Therefore, `unload` must be called if we're switching from SG usage to MG usage in order for the `dataset` object to store the correct type of edge list for its current use.

As a side-note, the weird bug I was seeing in CI was also related to the fact that certain `dataset` objects did not properly unload their internal edge-lists, so in subsequent tests the methods being called on `self.edgelist` were returning various errors.

Perhaps down the line, a more elegant solution can be implemented to remove the need for so many calls to `unload` within tests. Perhaps a flag (similar to how algorithms have `perform_expensive_check` where edgelists aren't persisted within the dataset object)

acostadon (approved):
Looks Great

python/cugraph/cugraph/tests/centrality/test_betweenness_centrality_mg.py
rlratzel: Thanks, Ralph.  Good detective work tracking that down!

rlratzel (approved):

jnke2016 (approved):
Looks good to me. And nice work tracking down that PR failure on CI

PR #4206, opened 2024-02-29
Trim trailing whitespace

bdice:
This PR trims trailing whitespace and normalizes file endings in all files, and adds pre-commit hooks to enforce this for future contributions.

These pre-commit hooks already exist in most other RAPIDS repositories, so it would be good to standardize.


jakirkham:
Wonder if we should add the final commit produced by merging this PR to `.git-blame-ignore-revs`

That could help keep the line noise down when reviewing `git blame`

https://github.blog/changelog/2022-03-24-ignore-commits-in-the-blame-view-beta/

BradReesWork (approved):

AyodeAwe (approved):

bdice:
> Wonder if we should add the final commit produced by merging this PR to `.git-blame-ignore-revs`

I think +/- 700 lines isn't enough to bother. There are larger PRs than this on a frequent basis. But feel free to file a follow-up PR after this merges and the commit hash is known, if you would like. (Maybe include #4205, too.)

ChuckHastings (approved):

PR #4215, opened 2024-03-05
Fix `louvain_communities`, `PropertyGraph`, cudf `column.full`, dgl 2.1.0 CI failures

eriknw:
* Updates cudf usage in hypergraph to use `cudf.core.column.as_column` instead of the deprecated `cudf.core.column.full` API
* Adds code to call louvain using a signature that's compatible with both pre and post NX 3.3
* Adds an upper bound to the DGL version pin to ensure a compatible version is used
* Updates a PropertyGraph test to handle minor ordering changes in the result DataFrame
* Test matrices starting 24.04 have changed, resulting in an "empty matrix" error when attempting to use the supported combination used for PyG (PyG wheel tests need pytorch with CUDA 12.2 on arm64), so the matrix was simplified to remove the CUDA 11.8 requirement. cc @tingyu66

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -162,7 +162,7 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_wheel_cugraph-pyg.sh
-      matrix_filter: map(select(.ARCH == "amd64" and .CUDA_VER == "11.8.0"))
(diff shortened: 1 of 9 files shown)


raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #4216, opened 2024-03-05
Treat nx-cugraph, cugraph-pyg, cugraph-dgl, and cugraph-equivariant CI artifacts as pure wheels

bdice:
## Description
This marks `nx-cugraph`, `cugraph-dgl`, `cugraph-pyg`, and `cugraph-equivariant` as pure wheels, meaning that the CI artifacts are not specific to a Python version or CPU architecture. This change depends on https://github.com/rapidsai/gha-tools/pull/96, and makes CI workflows more robust by allowing the test matrix to be separated from the build matrix.

### DIFF
--- a/ci/build_wheel.sh
+++ b/ci/build_wheel.sh
@@ -54,12 +54,12 @@ cd "${package_dir}"
 
 python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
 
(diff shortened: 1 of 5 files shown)


bdice:
Depends on #4215.

raydouglass (approved):

PR #4218, opened 2024-03-06
Use public pandas and cudf APIs where possible

mroeschke:
Avoiding `core` imports here as they tend to be "private"


rlratzel (approved):
Thank you!

PR #4223, opened 2024-03-08
Enable pytest failures on warnings from cudf

mroeschke:
In 24.04, cudf issues a FutureWarning when using positional indexing with `__getitem__`. The change here is to use `iloc` instead.

Additionally the test suite will now fail when a warning comes from cudf as apart of https://github.com/rapidsai/build-planning/issues/26


python/cugraph/pytest.ini
acostadon, line 1: License update needed

acostadon (changes requested):
What is the schedule to merge this?
Is it for 24.04?

mroeschke:
> What is the schedule to merge this? What is the schedule to merge this?
Is it for 24.04?

This is not critical for 24.04. Fine to de-prioritize over other PRs

rlratzel (approved):
Thank you very much!
We should have treated cudf warnings as test failures long ago.

BradReesWork:
All - can we get this is as soon as possible.  while the warnings do not stop an analytic, is still looks bad on us to even have warrings

acostadon (approved):
Looks great !!!

BradReesWork:
@mroeschke there is a merge conflict

PR #4225, opened 2024-03-08
Updates nx-cugraph `README` for latest h/w, CUDA, python, NX requirements, moves updater to pre-commit

rlratzel:
* Updates nx-cugraph `README` for latest h/w, CUDA, python, NX requirements
* Moves the call to the nx-cugraph `README` check and auto-updater script from the python CI test script to a pre-commit hook alongside the other codegen calls/checks.  The pre-commit checks are run by the dev at commit-time and by CI as part of the style check job.  This check will only run for changes to files under the `nx-cugraph` subdir.
* Updates the `update_readme.py` script to optionally download the NX `objects.inv` file itself so calling `make` is not required as part of the pre-commit hook.
  * _NOTE: This is less important though, since the pre-commit hook still calls `bash` in order to set the `PYTHONPATH` env var to use the local version of the `update_readme.py` script, which is what I was trying to avoid in the first place. This still has some advantages in making the `update_readme.py` script more usable if the user isn't aware of how to obtain a `objects.inv` file, but feedback is welcome. The hook could also be changed to just call `make` if that ends up being preferred, in which case we'd have to consider if we still want the updates to `update_readme.py` to download the `objects.inv` file itself or not._

Here's the check running as part of the style-check job for this PR:
image


python/nx-cugraph/Makefile
```diff
@@ -16,9 +16,6 @@ lint-update:
 plugin-info:
 	python _nx_cugraph/__init__.py
 
-objects.inv:
-	wget https://networkx.org/documentation/stable/objects.inv
-
 .PHONY: readme
-readme: objects.inv
-	python scripts/update_readme.py README.md objects.inv
```
rlratzel, line 24: @eriknw I removed the `objects.inv` download from the Makefile since it is no longer necessary, but I can't remember if you said you preferred to keep it or not.  The script will still accept an externally-provided `objects.inv` file specified as an arg, so restoring the Makefile will still work as intended.

rlratzel: ...and of course, running `make readme` will still work as expected.

Running `make readme` and the pre-commit hook will not download an `objects.inv` file if one already exists, so in both cases downloads only happen when necessary.

eriknw: I don't have a strong opinion. The nice thing about having a makefile target is it explicitly captures state and dependencies. Otherwise, this `objects.inv` will automagically appear when running `make`, which may be surprising.

rlratzel: I restored `Makefile` in the latest commit.

AyodeAwe (approved):

python/nx-cugraph/README.md
```diff
@@ -7,11 +7,10 @@ to run supported algorithms with GPU acceleration.
 ## System Requirements
 
 nx-cugraph requires the following:
-
- * NVIDIA GPU, Pascal architecture or later
+ * NVIDIA GPU, Volta architecture or later, with [compute capability](https://developer.nvidia.com/cuda-gpus) 7.0+
  * CUDA 11.2, 11.4, 11.5, 11.8, or 12.0
- * Python versions 3.9, 3.10, or 3.11
- * NetworkX >= version 3.2
+ * Python version 3.10 or 3.11
```
eriknw, line 12: Why drop Python 3.9? We still support Python 3.9, which is usable with older networkx versions. Even NetworkX 3.2 supports Python 3.9.

ci/test_python.sh
```diff
@@ -127,13 +127,6 @@ python -m nx_cugraph.scripts.print_tree --dispatch-name --plc --incomplete --dif
 python -m nx_cugraph.scripts.print_table
 popd
 
-rapids-logger "ensure nx-cugraph autogenerated files are up to date"
-pushd python/nx-cugraph
-make || true
-git diff --exit-code .
-git checkout .
-popd
-
```
eriknw, line 136: This check can still serve a purpose. The README is not the only thing that gets automatically updated. `make plugin-info` is used to update `_nx_cugraph/__init__.py`, and this emphatically should _not_ be a pre-commit in a dev environment, because a dev environment may be using a different version of networkx than CI uses. Running `make` also performs linting, which is a nice-to-check.

python/nx-cugraph/scripts/update_readme.py
```diff
@@ -190,14 +193,36 @@ def get_payload_internal(keys):
...
+
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         "Update README.md to show NetworkX functions implemented by nx-cugraph"
     )
     parser.add_argument("readme_filename", help="Path to the README.md file")
     parser.add_argument(
-        "networkx_objects", help="Path to the objects.inv file from networkx docs"
+        "networkx_objects", nargs="?", default=None,
+        help="Path to the objects.inv file from networkx docs. Optional."
```
eriknw, line 217: Minor nit: it would be nice to say that it's optional b/c it will search for and download the file if necessary.

eriknw (commented):
Thanks Rick. It's nice to move what we can into pre-commit.

Any appetite for updating linting in pre-commit to better match what we have in `nx-cugraph`?

Overall looks good, but I shared a few minor comments and nits.

python/nx-cugraph/README.md
rlratzel: Oops, you're right, that was an oversight.  Fixed.

ci/test_python.sh
rlratzel: Thanks for pointing that out.  I was able to add an additional hook to do the plugin-info update too.  I also agree about the need to isolate a known version of NetworkX to use for these codegen steps from the dev env, so I added a specific NetworkX dependency to the hook environment, which should guarantee consistency.  Let me know what you think, and I can still revert this if necessary since the only requirements for this PR are the README updates.

python/nx-cugraph/scripts/update_readme.py
rlratzel: Agreed, and I added text to explain the behavior better.

ci/test_python.sh
rlratzel: But, I guess we still don't have the same level of linting.  I did update `_nx_cugraph/core.py` to generate black-compatible strings, which now pass `black`, but this is still not running `ruff`, `isort`, etc.

> Any appetite for updating linting in pre-commit to better match what we have in nx-cugraph?

I like the idea of the nx-cugraph level of linting though.  I can either open an issue for 24.06 to update the cugraph pre-commit to match, or I can add `make lint` to the `test_python.sh` script (at least until cugraph catches up).  What do you think?

rlratzel: We discussed this offline and decided to file [this issue](https://github.com/rapidsai/cugraph/issues/4234) rather than change this PR.

python/nx-cugraph/_nx_cugraph/core.py
```diff
@@ -45,18 +45,27 @@ def update_text(text, lines_to_add, target, indent=" " * 8):
     return f"{text[:start]}{begin}{to_add}\n{indent}{text[stop:]}"
 
 
+def dq_repr(s):
+    """Return repr(s) quoted with the double quote preference used by black."""
+    rs = repr(s)
+    if rs.startswith("'") and '"' not in rs:
+        rs = rs.strip("'").replace('"', '\\"')
```
eriknw, line 52: Why do we need `.replace('"', '\\"')`? We skip this if `'"' not in rs`
```suggestion
        rs = rs.strip("'")
```

acostadon (approved):
Looks great.

eriknw (approved):

PR #4227, opened 2024-03-09
Make external vertex and edge shuffling function public

naimnv:
Make vertex and edge shuffling function public


cpp/include/cugraph/graph_functions.hpp
```diff
@@ -1052,4 +1052,60 @@ remove_multi_edges(raft::handle_t const& handle,
                    std::optional<rmm::device_uvector<edge_type_t>>&& edgelist_edge_types,
                    bool keep_min_value_edge = false);
 
+/**
+ * @brief Shuffle external vertex ids and values to the proper GPU.
+ * *
```
seunghwak, line 1057: Better document the supported value types.

And sorry for nitpicking, but it might be more natural to list `shuffle_external_vertices` first and `shuffle_external_vertices_and_values` second.

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -1052,4 +1052,60 @@ remove_multi_edges(raft::handle_t const& handle,
...
+ * GPU.
+ */
+template <typename vertex_t, typename weight_t>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<weight_t>>>
+shuffle_external_edges(raft::handle_t const& handle,
+                       rmm::device_uvector<vertex_t>&& edge_srcs,
+                       rmm::device_uvector<vertex_t>&& edge_dsts,
+                       std::optional<rmm::device_uvector<weight_t>>&& edge_weights);
```
seunghwak, line 1109: To match with `create_graph_from_edgelist`, we may better add `edge_ids` & `edge_types`.

cpp/src/detail/shuffle_vertices.cu
```diff
@@ -249,4 +249,74 @@ shuffle_ext_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
...
+{
+  return detail::shuffle_ext_vertices_to_local_gpu_by_vertex_partitioning(handle,
+                                                                          std::move(vertices));
+}
+
+template rmm::device_uvector<int32_t> shuffle_external_vertices(
+  raft::handle_t const& handle, rmm::device_uvector<int32_t>&& d_vertices);
+
+template rmm::device_uvector<int64_t> shuffle_external_vertices(
+  raft::handle_t const& handle, rmm::device_uvector<int64_t>&& d_vertices);
```
seunghwak, line 320: It will be more natural to list `shuffle_external_vertices` before `shuffle_external_vertices_and_values`

cpp/include/cugraph/graph_functions.hpp
seunghwak: And `vertex_value_pairs` is more consistent than `vertices_and_values` in our codebase.

seunghwak (commented):
Looks good to me in general but I have few minor suggestions about code cosmetics.

cpp/src/detail/shuffle_vertex_pairs.cu
```diff
@@ -519,4 +519,69 @@ shuffle_int_vertex_pairs_with_values_to_local_gpu_by_edge_partitioning(
   std::vector<int64_t> const& vertex_partition_range_lasts);
 
 }  // namespace detail
+
```
ChuckHastings, line 522: It might make sense to move this file out of `src/detail`.  Everything in this directory is in the detail namespace, if we're exposing it to the public namespace then we might want to move the source file to `src/utilities`.

cpp/src/detail/shuffle_vertices.cu
```diff
@@ -249,4 +249,74 @@ shuffle_ext_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
   rmm::device_uvector<double>&& values);
 
 }  // namespace detail
+
```
ChuckHastings, line 252: Same comment here

cpp/include/cugraph/graph_functions.hpp
naimnv: Added

cpp/src/detail/shuffle_vertices.cu
naimnv: Changed.

cpp/include/cugraph/graph_functions.hpp
naimnv: Changed.

ChuckHastings (approved):

PR #4228, opened 2024-03-09
Adds nx-cugraph benchmarks for APIs added to prior releases that were never benchmarked

rlratzel:
* Adds benchmark for nx-cugraph `pagerank` with a personalization dict (see results below)
* Adds several placeholder benchmarks for other nx-cugraph APIs added to prior releases that were never benchmarked.

image


eriknw (approved):
LGTM.

What would you think of adding metadata to benchmarks so we know which algorithms they benchmark? This could answer the more important question of what algorithms _are not_ currently benchmarked.

PR #4229, opened 2024-03-10
remove NVTX calls in libcugraph_etl

ChuckHastings:
NVTX calls are not necessary for this prototype capability.

### DIFF
--- a/cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
+++ b/cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
@@ -25,7 +25,6 @@
 #include <hash/helper_functions.cuh>
 #include <hash/managed.cuh>
 
-#include <cudf/detail/nvtx/ranges.hpp>
 #include <cudf/detail/utilities/device_atomics.cuh>
 #include <cudf/hashing/detail/default_hash.cuh>
 #include <cudf/hashing/detail/hash_functions.cuh>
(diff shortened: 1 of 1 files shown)


naimnv (approved):
LGTM

bdice (approved):
This should work as long as none of the other cudf `detail` headers being included have any NVTX references. If this passes CI, it should be good to merge.

cpp/libcugraph_etl/include/hash/concurrent_unordered_map.cuh
```diff
@@ -25,7 +25,6 @@
 #include <hash/helper_functions.cuh>
 #include <hash/managed.cuh>
 
-#include <cudf/detail/nvtx/ranges.hpp>
 #include <cudf/detail/utilities/device_atomics.cuh>
```
naimnv, line 28: cudf/detail header is used here. Doesn't it require changes in CCmakelistList.xt? 
 ` target_link_libraries(your_target PRIVATE nvtx3-cpp)`
 https://github.com/rapidsai/cudf/pull/15178#issue-2159770108

bdice: If there are any NVTX references, then we'd need this to link to `nvtx3-cpp`. However, if the (small number of) NVTX references are removed, then we don't need to alter the CMake at all.

https://github.com/rapidsai/cugraph/pull/4229#pullrequestreview-1926583836

naimnv: Thanks for clarifying. I thought If we include any cudf/detail header, in this case cudf/detail/utilities/device_atomics.cuh, would require to link to nvtx3-cpp.

bdice: If those cudf/detail headers include NVTX, then yes. But the cudf/detail headers in this file are not using NVTX, so linking to NVTX isn't necessary (we know that this is a good solution because C++ CI jobs are passing 😄).

seunghwak (approved):
LGTM and thanks for the quick fix!!!

Issue #4236, opened 2024-03-13
MTMG Use of streams


ChuckHastings:
Review MTMG use of streams.  We allocate a stream pool for MTMG to allow for different host threads to copy data to the same GPU in parallel.  We have a couple of different ways we access streams which sometimes make it easy to grab the wrong stream.

PR #4237, opened 2024-03-13
MTMG gather use of streams

ChuckHastings:
`mtmg::vertex_result_t::gather` was causing intermittent flaky results.

A couple of issues were identified:
* Stream use was incorrect.  We were using a stream from the stream pool AND the default stream.  Modified to use the default stream throughout to resolve the issue.  Created #4236 to design a long-term solution.
* Customer use case was trying to gather results from vertices that didn't exist.  Added a mechanism to allow for that and return a default value for vertices that don't exist.

Closes https://github.com/rapidsai/graph_dl/issues/464


seunghwak (commented):
LGTM

seunghwak (approved):

naimnv (approved):
LGTM.
Quite a bit of bookkeeping to get things right :)

PR #4244, opened 2024-03-15
Refactor Several MG Tests

nv-rliu:
Addresses #4187 

This PR makes improvements to old MG testing conventions used in these testing directories:
 - centrality
 - comms
 - community
 - components
 - core
 - internals

The goals of this PR is to improve readability and use of the MG tests. Changes are similar to #4197 
> Instead of using nested fixtures. eg. `input_expected_output -> input_combo -> fixture_params` which can be confusing, the `@pytest.mark.parametrize` marker is used to iterate through combinations of parameters used for testing. The fixtures are also replaced by functions used to create SG and MG graphs.

By using the Datasets API (which now supports MG Graphs thanks to @huiyuxie), the number of imports and `testing.utils` functions can be significantly reduced.


huiyuxie:
Thanks for mentioning me, Ralph! :) I did a rough review and please find it above. 

It’s so great to see nice and clean code coming in!

rlratzel (approved):
LGTM, thanks for all the cleanup!

PR #4260, opened 2024-03-19
Always scatter in FP32 in TPConv

tingyu66:
Always scatter in FP32 in TPConv due to poor performance of torch's native scatter_add in bfloat16.


alexbarghi-nv (approved):
👍

PR #4261, opened 2024-03-20
Update triangle count to internally use edge masking and update Triangle counting & K-hop neighbor tests to include edge masked graph view inputs

seunghwak:
This PR pulls updates from #4253
Better be reviewed after #4253 is merged.

* Update triangle count to internally use edge masking (this is cheaper than creating a new graph in both memory footprint and compute) to
  * include only the edges with endpoints reachable within two hops (if the list of vertices to compute triangle counts is provided)
  * Exclude self-loops
  * Exclude vertices that do not belong to 2-Core
  * Note that we are still creating a new graph when we are creating a degree ordered directed graph (as this graph is now significantly smaller than the original graph and edge masking has non-negligible overhead and computing triangle counts on the final degree ordered directed graph is computationally intensive).
* Update SG/MG Triangle counting & K-hop neighbors C++ tests to include edge masked graph view inputs


cpp/tests/community/triangle_count_test.cpp
```diff
@@ -278,7 +290,10 @@ INSTANTIATE_TEST_SUITE_P(
   Tests_TriangleCount_Rmat,
   ::testing::Combine(
     // enable correctness checks
-    ::testing::Values(TriangleCount_Usecase{0.1}, TriangleCount_Usecase{1.0}),
+    ::testing::Values(TriangleCount_Usecase{0.1, false},
+                      TriangleCount_Usecase{0.1, true},
+                      TriangleCount_Usecase{1.0, false},
+                      TriangleCount_Usecase{1.0, true}),
```
naimnv, line 296: Why correctness checks are not either all false or all true?

cpp/tests/community/triangle_count_test.cpp
```diff
@@ -302,7 +320,10 @@ INSTANTIATE_TEST_SUITE_P(
   Tests_TriangleCount_Rmat,
   ::testing::Combine(
     // disable correctness checks for large graphs
-    ::testing::Values(TriangleCount_Usecase{0.1, false}, TriangleCount_Usecase{1.0, false}),
+    ::testing::Values(TriangleCount_Usecase{0.1, false, false},
+                      TriangleCount_Usecase{0.1, true, false},
+                      TriangleCount_Usecase{1.0, false, false},
+                      TriangleCount_Usecase{1.0, true, false}),
```
naimnv, line 326: For Larger graphs, I understand we want to have them all false.

jnke2016 (approved):

cpp/tests/community/triangle_count_test.cpp
seunghwak: See
```
struct TriangleCount_Usecase {
  double vertex_subset_ratio{0.0};

  bool edge_masking{false};
  bool check_correctness{true};
};
```
This has 3 member variables, here I am providing only two, so the third one (check_correctness) is set to the default value (which is false).

cpp/tests/community/triangle_count_test.cpp
seunghwak: Yes, for benchmarking, we disable correctness check.

cpp/tests/community/triangle_count_test.cpp
naimnv: Ah okay.

naimnv (approved):
LGTM

ChuckHastings (approved):

PR #4275, opened 2024-03-25
Fix stream synchronization in MTMG graph construction

### LINKED ISSUE #4236: MTMG Use of streams

ChuckHastings:
Restructure to pass stream instead of handle and synchronize appropriately.

Closes #4236


cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
```diff
@@ -62,13 +62,13 @@ class per_device_edgelist_t {
...
    *
-   * @param handle             MTMG resource handle - used to identify GPU resources
+   * @param stream_view        CUDA stream view
    * @param device_buffer_size Number of edges to store in each device buffer
    * @param use_weight         Whether or not the edgelist will have weights
    * @param use_edge_id        Whether or not the edgelist will have edge ids
    * @param use_edge_type      Whether or not the edgelist will have edge types
    */
-  per_device_edgelist_t(cugraph::mtmg::handle_t const& handle,
+  per_device_edgelist_t(rmm::cuda_stream_view stream_view,
```
seunghwak, line 71: Something tedious, but I guess our convention is to pass the handle as the first input argument and stream_view as the last input argument. Should we better keep this convention?

seunghwak (approved):
LGTM

cpp/include/cugraph/mtmg/detail/per_device_edgelist.hpp
ChuckHastings: Yes, I should have caught that.  I will make that change.

PR #4278, opened 2024-03-25
[FEA] cuGraph GNN NCCL-only Setup and Distributed Sampling

### LINKED ISSUE #3851: [IMP] Default Seeds Per Call Should be Proportional to the Number of GPUs
Right now the number of seeds per call is set to 200,000 by default, which is low and will result in less than optimal performance.  The default should at least be proportional to the number of GPUs so that users can have a logical default for that argument.  The average user won't know how to set `seeds_per_call` or what it does.

### LINKED ISSUE #4200: [FEA] Research Removing Dask Reliance in GNN Packages
Currently, we use Dask to manage MNMG processing throughout RAPIDS/cuGraph, but this causes some issues when integrating with PyTorch DDP workflows.

The biggest issue is that RMM pools can't be shared across processes, which means when using pools, each Dask worker would have a separate pool from the corresponding DDP worker on the same GPU.  This wastes a significant amount of memory, and so in most cases, we don't use pools.

Another issue is that the semantics of Dask make managing loaders awkward.  Originally, this wasn't supposed to be the case, but due to #4089 , we can't have multiple loaders calling `uniform_neighbor_sample`.  And furthermore, even if that did work, it is a poor use of GPU resources when we should be combining all of those loader calls.

Finally, managing Dask and DDP in the same workflow is complicated and challenging.  Other teams have complained that this makes the examples difficult to understand for new users, or even unreadable to someone unfamiliar with Dask.  For MNMG workflows, we have to start Dask in a separate process which clashes with the PyTorch way of doing things.

There are also a number of new planned features (i.e. overlapping sampling/loading) that would be simpler to implement without Dask.

This path would not be unique in RAPIDS; WholeGraph already does something similar, relying on DDP as the process manager, and using its own RAFT/NCCL comms.  cuGraph should be able to leverage RAFT and PyLibcuGraph to do the same.

### LINKED ISSUE #4201: [BUG] Bulk Sampler Batch Distribution is Uneven
Currently, the bulk sampler naively divides output batches across _N_ files based on the `batches_per_partition` parameter.  But this nearly always results in uneven batch distribution, which is theoretically acceptable, but in practice, leads to significant waste of GPU resources since a few workers get a much higher number of batches.

This problem compounds when the number of batches per worker is required to be equal in order to prevent a hang.  `model.join` is supposed to prevent this problem, but for many workflows, it does not work.  So the only solution is to throw out some batches.  Depending on how batches were distributed across partitions, this could result in a lot of batches being thrown out, leading to lower training accuracy.

The native GNN frameworks do not have this issue since each batch is tied to a worker.  A good workaround could be to make `batches_per_partition` affect only batches on a per-worker basis.  So if we had 27 output batches and 3 workers, we would first divide the batches across the workers (9 batches per worker).  Then we would use the `batches_per_partition` parameter (4) to create the final partitions, resulting in partitions of (4, 4, 1) per worker and a total of 9 partitions.  Although there are now more partitions (9 vs. 7), we better utilize GPU resources and ensure no batch is dropped.

### LINKED ISSUE #4246: [FEA] New Distributed Sampler for GNN Packages
The current cuGraph `BulkSampler` class relies heavily on dask, and forces the `cuGraph-DGL` and `cuGraph-PyG` packages to manage the graph on rank 0, which causes _many_ breakages of standard `PyTorch`/`DGL`/`PyG` features.  It is also unaligned with the new `PyG` distributed sampling framework, as well as standard `PyTorch DDP` best practices.

This new sampler should also resolve #4201 as it will force output minibatches to remain on the node that generated them, and it may also fix some of the issues where output accuracy is (falsely) reported as lower than expected.

This is step 1 in constructing a new GNN package workflow that does not include Dask (#4200).

alexbarghi-nv:
* Adds the ability to run `pylibcugraph` without UCX/dask within PyTorch DDP.
* Adds the new distributed sampler which uses the new nccl+ddp path to perform bulk sampling.

Closes #4200 
Closes #4201 
Closes #4246 
Closes #3851


ChuckHastings (commented):
Seems alright to me.  @seunghwak is a bit more familiar with some of the sampling output functions.

python/cugraph-pyg/cugraph_pyg/examples/cugraph_dist_sampling_mg.py
```diff
@@ -0,0 +1,111 @@
...
+
+    device = rank
+    cugraph_comms_init(rank, world_size, uid, device)
+
+    print(f"rank {rank} initialized cugraph")
+
+    src = cudf.Series(np.array_split(edgelist[0], world_size)[rank])
+    dst = cudf.Series(np.array_split(edgelist[1], world_size)[rank])
+
+    seeds = cudf.Series(np.arange(rank * 50, (rank + 1) * 50))
```
seunghwak, line 61: Just minor nitpicking suggestions.

What is 50? # seeds per rank? If this is an example,

```
num_seeds_per_rank = 50
seeds = cudf.Series(np.arange(rank * num_seeds_per_rank, (rank + 1) * num_seeds_per_rank))
```

will be more readable.

And just for completeness. Are we assuming that # ranks * # seeds per rank > # vertices in the input graph? If this code assumes anything about the input `edgelist`, we may specify that in comments or we may give a check.

python/cugraph-pyg/cugraph_pyg/examples/cugraph_dist_sampling_sg.py
```diff
@@ -0,0 +1,79 @@
...
+from pylibcugraph import SGGraph, ResourceHandle, GraphProperties
+
+from ogb.nodeproppred import NodePropPredDataset
+
+
+def sample(edgelist, directory):
+    src = cudf.Series(edgelist[0])
+    dst = cudf.Series(edgelist[1])
+
+    seeds = cudf.Series(np.arange(0, 50))
```
seunghwak, line 41: Same here. What happens if `edgelist` is empty?

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
```diff
@@ -0,0 +1,92 @@
...
+        return ni
+    except Exception as ex:
+        raise RuntimeError(f"A nccl error occurred: {ex}")
+
+
+def make_raft_handle(
+    rank, world_size, nccl_comms, n_streams_per_handle=0, verbose=False
+):
+    handle = Handle(n_streams=n_streams_per_handle)
+    inject_comms_on_handle_coll_only(handle, nccl_comms, world_size, rank, verbose)
```
seunghwak, line 41: So, we don't need p2p?

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
```diff
@@ -0,0 +1,92 @@
...
+
+    return handle
+
+
+def __get_2D_div(ngpus):
+    prows = int(math.sqrt(ngpus))
+    while ngpus % prows != 0:
+        prows = prows - 1
+    return prows, int(ngpus / prows)
```
seunghwak, line 50: Just FYI, maybe in the future, we may use a common C++ utility function to compute a desirable 2D partition (with an user option to override the default). Setting prows as close as possible to math.sqrt(ngpus) is just one possibility.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+            edge_type_array_p = (
+                minibatch_dict["edge_type"][start_ix:end_ix]
+                if has_edge_types
+                else cupy.array([], dtype="int32")
+            )
+            weight_array_p = (
+                minibatch_dict["weight"][start_ix:end_ix]
+                if has_weights
+                else cupy.array([], dtype="float32")
+            )
```
seunghwak, line 107: So, are we assuming that edge IDs are always int64, edge types are always int32 (this may make sense), and edge weights are always float32?

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+        self.__writer = writer
+        self.__local_seeds_per_call = local_seeds_per_call
+
+    def sample_batches(
+        self, seeds: TensorType, batch_ids: TensorType, random_state: int = 0
+    ):
+        raise NotImplementedError("Must be implemented by subclass")
+
+    def sample_from_nodes(
+        self, nodes: TensorType, *, batch_size: int = 16, random_state: int = 62
```
seunghwak, line 186: Is random state a seed for random number generation? If yes, make sure that you are providing different seeds for different ranks in random number generation. If not and for deterministic random number generators, you are generating same random number sequences in every GPU.

seunghwak: In C++, we take base_seed and use base_seed + rank in random number generator initialization.

seunghwak (commented):
Looks good to me in general, I have some minor suggestions and questions.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
VibhuJawa, line 186: Can we a docstring here.  Hard to understand the inputs of this function. Are nodes all the seed we need to sample, if yes wont the below lines at 216 be very slow ? 

```
for i, current_seeds in enumerate(nodes):
```

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+                warnings.warn(
+                    "Not all ranks received the same number of batches. "
+                    "This might cause your training loop to hang due to uneven inputs."
+                )
+
+            batch_id_start = t.cumsum(dim=0)[rank] - t[0]
+        else:
+            batch_id_start = 0
+
+        for i, current_seeds in enumerate(nodes):
```
VibhuJawa, line 216: Wont this be slow and  is this a singular node ? 
```suggestion
        for i, current_seed in enumerate(nodes):
```

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+    def sample_batches(
+        self, seeds: TensorType, batch_ids: TensorType, random_state: int = 0
+    ):
+        # FIXME allow skipping of the synchronization logic with a boolean
+        # flag that assumes all ranks have the same number of batches.
+        if self.is_multi_gpu:
+            rank = torch.distributed.get_rank()
+            handle = pylibcugraph.ResourceHandle(
+                cugraph_comms_get_raft_handle().getHandle()
+            )
```
VibhuJawa, line 279: We should probably cache this and not get it for every call ?

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+
+            local_label_list = torch.unique(batch_ids)
+            local_label_to_output_comm_rank = torch.full(
+                (len(local_label_list),), rank, dtype=torch.int32, device="cuda"
+            )
+
+            num_batches = torch.tensor(
+                [len(local_label_list)], device="cuda", dtype=torch.int64
+            )
+            torch.distributed.all_reduce(num_batches, op=torch.distributed.ReduceOp.SUM)
```
VibhuJawa, line 289: Strongly suggest making functions for gatherig num_batches, label_list etc

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+            )
+            w2 = torch.distributed.all_gather_into_tensor(
+                label_to_output_comm_rank,
+                local_label_to_output_comm_rank,
+                async_op=True,
+            )
+
+            w1.wait()
+            w2.wait()
+
```
VibhuJawa, line 307: Should we have a synchronization call here ? Given above are multi-node operations , this will probably lag at different places and the  `pylibcugraph.uniform_neighbor_sample` can probably hang when there is too much time b/w two calls . 

Or at-least it used to, will let @jnke2016 confirm about this as he has tested this the most.

VibhuJawa (changes requested):
Thanks for working on this @alexbarghi-nv .  

Left an initial review .

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
alexbarghi-nv: I don't think so.  As far as I can now we can run all the GNN algorithms without UCX p2p comms.

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
alexbarghi-nv: I copied this from the existing dask config.  I would be happy to replace this with a better function if that becomes available.  We can make an issue for it.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: I think C++ only supports int32 edge types.  For edge id, the frameworks always use int64.  In any case, this does actually allow for int32 edge ids, we just have to set a default empty dtype, and I just went with int64.

alexbarghi-nv: Same for edge weights, I'm not technically restricting them to float32 here.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: Nodes is a list of tensors.  Except it's actually a collection of zero-copy views of the original tensor created using `torch.split`.  Each tensor in nodes is the input for one sampling call.  So this isn't any slower than the bulk sampler was.  In fact, I think this approach is slightly more efficient.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,349 @@
...
+        if self.is_multi_gpu:
+            rank = torch.distributed.get_rank()
+            world_size = torch.distributed.get_world_size()
+
+            t = torch.empty((world_size,), dtype=torch.int64, device="cuda")
+            local_size = torch.tensor(
+                [int(ceil(num_seeds / batch_size))], dtype=torch.int64, device="cuda"
+            )
+
+            torch.distributed.all_gather_into_tensor(t, local_size)
```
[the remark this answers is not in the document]
alexbarghi-nv: We need to know where the batch ids start for the rank.  Otherwise we'll get multiple ranks using the same batch ids, which will cause invalid samples to be produced.  So this all gather gets the number of batches for each rank, and figures out the starting batch id for this rank by doing a cumulative sum (line 212).

I think I may split this into a separate function.  Also I may add a way to skip this logic if you're 100% confident all ranks have the same number of batches (I have a FIXME in line 273 to do the same thing in another part of the code)

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: `current_seeds` is a view of all the seeds.  It contains the seeds for the current sampling call.  This is not slow.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: I'll add a function for this.

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
VibhuJawa: I thought we initialized `p2p` comms in our benchmarks .

https://github.com/rapidsai/cugraph/blob/branch-24.04/benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py#L838


>  without UCX p2p comms.

Are we configuring UCX here? I thought this was for NCCL

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: Agreed, and I need to add the flag to skip this logic.

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
alexbarghi-nv: We're only using nccl here.  No UCX.  I think that's sufficient for what we're doing.

alexbarghi-nv: At least that's what @ChuckHastings told me.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
VibhuJawa: Can we add a docstring to reflect that please and update the type hint accordingly.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: Doesn't that only apply if this is launched from a dask worker, which can die after a certain time?  I thought there is an implicit synchronization that occurs in the C++ code right before the call returns.  All the calls should end at roughly the same time.

python/cugraph-pyg/cugraph_pyg/examples/cugraph_dist_sampling_mg.py
alexbarghi-nv: I think you mean < # vertices?  But this is really just meant to be a toy example.  It's using a dataset that has far more than that number of vertices.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
VibhuJawa: > which can die after a certain time? 

I think the scenario you are referring to timing out after hanging (which is actually a better practice for distributed frameworks IMO  and we should follow)  

>is an implicit synchronization that occurs in the C++ code right before the call returns. 
I am referring to before the `pylibcugraph.uniform_neighbor_sample`, will let @jnke2016 confirm about his experiments

python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
seunghwak: OK, yeah, so p2p here means UCX p2p. NCCL has P2P as well. I think the naming here is confusing. NCCL-only might be more appropriate.

seunghwak (approved):
LGTM

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: Also changed the variable name instead of overwriting

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
alexbarghi-nv: Done.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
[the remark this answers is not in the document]
alexbarghi-nv: Done.

alexbarghi-nv:
I'm planning to add tests before merging, and also adding the scripts that run our examples to CI.

rlratzel (approved):
LGTM, thanks for the new tests and examples.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,577 @@
...
+from typing import Union, List, Dict, Tuple
+from cugraph.utilities import import_optional
+from cugraph.gnn.comms import cugraph_comms_get_raft_handle
+
+from cugraph.gnn.data_loading.bulk_sampler_io import create_df_from_disjoint_arrays
+
+# PyTorch is NOT optional but this is required for container builds.
+torch = import_optional("torch")
+
+TensorType = Union[torch.Tensor, cupy.ndarray, cudf.Series]
```
rlratzel, line 32: Looks like this forces `torch` to be required instead of optional. We're seeing this in CI:
```
/opt/conda/envs/test/lib/python3.9/site-packages/cugraph/gnn/data_loading/dist_sampler.py:32: in <module>
    TensorType = Union[torch.Tensor, cupy.ndarray, cudf.Series]
/opt/conda/envs/test/lib/python3.9/site-packages/cugraph/utilities/utils.py:410: in __getattr__
    raise RuntimeError(f"This feature requires the {self.name} " "package/module")
E   RuntimeError: This feature requires the torch package/module
```
We should either update `import_optional` to allow it to support importing objects from modules, or maybe just treat `torch.Tensor` like a `ForwardRef` type by adding quotes?
```suggestion
TensorType = Union["torch.Tensor", cupy.ndarray, cudf.Series]
```

rlratzel (changes requested):

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -0,0 +1,577 @@
...
+        Parameters
+        ----------
+        nodes: TensorType
+            Input seeds (node ids).
+        batch_size: int
+            The size of each batch.
+        random_state: int
+            The random seed to use for sampling.
+        """
+        nodes = torch.as_tensor(nodes, device="cuda")
```
rlratzel, line 396: This is also causing problems in test environments that don't have `torch`:
```
/opt/conda/envs/test/lib/python3.9/site-packages/cugraph/gnn/data_loading/dist_sampler.py:396: in sample_from_nodes
    nodes = torch.as_tensor(nodes, device="cuda")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <cugraph.utilities.utils.MissingModule object at 0x7fed4e169d90>
attr = 'as_tensor'

    def __getattr__(self, attr):
>       raise RuntimeError(f"This feature requires the {self.name} " "package/module")
E       RuntimeError: This feature requires the torch package/module

/opt/conda/envs/test/lib/python3.9/site-packages/cugraph/utilities/utils.py:410: RuntimeError
```
I don't have a good suggestion other than to skip all tests that use this method.  Looks like `torch.split()`, `torch.distributed.all_reduce()` and maybe others are also used and would cause problems.

rlratzel (changes requested):
Looks like CI got further, but more failures that might mean we have to skip tests?

alexbarghi-nv:
> Looks like CI got further, but more failures that might mean we have to skip tests?

Yeah, we're skipping for now but I think long-term we may migrate some or all of this code to WholeGraph.  But that's a discussion for another time.

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
jnke2016: @VibhuJawa are you referring to the Jaccard/Two_hop_neighbor hang which transiently occurred if one or more processes are taking longer to return?

VibhuJawa: Yes, I think that is an issue which should be true for other algos too unless i am mistaken.

rlratzel (approved):

AyodeAwe (approved):

python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
jnke2016: Right, the workaround I had was using `client.map()` with iterables (like you did a while ago experimentaly for neighborhood sampling).

alexbarghi-nv: So what's the conclusion here then?  Do I need to add a delay somewhere?  Or is this way of calling sampling acceptable?

jnke2016: But this workflow is quiet different than our MG algos as there is no dask. If the `torch` `all_gather` call is similar to `MPI/NCCL` then I don't think there is an issue. In fact, I had a C test adding a long delay to one or more ranks from the communicator prior to a blocking collective call and the other ranks simply waited and resumed afterward once the delayed rank were ready.

alexbarghi-nv: `all_gather_into_tensor` here is a NCCL call

jnke2016: Ok then I think this is fine.

VibhuJawa: >I had a C test adding a long delay to one or more ranks from the communicator prior to a blocking collective call and the other ranks simply waited and resumed afterward once the delayed rank were ready.

Perfect, in that case we are good to go. Thanks for confirming @jnke2016 .

jnke2016 (approved):

PR #4283, opened 2024-03-26
[FIX] Remove Deprecated Sampling Options

alexbarghi-nv:
This PR removes the various sampling options that have been deprecated for several releases.

### DIFF
--- a/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
+++ b/python/cugraph/cugraph/dask/sampling/uniform_neighbor_sample.py
@@ -60,32 +60,19 @@ start_col_name = "_START_"
 batch_col_name = "_BATCH_"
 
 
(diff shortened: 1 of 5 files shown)


alexbarghi-nv:
Unfortunately, after a lengthy investigation, I've determined this PR is blocked until we support heterogeneous sampling as a separate call, since the bulk sampler and GNN frameworks are currently depending on some of these deprecated options to handle heterogeneous graphs.

As it stands, if we merge this PR now, we would be forced to remove heterogeneous graph support and then add it back once distributed sampling and the new heterogeneous sampling function are ready.  Since we don't want to regress, we are stuck with the current set of deprecated sampling options for the time being.  The distributed sampling path will have to go through a separate I/O system and initially won't support heterogeneous graphs.

alexbarghi-nv:
This PR is abandoned since we are going for a complete refactor (#4358)

PR #4309, opened 2024-04-02
Refactor CMakeLists.txt to better express usage requirements

robertmaynard:
Previously the cugraph CMakeLists.txt genernated non-relocatable targets due to embedding absolute paths to RAFT and the CUDAToolkit. While refactoring those issues out I also made a general cleanup pass over the rest of the CMake code.


cpp/tests/CMakeLists.txt
```diff
@@ -38,28 +47,97 @@ add_library(cugraphtestutil STATIC
             ../../thirdparty/mmio/mmio.c)
 
 target_compile_options(cugraphtestutil
-            PUBLIC "$<$<COMPILE_LANGUAGE:CXX>:${CUGRAPH_CXX_FLAGS}>"
-                   "$<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:CUDA>:${CUGRAPH_CUDA_FLAGS}>>"
+    PUBLIC "$<$<COMPILE_LANGUAGE:CXX>:${CUGRAPH_CXX_FLAGS}>"
+           "$<$<COMPILE_LANGUAGE:CUDA>:${CUGRAPH_CUDA_FLAGS}>"
```
vyasr, line 51: Also confused as to how this was working for installs before... I guess nobody was building against the test util?

cpp/tests/CMakeLists.txt
```diff
@@ -38,28 +47,97 @@ add_library(cugraphtestutil STATIC
...
 )
 
 set_property(TARGET cugraphtestutil PROPERTY POSITION_INDEPENDENT_CODE ON)
 
 target_include_directories(cugraphtestutil
     PUBLIC
-        "${CMAKE_CURRENT_SOURCE_DIR}/../../thirdparty/mmio"
         "${CMAKE_CURRENT_SOURCE_DIR}"
+        "${CMAKE_CURRENT_SOURCE_DIR}/../../thirdparty/mmio"
```
vyasr, line 59: Should we move this file into the tests dir since it's only used for tests?

cpp/tests/CMakeLists.txt
```diff
@@ -38,28 +47,97 @@ add_library(cugraphtestutil STATIC
...
+target_link_libraries(cugraph_c_testutil
+    PUBLIC
+        CUDA::toolkit
+    PRIVATE
+        cugraph::cugraph_c
+        cugraph::cugraph
+)
+
+if(BUILD_CUGRAPH_MG_TESTS)
+    include(../cmake/thirdparty/get_nccl.cmake)
```
vyasr, line 100: Another module we could move into the tests directory if NCCL is no longer linked to by cugraph itself.

cpp/tests/CMakeLists.txt
```diff
@@ -38,28 +47,97 @@ add_library(cugraphtestutil STATIC
...
+)
+
+set_property(TARGET cugraph_c_testutil PROPERTY POSITION_INDEPENDENT_CODE ON)
+
+target_link_libraries(cugraph_c_testutil
+    PUBLIC
+        CUDA::toolkit
+    PRIVATE
+        cugraph::cugraph_c
+        cugraph::cugraph
```
vyasr, line 96: Should this be specified explicitly or should this testutil just depend on cugraph_c and pull cugraph transitively? Does it have a direct dependency on cugraph?

cpp/CMakeLists.txt
```diff
@@ -339,49 +340,26 @@ target_include_directories(cugraph
...
-            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:raft::raft>
-            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:${COMPILED_RAFT_LIB}>
-            cuco::cuco
-            NCCL::NCCL
+target_link_libraries(cugraph
+    PUBLIC
+        rmm::rmm
+        raft::raft
+        $<BUILD_LOCAL_INTERFACE:CUDA::toolkit>
+        $<BUILD_LOCAL_INTERFACE:$<TARGET_NAME_IF_EXISTS:cugraph-ops::cugraph-ops++>>
```
vyasr, line 358: Based on the test failures we either need this one to not be `BUILD_LOCAL_INTERFACE` or we need to update the Python CMake (probably for both pylibcugraph and cugraph) to link to cugraph-ops themselves. What's the rationale for marking cugraph-ops as `BUILD_LOCAL_INTERFACE` here? Is there a reason to expect that we shouldn't be exporting the cugraph-ops dependency?

cpp/CMakeLists.txt
```diff
@@ -339,49 +340,26 @@ target_include_directories(cugraph
...
-            $<$<NOT:$<BOOL:${CUDA_STATIC_RUNTIME}>>:${COMPILED_RAFT_LIB}>
-        PRIVATE
-            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:raft::raft>
-            $<$<BOOL:${CUDA_STATIC_RUNTIME}>:${COMPILED_RAFT_LIB}>
-            cuco::cuco
-            NCCL::NCCL
+target_link_libraries(cugraph
+    PUBLIC
+        rmm::rmm
+        raft::raft
```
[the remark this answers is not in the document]
robertmaynard: I don't know but cugraph public headers include raft headers, so raft is an unconditional dependency to me.

cpp/tests/CMakeLists.txt
robertmaynard: That is my assumption as well

cpp/tests/CMakeLists.txt
robertmaynard: I think the goal was to have a singular thirdparty folder for the entire project, instead of having two.

cpp/tests/CMakeLists.txt
robertmaynard: It has a direct dependency on cugraph::cugraph

robertmaynard: cugraph_c doesn't bring in cugraph as a transitive dep as well

cpp/tests/CMakeLists.txt
robertmaynard: I am ambivalant about `get_nccl.cmake` being in the normal `cmake` folder. We do the same for other test conditional dependencies like GTest

cpp/CMakeLists.txt
robertmaynard: Oversight on my behalf, will make it public

ChuckHastings (commented):
This looks good to me.  I provided a little feedback.  Happy to approve unless there are reactions to my comments.

bdice (approved):

cpp/tests/CMakeLists.txt
robertmaynard: > Should we have cugraph_c bring in cugraph as a dependency? This was an area of indecision for me when we first added it.

When talking about CMake dependencies ( via `PUBLIC` vs `PRIVATE` ) keyword we need to think about only what is required for correct compilation and linking, and not what is required at runtime. `cugraph_c` is designed as a pure `C` language firewall. No cugraph C++ types are allowed to pass this firewall, and therefore no build/link time consumers of `cugraph_c` should need to know about `cugraph`.

Therefore to ensure this firewall is correct, we want a `PRIVATE` linkage from `cugraph_c` to `cugraph`

ChuckHastings (approved):

PR #4315, opened 2024-04-04
MNMG Approximation Algorithm for the Weighted Matching Problem

naimnv:
MNMG [Approximation Algorithm for the Weighted Matching Problem](https://web.archive.org/web/20081031230449id_/http://www.ii.uib.no/~fredrikm/fredrik/papers/CP75.pdf)


cpp/include/cugraph/algorithms.hpp
```diff
@@ -2366,6 +2366,33 @@ rmm::device_uvector<vertex_t> vertex_coloring(
...
+ * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
+ * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
+ * @tparam multi_gpu Flag indicating whether template instantiation should target single-GPU (false)
+ * @param[in] handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator,
+ * and handles to various CUDA libraries) to run graph algorithms.
+ * @param[in] graph_view Graph view object.
+ * @param[in] edge_weight_view View object holding edge weights for @p graph_view.
+ * @param[out] partners Device span to matched vetex ids, memory needs to be pre-allocated by
+ * caller.
+ * @return Sum of the weights of the matched edges.
```
seunghwak, line 2389: Should we better return `std::tuple<rmm::device_uvector<vertex_t>, weight_t>`?

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2366,6 +2366,33 @@ rmm::device_uvector<vertex_t> vertex_coloring(
...
+ */
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+weight_t matching(raft::handle_t const& handle,
```
seunghwak, line 2392: Isn't this function name to generic? Better be more specific? (e.g. `weighted_maximum_matching`)

cpp/src/community/matching_impl.cuh
```diff
@@ -0,0 +1,382 @@
...
+  auto constexpr invalid_partner = invalid_vertex_id<vertex_t>::value;
+  rmm::device_uvector<weight_t> offers_from_partners(
+    current_graph_view.local_vertex_partition_range_size(), handle.get_stream());
+  thrust::fill(handle.get_thrust_policy(), partners.begin(), partners.end(), invalid_partner);
+  thrust::fill(handle.get_thrust_policy(),
+               offers_from_partners.begin(),
+               offers_from_partners.end(),
+               weight_t{0.0});
+
+  edge_src_property_t<graph_view_t, vertex_t> src_key_cache(handle);
```
seunghwak, line 74: Better minimize the scope of each variable, you can define this after `sequence_fill`.

cpp/src/community/matching_impl.cuh
```diff
@@ -0,0 +1,382 @@
...
+    current_graph_view.local_vertex_partition_range_size(), handle.get_stream());
+  detail::sequence_fill(handle.get_stream(),
+                        local_vertices.begin(),
+                        local_vertices.size(),
+                        current_graph_view.local_vertex_partition_range_first());
+
+  vertex_t loop_counter = 0;
+  while (true) {
+    if constexpr (graph_view_t::is_multi_gpu) {
+      src_key_cache = edge_src_property_t<graph_view_t, vertex_t>(handle, current_graph_view);
```
seunghwak, line 85: Any reason you're reallocating this in every loop? Can't you allocate this outside the loop and just update inside the loop?

cpp/src/community/matching_impl.cuh
```diff
@@ -0,0 +1,382 @@
...
+                        d_tx_value_counts.size(),
+                        handle.get_stream());
+      handle.sync_stream();
+
+      std::forward_as_tuple(std::tie(candidates, offers_from_candidates, targets), std::ignore) =
+        shuffle_values(handle.get_comms(),
+                       thrust::make_zip_iterator(
+                         candidates.begin(), offers_from_candidates.begin(), targets.begin()),
+                       h_tx_value_counts,
+                       handle.get_stream());
```
seunghwak, line 165: `shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning`?

cpp/src/community/matching_impl.cuh
```diff
@@ -0,0 +1,382 @@
...
+    }
+
+    using flag_t                        = uint8_t;
```
seunghwak, line 168: Why `uint8_t` instead of `bool`?

seunghwak (commented):
We need C++ tests as well.

cpp/include/cugraph/algorithms.hpp
naimnv: Changed.

cpp/src/community/matching_impl.cuh
naimnv: Changed.

cpp/include/cugraph/algorithms.hpp
naimnv: Changed.

cpp/src/community/matching_impl.cuh
naimnv: Here we are shuffling two values associated with one vertex, shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning only works for shuffling one value for each vertex.

cpp/src/community/matching_impl.cuh
naimnv: If I use `cugraph::edge_src_property_t<graph_view_t, bool> src_match_flags(handle);`
 and `is_vertex_matched ` is a vertex property of type `uint8_t `
 The following statement wouldn't work.
```
    graph_view_t::is_multi_gpu ? src_match_flags.view()
          : detail::edge_major_property_view_t<vertex_t, flag_t const*>(is_vertex_matched.begin()),
```

cpp/src/community/matching_impl.cuh
naimnv: Changed to allocate it once.

naimnv:
@ChuckHastings  @seunghwak It's ready for a re-review.

ChuckHastings (approved):

cpp/src/community/approx_weighted_matching_impl.cuh
```diff
@@ -0,0 +1,380 @@
...
+
+    //
+    // For each candidate vertex, find the best possible target
+    //
+
+    rmm::device_uvector<vertex_t> candidates(0, handle.get_stream());
+    rmm::device_uvector<weight_t> offers_from_candidates(0, handle.get_stream());
+    rmm::device_uvector<vertex_t> targets(0, handle.get_stream());
+
+    std::forward_as_tuple(candidates, std::tie(offers_from_candidates, targets)) =
```
seunghwak, line 116: May better add a FIXME statement saying that this can be implemented more efficiently if per_v_transform_reduce_incoming|outgoing_e is updated to support reduction on thrust::tuple.

cpp/src/community/matching_impl.cuh
seunghwak: You mean shuffling a tuple associated with one vertex, right? Can't you update the existing function to support a tuple?

cpp/tests/community/mg_weighted_matching_test.cpp
```diff
@@ -0,0 +1,225 @@
...
+#include <raft/random/rng_state.hpp>
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <iostream>
+#include <random>
+
+struct WeightedMatching_UseCase {
+  bool check_correctness{true};
```
seunghwak, line 40: Now we should test with bool edge_masking as well.

cpp/tests/community/weighted_matching_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+#include <raft/random/rng_state.hpp>
+
+#include <gtest/gtest.h>
+
+#include <chrono>
+#include <iostream>
+#include <random>
+
+struct WeightedMatching_UseCase {
+  bool check_correctness{true};
```
seunghwak, line 38: Now we should test with bool edge_masking as well.

cpp/tests/community/weighted_matching_test.cpp
```diff
@@ -0,0 +1,175 @@
...
+      cugraph::approximate_weighted_matching<vertex_t, edge_t, weight_t, multi_gpu>(
+        handle, sg_graph_view, (*sg_edge_weights).view());
+
+    if (weighted_matching_usecase.check_correctness) {
+      auto h_partners                = cugraph::test::to_host(handle, d_partners);
+      auto constexpr invalid_partner = cugraph::invalid_vertex_id<vertex_t>::value;
+
+      std::for_each(h_partners.begin(), h_partners.end(), [&invalid_partner, h_partners](auto& v) {
+        if (v != invalid_partner) ASSERT_TRUE(h_partners[h_partners[v]] == v);
+      });
```
seunghwak, line 103: Can't we at least check that this satisfies requirements to be a proper matching?

naimnv: Do you mean [perfect matching](https://en.wikipedia.org/wiki/Perfect_matching)?
It's not guarantee to a perfect matching.

cpp/src/community/matching_impl.cuh
naimnv:
```
candidates
offers_from_candidates 
 targets
```
Each target in `targets` has  a candidate and offer from the candidate. 

@seunghwak  Do you mean updating `shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning` to handle tuple of arbitrary length, and containing of any types?

cpp/tests/community/weighted_matching_test.cpp
seunghwak: "In graph theory, a matching in a graph is a set of edges that do not have a set of common vertices." Shouldn't this be enforced even for approximate matching?

cpp/src/community/matching_impl.cuh
seunghwak: Yes, for the code (if this is not too complicated), no for explicit instantiation. For explicit instantiation, you can just add an explicit instantiation for your use case.

cpp/tests/community/weighted_matching_test.cpp
naimnv: This enforced is already enforced by the check  `h_partners[h_partners[v]] == v`

cpp/src/community/approx_weighted_matching_impl.cuh
naimnv: Added.

cpp/src/community/matching_impl.cuh
naimnv: Changed to groupby_gpu_id_and_shuffle_values

cpp/tests/community/mg_weighted_matching_test.cpp
naimnv: Tested.

cpp/tests/community/weighted_matching_test.cpp
naimnv: Tested

naimnv:
@seunghwak it's read for a final review. I tested the code on dgx machin with 8 GPUs.

seunghwak (approved):
LGTM

PR #4320, opened 2024-04-08
Bug fix in cugraph::detail::extract_transform_v_froniter_e

seunghwak:
Fix a bug potentially accessing thrust::optional object's value when the optional object does not contain a value (thrust::nullopt).

This may manifest only when cugraph is built with a debug option (no deadcode elimination optimization).


vyasr (approved):
I can confirm that this fixes the bug in the debug build that I found. Looks like style checks need updating before CI will pass, though.

jnke2016 (approved):

naimnv (approved):

naimnv (approved):

ChuckHastings (approved):

PR #4321, opened 2024-04-09
Updates the nx-cugraph README.md with the latest upstream URLs

rlratzel:
This PR updates the nx-cugraph README.md with the latest upstream URLs by running the `update_readme.py` script on a current `objects.inv` file downloaded from NetworkX.  The latest `objects.inv` file has updated URLs, which results in links in the README being updated too.

This should resolve style check errors recently seen in [other PRs](https://github.com/rapidsai/cugraph/pull/4320) due to the upstream change in `objects.inv` in NetworkX.  Since CI runs pre-commit on all files, this caused the nx-cugraph checks to run on PRs that didn't change nx-cugraph files.


alexbarghi-nv (approved):
👍

PR #4325, opened 2024-04-09
[DNM][WIP] Debug expression-based Dask Dataframe support

rjzamora:
**[DO NOT MERGE]** I'm using this PR to debug/add support for `DASK_DATAFRAME__QUERY_PLANNING=True`.

**NOTES**:
- ...

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -3,10 +3,6 @@
 
 set -euo pipefail
 
-# TODO: Enable dask query planning (by default) once some bugs are fixed.
-# xref: https://github.com/rapidsai/cudf/issues/15027
-export DASK_DATAFRAME__QUERY_PLANNING=False
-
 # Support invoking test_python.sh outside the script directory
 cd "$(dirname "$(realpath "${BASH_SOURCE[0]}")")"/../
 
--- a/ci/test_wheel.sh
+++ b/ci/test_wheel.sh
@@ -3,10 +3,6 @@
 
 set -eoxu pipefail
 
-# TODO: Enable dask query planning (by default) once some bugs are fixed.
-# xref: https://github.com/rapidsai/cudf/issues/15027
-export DASK_DATAFRAME__QUERY_PLANNING=False
-
 package_name=$1
 package_dir=$2
 


python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -295,9 +295,13 @@ def __from_edgelist(
...
 
+        def _assign(x, y, names):
+            for name in names:
+                x[name] = y[name]
+            return x
+
         if value_col is not None:
-            for vc in value_col_names:
-                input_ddf[vc] = value_col[vc]
+            input_ddf = input_ddf.map_partitions(_assign, value_col, value_col_names)
```
rjzamora, line 304: TODO: This may be a dask-expr bug.

python/cugraph/cugraph/structure/symmetrize.py
```diff
@@ -298,7 +298,7 @@ def symmetrize(
             return (
                 output_df[source_col_name],
                 output_df[dest_col_name],
-                output_df[value_col.columns],
+                output_df[list(value_col.columns)],
```
rjzamora, line 301: TODO: This may be a dask-expr bug? Column projection using anything other than a list seems fragile.

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
```diff
@@ -344,7 +344,7 @@ def generate_rmat_dataset(
     del label_df
     gc.collect()
 
-    dask_label_df = dask_cudf.from_dask_dataframe(dask_label_df)
+    dask_label_df = dask_label_df.to_backend("cudf")
```
rjzamora, line 347: `from_dask_dataframe` is now deprecated.

python/cugraph/cugraph/dask/__init__.py
```diff
@@ -34,3 +36,6 @@
 from .link_prediction.sorensen import sorensen
 from .link_prediction.overlap import overlap
 from .community.leiden import leiden
+
+# Avoid "p2p" shuffling in dask for now
+config.set({"dataframe.shuffle.method": "tasks"})
```
rjzamora, line 41: "p2p" should work fine, but it will rarely provide a performance benefit. It seems best to minimize "optional" changes until the query-planning migration is finished.

python/cugraph/cugraph/dask/common/input_utils.py
```diff
@@ -15,8 +15,8 @@
 
 from collections.abc import Sequence
 from collections import OrderedDict
-from dask_cudf.core import DataFrame as dcDataFrame
-from dask_cudf.core import Series as daskSeries
+from dask_cudf import DataFrame as dcDataFrame
+from dask_cudf import Series as daskSeries
```
rjzamora, line 19: NOTE: **All** imports from `dask_cudf.core` should be avoided, because these imports are always using "legacy" dask-cudf. Importing from the top-level `dask_cudf` module are automatically routed to the proper API. There is no way to protect against `dask_cudf.core` imports **yet**, because some query-planning logic still needs to find/use specific legacy code.

python/cugraph/cugraph/structure/symmetrize.py
rjzamora: I'm not observing this bug locally anymore, but I'd still like to keep this precaution in place.

python/cugraph/cugraph/tests/structure/test_graph_mg.py
```diff
@@ -99,13 +99,13 @@ def test_nodes_functionality(dask_client, input_combo):
     expected_nodes = (
         dask_cudf.concat([ddf["src"], ddf["dst"]])
         .drop_duplicates()
-        .to_frame()
-        .sort_values(0)
+        .to_frame(name="0")
+        .sort_values("0")
```
rjzamora, line 103: Another "precaution" (using numerical column names still seems "fragile" in dask)

rlratzel (approved):
LGTM

python/cugraph/cugraph/structure/symmetrize.py
jakirkham: Should we leave a comment? Is it worth raising a tracking issue on cuGraph for follow up?

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
jakirkham: Does this mean that dask-expr has some dispatching/plugin support for different DataFrame implementations?

ci/test_python.sh
```diff
@@ -3,10 +3,6 @@
 
 set -euo pipefail
 
-# TODO: Enable dask query planning (by default) once some bugs are fixed.
-# xref: https://github.com/rapidsai/cudf/issues/15027
-export DASK_DATAFRAME__QUERY_PLANNING=False
-
```
jakirkham, line 9: AIUI this is one of the OPS relevant changes. Basically removing a workaround that is no longer needed

ci/test_wheel.sh
```diff
@@ -3,10 +3,6 @@
 
 set -eoxu pipefail
 
-# TODO: Enable dask query planning (by default) once some bugs are fixed.
-# xref: https://github.com/rapidsai/cudf/issues/15027
-export DASK_DATAFRAME__QUERY_PLANNING=False
-
```
jakirkham, line 9: This is the other one. So same change as before just in another place

jakirkham (commented):
Highlighting the OPS relevant changes. Namely dropping old Dask workarounds (environment variables that are no longer needed) as the underlying issue was resolved.

raydouglass (approved):

python/cugraph/cugraph/structure/symmetrize.py
rjzamora: I'll be honest: I don't actually think this "fix" is required, because removing it doesn't seem to cause test failures for me locally (was probably specific to an earlier combination of dask/dask-expr/dask-cudf). However, I left it for now because it will take a long time for "real CI" to tell me that it actually is a problem.

With that said, I'll be happy to give it a try now that I'm realizing it's only cudf/dask-cudf that is about to freeze (and not cugraph).

benchmarks/cugraph/standalone/bulk_sampling/cugraph_bulk_sampling.py
rjzamora: Correct - Dask documentation is here: https://docs.dask.org/en/latest/how-to/selecting-the-collection-backend.html, and Dask-cudf: https://docs.rapids.ai/api/dask-cudf/stable/#dataframe-creation-from-in-memory-formats

jakirkham (commented):
Thanks Rick! 🙏

Based on your comment above, do we want to drop this workaround?

python/cugraph/cugraph/structure/symmetrize.py
```diff
@@ -296,7 +296,7 @@ def symmetrize(
...
```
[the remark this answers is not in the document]
rjzamora: Yeah, thanks (sorry - got a bit distracted by something else)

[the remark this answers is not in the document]
jakirkham: No worries. Thanks Rick! 🙏

PR #4347, opened 2024-04-16
Update generation and flattening of dendrogram in Leiden

### LINKED ISSUE #4072: What is the expected repeatability of cugraph.leiden?
### What is your question?

I know that one cannot expect exact repeatability over multiple runs of `cugraph.leiden` (even setting `random_state` to a constant). But how similar can we rely on the results being? 

I've repeated 10 runs on a 15,000-vertex graph on 13 different GPUs, and got an average adjusted Rand index of 0.82 across the runs on a given GPU (min: 0.69, max: 0.95). That corresponds to a difference of ±1 cluster detected on average (and up to a 6-cluster difference in the worst case) for this graph. Is that as consistent as we can hope for?

### Code of Conduct

- [X] I have searched the [open issues](https://github.com/rapidsai/cugraph/issues) and have found no duplicates for this question

ChuckHastings:
The Leiden dendrogram was being populated with both Louvain cluster assignment and mapping between Leiden and Louvain clustering.  The flattening of the dendrogram was being accomplished by applying the Louvain clustering, then mapping between the Leiden clusters at each level.

Unfortunately, the Leiden to Louvain mapping allows a many-to-one relationship, so in certain cases the flattening was non-deterministic.  There wasn't enough information in the dendrogram to perform the mapping deterministically.

This PR modifies the dendrogram to be similar to Louvain, just keeping the cluster assignments at each level.  The mapping between Louvain and Leiden clusters is done when creating the dendrogram, where there is sufficient information to perform this translation deterministically.

Closes #4072


cpp/src/community/leiden_impl.cuh
```diff
@@ -466,34 +467,48 @@ std::pair<std::unique_ptr<Dendrogram<vertex_t>>, weight_t> leiden(
...
+        detail::sequence_fill(handle.get_stream(),
+                              louvain_of_refined_graph.data(),
+                              louvain_of_refined_graph.size(),
+                              current_graph_view.local_vertex_partition_range_first());
+
+        relabel<vertex_t, multi_gpu>(
+          handle,
+          std::make_tuple(static_cast<vertex_t const*>((*numbering_map).begin()),
+                          static_cast<vertex_t const*>(louvain_of_refined_graph.begin())),
+          current_graph_view.local_vertex_partition_range_size(),
```
naimnv, line 494: ` (*numbering_map).size() `<= `current_graph_view.local_vertex_partition_range_size()`
 I think we need to use ` (*numbering_map).size() ` here

seunghwak: When can `(*numbering_map).size() < current_graph_view.local_vertex_partition_range_size()`???

Don't miss https://github.com/rapidsai/cugraph/pull/4347/files#diff-609b314aeaec6ca07d296e41bdf9f67407b5ffe539ce568506290925e9b169abR464

I think `(*numbering_map).size() == current_graph_view.local_vertex_partition_range_size()`

seunghwak (approved):
LGTM

cpp/src/community/leiden_impl.cuh
naimnv: You are are right, they are equal.

naimnv (approved):

rlratzel (approved):

PR #4355, opened 2024-04-17
[FEA] Support Seed Retention for Sampling with Renumbering

alexbarghi-nv:
Exposes the ability to retain seeds even if they have no outgoing edges (and therefore are not sampled).  Required to fix the current bug in cuGraph-PyG involving batch size and dropping seeds.

Currently, this functionality can't be exposed through the MG Python API (#4358) but exposing it through the pylibcugraph API is sufficient to resolve this issue.  This PR does expose it through the SG Python API.


naimnv (commented):
C++ and C part LGTM
Should we add test for the added option?

seunghwak (approved):
LGTM (C/C++ part)

ChuckHastings:
> LGTM Shouldn't we add C++ test for the added option?

This is not an added C++ option, it already existed.  It was never exposed/used through the C API.  I believe (didn't exhaustively look) that it is already tested in `sampling_post_processing_test.cu`.

naimnv (approved):

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
```diff
@@ -142,6 +144,15 @@ def uniform_neighbor_sample(
...
         as an additional dataframe.
 
+    retain_seeds: bool, optional (default=False)
+        If True, will retain the original seeds (original source vertices)
+        in the output even if they do not have outgoing neighbors.
+
+    label_offsets: integer sequence, optional (default=None)
+        Offsets of each label within the start vertex list.
+        Only used if retain_seeds is True.  Required if retain_seeds
+        is True.
```
rlratzel, line 154: Just a suggestion: maybe change this so that if this is specified, the `retain_seeds=True` behavior is assumed and then drop the `retain_seeds` arg completely?  If you decide to do that, then be sure to also include the description of the `retain_seeds` behavior in this description.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
```diff
@@ -342,13 +353,15 @@ def uniform_neighbor_sample(
         else None,
         h_fan_out=fanout_vals,
         with_replacement=with_replacement,
-        do_expensive_check=False,
+        do_expensive_check=True,
```
rlratzel, line 356: was this intentional?

rlratzel (approved):
I just reviewed the python changes.  I had one question and one optional suggestion, but neither should hold up approval.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
alexbarghi-nv: not intentional, whoops.  We need to fix that.

python/cugraph/cugraph/sampling/uniform_neighbor_sample.py
alexbarghi-nv: That's a reasonable suggestion.  I'll add that to the new PR.

PR #4362, opened 2024-04-19
Erdos-Renyi generator had bad logic in thrust calls

### LINKED ISSUE #4359: [BUG]: erdos_renyi_generator not working properly
### Version

v23.02.00

Conda

### Describe the bug.

The source is cugraph/cpp/src/generators/erdos_renyi_generator.cu, and the test is cugraph/cpp/tests/generators/erdos_renyi_test.cpp. When random graph generation probability is changed from 0.1 to 0.5, there is some test error, e.g., with er_test<int32_t>(size_t{10}, float{0.5}). Moreover, the strange thing is that almost all generated source-to-destination vertice pairs are 0-0s, meaning there is almost no edge between vertex 0 and other vertices. I wonder if the function is mature enough for usage or is it to be further developed? Thank you for your time.

_No response_

### Relevant log output

_No response_

### Environment details

_No response_

### Other/Misc.

_No response_

### Code of Conduct

ChuckHastings:
The `random_iterator`, a thrust transform iterator that was used for counting and filtering returns the probability, but they `copy_if` call was assuming it was returning the index.

Modified the logic and then consolidated the `copy_if` and `transform` into a single call with an output iterator.

Closes #4359

### DIFF
--- a/cpp/src/generators/erdos_renyi_generator.cu
+++ b/cpp/src/generators/erdos_renyi_generator.cu
@@ -23,10 +23,9 @@
 #include <thrust/copy.h>
 #include <thrust/count.h>
 #include <thrust/iterator/counting_iterator.h>
-#include <thrust/iterator/transform_iterator.h>
+#include <thrust/iterator/transform_output_iterator.h>
 #include <thrust/iterator/zip_iterator.h>
 #include <thrust/random.h>
-#include <thrust/transform.h>
 #include <thrust/tuple.h>
 
 namespace cugraph {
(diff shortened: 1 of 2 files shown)


cpp/src/generators/erdos_renyi_generator.cu
```diff
@@ -42,45 +41,38 @@ generate_erdos_renyi_graph_edgelist_gnp(raft::handle_t const& handle,
...
+  size_t max_num_edges = static_cast<size_t>(num_vertices) * num_vertices;
 
-  size_t count = thrust::count_if(handle.get_thrust_policy(),
-                                  random_iterator,
-                                  random_iterator + num_vertices * num_vertices,
-                                  [p] __device__(float prob) { return prob < p; });
-
-  rmm::device_uvector<size_t> indices_v(count, handle.get_stream());
+  auto generate_random_value = cuda::proclaim_return_type<float>([seed] __device__(size_t index) {
+    thrust::default_random_engine rng(seed);
```
seunghwak, line 47: This might be out side the scope of this bug fix PR, but should we better use `raft::random::RngState`? Using two different random number generation mechanism does not sound ideal.

seunghwak (approved):

naimnv (approved):

jnke2016 (approved):

cpp/src/generators/erdos_renyi_generator.cu
ChuckHastings: Created an issue to address this.

cpp/src/generators/erdos_renyi_generator.cu
```diff
@@ -42,45 +41,38 @@ generate_erdos_renyi_graph_edgelist_gnp(raft::handle_t const& handle,
...
                     cuda::proclaim_return_type<thrust::tuple<vertex_t, vertex_t>>(
                       [num_vertices] __device__(size_t index) {
-                        size_t src = index / num_vertices;
-                        size_t dst = index % num_vertices;
-
-                        return thrust::make_tuple(static_cast<vertex_t>(src),
-                                                  static_cast<vertex_t>(dst));
-                      }));
-
-  handle.sync_stream();
```
lvjing421, line 83: Is it OK to leave out handle.sync_stream() here, as there is one after the kernel call in cpp/tests/generators/erdos_renyi_test.cpp?

ChuckHastings: Yes.  The `sync_stream` call here was never necessary.  Any asynchronous work that is active on this stream will continue to be active when we leave this function call.

PR #4271, opened 2024-03-22
Enable FutureWarnings and Deprecations as errors in cugraph

mroeschke:
Follow up to https://github.com/rapidsai/cugraph/pull/4223 and https://github.com/rapidsai/build-planning/issues/26, cugraph tests will fail if a `FutureWarning` or `DeprecationWarning` is raised

PR #4379, opened 2024-04-29
Address dask_cudf.read_csv chunksize deprecation

mroeschke:
xref https://github.com/rapidsai/cugraph/pull/4271

`chunksize` was deprecated in favor of `blocksize`

Also removed an unsupported `chunksize` in a `cudf.read_csv` call

### DIFF
--- a/benchmarks/cugraph/standalone/cugraph_dask_funcs.py
+++ b/benchmarks/cugraph/standalone/cugraph_dask_funcs.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2023, NVIDIA CORPORATION.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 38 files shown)


rlratzel (commented):
Thank you very much, @mroeschke !

So is `blocksize` simply a rename of the old `chunksize` arg, or are there bigger differences?  I'm just asking in case we also need to update [get_chunksize()](https://github.com/rapidsai/cugraph/blob/581bd9cfa8cf370016e57d7cc2647c93c430e4bd/python/cugraph/cugraph/dask/common/read_utils.py#L23) to return something that's computed differently.

mroeschke:
> So is blocksize simply a rename of the old chunksize arg, or are there bigger differences?

I believe the change is just a pure renamed based on https://github.com/rapidsai/cudf/pull/12394, but tagging @rjzamora just to confirm

rjzamora:
>I believe the change is just a pure renamed based on https://github.com/rapidsai/cudf/pull/12394, but tagging @rjzamora just to confirm

Yeah, I believe someone arbitrarily chose the name `chunksize` in the early days of dask-cudf, even though the upstream `dask.dataframe` function was already using the name `blocksize` for the exact same purpose. When we started closing the gap between the `dask_cudf` and `dask.dataframe` APIs a couple years ago, we deprecated `chunksize` in favor of `blocksize`.

rlratzel (approved):
Thank you!

PR #4382, opened 2024-04-30
Move edge triangle count to the stable API

### LINKED ISSUE #4370: Edge Triangle Counts in chunks
Neighbor intersection can be a memory intensive operation especially when intersecting high degree vertices. 

1. Perform neighbor intersection in chunks

### LINKED ISSUE #4371: Move edge triangle count to stable api
Edge triangle count is not part of the stable API hence:
- Move edge triangle count to stable API
- Add tests
- Add MG support

jnke2016:
This PR

1. Performs edge triangle count in chunk
2. Enables k - 1 core optimization
3. Add C++ tests for edge triangle count
4. Move edge triangle count to the stable API
5. Implement MG edge triangle count and add tests
6. Update 'mg_graph_to_sg_graph' to support 'edge_ids' along with tests

closes #4370
closes #4371


cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -81,74 +76,103 @@ struct update_edges_p_r_q_r_num_triangles {
...
 {
+  using weight_t = float;
+  rmm::device_uvector<vertex_t> edgelist_srcs(0, handle.get_stream());
+  rmm::device_uvector<vertex_t> edgelist_dsts(0, handle.get_stream());
+  std::tie(edgelist_srcs, edgelist_dsts, std::ignore, std::ignore) = decompress_to_edgelist(
+    handle,
+    graph_view,
+    std::optional<edge_property_view_t<edge_t, weight_t const*>>{std::nullopt},
+    std::optional<edge_property_view_t<edge_t, edge_t const*>>{std::nullopt},
+    std::optional<raft::device_span<vertex_t const>>(std::nullopt));
```
seunghwak, line 89: Just a tip,

If you explicitly provide the template parameter types, you can just pass std::nullopt.
```
std::tie(edgelist_srcs, edgelist_dsts, std::ignore, std::ignore) = 
decompress_to_edgelist<vertex_t, edge_t, weight_t>(
handle, graph_view, std::nullopt, std::nullopt, std::nullopt);
```
will also compile.

seunghwak: Sometimes, explicitly providing template parameter types leads to more concise code.

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -81,74 +76,103 @@ struct update_edges_p_r_q_r_num_triangles {
...
-  auto [intersection_offsets, intersection_indices] =
-    detail::nbr_intersection(handle,
-                             graph_view,
-                             cugraph::edge_dummy_property_t{}.view(),
-                             edge_first,
-                             edge_first + edgelist_srcs.size(),
-                             std::array<bool, 2>{true, true},
-                             false /*FIXME: pass 'do_expensive_check' as argument*/);
+  size_t approx_edges_to_intersect_per_iteration =
+    static_cast<size_t>(handle.get_device_properties().multiProcessorCount) * (1 << 17);
```
seunghwak, line 96: This isn't approximate, right? (I mean except for the last chunk). This `approx_` naming convention is used when we are dealing with the CSR data structure and # edges should need to be aligned with the CSR offset array boundaries.

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -81,74 +76,103 @@ struct update_edges_p_r_q_r_num_triangles {
...
+  std::tie(edgelist_srcs, edgelist_dsts, std::ignore, std::ignore) = decompress_to_edgelist(
+    handle,
+    graph_view,
+    std::optional<edge_property_view_t<edge_t, weight_t const*>>{std::nullopt},
+    std::optional<edge_property_view_t<edge_t, edge_t const*>>{std::nullopt},
+    std::optional<raft::device_span<vertex_t const>>(std::nullopt));
+
   auto edge_first = thrust::make_zip_iterator(edgelist_srcs.begin(), edgelist_dsts.begin());
 
   thrust::sort(handle.get_thrust_policy(), edge_first, edge_first + edgelist_srcs.size());
```
seunghwak, line 93: What's the purpose of sort here?

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -81,74 +76,103 @@ struct update_edges_p_r_q_r_num_triangles {
...
-                             edge_first,
-                             edge_first + edgelist_srcs.size(),
-                             std::array<bool, 2>{true, true},
-                             false /*FIXME: pass 'do_expensive_check' as argument*/);
+  size_t approx_edges_to_intersect_per_iteration =
+    static_cast<size_t>(handle.get_device_properties().multiProcessorCount) * (1 << 17);
 
+  auto num_chunks        = ((edgelist_srcs.size() % approx_edges_to_intersect_per_iteration) == 0)
+                             ? (edgelist_srcs.size() / approx_edges_to_intersect_per_iteration)
+                             : (edgelist_srcs.size() / approx_edges_to_intersect_per_iteration) + 1;
```
seunghwak, line 100: `= (edgelist_srcs.size() + edges_to_intersect_per_iteration - 1) / edges_to_intersection_per_iteration.` will be little more compact.

Or even better, I think we should use a raft utility function (https://github.com/rapidsai/raft/blob/branch-24.06/cpp/include/raft/util/integer_utils.hpp#L118)
`div_rounding_up_safe(edgelist_srcs.size(), edge_to_intersect_per_iteration)`

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -81,74 +76,103 @@ struct update_edges_p_r_q_r_num_triangles {
...
+  size_t prev_chunk_size = 0;
+  auto num_edges         = edgelist_srcs.size();
```
seunghwak, line 102: num_remaining_edges to be clearer?

seunghwak: And prev_chunk_size = edgelist_srcs.size() - num_remaining_edges, so redundant.

seunghwak (commented):
Reviewed the edge triangle count implementation (Haven't reviewed the K-Truss implementation updates yet).

cpp/src/community/k_truss_impl.cuh
```diff
@@ -726,92 +717,114 @@ k_truss(raft::handle_t const& handle,
...
-                         transposed_edge_first + edgelist_srcs.size()});
-
+      size_t approx_edges_to_intersect_per_iteration =
+        static_cast<size_t>(handle.get_device_properties().multiProcessorCount) * (1 << 17);
+
+      size_t prev_chunk_size         = 0;
+      size_t chunk_num_invalid_edges = num_invalid_edges;
+      auto num_chunks = ((num_invalid_edges % approx_edges_to_intersect_per_iteration) == 0)
+                          ? (num_invalid_edges / approx_edges_to_intersect_per_iteration)
+                          : (num_invalid_edges / approx_edges_to_intersect_per_iteration) + 1;
```
seunghwak, line 727: I think we can use `div_roundiung_up_safe` here.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -2007,6 +2007,10 @@ void triangle_count(raft::handle_t const& handle,
                     raft::device_span<edge_t> counts,
                     bool do_expensive_check = false);
 
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+rmm::device_uvector<edge_t> edge_triangle_count(
```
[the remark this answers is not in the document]
jnke2016: > How do you match the returned output vector elements to the edges in the graph?

I see what you mean.The reason why my tests are passing is because the edges are sorted by `src` so the corresponding count will be return however this is not a correct assumption. I will return an `edge_property_t` instead and I believe you suggested before.

cpp/tests/community/edge_triangle_count_test.cpp
```diff
@@ -0,0 +1,250 @@
...
+#include <algorithm>
+#include <iterator>
+#include <limits>
+#include <numeric>
+#include <random>
+#include <set>
+#include <vector>
+
+struct EdgeTriangleCount_Usecase {
+  bool test_weighted_{false};
```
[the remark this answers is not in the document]
jnke2016: > Need to include edge_masking.

I included edge mask tests

> is this code MNMG ready?

Yes it is

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: I see. Thanks for the tip

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: > I mean except for the last chunk

Ya or if it is bigger than the number of edges. But I agree it doesn't align with the intended context (CSR data structure). I renamed it to `edges_to_intersect_per_iteration`

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: Good catch. We no longer need this as we do not pass an edgelist to `edge_triangle_count`. In fact when we decompress the edges from the graph, they are already sorted.

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: I used the raft utility function

cpp/src/community/k_truss_impl.cuh
```diff
@@ -561,17 +554,17 @@ k_truss(raft::handle_t const& handle,
...
         std::nullopt,
         std::move(srcs),
         std::move(dsts),
-        std::nullopt,
+        std::move(wgts),
         std::nullopt,
         std::nullopt,
         cugraph::graph_properties_t{true, graph_view.is_multigraph()},
-        true);
+        false);  // FIXME: Renumbering should not be hardcoded.
```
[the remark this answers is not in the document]
jnke2016: I wasn't sure if renumbering should be hardcoded but I removed the FIXME.

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: > num_remaining_edges to be clearer?

Done

> And prev_chunk_size = edgelist_srcs.size() - num_remaining_edges, so redundant

Can you be more clear here?

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -78,77 +78,298 @@ struct update_edges_p_r_q_r_num_triangles {
...
+
+      std::optional<rmm::device_uvector<edge_t>> opt_increase_count =
+        std::make_optional(rmm::device_uvector<edge_t>(increase_count.size(), handle.get_stream()));
+
+      raft::copy<edge_t>((*opt_increase_count).begin(),
+                         increase_count.begin(),
+                         increase_count.size(),
+                         handle.get_stream());
+
+      // There are still multiple copies here but is it worth sorting and reducing again?
```
seunghwak, line 253: You mean after shuffle, right?

So, this is a trade-off. If you sort-and-reduce, no need to use atomics. You can experiment which will be faster.

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -78,77 +78,298 @@ struct update_edges_p_r_q_r_num_triangles {
...
+
+  cugraph::edge_property_t<graph_view_t<vertex_t, edge_t, false, multi_gpu>, edge_t> counts(
+    handle, graph_view);
+
+  cugraph::edge_bucket_t<vertex_t, void, true, multi_gpu, true> valid_edges(handle);
+  valid_edges.insert(edgelist_srcs.begin(), edgelist_srcs.end(), edgelist_dsts.begin());
+
+  auto cur_graph_view = graph_view;
+
+  auto edge_last = edge_first + edgelist_srcs.size();  // FIXME: Remove this unnecessary variable
```
seunghwak, line 337: You can make this update in the capture list (i.e. line 346) to minimize the scope of this temporary..

cpp/tests/community/edge_triangle_count_test.cpp
```diff
@@ -0,0 +1,260 @@
...
+}
+
+INSTANTIATE_TEST_SUITE_P(
+  simple_test,
+  Tests_EdgeTriangleCount_File,
+  ::testing::Combine(
+    // enable correctness checks
+    ::testing::Values(EdgeTriangleCount_Usecase{false, false, true},
+                      EdgeTriangleCount_Usecase{
+                        true, false, true}),  // FIXME: Still debugging edge_mask
```
seunghwak, line 234: I think we don't need test_weighted here. Weights have no role in TC. Need to include both edge_mask = false. and true  simple_test, rmat_small_test, and rmat_benchmark.

cpp/tests/community/mg_edge_triangle_count_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+#include <rmm/device_scalar.hpp>
+#include <rmm/device_uvector.hpp>
+
+#include <gtest/gtest.h>
+
+#include <random>
+
+struct EdgeTriangleCount_Usecase {
+  bool test_weighted_{false};
```
seunghwak, line 43: I don't think we need this. Weights are irrelevant in TC.

cpp/src/community/edge_triangle_count_impl.cuh
jnke2016: > You mean after shuffle, right?

Yep

> So, this is a trade-off. If you sort-and-reduce, no need to use atomics. You can experiment which will be faster

Right

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -78,77 +78,298 @@ struct update_edges_p_r_q_r_num_triangles {
...
+        get_dataframe_buffer_begin(vertex_pair_buffer_tmp),
+        get_dataframe_buffer_begin(vertex_pair_buffer_tmp) + intersection_indices.size(),
+        extract_p_r_q_r<vertex_t, edge_t, decltype(edge_first)>{
+          prev_chunk_size,
+          0,
+          raft::device_span<size_t const>(intersection_offsets.data(), intersection_offsets.size()),
+          raft::device_span<vertex_t const>(intersection_indices.data(),
+                                            intersection_indices.size()),
+          edge_first});
+      // FIXME: Consolidate both functions
```
seunghwak, line 201: And please address FIXMEs you can quickly address. We can leave performance optimizations or something that requires new primitives for future FIXMEs, but something you can independently and quickly address, we should address.

jnke2016: Right. And this FIXME was already addressed. Prior to this, I was launching two separate kernels to extract `p, r` and `q, r` edges which I consolidated to `extract_p_r_q_r`. I will remove the FIXME in my next commit

cpp/tests/utilities/conversion_utilities.hpp
```diff
@@ -216,15 +216,20 @@ graph_to_host_csc(
...
+                             weight_t>>,
+  std::optional<
+    cugraph::edge_property_t<cugraph::graph_view_t<vertex_t, edge_t, store_transposed, false>,
+                             edge_t>>,
+  std::optional<rmm::device_uvector<vertex_t>>>
 mg_graph_to_sg_graph(
   raft::handle_t const& handle,
   cugraph::graph_view_t<vertex_t, edge_t, store_transposed, true> const& graph_view,
   std::optional<cugraph::edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
+  std::optional<cugraph::edge_property_view_t<edge_t, edge_t const*>> edge_id_view,
```
ChuckHastings, line 232: Should we also add edge_type here, since we're modifying all of the functions?  We do currently support edge weight, edge id and edge type in most of our functions.

cpp/src/community/edge_triangle_count_impl.cuh
```diff
@@ -78,77 +78,297 @@ struct update_edges_p_r_q_r_num_triangles {
...
+      std::tie(pair_srcs, pair_dsts, std::ignore, pair_count, std::ignore) =
+        shuffle_int_vertex_pairs_with_values_to_local_gpu_by_edge_partitioning<vertex_t,
+                                                                               edge_t,
+                                                                               weight_t,
+                                                                               int32_t>(
+          handle,
+          std::move(std::get<0>(vertex_pair_buffer)),
+          std::move(std::get<1>(vertex_pair_buffer)),
+          std::nullopt,
+          // FIXME: Update 'shuffle_int_...' to support int32_t and int64_t values
```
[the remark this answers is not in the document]
jnke2016: >  You can use int32_t and int64_t values as you are currently doing this.

Right. This is currently how we proceed but based on Seunghwa feedbacks on `decompress_edgelist,` we should have something separate from `edge_id` to handle this in case `edge_id` and count don't hold the same type. I thought the same holds for the shuffling function.

cpp/tests/utilities/conversion_utilities.hpp
jnke2016: Right

cpp/src/community/edge_triangle_count_impl.cuh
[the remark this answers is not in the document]
ChuckHastings: I guess I'm asking for a more descriptive FIXME.  As worded it is not clear to me what you are asking to be changed at some point in the future.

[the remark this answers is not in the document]
jnke2016: Right. I updated the FIXME and it will be available in the next commit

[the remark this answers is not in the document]
jnke2016: Done

ChuckHastings (approved):

seunghwak (approved):
LGTM

PR #4387, opened 2024-05-03
Address deprecated summetrize_df(multi=False) calls in Python tests

mroeschke:
Broken off from https://github.com/rapidsai/cugraph/pull/4271

Deprecated in favor of `multi=True` being the default


mroeschke:
I am not sure really how to address these so closing this PR

PR #4389, opened 2024-05-06
Address k_truss_subgraph(use_weights=) deprecation

mroeschke:
This deprecated parameter appears to be a no-op but can raise the warning through `k_truss` with no addressable action from a user.


rlratzel (approved):
Thanks!

PR #4393, opened 2024-05-07
Use rapids-build-backend

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/build-planning/issues/31
Contributes to https://github.com/rapidsai/dependency-file-generator/issues/89

Proposes introducing `rapids-build-backend` as this project's build backend, to reduce the complexity of various CI/build scripts.


rlratzel:
It's my understanding that this PR should now target 24.08 (see [here](https://github.com/rapidsai/build-planning/issues/31#issuecomment-2113037773))

python/cugraph-equivariant/cugraph_equivariant/_version.py
```diff
@@ -10,18 +11,21 @@
...
-# source files.
 __version__ = (
-    importlib.resources.files("cugraph_equivariant")
-    .joinpath("VERSION")
-    .read_text()
-    .strip()
+    importlib.resources.files(__package__).joinpath("VERSION").read_text().strip()
 )
-__git_commit__ = ""
+try:
```
jameslamb, line 21: Could you please add unit tests for all the projects that already have unit test setups, like this:

```python
import cugraph_equivariant


def test_version_constants_are_populated():
    # __git_commit__ will only be non-empty in a built distribution
    assert isinstance(import cugraph_equivariant.__git_commit__, str)

    # __version__ should always be non-empty
    assert isinstance(import cugraph_equivariant.__version__, str)
    assert len(import cugraph_equivariant.__version__) > 0
```

I did this over in `cudf` (https://github.com/rapidsai/cudf/pull/15245 and https://github.com/rapidsai/cudf/pull/15853) and from conversations offline, we'd like to roll that out with all the `rapids-build-backend` updates.

python/cugraph-equivariant/cugraph_equivariant/_version.py
```diff
@@ -10,18 +11,21 @@
...
+    __git_commit__ = (
+        importlib.resources.files(__package__)
+        .joinpath("GIT_COMMIT")
+        .read_text()
+        .strip()
+    )
+except FileNotFoundError:
+    __git_commit__ = ""
+
+__all__ = ["__version__", "__git_commit__"]
```
jameslamb, line 31: If you do push another commit here, could you please re-order this (and all the similar `__all__` constants) so it's in alphabetical order? Just so it'd be a little easier to read at a glance if it grows in the future.

jameslamb (commented):
I left 2 comments for your consideration, leaving a non-blocking "comment" review so my review doesn't block merging whenever those (and other reviewers' comments) are addressed.

jameslamb:
The `conda-cpp-build` failures look like the same MPI issue reported in #4474, and not directly related to this PR's changes.

Once that's fixed, we'll be able to test conda builds here.

python/cugraph-service/client/pyproject.toml
```diff
@@ -43,3 +44,8 @@ version = {file = "cugraph_service_client/VERSION"}
 include = [
     "cugraph_service_client",
 ]
+
+[tool.rapids-build-backend]
+build-backend = "setuptools.build_meta"
+dependencies-file = "../../../dependencies.yaml"
+disable-cuda = true
```
jameslamb, line 51: I just added this.

`cugraph-service-client` doesn't have any CUDA-specific dependencies and we don't publish any wheels for it, so it should always have `disable-cuda=true`.

jameslamb:
The test failures look like this issue unrelated to this PR: https://github.com/rapidsai/cugraph/pull/4496#issuecomment-2192783353.

I believe this PR is now ready for review.

msarahan (approved):

vyasr (approved):

BradReesWork:
LGTM from a doc codeowner perspective

BradReesWork (approved):

jameslamb:
The other changes on `branch-24.08` since this was last built don't look like they'd affect or be affected by these changes, so I'm going to merge this. Thanks everyone!

PR #4395, opened 2024-05-08
nx-cugraph: add `ego_graph`

eriknw:


rlratzel (approved):
LGTM, although I'm not familiar with ego_graph implementations.  

I am curious why this isn't using `plc.ego_graph()`, is there a reason?

PR #4397, opened 2024-05-08
Update decompress_to_edgelist to handle edge types

naimnv:
Update decompress_to_edgelist to handle edge types

### DIFF
--- a/cpp/include/cugraph/detail/decompress_edge_partition.cuh
+++ b/cpp/include/cugraph/detail/decompress_edge_partition.cuh
@@ -209,7 +209,11 @@ void decompress_edge_partition_to_fill_edgelist_majors(
   }
 }
 
-template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+template <typename vertex_t,
+          typename edge_t,
+          typename weight_t,
+          typename edge_type_t,
+          bool multi_gpu>
 void decompress_edge_partition_to_edgelist(
   raft::handle_t const& handle,
   edge_partition_device_view_t<vertex_t, edge_t, multi_gpu> edge_partition,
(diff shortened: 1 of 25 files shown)


seunghwak (commented):
LGTM but @jnke2016 needs a decompress_to_edgelist function that takes just a graph view object and an edge_property_device_view_t object of integer types (i.e. int32_t and int64_t for now, and in the future, we may need to support floating point types as well). Do you think it is better to implement this in this PR or address this in a separate PR?

naimnv:
> LGTM but @jnke2016 needs a decompress_to_edgelist function that takes just a graph view object and an edge_property_device_view_t object of integer types (i.e. int32_t and int64_t for now, and in the future, we may need to support floating point types as well). Do you think it is better to implement this in this PR or address this in a separate PR?

He can just pass it instead of edge_id_view if edge_property_device_view_t object is of integer types (i.e. int32_t and int64_t 
or pass it instead of edge_weights_view if edge_property_device_view_t object is of floating point types. And he needs for triangle count tests, IIRC

naimnv:
> > LGTM but @jnke2016 needs a decompress_to_edgelist function that takes just a graph view object and an edge_property_device_view_t object of integer types (i.e. int32_t and int64_t for now, and in the future, we may need to support floating point types as well). Do you think it is better to implement this in this PR or address this in a separate PR?
> 
> He can just pass it instead of edge_id_view if edge_property_device_view_t object is of integer types (i.e. int32_t and int64_t or pass it instead of edge_weights_view if edge_property_device_view_t object is of floating point types. And he needs for triangle count tests, IIRC

Otherwise we need an additional 72 explicit instantiations

seunghwak:
> > LGTM but @jnke2016 needs a decompress_to_edgelist function that takes just a graph view object and an edge_property_device_view_t object of integer types (i.e. int32_t and int64_t for now, and in the future, we may need to support floating point types as well). Do you think it is better to implement this in this PR or address this in a separate PR?
> 
> He can just pass it instead of edge_id_view if edge_property_device_view_t object is of integer types (i.e. int32_t and int64_t or pass it instead of edge_weights_view if edge_property_device_view_t object is of floating point types. And he needs it for triangle count tests, IIRC.
> 
> Otherwise we need an additional 72 explicit instantiations and I can address it in a separate PR.

Yeah... let me think more about this... There is no guarantee that the integer type of the edge_property_device_view_t object and edge_id_t will coincide, and this solution sounds like a hack than a proper solution. But too many explicit instantiations are also problematic.

seunghwak (approved):

jnke2016 (approved):

ChuckHastings (approved):

PR #4404, opened 2024-05-08
Fix building cugraph with CCCL main

trxcllnt:
Similar to https://github.com/rapidsai/cudf/pull/15552, we are testing [building RAPIDS with CCCL's main branch](https://github.com/NVIDIA/cccl/pull/1667) to get ahead of any breaking changes.

### DIFF
--- a/cpp/include/cugraph/utilities/mask_utils.cuh
+++ b/cpp/include/cugraph/utilities/mask_utils.cuh
@@ -160,7 +160,7 @@ size_t count_set_bits(raft::handle_t const& handle, MaskIterator mask_first, siz
     handle.get_thrust_policy(),
     thrust::make_counting_iterator(size_t{0}),
     thrust::make_counting_iterator(packed_bool_size(num_bits)),
-    [mask_first, num_bits] __device__(size_t i) {
+    [mask_first, num_bits] __device__(size_t i) -> size_t {
       auto word = *(mask_first + i);
       if ((i + 1) * packed_bools_per_word() > num_bits) {
         word &= packed_bool_partial_mask(num_bits % packed_bools_per_word());
--- a/cpp/src/community/detail/common_methods.cuh
+++ b/cpp/src/community/detail/common_methods.cuh
@@ -178,7 +178,7 @@ weight_t compute_modularity(
     handle.get_thrust_policy(),
     cluster_weights.begin(),
     cluster_weights.end(),
-    [] __device__(weight_t p) { return p * p; },
+    [] __device__(weight_t p) -> weight_t { return p * p; },
     weight_t{0},
     thrust::plus<weight_t>());
 
--- a/cpp/src/community/legacy/louvain.cuh
+++ b/cpp/src/community/legacy/louvain.cuh
@@ -142,7 +142,7 @@ class Louvain {
       thrust::make_counting_iterator(0),
       thrust::make_counting_iterator(graph.number_of_vertices),
       [d_deg = deg.data(), d_inc = inc.data(), total_edge_weight, resolution] __device__(
-        vertex_t community) {
+        vertex_t community) -> weight_t {
         return ((d_inc[community] / total_edge_weight) - resolution *
                                                            (d_deg[community] * d_deg[community]) /
                                                            (total_edge_weight * total_edge_weight));
--- a/cpp/src/components/weakly_connected_components_impl.cuh
+++ b/cpp/src/components/weakly_connected_components_impl.cuh
@@ -400,7 +400,7 @@ void weakly_connected_components_impl(raft::handle_t const& handle,
         handle.get_thrust_policy(),
         new_root_candidates.begin(),
         new_root_candidates.begin() + (new_root_candidates.size() > 0 ? 1 : 0),
-        [vertex_partition, degrees = degrees.data()] __device__(auto v) {
+        [vertex_partition, degrees = degrees.data()] __device__(auto v) -> edge_t {
           return degrees[vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(v)];
         },
         edge_t{0},
@@ -642,7 +642,7 @@ void weakly_connected_components_impl(raft::handle_t const& handle,
         handle.get_thrust_policy(),
         thrust::get<0>(vertex_frontier.bucket(bucket_idx_cur).begin().get_iterator_tuple()),
         thrust::get<0>(vertex_frontier.bucket(bucket_idx_cur).end().get_iterator_tuple()),
-        [vertex_partition, degrees = degrees.data()] __device__(auto v) {
+        [vertex_partition, degrees = degrees.data()] __device__(auto v) -> edge_t {
           return degrees[vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(v)];
         },
         edge_t{0},
--- a/cpp/src/detail/utility_wrappers.cu
+++ b/cpp/src/detail/utility_wrappers.cu
@@ -139,7 +139,7 @@ vertex_t compute_maximum_vertex_id(rmm::cuda_stream_view const& stream_view,
     rmm::exec_policy(stream_view),
     edge_first,
     edge_first + num_edges,
-    [] __device__(auto e) { return std::max(thrust::get<0>(e), thrust::get<1>(e)); },
+    [] __device__(auto e) -> vertex_t { return std::max(thrust::get<0>(e), thrust::get<1>(e)); },
     vertex_t{0},
     thrust::maximum<vertex_t>());
 }
--- a/cpp/src/structure/graph_view_impl.cuh
+++ b/cpp/src/structure/graph_view_impl.cuh
@@ -353,7 +353,7 @@ edge_t count_edge_partition_multi_edges(
         execution_policy,
         thrust::make_counting_iterator(edge_partition.major_range_first()) + (*segment_offsets)[2],
         thrust::make_counting_iterator(edge_partition.major_range_first()) + (*segment_offsets)[3],
-        [edge_partition] __device__(auto major) {
+        [edge_partition] __device__(auto major) -> edge_t {
           auto major_offset = edge_partition.major_offset_from_major_nocheck(major);
           vertex_t const* indices{nullptr};
           [[maybe_unused]] edge_t edge_offset{};
@@ -374,7 +374,8 @@ edge_t count_edge_partition_multi_edges(
         execution_policy,
         thrust::make_counting_iterator(vertex_t{0}),
         thrust::make_counting_iterator(*(edge_partition.dcs_nzd_vertex_count())),
-        [edge_partition, major_start_offset = (*segment_offsets)[3]] __device__(auto idx) {
+        [edge_partition,
+         major_start_offset = (*segment_offsets)[3]] __device__(auto idx) -> edge_t {
           auto major_idx =
             major_start_offset + idx;  // major_offset != major_idx in the hypersparse region
           vertex_t const* indices{nullptr};
@@ -398,7 +399,7 @@ edge_t count_edge_partition_multi_edges(
       thrust::make_counting_iterator(edge_partition.major_range_first()),
       thrust::make_counting_iterator(edge_partition.major_range_first()) +
         edge_partition.major_range_size(),
-      [edge_partition] __device__(auto major) {
+      [edge_partition] __device__(auto major) -> edge_t {
         auto major_offset = edge_partition.major_offset_from_major_nocheck(major);
         vertex_t const* indices{nullptr};
         [[maybe_unused]] edge_t edge_offset{};
--- a/cpp/tests/sampling/sampling_post_processing_test.cu
+++ b/cpp/tests/sampling/sampling_post_processing_test.cu
@@ -402,7 +402,7 @@ bool check_renumber_map_invariants(
          raft::device_span<vertex_t const>(sorted_org_vertices.data(), sorted_org_vertices.size()),
        matching_renumbered_vertices = raft::device_span<vertex_t const>(
          matching_renumbered_vertices.data(),
-         matching_renumbered_vertices.size())] __device__(vertex_t major) {
+         matching_renumbered_vertices.size())] __device__(vertex_t major) -> vertex_t {
         auto it = thrust::lower_bound(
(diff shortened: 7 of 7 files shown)


cpp/tests/CMakeLists.txt
```diff
@@ -661,7 +661,7 @@ if(BUILD_CUGRAPH_MG_TESTS)
 
     ###############################################################################################
     # - MG PRIMS EXTRACT_TRANSFORM_E tests --------------------------------------------------------
-    ConfigureTestMG(MG_EXTRACT_TRANSFORM_E_TEST prims/mg_extract_transform_e.cu)
+    # ConfigureTestMG(MG_EXTRACT_TRANSFORM_E_TEST prims/mg_extract_transform_e.cu)
```
KyleFromNVIDIA, line 664: Why has this been commented out?

trxcllnt: Currently this test doesn't build, so it's commented out (for now) to unblock CCCL CI. I will be posting a longer comment about why it doesn't build, because it may involve fixing a larger issue around cugraph relying on UB by constructing `thrust::tuple` from a scalar.

trxcllnt:
The [new `kv_store_t` constructor](https://github.com/rapidsai/cugraph/pull/4404/files#diff-2dd0046c3b6293ec7dbc6be602d5633ec3a7cfd9026e441aadbd63a0a8fab017R948-R971) is necessary because there are code paths that attempt to construct a `thrust::tuple` [from a scalar (`invalid_vertex_id<vertex_t>::value` here)](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/src/prims/per_v_transform_reduce_dst_key_aggregated_outgoing_e.cuh#L753-L759).

While this [is possible](https://godbolt.org/z/anjdvY4az) in CCCL 2.2.0's version of `thrust::tuple`, I checked with @jrhemstad it was definitely never intended to work. CCCL 2.4+ makes `thrust::tuple` an alias of `cuda::std::tuple`, so this behavior is now gone, and cuGraph needs to be updated.

I [commented out](https://github.com/rapidsai/cugraph/pull/4404/files#diff-7caf2bdc6aa44337cf6e4b04392bc3e5940a7356512b4d525aba3bcda0be6c3bR664) the `prims/mg_extract_transform_e.cu` test for now, because the compiler resisted my attempts to rewrite [this functor](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/tests/prims/mg_extract_transform_e.cu#L69-L116) to be compatible with the new CCCL behavior.

The problem case is when the `key_t` template parameter is `thrust::tuple<vertex_t, int32_t>`, but the functor is called with the invalid value scalar `-1`.

Alternatively, I attempted to specialize the [`invalid_idx` struct](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/include/cugraph/graph.hpp#L307-L326) to return a `thrust::tuple` if the `vertex_t` template parameter is a tuple, but that seemed to break in other ways.

cc: @BradReesWork if you have any ideas or know the right person to help fix this.

trxcllnt:
The [nvcc error](https://github.com/rapidsai/cugraph/actions/runs/9023529776/job/24795543523?pr=4404#step:7:1813) tells us half of the issue:
```
mg_extract_transform_e.cu(210): error: no instance of function template "cugraph::extract_transform_e" matches the argument list
              argument types are: (raft::handle_t, cugraph::graph_view_t<int32_t, int32_t, false, true, void>, cugraph::detail::edge_major_property_view_t<int32_t, const int32_t *, int32_t>, cugraph::detail::edge_minor_property_view_t<int32_t, const int32_t *, int32_t>, cugraph::edge_dummy_property_view_t, e_op_t<thrust::tuple<int32_t, int32_t>, int32_t, int32_t, int32_t>)
          cugraph::extract_transform_e(*handle_,
          ^
```

The error from clang tells us the other half:
```
extract_transform_e.cuh(91, 1): Candidate template ignored: substitution failure [with
  GraphViewType = graph_view_t<int, int, false, true>,
  EdgeSrcValueInputWrapper = edge_major_property_view_t<int, const_value_iterator, int>,
  EdgeDstValueInputWrapper = edge_minor_property_view_t<int, const_iterator, int>,
  EdgeValueInputWrapper = edge_dummy_property_view_t,
  EdgeOp = e_op_t<key_t, int, result_t, int>
]: implicit instantiation of undefined template 'cugraph::detail::edge_op_result_type<int, int, int, int, thrust::nullopt_t, e_op_t<int, int, int, int>>'
```

Combining info from both errors, the issue is that when `extract_transform_e` [instantiates](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/src/prims/extract_transform_e.cuh#L100-L106) `detail::edge_op_result_type`, it tries to get the return type of invoking `e_opt_t` with `int, int, int, int`.

But `e_opt_t` was declared with template parameters `thrust::tuple<int32_t, int32_t>, int32_t, int32_t, int32_t`, and since thrust tuples can no longer be default constructed from a scalar, the `operator()()` overload cannot be invoked with `int` as the first argument.

ChuckHastings:
> The [new `kv_store_t` constructor](https://github.com/rapidsai/cugraph/pull/4404/files#diff-2dd0046c3b6293ec7dbc6be602d5633ec3a7cfd9026e441aadbd63a0a8fab017R948-R971) is necessary because there are code paths that attempt to construct a `thrust::tuple` [from a scalar (`invalid_vertex_id<vertex_t>::value` here)](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/src/prims/per_v_transform_reduce_dst_key_aggregated_outgoing_e.cuh#L753-L759).
> 
> While this [is possible](https://godbolt.org/z/anjdvY4az) in CCCL 2.2.0's version of `thrust::tuple`, I checked with @jrhemstad it was definitely never intended to work. CCCL 2.4+ makes `thrust::tuple` an alias of `cuda::std::tuple`, so this behavior is now gone, and cuGraph needs to be updated.
> 
> I [commented out](https://github.com/rapidsai/cugraph/pull/4404/files#diff-7caf2bdc6aa44337cf6e4b04392bc3e5940a7356512b4d525aba3bcda0be6c3bR664) the `prims/mg_extract_transform_e.cu` test for now, because the compiler resisted my attempts to rewrite [this functor](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/tests/prims/mg_extract_transform_e.cu#L69-L116) to be compatible with the new CCCL behavior.
> 
> The problem case is when the `key_t` template parameter is `thrust::tuple<vertex_t, int32_t>`, but the functor is called with the invalid value scalar `-1`.
> 
> Alternatively, I attempted to specialize the [`invalid_idx` struct](https://github.com/rapidsai/cugraph/blob/af749c3974f658039b5dda5742d407835d1e30f4/cpp/include/cugraph/graph.hpp#L307-L326) to return a `thrust::tuple` if the `vertex_t` template parameter is a tuple, but that seemed to break in other ways.
> 
> cc: @BradReesWork if you have any ideas or know the right person to help fix this.

@seunghwak is our expert in this area

.github/workflows/pr.yaml
```diff
@@ -195,6 +195,21 @@ jobs:
...
+          | jq -r ".packages.cuco *= {\"git_url\": \"https://github.com/trxcllnt/cuCollections.git\", \"git_tag\": \"rapids-${rapids_version}-cccl-2.5.0\", \"always_download\": true}" \
+          | tee ~/rapids-cmake-override-versions.json;
         sccache -z;
-        build-all --verbose -j$(nproc --ignore=1);
+        build-all \
+          -j$(nproc --ignore=1) -v \
+          -DBUILD_CUGRAPH_MG_TESTS=ON \
+          -DCMAKE_CXX_FLAGS="-ftemplate-backtrace-limit=0" \
+          -DCMAKE_CUDA_FLAGS="-ftemplate-backtrace-limit=0" \
+          -DRAPIDS_CMAKE_CPM_DEFAULT_VERSION_FILE="${HOME}/rapids-cmake-override-versions.json";
```
trxcllnt, line 214: Need to remove all this before merging

rapids_config.cmake
```diff
@@ -25,6 +25,9 @@ else()
       "Could not determine RAPIDS version. Contents of VERSION file:\n${_rapids_version_formatted}")
 endif()
 
+set(rapids-cmake-repo trxcllnt/rapids-cmake)
+set(rapids-cmake-branch fea/cccl-2.5)
+
```
trxcllnt, line 30: Need to revert this before merging:

```suggestion
```

cpp/libcugraph_etl/cmake/thirdparty/get_cudf.cmake
```diff
@@ -42,6 +42,6 @@ set(CUGRAPH_ETL_BRANCH_VERSION_cudf "${CUGRAPH_ETL_VERSION_MAJOR}.${CUGRAPH_ETL_
 # To use a different RAFT locally, set the CMake variable
 # RPM_cudf_SOURCE=/path/to/local/cudf
 find_and_configure_cudf(VERSION    ${CUGRAPH_ETL_MIN_VERSION_cudf}
-                        FORK       rapidsai
-                        PINNED_TAG branch-${CUGRAPH_ETL_BRANCH_VERSION_cudf}
+                        FORK       trxcllnt
+                        PINNED_TAG fix/cccl-2.5
```
trxcllnt, line 46: Need to revert this before merging:

```suggestion
                        FORK       rapidsai
                        PINNED_TAG branch-${CUGRAPH_ETL_BRANCH_VERSION_cudf}
```

cpp/libcugraph_etl/cmake/thirdparty/get_cudf.cmake
```diff
@@ -1,5 +1,5 @@
 #=============================================================================
-# Copyright (c) 2021, NVIDIA CORPORATION.
+# Copyright (c) 2021-2024, NVIDIA CORPORATION.
```
trxcllnt, line 2: Need to revert this before merging:

```suggestion
# Copyright (c) 2021, NVIDIA CORPORATION.
```

ci/build_cpp.sh
```diff
@@ -17,6 +17,7 @@ version=$(rapids-generate-version)
 
 rapids-logger "Begin cpp build"
 
+PARALLEL_LEVEL="8" \
```
trxcllnt, line 20:
```suggestion
PARALLEL_LEVEL="16" \
```

ci/build_wheel.sh
```diff
@@ -56,6 +56,8 @@ fi
 
 cd "${package_dir}"
 
+PARALLEL_LEVEL="8" \
+CMAKE_BUILD_PARALLEL_LEVEL="8" \
```
trxcllnt, line 60:
```suggestion
PARALLEL_LEVEL="16" \
CMAKE_BUILD_PARALLEL_LEVEL="16" \
```

ChuckHastings (approved):

cpp/src/prims/kv_store.cuh
```diff
@@ -945,6 +946,30 @@ class kv_store_t {
   {
   }
 
+  /* when use_binary_search = true */
+  template <bool binary_search = use_binary_search>
+  kv_store_t(rmm::device_uvector<key_t>&& keys,
+             decltype(allocate_dataframe_buffer<value_t>(0, rmm::cuda_stream_view{}))&& values,
+             decltype(cugraph::invalid_idx<key_t>::value)
+               invalid_value /* invalid_value is returned when match fails for the given key */,
```
seunghwak, line 954: This should be `value_t invalid_value`, and `invalid_value` should be a `thrust::tuple_type`. I think you introduced to fix a compile error, could you specify the code causing the error? I think we should fix this in the caller site, not here.

seunghwak: https://github.com/rapidsai/cugraph/blob/branch-24.06/cpp/src/utilities/collect_comm.cuh#L302

```
  kv_store_t<vertex_t, value_t, true> kv_map(std::move(unique_int_vertices),
                                             std::move(tmp_value_buffer),
                                             invalid_vertex_id<vertex_t>::value,
                                             false,
                                             handle.get_stream());
```

Is this the culprit? Let me think about how to fix this.

seunghwak: You can just provide `value_t{}`. Here, kv_map is guaranteed to have values for all the input keys, so `invalid_value` should never be used.

seunghwak: We never tried to update a thrust::tuple object from a scalar. This is just an undetected bug as the old CCCL allowed this behavior.

trxcllnt: @seunghwak it's possible your changes fixed this, but IIRC the issue was from this callsite: https://github.com/rapidsai/cugraph/blob/562b5a5b9f3db29184390c319468ccb488d21056/cpp/src/prims/per_v_transform_reduce_dst_key_aggregated_outgoing_e.cuh#L753-L759

seunghwak: Yes, same here, let me double check but I think `invalid_vertex_id<vertex_t>::value` should be `edge_src_value_t{}`

seunghwak: Yes, this caller site code mistakenly passes an invalid key (instead of an invalid value).

`invalid_vertex_id<vertex_t>::value => edge_src_value_t{}` (as invalid values are irrelevant in this context).

trxcllnt: Thanks, I'll rebuild locally and verify these changes work with the new constructor removed.

seunghwak: Thanks for fixing cugraph bugs :-)

trxcllnt: No problem! [My latest commit](https://github.com/rapidsai/cugraph/pull/4404/commits/06dee7ebae8b87e1198eabf1b7837bbcc1388a3d) builds for me locally w/ CCCL 2.5, so I think we're good to go.

seunghwak (approved):
LGTM

AyodeAwe (approved):

PR #4408, opened 2024-05-09
Updates SG `PropertyGraph` and `cugraph-service` to apply `DataFrame.fillna()` based on latest cuDF changes

rlratzel:
This handles a [recent cuDF change](https://github.com/rapidsai/cudf/pull/15683) by applying non-dict and non-Series values for a `fillna()` call on `PropertyGraph` instances only to the user-defined columns, with the assumption that savvy users that intend to update the "internal" columns, or users that are aware of their own categorical dtype columns, will use a dict or Series value to properly apply dtypes as needed.

This also updates code in `cugraph-service` that serializes dataframes to numpy bytes to properly convert NA values when categoricals are present.

Notes:
* This is only applied to the SG `PropertyGraph` class.  The MG class needs further review as to how to best apply the same policy (and because there are other MG failing tests that need addressed).  Since this is blocking CI for the SG case only, this PR is being submitted now and MG will be addressed later, which should be okay since `PropertyGraph` is experimental.
* This could be considered a breaking change if `PropertyGraph` was not experimental.

### DIFF
--- a/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
+++ b/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2023, NVIDIA CORPORATION.
+# Copyright (c) 2022-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


alexbarghi-nv (approved):
👍

PR #4411, opened 2024-05-10
Add MTMG support for Similarity and test for all-pairs Jaccard

ChuckHastings:
Added to the MTMG library support for results that contain a vertex pair and a result.  This will allow for support for all of the similarity algorithms.

Added an MTMG test that demonstrates how to call all-pairs Jaccard.

Also cleaned up the cmake files to better find ucx/ucs/ucp.

Closes https://github.com/rapidsai/graph_dl/issues/331


cpp/CMakeLists.txt
```diff
@@ -77,6 +77,16 @@ rapids_find_package(CUDAToolkit REQUIRED
     INSTALL_EXPORT_SET  cugraph-exports
     )
 
+if (BUILD_CUGRAPH_MTMG_TESTS)
+    if(NOT TARGET ucx::ucp)
+        find_package(ucx REQUIRED)
```
KyleFromNVIDIA, line 82: Does UCP have its own CMake buildsystem now?

ChuckHastings: I have no idea.  We install ucx/ucp/ucs as part of the conda environment, this should just find it or fail.

KyleFromNVIDIA: Thanks. I just installed the `ucxx` package from Conda and see that it provides `ucxx-config.cmake` - that's what's needed for this to work.

KyleFromNVIDIA (approved):

seunghwak (approved):
LGTM

PR #4414, opened 2024-05-13
Fix a build error in MG extract_transform_e tests (with the new CCCL)

seunghwak:
extract_transform_e cannot take a tagged key in e_op but tests including a tagged key compiled in the previous test because of a thrust bug (which implicitly converts a scalar to a thrust tuple of size 2). This bug is detected with the new CCCL and this PR removes a test case in extract_transform_e that includes a tagged key.


seunghwak:
This PR is merged into https://github.com/rapidsai/cugraph/pull/4404

PR #4426, opened 2024-05-16
Expose graph_view method to count multi edges through C API and PLC

### LINKED ISSUE #2417: [FEA] Expose `count_multi_edges `  for cuGraph C++ graph_view_t  to python

We should expose count_multi_edges  for cuGraph C++ graph_view_t  to python to check if a graph has multi_edges. 

It is possible  for that  it can be used to `has_duplicates` in Property Graphs.  
Currently we do expensive `has_duplicates` call  [here](https://github.com/rapidsai/cugraph/blob/049d44197ae775fbc5707a8fdee8cba94ebadb97/python/cugraph/cugraph/structure/property_graph.py#L807-L816 ) and using that `count_multi_edges` might be more optimial. 

For more context see discussion [here](https://github.com/rapidsai/cugraph/pull/2409#issuecomment-1184929827)

ChuckHastings:
Python property graph class does some expensive dask calls to determine if a graph is a multi-graph.

This PR exposes to the C API and the PLC layer the graph_view method that counts the number of multi-edges in the graph.  Specifically, the function returns the number of extra edges exist.  So if the edge `(u,v)` exists `k` times, that edge contributes `k-1` to the return value.

If the return value of this function is greater than 0 then the graph is a multi-graph.  If the return value is 0 then while the graph might be labeled a multi-graph, there are no multi-edges so it could be treated as a graph.

Closes #2417 

Note that the property graph python code would still need to be modified to use this function.


seunghwak (approved):

jnke2016 (approved):

alexbarghi-nv (approved):
👍

PR #4427, opened 2024-05-17
Mace conv layer

tingyu66:


tingyu66:
Close this as channelwise TP and relevant mace operators will land in cuequivariance repo.

PR #4429, opened 2024-05-17
Fix MG Betweenness Centrality Test Bug

nv-rliu:
Fixes bug seen in MG tests due to incorrect call to `enable_batch`


alexbarghi-nv (approved):
👍

PR #4443, opened 2024-05-24
Biased sampling

### LINKED ISSUE #4290: Implement C++ Biased Sampling
Implement C++ biased sampling including C++ tests.

seunghwak:
Implement biased sampling using the biased sampling primitive. The public biased sampling function takes bias values as edge property values; otherwise same to the uniform neighbor sampling function.

Closes #4290


cpp/src/sampling/detail/check_edge_bias_values.cuh
```diff
@@ -0,0 +1,73 @@
...
+               edge_dst_dummy_property_t{}.view(),
+               edge_bias_view,
+               [] __device__(vertex_t, vertex_t, auto, auto, bias_t b) { return b < 0.0; });
+
+  size_t num_overflows{0};
+  {
+    auto bias_sums = compute_out_weight_sums(handle, graph_view, edge_bias_view);
+    num_overflows  = thrust::count_if(
+      handle.get_thrust_policy(), bias_sums.begin(), bias_sums.end(), [] __device__(auto sum) {
+        return sum > std::numeric_limits<bias_t>::max();
```
ChuckHastings, line 58: This check seems suspect to me.  `sum` is going to be of type `bias_t`.  I don't see how it could ever have a value that is strictly greater than `std::numeric_limits<bias_t>::max()`.

Should this check for `std::isinf(sum)` or equality to `std::numeric_limits<bias_t>::max()`?

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -65,8 +69,10 @@ uniform_neighbor_sample_impl(
 {
 #ifdef NO_CUGRAPH_OPS
```
ChuckHastings, line 70: Is this check still correct?

I believe cugraph-ops is only required if edge_bias_view is `std::nullopt`.   If edge_bias_view is specified then we're not using cugraph-ops.

Of course, removing this ifdef and checking at runtime may require creating an #ifdef in the sampling primitive somewhere to make sure that we don't try and include the cugraph-ops code.  Haven't checked to see what the plan is.

Perhaps a FIXME would be sufficient, since we hope to break the dependency on cugraph-ops in the release anyway.

cpp/src/sampling/detail/check_edge_bias_values.cuh
seunghwak: If overflows, it will become std::numeric_limits<bias_t>::infinity();

cpp/src/sampling/neighbor_sampling_impl.hpp
seunghwak: Yes, this `ifdef` is relevant only when `edge_bias_view` is std::nullopt and we can defer this check till sampling primitives.

My thought was that we will fix this anyways when we migrate cugraph-ops functions to cugraph, so better fix this at once then creating an intermediate fix.

I can add FIXME as suggestsed.

cpp/src/sampling/detail/check_edge_bias_values.cuh
ChuckHastings: Here's an example...
```
#include <iostream>
#include <limits>

int main(int argc, char **argv) {
  double x = std::numeric_limits<double>::max();
  double y = std::numeric_limits<double>::infinity();
  x += 1;

  if (x > std::numeric_limits<double>::max())
    std::cout << "x comparison works..." << std::endl;
  else
    std::cout << "x comparison fails" << std::endl;

  if (y > std::numeric_limits<double>::max())
    std::cout << "y comparison works..." << std::endl;
  else
    std::cout << "y comparison fails" << std::endl;
}
```

Output when I run this is:
```
x comparison fails
y comparison works...
```

I'm not sure that overflow guarantees that it gets set to infinity.  Of course that's a numerical instability issue more than anything else, so perhaps not worthy of struggling with.

seunghwak: Oh, max() + 1 is still max() as max() for double is 1.79769e+308 and # significant digits for double is 16-ish. Adding 1 won't affect any numbers within the significant digit.

And AFAIK, how floating point numbers handle overflow is not strictly defined, but this is what I got from the following code.

```
#include <iostream>
#include <limits>

int main() {
  auto val = std::numeric_limits<double>::max();
  std:: cout << "val=" << val << " val + 1e308=" << (val + 1e308) << " (val + 1e308) - val=" << ((val + 1e308) - val) << " val * 2=" << val * 2 << " val + val=" << (val + val) << std::endl;
  return 0;
}
```

```
val=1.79769e+308 val + 1e308=inf (val + 1e308) - val=inf val * 2=inf val + val=inf
```

seunghwak: If I add 1e291, I don't get inf. If I add 1e292, I get inf. This test isn't perfect as floating point arithmetic gets highly inaccurate around their representation limit, but this test detects at least some overflow events.

ChuckHastings (approved):

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -25,6 +28,217 @@
 
 namespace cugraph {
 
+/**
+ * @brief Controls how we treat prior sources in sampling
+ *
+ * @param DEFAULT    Add vertices encounted while sampling to the new frontier
```
jnke2016, line 34: small  typo here `encountered`

jnke2016 (approved):

alexbarghi-nv (approved):
👍

naimnv (approved):

raydouglass:

PR #4447, opened 2024-05-27
Pin torch version in `cugraph-dgl` wheel test

tingyu66:
To fix the [CI nightly issue](https://github.com/rapidsai/cugraph/actions/runs/9188624604/job/25298484338#step:8:837) in cugraph-dgl wheel test for CUDA 11.


rlratzel (approved):
LGTM, thanks!  Too bad this can't be consolidated somehow in `dependencies.yaml`, but I don't think wheel installs and the rapids-dependency-file-generator have the ability to inspect things like the NCCL install to do this.

BradReesWork (approved):

AyodeAwe (approved):

PR #4451, opened 2024-05-29
Adds benchmark for `nx_cugraph.ego_graph`

rlratzel:
Adds benchmark for `nx_cugraph.ego_graph`

_Note: this code is not part of any installed package and does not affect the release.  It is being added so it can be included in the `branch-24.06` sources for reference._


alexbarghi-nv (approved):
👍

eriknw (approved):
LGTM, with the common caveat that this algorithm seems difficult to benchmark well in ways that match "real world" use cases (i.e., how should we choose initial nodes and `k`?), and `nx.ego_graph` has more functionality than is exercised here.

PR #4463, opened 2024-06-05
Define heterogeneous renumbering API

seunghwak:
Defines a heterogeneous renumbering function that renumbers sampled edge list vertex & edge IDs to consecutive integers per vertex & edge type, respectively.

The renumbered edges are sorted. New functions that sort & **compress** will be added later.


seunghwak:
@alexbarghi-nv In heterogeneous renumbering, can I assume that vertex types are always used? Or there can be a use case with no vertex type but just edge types?

Similarly, is there a use case with vertex types but no edge types & IDs? (i.e. should the input `edgelist_edge_ids` and `edgelist_edge_types` be std::optional?)

alexbarghi-nv:
> @alexbarghi-nv In heterogeneous renumbering, can I assume that vertex types are always used? Or there can be a use case with no vertex type but just edge types?
> 
> Similarly, is there a use case with vertex types but no edge types & IDs? (i.e. should the input `edgelist_edge_ids` and `edgelist_edge_types` be std::optional?)



> @alexbarghi-nv In heterogeneous renumbering, can I assume that vertex types are always used? Or there can be a use case with no vertex type but just edge types?
> 
> Similarly, is there a use case with vertex types but no edge types & IDs? (i.e. should the input `edgelist_edge_ids` and `edgelist_edge_types` be std::optional?)

It is safe to assume that if there are edge types, then there are also vertex types.

alexbarghi-nv:
This looks good to me.  I had a long question typed out about compression, but then my tab crashed and I lost it.  Anyways, we can worry about compression later.  This API looks good.

alexbarghi-nv (approved):
👍

alexbarghi-nv:
Because the performance gain of CSR format is pretty small for PyG, I think implementing this function is sufficient for our 24.08 goals for heterogeneous renumbering.

ChuckHastings (approved):

Issue #4474, opened 2024-06-07
Unpin `openmpi` Once Build Issues are Resolved


nv-rliu:
### Version

24.08

### Which installation method(s) does this occur on?

Source

### Describe the bug.

Currently `openmpi=5.0.3-hfd7b305_105` is blocking our CI `cpp_build` job.

Most likely introduced by this PR: https://github.com/conda-forge/openmpi-feedstock/pull/158

A PR to unblock this issue is up here (https://github.com/rapidsai/cugraph/pull/4473) to unblock cugraph development until the issues are fixed. Once that happens, the version pinning should be removed.

### Relevant log output

```shell
-- Could NOT find MPI_CXX (missing: MPI_CXX_LIB_NAMES MPI_CXX_WORKS) 
-- Configuring incomplete, errors occurred!
Traceback (most recent call last):
  File "/opt/conda/bin/conda-mambabuild", line 10, in <module>
    sys.exit(main())
  File "/opt/conda/lib/python3.10/site-packages/boa/cli/mambabuild.py", line 301, in main
    call_conda_build(action, config)
  File "/opt/conda/lib/python3.10/site-packages/boa/cli/mambabuild.py", line 273, in call_conda_build
    result = api.build(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/api.py", line 250, in build
    return build_tree(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 3762, in build_tree
    packages_from_this = build(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/build.py", line 2634, in build
    utils.check_call_env(
  File "/opt/conda/lib/python3.10/site-packages/conda_build/utils.py", line 408, in check_call_env
    return _func_defaulting_env_to_os_environ("call", *popenargs, **kwargs)
  File "/opt/conda/lib/python3.10/site-packages/conda_build/utils.py", line 384, in _func_defaulting_env_to_os_environ
    raise subprocess.CalledProcessError(proc.returncode, _args)
subprocess.CalledProcessError: Command '['/bin/bash', '-o', 'errexit', '/opt/conda/conda-bld/work/conda_build.sh']' returned non-zero exit status 1.
[rapids-conda-retry] conda returned exit code: 1
[...]

PR #4477, opened 2024-06-10
Forward merge 24.06 into 24.08

naimnv:
Forward merge 24.06 into 24.08


msarahan:
This is fine. It means you'll need to unpin stuff at some point, but if this passes CI, it's fine for now. https://github.com/rapidsai/cugraph/issues/4474

Issue #4459, opened 2024-06-03
[FEA]: clarify cugraph-equivariant's src vs dst semantics in weight-sharing bidirectionaly message passing


DejunL:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

High

### Please provide a clear description of problem this feature solves

In cugraph-equivariant's `FullyConnectedTensorProductConv` layer (FCTPConv), the 1st layer of the underlying MLP is decomposed into edge embedding, src and dst scalar weight blocks in that order (see code [here](https://github.com/rapidsai/cugraph/blob/ee4cbb7ffea4394fef4367cc02782657a0e94c16/python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py#L232-L236)). In one of our applications, DiffDock-PP, the same FCTPConv layer is called bidirectionally, sharing the same set of MLP weights (see the original code [here](https://github.com/ketatam/DiffDock-PP/blob/25a28900736c0730821e45265ee8e409751c358a/src/model/diffusion.py#L327) for one direction and [here](https://github.com/ketatam/DiffDock-PP/blob/25a28900736c0730821e45265ee8e409751c358a/src/model/diffusion.py#L348) for another.) Note that these links point to the code that hasn't been replaced with FCTPConv but imagine that situation once they are using FCTPConv: one of the two directions would have to violate the src vs dst assignment if the two directions keep sharing the same set of weights. 

The crux here is cugraph-equivariant's FCTPConv layer is inherently single-directional even though the user could shoehorn it into a bidirectionaly usage case by paying attentions to how the underlying MLP computation is carried out, e.g.:
```python
# first direction
a2b_edge_emb = torch.hstack([edge_emb, a_scalars, b_scalars])
cross_conv(a_node_attrs, edge_sh, a2b_edge_emb, (a2b_edge_index, (size_a, size_b)))

# second direction sharing the same weight as the first direction
b2a_edge_index = a2b_edge_index.flip(dims=(0,))
cross_conv(b_node_attrs, edge_sh, a2b_edg_emb, (b2a_edge_index, (size_b, size_a)))
```
where the semantic violation can be seen `a2b_edg_emb` being used in the case of a `b->a` convolution

### Describe your ideal solution

Not sure if there is any

### Describe any alternatives you have considered

[...]

PR #4480, opened 2024-06-11
Fix TensorProductConv test and improve docs

tingyu66:
Closes https://github.com/rapidsai/cugraph/issues/4459


python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -74,26 +79,28 @@ class FullyConnectedTensorProductConv(nn.Module):
...
-    >>> #         the input layer: 6
+    >>> # having 16 channels. edge_emb.size(1) must match the size of the input layer: 6
     >>>
     >>> conv1 = FullyConnectedTensorProductConv(in_irreps, sh_irreps, out_irreps,
     >>>     mlp_channels=[6, 16, 16], mlp_activation=nn.ReLU()).cuda()
     >>> out = conv1(src_features, edge_sh, edge_emb, graph)
     >>>
     >>> # Case 2: Same as case 1 but with the scalar features from edges, sources
-    >>> #         and destinations passed in separately.
+    >>> # and destinations passed in separately. This option allows a smaller GEMM
```
tingyu66, line 89: @DejunL Can you check if the description here is enough to clarify the usage?

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -74,26 +79,28 @@ class FullyConnectedTensorProductConv(nn.Module):
...
+    >>> # in the first MLP layer by performing GEMM on each component before indexing.
+    >>> # The first-layer weights are split into sections for edges, sources and
+    >>> # destinations, in that order.
```
DejunL, line 92: (in that order), which are to be multiplied with the respective input of `edge_emb`, `src_scalars` and `dst_scalars`. The following example is equivalent to calling the forward function in usage case 1 with  `conv1(src_features, edge_sh, torch.hstack([edge_emb, src_scalars[graph[0][0]], dst_scalars[graph[0][1]]]) , graph)`.

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
DejunL: Added a comment below. Can you also add to the input arguments `src_scalars` and `dst_scalars`'s docstring:

```
(Scalar features of source nodes.) See example usage case for details
```
so that the user knows where to look for the examples when then read the docstring

DejunL (changes requested):
Added a few comments about the docstring but otherwise LGTM

tingyu66:
@DejunL Feel free to take another look.

DejunL:
LGTM. Thank you for the quick update!

DejunL (approved):

rlratzel (approved):

python/cugraph-equivariant/cugraph_equivariant/nn/tensor_product_conv.py
```diff
@@ -31,6 +31,11 @@
     ) from exc
 
 
+class Graph(NamedTuple):
+    edge_index: torch.Tensor
+    size: tuple[int, int]
```
mariogeiger, line 36: This is unused?

PR #4482, opened 2024-06-11
compute cosine similarity for vertex pairs

naimnv:
compute cosine similarity for vertex pairs


cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -81,84 +81,148 @@ rmm::device_uvector<weight_t> similarity(
...
-        weight_t sum_of_intersected_b            = weight_t{0};
-
-        auto pair_first = thrust::make_zip_iterator(intersected_properties_a.data(),
-                                                    intersected_properties_b.data());
-        thrust::tie(sum_of_min_weight_a_intersect_b,
-                    sum_of_max_weight_a_intersect_b,
-                    sum_of_intersected_a,
-                    sum_of_intersected_b) =
-          thrust::transform_reduce(
+    if constexpr (functor_t::is_order_two) {
```
seunghwak, line 84: What is order_two here?

if is_order_two == true, we do this
```
return thrust::make_tuple(prop_a * prop_a, prop_b * prop_b, prop_a * prop_b);
```

and if is_order_two == false, we do the following?

```
return thrust::make_tuple(min(prop_a, prop_b), max(prop_a, prop_b), prop_a, prop_b);
```

cpp/src/link_prediction/overlap_impl.cuh
```diff
@@ -24,8 +24,9 @@
 namespace cugraph {
 namespace detail {
 
+template <typename weight_t, bool order_two>
 struct overlap_functor_t {
-  template <typename weight_t>
+  static constexpr bool is_order_two = order_two;
```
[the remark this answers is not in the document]
naimnv: Do you mean as a runtime variable?

cpp/src/link_prediction/similarity_impl.cuh
naimnv: cosine similarity requires second order terms. I couldn't come up with a better name.

cpp/src/link_prediction/overlap_impl.cuh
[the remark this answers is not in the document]
seunghwak: It can either be template parameter or a runtime variable... depends on the runtime overhead.

May better be template parameter there as this doesn't affect code size/compile time (order_two is fixed for each similarity type, right?) and this variable is queried inside the innermost loop.

cpp/src/link_prediction/similarity_impl.cuh
seunghwak: Naming is fine, but we should better avoid code replication if the difference is only in `prop_a * prop_a, prop_b * prop_b` vs `min(prop_a, prop_b), max(prop_a, prop_b)`. Code replication makes it tougher to figure out what order_two means.

cpp/src/link_prediction/overlap_impl.cuh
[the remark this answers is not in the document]
seunghwak: And we know order_two value when we call detail::similarity and this parameter is used in detail::similarity. Better directly pass in detail::similarity than indirectly passing through a functor (and the functor does not use this parameter internally...),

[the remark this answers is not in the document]
naimnv: Changed to pass it as runtime param.

cpp/src/link_prediction/similarity_impl.cuh
naimnv: Changed.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -36,13 +36,16 @@
 namespace cugraph {
 namespace detail {
 
+enum Coefficient { JACCARD, SORENSEN, OVERLAP, COSINE };
```
seunghwak, line 39: `enum class (similarity_)coefficient_t` to be consistent with the rest of the codebase.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -36,13 +36,16 @@
...
+
 template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu, typename functor_t>
 rmm::device_uvector<weight_t> similarity(
   raft::handle_t const& handle,
   graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
   std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
   std::tuple<raft::device_span<vertex_t const>, raft::device_span<vertex_t const>> vertex_pairs,
   functor_t functor,
+  Coefficient coff,
```
seunghwak, line 48: I google searched for "coefficient abbreviation" and got coeff or coef.

coff returned "common object file format".

coeff might be a better abbreviation.

seunghwak (approved):
LGTM except for few minor cosmetic issues.

cpp/src/link_prediction/similarity_impl.cuh
naimnv: Changed

cpp/src/link_prediction/similarity_impl.cuh
naimnv: Changed.

cpp/src/link_prediction/similarity_impl.cuh
```diff
@@ -36,13 +36,16 @@
 namespace cugraph {
 namespace detail {
 
+enum coefficient_t { JACCARD, SORENSEN, OVERLAP, COSINE };
```
seunghwak, line 39: enum **class**

https://stackoverflow.com/questions/18335861/why-is-enum-class-considered-safer-to-use-than-plain-enum

naimnv: Changed.

naimnv:
@BradReesWork I'm waiting for the CI being unblocked before triggering the CI.

PR #4484, opened 2024-06-12
Cut peak memory footprint in per_v_transform_reduce_dst_key_aggregated_outgoing_e

seunghwak:
Cut peak memory usage.

For an ultimate solution, we need to implement our own function to emulate ncclReduce with a custom reduction operator.


seunghwak:
@jnke2016 After confirming that the input size of sort_by_key grows as we increase # GPUs (and graph size grows proportional to # GPUs), you can try this. With this approach, we might be able to avoid OOM for 2-4x more GPUs. But we may encounter OOM again later especially for graphs with small E / V. To solve, this we need a multi-GPU reduce function that supports a custom reduce function (@naimnv also needs this to improve the performance of matching).

jnke2016:
Can you also fix the out of memory allocation [bug](https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/src/community/louvain_impl.cuh#L295) in Louvain in this PR please?

jnke2016 (approved):

PR #4489, opened 2024-06-14
Forward merge branch-24.06 into branch-24.08

nv-rliu:
Replaces https://github.com/rapidsai/cugraph/pull/4476


alexbarghi-nv (approved):
👍

alexbarghi-nv:
@nv-rliu I think this should be renamed "Forward merge branch-*24.06* into branch-*24.08*

ci/build_wheel.sh
```diff
@@ -56,7 +56,19 @@ fi
 
 cd "${package_dir}"
 
-python -m pip wheel . -w dist -vvv --no-deps --disable-pip-version-check
+
+PARALLEL_LEVEL=$(python -c \
+  "from math import ceil; from multiprocessing import cpu_count; print(ceil(cpu_count()/2))")
```
ajschmidt8, line 61: we recently added a `PARALLEL_LEVEL` environment variable to the `rapids-configure-sccache` script below:

- https://github.com/rapidsai/gha-tools/blob/dcd134953586fbcbe9924131e07fe443088da0a5/tools/rapids-configure-sccache#L7

That script is `source`d earlier in this file.

Therefore you could simply use the value defined from that script instead of redefining it here.

tingyu66: Great. Just wanted to be aggressive here to rule out the impact of hyperthreading. Using all 64 logical cores to compile cugraph did fail at times on my workstation (w/ Threadripper 3975WX).

ci/build_wheel.sh
```diff
@@ -56,7 +56,19 @@ fi
...
+
+SKBUILD_BUILD_OPTIONS="-j${PARALLEL_LEVEL}" \
+python -m pip wheel \
+    -w dist \
+    -vvv \
+    --no-deps \
+    --disable-pip-version-check \
+    --extra-index-url https://pypi.nvidia.com \
+    --global-option="-j${PARALLEL_LEVEL}" \
```
bdice, line 70: This `--global-option` is breaking the installation process. It issues this warning:

```
WARNING: Implying --no-binary=:all: due to the presence of --build-option / --global-option.
```

which in turn disallows us from using binary wheels of packages like pylibraft-cu12 and rmm-cu12.

```suggestion
```

ci/build_wheel.sh
bdice: I don't think hyperthreading is a factor on these CI machines -- we get the "real" number of cores, afaik. We should be safe to remove this. Let's wait to push until after CI runs on the current commit (I'm testing [something else](https://github.com/rapidsai/cugraph/pull/4489#discussion_r1643469633) at the moment).

ChuckHastings (approved):

bdice:
I'm going to go ahead and trigger a merge. Builds all succeeded.

bdice (approved):

PR #4499, opened 2024-06-24
Biased RW implementation and test

### LINKED ISSUE #2555: MNMG Random Walks - Implementation
Implement Random Walks.

Random walks implementation should support uniform random walks, biased random walks and node2vec random walks.

G-Cornett:
Biased random walk selector implemented.  Tests changed to accurately reflect expected behavior.

closes #2555


cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -483,7 +555,7 @@ biased_random_walks(raft::handle_t const& handle,
     std::optional<edge_property_view_t<edge_t, weight_t const*>>{edge_weight_view},
     start_vertices,
     max_length,
-    detail::biased_selector<weight_t>{(seed == 0 ? detail::get_current_time_nanoseconds() : seed)});
+    detail::biased_selector<weight_t>((seed == 0 ? detail::get_current_time_nanoseconds() : seed)));
```
seunghwak, line 558: Our API generally takes `rng_state` instead of seed (we have few functions taking seed instead, and we need to fix them as well).

See https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/include/cugraph/algorithms.hpp and search for rng_state. Random walk functions need to be updated to follow this convention.

e.g. Leiden takes `raft::random::RngState& rng_state` instead of seed.

And see https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/tests/prims/mg_per_v_random_select_transform_outgoing_e.cu#L173

This ensures that every GPU will use different seeds (so use different sequences of random numbers).

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -70,6 +72,18 @@ struct sample_edges_op_t {
...
+struct e_bias_op_t {
+
+  raft::device_span<bias_t const> vertex_weight_sum{};
+
+  __device__ bias_t
+  operator()(vertex_t src, vertex_t, thrust::nullopt_t, thrust::nullopt_t, bias_t weight) const
+  {
+    return weight / vertex_weight_sum[src];
+  }
+};
```
seunghwak, line 85: You may have another e_bias_op_t in the future for node2vec random walk. You may name this considering this future updates (e.g. biased_random_walk_e_bias_op_t).

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -139,7 +153,9 @@ struct uniform_selector {
 
 template <typename weight_t>
 struct biased_selector {
-  uint64_t seed_{0};
+  raft::random::RngState rng_state_;
+  
+  biased_selector(uint64_t seed) : rng_state_(seed) {}
```
seunghwak, line 158: Better delete this. We won't take a seed anymore.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -70,6 +72,18 @@ struct sample_edges_op_t {
...
 };
 
+template <typename vertex_t, typename bias_t>
+struct e_bias_op_t {
+
+  raft::device_span<bias_t const> vertex_weight_sum{};
+
+  __device__ bias_t
+  operator()(vertex_t src, vertex_t, thrust::nullopt_t, thrust::nullopt_t, bias_t weight) const
+  {
```
seunghwak, line 82: `vertex_weight_sum[src]` won't work in multi-GPU. Each GPU holds roughly V/P (V: # vertices, P: # GPUs) weight values, not the entire V.

You need to use `edge_src_property_t`. 


See https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/src/link_analysis/pagerank_impl.cuh#L233

https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/src/link_analysis/pagerank_impl.cuh#L264

and


https://github.com/rapidsai/cugraph/blob/branch-24.08/cpp/src/link_analysis/pagerank_impl.cuh#L279

seunghwak (commented):
Congratulations on your first PR!!!

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1583,7 +1583,7 @@ uniform_random_walks(raft::handle_t const& handle,
                      std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
                      raft::device_span<vertex_t const> start_vertices,
                      size_t max_length,
-                     uint64_t seed = std::numeric_limits<uint64_t>::max());
+                     raft::random::RngState& rng_state);
```
seunghwak, line 1586: Not a request for a change, mainly a comment for @ChuckHastings 

`rng_state` is mostly a parameter before `graph_view` in `algorithms.hpp` (exceptions, `maximal_independent_set` and `vertex_coloring`). In `sampling.hpp`, `rng_state` appears close to the end of the input parameter lists for uniform and biased sampling functions. We should better be consistent on this, but not sure placing this right before `graph_view` (or right after `handle`) make sense or not. This may make sense if we consier `rng_state` (sort of a common resource) as an extension of `handle`.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -139,7 +185,7 @@ struct uniform_selector {
 
 template <typename weight_t>
 struct biased_selector {
-  uint64_t seed_{0};
+  raft::random::RngState rng_state_;
```
seunghwak, line 188: This will copy-construct a new `RngState` object. You need to store a reference (or a pointer) to the passed `rng_state` object.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -70,12 +74,54 @@ struct sample_edges_op_t {
...
+struct biased_sample_edges_op_t {
+  __device__ thrust::tuple<vertex_t, weight_t>
+  operator()(vertex_t, vertex_t dst, weight_t, thrust::nullopt_t, weight_t weight) const
+  {
+    return thrust::make_tuple(dst, weight);
+  }
+};
+
+template <typename vertex_t, typename weight_t>
+struct node2vec_sample_edges_op_t {
```
seunghwak, line 97: Do you plan to implement both biased & node2vec random walks in a single PR?

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -156,15 +202,73 @@ struct biased_selector {
...
+          biased_sample_edges_op_t<vertex_t, weight_t>{},
+          rng_state_,
+          size_t{1},
+          true,
+          std::make_optional(
+            thrust::make_tuple(vertex_t{cugraph::invalid_vertex_id<vertex_t>::value}, weight_t{0.0})));
+      minors = std::move(std::get<0>(sample_e_op_results));
+      weights = std::move(std::get<1>(sample_e_op_results));
+    } else {
+      //  Just uniform random walk
```
seunghwak, line 242: We have an existing uniform random walk function, right?

I think we'd better make `edge_weight_view` non-optional. If the caller doesn't have a valid `edge_weight_view` object, the caller should invoke the uniform random walk function rather than calling this function with std::nullopt. No need to implement uniform random walk in two different places.

cpp/src/sampling/random_walks_impl.cuh
G-Cornett: Yes, I was instructed to just make them a single PR

cpp/include/cugraph/algorithms.hpp
ChuckHastings: I agree that consistency is better.

I'm inclined to prefer right after the handle.  It does seem like the RNG state is a common resource/extension of `handle` more than a regular parameter.  I think we have an optional case as well (Louvain/Leiden).  If we try and put it anywhere else in the argument list it seems like it could be pretty variable (e.g. before or after properties, etc).

cpp/src/sampling/random_walks_impl.cuh
ChuckHastings: We could split them up if that's easier.  Smaller pull requests are easier to review... but if the code changes are pretty small it wouldn't be too bad to have them together.

G-Cornett:
@seunghwak @ChuckHastings 
As requested, here is the sanitized node2vec implementation and generic refactoring.  Node2Vec, as currently written, passes the karate dataset test but fails larger tests

cpp/include/cugraph/algorithms.hpp
ChuckHastings: Let's move the rng_state parameter (per the discussion above) so that it becomes the second parameter (between the handle and the graph view).  That will necessitate updating the implementation and any tests or other calls to this function.

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1627,7 +1627,7 @@ biased_random_walks(raft::handle_t const& handle,
                     edge_property_view_t<edge_t, weight_t const*> edge_weight_view,
                     raft::device_span<vertex_t const> start_vertices,
                     size_t max_length,
-                    uint64_t seed = std::numeric_limits<uint64_t>::max());
+                    raft::random::RngState& rng_state);
```
ChuckHastings, line 1630: Same here

cpp/include/cugraph/algorithms.hpp
```diff
@@ -1676,7 +1676,7 @@ node2vec_random_walks(raft::handle_t const& handle,
                       size_t max_length,
                       weight_t p,
                       weight_t q,
-                      uint64_t seed = std::numeric_limits<uint64_t>::max());
+                      raft::random::RngState& rng_state);
```
ChuckHastings, line 1679: Same here

cpp/src/c_api/random_walks.cpp
```diff
@@ -156,7 +157,7 @@ struct uniform_random_walks_functor : public cugraph::c_api::abstract_functor {
   cugraph::c_api::cugraph_graph_t* graph_{nullptr};
   cugraph::c_api::cugraph_type_erased_device_array_view_t const* start_vertices_{nullptr};
   size_t max_length_{0};
-  size_t seed_{0};
+  cugraph::c_api::cugraph_rng_state_t* rng_state_{nullptr};
```
ChuckHastings, line 160: Can you add a FIXME comment here to indicate that the rng state should be passed in as a parameter.  We can make that a separate PR, since it will affect a bunch more code (including python and PLC calls).

cpp/src/c_api/random_walks.cpp
```diff
@@ -228,7 +229,7 @@ struct uniform_random_walks_functor : public cugraph::c_api::abstract_functor {
         (edge_weights != nullptr) ? std::make_optional(edge_weights->view()) : std::nullopt,
         raft::device_span<vertex_t const>{start_vertices.data(), start_vertices.size()},
         max_length_,
-        seed_);
+        rng_state_->rng_state_);
```
ChuckHastings, line 232: This will fail in the C API.  You set `rng_state_` above to nullptr, so you're dereferencing a nullptr.

Until we add the rng_state as a parameter, I would suggest that you create a `raft::random::RngState{0}` either inline here or as a local variable above this call and pass it here.

cpp/src/c_api/random_walks.cpp
```diff
@@ -332,7 +333,7 @@ struct biased_random_walks_functor : public cugraph::c_api::abstract_functor {
         edge_weights->view(),
         raft::device_span<vertex_t const>{start_vertices.data(), start_vertices.size()},
         max_length_,
-        seed_);
+        rng_state_->rng_state_);
```
ChuckHastings, line 336: Same

cpp/src/c_api/random_walks.cpp
```diff
@@ -439,7 +440,7 @@ struct node2vec_random_walks_functor : public cugraph::c_api::abstract_functor {
         max_length_,
         static_cast<weight_t>(p_),
         static_cast<weight_t>(q_),
-        seed_);
+        rng_state_->rng_state_);
```
ChuckHastings, line 443: Same

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -174,16 +352,203 @@ struct node2vec_selector {
...
+      detail::nbr_intersection(handle,
+                               graph_view,
+                               cugraph::edge_dummy_property_t{}.view(),
+                               intersection_pairs,
+                               intersection_pairs + current_vertices.size(),
+                               std::array<bool, 2>{true, true},
+                               false);
+
+    if constexpr (GraphViewType::is_multi_gpu) {
+      //  Create COO data structures
```
ChuckHastings, line 390: @seunghwak - are there functions in `cugraph::detail` under the structure utilities that would be helpful here?

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -221,6 +587,13 @@ random_walk_impl(raft::handle_t const& handle,
                        ? std::make_optional<rmm::device_uvector<weight_t>>(0, handle.get_stream())
                        : std::nullopt;
 
+  auto previous_vertices = (walk_type == random_walk_t::NODE2VEC)
```
ChuckHastings, line 590: It feels like we could use `std::is_same` on `random_selector_t` to determine this and not need to add the `random_walk_t` type at all.  Or if we want more flexibility, you could either:
1. Add a `constexpr` to the `random_selector_t` classes that is true for node2vec and false otherwise that you could use here, or
2. Add a function in `random_selector_t` that returns `std::nullopt`, except in node2vec where it allocates the optional device uvector

Seems like adding an enumerated type just for this one if statement is overkill.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -255,147 +628,291 @@ random_walk_impl(raft::handle_t const& handle,
...
+      random_selector.follow_random_edge(handle, graph_view, edge_weight_view, current_vertices, previous_vertices);
 
     // FIXME: remove_if has a 32-bit overflow issue
     // (https://github.com/NVIDIA/thrust/issues/1302) Seems unlikely here (the goal of
     // sampling is to extract small graphs) so not going to work around this for now.
     CUGRAPH_EXPECTS(
       current_vertices.size() < static_cast<size_t>(std::numeric_limits<int32_t>::max()),
       "remove_if will fail, current_vertices.size() is too large");
-
+    size_t compacted_length{0};
```
ChuckHastings, line 703: Scope of this seems off.  Make it as small as possible, perhaps inside the constexpr(multi_gpu) right below this.

ChuckHastings (commented):
Great job!

A number of things to improve the quality.  A question for @seunghwak to perhaps improve code reuse.

cpp/src/sampling/random_walks_impl.cuh
G-Cornett: Compacted length is needed regardless of single or multi-GPU, it is used to remove invalid vertex IDs from the results span and resize the number of walks for future steps (starting Line 834).  The scope was expanded to cut down on code duplication, as the original code had the resize calls in each if-else branch block.

ChuckHastings: OK.  I see that both paths through this if set it.

seunghwak (commented):
Review Part 1

cpp/src/sampling/random_walks_impl.cuh
seunghwak: Let me think more but I think

1) running adjacent_difference on the offsets array first,

2) call device_allgatherv using minor comm on tagged current vertices, counts (you get this by applying adjacent_difference on offsets) and indices

3) and converting counts to offsets

will work. Let me read this code more thoroughly.

seunghwak: Yeah... so the entire code in the `if constexpr (GraphViewType::is_multi_gpu)` path is to construct a (tagged-current vertices, offsets, indices) array triplet for all the edges local to this GPU.

I think we should eventually hide this in our primitives, but currently you need to understand our 2D partitioning to see you actually don't need to collect array triplets from the entire set of GPUs.

You need to collect intersection indices for (current vertex, previous vertex) pairs only for the current vertices that can appear as edge sources.

For this purpose, you need to gather array triplets only from the GPUs in the same minor_comm.

This could be pretty confusing without proper understanding of our 2D partitioning scheme. You may speak with me or Joseph if you want to understand this more.

Anyways,

If you run thrust::adjacent_difference on `intersection_offsets`, you get `intersection_counts`.

Now you have (current vertex, previous vertex) pairs, intersection counts, and intersection indices.

You can run allgatherv on minor_comm.

Then you get aggregate (current vertex, previous vertex) pairs, aggregate intersection counts, and aggregate intersection indices.

And luckily, if (current_vertex, previous vertex) pairs are sorted in each GPU, aggregate (current vertex, previous vertex) pairs are also sorted.

Now you run thrust::exclusive|inclusive_scan on the aggregate intersection counts, you now have intersection offsets.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -174,16 +352,203 @@ struct node2vec_selector {
...
+        true,
+        std::make_optional(vertex_t{cugraph::invalid_vertex_id<vertex_t>::value}));
+      minors = std::move(sample_e_op_results);
+    }
+
+    //  Copy current vertices to previous vertices for two-order walk
+    thrust::copy(handle.get_thrust_policy(), 
+                 current_vertices.begin(), 
+                 current_vertices.end(), 
+                 (*previous_vertices).data());
```
seunghwak, line 549: You may better perform this copy outside this function. This is really not part of a single step random walk but more for preparing for the next step. And if we perform this outside this function, previous_vertices can be const, so all the input parameters are const. This makes this function more functional (all the input parameters are non-modifiable, and you return updates only through return values).

seunghwak (commented):
Review Part 2.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -70,11 +77,141 @@ struct sample_edges_op_t {
...
+  __device__ std::enable_if_t<!std::is_same_v<W, void>, thrust::tuple<vertex_t, W>> operator()(
+        thrust::tuple<vertex_t, vertex_t> tagged_src,
+        vertex_t dst,
+        thrust::nullopt_t,
+        thrust::nullopt_t,
+        W w) const
+  {
+    return thrust::make_tuple(dst, w);
+  }
+};
```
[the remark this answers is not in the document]
G-Cornett: See above

cpp/src/sampling/random_walks_impl.cuh
G-Cornett: The function call on Line 682 uses the return to overwrite `current_vertices`, this is the latest point to make the copy before the data is lost.  In order to perform the copy outside of the function, I would need to have a separate variable to temporarily store the results so that the copy could be performed, then copy the data from the temp variable to 'current_vertices'.  Would that suffice?

seunghwak: What about taking (current_vertices, previous vertices) as R-values and return new (current_vertices, previous vertices)?
```
std::tuple<
rmm::device_uvector<vertex_t> /* new current vertices */,
rmm::device_uvector<vertex_t> /* new previous vertices */,
... >
follow_random_edge(...,
rmm::device_uvector<vertex_t>&& /* current_vertices */,
rmm::device_uvector<vertex_t>&& /* previous vertices */,
...) {
...
previous_vertices = std::move(current_vertices);
return std::make_tuple(std::move(minors), std::move(previous_vertices), ...);
}
```

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -70,21 +68,132 @@ struct sample_edges_op_t {
...
+  }
+
+  //  Weighted Bias Operator
+  template <typename W = weight_t>
+  __device__ std::enable_if_t<!std::is_same_v<W, void>, bias_t> operator()(
+        thrust::tuple<vertex_t, vertex_t> tagged_src,
+        vertex_t dst,
+        thrust::nullopt_t,
+        thrust::nullopt_t,
+        W w) const
```
seunghwak, line 134: Where are we using `w`? If this is not used, `W w`=> `W` will be sufficient (and also make it clear that `w` is not used).

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -84,7 +221,8 @@ struct uniform_selector {
     GraphViewType const& graph_view,
     std::optional<edge_property_view_t<typename GraphViewType::edge_type, weight_t const*>>
       edge_weight_view,
-    rmm::device_uvector<typename GraphViewType::vertex_type> const& current_vertices)
+    rmm::device_uvector<typename GraphViewType::vertex_type> const& current_vertices,
+    std::optional<rmm::device_uvector<typename GraphViewType::vertex_type>> &previous_vertices)
```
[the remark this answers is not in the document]
seunghwak: Any reason to use `rmm::device_uvector const&` over `raft::device_span`?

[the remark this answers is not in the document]
seunghwak: And for `previous_vertices`, use `std::optional<rmm::device_uvector<typename GraphViewType::vertex_type>> &&` instead of `std::optional<rmm::device_uvector<typename GraphViewType::vertex_type>> &`.  (&& instead of just &).

See https://stackoverflow.com/questions/37935393/pass-by-value-vs-pass-by-rvalue-reference

You're passing the ownership of `previous_vertices` to `folow_random_edge` so `follow_random_edge` can pass it pack to the caller.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -149,22 +287,62 @@ struct biased_selector {
     GraphViewType const& graph_view,
     std::optional<edge_property_view_t<typename GraphViewType::edge_type, weight_t const*>>
       edge_weight_view,
-    rmm::device_uvector<typename GraphViewType::vertex_type> const& current_vertices)
+    rmm::device_uvector<typename GraphViewType::vertex_type> const& current_vertices,
+    std::optional<rmm::device_uvector<typename GraphViewType::vertex_type>>& previous_vertices)
```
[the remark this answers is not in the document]
seunghwak: Same here.

cpp/src/sampling/random_walks_impl.cuh
```diff
@@ -133,57 +242,284 @@ struct uniform_selector {
...
+
+    using tag_t = void;
+
+    cugraph::vertex_frontier_t<vertex_t, tag_t, GraphViewType::is_multi_gpu, false> vertex_frontier(handle, 1);
+
+    vertex_frontier.bucket(0).insert(current_vertices.begin(), current_vertices.end());
+
+    auto vertex_weight_sum  = compute_out_weight_sums(handle, graph_view, *edge_weight_view);
+    edge_src_property_t<GraphViewType, weight_t> edge_src_out_weight_sums(handle, graph_view);
+    update_edge_src_property(handle, graph_view, vertex_weight_sum.data(), edge_src_out_weight_sums.mutable_view());
```
seunghwak, line 283: You're using source out weight sums for only the vertices in the `vertex_frontier`, right?

https://github.com/rapidsai/cugraph/blob/branch-24.10/cpp/src/prims/update_edge_src_dst_property.cuh#L938

This updates src property values for a subset of vertices. If the aggregate vertex frontier size is much smaller than the number of vertices in the graph, using this function will save both communication/computation.

You may just add a FIXME statement if you don't have time to fix it.

seunghwak (approved):
I have few suggestions for minor fixes. Otherwise, looks good to me and thank you for your contribution!!!

cpp/src/sampling/random_walks_impl.cuh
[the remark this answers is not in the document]
G-Cornett: > Any reason to use rmm::device_uvector const& over raft::device_span?
For the uniform and biased selectors, there is not. But for the node2vec selector, on Line 518, the copy requires that current vertices be a `rmm::device_uvector` type.  Since each selector uses the same call, they all have the `rmm::device_uvector` type. If there is a way to avoid this, I am unaware of it but would love to learn.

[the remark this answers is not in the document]
ChuckHastings: Interestingly, the uniform and biased selectors pass `rmm::device_uvector<vertex_t> const&`, where the node2vec selector passes `rmm::device_uvector<vertex_t>&` and has a side effect of doing `std::move` on it and invalidating the contents.  So it's not really the same.

I'd suggest marking this with a FIXME (in all 3 places).  My inclination would be to pass a span and do something different in the node2vec implementation (perhaps populate a new `rmm::device_uvector` or perhaps moving `rmm::device_uvector` in the function... we need to look at the context of what's going on to decide).

ChuckHastings (approved):
Great job @G-Cornett.  Thanks for your work this summer!

PR #4502, opened 2024-06-24
Expose new all-pairs Similarity algorithms

### LINKED ISSUE #4470: Expose new all-pairs Similarity algorithms to PLC
We implemented a C++ capability to initiate an all-pairs similarity algorithms directly without first creating a list of 2-hop neighbors.  This feature also adds a `topk` parameter which allows for executing the all-pairs computation in batches and only keeping the top `k` results.  This would enable execution of all-pairs similarity algorithms on larger graphs (when the `topk` parameter is specified) than would otherwise be possible.

Expose this feature to PLC.

jnke2016:
A variation of the Similarity algorithms leveraging of all-pairs between vertices is available in the C and C++ API. It also enable the retrieval of the top k vertices.

This PR:
- Exposes the new all-pairs Similarity algorithm to the PLC and python API
- Add SG and MG python tests
- Add docstrings with examples

closes #4470


seunghwak (approved):
Approving the C part.

python/cugraph/cugraph/dask/link_prediction/jaccard.py
```diff
@@ -140,21 +161,148 @@ def jaccard(input_graph, vertex_pair=None, use_weight=False):
...
+
+
+def all_pairs_jaccard(
+    input_graph,
+    vertices: cudf.Series = None,
+    use_weight: bool = False,
+    topk: int = None,
+):
+    """
+    Compute the All Pairs Jaccard similarity between all pairs of vertices specified.
```
rlratzel, line 205: The description of the algo is identical to `all_pairs_cosine`, but the word "cosine" is replaced with "jaccard".  Shouldn't there be a different description?  Otherwise, it's not clear when I should use one over the other.

rlratzel (changes requested):
Thanks, Joseph!  I didn't see any issues with the implementations but I have some comments about the docstrings.

Also, moving forward, we should probably have examples in the docstrings for each one.  It would also be helpful if the examples showed at least some of the results so users can see what to expect.

python/cugraph/cugraph/dask/link_prediction/cosine.py
```diff
@@ -0,0 +1,325 @@
...
+    Parameters
+    ----------
+    input_graph : cugraph.Graph
+        cuGraph Graph instance, should contain the connectivity information
+        as an edge list (edge weights are not supported yet for this algorithm). The
+        graph should be undirected where an undirected edge is represented by a
+        directed edge in both direction. The adjacency list will be computed if
+        not already present.
+
+        This implementation only supports undirected, unweighted Graph.
```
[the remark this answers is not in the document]
jnke2016: Good catch. Weight are not supported. If you recall the initial refactored similarity algorithm didn't support weights. I will update the docstrings for all of them

python/cugraph/cugraph/dask/link_prediction/jaccard.py
jnke2016: I provided further description in the documentation

rlratzel (approved):

PR #4503, opened 2024-06-25
Tweak rmm configuration for C++ unit tests

ChuckHastings:
We are seeing intermittent failures in CI from having trouble allocating the RMM pool allocator.

Dropping the memory usage by default from 1/6 to 1/10.

Added an option `maxpool` that will use 1/2 of the available memory, since we use the unit tests in larger configurations to do scale testing of algorithms.


cpp/tests/utilities/base_fixture.hpp
```diff
@@ -68,14 +68,18 @@ inline auto make_cuda() { return std::make_shared<rmm::mr::cuda_memory_resource>
...
+// use_max set to true will use half of available GPU memory for RMM, otherwise
+// otherwise we'll use 1/10.
+inline auto make_pool(bool use_max = false)
 {
   // Reduce the default pool allocation to 1/6th of the GPU memory so that we can
   // run more than 2 tests in parallel at the same time. Changes to this value could
   // effect the maximum amount of parallel tests, and therefore `tests/CMakeLists.txt`
   // `_CUGRAPH_TEST_PERCENT` default value will need to be audited.
   auto const [free, total] = rmm::available_device_memory();
-  auto const min_alloc = rmm::align_down(std::min(free, total / 6), rmm::CUDA_ALLOCATION_ALIGNMENT);
```
seunghwak, line 78: We should update the comments above (or delete it).

ChuckHastings: Yes, I'll update.

seunghwak (approved):

jameslamb (approved):

PR #4509, opened 2024-06-27
Avoid --find-links in wheel jobs

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/69
Contributes to https://github.com/rapidsai/build-planning/issues/33

Proposes a stricter pattern for passing a `pylibcugraph` wheel from the `wheel-build-cpp` job that produced it into the `wheel-build-python` job wanting to use it (as a build dependency of `cugraph`). This change improves the likelihood that issues with the `pylibcugraph` wheels will be caught in CI.

## Notes for Reviewers

See https://github.com/rapidsai/rmm/pull/1586 and https://github.com/rapidsai/build-planning/issues/69#issuecomment-2174556118 for details on how I tested this approach.


msarahan (approved):
It's counterintuitive to me how this works, but I also think that it is a good idea. The counterintuitive part is how we're using constraints to specify the path to the file. It's not wrong, it just isn't a use for the constraints file that I had considered.

jameslamb:
> *It's counterintuitive to me how this works, but I also think that it is a good idea. The counterintuitive part is how we're using constraints to specify the path to the file. It's not wrong, it just isn't a use for the constraints file that I had considered.*

Thanks for the careful review @msarahan .

I agree, it is definitely counterintuitive at first (I tried several other ideas before landing on this one). I think because we're in this situation where we both want to use build isolation but also want very fine-grained control over some characteristics of the build environment. If we wanted much more control than this, it'd probably be time to just abandon build isolation (I mentioned that near the end of https://github.com/rapidsai/build-planning/issues/79, a potential follow-up to this).

PR #4518, opened 2024-07-02
[FIX] Skip Distributed Sampler Tests if PyTorch with CUDA is not Available

alexbarghi-nv:
Some nightly tests are failing because there is no CUDA-supporting version of PyTorch available (as expected, i.e. on CUDA 11.4).  Instead, the CPU version of PyTorch gets installed, and the test crashes when attempting to set the CUDA allocator.  This PR disables those tests when only a CPU version of PyTorch is available to prevent this from happening.


vyasr (approved):

vyasr:
Thanks!

rlratzel (approved):

alexbarghi-nv:
C++ tests failed :(

vyasr:
CC @ChuckHastings this looks like the same issue we were discussing last week and this morning. Is the error intermittent for now, i.e. will rerunning be an OK approach until the test's memory usage gets fixed?

ChuckHastings:
I would try rerunning, I believe this is non-deterministic.  We are discussing a fix, I'll try and link the PR here.

PR #4496, opened 2024-06-24
remove openmpi ceiling

jameslamb:
fixes #4474 

#4496 and related PRs introduced a ceiling on `openmpi`, a dependency that's only pulled in at test time, because `cugraph`'s builds were struggling to find it.

This proposes removing that pin, as the fixes in https://github.com/conda-forge/openmpi-feedstock/pull/159 should allow the package to again be found by e.g. `find_package(MPI)` in CMake scripts.

PR #4478, opened 2024-06-10
Doc cleanup for nx-cugraph: fixed typos, cleaned up various descriptions, renamed notebook to match naming convetion.

rlratzel:
closes #4466 

* Fixed typos
* Changed various descriptions to match existing terminology and hopefully clarify
* Renamed notebook to match file name conventions

PR #4521, opened 2024-07-05
remove thriftpy2 ceiling

jameslamb:
#4496 introduced a ceiling on `thriftpy2`. Context: https://github.com/rapidsai/cugraph/pull/4496#issuecomment-2192783353

The bug that ceiling was added to avoid was fixed in v0.5.2 of `thriftpy2`, which was just released (https://github.com/rapidsai/cugraph/pull/4478#issuecomment-2210762587).

This removes that, adding `!=` constraints to skip the 2 versions that `cugraph` was not compatible with.

### Why not a floor?

I'm proposing adding `!=` constraints to skip v0.5.0 and v0.5.1 to maximize `cugraph`'s compatibility with other environments... that'd allow it to be used in environments with `thriftpy2<0.5.0` and in environments with `thriftpy2>0.5.2`.

Let me know if you'd prefer the simplicity of a floor like `>=0.5.2` instead.


jameslamb:
This is blocked until #4522 is merged.

bdice (approved):

nv-rliu (approved):

rlratzel (approved):

PR #4522, opened 2024-07-05
Skip the benchmark ctests within CI

ChuckHastings:
We have been hitting OOM failures in CI due to some tests using larger amounts of memory.  Our C++ testing includes benchmark tests that allow us to test on larger data sets.

This PR will suppress execution of the benchmark tests within CI.  Disabling these tests shouldn't reduce code coverage in testing.  They are mainly used for developers to test larger scale graphs and see the impact on performance of different updates we make.  We plan on adding these to our nightly tests where we will probably only run the benchmark results, perhaps with larger data sets since we'll run serially on larger GPUs.

Also updated dependencies to skip cmake version 3.30.0

### DIFF
--- a/ci/test_cpp.sh
+++ b/ci/test_cpp.sh
@@ -43,6 +43,9 @@ popd
 
 export GTEST_OUTPUT=xml:${RAPIDS_TESTS_DIR}/
 
+# Skip benchmark tests inside of CI
+export GTEST_FILTER="-*benchmark*"
+
 # Run libcugraph gtests from libcugraph-tests package
 rapids-logger "Run gtests"
 ./ci/run_ctests.sh -j10 && EXITCODE=$? || EXITCODE=$?;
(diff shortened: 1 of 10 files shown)


jameslamb (approved):
This `!=3.30.0` constraint on CMake should be added in all the conda recipes using CMake as well.

e.g. https://github.com/rapidsai/cugraph/blob/b9001f934be1ba3b2d5a6eb479dc2bfb9adb3214/conda/recipes/cugraph/conda_build_config.yaml#L13-L14 and others in `conda/recipes/`

Approving so you don't need to wait on another review from me after addressing that.

ChuckHastings:
> This `!=3.30.0` constraint on CMake should be added in all the conda recipes using CMake as well.
> 
> e.g.
> 
> https://github.com/rapidsai/cugraph/blob/b9001f934be1ba3b2d5a6eb479dc2bfb9adb3214/conda/recipes/cugraph/conda_build_config.yaml#L13-L14
> 
> and others in `conda/recipes/`
> Approving so you don't need to wait on another review from me after addressing that.

So these files aren't managed by the dependency generator.  I have updated them.

naimnv (approved):

Issue #4510, opened 2024-06-28
Jaccard, Sorensen, and Overlap OOM in bench_algos.py


nv-rliu:
### Version

24.08

### Which installation method(s) does this occur on?

Source

### Describe the bug.

The three algorithm calls (Jaccard, Overlap, Sorenson) are encountering OOM errors when called inside `cugraph/benchmarks/cugraph/pytest-based/bench_algos.py`

### Minimum reproducible example

```shell
pytest --import-mode=append -v bench_algos.py
```

### Relevant log output

```shell
06/28/24-11:18:48.121397660_UTC>>>> NODE 0: ******** STARTING BENCHMARK FROM: ./bench_algos.py::bench_jaccard, using 8 GPUs
============================= test session starts ==============================
platform linux -- Python 3.10.14, pytest-8.2.2, pluggy-1.5.0 -- /opt/conda/bin/python3.10
cachedir: .pytest_cache
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rapids_pytest_benchmark: 0.0.15
rootdir: /root/cugraph/benchmarks
configfile: pytest.ini
plugins: cov-5.0.0, benchmark-4.0.0, rapids-pytest-benchmark-0.0.15
collecting ... collected 720 items / 719 deselected / 1 selected

bench_algos.py::bench_jaccard[ds:rmat_mg_20_16-mm:False-pa:True] [1719573540.937835] [rno1-m02-b07-dgx1-012:2540068:0]          parser.c:2036 UCX  WARN  unused environment variable: UCX_MEMTYPE_CACHE (maybe: UCX_MEMTYPE_CACHE?)
[1719573540.937835] [rno1-m02-b07-dgx1-012:2540068:0]          parser.c:2036 UCX  WARN  (set UCX_WARN_UNUSED_ENV_VARS=n to suppress this warning)
[1719573540.938259] [rno1-m02-b07-dgx1-012:2540068:0]            sock.c:470  UCX  ERROR bind(fd=143 addr=0.0.0.0:38102) failed: Address already in use
[1719573540.936350] [rno1-m02-b07-dgx1-012:2540076:0]          parser.c:2036 UCX  WARN  unused environment variable: UCX_MEMTYPE_CACHE (maybe: UCX_MEMTYPE_CACHE?)
[...]

PR #4524, opened 2024-07-08
Fix OOM Bug for Jaccard, Sorensen, and Overlap benchmarks

nv-rliu:
Closes https://github.com/rapidsai/cugraph/issues/4510

This PR fixes a bug in `benchmarks/cugraph/pytest-based/bench_algos.py` for Jaccard and Overlap (and adds Sorensen to the list of algos to be run).

The reason for the error was that the benchmarks were trying to compute two_hop_neighbors on all nodes which would cause the GPU to run out of memory.


benchmarks/cugraph/pytest-based/bench_algos.py
```diff
@@ -268,6 +268,25 @@ def is_graph_distributed(graph):
     return isinstance(graph.edgelist.edgelist_df, dask_cudf.DataFrame)
 
 
+def get_vertex_pairs(G, subset_factor=3):
+    """
+    Return a DF of two columns representing random vertex pairs
+    """
+    sample_num = G.number_of_nodes() // subset_factor
+    col_a = (
+        G.select_random_vertices(num_vertices=sample_num)
```
jnke2016, line 277: You can instead call [get_two_hop_neighbors](https://github.com/rapidsai/cugraph/blob/branch-24.08/python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py#L853) which returns `vertex_pairs` based on a given set of `vertices`. You can still randomly generate the set of vertices with `select_random_vertices` and pass it to `get_two_hop_neighbors`

jnke2016 (changes requested):
I left a suggestion regarding the vertex_pairs generation.

benchmarks/cugraph/pytest-based/bench_algos.py
nv-rliu: Ah, thank you for the tip. I'll check that out.

nv-rliu: @jnke2016 I understand that `get_two_hop_neighbors` takes in a subset of `node_ids` from the Graph and returns a sorted list of all two-hop neighbor pairs. I was wondering what the advantage/goal of using this instead of constructing a random `src`/`dst` pair like I am doing now? Is this something to do with how the algs work?

jnke2016: >  Is this something to do with how the algs work?

Not necessarily but we already have and efficient code generating pairs. I don't see the use of creating a new one.

jnke2016 (approved):
Looks good to me

PR #4525, opened 2024-07-08
Better handle cudf.pandas in `from_pandas_edgelist`

eriknw:
Optimistically use cupy, but fall back to numpy if necessary.

Also, bump lint versions.

CC @rlratzel

### DIFF
--- a/python/nx-cugraph/nx_cugraph/convert_matrix.py
+++ b/python/nx-cugraph/nx_cugraph/convert_matrix.py
@@ -35,11 +35,21 @@ def from_pandas_edgelist(
 ):
     """cudf.DataFrame inputs also supported; value columns with str is unsuppported."""
     graph_class, inplace = _create_using_class(create_using)
+    # Try to be optimal whether using pandas, cudf, or cudf.pandas
     src_array = df[source].to_numpy()
     dst_array = df[target].to_numpy()
+    try:
+        # Optimistically try to use cupy, but fall back to numpy if necessary
+        src_array = cp.asarray(src_array)
+        dst_array = cp.asarray(dst_array)
+        np_or_cp = cp
+    except ValueError:
+        src_array = np.asarray(src_array)
+        dst_array = np.asarray(dst_array)
+        np_or_cp = np
     # TODO: create renumbering helper function(s)
     # Renumber step 0: node keys
-    nodes = np.unique(np.concatenate([src_array, dst_array]))
+    nodes = np_or_cp.unique(np_or_cp.concatenate([src_array, dst_array]))
     N = nodes.size
     kwargs = {}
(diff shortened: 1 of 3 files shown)


python/nx-cugraph/nx_cugraph/convert_matrix.py
```diff
@@ -35,25 +35,35 @@ def from_pandas_edgelist(
 ):
     """cudf.DataFrame inputs also supported; value columns with str is unsuppported."""
     graph_class, inplace = _create_using_class(create_using)
+    # Try to be optimal whether using pandas, cudf, or cudf.pandas
     src_array = df[source].to_numpy()
     dst_array = df[target].to_numpy()
```
eriknw, line 40: `copy=False` is currently the default here, but should we add `copy=False` to be clear about potential data movement? I wonder whether pandas will eventually follow numpy 2 semantics and use `copy=None` to only copy if necessary and `copy=False` to raise if a copy is necessary.

rlratzel: > should we add copy=False to be clear about potential data movement?

That seems like a good idea to me, unless there's a reason you think why it might be better not to.

rlratzel (approved):
Thanks!

python/nx-cugraph/nx_cugraph/convert_matrix.py
eriknw: NumPy 2 now has these semantics: https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword

python/nx-cugraph/nx_cugraph/convert_matrix.py
```diff
@@ -35,25 +35,35 @@ def from_pandas_edgelist(
...
-        # We need to renumber indices--np.searchsorted to the rescue!
+        # We need to renumber indices--np_or_cp.searchsorted to the rescue!
         kwargs["id_to_key"] = nodes.tolist()
-        src_indices = cp.array(np.searchsorted(nodes, src_array), index_dtype)
-        dst_indices = cp.array(np.searchsorted(nodes, dst_array), index_dtype)
+        src_indices = cp.asarray(np_or_cp.searchsorted(nodes, src_array), index_dtype)
+        dst_indices = cp.asarray(np_or_cp.searchsorted(nodes, dst_array), index_dtype)
     else:
         src_indices = cp.array(src_array)
         dst_indices = cp.array(dst_array)
```
eriknw, line 69: Note that this may perform an extra/unnecessary copy if `cp.asarray` above performed a copy. Maybe we should figure out best practice for determining whether a copy was made or not and avoid do extra copies.

PR #4527, opened 2024-07-08
Performance optimize BFS (including direction optimizing BFS implementation, mainly for single-GPU)

seunghwak:
This PR includes multiple performance optimizations to improve BFS performance (including direction optimizing BFS for symmetric graphs).

* Implement direction optimizing BFS
* Add the transform_reduce_v_frontier_outgoing_e_by_src primitive (this resembles transform_reduce_v_frontier_outgoing_e_by_dst but performs reduction based on src)
* Update the fill_edge_src|dst_property to take the fill value as the last input parameter (to be consistent with thrust::fill or std::fill)
* Update the fill_edge_src|dst_property to take a vertex list to fill only a subset of vertex property values.
* Update the update_edge_src|dst_property primitive to take a mutable view (to be consistent with other primitives)
* Update the vertex_frontier bucket to take a raft::device_span (supports creation of non-owning bucket, this saves a copy operation in some cases)
* Fix compiler warnings (cuda 12.4)
* Few updates in code cosmetics.


ChuckHastings (approved):

jnke2016 (approved):
Looks good to me

PR #4528, opened 2024-07-09
Further optimize `from_pandas_edgelist` with cudf

eriknw:
This continues #4525 (and [this comment](https://github.com/rapidsai/cugraph/pull/4525#discussion_r1669847549)) to avoid copies and to be more optimal whether using pandas, cudf, or cudf.pandas. Notably, using `s.to_numpy` with cudf will return a _numpy_ array, but `cudf.pandas` may return a _cupy_ array (proxy).

Also, `s.to_numpy(copy=False)` ([from comment](https://github.com/rapidsai/cugraph/pull/4525#discussion_r1669020347)) is not used, b/c cudf's `to_numpy` raises if `copy=False`. We get the behavior we want by not specifying `copy=`.

I don't know if this is the best way to determine whether a copy occurred or not, but this seems like a useful pattern to establish, because we want to make ingest more efficient.

CC @rlratzel


python/nx-cugraph/nx_cugraph/convert_matrix.py
```diff
@@ -65,8 +77,14 @@ def from_pandas_edgelist(
         src_indices = cp.asarray(np_or_cp.searchsorted(nodes, src_array), index_dtype)
         dst_indices = cp.asarray(np_or_cp.searchsorted(nodes, dst_array), index_dtype)
     else:
-        src_indices = cp.array(src_array)
-        dst_indices = cp.array(dst_array)
+        if is_src_copied:
+            src_indices = src_array
+        else:
+            src_indices = cp.array(src_array)
```
rlratzel, line 83: Maybe I'm missing something obvious, but couldn't the same behavior be achieved by just doing the following?
```suggestion
        src_indices = cp.array(src_array, copy=False)
```

rlratzel:
Thanks @eriknw , I really like the attention to detail with the tests!  I just had one question (see review) and I'm wondering if the answer to it could significantly change the PR.

python/nx-cugraph/nx_cugraph/convert_matrix.py
eriknw: I think you meant:
```python
        src_indices = cp.array(src_array, copy=not is_src_copied)
```
which should work as expected.

I found if-else branches more clear to use here, and easy for us to do given that we already have the booleans around.

eriknw: And once again, if you didn't catch it from one of my previous comments in previous PRs, numpy 2 changed semantics:
https://numpy.org/doc/stable/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
so I think it's okay to be extra clear here.

rlratzel: Thanks for the explanation, but no, actually that snippet wasn't what I was thinking.  But after re-reading again I think there's some necessary side effects that I was missing, but I'd rather not assume I know what they are.  Let's chat offline and I'll update the comment afterwards.

rlratzel (commented):
Thanks for meeting offline and clarifying things! I updated the review with the explanation. LGTM overall, but I have just one request for a comment to add.

python/nx-cugraph/nx_cugraph/convert_matrix.py
eriknw: Thanks, code comments added. NetworkX doesn't share ownership with input objects, so I figure neither should we.

rlratzel (approved):
Thanks for the new comments.

PR #4531, opened 2024-07-11
nx-cugraph: add `relabel_nodes` and `convert_node_labels_to_integers`

eriknw:
This was pretty tricky in places! This begins with `Graph` and `DiGraph`, which are probably tricker, because of the way edge data get merged when nodes are combined (and I wouldn't be surprised if there are other ways to do this operation).

This is one of the most heavily used functions in networkx and also networkx dependents.

Up next: multigraphs


python/nx-cugraph/nx_cugraph/interface.py
```diff
@@ -91,6 +94,42 @@ def key(testpath):
...
+            key("test_gml.py:TestGraph.test_parse_gml"): no_object_dtype_for_edges,
+            key("test_gml.py:TestGraph.test_read_gml"): no_object_dtype_for_edges,
+            key("test_gml.py:TestGraph.test_data_types"): no_object_dtype_for_edges,
+            key(
+                "test_gml.py:TestPropertyLists.test_reading_graph_with_list_property"
+            ): no_object_dtype_for_edges,
+            key(
+                "test_relabel.py:"
+                "test_relabel_preserve_node_order_partial_mapping_with_copy_false"
+            ): "Node order is preserved when relabeling with partial mapping",
```
eriknw, line 132: I don't fully grok why this test is checking that the node order _changed_. It seems like we'd want to keep the node order the same (and we do). See:
- https://github.com/networkx/networkx/pull/5258
- https://github.com/networkx/networkx/blob/ede3025392138cd0602894592a80b595a0e248d2/networkx/tests/test_relabel.py#L341-L347

rlratzel (approved):
Thanks, glad to see the update to handle `copy=False` for NX graphs, and also glad to see the updates for NX 3.4 (we should also add test runs using NX from its main branch to our nightlies, cc @nv-rliu ).

It was a bit hard to review though since `relabel_nodes` is a dense ~180 lines long :)  It's also too bad there's so many new xfails, but I guess that's understandable.

Also, I'm very curious what the speedup over NX will be, especially if they pass in a NX graph.  Do you have any benchmark results to share?

python/nx-cugraph/nx_cugraph/algorithms/traversal/breadth_first_search.py
```diff
@@ -57,23 +57,41 @@ def _bfs(G, source, *, depth_limit=None, reverse=False):
...
-        raise NotImplementedError(
-            "sort_neighbors argument in generic_bfs_edges is not currently supported"
-        )
-    return bfs_edges(G, source, depth_limit=depth_limit)
-
-
-@generic_bfs_edges._can_run
-def _(G, source, neighbors=None, depth_limit=None, sort_neighbors=None):
-    return neighbors is None and sort_neighbors is None
+if nx.__version__[:3] <= "3.3":
```
bdice, line 60: String comparison isn’t appropriate here (breaks with versions like 3.30). Try comparing version specifiers like this:

```python
from packaging.version import parse

if parse(nx.__version__) <= parse("3.3"):
    …

bdice (approved):
Approving packaging changes. One suggestion for version comparison.

PR #4534, opened 2024-07-11
Build and test with CUDA 12.5.1

KyleFromNVIDIA:
This PR updates the latest CUDA build/test version 12.2.2 to 12.5.1.

### DIFF
--- a/.devcontainer/cuda12.2-conda/devcontainer.json
+++ b/.devcontainer/cuda12.2-conda/devcontainer.json
@@ -3,7 +3,7 @@
     "context": "${localWorkspaceFolder}/.devcontainer",
     "dockerfile": "${localWorkspaceFolder}/.devcontainer/Dockerfile",
     "args": {
-      "CUDA": "12.2",
+      "CUDA": "12.5",
       "PYTHON_PACKAGE_MANAGER": "conda",
       "BASE": "rapidsai/devcontainers:24.08-cpp-mambaforge-ubuntu22.04"
     }
@@ -11,7 +11,7 @@
   "runArgs": [
     "--rm",
     "--name",
-    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-24.08-cuda12.2-conda"
+    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-24.08-cuda12.5-conda"
   ],
   "hostRequirements": {"gpu": "optional"},
   "features": {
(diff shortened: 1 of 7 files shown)


docs/cugraph/source/installation/source_build.md
```diff
@@ -61,7 +61,7 @@ conda env update --name cugraph_dev --file $CUGRAPH_HOME/conda/environments/all_
 conda activate cugraph_dev
 
 # for CUDA 12.x
-conda env update --name cugraph_dev --file $CUGRAPH_HOME/conda/environments/all_cuda-122_arch-x86_64.yaml
+conda env update --name cugraph_dev --file $CUGRAPH_HOME/conda/environments/all_cuda-125_arch-x86_64.yaml
```
jameslamb, line 64: There's one more right above this that also needs to be fixed:

https://github.com/rapidsai/cugraph/blob/5cdef4ef7f0df564fd85088c86e0e961958c5ef5/docs/cugraph/source/installation/source_build.md#L46

Looks like `rapids-reviser` wouldn't have caught it because it was (incorrectly) using `-12.2` instead of `-122` (so pointing to a file that doesn't exist).

jameslamb (approved):
Looks good, just one more small docs thing to fix.

jakirkham:
Should we update these lines?

https://github.com/rapidsai/cugraph/blame/5cdef4ef7f0df564fd85088c86e0e961958c5ef5/notebooks/README.md#L70

https://github.com/rapidsai/cugraph/blame/5cdef4ef7f0df564fd85088c86e0e961958c5ef5/notebooks/README.md#L76

https://github.com/rapidsai/cugraph/blame/5cdef4ef7f0df564fd85088c86e0e961958c5ef5/notebooks/README.md#L92

jameslamb:
> should we update these lines from CUDA 12.0 to 12.5

We couldn't update them to 12.5 yet, as we aren't yet publishing 12.5 images from `rapidsai/docker` (https://github.com/rapidsai/build-planning/issues/73#issuecomment-2231443095).

Since it seems like the offline discussion on that issue is trending towards "let's add 12.5 images", and since `cugraph`'s CI takes so long to run, I think we should NOT update those lines from 12.0 to 12.5 in this PR, or block this PR until those new images are available.

I think we should merge this once CI passes, and go back over RAPIDS with a pass for these image tags once we're publishing 12.5 images from `rapidsai/docker`.

jakirkham:
Given we didn't update them before, not updating would be ok

> I think we should merge this once CI passes, and go back over RAPIDS with a pass for these image tags once we're publishing 12.5 images from rapidsai/docker.

SGTM

ChuckHastings (approved):

jakirkham:
Thanks all! 🙏

PR #4535, opened 2024-07-12
Define and Implement C API for biased sampling

ChuckHastings:
Defined and implements C API for biased sampling.

API includes a parameter for handling edge bias input that is not the weight of the edge.  However, there is currently no mechanism to create the `cugraph_edge_property_view_t *` value that would need to be passed in.  There are some proposed functions to support that included in `property.h`, but they are inside an `#ifdef 0` block so that they won't be compiled.


cpp/include/cugraph_c/properties.h
```diff
@@ -0,0 +1,146 @@
...
+ * @param [out] error            Pointer to an error object storing details of any error.  Will
+ *                               be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_edge_property_create(
+  const cugraph_resource_handle_t* handle,
+  const cugraph_graph_t * graph,
+  const cugraph_lookup_container_t* lookup_container,
+  const cugraph_type_erased_device_array_t* edge_ids,
+  const cugraph_type_erased_device_array_t* properties,
```
seunghwak, line 93: What happens if edge IDs are provided only for a subset of edges? Will those edges have undefined values? Should we ask for default values?

And no need for an update function? (e.g. update bias values only for a small subset of edges).

ChuckHastings: Reasonable ideas.  I'll look at defining update functions and a mechanism for a default value.  I'll add it to both vertex and edge functions.

seunghwak: Just fyi. In the C++ primitives level, we have `fill_edge_property_value` (which sets edge property values for the entire set of edges, we may add a function that works on a subset of the edges in the future) and `update_edge_property` function that works either on a subset of the edges or the entire set of edges. Following this may make implementing the API easier.

ChuckHastings: I pushed an update that adds these functions.  Plan is to implement them at some point in the future.

alexbarghi-nv (approved):
👍

seunghwak (approved):

PR #4538, opened 2024-07-15
nx-cugraph: add dijkstra sssp functions

eriknw:
This does add new functionality (so it's not just sugar!), since dijstra methods have `cutoff=` arguments.

Note that we are missing multi-source dijstra methods, since `plc.sssp` only accepts a single source.

We also do not have `bidirectional_dijkstra`, since PLC does not implement this algorithm.


rlratzel (approved):
LGTM

PR #4543, opened 2024-07-18
Support non p2p configuration when initializing the comms

### LINKED ISSUE #4490: [FEA]: Support Cugraph dask algorithms without creating ucx endpoints.

Improvement

Medium

Cugraph dask algorithms like `wcc` are used in downstream libraries like [NeMo-Curator](https://github.com/NVIDIA/NeMo-Curator/tree/main) and require intializing comms with something like `cugraph.dask.comms.comms.initialize(p2p=True)`.

Setting `p2p=False` is currently unsupported and `p2p=True` creates UCX endpoints to establish connections to workers even when working with a TCP based dask cluster. 

Many systems & clusters have varying network topologies and require different UCX specific configurations to be set up correctly for the setup to work as expected & therefor `p2p=True` fails in many cluster unless the right UCX configurations are set.

It would improve usability if there was an option to run mnmg algorithms on a TCP cluster without any UCX config or endpoint creation.

```python
cugraph.dask.comms.comms.initialize(p2p=False) # False works, no ucx created

# Proceed with mnmg algorithms.
```

_No response_

### Additional context

_No response_

### Code of Conduct

jnke2016:
closes #4490


ChuckHastings (approved):

python/cugraph/cugraph/tests/conftest.py
```diff
@@ -52,6 +52,21 @@ def dask_client():
     stop_dask_client(dask_client, dask_cluster)
 
 
+# FIXME: Add tests leveraging this fixture
+@pytest.fixture(scope="module")
+def dask_client_non_p2p():
```
rlratzel, line 57: Do we know if we'd like to run tests that don't create UCX endpoints by hardcoding a different fixture, or would it be better to pass a CLI option, or something else?  I think it's okay to add this fixture now if we're going to use it, but I wonder if we should think of an alternative if something like a CLI option to pytest would be better.

rlratzel (approved):

python/cugraph/cugraph/tests/conftest.py
ChuckHastings: I think we should eliminate the p2p feature entirely.  IIRC, the only reason we required the P2P to be configured was because at the time we initially implemented it was required to properly set up the NCCL communication that the C++ library uses.  NCCL has been improved so that the P2P feature is no longer required to make this configuration.  Joseph's changes here and his testing have validated that belief.

I'd be inclined to merge this PR and create a new issue to address dropping the P2P configuration option and convert all of the tests to configure with the non-P2P configuration.

PR #4548, opened 2024-07-22
[BUG] Use the Correct WG Communicator

alexbarghi-nv:
cuGraph-PyG's WholeFeatureStore currently uses the local communicator, when it should be using the global communicator, as was originally intended.  This PR modifies the feature store so it correctly calls `get_global_node_communicator()`.

This also fixes another bug where torch.int32 was used to store the number of edges in the graph, which resulted in an overflow error when the number of edges exceeded that datatype's maximum value.  The datatype is now correctly set to int64.


rlratzel (approved):
LGTM  (disclaimer: I'm not familiar with the behavior, side-effects, and intent, so feel free to also pull in others if necessary)

alexbarghi-nv:
> LGTM (disclaimer: I'm not familiar with the behavior, side-effects, and intent, so feel free to also pull in others if necessary)

The tl;dr is that the local communicator is used for intra-node communication and the global communicator is used for inter-node communication.  If there's only one node, they are the same communicator, basically.  But if there are multiple nodes, then the local communicator won't include all workers.  So we could potentially hang if we use the local communicator here.  There usually isn't any reason to use the local communicator in this context unless we're doing something that we know only involves the workers on the current node.

PR #4551, opened 2024-07-22
Replace conda-build with rattler-build

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/47.

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -15,6 +15,8 @@ rapids-print-env
 
 rapids-logger "Begin cpp build"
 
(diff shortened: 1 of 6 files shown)


vyasr:
There are additional changes that will be needed for CUDA 11 compatibility that I didn't bother with yet. I also haven't updated the Python recipes at all, but that should be much simpler once the C++ recipe is finalized.

vyasr:
@msarahan I've assigned this PR to you since you were planning to take this over. Please reassign it if you expect someone else to wind up being the one to take this over the finish line.

PR #4553, opened 2024-07-25
Remove ODR violations due to cuda kernels being public

robertmaynard:


robertmaynard:
Closed as #4609 replaces it

PR #4556, opened 2024-07-26
Add an Explanatory Error Message for uint Types

alexbarghi-nv:
Clarify that unsigned integer types are not supported as vertex ids in the error message.  Resolves issue reported by @ogreen.


python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py
```diff
@@ -255,6 +255,11 @@ def __from_edgelist(
             elif elist[source].dtype not in [np.int32, np.int64] or elist[
                 destination
             ].dtype not in [np.int32, np.int64]:
+                if elist[destination].dtype in [np.uint32, np.uint64]:
```
rlratzel, line 258: should this also check `elist[source].dtype` ?

alexbarghi-nv: fixed

alexbarghi-nv:
@rlratzel can you re-review?  All changes should be addressed now.

PR #4557, opened 2024-07-26
Use stream_allocator_adaptor constructor instead of factory.

bdice:
This PR uses the `stream_allocator_adaptor` constructor instead of a factory. With CTAD, we do not need the factory and can use the constructor directly. The factory will be deprecated in https://github.com/rapidsai/rmm/pull/1626.


jameslamb (dismissed):
It looks like most of the diff here (and `packaging-codeowners` being added as a reviewer) is coming from the fact that this has `branch-24.10` commits on it but is targeting `branch-24.08`. I suspect this was meant to target `branch-24.10`.

jameslamb (approved):
Ok yeah I figured, thanks. Approved from me (not that you need it any more).

seunghwak (approved):

PR #4561, opened 2024-07-29
Re-configure benchmarking devices & add markers to `bench_cugraph_uniform_neighbor_sample`

nv-rliu:
This PR re-enables the `benchmarks/cugraph/pytest-based/bench_cugraph_uniform_neighbor_sample.py` benchmark in the MNMG nightlies. 

This benchmark was previously being skipped due to the fact that the benchmarks are configured to run with Pytest markers `"managedmem_off and poolallocator_on"`. Thanks to @jameslamb for spotting this inside `conftest.py`

By adding the missing markers to the test (and removing some `dgx` machine specific dask-client configs), the benchmark should properly run in the nightly jobs.


nv-rliu:
This PR is targeting `24.10` and should not be reviewed until the forward-merge has been completed.

rlratzel (approved):

nv-rliu:
> I do think the `pytest` marker changes will help ensure the benchmark actually runs.
> 
> For my own understanding ... why is it desirable to remove the ability to configure which GPUs on the system are used? (this is just for my own curiosity... can answer this after you merge)

Hi James. Thanks for the question.
When we run these benchmarks on our clustered machines, we run then in multiple configurations (2-GPU, 8-GPU, 32, 64, etc) and ideally want to test the algorithm while using all of the available GPUs (allocated by the job scheduler). 
As for the old code that selected 4-GPUs (Devices 1-4), my guess is that this was for running this benchmark locally on the lab machines.

PR #4562, opened 2024-07-29
Fix ucx-py version, use UCX 1.17.0 in pip devcontainers

bdice:
Fixes ucx-py dependency in `dependencies.yaml` and `update-version.sh`.

Updates to UCX 1.17.0 in pip devcontainers. (context: https://github.com/rapidsai/cugraph/pull/4562#issuecomment-2258608174).

PR #4565, opened 2024-07-30
Merge branch-24.08 into branch-24.10

jameslamb:
Replaces #4546

Fixes forward-merge conflicts, per the process described at https://docs.rapids.ai/maintainers/forward-merger/.

Tested to be sure I got everything like this:

```shell
git grep -E '24\.8|24\.08|0\.39'
```


dependencies.yaml
```diff
@@ -666,7 +674,7 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - &pytorch_conda pytorch>=2.0,<2.2.0a0
+          - &pytorch_unsuffixed pytorch>=2.0,<2.2.0a0
```
bdice, line 677: Not sure if we discussed this in another PR (lost track) but we don’t need the suffix here.
```suggestion
          - &pytorch pytorch>=2.0,<2.2.0a0
```

bdice (approved):
I expect CI to be broken without #4562 so please request an admin merge after looking at my comment on pytorch.

dependencies.yaml
jameslamb: we definitely did! It must have been on a different PR (maybe `wholegraph`?). I agree, will fix this up in a follow-up PR.

PR #4572, opened 2024-07-31
Use RMM adaptor constructor rather than factory function

harrism:
RMM will be deprecating and removing its adaptor factory functions (`make_*_adaptor`). These were needed prior to C++17 due to the lack of CTAD, but since C++17 is our minimum supported version, we can drop them now. Users can simply call the constructor for the adaptors.

This PR replaces calls to `make_stream_allocator_adaptor()` with `stream_allocator_adaptor` ahead of the deprecation of these functions in RMM 24.10.


bdice:
Duplicate of #4557. I merged the change to add `<cstddef>` into my PR, so they should be identical now.

harrism:
My bad!

PR #4582, opened 2024-07-31
add setuptools to host requirements for conda packages that need it

jameslamb:
@caryr35 pointed out to me this morning that `cugraph`'s nightly CI is failing. All `conda-python-build` jobs are failing like this, when building `nx-cugraph`

> ModuleNotFoundError: No module named 'setuptools'
> ...
> ValueError: Could not import build backend specified in pyproject.toml's tool.rapids-build-backend table. Make sure you specified the right optional dependency in your build-system.requires entry for rapids-build-backend.

([build link](https://github.com/rapidsai/cugraph/actions/runs/10176436224/job/28166520972))

**suspected root cause:**  `nx-cugraph` uses `setuptools.build_meta`, but `setuptools` isn't present in the conda build environment

### Why is this targeting `branch-24.08`?

Looks like CI is failing there too: https://github.com/rapidsai/cugraph/actions/runs/10183681336/job/28171285190

### Why is this just breaking now?

I suspect that prior to this we were getting `setuptools` because it was a transitive dependency of one of `nx-cugraph`'s other build/host dependencies.

### How could we prevent stuff like this in the future?

We could add support for updating conda recipe files in `rapids-dependency-file-generator` (https://github.com/rapidsai/dependency-file-generator/issues/7).

`setuptools` was correctly added as a build dependency in `pyproject.toml` files here (automatically, via `dependencies.yaml` + `rapids-dependency-file-generator`), but the conda recipe `meta.yaml` files were missed.

### DIFF
--- a/conda/recipes/cugraph-dgl/meta.yaml
+++ b/conda/recipes/cugraph-dgl/meta.yaml
@@ -22,6 +22,7 @@ requirements:
   host:
     - python
     - rapids-build-backend>=0.3.1,<0.4.0.dev0
+    - setuptools>=61.0.0
   run:
     - cugraph ={{ version }}
     - dgl >=1.1.0.cu*
--- a/conda/recipes/cugraph-equivariant/meta.yaml
+++ b/conda/recipes/cugraph-equivariant/meta.yaml
@@ -22,6 +22,7 @@ requirements:
   host:
     - python
     - rapids-build-backend>=0.3.1,<0.4.0.dev0
+    - setuptools>=61.0.0
   run:
     - pylibcugraphops ={{ minor_version }}
     - python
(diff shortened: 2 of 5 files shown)


alexbarghi-nv:
It looks like cugraph-pyg has not been updated

jameslamb:
> It looks like cugraph-pyg has not been updated

Ah! Thanks, updated in https://github.com/rapidsai/cugraph/pull/4582/commits/6774dcd5e705e6479b66af748b5f7d5729bc8917. I'd missed that one because I saw `scikit-build-core` in its conda recipe. Looks like that was incorrect, it really does use `setuptools`:

https://github.com/rapidsai/cugraph/blob/b38e010bd81e8d53f953970e5603f5890efbc4b6/python/cugraph-pyg/pyproject.toml#L68

alexbarghi-nv (approved):
👍

bdice (approved):

rlratzel (approved):
Thanks for doing this, @jameslamb !

PR #4586, opened 2024-08-01
[FEA] Biased Sampling in cuGraph-PyG

alexbarghi-nv:
Enables biased sampling in PylibcuGraph and cuGraph-PyG.  Also fixes a memory issue that was revealed by calling biased sampling instead of uniform neighbor sampling; the loader was pre-allocating more memory than needed.

- [x] Update pylibcugraph
- [x] Update dist sampler in cugraph
- [x] Correctness testing
- [x] Make issue for PyG upstream ([here](https://github.com/pyg-team/pytorch_geometric/issues/9566))


python/pylibcugraph/pylibcugraph/biased_neighbor_sample.pyx
```diff
@@ -0,0 +1,448 @@
...
+        referencing data and running algorithms.
+
+    input_graph : SGGraph or MGGraph
+        The input graph, for either Single or Multi-GPU operations.
+
+    start_list: device array type
+        Device array containing the list of starting vertices for sampling.
+
+    h_fan_out: numpy array type
+        Device array containing the brancing out (fan-out) degrees per
```
rlratzel, line 121: Is this a device array?  It has an `h_` prefix and the assertion below looks for (non-CUDA) array_interface.

rlratzel (approved):
Just some minor things which need not hold up my approval.  I suspect @VibhuJawa might be a better reviewer.

python/pylibcugraph/pylibcugraph/biased_neighbor_sample.pyx
alexbarghi-nv: This is a host array.  I naively copied this from `uniform_neighbor_sample` so I'll fix the docstring there too.

python/cugraph-pyg/cugraph_pyg/data/graph_store.py
```diff
@@ -58,13 +58,19 @@ def __init__(self, is_multi_gpu: bool = False):
...
-        self.__vertex_offsets = None
+
         self.__handle = None
         self.__is_multi_gpu = is_multi_gpu
 
+        self.__invalidate_graph()
+
         super().__init__()
 
+    def __invalidate_graph(self):
```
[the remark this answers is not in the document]
alexbarghi-nv: It's a bad name, you're right.  I'll change it to "clear"

PR #4597, opened 2024-08-07
Use tool.scikit-build.cmake.version, set scikit-build-core minimum-version

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/58.

`scikit-build-core==0.10.0` was released today (https://github.com/scikit-build/scikit-build-core/releases/tag/v0.10.0), and wheel-building configurations across RAPIDS are incompatible with it.

This proposes upgrading to that version and fixing configuration here in a way that:

* is compatible with that new `scikit-build-core` version
* takes advantage of the forward-compatibility mechanism (`minimum-version`) that `scikit-build-core` provides, to reduce the risk of needing to do this again in the future


jakirkham:
Thanks James! 🙏

Appears to be marked draft/WIP. What else is still needed here?

jameslamb:
> Appears to be marked draft/WIP. What else is still needed here?

Most of the CI jobs failed, in a way that looked to me like what we've seen before when the runners run out of memory. I'm not sure if that was because of some infrastructure-level issues or because we just weren't getting many cache hits from `sccache`.

Didn't want to put this up for review until more things were passing.

I've just restarted CI, let's see how it goes.

msarahan (approved):
test failures look like CUDF API stuff. Definitely orthogonal to this change. I think this PR is fine.

jameslamb:
Thanks @msarahan !

And in fact, looks like @nv-rliu just merged a PR that should fix those: CI issues: #4600

I just merged latest `branch-24.10` in here to pull those fixes in.

jameslamb:
(just queuing that to auto-merge whenever the tests run)

PR #4599, opened 2024-08-07
Improve update-version.sh

bdice:
This PR improves `update-version.sh` by clarifying suffix handling and ucxx/ucx-py version handling.


jameslamb (approved):
Agree, this is an improvement. Thanks!

PR #4600, opened 2024-08-07
Fix `test_property_graph_mg` Usage of Util Function

nv-rliu:
This PR fixes an issue that recently arose due to a change in constructing a `cudf.Series` object.

`test_property_graph_mg.py` uses a function from `utilities/utils.py` to construct a Series object from 2d arrays. The function is now using the correct `cudf.Series._from_column` API.


rlratzel (approved):

rlratzel:
@nv-rliu mentioned that the change here uses a non-public API that was recommended by cuDF devs (cc @bdice )

PR #4602, opened 2024-08-08
Heterogeneous renumbering implementation

### LINKED ISSUE #4412: Support Heterogeneous Renumbering in the C++ API
We currently support heterogeneous renumbering in the C++ sampling API.  This issue is for extending that to heterogeneous graphs.  Fundamentally, the process is the same with a few differences:

Input:
* A heterogeneous graph can have multiple node types
  * BUT we can always assume they will be renumbered from 0 to N by cuGraph-PyG and cuGraph-DGL
  * We can also assume that each node type has a starting and ending offset (i.e. type 0 is nodes 0-30, type 1 is nodes 31-50, etc.) and that the user must provide this in sampling.
* A heterogeneous graph can have multiple edge types
  * AND each edge type has its own numbering from 0 to #edges_of_type (cuGraph-PyG/cuGraph-DGL don't renumber these)
  * So from the libcugraph perspective, there may be multiple edges with id 0, but there can only be one id 0 per edge type
* For each edge type, there is a single input node type and output node type
  * i.e. (author, writes, paper) has "author" as the input node type and "paper" as the output node type.
  * The user will be required to provide a mapping of edge type id to src/dst node type id.

Output:
* Every edge type and every node type has its own renumbering for each output minibatch
  * i.e. "paper" has node ids from 0 to #papers_in_minibatch - 1 and "author" has node ids from 0 to #authors_in_minibatch - 1
  * These node numberings are consistent across edge types.  We only renumber the nodes once, globally, or per hop in the DGL case.
  * This obviously means that each node type needs a renumber map.
* Each edge type corresponds to a separate subgraph in the output minibatch.
  * libcugraph needs to output a list of offsets for where each edge type graph starts and ends
  * If there is no output for a particular edge type, it should still be noted (i.e. the offsets [0,30,30,70] means that edge type 1 had no output edges in the current minibatch.
  * Each edge type needs a renumber map.  These renumber maps are created the same way they are for nodes.
  * In the case of a separate graph outputted per hop (DGL), every hop would still need a separate renumbering, just as it is now.
* We also use the user-provided node offsets array to subtract off the starting offset for each node type in each node's renumber map

Other Notes:
- We are probably only going to expose this through pylibcugraph for the time being.
- We should make sure the node offsets requirement is well-documented since it requires an extra renumbering step that cuGraph-DGL/cuGraph-PyG does (even though this is trivial and already implemented by both frameworks)
- For MG, the node offsets and edge mapping are global.  Each worker will get the exact same values for each array.  These should be respected globally when doing renumbering.  No concatenation should occur.  I suppose we could have an expensive check that verifies these arrays are identical across workers.

seunghwak:
This PR implements heterogeneous renumbering for GNN.

In addition,

* Update the existing (homogeneous) sampling post processing function test file extension from .cu to .cpp.
* Remove the unused `renumber_sampled_edgelist` function (breaking because this function is removed)
* Add a `stride_fill` utility function (thrust wrapper)
* Add test utility functions to generate edge types & IDs. 

Closes #4412


cpp/tests/CMakeLists.txt
```diff
@@ -576,7 +582,8 @@ if(BUILD_CUGRAPH_MG_TESTS)
     ###############################################################################################
     # - MG BETWEENNESS CENTRALITY tests -----------------------------------------------------------
     ConfigureTestMG(MG_BETWEENNESS_CENTRALITY_TEST centrality/mg_betweenness_centrality_test.cpp)
-    ConfigureTestMG(MG_EDGE_BETWEENNESS_CENTRALITY_TEST centrality/mg_edge_betweenness_centrality_test.cpp)
+    ConfigureTestMG(MG_EDGE_BETWEENNESS_CENTRALITY_TEST
```
ChuckHastings, line 585: This seems like an odd edit.

cpp/tests/sampling/sampling_heterogeneous_post_processing_test.cpp
```diff
@@ -0,0 +1,828 @@
+/*
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Just 2024?

ChuckHastings (approved):
A few minor things.  Looks good.

cpp/tests/CMakeLists.txt
seunghwak: This exceeded the line length limit :-) Should we allow long lines in CMakeLists.txt?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -510,6 +514,11 @@ renumber_and_sort_sampled_edgelist(
  * true.
  * 2. Edges in each label are sorted independently if @p edgelist_label_offsets.has_value() is true.
  *
+ * This function assumes that there is a single edge source vertex type and a single edge
+ * destination vertex type for each edge. If @p edgelist_edge_types.has_value() is false (i.e. there
+ * is only one edge type), there shouldb be only one edge source vertex type and only one edge
```
alexbarghi-nv, line 519: typo here
```suggestion
 * is only one edge type), there should be only one edge source vertex type and only one edge
```

alexbarghi-nv (approved):
Looks good.  I assume you're delegating updating the C API sampling function to properly handle heterogeneous renumbering to me.

cpp/tests/CMakeLists.txt
ChuckHastings: I didn't realize we had one.

ChuckHastings:
> Looks good. I assume you're delegating updating the C API sampling function to properly handle heterogeneous renumbering to me.

Do you have time to do that @alexbarghi-nv ?  We can do that as a separate PR if you are busy with other things.  Would you want to always use this new heterogeneous renumbering logic, add an option for it, or detect it in some other way?

alexbarghi-nv:
> > Looks good. I assume you're delegating updating the C API sampling function to properly handle heterogeneous renumbering to me.
> 
> Do you have time to do that @alexbarghi-nv ? We can do that as a separate PR if you are busy with other things. Would you want to always use this new heterogeneous renumbering logic, add an option for it, or detect it in some other way?

We only want to use it for heterogeneous graphs.  I think we can just add an optional parameter for the edge type mappings, and if they are present, then the C API sampling function will call the heterogeneous renumbering with the mappings instead of the homogeneous renumbering.

alexbarghi-nv:
@seunghwak it looks like your PR changed the way we're handling empty start lists which is causing CI to fail:
```
FAILED tests/sampling/test_uniform_neighbor_sample.py::test_uniform_neighbor_sample_simple[graph_file:email_Eu_core-directed:True-with_replacement:False-indices_type:int32] - RuntimeError: non-success value returned from cugraph_uniform_neighbor_sample: CUGRAPH_UNKNOWN_ERROR cuGraph failure at file=/opt/conda/conda-bld/work/cpp/src/sampling/sampling_post_processing_impl.cuh line=235: Invalid input arguments: num_labels should be 0 if the input edge list is empty.
```

cpp/tests/CMakeLists.txt
seunghwak: I guess we don't have one officially, but I implicitly consider the end of `###############################################################################################` as the line limit.

seunghwak:
> @seunghwak it looks like your PR changed the way we're handling empty start lists which is causing CI to fail:
> 
> ```
> FAILED tests/sampling/test_uniform_neighbor_sample.py::test_uniform_neighbor_sample_simple[graph_file:email_Eu_core-directed:True-with_replacement:False-indices_type:int32] - RuntimeError: non-success value returned from cugraph_uniform_neighbor_sample: CUGRAPH_UNKNOWN_ERROR cuGraph failure at file=/opt/conda/conda-bld/work/cpp/src/sampling/sampling_post_processing_impl.cuh line=235: Invalid input arguments: num_labels should be 0 if the input edge list is empty.
> ```

@ChuckHastings I assume this is also due to the fix in "num_hops == 0," right?

cpp/src/sampling/sampling_post_processing_impl.cuh
```diff
@@ -209,13 +221,26 @@ void check_input_edges(raft::handle_t const& handle,
...
+                    "Invalid input arguments: vertex_type_offsets.has_value() should be true if "
+                    "num_vertex_types >= 2.");
   } else {
     CUGRAPH_EXPECTS(
-      "num_labels == 0",
+      num_labels == 0,
       "Invalid input arguments: num_labels should be 0 if the input edge list is empty.");
     CUGRAPH_EXPECTS(
-      "num_hops == 0",
+      num_hops == 0,
```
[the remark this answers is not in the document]
seunghwak: Yeah... I guess you're right. We can check only for nun_labels == 0 for an empty edge list and allow num_hops & num_vertex_types to have a non zero value.

PR #4603, opened 2024-08-08
Fix ucx-py update-version script

raydouglass:
Back ports the `update-version.sh` fixes from https://github.com/rapidsai/cugraph/pull/4562


raydouglass:
Admin merging to unblock the release

PR #4606, opened 2024-08-08
[FIX] Shuffle Inputs Globally in cuGraph-PyG

alexbarghi-nv:
Previously, inputs were only being shuffled locally.  This PR uses NCCL to shuffle inputs across workers for better accuracy.  It also exposes an option to disable global shuffling if desired.


tingyu66 (approved):

alexbarghi-nv:
This will be refactored and merged into `cugraph-gnn` instead.

PR #4610, opened 2024-08-14
Updates to `cugraph.hypergraph` (Duplicate Col Labels Bug)

nv-rliu:
cc: @rlratzel @ChuckHastings 

This PR addresses failures seen in certain PRs (like [here](https://github.com/rapidsai/cugraph/actions/runs/10372270389/job/28718471674?pr=4606#step:7:5269)) due to a [recent change](https://github.com/rapidsai/cudf/pull/16514) to `cudf` that disallows selecting duplicate column labels.

---

In `hypergraph.py`, this PR modifies `_create_hyper_edges` and `_create_direct_edges` to ensure that DataFrames are being indexed by non-duplicate column values.

This is done by taking a list that includes duplicates (`fs`), and removing the non-unique values
```python
fs = list(set(fs))
```

_This part requires some attention from the author of the unit test @jnke2016_

In `test_hypergraph.py`, this PR adds the `check_like=True` arg to `assert_frame_equals` function because the ordering of the columns is different for the two DFs.

### DIFF
--- a/python/cugraph/cugraph/structure/hypergraph.py
+++ b/python/cugraph/cugraph/structure/hypergraph.py
@@ -441,7 +441,6 @@ def _create_hyper_edges(
         cat = categories.get(key, key)
         fs = [EVENTID] + ([key] if drop_edge_attrs else edge_attrs)
         fs = list(set(fs))
-        # breakpoint()
         df = events[fs].dropna(subset=[key]) if dropna else events[fs]
         if len(df) == 0:
             continue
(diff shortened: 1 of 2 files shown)


rlratzel (approved):
Thanks Ralph!

ChuckHastings (approved):

trxcllnt (approved):

jnke2016 (approved):

PR #4612, opened 2024-08-15
Set CUDA_STATIC_MATH_LIBRARIES in Python builds

KyleFromNVIDIA:
In https://github.com/rapidsai/cugraph/pull/4526, we introduced the CUDA_STATIC_MATH_LIBRARIES option and set it to OFF by default. We did not set it to ON in the Python builds, and as a result ended up bundling dynamic CUDA libraries. Set the option to ON in the Python build.


vyasr:
Thanks Kyle, before I approve this, let's wait until a wheel build completes and we can inspect the resulting outputs. If the sizes go back down then we'll know things worked as expected.

KyleFromNVIDIA:
Comparing https://downloads.rapids.ai/ci/cugraph/branch/branch-24.10/a538745/ (before this change) to https://downloads.rapids.ai/ci/cugraph/pull-request/4612/aa83fbf/ (after this change), it looks like there was a slight reduction in wheel size, but not massive.

vyasr:
Interestingly, both look much smaller than what Robert was observing in his document.

vyasr (approved):
I'm recommending that we merge this fix (since it's definitely needed) and then we reassess package sizes afterwards to see if there's some other additional cause for increased wheel sizes in 24.08.

vyasr (changes requested):
Actually sorry hold on, I'm seeing some odd things that make me want to investigate a bit more.

vyasr (approved):
OK, so strictly speaking this isn't related to #4526, but rather https://github.com/rapidsai/raft/pull/2376. #4526 changed the build options for cugraph_etl, which doesn't get built into the wheel anyway right, but the changes in raft needed to be propagated up since we rebuild/statically link raft in our wheels.

However, that still doesn't account for the full binary size increase. Most of that is coming from the fact that libcugraph itself seems to have nearly doubled in size from 24.06 to 24.08. We need to look into that, but we can do that independently since it's not a wheels issue (the size difference is evident even in [our libcguraph conda packages](https://anaconda.org/rapidsai/libcugraph/files)).

PR #4613, opened 2024-08-15
Add Additional Check for `NetworkX` Release Candidate Versions

nv-rliu:
Part of https://github.com/rapidsai/graph_dl/issues/579

This PR allows `nx-cugraph` to operate with release candidate (under development) versions of `networkx`

What This Pattern Allows:
- Just a single digit: `3.4`, `3.9`, etc

- Release Candidate format: `3.4rc0`. `3.7rc2`

This is needed to complete adding nightly test coverage for development branches of `nx`

### DIFF
--- a/python/nx-cugraph/_nx_cugraph/__init__.py
+++ b/python/nx-cugraph/_nx_cugraph/__init__.py
@@ -298,6 +298,7 @@ def get_info():
 
 def _check_networkx_version():
     import warnings
+    import re
 
     import networkx as nx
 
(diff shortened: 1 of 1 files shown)


eriknw (approved):
Thanks @nv-rliu, and my bad for missing this!

Btw, `re.match` already matches only from the beginning of the line, so beginning the pattern with `"^"` is not strictly necessary, but it's fine to be explicit.

nv-rliu:
> Thanks @nv-rliu, and my bad for missing this!
> 
> Btw, `re.match` already matches only from the beginning of the line, so beginning the pattern with `"^"` is not strictly necessary, but it's fine to be explicit.

Thanks for the review. And neat tip! TIL 😄

rlratzel (approved):

PR #2056, opened 2022-02-02
PropertyGraph updates: added features for DGL, improved `extract_subgraph()` and `num_vertices` performance

rlratzel:
Various updates needed for adding new features for DGL, improving `extract_subgraph()` and `num_vertices` performance:

* made `PropertyGraph` column name constants public for use by clients
* changed `num_vertices` property impl for better performance
* added `get_vertices()`
* allow a `None` type_name for `add_*_data()`
* updated docstrings
* changed the `edge_data` added to Graphs to be a DataFrame for efficiency
* added tests and benchmarks
* refactored and added test fixtures
* updated `pytest.ini` for new benchmarks

Tested by ensuring all new tests added passed, benchmarks ran, and existing `test_graph_store.py` tests still passed.

PR #1999, opened 2022-01-03
Initial PropertyGraph implementation and tests

rlratzel:
Initial implementation of the `PropertyGraph` graph class and corresponding unit tests.
```
pG = cugraph.experimental.PropertyGraph()

# Add properties
pG.add_vertex_data(...)
pG.add_edge_data(...)

# Select specific vertices and edges based on conditions
selection = pG.select_vertices("(user_location==47906) | (user_location==78750)")
selection += pG.select_edges("_TYPE_=='referrals'")

# Extract a subgraph using selection
G = pG.extract_subgraph(selection=selection,
                        create_using=DiGraph_inst,
                        edge_weight_property="review_rating")

# run some algo on G, get algo_results

# annotate results with properties
algo_results_with_properties = pG.annotate_dataframe(algo_results, G, edge_vertex_id_columns=("from", "to"))

# Add algo results back to the PropertyGraph for further analysis
pG.add_edge_data(algo_results, type_name="referral_data", vertex_id_columns=("from", "to"))
```

closes #1876

PR #4615, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).


seberg:
Updating branch to trigger CI.  I _think_ RAPIDS dependencies should now be resolved sufficiently to actually get NumPy 2 (not sure about all runs, as torch may get in the way depending on the version).

seberg:
I realized that NumPy and torch are not runtime requirements for `wholegraph`.  Is that is correct in that they are optional dependencies only?

(I don't think it affects this PR, it actually means that `wholegraph` should not block test `cugraph` with NumPy 2.1.)

jakirkham:
Updating branch to pull in the latest upstream changes and restart CI now that cuDF is done: https://github.com/rapidsai/cudf/pull/16300

jakirkham:
Documenting for posterity

Seeing some unrelated C++ failures [on CI]( https://github.com/rapidsai/cugraph/actions/runs/10536202066/job/29197447050?pr=4615#step:7:870 ):

```cpp
The following tests FAILED:
	 50 - EDGE_TRIANGLE_COUNT_TEST (Failed)
	 51 - LOOKUP_SRC_DST_TEST (Failed)
	 52 - K_HOP_NBRS_TEST (Failed)
```

Will restart the failed job after CI completes

seberg:
Just to note, the test failures seem all due to an out of memory.
```
terminate called after throwing an instance of 'rmm::out_of_memory'
  what():  std::bad_alloc: out_of_memory: RMM failure at:/opt/conda/envs/test/include/rmm/mr/device/pool_memory_resource.hpp:257: Maximum pool size exceeded
```

jakirkham:
So potentially related to general CI issues we have seen: https://github.com/rapidsai/cuml/pull/6031#issuecomment-2308158465

dependencies.yaml
```diff
@@ -710,36 +708,6 @@ dependencies:
...
-          - matrix: {cuda: "12.*"}
-            packages:
-              - &pytorch_pip torch>=2.0,<2.2.0a0
-              - *tensordict
-          - matrix: {cuda: "11.*"}
-            packages:
-              - *pytorch_pip
-              - *tensordict
-          - {matrix: null, packages: [*pytorch_pip, *tensordict]}
-
```
jakirkham, line 742: Think it is ok to keep this. We just want to relax the upper bound. At a minimum we need PyTorch 2.3.0 for NumPy 2 support ( https://github.com/pytorch/pytorch/issues/107302 ). Though wonder if this upper bound can be dropped altogether

dependencies.yaml
```diff
@@ -662,10 +658,12 @@ dependencies:
       - output_types: [conda, pyproject]
         packages:
           - pandas
+          - pydantic
```
jakirkham, line 661: Curious what `pydantic` is being used for

jakirkham:
Have gone ahead and [restarted CI]( https://github.com/rapidsai/cugraph/actions/runs/10571793872?pr=4615 )

seberg:
Seems those failures did not clear up, although it isn't clear to me why (or if) they are related to these changes.

jakirkham:
Including the full traceback of one such error with cuDF [on CI]( https://github.com/rapidsai/cugraph/actions/runs/10571793872/job/29321491712?pr=4615#step:7:5263 )

Note there is also a Pandas failure, which looks identical

```python
___________ bench_extract_subgraph_for_rmat[df_type=cudf.DataFrame] ____________

gpubenchmark = <pytest_benchmark.fixture.BenchmarkFixture object at 0x7fe8f938f950>
rmat_PropertyGraph = (<cugraph.experimental.PropertyGraph object at 0x7fe90cf59410>,              src      dst    weight
0         632844  ... 543777  0.816715
16777214  218373   734889  0.940356
16777215  762125  1012603  0.426682

[16777216 rows x 3 columns])

    def bench_extract_subgraph_for_rmat(gpubenchmark, rmat_PropertyGraph):
    
        (pG, generated_df) = rmat_PropertyGraph
        scn = PropertyGraph.src_col_name
        dcn = PropertyGraph.dst_col_name
    
        verts = []
        for i in range(0, 10000, 10):
            verts.append(generated_df["src"].iloc[i])
    
>       selected_edges = pG.select_edges(f"{scn}.isin({verts}) | {dcn}.isin({verts})")

tests/data_store/test_property_graph.py:2583: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/conda/envs/test/lib/python3.11/site-packages/cugraph/structure/property_graph.py:1555: in select_edges
    selected_col = eval(expr, globals, locals)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   NameError: name 'np' is not defined

<string>:1: NameError
```

Seems to be pointing to this line, which uses `eval`. Unfortunately it is hard to debug past there

https://github.com/rapidsai/cugraph/blob/f7e1ceafaa8483f6aeb7353373dcccc2901c33ba/python/cugraph/cugraph/structure/property_graph.py#L1555

Best guess is the globals need to include some `import`ed modules (like `np`)

https://github.com/rapidsai/cugraph/blob/f7e1ceafaa8483f6aeb7353373dcccc2901c33ba/python/cugraph/cugraph/structure/property_graph.py#L1552

jakirkham:
Looks like `cugraph.structure.property_graph` (and `select_edges` specifically) originates in PR: https://github.com/rapidsai/cugraph/pull/1999

The test that fails, `bench_extract_subgraph_for_rmat`, comes from PR: https://github.com/rapidsai/cugraph/pull/2056

Both authored by @rlratzel. Perhaps Rick can advise 🙂

dependencies.yaml
alexbarghi-nv: DGL requires it.  It tries to import it and raises an exception if it isn't available.  It's used by GraphBolt.

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2576,9 +2576,12 @@ def bench_extract_subgraph_for_rmat(gpubenchmark, rmat_PropertyGraph):
     scn = PropertyGraph.src_col_name
     dcn = PropertyGraph.dst_col_name
 
+    # Build a query string to extract a graph with only specific edges based on
+    # the integer vertex IDs. Other edge and/or vertex properties can be
+    # included in the query as well.
     verts = []
     for i in range(0, 10000, 10):
```
jakirkham, line 2583:
```suggestion
```

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2576,9 +2576,12 @@ def bench_extract_subgraph_for_rmat(gpubenchmark, rmat_PropertyGraph):
...
-        verts.append(generated_df["src"].iloc[i])
+        verts.append(int(generated_df["src"].iloc[i]))
```
jakirkham, line 2584:
```suggestion
    verts = [int(generated_df["src"].iloc[i]) for i in range(0, 10000, 10)]
```

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2618,9 +2621,12 @@ def bench_extract_subgraph_for_rmat_detect_duplicate_edges(
     scn = PropertyGraph.src_col_name
     dcn = PropertyGraph.dst_col_name
 
+    # Build a query string to extract a graph with only specific edges based on
+    # the integer vertex IDs. Other edge and/or vertex properties can be
+    # included in the query as well.
     verts = []
     for i in range(0, 10000, 10):
```
jakirkham, line 2628:
```suggestion
```

dependencies.yaml
alexbarghi-nv: As long as we have cupy>13.2 in our test environment we are ok to remove the upper bound

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2618,9 +2621,12 @@ def bench_extract_subgraph_for_rmat_detect_duplicate_edges(
...
-        verts.append(generated_df["src"].iloc[i])
+        verts.append(int(generated_df["src"].iloc[i]))
```
jakirkham, line 2629:
```suggestion
    verts = [int(generated_df["src"].iloc[i]) for i in range(0, 10000, 10)]
```

dependencies.yaml
alexbarghi-nv: I'm also ok if we enforce pytorch >=2.3 for both test and runtime

jakirkham (commented):
Thanks Rick! 🙏

Would it make sense to use `list` comprehensions here?

python/cugraph/cugraph/tests/data_store/test_property_graph.py
rlratzel: Thanks, good suggestions, I just pushed a commit to do this and the others below.

jakirkham:
Looks like there were a couple CI jobs with network errors

Also seeing some test failures related to cuDF. Think these are likely fixed by PR: https://github.com/rapidsai/cugraph/pull/4618

Will wait for that PR to land and then merge in latest upstream changes to retest

jakirkham:
Now that the aforementioned PR is in, updating this PR to include that fix and rerun CI

jakirkham:
There was a networking error in one of the conda builds

Also one of the wheel builds appears to have gotten stuck

Canceled the stuck job and restarted

alexbarghi-nv (approved):
👍

seberg:
Close/re-open to trigger CI.  Were still related issues here or is it in principle good to go?

hcho3:
Got two errors:

* Rockylinux8 amd64: `AssertionError: Torch not compiled with CUDA enabled`
* Ubuntu20.04 arm64: `RuntimeError: PyTorch was compiled without NumPy support`

ci/test_python.sh
```diff
@@ -160,6 +160,7 @@ if [[ "${RAPIDS_CUDA_VERSION}" == "11.8.0" ]]; then
       cugraph-dgl \
       'dgl>=1.1.0.cu*,<=2.0.0.cu*' \
       'pytorch>=2.0' \
+      'numpy>=2.0' \
```
hcho3, line 163: Added this line temporarily to figure out the root cause of the downgrade:
```
2024-09-20T01:06:45.8567605Z   Downgrade:
2024-09-20T01:06:45.8568641Z ──────────────────────────────────────────────────────────────────────────────────────────────────────────────────
2024-09-20T01:06:45.8569407Z 
2024-09-20T01:06:45.8570152Z   - libtorch                          2.4.0  cuda118_h8db9d67_301                 conda-forge             Cached
2024-09-20T01:06:45.8571690Z   + libtorch                          2.1.2  cpu_mkl_hff68eba_104                 conda-forge               47MB
2024-09-20T01:06:45.8573188Z   - numpy                             2.0.2  py310hd6e36ab_0                      conda-forge             Cached
2024-09-20T01:06:45.8574693Z   + numpy                            1.26.4  py310hb13e2d6_0                      conda-forge             Cached
2024-09-20T01:06:45.8576206Z   - pytorch                           2.4.0  cuda118_py310h954aa82_301            conda-forge             Cached
2024-09-20T01:06:45.8577750Z   + pytorch                           2.1.2  cpu_mkl_py310hcb3bde6_104            conda-forge               27MB
2024-09-20T01:06:45.8579524Z   - torchdata                         0.7.1  py310h68cf1eb_9                      conda-forge             Cached
2024-09-20T01:06:45.8581034Z   + torchdata                         0.4.1  pyh8b8bddf_0                         conda-forge             Cached
```

hcho3:
Adding `pytorch<2.4` pin to work around https://github.com/conda-forge/pytorch-cpu-feedstock/issues/254#issuecomment-2319090902

dependencies.yaml
jameslamb: Let's keep the discussion about what to do with the `pytorch` dependency in one thread, for ease of linking.

@hcho3 would `>=2.3.0`, as @alexbarghi-nv suggested here, work? A floor would be preferable to a ceiling, I think...otherwise this will have to be changed again in the next release.

Note: we should match whatever's done here in https://github.com/rapidsai/wholegraph/pull/218

hcho3: No, because 2.4.0 has issues with Conda-forge packaging. See https://github.com/conda-forge/pytorch-cpu-feedstock/issues/254#issuecomment-2319090902

dependencies.yaml
```diff
@@ -696,7 +696,7 @@ dependencies:
       - output_types: [conda]
         packages:
           - *cugraph_unsuffixed
-          - pytorch>=2.0
+          - pytorch>=2.0,<2.4
```
jameslamb, line 699:
```suggestion
          # ceiling could be removed when this is fixed:
          # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/254
          - &pytorch_conda pytorch>=2.0,<2.4.0a0
```

Two suggestions:

1. could you please leave a comment in code here pointing back to the root cause, to hopefully make it easier for someone to notice that loosening this pin is worth investigating in the future?
2. To keep all these `pytorch` dependencies consistent, could you please use an anchor `&pytorch_conda` instead of `&pytorch_unsuffixed`? I know that use of `_unsuffixed` existed before this PR (probably my fault, honestly), but let's clean it up now that these exact lines are being touched.

dependencies.yaml
```diff
@@ -705,7 +705,7 @@ dependencies:
...
```
jameslamb, line 708:
```suggestion
          - *pytorch_conda
```

dependencies.yaml
```diff
@@ -714,7 +714,7 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - &pytorch_unsuffixed pytorch>=2.0,<2.2.0a0
+          - &pytorch_unsuffixed pytorch>=2.0,<2.4
```
jameslamb, line 717:
```suggestion
          - *pytorch_conda
```

dependencies.yaml
```diff
@@ -734,7 +734,7 @@ dependencies:
         matrices:
           - matrix: {cuda: "12.*"}
             packages:
-              - &pytorch_pip torch>=2.0,<2.2.0a0
+              - &pytorch_pip torch>=2.0,<2.4
```
jameslamb, line 737:
```suggestion
              - &pytorch_pip torch>=2.0,<2.4.0a0
```

I think the alpha spec should be preserved here, to avoid accidentally pulling in alpha releases of `torch` 2.4

jameslamb (changes requested):
Thanks! I left some suggestions with the assumption that we're going to adopt the `<2.4` pin on `pytorch` for this release. Will be curious to hear what @alexbarghi-nv says about that.

I looked through CI and it seemed that the failures were builds timing out after 6 hours....which is something I've seen in this repo occasionally over the last 2 months, and unlikely to be related to these changes. Hopefully a new build will get more sccache hits and complete faster.

jameslamb (changes requested):
@hcho3 do you want any help investigating the failing CI jobs?

I see 3 issues here. I think 2 might be resolved by a restart of CI, so I've done that by merging `branch-24.10` in here.

---

**1: `conda-cpp-tests / tests (arm64, 3.11, 12.0.1, ubuntu20.04, a100, latest, latest)`**

```text
The following packages are incompatible
├─ libcugraph is not installable because there are no viable options
│  ├─ libcugraph [24.10.00a101|24.10.00a76|...|24.10.00a81] would require
│  │  └─ libraft-headers 24.10.*  but there are no viable options
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/11021389414/job/30623854329?pr=4615))

^ this would probably be solved by just merging the latest `branch-24.10` and doing a full re-run of CI. I suspect the issue here is that you got a mix of old and new raft packages based on the timing of when `fmt` / `spdlog` changes like https://github.com/rapidsai/raft/pull/2433 were merged.

---

**2: `conda-python-tests / 11.8.0, 3.10, amd64, rockylinux8, v100, earliest-driver, oldest-deps`**

No logs ([build link](https://github.com/rapidsai/cugraph/actions/runs/11021389414/job/30623863433?pr=4615)), usually implies an out-of-memory on the runner or some other infrastructure-level issue. Hopefully would be fixed by a re-run.

---

**3: `wheel-tests-cugraph-pyg / 12.5.1, 3.10, amd64, ubuntu22.04, v100, latest-driver, oldest-deps`**

Dozens of tests failing like this:

```text
RuntimeError: Numpy is not available
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/11021389414/job/30623874191?pr=4615))

I believe that's coming from `pytorch`, e.g.:

* https://github.com/pytorch/pytorch/blob/5a29a06aa3c2dbb3afef730fbf9e7c0e2f308bf7/torch/csrc/utils/tensor_numpy.cpp#L127
* https://github.com/pytorch/pytorch/blob/5a29a06aa3c2dbb3afef730fbf9e7c0e2f308bf7/torch/csrc/utils/tensor_numpy.cpp#L218-L220
* https://github.com/pytorch/pytorch/blob/5a29a06aa3c2dbb3afef730fbf9e7c0e2f308bf7/torch/csrc/utils/tensor_numpy.cpp#L381-L383

In those builds, it looks like we're getting the following:

```text
Successfully installed ...
cuda-python-12.6.0
...
numba-0.60.0
numpy-2.0.2
...
pandas-2.2.2
...
torch-2.1.0+cu121
```

I'm not sure what the root cause is. Can you please try to figure that out, or at least get to a minimum reproducible example that doesn't require a full CI run to test?

Maybe allowing that to float by loosening this would help:

https://github.com/rapidsai/cugraph/blob/1e5b3287e628a5cf8e60da14c25174baf0e5cc1a/ci/test_wheel_cugraph-pyg.sh#L38

If we cannot figure it out in the next day or two, we should consider leaving the tighter pin on `pytorch` in place only for `cugraph-pyg` and then addressing that separately (separate issue, follow-up PR). That would at least allow `cugraph` and `nx-cugraph` to be installed alongside NumPy 2, which would allow the `rapids` metapackage to be installed with NumPy 2.

jakirkham:
Thanks for taking a look James! 🙏

Think the last issue is due to the fact that NumPy 2 + PyTorch 2.1.0 are incompatible. PyTorch did not start supporting NumPy 2 until PyTorch 2.3.0 ( https://github.com/pytorch/pytorch/issues/107302 )

There are some wrinkles with Windows wheels ( https://github.com/pytorch/pytorch/issues/131668 ), which were not fixed until later, but think we can ignore those

In any event, the simplest path is probably to bump the lower bound on PyTorch to 2.3.0. Arguably that shouldn't be needed if the PyTorch wheels properly constrain NumPy, but they may not all do that. So the lower bound is the more reliable approach

jameslamb:
> In any event, the simplest path is probably to bump the lower bound on PyTorch to 2.3.0. 

Ah great! Sorry, I remember now that you said that a while ago: https://github.com/rapidsai/cugraph/pull/4615#discussion_r1730635145

I guess these were just missed in earlier versions of this PR because it's in a script instead of `pyproject.toml` / `dependencies.yaml` / `conda/recipes/*/meta.yaml`:

https://github.com/rapidsai/cugraph/blob/1e5b3287e628a5cf8e60da14c25174baf0e5cc1a/ci/test_wheel_cugraph-dgl.sh#L46-L47

https://github.com/rapidsai/cugraph/blob/1e5b3287e628a5cf8e60da14c25174baf0e5cc1a/ci/test_wheel_cugraph-pyg.sh#L38-L39

@hcho3 could you update those to have a floor of pytorch 2.3.0? And it would probably also be helpful to combine them instead of doing multiple consecutive `pip install` calls, to reduce the risk of a later call downgrading a dependency installed by an earlier call.

ci/test_wheel_cugraph-dgl.sh
```diff
@@ -32,18 +32,8 @@ fi
...
 
-# Starting from 2.2, PyTorch wheels depend on nvidia-nccl-cuxx>=2.19 wheel and
-# dynamically link to NCCL. RAPIDS CUDA 11 CI images have an older NCCL version that
-# might shadow the newer NCCL required by PyTorch during import (when importing
-# `cupy` before `torch`).
-if [[ "${NCCL_VERSION}" < "2.19" ]]; then
-  PYTORCH_VER="2.1.0"
-else
-  PYTORCH_VER="2.3.0"
-fi
```
jakirkham, line 43: We recently bumped the NCCL floor to 2.18.1.1 in PR: https://github.com/rapidsai/cugraph/pull/4661

Reading this comment, am wondering whether we need to bump it further (namely to 2.19)

dependencies.yaml
```diff
@@ -695,7 +695,9 @@ dependencies:
       - output_types: [conda]
         packages:
           - *cugraph_unsuffixed
-          - pytorch>=2.0
+          # ceiling could be removed when this is fixed:
+          # https://github.com/conda-forge/pytorch-cpu-feedstock/issues/254
+          - &pytorch_conda pytorch>=2.0,<2.4.0a0
```
jakirkham, line 700: Do we want to bump the floor on these as well?

ci/test_wheel_cugraph-dgl.sh
jakirkham: Raising in the NCCL update issue: https://github.com/rapidsai/build-planning/issues/102#issuecomment-2375595743

ci/test_python.sh
```diff
@@ -159,7 +159,8 @@ if [[ "${RAPIDS_CUDA_VERSION}" == "11.8.0" ]]; then
       cugraph \
       cugraph-dgl \
       'dgl>=1.1.0.cu*,<=2.0.0.cu*' \
-      'pytorch>=2.0' \
+      'pytorch>=2.0,<2.4' \
+      'numpy>=2.0' \
```
jakirkham, line 163: Are we able to relax the NumPy pin now?

Also may want to bump the lower bound on PyTorch here to ensure NumPy 2 compatibility, which was added to the conda-forge package in PR: https://github.com/conda-forge/pytorch-cpu-feedstock/pull/238

There are a few more cases where this happens below, which we may want to give similar treatment

```suggestion
      'pytorch>=2.3,<2.4' \
```

alexbarghi-nv: PyTorch 2.3+ is ok with me

jameslamb: Thank you! I know you'd said that in https://github.com/rapidsai/cugraph/pull/4615#discussion_r1735256909 as well, this conversation is getting hard to follow split across multiple similar threads.

@hcho3 @jakirkham it seems to me like we have a clear path forward here:

* remove all the `numpy>2` that were temporarily added for testing
* set all `pytorch` / `torch` pins everywhere to `>=2.3,<2.4`
* bump `nccl` to `nccl>=2.19` (ref: https://github.com/rapidsai/build-planning/issues/102#issuecomment-2377268840)

jameslamb: I just pulled this branch and searched like this:

```text
git grep -E 'orch.*2\.'
```

Looks like there's one more `pytorch=2.1.0` reference that probably should be updated to 2.3:

https://github.com/seberg/cugraph/blob/a6a84ebca91d0b0e8f1dabd0fa5f8be08283d82b/ci/test_python.sh#L198-L205

In a future PR, it'd be great to drive more of these things through `dependencies.yaml` and `rapids-dependency-file-generator`, to make it less likely that things like this are missed in upgrades.

alexbarghi-nv: Yeah, I think this should be resolved in `cugraph-gnn`.  But if it isn't I agree we should fix it.  This got in there because of some of the quirks of the conda environment we use to test cuGraph.  `cugraph-gnn` aligns the package versions with PyTorch, and shouldn't have to do this.

hcho3: Done

jakirkham:
Thanks Philip! 🙏

Think this is looking pretty good

Would be great to hear from James and Alex as well 🙂

jameslamb (approved):
Approving based on what I see in the diff and my own search through the repo... I don't see any references that were missed, and I do believe the new floors on `pytorch` + NCCL will be enough to fix the issues we saw.

But let's please not merge until we see NumPy 2 in the logs of some test jobs (especially `conda-python-tests`).

jakirkham:
Two of the test jobs errored during source checkout. One of the other builds got hung. The rest of the jobs passed. Canceled to clear out the hung job. Have now restarted those 3 jobs

jakirkham:
That said, [one of the Conda test jobs]( https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30740287411?pr=4615#step:7:446 ) does show NumPy 2 is installed

```
numpy                     2.0.2           py310hd6e36ab_0    conda-forge
```

jameslamb:
This wheel test job was failing, so I restarted it:

````text
wheel-tests-cugraph / 12.5.1, 3.10, amd64, ubuntu22.04, v100, latest-driver, oldest-deps
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30768785958?pr=4615))

It's been running for 3+ hours and I see a ton of error there (no logs with specific tracebacks reported yet, but it's 98% done so close). e.g. like this:

```text
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset0] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset1] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset2] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset0] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset1] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset2] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset0] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset1] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset2] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset0] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset1] ERROR [ 83%]
python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset2] ERROR [ 83%]
```

The other `wheel-tests-cugraph` job succeeded in under 9 minutes.

```text
wheel-tests-cugraph / 11.8.0, 3.12, arm64, ubuntu20.04, a100, latest-driver, latest-deps
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30768786476?pr=4615))

We'll know more once the job completes and we can see tracebacks, but I think that needs some investigation.

alexbarghi-nv:
> This wheel test job was failing, so I restarted it:
> 
> ```
> wheel-tests-cugraph / 12.5.1, 3.10, amd64, ubuntu22.04, v100, latest-driver, oldest-deps
> ```
> 
> ([build link](https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30768785958?pr=4615))
> 
> It's been running for 3+ hours and I see a ton of error there (no logs with specific tracebacks reported yet, but it's 98% done so close). e.g. like this:
> 
> ```
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset0] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset1] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame0-dataset2] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset0] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset1] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[16-DataFrame1-dataset2] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset0] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset1] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame0-dataset2] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset0] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset1] ERROR [ 83%]
> python/cugraph/cugraph/tests/sampling/test_uniform_neighbor_sample_mg.py::test_uniform_neighbor_sample_batched[32-DataFrame1-dataset2] ERROR [ 83%]
> ```
> 
> The other `wheel-tests-cugraph` job succeeded in under 9 minutes.
> 
> ```
> wheel-tests-cugraph / 11.8.0, 3.12, arm64, ubuntu20.04, a100, latest-driver, latest-deps
> ```
> 
> ([build link](https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30768786476?pr=4615))
> 
> We'll know more once the job completes and we can see tracebacks, but I think that needs some investigation.

I'm seeing similar failures in some of the other PRs.  Usually these pass and it's the PyG tests that fail.  But the failure doesn't seem PyG specific.  Is there any chance we pulled the wrong version of dask somewhere?

jameslamb:
Maybe! The errors do look Dask-y

```text
E           Exception: Communicator is already initialized
/pyenv/versions/3.10.15/lib/python3.10/site-packages/cugraph/dask/comms/comms.py:156: Exception
_ ERROR at setup of test_dask_mg_all_pairs_sorensen[graph_file:/__w/cugraph/cugraph/datasets/email-Eu-core.csv-directed:False-has_vertex_pair:False-has_vertices:True-has_topk:False-is_weighted:True] _
    def dask_client():
        # start_dask_client will check for the SCHEDULER_FILE and
        # DASK_WORKER_DEVICES env vars and use them when creating a client if
        # set. start_dask_client will also initialize the Comms singleton.
>       dask_client, dask_cluster = start_dask_client(
            worker_class=IncreasedCloseTimeoutNanny
        )
```

It seems that in that job, we got these:

```text
dask 2024.9.0
distributed 2024.9.0
dask-expr 1.1.14
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/11060465303/job/30768785958?pr=4615))

Same versions I see in the successful wheel-testing job :/

jakirkham:
Well before that we had a wheel build that was taking a really long time. Honestly am wondering if there is just a bad node in the GHA pool that needs a reboot or perhaps exclusion until it can be repaired

jakirkham:
Rerunning the failed job again

The fact that there are two test with slightly different configurations running this test suite with one passing and the other failing is odd. It would be easier to believe Dask was the issue if both jobs were failing (or even more jobs were affected)

Given the flakiness we have seen here with CI, suspect there is an issue in the CI pool itself. Flagging this offline for follow up with the appropriate team

jakirkham:
Am noticing this looks a little out-of-date relative to `branch-24.10`. Going to try merging in latest in case that helps

jameslamb (approved):
Re-approving. Seeing all CI passing again on a re-run, and NumPy 2 being used in all the wheel and conda test jobs, has me feeling confident that this is working as intended.

I also `git grep`'d around one more time to check for missed references to NumPy, torch/PyTorch, and NCCL... didn't find any.

From my perspective, this can be merged.

jameslamb:
There are several other PRs in `cugraph` trying to get into `24.10` in the next few days. So if this really did introduce flakiness (instead of it, say, being the result of some issue with the CI runners), we'll find out about it soon.

Given that and @alexbarghi-nv 's written approval of the main changes here in the threads on this PR, I'm confident we can merge this, so I'm going to merge it.

@alexbarghi-nv please `@` me any time if you see other issues like https://github.com/rapidsai/cugraph/pull/4615#issuecomment-2379854819 on the other PRs, I'd be happy to come help investigate.

jakirkham:
Thanks all! 🙏

dependencies.yaml
jakirkham: This was recently resolved. So have submitted PR ( https://github.com/rapidsai/cugraph/pull/4703 ) to relax this in 24.12

PR #4616, opened 2024-08-19
Add `nx-cugraph` Benchmarking Scripts

nv-rliu:
Closes https://github.com/rapidsai/graph_dl/issues/596
This PR adds scripts written by @rlratzel to generate benchmarking numbers for `nx-cugraph`


benchmarks/nx-cugraph/pytest-based/run-gap-benchmarks.sh
```diff
@@ -0,0 +1,65 @@
...
+    triangles
+    bfs_predecessors
+"
+datasets="
+   netscience
+   email_Eu_core
+   cit_patents
+   hollywood
+   soc-livejournal
+"
```
acostadon, line 36: Why is this repeated?

acostadon (approved):
Looks good. It might be good to put an approximate runtime just so users know what they are in for. Even a single example with a specified architecture might set expectations.

This looks great though  !!

benchmarks/nx-cugraph/pytest-based/run-gap-benchmarks.sh
nv-rliu: Hi Don. That was removed once this PR was moved out of draft!

benchmarks/nx-cugraph/pytest-based/README.md
```diff
@@ -0,0 +1,45 @@
...
+| email-Eu-core  | 1,005    | 25,571 | Yes |
+| cit-Patents  | 3,774,768    | 16,518,948 | Yes |
+| hollywood  | 1,139,905    | 57,515,616 | No |
+| soc-LiveJournal1  | 4,847,571    | 68,993,773 | Yes |
+
+
+
+### Scripts
+
+#### 1. `run-gap-benchmarks.sh`
```
rlratzel, line 21: I like that we're using many of the same algos covered by the GAP benchmark, but I think it might be misleading to refer to this as "gap" since GAP also specifies the graph inputs, has a reference implementation to compare to, and possibly other things a user might expect to see.

Can we use another name for this script, perhaps just "run-main-benchmarks.sh" or something to indicate it's just the top n algos we've selected?

benchmarks/nx-cugraph/pytest-based/create_results_summary_page.py
```diff
@@ -0,0 +1,245 @@
...
+        }
+    </style>
+    </head>
+    <table>
+    <thead>
+    <tr>
+        <th></th>"""
+    )
+    for ds in ordered_datasets:
+        print(f"      <th>{ds}</th>")
```
rlratzel, line 199: Can you make this a multi-line cell and also include the number of nodes and edges and if it's directed or undirected?

rlratzel (commented):
Thanks for adding the README.  I just had a few hopefully small requests.

rlratzel (approved):

PR #4620, opened 2024-08-20
Fix ListColumn constructor argument

mroeschke:
A recent API change in cudf (https://github.com/rapidsai/cudf/pull/16465) now requires the `data` argument to be passed (consistent with the other Column subclasses)


rlratzel (approved):
Thanks!

PR #4625, opened 2024-08-21
Drop Python 3.9 support

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/88

Finishes the work of dropping Python 3.9 support.

This project stopped building / testing against Python 3.9 as of https://github.com/rapidsai/shared-workflows/pull/235.
This PR updates configuration and docs to reflect that.

### How I tested this

Checked that there were no remaining uses like this:

```shell
git grep -E '3\.9'
git grep '39'
git grep 'py39'
```

And similar for variations on Python 3.8 (to catch things that were missed the last time this was done).


python/nx-cugraph/pyproject.toml
```diff
@@ -90,7 +89,7 @@ matrix-entry = "cuda_suffixed=true"
 
 [tool.black]
 line-length = 88
-target-version = ["py39", "py310", "py311"]
+target-version = ["py310", "py310", "py311"]
```
KyleFromNVIDIA, line 92: Why is `py310` listed here twice?

jameslamb:
```suggestion
target-version = ["py310", "py311"]
```

oh wow, thank you! Think my `rapids-reviser` code did not expect this to ever include more than one value. But sure enough, it can (and in fact the `black` docs even say it *should*): https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version

docs/cugraph/source/tutorials/basic_cugraph.md
```diff
@@ -5,7 +5,7 @@
 CuGraph is part of [Rapids](https://docs.rapids.ai/user-guide) and has the following system requirements:
  * NVIDIA GPU, Volta architecture or later, with [compute capability](https://developer.nvidia.com/cuda-gpus) 7.0+
  * CUDA 11.2, 11.4, 11.5, 11.8, 12.0 or 12.2
```
bdice, line 7: We might want to slip this update in here too.
```suggestion
 * CUDA 11.2, 11.4, 11.5, 11.8, 12.0, 12.2, or 12.5
```

bdice (approved):

docs/cugraph/source/tutorials/basic_cugraph.md
jameslamb: sure! fine with me

BradReesWork (approved):

PR #4630, opened 2024-08-24
Update edge triangle count to call a non detail primitive

### LINKED ISSUE #3475: Create a non-detail namespace wrapper for detail::nbr_intersection
https://github.com/rapidsai/cugraph/issues/3446 Outlined an approach for implementing k-truss which requires use of the nbr_intersection primitive and edge masking.

Create a non-detail method that wraps the detail::nbr_intersection primitive, incorporating edge masking, that can be used to implement k-truss.

jnke2016:
Create a non detail primitive that iterates over each input vertex pair and returns the common destination neighbor list
pair in a CSR-like format.

closes #3475


cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,105 @@
...
+#include <thrust/reduce.h>
+#include <thrust/sort.h>
+#include <thrust/tabulate.h>
+#include <thrust/tuple.h>
+
+#include <type_traits>
+
+namespace cugraph {
+
+namespace detail {
```
ChuckHastings, line 56: Don't need to define an empty detail namespace.

cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,105 @@
...
+#include <cugraph/edge_src_dst_property.hpp>
+#include <cugraph/graph_view.hpp>
+#include <cugraph/utilities/device_functors.cuh>
+#include <cugraph/utilities/error.hpp>
+
+#include <raft/core/handle.hpp>
+
+#include <rmm/exec_policy.hpp>
+
+#include <thrust/binary_search.h>
```
ChuckHastings, line 35: I don't think any of these thrust includes are necessary.

You might want to review all of your includes to reduce them just to what's directly used in this file.

cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,105 @@
+/*
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
ChuckHastings, line 2: This should probably just be 2024.

cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,105 @@
...
+std::tuple<rmm::device_uvector<size_t>, rmm::device_uvector<typename GraphViewType::vertex_type>>
+per_v_pair_dst_nbr_intersection(raft::handle_t const& handle,
+                                GraphViewType const& graph_view,
+                                VertexPairIterator vertex_pair_first,
+                                VertexPairIterator vertex_pair_last,
+                                bool do_expensive_check = false)
+{
+  static_assert(!GraphViewType::is_storage_transposed);
+
+  if (do_expensive_check) {
```
ChuckHastings, line 86: Is there a reason to duplicate this error check?  Couldn't you just pass `do_expensive_check` directly to the `detail::nbr_intersection` and let it do this check (it already does).

cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,105 @@
...
+    auto num_invalids =
+      detail::count_invalid_vertex_pairs(handle, graph_view, vertex_pair_first, vertex_pair_last);
+    CUGRAPH_EXPECTS(num_invalids == 0,
+                    "Invalid input arguments: there are invalid input vertex pairs.");
+  }
+
+  auto [intersection_offsets, intersection_indices] =
```
ChuckHastings, line 93: You should be able to just return the results directly...
```
return detail::nbr_intersection(handle,
...
```
instead of capturing and then immediately returning the result.

ChuckHastings (approved):

cpp/src/prims/per_v_pair_dst_nbr_intersection.cuh
```diff
@@ -0,0 +1,58 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "prims/detail/nbr_intersection.cuh"
+
```
seunghwak, line 19: Add include statements for raft::handle_t, std::tuple, and rmm::device_uvector.

seunghwak (approved):

PR #4632, opened 2024-08-26
Update examples to build with latest changes to cugraph

ChuckHastings:
We don't build the examples as part of CI.  Some changes were made to the cugraph functions that cause the examples to be out of date.

Also, we should release the RAFT handle before calling MPI_Finalize.


seunghwak (approved):
LGTM

PR #4634, opened 2024-08-27
Prune workflows based on changed files

KyleFromNVIDIA:
Only run tests based on things that have actually changed. For example, if only Python files have changed, we don't need to run the C++ tests.

Contributes to https://github.com/rapidsai/build-planning/issues/94

Depends on https://github.com/rapidsai/shared-workflows/pull/239


.github/workflows/pr.yaml
```diff
@@ -35,6 +36,58 @@ jobs:
...
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v45
+        with:
+          base_sha: ${{ fromJSON(steps.get-pr-info.outputs.pr-info).base.sha }}
+          files_yaml: |
+            cpp:
+              - '**'
+              - '!CONTRIBUTING.md'
+              - '!README.md'
```
robertmaynard, line 68: I think we can add the following folders for cpp:

- mg_utils
- readme_pages

.github/workflows/pr.yaml
```diff
@@ -105,9 +161,10 @@ jobs:
       extra-repo-deploy-key: CUGRAPH_OPS_SSH_PRIVATE_DEPLOY_KEY
       node_type: cpu32
   wheel-tests-pylibcugraph:
-    needs: wheel-build-pylibcugraph
+    needs: [wheel-build-pylibcugraph, changed-files]
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-24.10
+    if: needs.changed-files.outputs.test_python == 'true'
```
KyleFromNVIDIA, line 167: This change results in the `cugraph` wheel not being built if Python tests aren't needed, because `wheel-build-cugraph` depends on `wheel-tests-pylibcugraph`. Is it OK to skip building the `cugraph` wheel in that case?

notebooks/algorithms/structure/Renumber.ipynb
```diff
@@ -57,7 +57,11 @@
     "import pandas as pd\n",
     "import numpy as np\n",
     "import networkx as nx\n",
-    "from cugraph.structure import NumberMap"
+    "from cugraph.structure import NumberMap\n",
+    "\n",
+    "# Force this notebook to fail\n",
+    "# TODO: Remove this before merging\n",
+    "exit(1)"
```
acostadon, line 64: Should this have been removed?

KyleFromNVIDIA: I put it in for testing. The thing I tested was successful, so I'll remove it.

.github/workflows/pr.yaml
```diff
@@ -35,6 +36,42 @@ jobs:
...
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-24.10
+    if: always()
+    with:
+      needs: ${{ toJSON(needs) }}
+  changed-files:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/changed-files.yaml@branch-24.10
+    with:
+      files_yaml: |
```
jameslamb, line 46: I think `!.devcontainers/**` could also safely be added to these lists. A change that just updates the base image for devcontainers shouldn't require running any of the other tests in this repo.

.github/workflows/pr.yaml
jameslamb: > Is it OK to skip building the cugraph wheel in that case?

I think it is ok, yes.

But only if I've understood how this works correctly.

Seeing these paths in the `test_python` config:

```text
CONTRIBUTING.md
README.md
docs/**
img/**
notebooks/
```

I'm assuming that translates to "if any file was changed that doesn't match those patterns, then run the builds and test".

I cannot think of a case where a PR touching only those files would require rebuilding wheels...given that the notebooks are tested with conda packages, not wheels.

jameslamb (commented):
I'm not up to date on the changed-files work, so my review shouldn't count towards a merge. I left one other comment for your consideration.

I'll get caught up today and could leave a more thoughtful review later today, if you still don't have a `ci-codeowners` approval by then.

.github/workflows/pr.yaml
vyasr: Yes, I think that is OK, although it is a bit incongruous to build pylibcugraph but not cugraph. We could also reorder the jobs to make `wheel-build-cugraph` depend on `wheel-build-pylibcugraph` instead.

.github/workflows/pr.yaml
KyleFromNVIDIA: I've added `!.devcontainers/**`.

.github/workflows/pr.yaml
KyleFromNVIDIA: If nobody has any objections, I'll merge this as-is, and we can consider revisiting the build graph at some point in the future if it becomes a problem.

robertmaynard (approved):
C++ build side rules look good

AyodeAwe (approved):

PR #4639, opened 2024-08-29
Fix build strings in nx-cugraph

bdice:
Build strings aren't being published correctly for `nx-cugraph` (see missing info in current nightlies: https://anaconda.org/rapidsai-nightly/nx-cugraph/files). This is due to a formatting or copy-paste problem in the `meta.yaml` recipe file. This PR fixes the problem.


jameslamb (approved):
oh wow, never noticed this! Thank you.

PR #4619, opened 2024-08-20
[WIP] Improve multi-GPU BFS performance

seunghwak:

PR #4642, opened 2024-08-30
[WIP] Memory footprint reduction in graph creation

seunghwak:
When we create a graph from just edge lists (no vertex lists), we find unique vertices from the given edge lists. This can significantly increase the peak memory requirement under the current implementation.

We currently find unique vertices locally in each GPU. We create a local copy of the edges (for each partition), and use sort& unique to find unique vertices. We can cut memory footprint here by hash based binning, and copying only the edges that belong to the current bin in finding unique vertices that belong to the current bin.

Once we find locally unique vertices, we need to shuffle the unique vertices to different GPUs. In each GPU, multiple locally unique vertex vectors are gathered, and there can be multiple duplicate vertices across the vectors. These duplicates can significantly increase the peak memory usage especially when the average vertex degree is low and the number of GPUs is large. This PR performs shuffling separately for different bins keep the temporary memory usage as `edge list size` * `max(1/# bins, 2/# local edge partitions)`.

We can control # bins to trade-off computing (we need to scan edge list # bins times) vs peak memory usage.

### DIFF
--- a/cpp/include/cugraph/partition_manager.hpp
+++ b/cpp/include/cugraph/partition_manager.hpp
@@ -42,7 +42,8 @@ class partition_manager {
   // partitioning along the major axis (major sub-communicator is responsible for this) and along
   // the minor axis (minor sub-communicator is responsible for this). This variable controls whether
   // to map the major sub-communicator to the GPU row communicator or the GPU column communicator.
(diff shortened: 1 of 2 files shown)


seunghwak:
This PR is merged into PR #4619

PR #4647, opened 2024-09-04
Add support for Python 3.12

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/40

This PR adds support for Python 3.12.

## Notes for Reviewers

This is part of ongoing work to add Python 3.12 support across RAPIDS.
It temporarily introduces a build/test matrix including Python 3.12, from https://github.com/rapidsai/shared-workflows/pull/213.

A follow-up PR will revert back to pointing at the `branch-24.10` branch of `shared-workflows` once all
RAPIDS repos have added Python 3.12 support.

### This will fail until all dependencies have been updates to Python 3.12

CI here is expected to fail until all of this project's upstream dependencies support Python 3.12.

This can be merged whenever all CI jobs are passing.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@python-3.12
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -38,7 +38,7 @@ jobs:
   python-build:
     needs: [cpp-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@python-3.12
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -47,7 +47,7 @@ jobs:
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@python-3.12
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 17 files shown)


jameslamb:
Restarted CI here now that there are Python 3.12 packages for `raft`: https://github.com/rapidsai/raft/pull/2428

Will put this up for review once I see enough of CI passing.

python/cugraph/cugraph/testing/resultset.py
```diff
@@ -108,7 +109,11 @@ def load_resultset(resultset_name, resultset_download_url):
         if not compressed_file_path.exists():
             urllib.request.urlretrieve(resultset_download_url, compressed_file_path)
         tar = tarfile.open(str(compressed_file_path), "r:gz")
-        tar.extractall(str(curr_resultset_download_dir))
+        # TODO: pass filter="fully_trusted" when minimum supported Python version >=3.12
+        #  ref: https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter
+        with warnings.catch_warnings():
```
jameslamb, line 114: Starting with Python 3.12, using `tar.extractall()` without passing the `filter` argument raises this `DeprecationWarning`:

```text
DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior.
```

That warning is treated as an error for CI here, so I added this warning filter to allow these tests to pass.

([build link](https://github.com/rapidsai/cugraph/actions/runs/10781109765/job/29911632066))

bdice (approved):

.github/workflows/build.yaml
```diff
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-24.10
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@python-3.12
```
rlratzel, line 31: just curious if this is a new naming convention (going from RAPIDS branch name to python version) or just a temporary change for python 3.12

rlratzel (approved):
Thanks!  LGTM, just one question which need not hold up approval.

.github/workflows/build.yaml
jameslamb: Fair question! Nope this is only temporary.

The process I'm using is like this:

* put up CI changes on a new `shared-workflows` branch: https://github.com/rapidsai/shared-workflows/pull/213
* migrate all the repos to using those new CI configs one-by-one (with PRs like this one)
* once everything is migrated, merge those changes into `shared-workflows`
* change all the repos back to referencing a `branch-{YY}.{MM}` branch

This way, we can do the work incrementally. Instead of, say, adding Python 3.12 jobs that then all RAPIDS repos have to immediately become compliant with.

PR #4649, opened 2024-09-10
Refactor the python function symmetrizing the edgelist

### LINKED ISSUE #4588: Remove deprecation warning for multigraph edges / refactor symmetrization
The removal of multi edges is no longer supported in `symmetrize.py` for both single and multi GPU workflows. Instead, multi edges can be removed at the graph creation through the flag [drop_multi_edges](https://github.com/rapidsai/cugraph/blob/branch-24.10/python/cugraph/cugraph/structure/graph_implementation/simpleGraph.py#L307).

jnke2016:
This PR updates the python API to symmetrize the edge list through the CAPI for PLC algorithms. This PR also deprecates the legacy python function symmetrizing the edge list 

closes #4588


cpp/include/cugraph_c/graph_functions.h
```diff
@@ -361,6 +361,30 @@ cugraph_type_erased_device_array_view_t* cugraph_degrees_result_get_out_degrees(
...
+ *                                be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t symmetrize_edgelist(const cugraph_resource_handle_t* handle,
+                                         const cugraph_type_erased_device_array_view_t* src,
+                                         const cugraph_type_erased_device_array_view_t* dst,
+                                         const cugraph_type_erased_device_array_view_t* weights,
+                                         bool_t reciprocal,
+                                         cugraph_induced_subgraph_result_t** result,
+                                         cugraph_error_t** error);
```
seunghwak, line 386: Aren't we planning to remove this function from the python API? Should we expose this in the C-API?

seunghwak (approved):
Reviewed the C part, LGTM (besides one question)

cpp/include/cugraph_c/graph_functions.h
ChuckHastings: Agreed, I thought we decided this was not going to be exposed via the C API.

cpp/include/cugraph_c/graph.h
```diff
@@ -150,6 +152,9 @@ cugraph_error_code_t cugraph_graph_create_sg(
  *    If false, do not renumber.  Renumbering enables some significant optimizations within
  *    the graph primitives library, so it is strongly encouraged.  Renumbering is required if
  *    the vertices are not sequential integer values from 0 to num_vertices.
+ * @param [in]  symmetrize     If true, symmetrize the edgelist.
+ *    or take the maximum weight), the caller should remove specific edges themselves and not rely
```
ChuckHastings, line 156: This comment doesn't read correctly, looks like we didn't get the entire comment from wherever it was copied.

We should add in the assumption that only weights can be symmetrized currently... if you have `edge_ids` or `edge_types` you can't specify symmetrize as `true`.

cpp/include/cugraph_c/graph.h
```diff
@@ -190,6 +196,9 @@ cugraph_error_code_t cugraph_sg_graph_create_from_csr(
  *    If false, do not renumber.  Renumbering enables some significant optimizations within
  *    the graph primitives library, so it is strongly encouraged.  Renumbering is required if
  *    the vertices are not sequential integer values from 0 to num_vertices.
+ * @param [in]  symmetrize     If true, symmetrize the edgelist.
```
ChuckHastings, line 199: Same as above.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -207,6 +210,22 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
                                         : false);
       }
 
+      if (symmetrize_) {
```
ChuckHastings, line 213: In addition to this logic, if `symmetrize_` is true then we should override the graph properties to set `is_symmetric` to true.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -398,6 +420,22 @@ struct create_graph_csr_functor : public cugraph::c_api::abstract_functor {
         cugraph::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>,
         edge_type_id_t>(handle_);
 
+      if (symmetrize_) {
```
ChuckHastings, line 423: Same comment as above regarding `is_symmetric`

cpp/src/c_api/graph_sg.cpp
```diff
@@ -398,6 +420,22 @@ struct create_graph_csr_functor : public cugraph::c_api::abstract_functor {
...
+        if (edgelist_edge_ids || edgelist_edge_types) {
```
ChuckHastings, line 424: We should add an issue to discuss what symmetrization means with respect to edge properties.

ChuckHastings (commented):
A few things to address.

jnke2016:
> Agreed, I thought we decided this was not going to be exposed via the C API.

Right we are not exposing it to the python API. I forgot to remove it from the CAPI as well

cpp/include/cugraph_c/graph.h
jnke2016: I removed the unrelated comments. I also update the doc strings to reflect that the symmetrization of edges with edge and type ids is currently not supported. In the CAPI, I return an `unsupported` error if either or is specified

cpp/src/c_api/graph_sg.cpp
jnke2016: I tracked this in an [issue](https://github.com/rapidsai/cugraph/issues/4675)

python/cugraph/cugraph/components/connectivity.py
```diff
@@ -287,6 +287,8 @@ def strongly_connected_components(
         "strongly_connected_components", G, directed, connection, return_labels
     )
 
+    print("G = ", G)
+
```
rlratzel, line 291: Oops

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -174,6 +175,12 @@ def from_cudf_edgelist(
 
             This parameter is deprecated and will be removed.
 
+        symmetrize: bool, optional (default=None)
+            If True, symmetrize the edge list for an undirected graph. Setting
+            this flag to True for a directed graph returns an error. The default
+            behavior symmetrizes the edges if the graph is undirected. If the edges
+            are already symmetric, set this flag to False.
```
rlratzel, line 182: Should you also mention the restriction on edgelists with edge IDs and edge types?  Also, the way this is worded makes it seem like the user must figure out if the edgrs are symmetric or not and set the flag accordingly.  Maybe it would be better if it was worded like this:
```suggestion
            behavior symmetrizes the edges if the graph is undirected. This flag cannot be set to True if the edgelist contains edge IDs or edge Types.  If the incoming edgelist is intended for an undirected graph and it is known to be symmetric, this flag can be set to False to skip the symmetrization step for better performance.
```
Same suggestion for `from_cudf_adjlist` below.

python/cugraph/cugraph/utilities/nx_factory.py
```diff
@@ -117,8 +119,10 @@ def convert_from_nx(
     """
 
     if isinstance(nxG, nx.classes.digraph.DiGraph):
+        print("creating a directed graph")
         G = cugraph.Graph(directed=True)
     elif isinstance(nxG, nx.classes.graph.Graph):
+        print("creating an undirected graph")
```
rlratzel, line 125: more leftover debug prints.

rlratzel (changes requested):
Thanks!  Just a few requests below.

python/cugraph/cugraph/tests/structure/test_multigraph.py
```diff
@@ -76,7 +76,7 @@ def test_Graph_from_MultiGraph(graph_file):
 
     G = cugraph.Graph(GM)
     Gnx = nx.Graph(GnxM)
-    assert Gnx.number_of_edges() == G.number_of_edges()
+    assert Gnx.number_of_edges() == G.number_of_edges(True)
```
rlratzel, line 79: Can you add the kwarg name here so it's more obvious?
```suggestion
    assert Gnx.number_of_edges() == G.number_of_edges(directed_edges=True)
```

rlratzel (changes requested):
Thanks, almost done, still see a debug print and I had one minor request.

python/cugraph/cugraph/tests/structure/test_multigraph.py
jnke2016: Right

python/cugraph/cugraph/structure/graph_classes.py
jnke2016: Got it. I reworded it as you suggested

rlratzel (approved):

PR #4651, opened 2024-09-11
reduce pip verbosity in wheel builds

jameslamb:
Proposes reducing the verbosity of `pip wheel` called in wheel builds from `-vvv` to `-v`.

This eliminates the 1000s of lines like this in CI logs:

image

to hopefully make it easier to view those logs interactively


jameslamb:
Had to kick off another CI run because this was too old. I hope this can be merged, I think it's non-controversial and will make CI a bit easier to use.

vyasr (approved):

PR #4652, opened 2024-09-17
Update flake8 to 7.1.1.

bdice:
We need to update flake8 to fix a false-positive that appears with older flake8 versions on Python 3.12.


alexbarghi-nv (approved):
👍

KyleFromNVIDIA (approved):

PR #4653, opened 2024-09-17
Fix `from_pandas_edgelist` with duplicate edges

eriknw:
When creating a `Graph` or `DiGraph`, duplicate edges should be dropped. Also, ensure arrays are contiguous when creating the PLC graph, which can lead to incorrect results.

For now, this was done by splitting out functionality from `relabel_nodes`, but I wonder if there could be a faster way to deduplicate when edge values also need handled.

Also, `cp.unique` was sometimes problematic for me on larger data sets, but I may have been running out of device memory. @rlratzel I'm curious how this will perform for you.

### DIFF
--- a/python/nx-cugraph/_nx_cugraph/__init__.py
+++ b/python/nx-cugraph/_nx_cugraph/__init__.py
@@ -297,8 +297,8 @@ def get_info():
 
 
 def _check_networkx_version():
-    import warnings
     import re
+    import warnings
 
     import networkx as nx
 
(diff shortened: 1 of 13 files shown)


eriknw:
I added a workaround for https://github.com/cupy/cupy/issues/8326: we now use `np.unique` when `axis=` is not None.

rlratzel:
Thanks again for working on this, @eriknw .  I'm going to close it now since [this fix](https://github.com/rapidsai/cugraph/pull/4658) has been merged, but feel free to re-open or create a new PR if other changes in this PR need to be done for 24.10.

PR #4657, opened 2024-09-19
Remove NumPy <2 pin

hcho3:
Continue https://github.com/rapidsai/cugraph/pull/4615


jameslamb:
Hey @hcho3 thanks for working on this! Would you consider continuing this work by just pushing to #4615 instead? That way, all the prior discussion about the code changes would be in the same place as the changes themselves.

Now that you have write access to this repo, you should be able to push to PR branches from forks. And I'm confident @seberg will be ok with it, as that's how he, @jakirkham , and I have been collaborating on the other NumPy 2.x PRs like https://github.com/rapidsai/cuspatial/pull/1441

PR #4641, opened 2024-08-30
[FEA] Implement Link Prediction in cuGraph-PyG

alexbarghi-nv:
Updates the distributed sampler and cuGraph-PyG samplers and loaders to support link prediction.  Currently does not support negative sampling.  Resolves an issue where very small minibatches were interpreted as COO instead of CSR.

Also fixes a bug where the cuGraph-PyG `GraphStore` object calculated the number of vertices incorrectly for homogeneous graphs, which was encountered when testing link prediction on small batches.

Merge after #4628 
Closes #2992

PR #4660, opened 2024-09-20
Support Negative Sampling in pylibcugraph and cuGraph-PyG

alexbarghi-nv:
Adds support for negative sampling in `pylibcugraph` and `cugraph-pyg`.

Closes rapidsai/cugraph-gnn#39
Merge after #4641


ChuckHastings (approved):
Approving C API updates.

python/cugraph-pyg/cugraph_pyg/examples/rgcn_link_class_mnmg.py
```diff
@@ -0,0 +1,418 @@
...
+    finalize as wm_finalize,
+)
+
+
+# Enable cudf spilling to save gpu memory
+from cugraph.testing.mg_utils import enable_spilling
+
+# Ensures that a CUDA context is not created on import of rapids.
+# Allows pytorch to create the context instead
+os.environ["RAPIDS_NO_INITIALIZE"] = "1"
```
rlratzel, line 49: I'm assuming you tested this but I'm surprised this works since it comes *after* various cugraph imports above.  Would it be safer to move up, in case something in cugraph is refactored to pull in a module that creates a CUDA context?

rlratzel (approved):
I'm not familiar with this code and the PR is quite large, but I didn't see anything that jumped out as a problem.  I did have one question/suggestion though which need not hold up approval.

python/cugraph-pyg/cugraph_pyg/examples/rgcn_link_class_mnmg.py
alexbarghi-nv: You're right, this got pulled in from a similar file and ended up in the right place.  I'll change it.

alexbarghi-nv:
> I'm not familiar with this code and the PR is quite large, but I didn't see anything that jumped out as a problem. I did have one question/suggestion though which need not hold up approval.

This PR includes the other PRs (in-memory sampling, link prediction) which is why there are a lot of changes.  But if you've reviewed everything we could merge this in place of the other ones.

PR #4661, opened 2024-09-20
bump NCCL floor to 2.18.1.1, include nccl.h where it's needed

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/102

Some RAPIDS libraries are using `ncclCommSplit()`, which was introduced in `nccl==2.18.1.1`. This is part of a series of PRs across RAPIDS updating libraries' pins to `nccl>=2.18.1.1` to ensure they get a new-enough version that supports that.

### DIFF
--- a/conda/environments/all_cuda-118_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-118_arch-x86_64.yaml
@@ -33,7 +33,7 @@ dependencies:
 - libraft==24.10.*,>=0.0.0a0
 - librmm==24.10.*,>=0.0.0a0
 - nbsphinx
-- nccl>=2.9.9
+- nccl>=2.18.1.1
 - networkx>=2.5.1
 - networkx>=3.0
 - ninja
--- a/conda/environments/all_cuda-125_arch-x86_64.yaml
+++ b/conda/environments/all_cuda-125_arch-x86_64.yaml
@@ -39,7 +39,7 @@ dependencies:
 - libraft==24.10.*,>=0.0.0a0
 - librmm==24.10.*,>=0.0.0a0
 - nbsphinx
-- nccl>=2.9.9
+- nccl>=2.18.1.1
 - networkx>=2.5.1
 - networkx>=3.0
 - ninja
--- a/conda/recipes/libcugraph/conda_build_config.yaml
+++ b/conda/recipes/libcugraph/conda_build_config.yaml
@@ -17,7 +17,7 @@ doxygen_version:
   - ">=1.8.11"
 
 nccl_version:
-  - ">=2.9.9"
+  - ">=2.18.1.1"
 
 c_stdlib:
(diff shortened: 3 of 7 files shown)


vyasr (changes requested):
cugraph also ought to be able to rely on raft for its NCCL handling rather than doing anything on its own AFAICT. It does use some raft APIs that accept NCCL types directly as arguments, but I still think that's OK.

jameslamb:
I'm not convinced that `cugraph` could/should drop its NCCL dependency and only get it transitively through raft.

It has some direct uses, like these:

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/include/cugraph/mtmg/instance_manager.hpp#L52

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/include/cugraph/mtmg/resource_manager.hpp#L200

As well as others in tests and examples:

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/tests/mtmg/threaded_test.cu#L41

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/tests/mtmg/threaded_test_jaccard.cu#L41

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/tests/mtmg/threaded_test_louvain.cu#L41

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/examples/developers/graph_operations/CMakeLists.txt#L28

https://github.com/rapidsai/cugraph/blob/7e058e2961b5b8be05877282d26051f36dd73c53/cpp/examples/developers/graph_operations/CMakeLists.txt#L38

Those look like direct-enough uses that `cugraph` should have its own pinnings, in my opinion.

vyasr:
I agree, those are direct uses. In that case, though, the files should be including nccl.h rather than relying on it being transitively included by raft's std_comms.h.

ChuckHastings (approved):

jameslamb:
> the files should be including nccl.h rather than relying on it being transitively included by raft's std_comms.h.

Agreed! Pushed https://github.com/rapidsai/cugraph/pull/4661/commits/7fca13ffb6ad20230c478a0009e3cc504183f7dc adding that.

Maybe in the future we can explore using `include-what-you-use` (https://github.com/include-what-you-use/include-what-you-use) in CI. I found from search that you'd tried that last year: https://github.com/rapidsai/cudf/issues/581#issuecomment-1362133285, maybe at some point we can look into it again.

vyasr (approved):

jakirkham (approved):

PR #4677, opened 2024-09-27
Add `nx-cugraph` introduction notebook to repo

nv-rliu:
## Proposed Changes

This PR adds an introduction notebook to the `notebooks/demo` directory of the repository. Click the link to view the files directory in the dev branch.
 * [The introduction to `nx-cugraph` notebook](https://github.com/rapidsai/cugraph/blob/1a70b7ad588bcc1dc6f65af8fe07149ed8083ef0/notebooks/demo/accelerating_networkx.ipynb)


eriknw (commented):
I think these are great! I suggested a fix for a couple typos, and left a couple minor questions or nits. The only thing blocking approval is the "TODO" remaining in a notebook.

notebooks/demo/accelerating_networkx.ipynb
```diff
@@ -0,0 +1,614 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "3atL3tI0frYm"
+      },
+      "source": [
+        "## Betweenness Centrality\n",
+        "[Betweenness Centrality](https://en.wikipedia.org/wiki/Betweenness_centrality) is a graph algorithm that computes a centrality score for each node (`v`) based on how many of the shortest paths between pairs of nodes in the graph pass through `v`. A higher centrality score represents a node that \"connects\" other nodes in a network more than that of a node with a lower score.\n",
```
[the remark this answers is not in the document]
nv-rliu: IMO from the POV of someone who isn't very familiar with graph algorithms, the Wikipedia article does provide a better explanation of the algorithm and its usage. That's just my two cents.

notebooks/demo/accelerating_networkx.ipynb
```diff
@@ -0,0 +1,614 @@
...
+        "nxcg_bc_results = nx.betweenness_centrality(G)"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "z1hxut3GTj5A"
+      },
+      "source": [
+        "You may have noticed that using the `nx-cugraph` backend resulted in a slightly slower execution time. This is not surprising when working with a graph this small, since the overhead of converting the graph for the first time and launching the algorithm kernel on the GPU is actually significantly more than the computation time itself.  We'll see later that this overhead is negligeble when compared to the time saved when running on a GPU for larger graphs.\n",
```
eriknw, line 195: `negligeble` -> `negligible`
```suggestion
        "You may have noticed that using the `nx-cugraph` backend resulted in a slightly slower execution time. This is not surprising when working with a graph this small, since the overhead of converting the graph for the first time and launching the algorithm kernel on the GPU is actually significantly more than the computation time itself.  We'll see later that this overhead is negligible when compared to the time saved when running on a GPU for larger graphs.\n",
```

notebooks/demo/accelerating_networkx.ipynb
```diff
@@ -0,0 +1,614 @@
...
+      "metadata": {
+        "id": "lyYF0LbtFwjh"
+      },
+      "outputs": [],
+      "source": [
+        "# The locale encoding may have been modified from the plots above, reset here to run shell commands\n",
+        "import locale\n",
+        "locale.getpreferredencoding = lambda: \"UTF-8\"\n",
+        "!wget https://data.rapids.ai/cugraph/datasets/cit-Patents.csv  # Skip if cit-Patents.csv already exists.\n",
+        "# !wget https://snap.stanford.edu/data/cit-Patents.txt.gz  # Skip if cit-Patents.txt.gz already exists."
```
eriknw, line 392: Why do we have this line if it's commented out?

notebooks/demo/accelerating_networkx.ipynb
```diff
@@ -0,0 +1,614 @@
...
+        "                names=[\"src\", \"dst\"],\n",
+        "                dtype=\"int32\",\n",
+        ")\n",
+        "# df = pd.read_csv(\"cit-Patents.txt.gz\",\n",
+        "#                  compression=\"gzip\",\n",
+        "#                  skiprows=4,\n",
+        "#                  sep=\"\\t\",\n",
+        "#                  names=[\"src\", \"dst\"],\n",
+        "#                  dtype=\"int32\",\n",
+        "# )"
```
eriknw, line 427: And here?

eriknw (changes requested):
Almost there... a couple minor nits.

notebooks/demo/accelerating_networkx.ipynb
nv-rliu: It was to give the users the option to download the compressed dataset directly from the SNAP source. I believe it was relevant in our recording as an option.

notebooks/demo/accelerating_networkx.ipynb
nv-rliu: Same as the `wget` command. Users are given the option to directly load in the compressed dataset from the original source if they desire.

eriknw (approved):

notebooks/demo/accelerating_networkx.ipynb
```diff
@@ -0,0 +1,614 @@
...
+        "!pip install nx-cugraph-cu12 --extra-index-url=https://pypi.nvidia.com"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "0h1K-7tI_AZH"
+      },
+      "source": [
+        "Of course, we'll also be using `networkx`, which is already provided in the Colab environment. This notebook will be using features added in version 3.3, so we'll import it here to verify we have a compatible version."
```
acostadon, line 76: You could mention that the feature added inversion 3.3 is the networkx.config package which is referenced here.
https://networkx.org/documentation/stable/reference/backends.html#module-networkx.utils.configs

acostadon (approved):
Looks great..only 2 recommendations, neither of which should delay merging.
1) I commented on adding the link to the NetworkX config package.
https://networkx.org/documentation/stable/reference/backends.html#module-networkx.utils.configs

2) It would be nice to have a link to to documentation of the wikipedia dataset. I looked but couldn't find it. It would enable users to experiment with the notebook.

notebooks/demo/accelerating_networkx.ipynb
nv-rliu: In the interest of getting these merged, I think it would be best to add these in a future PR. I'll make note of these, thak you.

rlratzel (approved):
Thanks.  It's too bad it's not using new NX 3.4 features, but this way it'll work with old and new versions.

PR #4685, opened 2024-10-01
`nx-cugraph`: add `NX_CUGRAPH_AUTOCONFIG=True` env var to enable full zero-code change

eriknw:
This is for convenience and sets or updates `NETWORKX` environment variables.

Do we like `NX_CUGRAPH` as the env var name? What should we consider a true value: "True" (case-insensitive) or a non-empty value?

This works with the latest dev version of NetworkX. I have not yet tried it with older NetworkX versions.


rlratzel (changes requested):
Thanks, I think the choices of what/how to update here is good, especially the (untested) support for older versions.

I do have a few requests, questions, and concerns:
* Since this has a lot of potentially surprising side effects, I think the env var name should be less ambiguous (as you also wondered).  I'd say `NX_CUGRAPH_AUTOCONFIG` is much more self-documenting.
* I think the value should be `True`/`true`/`TRUE` instead of just non-empty.  I think that matches NetworkX too?  Either way, I'd hate someone to set it to `False` and have it evaluate to True because it's non-empty.
* When does this get applied relative to NetworkX processing its env vars?  If/when a user sets this and the standard NX vars, what wins, and is that expected?
* concern: I hope this doesn't set a precedent for other backends, since I can already see multiple backends with their own presets clobber each other in fun and mysterious ways. I hope this shows that these can be the expected defaults and we can get back to one env var: `NETWORKX_BACKEND_PRIORITY`.
* Unfortunately, I think we need this preset.  Early feedback from new users helping with docs indicate that the 3 vars NX now requires for full ZCC is *this close* :pinching_hand:  to being a non-starter.

eriknw:
Thanks for the quick feedback.

> * Since this has a lot of potentially surprising side effects, I think the env var name should be less ambiguous (as you also wondered).  I'd say `NX_CUGRAPH_AUTOCONFIG` is much more self-documenting.

:+1: I updated it to that. I want "AUTO" in the variable name as well.

> * I think the value should be `True`/`true`/`TRUE` instead of just non-empty.  I think that matches NetworkX too?  Either way, I'd hate someone to set it to `False` and have it evaluate to True because it's non-empty.

Yeah, I think this generally matches networkx.

> * When does this get applied relative to NetworkX processing its env vars?  If/when a user sets this and the standard NX vars, what wins, and is that expected?

`nx-cugraph`'s `get_info` (which updates the env vars) gets called here:
https://github.com/networkx/networkx/blob/74af3e63d21406680ae21a60975135472d3ca412/networkx/utils/backends.py#L415 which is before handling of the environment variables here:
https://github.com/networkx/networkx/blob/74af3e63d21406680ae21a60975135472d3ca412/networkx/utils/backends.py#L475

> * concern: I hope this doesn't set a precedent for other backends, since I can already see multiple backends with their own presets clobber each other in fun and mysterious ways. I hope this shows that these can be the expected defaults and we can get back to one env var: `NETWORKX_BACKEND_PRIORITY`.

I think such an "easy-mode" environment variable should only be expected to be useful when using a single backend. I think we can get a lot of mileage out of this. But, yeah, if multiple backends start doing the same thing, clobbering and undefined behavior become possible unless the backends can coordinate. Let's see how it goes.

> * Unfortunately, I think we need this preset.  Early feedback from new users helping with docs indicate that the 3 vars NX now requires for full ZCC is _this close_ 🤏  to being a non-starter.

Yeah, one environment variable is nice.

rlratzel (approved):

nv-rliu (approved):

eriknw:
I wonder whether (and when) it might make sense to introduce a RAPIDS-wise environment variable (such as `RAPIDS_MAKE_THINGS_GO_BRRR=True`) to automatically accelerate multiple libraries. We could support it.

jakirkham:
Updated now that PR ( https://github.com/rapidsai/cugraph/pull/4690 ) has merged

PR #4687, opened 2024-10-01
[FEA] Support Edge ID Lookup in PyLibcuGraph

alexbarghi-nv:
Support Edge ID lookup in `pylibcugraph`.  Also fixes some bugs in the C API (i.e. lookup table not being cleaned up correctly, container being incorrectly dereferenced as graph).

Verified in rapidsai/cugraph-gnn#50


ChuckHastings (approved):
Thanks for catching these!

seunghwak (approved):
LGTM

rlratzel (commented):
The Python/Cython changes look fine, just a couple of questions/requests:
* Would it be better to be more explicit and replace the word "container" with "map", since (I'm assuming) it has to be a map?
* Can you add a smoketest of some kind for the PLC updates?  This would also be great for users looking for an example of how to use the new APIs.

alexbarghi-nv:
> * Would it be better to be more explicit and replace the word "container" with "map", since (I'm assuming) it has to be a map?

We're calling it `EdgeIdLookupTable` at the Python level.  That should be ok, right?

rlratzel (approved):
Looks great, thanks @alexbarghi-nv !

PR #4690, opened 2024-10-03
Constrain versions of PyTorch and CI artifacts in CI Runs, upgrade to dgl 2.4

alexbarghi-nv:
We were pulling the wrong packages because the PyTorch version constraint wasn't tight enough.  Hopefully these sorts of issues will be resolved in the `cugraph-gnn` repository going forward, where we can pin a specific pytorch version for testing.

PR #4701, opened 2024-10-07
combine pip install calls in wheel-testing scripts

jameslamb:
## Summary

Follow-up to #4690.

Proposes consolidating stuff like this in CI scripts:

```shell
pip install A
pip install B
pip install C
```

Into this:

```shell
pip install A B C
```

## Benefits of these changes

Reduces the risk of creating a broken environment with incompatible packages. Unlike `conda`, `pip` does not evaluate the requirements of all installed packages when you run `pip` install.

Installing `torch` and `cugraph-dgl` at the same time, for example, gives us a chance to find out about packaging issues like *"`cugraph-dgl` and `torch` have conflicting requirements on `{other_package}`"* at CI time.

Similar change from `cudf`: https://github.com/rapidsai/cudf/pull/16575


KyleFromNVIDIA (approved):

alexbarghi-nv (approved):
This is fine with me.  I agree we should have already been doing this.

jameslamb:
Ok cool, thank you both!

PR #4702, opened 2024-10-07
New notebook to use full patent repository data

acostadon:
Creates a new notebook intended for a data science toolkit exercise.


rlratzel:
Hey @acostadon , the latest version of nx-cugraph includes some features which improve the UX and we'd like to include them in this notebook. For example, there's a new env var (`NX_CUGRAPH_AUTOCONFIG`) which simplifies how a user configures the latest version NetworkX for optimum performance, and we probably no longer need to mention things like caching quite as prominently.  We'll also be adding this notebook to the new [nx-cugraph repo](https://github.com/rapidsai/nx-cugraph) now that it's set up.  Can you move this PR to `Draft` and we can work together offline to update it?

acostadon:
I moved it back to draft

acostadon:
needs to be moved to nx-cugraph repository.

PR #4705, opened 2024-10-08
Update `nx-cugraph` README with New Env Variable

nv-rliu:
This PR updates the nx-cugraph README with the new `NX_CUGRAPH_AUTOCONFIG` variable.

### DIFF
--- a/python/nx-cugraph/README.md
+++ b/python/nx-cugraph/README.md
@@ -1,7 +1,7 @@
 # nx-cugraph
 
 ## Description
-[RAPIDS](https://rapids.ai) nx-cugraph is a [backend to NetworkX](https://networkx.org/documentation/stable/reference/utils.html#backends)
+[RAPIDS](https://rapids.ai) nx-cugraph is a [backend to NetworkX](https://networkx.org/documentation/stable/backends.html)
 to run supported algorithms with GPU acceleration.
 
 ## System Requirements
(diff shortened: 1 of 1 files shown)


rlratzel (approved):

python/nx-cugraph/README.md
```diff
@@ -45,18 +45,18 @@ Notes:
 NetworkX will use nx-cugraph as the graph analytics backend if any of the
 following are used:
 
-### `NETWORKX_AUTOMATIC_BACKENDS` environment variable.
-The `NETWORKX_AUTOMATIC_BACKENDS` environment variable can be used to have NetworkX automatically dispatch to specified backends an API is called that the backend supports.
-Set `NETWORKX_AUTOMATIC_BACKENDS=cugraph` to use nx-cugraph to GPU accelerate supported APIs with no code changes.
+### `NX_CUGRAPH_AUTOCONFIG` environment variable.
+The `NX_CUGRAPH_AUTOCONFIG` environment variable can be used to have NetworkX automatically dispatch to specified backends if an API is called that the backend supports.
+Set `NX_CUGRAPH_AUTOCONFIG=True` to use nx-cugraph to GPU accelerate supported APIs with zero code change.
```
eriknw, line 50: I think we should update the language to say this specifically automatically dispatches to nx-cugraph. Previously, this section discussed `NETWORKX_AUTOMATIC_BACKENDS` which is more generic.

Also, do you think we should share _how_ this works? That is, it works by setting or updating environment variables that control the config: https://networkx.org/documentation/stable/reference/backends.html . It's also fine if we want to keep the README minimal, but someday we may want a reference page that explains how.

eriknw (commented):
Looks pretty good, thanks @nv-rliu!

Can we also change the "backend to NetworkX" link at the top to point to
- https://networkx.org/documentation/stable/reference/backends.html
or
- https://networkx.org/documentation/stable/backends.html ?

The previous link is stale.

python/nx-cugraph/README.md
nv-rliu: I just updated the README based on your feedback.

I updated the URL at the top to -> https://networkx.org/documentation/stable/backends.html

And for the _environment variable_ section, I updated the description on `NX_CUGRAPH_AUTOCONFIG` and linked to the Backends page so users can read more on how backends work.

nv-rliu: See new lines 49-51

PR #4718, opened 2024-10-14
Remove deprecated C API functions for graph creation/graph free

ChuckHastings:
Prep work for temporal, cleaning up some old functions that we deprecated a while back.


seunghwak (approved):
LGTM

rlratzel (approved):
Cython changes LGTM.  If CI passes, then I'll assume these removals are safe, since I'm not aware of anyone else using those functions.

PR #4720, opened 2024-10-14
Don't compile int32_t/int64_t vertex_t/edge_t combinations

ChuckHastings:
To reduce compile time and library size... we're dropping pre-compiled support for vertex_t as int32_t and edge_t as int64_t.  This is a small edge case in memory utilization.

The data structures and primitives would still support such a configuration, but we're not compiling it into libcugraph.so with this PR. 

Seems to improve C++ compile time by 25% on my dgx18 and a 30% reduction in the size of libcugraph.so.


seunghwak (approved):
LGTM and thanks!!!

seunghwak (approved):

PR #4724, opened 2024-10-16
Streamline graph creation in C API to allow for easier addition of new parameters

ChuckHastings:
This PR creates an edgelist object that will contain the COO/CSR information along with all edge properties to support graph creation.  This will allow us to add new features (and new edge attributes) with a smaller impact on the C API footprint.  It should also simplify the API a bit.

The PR also marks the old methods as deprecated so that we can phase them out.


ChuckHastings:
Superseded by #5161

PR #4721, opened 2024-10-15
Update `nx-cugraph` Benchmarks for Additional BC K-Values

nv-rliu:
Part of https://github.com/rapidsai/graph_dl/issues/597

## Proposed Changes:
 - This PR adds additional k-values for `bench_algos.py::bench_betweenness_centrality`. 
 - These k-values can be selected/turned off by using `pytest -k`. When using `run-main-benchmarks.sh`, there is a `bc-k-values` string that can be edited to perform the selection. 
 - `README` update

PR #4726, opened 2024-10-16
Adds new dataset for benchmarking in the 100k node 1M edge range, adds additional k-values for BC benchmarks

rlratzel:
Adds new dataset for benchmarking in the 100k node 1M edge range.
This also updates the benchmark fixture to download the dataset instead of requiring a separate script.
This also includes changes from #4721 to add additional k-values for BC benchmarks to run. 

The addition was done with an inline yaml file and a followup PR will be done to properly add this dataset to the cugraph.datasets API.  The inline yaml was done in order to allow benchmarks to run on a system with an existing cugraph installation by simply copying this bench_algos.py, rather than require an updated cugraph install with the new dataset metadata.


nv-rliu (approved):
Looks good to me

acostadon (approved):
Looks good

PR #4736, opened 2024-10-22
Updates README with new dataset, removes mention of script no longer used

rlratzel:
This is a doc-only PR which updates the README for nx-cugraph pytest-based benchmarks:
* Adds description of a dataset that's being used
* Removes mention of script no longer used


acostadon (approved):
Looks good

PR #4737, opened 2024-10-23
Delete the deprecated data_type_id_t enum

ChuckHastings:
Prep work for temporal graphs.  We renamed `data_type_id_t` to `cugraph_data_id_t` about 8 months ago, this finishes the transition.  We also split the types out into their own file in the C API.

Marked as breaking since we are deleting a deprecated type from the C API.


cpp/include/cugraph_c/types.h
```diff
@@ -0,0 +1,41 @@
...
+extern "C" {
+#endif
+
+typedef enum bool_ { FALSE = 0, TRUE = 1 } bool_t;
+
+typedef int8_t byte_t;
+
+typedef enum data_type_id_ {
+  INT32 = 0,
+  INT64,
```
seunghwak, line 31: No need to add INT8, INT16, UINT8, UINT16, UINT32, UINT64?

These may become necessary in some algorithms in the future.

seunghwak: If we add these types but don't explicitly instantiate for these types, the cost is pretty minimal, right?

seunghwak (approved):
Besides adding more integer types to be future proof, LGTM.

cpp/include/cugraph_c/types.h
ChuckHastings: We can keep the cost minimal.  I have added them in the next push.

rlratzel (approved):
cython/python changes LGTM, CI should tell us if anything was missed.

PR #4742, opened 2024-11-01
Optimize K-Truss

jnke2016:
This PR introduces several optimization to speed up K-Truss. In fact, our K-Truss implementation computes the intersection of all edges regardless they are weak or not which can be very expensive if only few edges need to be invalidated. By running `nbr_intersection` on the weak edges, this considerably improves the runtime.


cpp/src/community/k_truss_impl.cuh
```diff
@@ -43,6 +44,184 @@
...
+
+    // Extract (p, r) edges
+    auto edge_p_r = thrust::make_tuple(weak_srcs[chunk_start + idx], intersection_indices[i]);
+
+    // check for edge existance in the DODG edgelist- FIXME: Create a function
+    // Check in the valid edge range
+    auto has_edge = thrust::binary_search(
+          thrust::seq, edgelist_first, weak_edgelist_first, edge_p_r);
+    
+    if (!has_edge) { // FIXME: Do binary search instead
```
ChuckHastings, line 125: Is this FIXME still relevant?  It looks like you're doing binary search

cpp/src/community/k_truss_impl.cuh
```diff
@@ -43,6 +44,184 @@
...
+  {
+    return count < k - 2
+             ? thrust::optional<thrust::tuple<vertex_t, vertex_t>>{thrust::make_tuple(src, dst)}
+             : thrust::nullopt;
+  }
+};
+
+
+template <typename vertex_t, typename edge_t>
+struct extract_edges { // FIXME:  ******************************Remove this functor. For testing purposes only*******************
```
ChuckHastings, line 61: Should these be left in here?

cpp/src/community/k_truss_impl.cuh
```diff
@@ -293,63 +475,349 @@ k_truss(raft::handle_t const& handle,
...
 
-      if (prev_number_of_edges == cur_graph_view.compute_number_of_edges(handle)) { break; }
     }
 
-    rmm::device_uvector<vertex_t> edgelist_srcs(0, handle.get_stream());
-    rmm::device_uvector<vertex_t> edgelist_dsts(0, handle.get_stream());
     std::optional<rmm::device_uvector<weight_t>> edgelist_wgts{std::nullopt};
-
+    
+    //#if 0
```
ChuckHastings, line 798: Delete the commented out lines in this section

jnke2016: Right I still had a lot of debug prints which are now removed

ChuckHastings (approved):

seunghwak:
@jnke2016 Is this PR ready for review?

cpp/src/community/k_truss_impl.cuh
```diff
@@ -41,8 +42,175 @@
...
+    graph_view.has_edge(handle,
+                        raft::device_span<vertex_t const>(srcs.data(), srcs.size()),
+                        raft::device_span<vertex_t const>(dsts.data(), dsts.size()));
+
+  // Send the result back
+  std::tie(edge_exists, std::ignore) =
+    shuffle_values(handle.get_comms(), edge_exists.begin(), rx_counts, handle.get_stream());
+
+  // The 'edge_exists' array is ordered based on 'cp_edgelist_srcs' where the edges where group,
+  // hoever it needs to match 'edgelist_srcs', hence re-order 'edge_exists' accordingly.
```
seunghwak, line 126: Is this true? I guess `edge_exists` is ordered based on `edgelist_srcs` not `cp_edgelist_srcs`.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -41,8 +42,175 @@
...
+#include <chrono>
+using namespace std::chrono;
+
 namespace cugraph {
 
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+void order_edge_based_on_dodg(raft::handle_t const& handle,
+                              graph_view_t<vertex_t, edge_t, false, multi_gpu>& graph_view,
+                              raft::device_span<vertex_t> edgelist_srcs,
+                              raft::device_span<vertex_t> edgelist_dsts)
```
seunghwak, line 54: What are you doing in this function?

cpp/src/community/k_truss_impl.cuh
jnke2016: Yes other way around. It is `cp_edgelist_srcs` that is supposed to be groupby_and_count. The documentation is correct but made a mistake in the group_by_and_count call.  I don't want to mess up original edges in the triangles when grouping that's why I make a copy.

cpp/src/community/k_truss_impl.cuh
jnke2016: Essentially this function is ordering the edges obtained from nbr_intersection (On the undirected graph) based on the DODG edges

jnke2016: Once we identify weak edges, the next step is to retrieve the triangles incident to those weak edges and this is done with nbr_intersection (on the symmetric graph). Once we have the three endpoints of the triangle, the next step is to find the direction that matches the edges in the DODG and this is what that function does

cpp/src/community/k_truss_impl.cuh
```diff
@@ -41,7 +42,214 @@
...
+using namespace std::chrono;
+
 namespace cugraph {
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+void order_edge_based_on_dodg(raft::handle_t const& handle,
+                              graph_view_t<vertex_t, edge_t, false, multi_gpu>& graph_view,
+                              raft::device_span<vertex_t> edgelist_srcs,
+                              raft::device_span<vertex_t> edgelist_dsts)
+
+{
```
seunghwak, line 55: So, here, what we are actually doing is this.

For (src, dst) pairs in edgelist_srcs & edgelist_dsts, return (src, dst) if (src, dst) exists in graph_view, and return (dst, src) if not.

I think this code is overly complex.

If single-GPU, this code looks fine.

In multi-GPU, you first need to find unique (src, dst) pairs, shuffle, call has_edge, shuffle_back. Now you have (src, dst, exist) triplets. Sort them for binary search. Then, you can iterate over the input pairs and flip or not based on whether the edge exists or not.

In the current code, you are finding unique edges after shuffling. But better do this at the beginning to reduce computation/communication. And for temporary variables, try to minimize their scopes. Some variables' scopes are too large and this makes code harder to understand.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -41,7 +42,214 @@
...
+      [edge_exists    = edge_exists->data(),
+       edgelist_first = thrust::make_zip_iterator(edgelist_srcs.begin(), edgelist_dsts.begin()),
+       cp_edgelist_first =
+         thrust::make_zip_iterator(cp_edgelist_srcs->begin(), cp_edgelist_dsts->begin()),
+       cp_edgelist_last = thrust::make_zip_iterator(cp_edgelist_srcs->end(),
+                                                    cp_edgelist_dsts->end())] __device__(auto idx) {
+        auto src = thrust::get<0>(edgelist_first[idx]);
+        auto dst = thrust::get<1>(edgelist_first[idx]);
+
+        auto itr_pair = thrust::find(  // FIXME: replace by lower bound
```
seunghwak, line 181: This is a simple fix. Don't defer simple fixes for later updates.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -121,6 +314,8 @@ k_truss(raft::handle_t const& handle,
     edge_weight{std::nullopt};
```
seunghwak, line 314: Should we maintain `edge_weight` from here? We can treat the graph as an unweighted graph till we finish finding K-Truss edge sources & destinations. Once we have K-Truss edge sources & destinations, we can extract edge (src, dst, weight) triplets from the input graph.

seunghwak (commented):
I am reviewing the entire K-Truss code; more reviews coming.

ChuckHastings:
Moving this to 25.02... hopefully merging very early.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -214,121 +409,501 @@ k_truss(raft::handle_t const& handle,
...
+        handle.get_thrust_policy(),
+        get_dataframe_buffer_begin(triangles_from_weak_edges),
+        get_dataframe_buffer_end(triangles_from_weak_edges),
+        extract_triangles_from_weak_edges<vertex_t, edge_t>{
+          prev_chunk_size,
+          raft::device_span<size_t const>(intersection_offsets.data(), intersection_offsets.size()),
+          raft::device_span<vertex_t const>(intersection_indices.data(),
+                                            intersection_indices.size()),
+          raft::device_span<vertex_t const>(weak_edgelist_srcs.data(), weak_edgelist_srcs.size()),
+          raft::device_span<vertex_t const>(weak_edgelist_dsts.data(), weak_edgelist_dsts.size())});
```
seunghwak, line 543: Instead of storing all three edges in each triangle, what happens 

1) if we store only three triangle end points (a, b, c where out-degree(a) < out-degree(b) < out-degree(c) using the same tie breaking mechanism to create a DODG graph),
2) Shuffle based on edge partitioning using (a, b)
3) Remove duplicates
4) Re-create an edge list (a, b) (a, c), (b, c) from each (a, b, c)
5) Shuffle the result edge list
6) Decrease edge counts.

Won't this work? This sounds simpler & cheaper.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -214,121 +409,501 @@ k_truss(raft::handle_t const& handle,
...
+    cur_graph_view,
+    edgelist_dodg,
+    cugraph::edge_src_dummy_property_t{}.view(),
+    cugraph::edge_dst_dummy_property_t{}.view(),
+    cugraph::edge_dummy_property_t{}.view(),
+    [] __device__(auto src, auto dst, thrust::nullopt_t, thrust::nullopt_t, thrust::nullopt_t) {
+      return true;
+    },
+    dodg_mask.mutable_view(),
+    false);
```
seunghwak, line 471: Should we really extract `edgelist_dodg` and call `transform_e`?

Can't we combine the `extract_transform_e` and `transform_e` calls?

Call `transform_e` with `edge_src_out_degrees.view()` & `edge_dst_out_degrees.view()`?

cpp/src/community/k_truss_impl.cuh
seunghwak: And we haven't implemented truss decomposition, right?

K-Truss should be implemented as truss decomposition with min_k = max_k = k

cpp/src/community/k_truss_impl.cuh
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * Copyright (c) 2024, NVIDIA CORPORATION.
```
ChuckHastings, line 2: Looks like your copyright reverted... should include 2025.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -41,8 +42,68 @@
 #include <thrust/transform.h>
 #include <thrust/tuple.h>
 
+#include <chrono>
```
ChuckHastings, line 45: Is this used?

Rather not have dangling using statements like this in a header.  If we're actually using `std::chrono`, I think it's better to prefix the function calls/variable references (e.g. `std::chrono::duration`) rather than a using which loses the context of where `duration` comes from.

If this is just from some debugging/benchmarking that's not in use, best not to include it in the final code.

cpp/src/community/k_truss_impl.cuh
jnke2016: resolved

cpp/src/community/k_truss_impl.cuh
jnke2016: > If this is just from some debugging/benchmarking that's not in use, best not to include it in the final code.

Yes those were only for benchmarking purposes and I removed them

cpp/src/community/k_truss_impl.cuh
jnke2016: Right. resolved

cpp/src/community/k_truss_impl.cuh
jnke2016: Yes this is something we can work on and I already experimented it on my side. So it shouldn't be complex to add in the next releases. We can return the max truss value of each edge as an edge property.

jnke2016: > Won't this work? This sounds simpler & cheaper.

Yes this works and most if not all the steps you mentioned were added in my PR

cpp/src/community/k_truss_impl.cuh
jnke2016: Right. I treated the graph as unweighted and extracted the edges with weights(if exist) at the end.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -214,121 +409,501 @@ k_truss(raft::handle_t const& handle,
 
   // 3. Keep only the edges from a low-degree vertex to a high-degree vertex.
 
-  {
-    auto cur_graph_view = modified_graph_view ? *modified_graph_view : graph_view;
```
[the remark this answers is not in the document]
jnke2016: This was also addressed

cpp/src/community/k_truss_impl.cuh
```diff
@@ -121,6 +314,8 @@ k_truss(raft::handle_t const& handle,
     edge_weight{std::nullopt};
   std::optional<rmm::device_uvector<weight_t>> wgts{std::nullopt};
 
+  cugraph::edge_bucket_t<vertex_t, void, true, multi_gpu, true> edgelist_dodg(handle);
```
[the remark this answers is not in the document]
jnke2016: Right. Now I only maintain `cur_graph_view` and `unmasked_cur_graph_view`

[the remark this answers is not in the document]
jnke2016: > Better minimize the scope of edgelist_dodg

After updating my script, this variable was no longer need and removed

cpp/src/community/k_truss_impl.cuh
```diff
@@ -114,246 +188,547 @@ k_truss(raft::handle_t const& handle,
...
+      edge_dummy_property_t{}.view(),
+      [] __device__(auto src, auto dst, auto src_out_degree, auto dst_out_degree, auto) {
+        return (src_out_degree < dst_out_degree) ? true
+               : ((src_out_degree == dst_out_degree) &&
+                  (src < dst) /* tie-breaking using vertex ID */)
+                 ? true
+                 : false;
+      },
+      dodg_mask.mutable_view(),
+      false);
```
seunghwak, line 310:
```
dodg_mask.mutable_view());
```
No need to explicitly set do_expensive_check to false here.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -114,246 +188,547 @@ k_truss(raft::handle_t const& handle,
...
+      handle, unmasked_cur_graph_view, weak_edges_mask.mutable_view(), bool{true});
+
+    // Mask self loops and edges not being part of k-1 core
+    cugraph::transform_e(
+      handle,
+      unmasked_cur_graph_view,
+      edge_src_dummy_property_t{}.view(),
+      edge_dst_dummy_property_t{}.view(),
+      edge_mask.view(),
+      [] __device__(auto src, auto dst, auto src_out_degree, auto dst_out_degree, auto e_mask) {
```
seunghwak, line 331:
```
auto src_out_degree, auto dst_out_degree,
```
=>
```
auto, auto,
```
`src_out_degree` & `dst_out_degree` are  misnomers... They are dummy values.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -114,246 +188,547 @@ k_truss(raft::handle_t const& handle,
...
+        return e_mask ? true : false;
+      },
+      weak_edges_mask.mutable_view(),
+      false);
```
seunghwak, line 335: No need to explicitly set do_expensive_check to false here.

seunghwak: Won't weak_edge_mask = std::move(edge_mask) work here? (Or may just rename edge_mask to something like undirected_mask? So, we just maintain undirected_mask & dodg_mask.)

cpp/src/community/k_truss_impl.cuh
```diff
@@ -114,246 +188,547 @@ k_truss(raft::handle_t const& handle,
...
+      handle,
+      cur_graph_view,
+      cugraph::edge_src_dummy_property_t{}.view(),
+      cugraph::edge_dst_dummy_property_t{}.view(),
+      edge_triangle_counts.view(),
+      [] __device__(auto src, auto dst, thrust::nullopt_t, thrust::nullopt_t, auto count) {
+        return count == 0 ? false : true;
+      },
+      dodg_mask.mutable_view(),
+      false);
```
seunghwak, line 730: No need to explicitly set do_expensive_check = false.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -114,246 +188,547 @@ k_truss(raft::handle_t const& handle,
...
+          [key_func =
+             cugraph::detail::compute_gpu_id_from_int_edge_endpoints_t<vertex_t>{
+               raft::device_span<vertex_t const>(d_vertex_partition_range_lasts.data(),
+                                                 d_vertex_partition_range_lasts.size()),
+               comm_size,
+               major_comm_size,
+               minor_comm_size}] __device__(auto val) {
+            return key_func(thrust::get<0>(val), thrust::get<1>(val));
+          },
+
```
seunghwak, line 440: No need for an empty line here.

cpp/src/community/k_truss_impl.cuh
```diff
@@ -43,6 +44,69 @@
 
 namespace cugraph {
 
+template <typename vertex_t, typename edge_t>
+struct extract_weak_edges {
+  edge_t k{};
+  __device__ thrust::optional<thrust::tuple<vertex_t, vertex_t>> operator()(
+    vertex_t src, vertex_t dst, thrust::nullopt_t, thrust::nullopt_t, edge_t count) const
+  {
+    return ((count < k - 2) && (count > 0))
```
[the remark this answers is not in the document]
jnke2016: > Why should we check count > 0? A comment explaining this will be helpful.

`count` can only be >= 0 if we unrolled the edge count properly (without overcompensating). We don't need to process edges with `count == 0` that's why I check `count > 0`. I just replaced that check by `((count < k - 2) && (count != 0))` which makes more sense. 

Once I am done processing all the edges in the k-truss, I simply mask all the edges with `count ==0` as you can see below and then extract the remaining edges with is the final result.
```
cugraph::transform_e(
      handle,
      cur_graph_view,
      cugraph::edge_src_dummy_property_t{}.view(),
      cugraph::edge_dst_dummy_property_t{}.view(),
      edge_triangle_counts.view(),
      [] __device__(auto src, auto dst, cuda::std::nullopt_t, cuda::std::nullopt_t, auto count) {
        return count == 0 ? false : true;
      },
      dodg_mask.mutable_view(),
      false);
```

cpp/src/community/k_truss_impl.cuh
```diff
@@ -43,6 +44,69 @@
 
 namespace cugraph {
 
+template <typename vertex_t, typename edge_t>
+struct extract_weak_edges {
+  edge_t k{};
+  __device__ thrust::optional<thrust::tuple<vertex_t, vertex_t>> operator()(
```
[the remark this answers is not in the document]
jnke2016: Right. I updated my PR. As a matter of fact, I couldn't even build my branch once I merged the latest changes

cpp/src/community/k_truss_impl.cuh
jnke2016: Right. But since we have the expensive check exposed, I can just pass it to every function that support it/

cpp/src/community/k_truss_impl.cuh
jnke2016: Right, I am not passing anything for the src and dat property.

cpp/src/community/k_truss_impl.cuh
jnke2016: > weak_edge_mask = std::move(edge_mask)

Good catch. No need to repeat the same work in `edge_mask` with `transform_e`

seunghwak (approved):
LGTM

PR #4749, opened 2024-11-08
nx-cugraph: dispatch graph method to gpu or cpu

eriknw:
This is to avoid unnecessary conversions to networkx data structures.

Also, update to pass tests with networkx 3.4.2


eriknw:
Also here: https://github.com/rapidsai/nx-cugraph/pull/17

eriknw:
Closing in favor of https://github.com/rapidsai/nx-cugraph/pull/17

PR #4750, opened 2024-11-09
extract the edgelist from the graph

jnke2016:
This PR exposes the C++ function decompress_to_edgelist to the C, PLC and Python API. This will enable the extraction of the edgelist from a graph which is currently not supported. It also removes the deprecated parameter `legacy_renum_only`


python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -979,6 +980,83 @@ def convert_to_cudf(cp_arrays):
 
         return ddf
 
+    def decompress_to_edgelist(
+        self,
+        original: bool = True
```
rlratzel, line 985: I think this would be more descriptive if it was named `return_unrenumbered_edgelist`

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -979,6 +980,83 @@ def convert_to_cudf(cp_arrays):
...
+        Parameters
+        ----------
+        original    : bool (default=True)
+                    Flag determining whether to return the original input edgelist
+                    if 'True' or the renumbered one of 'False' and the edgelist was
+                    renumbered.
+
+        Returns
+        -------
+        ego_edge_lists : dask_cudf.DataFrame
```
rlratzel, line 999: Is this the right name for the return value?  Also, should it be plural?  I figured it would be just `edgelist`

python/pylibcugraph/pylibcugraph/decompress_to_edgelist.pyx
```diff
@@ -0,0 +1,169 @@
...
+    ...     [0.1, 2.1, 1.1, 5.1, 3.1, 4.1, 7.2, 3.2], dtype=numpy.float32)
+    >>> resource_handle = pylibcugraph.ResourceHandle()
+    >>> graph_props = pylibcugraph.GraphProperties(
+    ...     is_symmetric=False, is_multigraph=False)
+    >>> G = pylibcugraph.SGGraph(
+    ...     resource_handle, graph_props, srcs, dsts, weight_array=weights,
+    ...     store_transposed=False, renumber=False, do_expensive_check=False)
+    >>> (sources, destinations, edge_weights, _, _) =
+    ...     pylibcugraph.induced_subgraph(
+    ...         resource_handle, G, False)
```
rlratzel, line 98: Shouldn't this be calling `decompress_to_edgelist` somewhere?

python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -979,6 +980,83 @@ def convert_to_cudf(cp_arrays):
 
         return ddf
 
+    def decompress_to_edgelist(
```
rlratzel, line 983: I only see this on the MG Graph API, shouldn't this also be added to the SG Graph API?

jnke2016: I pushed the SG API

rlratzel (approved):

cpp/include/cugraph_c/graph_functions.h
```diff
@@ -324,6 +324,24 @@ cugraph_error_code_t cugraph_degrees(const cugraph_resource_handle_t* handle,
...
+ * true)
+ * @param [out] result              Opaque pointer to induced subgraph result
+ * @param [out] error               Pointer to an error object storing details of any error.  Will
+ *                                  be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_decompress_to_edgelist(const cugraph_resource_handle_t* handle,
+                                                    cugraph_graph_t* graph,
+                                                    bool_t do_expensive_check,
+                                                    cugraph_induced_subgraph_result_t** result,
```
seunghwak, line 342: Just a suggestion, but `cugraph_induced_subgraph_result_t` sounds a bit of misnomer here. We may used this data structure to just store edge list from the induced subgraph algorithm in the past, but if we wish to use this for other algorithms as well, we may better rename this now or sometime in the future.

cpp/src/c_api/decompress_to_edgelist.cpp
```diff
@@ -0,0 +1,139 @@
...
+  {
+    if constexpr (!cugraph::is_candidate<vertex_t, edge_t, weight_t>::value) {
+      unsupported();
+    } else {
+      if constexpr (store_transposed) {
+        error_code_ = cugraph::c_api::
+          transpose_storage<vertex_t, edge_t, weight_t, store_transposed, multi_gpu>(
+            handle_, graph_, error_.get());
+        if (error_code_ != CUGRAPH_SUCCESS)
+          ;
```
seunghwak, line 69: What's this?

cpp/include/cugraph_c/graph_functions.h
ChuckHastings: Agreed.  I would suggest `cugraph_edgelist_t`.  I have a draft PR (that I need to discard and start over with) that was going to create a `cugraph_edgelist_t` for the C API for graph construction.

I'd suggest the following:
* Replicate the `cugraph_induced_subgraph_result_t` type and C API accessor functions to call it `cugraph_edgelist_t` instead
* Mark the induced subgraph variations as deprecated
* Within the C API implementation (`src/c_api`)  you can either rename the `cugraph::c_api::cugraph_induced_subgraph_result_t` to `cugraph::c_api::cugraph_edgelist_t` and rename all references accordingly, or you can leave it using the legacy version and I'll fix it in 25.02 when I finish the C API graph construction updates.

ChuckHastings (approved):

seunghwak (approved):

PR #4767, opened 2024-11-15
Add generic edge properties object

ChuckHastings:
Adding features to support type-erased edge properties within cugraph.

Temporal graph support would require adding 2 additional edge properties.  This would result in 4x compile time and code size for functions that manipulate edge properties during the graph creation process.

This PR adds some new features to cugraph to reduce this compile time and provide the basis for adding the temporal edge properties to the graph in an efficient manner:
* device_vector_t - an analog to a cudf series... this is a type-erased structure for storing a vector that allows us to interpret it upon use rather than needing to templatize the types
* type_dispatcher - adapted from cudf to allow us to do dynamic type dispatching
* edge_properties_t - a single object that contains a collection of device_vector_t objects allowing us to pass the edge properties in a single unit between functions that need to operate on the

### DIFF
--- a/cpp/include/cugraph/device_vector.hpp
+++ b/cpp/include/cugraph/device_vector.hpp
@@ -16,7 +16,7 @@
 
 #pragma once
 
-#include "cugraph_c/types.h"
+#include <cugraph_c/types.h>
 
 #include <cugraph/utilities/cugraph_data_type_id.hpp>
 #include <cugraph/utilities/error.hpp>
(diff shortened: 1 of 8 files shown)


cpp/include/cugraph/device_vector.hpp
```diff
@@ -0,0 +1,157 @@
...
+
+namespace cugraph {
+
+namespace detail {
+
+inline rmm::device_buffer allocate_buffer(raft::handle_t const& handle,
+                                          cugraph_data_type_id_t t,
+                                          size_t size)
+{
+  return (t == BOOL) ? rmm::device_buffer(cugraph::packed_bool_size(size) * data_type_size(t),
```
seunghwak, line 39: `cugraph::packed_bool_size(size) * data_type_size(t)`, should `data_type_size(t)` here be `sizeof(uint32_t)`?

cpp/include/cugraph/device_vector.hpp
```diff
@@ -0,0 +1,157 @@
...
+                                          handle.get_stream())
+                     : rmm::device_buffer(size * data_type_size(t), handle.get_stream());
+}
+
+}  // namespace detail
+
+/**
+ * Class wrapping a type-erased device vector.
+ * */
+class device_vector_t {
```
seunghwak, line 49: I am not sure `device_vector` is a right name for type erased edge property vector. `device_vector` reminds me of `thrust::device_vector` or `rmm::device_uvector`. They are not type-erased vectors.

We may need a better name. My suggestions: `edge_property_buffer` or `property_buffer` or `cugraph::device_buffer`. `rmm::device_buffer` does not assume a specific type and may be more appropriate for type erased data types. Or we may use `column` (adopt column from `cudf::column`). In cuDF C++, they use `cudf::column` to store cuDF series. We may say `edge_property_column` or `property_column` or `cugraph::device_column` or just `cugraph::column`. In this case, we may rename the above `allocate_buffer` function to `allocate_column_buffer` as well.

seunghwak: I feel like we need more discussion about this naming.

cpp/include/cugraph/device_vector.hpp
```diff
@@ -0,0 +1,157 @@
...
+   *
+   * @tparam T      type for the array vector
+   *
+   * @return const pointer to the beginning of the vector
+   */
+  template <typename T>
+  T const* begin() const
+  {
+    return reinterpret_cast<T const*>(data_.data());
+  }
```
seunghwak, line 98: We may add cbegin() and cend() as well.

ChuckHastings:
We're taking a different approach to this.

PR #4769, opened 2024-11-18
Notebook to run Centrality against PatentsView data for Medium Blog

acostadon:


BradReesWork (changes requested):
Cell 1: please add text with license of dataset an steps for downloading and expanding
After cell 2 - add a new header - otherwise it looks like everything is related to downloading data
Cell 3: move the comment into the cell
Cell 13: I don't understand why you get the 2nd, 3rd, and 4th hops

You need more discriptive text bout what is going on

notebooks/demo/centrality_patentsview.ipynb
```diff
@@ -0,0 +1,590 @@
...
+    "# we are using the default argument values for all centrality functions except where noted.\n",
+    "\n",
+    "def compute_centrality(_graph, highest_degree=2000) :\n",
+    "\n",
+    "    # Compute Degree Centrality\n",
+    "    _d = cugraph.degree_centrality(_graph)\n",
+    "    print(\"Degree Centrality done\")\n",
+    "\n",
+    "    # Compute the Betweenness Centrality\n",
+    "    # The k value is needed in large graphs as by default it will calculate the for every pair in the graph.\n",
```
jnke2016, line 179: I think there is a missing word in this sentence between `the` and `for`. And I you can rephrase it slightly differently for better clarity. As you already know, the between centrality score of a vertex is a measure of the number of shortest paths that passes through it. Hence for large graphs, this can be really expensive because the algorithm will need to compute the all-pair shortest path. Setting a value of `k` limits the number of vertices involved in the all pairs shortest computation for `bc`. As a suggestion you can rephrase it along theses lines: 

> The k value is needed in large graphs as by default, it will calculate the betweenness centrality score based on every possible pairs in the graph. Limiting the number of vertices involved in the shortest path search leads to better runtime.

The second sentence is optional if you want to more descriptive

notebooks/demo/centrality_patentsview.ipynb
```diff
@@ -0,0 +1,590 @@
...
+    "    # in this case we limit it to 100 samples at most.\n",
+    "    _b = cugraph.betweenness_centrality(_graph,k=100)\n",
+    "    print(\"Between Centrality done\")\n",
+    "\n",
+    "    #Compute Katz Centrality\n",
```
jnke2016, line 184: Formatting typo here. forgot a space after `#`

notebooks/demo/centrality_patentsview.ipynb
```diff
@@ -0,0 +1,590 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Display the first five edges in the dataframe.\n",
+    "citation_df.iloc[:5]"
```
jnke2016, line 233: I believe `.head()` serves that purpose

jnke2016 (commented):
Looks good overall

jnke2016 (approved):

PR #4775, opened 2024-11-19
Remove edge renumber map from the homogeneous sampling API

jnke2016:
edge renumbering is only supported in heterogeneous neighborhood sampling hence trying to extract it from the PLC API leads to a segmentation fault

This PR :

1. fixes a bug in homogeneous neighborhood sampling (both uniform and bias).
2. properly handle vertex_type_offsets when performing heterogeneous renumbering
3. fixes a typo in the homogeneous neighborhood sampling docstrings


ChuckHastings (approved):

rlratzel (approved):

cpp/src/c_api/neighbor_sampling.cpp
```diff
@@ -1011,6 +1012,8 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
                                         options_.with_replacement_},
               do_expensive_check_);
         } else {
+          raft::print_device_vector(
+            "labels", (*start_vertex_labels).data(), (*start_vertex_labels).size(), std::cout);
```
seunghwak, line 1016: Do you need to keep this? If this is just for debugging, better delete.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -362,15 +369,45 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+                                                                 std::move(result_edge_ids),
+                                                                 std::move(result_edge_types),
+                                                                 std::move(result_hops),
+                                                                 std::move(result_labels),
+                                                                 label_to_output_comm_rank);
+
+  if (result_srcs.size() == 0) {
+    // Update the 'edgelist_label_offsets' array to be proportional to the
+    // number of labels
+    result_offsets->resize(num_labels + 1, handle.get_stream());
```
seunghwak, line 394: Is there any guarantee that if results_srcs.size() != 0, result_offsets.size() == num_labels + 1?  Say, we have 100 labels but only one label has a sampled edge, will this still work?

jnke2016: > Say, we have 100 labels but only one label has a sampled edge, will this still work?

This is a valid edge case to consider. I believe the current implementation of neighborhood sampling only keeps track of vertices that have outgoing edges. @ChuckHastings any thought?

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -58,11 +58,12 @@ prepare_next_frontier(
   bool dedupe_sources,
   bool do_expensive_check)
 {
+  // Create a single new frontier.
   vertex_partition_device_view_t<vertex_t, multi_gpu> d_vertex_partition(vertex_partition);
 
-  size_t frontier_size = sampled_dst_vertices.size();
+  size_t frontier_size = sampled_dst_vertices.size(); // std:reduce for all edge typw within hop / Reduce the size of each of the span. std::reduce for each of the size - example code that concatenate device vectors (look at that)
```
seunghwak, line 64: `all edge typew` => `all edge_types`

But not sure we need this comment here. This function doesn't need to know whether `sampled_dst_vertices` is from a single vector or multiple vectors.

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -58,11 +58,12 @@ prepare_next_frontier(
   bool dedupe_sources,
   bool do_expensive_check)
 {
+  // Create a single new frontier.
```
seunghwak, line 61: Is this comment still relevant?

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -105,6 +107,29 @@ neighbor_sample_impl(raft::handle_t const& handle,
   graph_view_t<vertex_t, edge_t, false, multi_gpu> modified_graph_view = graph_view;
   edge_masks_vector.reserve(num_edge_types);
 
+  label_t num_unique_labels = 0;
+
+  std::optional<rmm::device_uvector<label_t>> cp_starting_vertex_labels{std::nullopt};
```
seunghwak, line 112: What is `cp`? `copy`? Once the follow-up if block is executed, this is no longer a copy of the original starting vertex labels. This stores unique starting vertex labels, and I think `cp_starting_vertex_labels` is a misnomer.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -227,8 +252,12 @@ neighbor_sample_impl(raft::handle_t const& handle,
       if (labels) { (*level_result_label_vectors).push_back(std::move(*labels)); }
 
       if (num_edge_types > 1) { modified_graph_view.clear_edge_mask(); }
+
+      //vector.push_back (raft::device_span(vertex_t const)) //<- // level_result_dst_vectors
```
seunghwak, line 256: Delete the dead code.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -237,7 +266,7 @@ neighbor_sample_impl(raft::handle_t const& handle,
         handle,
         starting_vertices,
         starting_vertex_labels,
-        raft::device_span<vertex_t const>{level_result_dst_vectors.back().data(),
+        raft::device_span<vertex_t const>{level_result_dst_vectors.back().data(), // define a vector .
```
seunghwak, line 269: Are we still passing the sampled vertices for the last edge type only?

jnke2016: I think you are looking at old code. I no longer have it in my branch

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
jnke2016: These changes weren't meant to be committed. I will undo all changes to this file as I didn't modify anything here

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: Yes it is copy. 

> This stores unique starting vertex labels, and I think cp_starting_vertex_labels is a misnomer. 

the values in `cp_starting_vertex_labels` are not unique. I am only using this array to compute the number of unique labels

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: I think this was already addressed

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: You were right. I didn't commit my changes before pushing. All the concerns you had are addressed in my last commit

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -362,15 +480,73 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+                                                                 std::move(result_srcs),
+                                                                 std::move(result_dsts),
+                                                                 std::move(result_weights),
+                                                                 std::move(result_edge_ids),
+                                                                 std::move(result_edge_types),
+                                                                 std::move(result_hops),
+                                                                 std::move(result_labels),
+                                                                 label_to_output_comm_rank);
+
+  if (result_labels && (result_offsets->size() != num_unique_labels + 1)) {
```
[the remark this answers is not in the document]
jnke2016: I am not sure I understand this edge case. Aren't we specifying the labels for only the starting vertices? And starting vertex labels offsets is being passed from the CAPI hence the labels are contiguous.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -217,18 +262,93 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
-      level_result_src_vectors.push_back(std::move(srcs));
-      level_result_dst_vectors.push_back(std::move(dsts));
+      level_sizes_edge_types.push_back(srcs.size());
 
-      if (weights) { (*level_result_weight_vectors).push_back(std::move(*weights)); }
-      if (edge_ids) { (*level_result_edge_id_vectors).push_back(std::move(*edge_ids)); }
-      if (edge_types) { (*level_result_edge_type_vectors).push_back(std::move(*edge_types)); }
-      if (labels) { (*level_result_label_vectors).push_back(std::move(*labels)); }
+      level_result_src.resize(
+        std::reduce(level_sizes_edge_types.begin(), level_sizes_edge_types.end()),
```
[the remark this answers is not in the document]
jnke2016: No longer needed since I am now using `old_size`

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -217,18 +262,93 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+        handle.get_stream());
+      level_result_dst.resize(
+        std::reduce(level_sizes_edge_types.begin(), level_sizes_edge_types.end()),
+        handle.get_stream());
```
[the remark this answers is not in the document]
jnke2016: using `old_size + srcs.size()`

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -339,7 +457,7 @@ neighbor_sample_impl(raft::handle_t const& handle,
   if (return_hops) {
     result_hops   = rmm::device_uvector<int32_t>(result_size, handle.get_stream());
     output_offset = 0;
-    for (size_t i = 0; i < fan_out.size(); ++i) {
+    for (size_t i = 0; i < num_hops; ++i) {  // FIXME: replace this by the number of hops
```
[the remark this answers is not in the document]
jnke2016: No longer relevant

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -217,18 +262,93 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+                   std::reduce(level_sizes_edge_types.begin(), level_sizes_edge_types.end() - 1),
+                 srcs.begin(),
+                 srcs.size(),
+                 handle.get_stream());
+
+      raft::copy(level_result_dst.begin() +
+                   std::reduce(level_sizes_edge_types.begin(), level_sizes_edge_types.end() - 1),
+                 dsts.begin(),
+                 srcs.size(),
+                 handle.get_stream());
```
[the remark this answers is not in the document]
jnke2016: Did the same for all copies

alexbarghi-nv (approved):
👍

cpp/src/sampling/neighbor_sampling_impl.hpp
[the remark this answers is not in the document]
jnke2016: By global do you mean the multi-gpu case?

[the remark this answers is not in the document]
seunghwak: Think about the case `label_to_output_comm_rank` is defined.

In the multi-GPU case,

The unique labels in `starting_vertex_labels` and the labels mapped to this GPU (in `label_to_output_comm_rank`) may not coincide (I don't see any documentation about this).

What happens if the unique labels in `starting_vertex_labels` are 0, 7, 13 and the labels mapped to this GPU are 1, 4, 5?

Do you think this code will still work?

cpp/src/sampling/neighbor_sampling_impl.hpp
seunghwak: We'd better reduce the scope of this variable.

You are using `num_unique_labels` or `cp_starting_vertex_labels` only after invoking `shuffle_and_organize_output`.

The input vertex labels are `const` and won't be modified in this function (even though you're updating `starting_vertices` to point different arrays).

No reason to pre-allocate memory here and also define this variable way before it is actually used. This goes against minimizing memory footprint & code readability.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -217,18 +261,70 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+                   srcs.size(),
+                   handle.get_stream());
+      }
 
       if (num_edge_types > 1) { modified_graph_view.clear_edge_mask(); }
     }
 
+    level_sizes.push_back(level_result_src.size());
+    level_result_src_vectors.push_back(std::move(level_result_src));
+    level_result_dst_vectors.push_back(std::move(level_result_dst));
```
seunghwak, line 313: After calling std::move, the variable is in valid but arbitrary state. You should not use re-use the moved-out variable assuming that it's size is reset to 0. And to minimize the scope of per-level (per-hop) variables, we should better define these variables inside the `for (auto hop = 0; hop < num_hops; hop++) { ... }`.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -362,15 +458,80 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+  }
+
+  return std::make_tuple(std::move(result_srcs),
+                         std::move(result_dsts),
+                         std::move(result_weights),
+                         std::move(result_edge_ids),
+                         std::move(result_edge_types),
+                         std::move(result_hops),
+                         std::move(result_labels),
+                         std::move(result_offsets));
```
seunghwak, line 534: `result_offsets` here stores result label offsets, right? Shouldn't we better rename to `result_label_offsets`?

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: > The input vertex labels are const and won't be modified in this function (even though you're updating starting_vertices to point different arrays).

As you mentioned, vertex labels are constant hence it can't be sorted and I need the initial labels downstream. I reduced the scope of its copy

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: Right. It is more descriptive

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -135,6 +137,12 @@ neighbor_sample_impl(raft::handle_t const& handle,
 
   std::vector<rmm::device_uvector<vertex_t>> level_result_src_vectors{};
   std::vector<rmm::device_uvector<vertex_t>> level_result_dst_vectors{};
+
+  // Get the number of hop. If homogeneous neighbor sample, num_edge_types = 1.
+  auto num_hops = ((fan_out.size() % num_edge_types) == 0)
+                    ? (fan_out.size() / num_edge_types)
+                    : ((fan_out.size() / num_edge_types) + 1);
+
```
seunghwak, line 145: We may better define this before defining `level_result_src_vectors` and `level_result_dst_vectors`. So, we can define all `level_resutl_..._vectors` in one code block.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -252,11 +337,6 @@ neighbor_sample_impl(raft::handle_t const& handle,
 
     starting_vertices =
       raft::device_span<vertex_t const>(frontier_vertices.data(), frontier_vertices.size());
```
seunghwak, line 339: I assume we don't need this code anymore. We are directly creating a span from `frontier_vertices`.

seunghwak (approved):
LGTM except for minor cosmetic issues.

cpp/src/sampling/neighbor_sampling_impl.hpp
jnke2016: Should I initialize `frontier_vertices` to `starting_vertices`?

PR #4340, opened 2024-04-12
Build C++ wheel

vyasr:
This PR changes wheel building in RMM to create a separate C++ wheel that is then found from the Python wheel. The C++ wheel is now a hard dependency of the Python wheel. This allows Python packaging to more closely mirror the structure of our conda packaging, and the way we would normally wish to package these in any other manager. It also allows us to reduce package sizes by allowing better sharing of artifacts between different Python packages that rely on the same C++ components from other packages.

Contributes to rapidsai/build-planning#33

PR #4804, opened 2024-12-05
introduce libcugraph wheels

jameslamb:
Replaces #4340, contributes to https://github.com/rapidsai/build-planning/issues/33.

Proposes packaging `libcugraph` as a wheel, which is then re-used by `cugraph-cu{11,12}` and `pylibcugraph-cu{11,12}` wheels.

### Benefits of these changes

* smaller wheels (see "Size Changes" below)
  - *no more `pylibcugraph` and `cugraph` both holding copies of libcugraph.so*
* faster compile times
  - *no more re-compiling RAFT, thanks to https://github.com/rapidsai/raft/pull/2531*
  - *no more recompiling libcugraph.so in both `pylibcugraph` and `cugraph` wheel builds*
* other benefits mentioned in https://github.com/rapidsai/build-planning/issues/33

### Wheel contents

`libcugraph`:

* `libcugraph.so` (shared library)
* cuGraph headers
* vendored dependencies (`fmt`, `spdlog`, CCCL, `cuco`)

`pylibcugraph`:

* `pylibcugraph` Python / Cython code and compiled Cython extensions

`cugraph`:

* `cugraph` Python / Cython code and compiled Cython extension

### Dependency Flows

In short.... `libcugraph` contains `libcugraph.so` and `libcugraph_c.so` dynamic libraries and the headers to link against it.

* Anything that needs to link against cuGraph at build time pulls in `libcugraph` wheels as a build dependency.
* Anything that needs cuGraph's symbols at runtime pulls it in as a runtime dependency, and calls `libcugraph.load_library()`.

For more details and some flowcharts, see https://github.com/rapidsai/build-planning/issues/33#issuecomment-2590129852

### Size changes (CUDA 12, Python 3.12, x86_64)

| wheel                | num files (before) | num files (this PR) | size (before)  | size (this PR) |
|:---------------:|------------------:|-----------------:|--------------:|-------------:|
| `libcugraph`     |   ---                       |  1762                     | ---                   | 903M       |
| `pylibcugraph` |  190                      |   187                       | 901M              | 2M                 |
| `cugraph`         |  315                      |   313                      | 899M              | 3M                 |
|**TOTAL**          |   **505**         |   **2,262**               | **1,800M**                 | **908M**    |

*NOTES: size = compressed, "before" = 2025-01-13 nightlies*

*This is a cuGraph-specific slice of the table from https://github.com/rapidsai/raft/pull/2531. See that PR for details.*

### How I tested this

These other PRs:

* https://github.com/rapidsai/devcontainers/pull/435
* https://github.com/rapidsai/cugraph-gnn/pull/110


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -45,14 +45,25 @@ function(find_and_configure_raft)
...
+
+    # install the RAFT headers
+    # TODO(jameslamb): get exporting working
+    # TODO(jameslamb): move this under if(raft_ADDED)?
+    install(
+      TARGETS raft
+      EXPORT cugraph-exports
+      FILE_SET raft_headers
+      DESTINATION ${CMAKE_INSTALL_PREFIX}
     )
```
jameslamb, line 67: Alright, this is getting closer! But I need some CMake help. Here's what I'm trying to do in `libcugraph` wheels:

* build static RAFT (`libraft.a`) and statically link it into `libcugraph.so`
* do not install `libraft.a` in `libcugraph` wheels
* do not built `libraft.so` at all (or at least, don't install it into `libcugraph` wheels)
* install the RAFT headers into the `libcugraph` wheels, and export them
* use `libcugraph` wheels in the build env for `pylibcugraph` and `cugraph` wheels, and have `libcugraph`'s PUBLIC `raft::raft` dependency be served by the RAFT headers it vendors

I've got all the RAFT headers installing into `include/raft/` in the wheel the way I want, thanks to the combination of:

* setting `EXCLUDE_FROM_ALL TRUE` in the `rapids_cpm_find(raft)` call above
* this `install(TARGETS`
* using `target_sources()` + `FILE_SET` to associated the header files with the `raft` target in RAFT (https://github.com/rapidsai/raft/pull/2528)

But I can't figure out the exporting piece.

The current state of this branch results in stuff like this at configure time for `libcugraph`:

>   CMake Error: install(EXPORT "raft-compiled-exports" ...) includes target "raft_compiled" which requires target "raft" that is not in this export set, but in multiple other export sets: lib64/cmake/cugraph/cugraph-targets.cmake, lib64/cmake/raft/raft-targets.cmake.
  An exported target cannot depend upon another target which is exported multiple times. Consider consolidating the exports of the "raft" target to a single export.

>  CMake Error: install(EXPORT "raft-compiled-exports" ...) includes target "raft_compiled_static" which requires target "raft" that is not in this export set, but in multiple other export sets: lib64/cmake/cugraph/cugraph-targets.cmake, lib64/cmake/raft/raft-targets.cmake.

>   CMake Error: install(EXPORT "raft-compiled-lib-exports" ...) includes target "raft_lib" which requires target "raft" that is not in this export set, but in multiple other export sets: lib64/cmake/cugraph/cugraph-targets.cmake, lib64/cmake/raft/raft-targets.cmake.

If I remove the `EXPORT` line, `libcugraph` builds successfully but then the `pylibcugraph` builds cannot find those RAFT headers (because no `raft-config.cmake` is written out):

> -- Configuring done (3.9s)
CMake Error at /pyenv/versions/3.12.7/lib/python3.12/site-packages/libcugraph/lib64/cmake/cugraph/cugraph-targets.cmake:61 (set_target_properties):
  The link interface of target "cugraph::cugraph" contains:
>    raft::raft
>  but the target was not found.  Possible reasons include:
>    * There is a typo in the target name.
>    * A find_package call is missing for an IMPORTED target.
>    * An ALIAS target is missing.
> Call Stack (most recent call first):
>  /pyenv/versions/3.12.7/lib/python3.12/site-packages/libcugraph/lib64/cmake/cugraph/cugraph-config.cmake:73 (include)
>  CMakeLists.txt:30 (find_package)

If I remove this `install()` and remove `EXCLUDE_FROM_ALL TRUE` from the `rapids_cpm_find(raft)` call, then the RAFT headers, `libraft.a`, and `libraft.so` are all packaged in the `libcugraph` wheel and exported successfully, such that `pylibcugraph` and `cugraph` wheel builds find them and succeed ([without runtime issues, at least in the one test job I ran](https://github.com/rapidsai/cugraph/actions/runs/12241843897/job/34149484561)).

I think I'm just fundamentally misunderstanding some things about dependency tracking in CMake / `rapids-cmake`. I've tried various forms of:

* `rapids_export()`
* `rapids_export_package()`
* `install(EXPORT`

And haven't yet found the right combination.

If what I'm trying to do is not possible or just undesirable for something other reason, then other approaches to try (some of which are from an offline discussion with @bdice and @vyasr) include:

* omit `EXCLUDE_FROM_ALL TRUE` and just add a hack to the `libcugraph` CMakeLists.txt that manually deletes `libraft.a` and `libraft.so`
* have `pyblibcugraph` and `cugraph` wheel builds re-download all the RAFT headers via CPM at build time
* stop on this and first get `libraft` wheels set up, then use them at build time for `libcugraph`, `pylibcugraph`, and `cugraph`

jameslamb: I've started exploring whether we can avoid needing to worry about this by just creating `libraft` wheels: https://github.com/rapidsai/raft/pull/2531

We can come back to this thread if that doesn't work out.

python/libcugraph/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+
+option(USE_CUDA_MATH_WHEELS "Use the CUDA math wheels instead of the system libraries" OFF)
+
+# Check if cugraph is already available. If so, it is the user's responsibility to ensure that the
+# CMake package is also available at build time of Python packages that need libcugraph.
+find_package(cugraph "${RAPIDS_VERSION}")
+
+if(cugraph_FOUND)
+  return()
+endif()
```
jameslamb, line 33: This should have been enough to get devcontainer builds working (making this the type of "empty shell" wheel we've done with other C++ wheels like `libcudf`, and which we eventually want to stop doing with https://github.com/rapidsai/build-planning/issues/119).

But on https://github.com/rapidsai/devcontainers/pull/435, I saw this in the devcontainer build of `libcugraph`:

```text
  -- CPM: Using local package CCCL@2.7.0.0
  -- CPM: Using local package rmm@25.02.0
  -- CPM: Using local package cuco@0.0.1

  CMake Error at /home/coder/.conda/envs/rapids/share/cmake-3.31/Modules/Internal/CMakeCUDAArchitecturesValidate.cmake:9 (message):
    CMAKE_CUDA_ARCHITECTURES:

      RAPIDS

    is not one of the following:

      * a semicolon-separated list of integers, each optionally

        followed by '-real' or '-virtual'
      * a special value: all, all-major, native

  Call Stack (most recent call first):
    /home/coder/.conda/envs/rapids/share/cmake-3.31/Modules/CMakeDetermineCUDACompiler.cmake:112 (cmake_cuda_architectures_validate)

    /home/coder/raft/cpp/build/conda/cuda-12.0/latest/raft-config.cmake:170 (enable_language)
    build/conda/cuda-12.0/release/cmake/CPM_0.40.0.cmake:249 (find_package)
    build/conda/cuda-12.0/release/cmake/CPM_0.40.0.cmake:303 (cpm_find_package)
    /home/coder/cugraph/cpp/build/conda/cuda-12.0/latest/cugraph-dependencies.cmake:281 (CPMFindPackage)
    /home/coder/cugraph/cpp/build/conda/cuda-12.0/latest/cugraph-config.cmake:72 (include)
    CMakeLists.txt:29 (find_package)

  -- Configuring incomplete, errors occurred!

  *** CMake configuration failed
  error: subprocess-exited-with-error
  × Building editable for libcugraph (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /home/coder/.conda/envs/rapids/bin/python /home/coder/.conda/envs/rapids/lib/python3.12/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py build_editable /tmp/tmpcw086igy
  cwd: /home/coder/cugraph/python/libcugraph
  Building editable for libcugraph (pyproject.toml): finished with status 'error'
```

([logs link](https://productionresultssa19.blob.core.windows.net/actions-results/5fe50ddb-dbff-4cfb-b74e-4ed6022d9a03/workflow-job-run-c86cf860-88c0-577f-f339-cca88bd0bed3/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-01-14T16%3A08%3A21Z&sig=b%2B1yFp7N1AIc51eX5p0xRxF%2BarhSkEfLon9WoSFSspg%3D&ske=2025-01-15T02%3A26%3A35Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-01-14T14%3A26%3A35Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-11-04&sp=r&spr=https&sr=b&st=2025-01-14T15%3A58%3A16Z&sv=2024-11-04))

I'm not sure what caused that. In theory this should have just hit this early `return()` and left this with 0 targets to be built (just as happened successfully for the `libraft` wheel build further up in that same log).

I wouldn't expect that we need `LANGUAGES CUDA` or `rapids_cuda_init_architectures()` for `libcugraph` wheels... `libraft` is not doing that (https://github.com/rapidsai/raft/pull/2531) and neither is `libcudf` (https://github.com/rapidsai/cudf/blob/847029172ce47ef2109dc825f149ab58130a2fc6/python/libcudf/CMakeLists.txt#L19-L23).

python/libcugraph/CMakeLists.txt
```diff
@@ -0,0 +1,68 @@
...
+# --- cugraph ---#
+set(BUILD_TESTS OFF)
+set(BUILD_CUGRAPH_MG_TESTS OFF)
+set(CMAKE_CUDA_LINEINFO OFF)
+
+# --- raft -- #
+set(USE_RAFT_STATIC OFF)
+set(CUGRAPH_COMPILE_RAFT_LIB ON)
+
+add_subdirectory(../../cpp cugraph-cpp)
```
jameslamb, line 56: > **Wheel contents**
>   * _libcugraph.so (shared library)_
>   * _cuGraph headers_
>   * _vendored dependencies (fmt, spdlog, CCCL, cuco, nvtx3, rmm)_

Choosing a spot in the diff to start this conversation.... how do we want to handle these other header-only dependencies?

As of this set of related PRs (this, https://github.com/rapidsai/raft/pull/2531, and https://github.com/rapidsai/cuml/pull/6199), `libraft`, `libcugraph`, and `libcuml` wheels each vendor their own copies of things needed to link against them:

* `libcuml`: CCCL
* `libcugraph`: CCCL, cuco, fmt, nvtx3, rmm, spdlog
* `libraft`: CCCL, cuco, cute, cutlass, fmt

This happens in the conda packages too, though for a smaller set of packages. Here's what the latest 25.02 nightlies of those vendor:

* `libcuml`: CCCL
* `libcugraph`: CCCL, cuco, 
* `libraft-headers`: cuco, cute, cutlass

<details><summary>code to inspect those (click me)</summary>

```shell
LIBCUML_PKG="https://anaconda.org/rapidsai-nightly/libcuml/25.02.00a46/download/linux-64/libcuml-25.02.00a46-cuda12_250114_g47bac702c_46.conda"
LIBCUGRAPH_PKG="https://anaconda.org/rapidsai-nightly/libcugraph/25.02.00a67/download/linux-64/libcugraph-25.02.00a67-cuda12_250114_g8507cbf63_67.conda"
LIBRAFT_PKG="https://anaconda.org/rapidsai-nightly/libraft-headers-only/25.02.00a34/download/linux-64/libraft-headers-only-25.02.00a34-cuda12_250114_g5c826d73_34.conda"

wget -O ./libcugraph.conda ${LIBCUGRAPH_PKG}
wget -O ./libcuml.conda ${LIBCUML_PKG}
wget -O ./libraft.conda ${LIBRAFT_PKG}

cph list ./libcugraph.conda
cph list ./libcuml.conda
cph list ./libraft.conda
```

</details>

This PR should probably be updated to stop vendoring `rmm` (since we have wheels for it), but otherwise... I think this excessive vendoring should be addressed separately in follow-up work.

bdice: > This PR should probably be updated to stop vendoring rmm (since we have wheels for it).
>
> Otherwise... I think this excessive vendoring should be addressed separately in follow-up work.

I support this.

ci/validate_wheel.sh
bdice, line 1: Are we going to run the size checks for `libcugraph` wheels for each CUDA version? I see `max_allowed_size_compressed = '1.3G'` in the `pyproject.toml` but that seems like it would be for CUDA 11?

python/libcugraph/CMakeLists.txt
vyasr: This is happening because raft's config file [is explicitly enabling the CUDA language](https://github.com/rapidsai/raft/blob/branch-25.02/cpp/CMakeLists.txt#L495). The generated config file cannot embed a all to `rapids_cuda_init_architectures` without either imposing a requirement that rapids-cmake be part of the consumer's CMake build, or bundling this function directly into the config file. The former is a non-starter; exporting via rapids-cmake should not lead to projects requiring rapids-cmake, and the rest of RAPIDS shouldn't impose that requirement either. The second option could be workable, but it still feels like undesirable scope creep.

What this really boils down to is whose responsibility it is to support the extended CUDA architecture settings that rapids-cmake exposes. Since rapids-cmake is really not something that is part of exports but only something that is downloaded and used when a project is configured, I think we have to put the onus on the consumer. tl;dr I think this means that if a project does `find_project(raft)` (or in any way transitively winds up enabling the CUDA language) and if that project needs to support building in environments where the value of `CMAKE_CUDA_ARCHITECTURES` could be one of the values supported by rapids-cmake, that project has to call `rapids_cuda_init_architectures` here. It's a bit annoying, but I think it's the right division of responsibilities.

@robertmaynard would you agree?

robertmaynard: Excellent summary Vyas and I completley agree.

The reason that `add_subdirectory(../../cpp cugraph-cpp)` doesn't handle this for you is that `rapids_cuda_init_architectures` must be called before the root level project call

ci/validate_wheel.sh
jameslamb: I don't see a meaningful size difference across CUDA versions for `libcugraph` wheels here.

For example, look at these builds from this PR:

**CUDA 11.8.0, Python 3.12, amd64** ([build link](https://github.com/rapidsai/cugraph/actions/runs/12775494732/job/35612054813?pr=4804#step:10:2461))

```text
file size
  * compressed size: 0.9G
  * uncompressed size: 1.4G
```

**CUDA 12.5.1, Python 3.12, amd64** ([build link](https://github.com/rapidsai/cugraph/actions/runs/12775494732/job/35612055205?pr=4804#step:10:2488))

```text
file size
  * compressed size: 0.9G
  * uncompressed size: 1.4G
```

I think that cuGraph doesn't actually directly need any of the CUDA math libraries (cuBLAS, cuSolver, etc.)... just indirectly via RAFT needing them. We're observing this now that `libcugraph` wheels are linking against a pre-built RAFT delivered by the `libraft` wheels, instead of building RAFT from source here.

I don't see any of those libraries referenced in code here (except some cuRAND usage in C++ tests)

```shell
git grep -i blas cpp/
git grep -i blas python/

git grep -i fft cpp/
git grep -i fft python/

git grep -i solver cpp/
git grep -i solver python/

git grep -i curand cpp/
git grep -i curand python/
```

Does that sound plausible? If so, I'll try dropping the dependency on CUDA math wheels for `libcugraph` wheels here.

bdice: I agree with your assessment (confirmed independently). Let's try it.

python/libcugraph/CMakeLists.txt
jameslamb: Great, thank you. Pushed https://github.com/rapidsai/cugraph/pull/4804/commits/540c1a4a251a3e13e815b290439cba6e51b2c5e3 to do that, confirmed that RMM isn't showing up in any of the wheels produced from this PR.

The "wheel contents" table in this PR's description and on https://github.com/rapidsai/raft/pull/2531 reflect this.

ci/validate_wheel.sh
bdice: I did a bit more research to validate my understanding.

On the conda side, [RAFT has no run exports](https://github.com/rapidsai/raft/blob/branch-25.02/conda/recipes/libraft/meta.yaml) that would cause a `host:` dependency on RAFT to add `run:` dependencies on the related CUDA math libraries. We do this because we don't know which math libraries would be relevant for any given RAFT consumer. However, because libraft wheels already supply those math libraries, we won't need to specify them for cuGraph. Transitive discovery is sufficient.

cpp/cmake/thirdparty/get_raft.cmake
jameslamb: At this point it looks like we're moving forward with a `libraft` wheel instead (https://github.com/rapidsai/raft/pull/2531) and that that's working here... so this thread and https://github.com/rapidsai/raft/pull/2528 are no longer needed.

bdice:
I bumped this to use published libraft wheels.

ci/validate_wheel.sh
jameslamb: Ok perfect, thank you.

I've pushed commits here removing the direct dependencies on CUDA math libraries for cuGraph wheels. There's one caveat... they do need to be listed out explicitly in the `--exclude` args passed to `auditwheel repair`, or otherwise `auditwheel` ends up vendoring them. I've done that in `ci/build_wheel_libcugraph.sh`.

jameslamb:
Python tests here are failing as a result of these RMM changes:  https://github.com/rapidsai/rmm/pull/1775

> E   FutureWarning: The `rmm._cuda.stream` module is deprecated in 25.02 and will be removed in a future release. Use `rmm.pylibrmm.stream` instead.

([build link](https://github.com/rapidsai/cugraph/actions/runs/12820092417/job/35752921907?pr=4804#step:8:5666))

Think that might be coming through via `cudf`. It's being discussed at https://github.com/rapidsai/cuml/pull/6229#issuecomment-2597838403

python/libcugraph/CMakeLists.txt
jameslamb: Thank you both! Explanations make sense to me. I pushed commits here yesterday calling `rapids_cuda_init_architectures()` before `project(libcugraph-python)`. 

It it looks like the `libcugraph` wheel (and everything else) built successfully in devcontainers CI: https://github.com/rapidsai/devcontainers/actions/runs/12816845275/job/35738725807

ci/build_wheel_libcugraph.sh
```diff
@@ -0,0 +1,39 @@
...
+rapids-logger "Done build requirements"
+
+# build with '--no-build-isolation', for better sccache hit rate
+# 0 really means "add --no-build-isolation" (ref: https://github.com/pypa/pip/issues/5735)
+export PIP_NO_BUILD_ISOLATION=0
+
+RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
+
+PARALLEL_LEVEL=$(python -c \
+  "from math import ceil; from multiprocessing import cpu_count; print(ceil(cpu_count()/4))")
```
bdice, line 34: This is copied from pylibcugraph, so I want to question it a bit. Are we really limited to only 1/4 of the available cores? I am guessing it's due to memory requirements at compile time?

Can we try to be more aggressive in a follow-up PR? Or can we add comments to explain why this is so limited?

dependencies.yaml
```diff
@@ -133,16 +161,20 @@ files:
       key: requires
     includes:
       - common_build
+      - depends_on_libcugraph
+      - depends_on_libraft
+      - depends_on_librmm
```
bdice, line 166: It surprises me that we have so many C++ wheel dependencies in pylibcugraph. Does pylibcugraph directly use RAFT / RMM C++ headers? I would have expected that pylibraft / rmm would have sufficed here.

bdice (commented):
Some comments to address.

.github/workflows/pr.yaml
```diff
@@ -180,9 +189,11 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_wheel_cugraph.sh
+      # This selects "ARCH=amd64 + the latest supported Python + CUDA".
+      matrix_filter: map(select(.ARCH == "amd64")) | max_by([(.PY_VER|split(".")|map(tonumber)), (.CUDA_VER|split(".")|map(tonumber))]) | [.]
```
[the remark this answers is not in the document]
jameslamb: ah yep, this was temporary to limit the impact of this PR on CI resources while I was testing. It should be removed.

.github/workflows/pr.yaml
```diff
@@ -180,9 +189,11 @@ jobs:
...
   devcontainer:
     secrets: inherit
-    needs: telemetry-setup
+    # needs: telemetry-setup
```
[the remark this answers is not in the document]
jameslamb:
```suggestion
    needs: telemetry-setup
```

Nope, that's a mistake.

ci/build_wheel.sh
```diff
@@ -29,22 +30,31 @@ python -m pip wheel \
...
+EXCLUDE_ARGS=(
+  --exclude "libraft.so"
+)
+
+# Avoid picking up dependencies on CUDA wheels that come through
+# transitively from 'libraft'.
+#
+# 'libraft' wheels are responsible for carrying a runtime dependency on
+# these based on RAFT's needs.
+EXCLUDE_ARGS+=(
```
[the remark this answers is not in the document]
jameslamb: They should only *matter* for CUDA 12, but there's no harm in having them be applied unconditionally, right? If no linkage for these libraries comes through from `libraft` for CUDA 11 (which is what we expect), then these arguments just won't do anything... `auditwheel` won't detect e.g. a `libcublas.so.12` linkage and therefore it won't matter that we told it to `--exclude` that.

Happy to add the CUDA version guard back if you want, I just thought this was a little simpler... it makes it clearer that there's no situation in which we expect any of the wheels from this repo to vendor these CUDA libraries.

ci/build_wheel_libcugraph.sh
jameslamb: >  am guessing it's due to memory requirements at compile time?

Exactly, this was added back in https://github.com/rapidsai/cugraph/pull/4489#discussion_r1643392183 to deal with out-of-memory issues building wheels.

Since most of the heavy compilation would move to `libcugraph` wheels in this PR, I've moved this limiting (and requesting a `cpu32` node) over to `libcugraph` here.

HOWEVER... maybe now that these builds aren't also compiling RAFT and since https://github.com/rapidsai/cugraph/pull/4720 reduced the number of TUs being compiled, we could get away with a higher degree of parallelism.

I'll push a change here trying to remove this limit, let's see what happens.

ci/build_wheel.sh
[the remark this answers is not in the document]
bdice: Sounds fine to me. Just wanted to be sure. Thanks!

dependencies.yaml
jameslamb: > *Does pylibcugraph directly use RAFT / RMM C++ headers?*

`pylibcugraph` doesn't, but RAFT / RMM appear in `libcugraph`'s public headers and are therefore `PUBLIC` dependencies in CMake. So those headers need to get into the build environment to link against `libcugraph.so`.

https://github.com/rapidsai/cugraph/blob/0b50bf925614dd10a603918885e953c0dcf6759b/cpp/include/cugraph/algorithms.hpp#L26-L30

https://github.com/rapidsai/cugraph/blob/0b50bf925614dd10a603918885e953c0dcf6759b/cpp/CMakeLists.txt#L482-L485

This is related to the discussions from https://github.com/rapidsai/build-planning/issues/92 , and it's for similar reasons that e.g. `pylibcudf` has a `librmm` build dependency despite not directly including any RMM headers:

https://github.com/rapidsai/cudf/blob/a4bbd0930a0e4922f69586560b064a0bd9e6aedc/python/pylibcudf/pyproject.toml#L115

ci/build_wheel_libcugraph.sh
jameslamb: I just pushed a changed removing this limiting of `PARALLEL_LEVEL`. I left this requesting a `cpu32` node for now. Let's see how it goes.

BradReesWork (approved):

bdice (approved):

PR #4805, opened 2024-12-05
remove 'wget' conda dependency, re-organize dependencies.yaml

jameslamb:
In #4804, I've started working on adding `libcugraph` wheels.

This includes a few fixes for things I noticed while doing that:

* removes `wget` from `test_notebook` environment
  - *our CI images already have this system-installed, and removing it helps to remove it from the environment solve for the unified RAPIDS devcontainers*
* `dependencies.yaml` re-organization:
  - breaks `librmm` dependency out into a `depends_on_librmm` to reduce duplication, and for consistency with other RAPIDS dependencies
  - uses `depends_on_pylibwholegraph` group everywhere instead of repeating `pylibwholegraph` in multiple places
  - removes unused YAML anchors
  - alphabetizes lists

## Notes for Reviewers

I'd originally wanted to also add a `librmm` wheel dependency for wheel builds here, but looks like doing that resulted in a lot more `sccache` misses, I guess because of building with build isolation. That change will happen in #4804 .


jameslamb:
Tests here are failing like this:

```text
AttributeError: module 'cudf._lib' has no attribute 'transform'
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/12242819477/job/34200391264?pr=4805))

Documented that in https://github.com/rapidsai/cugraph/issues/4823 ... there are some changes in `cudf` that `cugraph` needs to adapt to.

jameslamb:
Oh great thanks for merging `brnach-25.02` into this @nv-rliu !

dependencies.yaml
```diff
@@ -565,6 +562,31 @@ dependencies:
               - pylibwholegraph-cu11==25.2.*,>=0.0.0a0
           - {matrix: null, packages: [*pylibwholegraph_unsuffixed]}
 
+  depends_on_librmm:
```
bdice, line 565: We probably want the same for other packages too? libcudf, libraft? It’s okay to defer on this in the name of gradual improvement.

bdice (approved):

dependencies.yaml
jameslamb: Sure, we can go further in this PR. I just pushed https://github.com/rapidsai/cugraph/pull/4805/commits/64d0fad48f3d07e6100eea13d008d4de985a7280 moving more of the RAPIDS dependencies into these `depends_on_*` lists. 

Some notes on other changes in that commit:

* proposes removing explicit `libraft-headers` listing
   - `libraft` is explicitly listed, and directly pulls that in ([recipe link](https://github.com/rapidsai/raft/blob/d7e68f55c58493bc9cf1fbe4eb775a9593891c58/conda/recipes/libraft/meta.yaml#L157))
   - normally I'd advise against relying on transitive relationships like this implicitly, but in this case I think it's ok... the `libraft` conda package's purpose is to pull in the compiled bits of RAFT and the headers needed to link against them
   - if you disagree I'll happily add a `depends_on_libraft_headers`, don't feel THAT strongly about it either way
* removes `pyproject` and `requirements` entries for `libcudf`, `libraft,` and `librmm`
  - because `libcudf` is only used for benchmarks here, which aren't built from wheels
  - because `libraft` wheels don't actually exist yet
  - because `librmm` wheels aren't being used here yet
  - I'll add `libraft` and `librmm` entires in #4804
  - these `requirements` dependencies would all end up in the devcontainers environment: https://github.com/rapidsai/devcontainers/blob/branch-25.02/features/src/rapids-build-utils/opt/rapids-build-utils/bin/make-pip-dependencies.sh#L46

Requesting a re-review, that's too much to merge on a stale approval.

bdice: Yes, it is okay to use libraft to get its headers transitively.

The other bits are fine, too. I’ll re-review, just make sure update-version.sh and conda recipes reflect things correctly.

dependencies.yaml
```diff
@@ -500,19 +462,68 @@ dependencies:
...
+              cuda_suffixed: "true"
+            packages:
+              - cugraph-cu11==25.2.*,>=0.0.0a0
+          - {matrix: null, packages: [*cugraph_unsuffixed]}
+
+  depends_on_cugraph_service_client:
+    common:
+      - output_types: conda
+        packages:
+          - &cugraph_service_client_unsuffixed cugraph-service-client==25.2.*,>=0.0.0a0
```
bdice, line 501: This package does not use CUDA or have CUDA dependencies, so it should never be suffixed. This can be simplified.

References:

https://github.com/rapidsai/cugraph/blob/6783aec7a56f05d5088073c822140bc0cd7d9665/python/cugraph-service/client/pyproject.toml#L51

https://anaconda.org/search?q=cugraph-service-client

bdice (commented):
I saw one error in cugraph-service-client (no suffix needed). Also check update-version.sh and meta.yaml to make sure everything is in sync.

dependencies.yaml
jameslamb: Ah good catch, thanks. Guess the suffixes that were in the file before were unnecessary. I pushed https://github.com/rapidsai/cugraph/pull/4805/commits/7fc94736184cf1fcee3349f650ef8ceecdf0ffc7 removing the suffix-related stuff for that.

dependencies.yaml
jameslamb: Alright great, thanks.

> update version.sh and conda recipes

Just did that in https://github.com/rapidsai/cugraph/pull/4805/commits/34697a69b91ff5b65b1556a3b8a557dba58674f9.

* removed `libraft-headers` from conda recipes and `update-version.sh`
* put a floor on `thriftpy2` in `dependencies.yaml` to match what conda recipes were using

bdice (approved):

PR #4808, opened 2024-12-05
[BUG] Set Max PyTorch Version, Skip 11.4 Tests Using WholeGraph

alexbarghi-nv:
Set Max PyTorch Version, Skip 11.4 Tests Using WholeGraph

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -507,7 +507,7 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - &pytorch_conda pytorch>=2.3
+          - &pytorch_conda pytorch>=2.3,<2.5
           - torchdata
           - pydantic
           - ogb
--- a/python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
+++ b/python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
@@ -15,6 +15,8 @@ import pytest
 import numpy as np
 import os
 
+from cuda import cudart
+
 from cugraph.gnn import FeatureStore
 
 from cugraph.utilities.utils import import_optional, MissingModule
@@ -66,6 +68,9 @@ def runtest(rank: int, world_size: int):
 @pytest.mark.skipif(
     isinstance(pylibwholegraph, MissingModule), reason="wholegraph not available"
 )
+@pytest.mark.skipif(
+    cudart.cudaRuntimeGetVersion()[1] < 11080, reason="not compatible with CUDA < 11.8"
+)
 def test_feature_storage_wholegraph_backend():
     world_size = torch.cuda.device_count()
     print("gpu count:", world_size)
@@ -80,6 +85,9 @@ def test_feature_storage_wholegraph_backend():
 @pytest.mark.skipif(isinstance(torch, MissingModule), reason="torch not available")
 @pytest.mark.skipif(
     isinstance(pylibwholegraph, MissingModule), reason="wholegraph not available"
+)
+@pytest.mark.skipif(
+    cudart.cudaRuntimeGetVersion()[1] < 11080, reason="not compatible with CUDA < 11.8"
 )
 def test_feature_storage_wholegraph_backend_mg():
     world_size = torch.cuda.device_count()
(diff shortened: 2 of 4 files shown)


dependencies.yaml
```diff
@@ -507,7 +507,7 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - &pytorch_conda pytorch>=2.3
+          - &pytorch_conda pytorch>=2.3,<2.5
```
bdice, line 510: Typically we spell upper pinnings with an alpha like:
```suggestion
          - &pytorch_conda pytorch>=2.3,<2.5a0
```

python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
```diff
@@ -66,6 +68,9 @@ def runtest(rank: int, world_size: int):
 @pytest.mark.skipif(
     isinstance(pylibwholegraph, MissingModule), reason="wholegraph not available"
 )
+@pytest.mark.skipif(
+    cudart.cudaRuntimeGetVersion()[1] < 11080, reason="not compatible with CUDA < 11.8"
```
bdice, line 72: Despite the name, this is not a reliable way to get the CUDA Runtime version. It gets the runtime version used to build cuda-python, more-or-less. See the below links for some details.

https://github.com/NVIDIA/cuda-python/issues/16
https://github.com/rapidsai/rmm/issues/948
https://github.com/rapidsai/rmm/pull/1675 

Meanwhile, the CUDA 11.4 jobs [actually use cuda-python 11.8](https://github.com/rapidsai/cugraph/actions/runs/12178064576/job/33967234457#step:8:7090) -- which is fine, as it is okay to use a newer cuda-python with an older CTK -- but it means we can't use this function to get the info you're wanting. It will report 11.8 even in the 11.4 CI jobs.

The best workaround I know of is from [this PR](https://github.com/rapidsai/rmm/pull/946):
```python
import numba.cuda

def get_cudart_version():
    major, minor = numba.cuda.runtime.get_version()
    return major * 1000 + minor * 10
```

bdice (changes requested):
A fix is probably needed for the conditionally skipped tests on earlier CUDA versions.

python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
vyasr: Ack I knew about this but totally just forgot while I recommended this solution. It looks like we now have [`getLocalRuntimeVersion`](https://github.com/NVIDIA/cuda-python/issues/16#issuecomment-1774395711) now though?

alexbarghi-nv: This should be resolved now

bdice: getLocalRuntimeVersion only works correctly with the latest release of cuda-python. We have an upper bound that excludes that in 24.12.

bdice (approved):

python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
```diff
@@ -15,6 +15,8 @@
 import numpy as np
 import os
 
+import numba.cuda
```
rlratzel, line 18: Do we need to update `dependencies.yaml` to also include this new dependency, or is it already included or covered some other way?

rlratzel: Ah, nevermind, I see it's already a dependency.

rlratzel (approved):

python/cugraph/cugraph/tests/data_store/test_gnn_feat_storage_wholegraph.py
alexbarghi-nv: Covered by `cudf` https://github.com/rapidsai/cudf/blob/38261f8509245f88bdeab193a1357d9c73d765f0/dependencies.yaml#L692

Issue #4818, opened 2024-12-09
[BUG]: 24.12 nightly wheel tests failing with SSL certificate issues


jameslamb:
### Version

24.12

### Describe the bug.

Many nightly wheel tests are failing on 24.12. I see dozens of failures, in two broad categories.

Stuff like this:

> FAILED python/cugraph/cugraph/tests/traversal/test_sssp.py::test_sssp_data_type_conversion[1-graph_file0] - KeyError: "results for {'algo': 'single_source_dijkstra_path_length', 'graph_dataset': 'karate', 'graph_directed': 'True', 'source': '1', 'test': 'data_type_conversion', 'resultset_name': 'traversal'} not found"

And stuff like this:

> FAILED python/cugraph/cugraph/tests/utils/test_dataset.py::test_download[dataset0] - urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1007)>

([build link](https://github.com/rapidsai/cugraph/actions/runs/12235138888/job/34144089538))

These failures only occur on the CUDA 11.8.0, Python 3.10 wheels test jobs.

### Minimum reproducible example

```shell
See CI links above.
```

### Relevant log output

```shell
See CI links above.
```

### Environment details

```shell
See CI links above.
```

### Other/Misc.

Because they looked network-related, I did try manually re-running these failed jobs... they failed in exactly the same way.

I know that `cugraph` tests do some downloading of externally-hosted datasets... maybe one of those sites hosting that data had an outage or changed some configuration?

### Code of Conduct

- [x] I agree to follow cuGraph's Code of Conduct
- [x] I have searched the [open bugs](https://github.com/rapidsai/cugraph/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

rlratzel:
I'm wondering if the solution [here](https://github.com/rapidsai/graph_dl/pull/656) is relevant for these SSL errors.

nv-rliu:
Can this be closed now?

jameslamb:
I think so, yes.

The latest 24.12 test run did fail: https://github.com/rapidsai/cugraph/actions/runs/12349459995

As did the latest 25.02 test run: https://github.com/rapidsai/cugraph/actions/runs/12372518014

[...]

PR #4825, opened 2024-12-10
Fix SSL Error

nv-rliu:
Addresses errors in `wheel-tests` seen [here](https://github.com/rapidsai/cugraph/issues/4818)

### DIFF
--- a/python/cugraph/cugraph/datasets/dataset.py
+++ b/python/cugraph/cugraph/datasets/dataset.py
@@ -11,15 +11,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import cudf
-import dask_cudf
-import yaml
 import os
 import pandas as pd
-import cugraph.dask as dcg
+import yaml
 from pathlib import Path
-import urllib.request
+from urllib.requests import urlretrieve
+
+import cudf
+import cugraph.dask as dcg
+import dask_cudf
 from cugraph.structure.graph_classes import Graph
+from cugraph.utilities import install_ssl_cert
 
 
 class DefaultDownloadDir:
@@ -142,7 +144,8 @@ class Dataset:
         filename = self.metadata["name"] + self.metadata["file_type"]
         if self._dl_path.path.is_dir():
             self._path = self._dl_path.path / filename
-            urllib.request.urlretrieve(url, str(self._path))
+            install_ssl_cert()
+            urlretrieve(url, str(self._path))
 
         else:
             raise RuntimeError(
@@ -458,7 +461,8 @@ def download_all(force=False):
                 filename = meta["name"] + meta["file_type"]
                 save_to = default_download_dir.path / filename
                 if not save_to.is_file() or force:
-                    urllib.request.urlretrieve(meta["url"], str(save_to))
+                    install_ssl_cert()
+                    urlretrieve(meta["url"], str(save_to))
 
 
 def set_download_dir(path):
--- a/python/cugraph/cugraph/utilities/utils.py
+++ b/python/cugraph/cugraph/utilities/utils.py
@@ -25,6 +25,10 @@ from rmm._cuda.gpu import getDeviceAttribute
 
 from warnings import warn
 
+import certifi
+from ssl import create_default_context
+from urllib.request import build_opener, HTTPSHandler, install_opener
+
 # optional dependencies
 try:
     import cupy as cp
@@ -549,3 +553,13 @@ def create_directory_with_overwrite(directory):
     if os.path.exists(directory):
         shutil.rmtree(directory)
     os.makedirs(directory)
+
+
+def install_ssl_cert():
+    """
+    Build and install an opener with the custom HTTPS handler. Use this when
+    downloading datasets to have the proper SSL certificate.
+    """
+    ssl_context = create_default_context(cafile=certifi.where())
+    https_handler = HTTPSHandler(context=ssl_context)
+    install_opener(build_opener(https_handler))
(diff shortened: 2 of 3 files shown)


python/cugraph/cugraph/utilities/utils.py
```diff
@@ -549,3 +553,13 @@ def create_directory_with_overwrite(directory):
     if os.path.exists(directory):
         shutil.rmtree(directory)
     os.makedirs(directory)
+
+
+def install_ssl_cert():
```
jameslamb, line 558: Putting this here in `cugraph.utilities` means that now `certifi` needs to be introduced as a runtime dependency of `cugraph`.

> ModuleNotFoundError: No module named 'certifi'

([build link](https://github.com/rapidsai/cugraph/actions/runs/12265067332/job/34222694003?pr=4825))

I'm really nervous about the idea of introducing a new runtime dependency this long after code freeze and this close to the release.

And even just in general... does this really need to be in the `cugraph` package? Are those code paths in `datasets.py` intended to be used by downstream users, or are they just in the `cugraph` package for convenience in its own testing (and testing for `cugraph-gnn` libraries)?

Here in 24.12, to minimize risk, I think we should only run this at test time in CI here.

That'd probably mean:

* make `certifi` a test-only dependency
* put this bit of Python code in a script like `ci/install-certifi-certs.py`
* run that in each `ci/test_*` script, before any tests

jameslamb (changes requested):
I'm supportive of using `certifi` certs to support the dataset-downloading you need to do on Rocky Linux 8, but don't think we should be adding a new runtime dependency to the `cugraph` Python package.

python/cugraph/cugraph/utilities/utils.py
nv-rliu: I agree that it is not a good time to add an additional runtime dependency..

The codepaths in `datasets.py` is used both by end users and for our own testing purposes. You can think of them as analogous to [NetworkX's built-in graphs](https://networkx.org/documentation/stable/auto_examples/graph/index.html)

I don't think that the SSL issue would affect our users. I think this is just affecting CI because of how our images are setup. It would probably be fine to go with the solution you suggested and install it for our tests. I'll go ahead and add that

jameslamb: Ok got it, thanks for the explanation! Then yeah, I think doing this only as a testing dependency is a good solution.

I tested whether this could just be run outside of of `pytest`, like I suggested... and you're absolutely right, it can't. I guess something in this solution must also modify `os.environ` or similar.

<details><summary>code I used to confirm that (click me)</summary>

```shell
docker run \
    --rm \
    -it rapidsai/citestwheel:cuda11.8.0-rockylinux8-py3.10 \
    bash

python -c 'import urllib.request; urllib.request.urlretrieve("https://data.rapids.ai/cugraph/results/resultsets.tar.gz", "foo.tar.gz")'
# urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED

cat > ./install-ca-certs.py <<EOF
import certifi
from ssl import create_default_context
from urllib.request import build_opener, HTTPSHandler, install_opener

ssl_context = create_default_context(cafile=certifi.where())
https_handler = HTTPSHandler(context=ssl_context)
install_opener(build_opener(https_handler))
EOF

python -m pip install certifi
python ./install-ca-certs.py

python -c 'import urllib.request; urllib.request.urlretrieve("https://data.rapids.ai/cugraph/results/resultsets.tar.gz", "foo.tar.gz")'
# urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED
```

</details>

Given that, I think the way you've set this up as a testing-only dependency is a good solution. Thanks for working through it with me.

jameslamb (approved):
I support this change, especially since you saw it fix the error locally. Thanks for finding a minimal reproducible example and then putting up a fix so quickly!

python/cugraph/cugraph/utilities/utils.py
nv-rliu: Thank _you_ for helping track this down and brainstorming a solution with me.

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -458,7 +459,7 @@ def download_all(force=False):
                 filename = meta["name"] + meta["file_type"]
                 save_to = default_download_dir.path / filename
                 if not save_to.is_file() or force:
-                    urllib.request.urlretrieve(meta["url"], str(save_to))
+                    urlretrieve(meta["url"], str(save_to))
```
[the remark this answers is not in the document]
nv-rliu: Hmm, I don't believe I have any more changes to push as of now. The PR is just about to pass CI so let me know if you think it'll be better to still revert the changes to have as minimal of a diff has possible. Thanks!

dependencies.yaml
```diff
@@ -470,6 +470,7 @@ dependencies:
     common:
       - output_types: [conda, requirements]
         packages:
+          - certifi
```
bdice, line 473: @nv-rliu Can I ask why this was added to notebook dependencies? It looks like the package is only imported in tests, not notebooks.

nv-rliu: Ah, my apologies. I thought that we needed this package in the notebook tests because we also use the datasets API in the notebooks a lot. It may not be necessary after all..

bdice: I don't know if "having certifi installed" is sufficient to avoid this issue or not. Were you seeing certificate failures on notebook jobs before? Are they resolved now?

python/cugraph/cugraph/datasets/dataset.py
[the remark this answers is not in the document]
jameslamb: For those finding this from search... we talked offline and agreed that given the time-sensitivity of this, it wasn't worth another round of CI to revert this change.

dependencies.yaml
nv-rliu: You're right. Just having it installed doesn't actually do anything for the notebooks. 

I am not seeing any SSL failures for the notebooks so we can remove this dependency. Should I go ahead and do that in 24.12 or should it be in 25.02?

bdice: 25.02 please!

PR #4827, opened 2024-12-10
Fix bug in MG Neighborhood sampling

jnke2016:
This PR fixes a bug when computing the global start vertex labels


cpp/src/c_api/neighbor_sampling.cpp
```diff
@@ -948,11 +948,13 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
           std::exclusive_scan(
             recvcounts.begin(), recvcounts.end(), displacements.begin(), size_t{0});
 
-          (*label_to_comm_rank)
-            .resize(displacements.back() + recvcounts.back(), handle_.get_stream());
+          rmm::device_uvector<label_t> tmp_label_to_comm_rank(
```
ChuckHastings, line 951: Why not just use `label_to_comm_rank.resize(...)` here?

jnke2016: It is an optional which is initialized to `std::nullopt`

alexbarghi-nv (approved):
👍

cpp/src/c_api/neighbor_sampling.cpp
seunghwak:
```
label_to_comm_rank =
  rmm::device_uvector<label_t>(displacements.back() + recvcounts.back(), handle_.get_stream());
```
is sufficient (no need to create a temporary and move).

seunghwak (approved):
Besides the above comment, looks good to me.

naimnv (approved):

seunghwak (approved):

ChuckHastings (approved):

PR #4830, opened 2024-12-11
Use cuda-python bindings for getting device properties.

bdice:
This PR uses `cuda-python` for getting device properties. These APIs are more stable than getting this information via `numba.cuda`.

Companion to #4829 (this is not dependent on that PR, though).


ci/notebook_list.py
```diff
@@ -30,20 +30,27 @@
...
+def _get_cuda_version_string():
+    status, version = runtime.getLocalRuntimeVersion()
+    if status != runtime.cudaError_t.cudaSuccess:
+        raise RuntimeError("Could not get CUDA runtime version.")
+    major = version // 1000
+    minor = (version % 1000) // 10
+    return f"{major}.{minor}"
+
+
+def _is_ampere_or_newer():
```
bdice, line 45: I implemented what was here before, but I would double-check this logic: are our notebooks still failing on Ampere and newer? Does this check need to be removed?

python/cugraph/cugraph/utilities/path_retrieval_wrapper.pyx
```diff
@@ -19,7 +19,6 @@
 from cugraph.utilities.path_retrieval cimport get_traversed_cost as c_get_traversed_cost
 from cugraph.structure.graph_primtypes cimport *
 from libc.stdint cimport uintptr_t
-from numba import cuda
```
bdice, line 22: This was an unused import.

ci/notebook_list.py
```diff
@@ -30,20 +30,27 @@
...
-        if Path(runtype_dict.get(runtype)).is_file():
-            return True
-    return False
 
+def _get_cuda_version_string():
+    status, version = runtime.getLocalRuntimeVersion()
+    if status != runtime.cudaError_t.cudaSuccess:
+        raise RuntimeError("Could not get CUDA runtime version.")
+    major = version // 1000
+    minor = (version % 1000) // 10
```
jakirkham, line 41: We can consolidate this somewhat with [`divmod`]( https://docs.python.org/3/library/functions.html#divmod )

```suggestion
    major, minor = divmod(version, 1000)
    minor //= 10
```

python/cugraph/cugraph/utilities/utils.py
```diff
@@ -210,45 +207,42 @@ def get_traversed_path_list(df, id):
     return answer
 
 
-def is_cuda_version_less_than(min_version=(10, 2)):
+def is_cuda_version_less_than(min_version):
```
bdice, line 210: Based on the function name, this should not have a default value. Its default was also outdated.

This function also appears to be unused. Do we want to keep it?

python/cugraph/cugraph/utilities/utils.py
```diff
@@ -210,45 +207,42 @@ def get_traversed_path_list(df, id):
...
+    status, version = runtime.getLocalRuntimeVersion()
+    if status != runtime.cudaError_t.cudaSuccess:
+        raise RuntimeError("Could not get CUDA runtime version.")
+    major = version // 1000
+    minor = (version % 1000) // 10
+    return (major, minor) < min_version
 
 
-def is_device_version_less_than(min_version=(7, 0)):
+def is_device_version_less_than(min_version):
```
bdice, line 222: Based on the function name, this should not have a default value.

It appears this is only used once, to guard against use on Pascal. However, we dropped Pascal a year ago. Can we remove this guard on the test? Then, can we delete this function if it is unused?

https://github.com/rapidsai/cugraph/blob/e03986c6a6876be4f1143ff1f4fd58b79ce9def5/benchmarks/cugraph/pytest-based/bench_algos.py#L366

jakirkham:
It seems like the same or similar version functions are in multiple places. Could we pick one and use that in the other places?

bdice:
> It seems like the same or similar version functions are in multiple places. Could we pick one and use that in the other places?

I think some of these are needed in odd places -- things like listing compatible notebooks shouldn't require a cugraph utility function. There may be some opportunities for removal, though. See my comments above.

jakirkham:
Agreed. Maybe @dantegd can provide us some guidance 😉

KyleFromNVIDIA (approved):

python/cugraph/cugraph/utilities/utils.py
rlratzel: Removing it seems like a good idea to me.

ci/notebook_list.py
rlratzel: LGTM.  I'm in favor of committing what's here now then checking the notebooks that use the "# Does not run on Ampere" comment on Ampere to see if it's still needed. If not needed, we can have a followup PR to remove it and/or the comments.

rlratzel (approved):

PR #4833, opened 2024-12-12
[DO NOT MERGE] Experiment to leverage CUDASTF

caugonnet:


cpp/cmake/thirdparty/get_cccl.cmake
```diff
@@ -15,6 +15,13 @@
 # This function finds CCCL and sets any additional necessary environment variables.
 function(find_and_configure_cccl)
   include(${rapids-cmake-dir}/cpm/cccl.cmake)
+  include(${rapids-cmake-dir}/cpm/package_override.cmake)
+
+  rapids_cpm_package_override("${CMAKE_CURRENT_FUNCTION_LIST_DIR}/cccl_override.json")
```
caugonnet, line 20: This may be something we surround with some CUGRAPH_USE_STF option for example

cpp/tests/utilities/check_utilities.hpp
```diff
@@ -19,6 +19,7 @@
 
 #include <raft/core/handle.hpp>
 #include <raft/core/span.hpp>
+#include <thrust/extrema.h>
```
caugonnet, line 22: remove ?

PR #4836, opened 2024-12-18
heterogeneous sampling primitive

### LINKED ISSUE #4590: Optimize sampling primitives to support heterogeneous fanout

### LINKED ISSUE #4735: Update heterogeneous sampling to push functionality into the sampling primitives.
https://github.com/rapidsai/cugraph/pull/4608 provides an initial implementation of heterogeneous sampling that implemented the public API by iterating over the edge types.  If there are `n` edge types, this will result in `n` iterations over the vertices in the sampling frontier to perform the sampling, each using a different edge mask.

It should be more efficient to push the heterogeneous sampling down into the graph primitives.  That way in one pass through the vertices we can randomly sample all required edges, regardless of type.

seunghwak:
We currently use edge masking for heterogeneous sampling. This is inefficient as we need scan the entire edges # edge types times every time we call a sampling function.

This PR fixes this. We extract a subgraph (seeds and their neighbors) with edge types. Then, we perform sampling on this subgraph so no need to scan the entire edge.

This PR also updated `cugraph::test::mg_graph_to_sg_graph` function to support edge types (which is necessary to test this primitive). A large number of files (in the test directory) are modified due to this change, but updates are just few lines for each file.

Closes #4735 
Closes #4590


ChuckHastings (approved):
Looks like you're still missing some copyright updates.

Glad we could more directly incorporate the Algorithm-R code.

ChuckHastings:
To be clear... this PR updates the primitive.  But the sampling algorithm is not modified in this PR to use the primitive.  We will need to add work in 25.04 to use this primitive, correct?

jnke2016 (approved):
Looks good to me

PR #4842, opened 2024-12-31
Get breathe from conda

vyasr:
See https://github.com/rapidsai/cuvs/pull/554, the switch to pip for Breathe in #4839 is no longer required.


jameslamb (approved):
💥

Issue #4846, opened 2025-01-06
[BUG]: Incorrect code in compute_vertex_renumber_map() in src/sampling/sampling_post_processing_impl.cuh


mcordery:
### Version

25.02

### Describe the bug.

```
In compute_vertex_renumber_map() in src/sampling/sampling_post_processing_impl.cuh there is the following code:

     thrust::stable_sort(
          handle.get_thrust_policy(),
          renumber_map.begin(),
          renumber_map.end(),
          [offsets = *vertex_type_offsets] __device__(auto lhs, auto rhs)  {
            auto lhs_v_type = thrust::distance(
              offsets.begin() + 1,
              thrust::upper_bound(
                thrust::seq, offsets.begin() + 1, offsets.end(), thrust::get<0>(lhs)));
            auto rhs_v_type = thrust::distance(
              offsets.begin() + 1,
              thrust::upper_bound(
                thrust::seq, offsets.begin() + 1, offsets.end(), thrust::get<0>(rhs)));
            return lhs_v_type < rhs_v_type; 
          });

```
From this, we can deduce that the lambda parameter list should be some kind of zip_iterator (as done earlier in the file) or some kind of tuple. However, renumber_map is device_uvector<vertex_t> where vertex_t is int32_t. Thus, when compiling, the thrust::get<> operators don't know what to do with either the lhs or rhs parameters.

### Minimum reproducible example

```shell

```

### Relevant log output

```shell
/home/xxx/Projects/cugraph/cpp/src/sampling/sampling_post_processing_impl.cuh:1400:66: error: no matching function for call to 'get'
 1400 |                 thrust::seq, offsets.begin() + 1, offsets.end(), thrust::get<0>(rhs)));
      |                                                                  ^~~~~~~~~~~~~~
/opt/rocm-6.3.0/lib/llvm/bin/../../../include/thrust/detail/pair.inl:218:7: note: candidate template ignored: could not match 'pair<T1, T2>' against 'int'
  218 |       get(pair<T1,T2> &p)
      |       ^
/opt/rocm-6.3.0/lib/llvm/bin/../../../include/thrust/detail/pair.inl:226:7: note: candidate template ignored: could not match 'pair<T1, T2>' against 'int'
  226 |       get(const pair<T1,T2> &p)
      |       ^
[...]

PR #4851, opened 2025-01-07
Remove thrust::get<0>() calls on a scalar variable.

seunghwak:
Close Issue #4846

Remove `thrust::get<0>()` calls on integer type variables. This shouldn't compile but this successfully compiles in the current build environment, so this hasn't been detected.


ChuckHastings (approved):
LGTM.  Need to update copyright year to pass CI.

PR #4854, opened 2025-01-08
pylibcugraph: declare cupy and numpy hard dependencies

jameslamb:
While testing stuff for #4804, I found that `pylibcugraph` has a hard runtime dependency on `cupy` and `numpy`, but isn't declaring them
 
```shell
docker run \
    --rm \
    --gpus 0 \
    -it rapidsai/ci-wheel:latest \
    bash

python -m pip install 'pylibcugraph-cu12==25.2.*,>=0.0.0a0'
python -c "import pylibcugraph"
```

```text
  File "/pyenv/versions/3.12.7/lib/python3.12/site-packages/pylibcugraph/__init__.py", line 18, in <module>
    from pylibcugraph.graphs import SGGraph, MGGraph
  File "graphs.pyx", line 1, in init pylibcugraph.graphs
  File "utils.pyx", line 20, in init pylibcugraph.utils
ModuleNotFoundError: No module named 'cupy'
```

This declares those dependencies.

It also promotes `cugraph-service-server`'s `numpy` dependency to a hard runtime dependency.

### Evidence that `pylibcugraph` *already* has a hard dependency on these libraries

They're used unconditionally here:

https://github.com/rapidsai/cugraph/blob/cddd69ea3f62cabdb3aa2b7b6676e0b74ab4eefc/python/pylibcugraph/pylibcugraph/utils.pyx#L19-L20

But have import guards in other places:

https://github.com/rapidsai/cugraph/blob/cddd69ea3f62cabdb3aa2b7b6676e0b74ab4eefc/python/pylibcugraph/pylibcugraph/sssp.pyx#L127-L139

So this PR doesn't introduce new hard dependencies... it just makes them explicit, to make it easier to install and run `pylibcugraph`.

### How was this not caught in CI?

Import tests aren't run here for conda packages, because conda builds happen on CPU-only nodes.

https://github.com/rapidsai/cugraph/blob/cddd69ea3f62cabdb3aa2b7b6676e0b74ab4eefc/ci/build_python.sh#L27-L30

And `numpy` and `cupy` are probably getting pulled in by some of the wheels' test dependencies, like `cudf`, here:

https://github.com/rapidsai/cugraph/blob/cddd69ea3f62cabdb3aa2b7b6676e0b74ab4eefc/ci/test_wheel.sh#L17

### Should we just make the other unconditional cases conditional with try-catching?

No. Talked with @rlratzel, @ChuckHastings, and @eriknw offline, and we agreed to declare these as hard runtime dependencies (and remove the try-catching in places that had it).


python/cugraph-service/server/pyproject.toml
```diff
@@ -3,6 +3,7 @@
 [build-system]
 
 requires = [
+    "numpy>=1.23,<3.0a0",
```
jameslamb, line 6: `cugraph-service-server` should have a hard runtime dependency on `numpy`, because of this:

https://github.com/rapidsai/cugraph/blob/a6eea2f7cc40165e584e459351d1a92406de7415/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py#L30

its conda recipe does:

https://github.com/rapidsai/cugraph/blob/a6eea2f7cc40165e584e459351d1a92406de7415/conda/recipes/cugraph-service/meta.yaml#L66

rlratzel: agreed, thanks for being so thorough!

dependencies.yaml
```diff
@@ -174,6 +178,7 @@ files:
     extras:
       table: build-system
     includes:
+      - depends_on_numpy
```
bdice, line 181: Since numpy doesn’t have any matrices for CUDA version or any extra index URLs, I would treat it as a normal dependency and add it to the corresponding lists. We don’t use `depends_on_numpy` in other repos because it doesn’t reduce complexity. You could use a YAML anchor for deduplication of the pinning declaration if desired.

bdice: Example: https://github.com/rapidsai/cudf/blob/dc2a75cba40d38f4a6ba66e652764e96fa6b593d/dependencies.yaml#L661

bdice (changes requested):

dependencies.yaml
jameslamb: Ok sure. 

That is not how I've been thinking about when to do these `depends_on_*` lists. At https://github.com/rapidsai/build-planning/blob/d9e3c606d95c835ee384ac6480a4af0ac6cb024a/docs/docs/packaging.md#L181 we have this:

>  *Dependencies appearing in several lists should be in their own standalone `depends_on_{project}` lists.*

I can update those docs with this added nuance, I don't disagree with you.

Pushed https://github.com/rapidsai/cugraph/pull/4854/commits/6a83f20686c4670b66f5c42cb98dbaac01e69cac here removing `depends_on_numpy` in favor of anchors. That revealed that there were some other issues in `cugrpah-service-server`'s `[test]` extra... it was pulling through all of `cugraph's` testing dependencies, which it doesn't need based on my read of https://github.com/rapidsai/cugraph/tree/branch-25.02/python/cugraph-service/server/cugraph_service_server/testing

bdice: Yes, I think those docs should be updated. I think it should be something like:

> Dependencies appearing in several lists can be handled in two ways. Dependencies with complex requirements such as extra index URLs or package names that vary by CUDA version (like `-cuXX` suffixes) or format (different conda and PyPI names) should be in their own standalone `depends_on_{project}` lists. Simpler dependencies can use YAML anchors to avoid duplication of the pinning specs.

I'll re-review and approve based on this. Feel free to update those docs as you see fit.

bdice (approved):

dependencies.yaml
jameslamb: I like that language! Put up a PR with it here: https://github.com/rapidsai/build-planning/pull/132

PR #4855, opened 2025-01-08
Remove unnecessary CUDA utilities

bdice:
This PR removes some utilities that were updated in https://github.com/rapidsai/cugraph/pull/4830 but are no longer needed.

xref: https://github.com/rapidsai/build-planning/issues/117

### DIFF
--- a/ci/notebook_list.py
+++ b/ci/notebook_list.py
@@ -17,6 +17,8 @@ import sys
 import glob
 from pathlib import Path
 
+from cuda.bindings import runtime
+
 # for adding another run type and skip file name add to this dictionary
 runtype_dict = {
     "all": "",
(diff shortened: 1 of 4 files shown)


python/cugraph/cugraph/traversal/ms_bfs.py
```diff
@@ -56,7 +56,6 @@ def _get_feasibility(G, sources, components=None, depth_limit=None):
 
     # Fixme not implemented in RMM yet
```
bdice, line 57: This should probably use `rmm.mr.available_device_memory()` but this function is currently unreachable due to the `NotImplementedError` thrown on line 149 below.

jameslamb (approved):

rlratzel (approved):
LGTM, thanks!

ci/notebook_list.py
```diff
@@ -17,8 +17,6 @@
 import glob
 from pathlib import Path
```
bdice, line 19:
```suggestion

from cuda.bindings import runtime

```

PR #4856, opened 2025-01-09
Renumber with global renumber map too big for GPU

ChuckHastings:
New function for arbitrarily renumbering a local edgelist using a global renumber map that will not fit in GPU memory.


alexbarghi-nv (approved):
👍

cpp/src/c_api/renumber_arbitrary_edgelist.cu
```diff
@@ -0,0 +1,215 @@
...
+                        renumber_map->as_type<vertex_t>() + chunk_base_offset,
+                        size,
+                        handle.get_stream());
+
+    rmm::device_uvector<vertex_t> renumbered_values(srcs_v.size(), handle.get_stream());
+
+    thrust::fill(handle.get_thrust_policy(),
+                 renumbered_values.begin(),
+                 renumbered_values.end(),
+                 cugraph::invalid_vertex_id<vertex_t>::value);
```
seunghwak, line 78: Where are you using `renumbered_values`?

cpp/src/c_api/renumber_arbitrary_edgelist.cu
```diff
@@ -0,0 +1,215 @@
...
+                    thrust::make_zip_iterator(dsts_v.begin(), dsts_pos.begin()),
+                    thrust::remove_if(handle.get_thrust_policy(),
+                                      thrust::make_zip_iterator(dsts_v.begin(), dsts_pos.begin()),
+                                      thrust::make_zip_iterator(dsts_v.end(), dsts_pos.end()),
+                                      [] __device__(auto t) {
+                                        return thrust::get<0>(t) ==
+                                               cugraph::invalid_vertex_id<vertex_t>();
+                                      })),
+                  handle.get_stream());
+    dsts_pos.resize(dsts_v.size(), handle.get_stream());
```
seunghwak, line 148: I feel like scanning src_v & dst_v multiple times sounds a bit inefficient.

We already sorted srcs & dsts, so it is easy to find unique srcs & dsts, and accordingly unique vertices. Then, for each chunk, we can call thrust::copy_if to find only the relevant (old_v, new_v) pairs. If we have this renumber map, we can just directly renumber so no need for position vectors as well.

cpp/src/c_api/renumber_arbitrary_edgelist.cu
ChuckHastings: Deleted.

cpp/src/c_api/renumber_arbitrary_edgelist.cu
ChuckHastings: Refactored based on this guidance.

cpp/src/c_api/renumber_arbitrary_edgelist.cu
```diff
@@ -0,0 +1,190 @@
...
+      thrust::make_counting_iterator<size_t>(0),
+      thrust::make_counting_iterator<size_t>(renumber_chunk.size()),
+      [chunk_base_offset,
+       renumber_chunk_span =
+         raft::device_span<vertex_t const>{renumber_chunk.data(), renumber_chunk.size()},
+       vertices_span,
+       ids_span] __device__(size_t idx) {
+        auto pos = thrust::lower_bound(
+          thrust::seq, vertices_span.begin(), vertices_span.end(), renumber_chunk_span[idx]);
+        if (pos != vertices_span.end()) {
```
seunghwak, line 91: Shouldn't we check `&& *pos = renumber_chunk_span[idx]` as well?

ChuckHastings: Yes.  Will fix.

ChuckHastings: Just pushed update

seunghwak (approved):

PR #3584, opened 2023-05-19
Sampling Performance Testing

alexbarghi-nv:
Adds performance benchmarking scripts for testing MNMG cuGraph GNN workflows.
This branch is the head branch for the cuGraph benchmarking effort.  All work supporting the benchmarks should be merged into this branch.  It will be merged into branch-24.02 once all features are ready.

Includes patches to cuGraph-PyG required for the latest DLFW container.

To-Do:

- [x] Refactor for branch-24.02
- [x] <s>Add WholeGraph training portion</s> Deferred to future PR (see https://github.com/alexbarghi-nv/cugraph/pull/6)
- [x] <s>Add WholeGraph generators</s> Included in above
- [x] <s>Support DGL</s> Deferred to future PR
- [x] <s>Use appropriate docker containers</s> Deferred, waiting on DLFW release

Closes #3839

PR #4863, opened 2025-01-13
Switch to `pynvml_utils.smi` for PyNVML 12

jakirkham:
In the PyNVML 12 upgrade, the `pynvml` package now depends on `nvidia-ml-py`. However `nvidia-ml-py` does not supply the `smi` submodule. To avoid clobbering the package, this module was moved from the `pynvml` namespace to `pynvml_utils`. As a result the `import` here needs the same tweak.

### DIFF
--- a/benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
+++ b/benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
@@ -36,9 +36,7 @@ from cugraph.testing.mg_utils import enable_spilling
 def init_pytorch_worker(rank: int, use_rmm_torch_allocator: bool = False) -> None:
     import cupy
     import rmm
-    from pynvml_utils.smi import nvidia_smi
 
-    smi = nvidia_smi.getInstance()
     pool_size = 16e9  # FIXME calculate this
 
     rmm.reinitialize(
--- a/benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
+++ b/benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2023-2024, NVIDIA CORPORATION.
+# Copyright (c) 2023-2025, NVIDIA CORPORATION.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -36,7 +36,7 @@ from cugraph.testing.mg_utils import enable_spilling
 def init_pytorch_worker(rank: int, use_rmm_torch_allocator: bool = False) -> None:
     import cupy
     import rmm
-    from pynvml.smi import nvidia_smi
+    from pynvml_utils.smi import nvidia_smi
 
     smi = nvidia_smi.getInstance()
     pool_size = 16e9  # FIXME calculate this
--- a/benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
+++ b/benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
@@ -201,7 +201,6 @@ class PyGTrainer(Trainer):
                         )
                         logger.info(f"total time: {total_time_iter}")
 
-
                     y_true = data.y
                     y_true = y_true.reshape((y_true.shape[0],))
                     x = data.x.to(torch.float32)
--- a/benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
+++ b/benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
@@ -201,9 +201,6 @@ class PyGTrainer(Trainer):
                         )
                         logger.info(f"total time: {total_time_iter}")
 
-                        # from pynvml_utils.smi import nvidia_smi
-                        # mem_info = nvidia_smi.getInstance().DeviceQuery('memory.free, memory.total')['gpu'][self.rank % 8]['fb_memory_usage']
-                        # logger.info(f"rank {self.rank} memory: {mem_info}")
 
                     y_true = data.y
                     y_true = y_true.reshape((y_true.shape[0],))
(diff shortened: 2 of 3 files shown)


python/utils/gpu_metric_poller.py
```diff
@@ -31,7 +31,7 @@
 import os
 import sys
 import threading
-from pynvml import smi
+from pynvml_utils import smi
```
rjzamora, line 34: Pretty sure we can just do `import pynvml`, and replace all `smi.` instances below with `pynvml.`.

In other words, non of the logic in this file is actually using "smi". As far as I can tell, the only reason this works now is that `smi.py` includes a `from pynvml import *` statement.

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
```diff
@@ -36,7 +36,7 @@
 def init_pytorch_worker(rank: int, use_rmm_torch_allocator: bool = False) -> None:
     import cupy
     import rmm
-    from pynvml.smi import nvidia_smi
+    from pynvml_utils.smi import nvidia_smi
 
     smi = nvidia_smi.getInstance()
```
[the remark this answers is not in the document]
jakirkham: Thanks Rick! 🙏

Also am not seeing uses later either

Are there situations where this would be used for its side effects?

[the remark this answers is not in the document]
jakirkham: Looks like this was added originally in @alexbarghi-nv 's PR: https://github.com/rapidsai/cugraph/pull/3584

Appears the `smi` object wasn't used there either (unless there is some side effect we are missing)

Hopefully Alex can let us know whether this is still needed

python/utils/gpu_metric_poller.py
jakirkham: Not sure this is true. Am seeing quite a lot of `smi` usage in this file. For example

https://github.com/rapidsai/cugraph/blob/e3054167d596927f94c5a62fba773d4d144b796a/python/utils/gpu_metric_poller.py#L94-L98

rjzamora: Right - I'm saying that all those calls are actually using `pynvml` (not `smi`).

For example, calling `smi.nvmlInit()` is actually calling `pynvml.nvmlInint()` (the main hint is that all commands start with `nvml`).

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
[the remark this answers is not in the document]
rjzamora: I suppose it's possible that there is some side effect I'm not considering, but nothing comes to mind. If so, there should probably be a comment explaining that the import is necessary.

python/utils/gpu_metric_poller.py
jakirkham: Ah ok. Thanks for clarifying! 🙏

Have made this change below

python/utils/gpu_metric_poller.py
```diff
@@ -147,34 +147,3 @@ def startGpuMetricPolling():
...
-    if gpuUtil > self.maxGpuUtil:
-        self.maxGpuUtil = gpuUtil
-
-    smi.nvmlShutdown()
-"""
-
-
-# if __name__ == "__main__":
-#     sto=stopGpuMetricPolling
-#     po = startGpuMetricPolling()
```
jakirkham, line 180: This code was unused (all commented or quoted). So went ahead and just dropped it instead of refreshing it

benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
```diff
@@ -201,7 +201,7 @@ def train(self):
                         )
                         logger.info(f"total time: {total_time_iter}")
 
-                        # from pynvml.smi import nvidia_smi
+                        # from pynvml_utils.smi import nvidia_smi
                         # mem_info = nvidia_smi.getInstance().DeviceQuery('memory.free, memory.total')['gpu'][self.rank % 8]['fb_memory_usage']
                         # logger.info(f"rank {self.rank} memory: {mem_info}")
```
jakirkham, line 206: Is this actually used or can we drop this?

```suggestion
```

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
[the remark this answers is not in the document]
alexbarghi-nv: It's not needed.  This code is slated for removal anyways.  It's just waiting on rapidsai/cugraph-gnn#82 to merge

alexbarghi-nv (approved):
👍

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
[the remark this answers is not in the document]
jakirkham: Ok removing this below

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
```diff
@@ -36,7 +36,7 @@
 def init_pytorch_worker(rank: int, use_rmm_torch_allocator: bool = False) -> None:
     import cupy
     import rmm
-    from pynvml.smi import nvidia_smi
+    from pynvml_utils.smi import nvidia_smi
```
jakirkham, line 39:
```suggestion
```

benchmarks/cugraph/standalone/bulk_sampling/bench_cugraph_training.py
jakirkham, line 41:
```suggestion
```

jakirkham (commented):
Adding suggestion to remove unused `smi`  code below

benchmarks/cugraph/standalone/bulk_sampling/trainers/pyg/trainers_pyg.py
```diff
@@ -201,9 +201,6 @@ def train(self):
                         )
                         logger.info(f"total time: {total_time_iter}")
```
jakirkham, line 203: Suggesting dropping extra blank line

```suggestion
```

jakirkham:
Thanks for the feedback! 🙏

Cleaned up the bits discussed above. Please take another look when you have a moment

PR #4865, opened 2025-01-14
Removes unused benchmark infra: rapids-pytest-benchmark and asvdb

rlratzel:
This PR removes `rapids-pytest-benchmark` and references to one of its dependencies, `asvdb`.

These packages were used for adding GPU memory usage to pytest-benchmark output and for updating ASV output files with pytest-benchmark results so the ASV front end can be used on pytest-benchmark suites.

Since we're not using the ASV frontend anymore, and the GPU memory usage mechanism used by rapids-pytest-benchmark is outdated and unmaintained (and slower than other tools available), we're removing references to these packages.

cc @vyasr @robertmaynard


acostadon (approved):
LGTM

alexbarghi-nv (approved):
Looks good 👍

jakirkham:
Thanks all! 🙏

PR #4867, opened 2025-01-17
Include a thrust header that was missing to use thrust::max

caugonnet:
The <thrust/extrema.h> header is needed to use thrust::max so we include it to avoid this error

cpp/tests/utilities/check_utilities.hpp: In member function ‘bool cugraph::test::device_nearly_equal<type_t>::operator()(type_t, type_t) const’:
cpp/tests/utilities/check_utilities.hpp:98:20: error: ‘max’ is not a member of ‘thrust’
98 | thrust::max(thrust::max(lhs, rhs) * threshold_ratio, threshold_magnitude);
| ^~~

### DIFF
--- a/cpp/tests/utilities/check_utilities.hpp
+++ b/cpp/tests/utilities/check_utilities.hpp
@@ -19,6 +19,7 @@
 
 #include <raft/core/handle.hpp>
 #include <raft/core/span.hpp>
+#include <thrust/extrema.h>
 
 #include <numeric>
 #include <type_traits>


cpp/tests/utilities/check_utilities.hpp
ChuckHastings, line 1: Need to update the copyright date on the file.

ChuckHastings (approved):
Also looks like you need to merge the latest 25.02 changes in order to pass CI.

caugonnet:
Per discussions with @seunghwak an even better solution is to get rid of thrust::max and use (cuda::)std::max nowadays

seunghwak:
https://github.com/rapidsai/cugraph/pull/4871 should replace this PR.

ChuckHastings:
Closed in favor of #4871

PR #4870, opened 2025-01-17
Added more doxygen groups and fixed some existing ones

acostadon:
Added more doxygen groups and fixed some existing ones


seunghwak (approved):
LGTM

PR #4871, opened 2025-01-17
Replace thrust::min/max with cuda::std::min/max

seunghwak:
cuda::std::min/max is a more C++ standard conformant way of computing min/max of two values than thrust::min/max.

We are using cuda::std::min/max in the src directory but thrust::min/max is still used in some files under the test directory.

This PR replaces thrust::min/max calls in the test directory with cuda::std::min/max.


ChuckHastings (approved):
Looks like CI wants a different order for the header files.

PR #4879, opened 2025-01-22
Revert "Temporarily skip CUDA 11 wheel CI"

bdice:
Reverts rapidsai/cugraph#4876 now that https://github.com/rapidsai/raft/pull/2548 has landed.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -174,9 +174,6 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_wheel_pylibcugraph.sh
-      # CUDA 11 wheel CI is disabled until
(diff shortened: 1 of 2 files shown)


jameslamb:
The re-runs of nightly tests failed like this:

> E   ImportError: libcugraph.so: cannot open shared object file: No such file or directory

Suspect that might be a bad mix of packages build at different time while rolling out the fixes from https://github.com/rapidsai/raft/pull/2548. If all other CI passes here on this PR, I think it'd be ok to admin-merge this.

And then I'll trigger a new branch test build after the packages are published, and confirm that passes.

vyasr:
@jameslamb yeah I agree. Let's go ahead and get this merged and try a new build.

PR #4883, opened 2025-01-22
Introduce options to disable some algorithms in cmake and reduce compilation time

caugonnet:


cpp/CMakeLists.txt
```diff
@@ -429,6 +301,175 @@ set(CUGRAPH_SOURCES
...
+    src/sampling/negative_sampling_mg_v64_e64.cu
+    src/sampling/sampling_post_processing_sg_v64_e64.cu
+    src/sampling/sampling_post_processing_sg_v32_e32.cu
+    src/sampling/detail/conversion_utilities.cu
+    src/sampling/random_walks_mg_v64_e64.cu
+    src/sampling/random_walks_mg_v32_e32.cu
+)
+endif()
+
+if (BUILD_CUGRAPH_CENTRALITY_ALGORITHMS)
```
caugonnet, line 403: maybe it's cleaner to define CUGRAPH_CENTRALITY_SOURCES, CUGRAPH_COMMON_SOURCES, etc.... and then append at the end ?

ChuckHastings: This seems cleaner to me.  Define `CUGRAPH_CENTRALITY_SOURCES` as empty if `BUILD_CUGRAPH_CENTRALITY_ALGORITHMS` is not defined, define it with the actual sources if it is defined.  Then when you fill in `CUGRAPH_SOURCES` you can just include the variable `${CUGRAPH_CENTRALITY_SOURCES}` in its definition.

cpp/tests/CMakeLists.txt
```diff
@@ -330,6 +330,7 @@ ConfigureTest(GRAPH_GENERATORS_TEST generators/generators_test.cpp)
 # - erdos renyi graph generator tests -------------------------------------------------------------
 ConfigureTest(ERDOS_RENYI_GENERATOR_TEST generators/erdos_renyi_test.cpp)
 
+if (BUILD_CUGRAPH_COMMUNITY_ALGORITHMS)
```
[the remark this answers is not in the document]
caugonnet: And i'll create a "COMMON" class too for everything that's always there

PR #4885, opened 2025-01-23
Update MG negative sampling to return random samples distributed as specified

### LINKED ISSUE #4672: Update Negative Sampling to request edges per worker
Instead of accepting an argument to indicate how many total negative samples to generate, allow each worker to specify how many they want.

Work with @alexbarghi-nv to identify the API change and implementation.

ChuckHastings:
Modifies the new negative sampling interface so that when called from MG, each rank specifies how many samples they wish to receive, and to randomly distribute the samples across the calling GPUs.

Marked breaking as it changes the C++ interface... although nothing uses it yet.

Closes #4672


seunghwak (commented):
Shouldn't we update the documentation?

ChuckHastings:
> Shouldn't we update the documentation?

Just pushed an update to the documentation.

alexbarghi-nv:
Can we hold this PR until I have a corresponding one in cuGraph-GNN?

cpp/src/sampling/negative_sampling_impl.cuh
```diff
@@ -412,16 +431,152 @@ std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>> negativ
...
+      while (reduction > 0) {
+        size_t est_reduction_per_gpu = (reduction + comm_size - 1) / comm_size;
+        for (size_t i = 0; i < samples_per_gpu.size(); ++i) {
+          if (samples_per_gpu[i] > est_reduction_per_gpu) {
+            samples_per_gpu[i] -= est_reduction_per_gpu;
+            reduction -= est_reduction_per_gpu;
+          } else {
+            reduction -= samples_per_gpu[i];
+            samples_per_gpu[i] = 0;
+          }
```
seunghwak, line 469: I think this logic has a flaw.

Say reduction = 3 & comm_size  = 2.
Then, est_reduction_per_gpu = 2
If samples_per_gpu[] > 2, reduction = 3 - 2 - 2 = overflow!

cpp/src/sampling/negative_sampling_impl.cuh
```diff
@@ -412,16 +431,152 @@ std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<vertex_t>> negativ
...
+      handle.get_stream());
+    reduced_counts.resize(reduced_ranks.size(), handle.get_stream());
+
+    rmm::device_uvector<size_t> send_count(comm_size, handle.get_stream());
+    thrust::fill(handle.get_thrust_policy(), send_count.begin(), send_count.end(), 0);
+    thrust::scatter(handle.get_thrust_policy(),
+                    reduced_counts.begin(),
+                    reduced_counts.end(),
+                    reduced_ranks.begin(),
+                    send_count.begin());
```
seunghwak, line 521: gpu_assignment is already sorted.

We can use thrust::upper_bound to find boundaries. This will require just comm_size binary searches and will be much faster.

ChuckHastings: Fixed in the next push.

cpp/src/sampling/negative_sampling_impl.cuh
seunghwak: > reduce set_reduction_per_gpu to 1.

You mean est_reduction_per_gpu? This is set before the for loop. Once it is set to 2 in line 460, it will be 2 till the end of the for loop in line 416. Am I missing something?

seunghwak: Oh, sorry, I missed the line 470.

seunghwak (approved):
LGTM

alexbarghi-nv:
@ChuckHastings doesn't this affect negative sampling in the C API?  So once this merges, I'll have to update `cugraph-pyg` to call this function with the number of samples on each GPU.

ChuckHastings:
> @ChuckHastings doesn't this affect negative sampling in the C API? So once this merges, I'll have to update `cugraph-pyg` to call this function with the number of samples on each GPU.

This changes the semantic meaning of the parameter.  Prior to this change, if you wanted this to run against an MG graph you would specify the total number of samples across all GPUs as the parameter for `num_samples` on each GPU (the same value on each GPU).  With this change, each GPU should specify the number of samples that they want.

So, depending on what you implemented in `cugraph-pyg`, you may need to change the value that's passed for `num_samples`, yes.

PR #4888, opened 2025-01-23
Fix issues when building with upcoming cccl

miscco:
We are removing a lot of deprecated thrust features, so replace them by the equivalent `cuda::std` ones


bdice (approved):

seunghwak (approved):
LGTM

PR #4891, opened 2025-01-23
Replace all uses of `thrust::optional` with `cuda::std::optional`

miscco:
`thrust::optional` is being deprecated and will be removed in a future version of CCCL

### DIFF
--- a/cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
+++ b/cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
@@ -29,10 +29,11 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <cuda/std/optional>
 #include <thrust/sort.h>
 #include <thrust/tuple.h>
 #include <thrust/unique.h>
+
+#include <optional>
 
 namespace cugraph {
 namespace detail {
--- a/cpp/src/sampling/detail/remove_visited_vertices_from_frontier.cuh
+++ b/cpp/src/sampling/detail/remove_visited_vertices_from_frontier.cuh
@@ -19,10 +19,11 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <cuda/std/optional>
 #include <thrust/binary_search.h>
 #include <thrust/remove.h>
 #include <thrust/tuple.h>
+
+#include <optional>
 
 namespace cugraph {
 namespace detail {
--- a/cpp/src/sampling/detail/remove_visited_vertices_from_frontier_sg_v32_e32.cu
+++ b/cpp/src/sampling/detail/remove_visited_vertices_from_frontier_sg_v32_e32.cu
@@ -20,10 +20,11 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <cuda/std/optional>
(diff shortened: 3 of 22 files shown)


cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -29,7 +29,7 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <thrust/optional.h>
+#include <cuda/std/optional>
```
ChuckHastings, line 32: Same comment

cpp/src/sampling/detail/remove_visited_vertices_from_frontier.cuh
```diff
@@ -19,8 +19,8 @@
 
 #include <rmm/device_uvector.hpp>
 
+#include <cuda/std/optional>
```
ChuckHastings, line 22: Same comment

cpp/src/sampling/detail/remove_visited_vertices_from_frontier_sg_v32_e32.cu
```diff
@@ -20,8 +20,8 @@
 
 #include <rmm/device_uvector.hpp>
 
+#include <cuda/std/optional>
```
ChuckHastings, line 23: Same comment

cpp/src/sampling/detail/remove_visited_vertices_from_frontier_sg_v64_e64.cu
```diff
@@ -20,8 +20,8 @@
 
 #include <rmm/device_uvector.hpp>
 
+#include <cuda/std/optional>
```
ChuckHastings, line 23: Same comment

cpp/src/sampling/detail/sampling_utils.hpp
```diff
@@ -22,7 +22,7 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <thrust/optional.h>
+#include <cuda/std/optional>
```
ChuckHastings, line 25: Same comment

cpp/src/sampling/detail/shuffle_and_organize_output_impl.cuh
```diff
@@ -32,7 +32,7 @@
 
 #include <rmm/device_uvector.hpp>
 
-#include <thrust/optional.h>
+#include <cuda/std/optional>
```
ChuckHastings, line 35: Same comment

ChuckHastings (approved):
Thanks for this update.

I identified several files, tagged with comments, where it looks like we could just drop the include file rather than changing it.  If you're willing to make those changes (and they compile) that would be great.  If you'd rather we do that, I can put it on the list to review those changes in 25.04.

miscco:
@ChuckHastings Thanks a lot for the review.

I went ahead and checked those files. It seems that those were actually using `std::optional` So I replaced the include with `<optional>` when appropriate and verified that each file properly includes `<cuda/std/optional>` when needed

seunghwak (approved):
LGTM

PR #4901, opened 2025-01-29
Suppress maybe unitialized warnigns found with CTK 12.8

robertmaynard:


seunghwak:
Do you plan to just suppress the warning using a compiler flag or update the code to work-around? If the latter, this work may overlap with https://github.com/rapidsai/cugraph/pull/4924.

PR #4903, opened 2025-01-30
Fixed links that pointed to old docs/source and removed perf page

acostadon:
Removed perf page and fixed links that pointed to old docs source that are moved to the new docs repository

### DIFF
--- a/.devcontainer/cuda11.8-pip/devcontainer.json
+++ b/.devcontainer/cuda11.8-pip/devcontainer.json
@@ -5,7 +5,7 @@
     "args": {
       "CUDA": "11.8",
       "PYTHON_PACKAGE_MANAGER": "pip",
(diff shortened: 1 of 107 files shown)


nv-rliu (commented):
Question: did we get rid of all the pages within cugraph and decide to move everything over to docs.rapids.ai during 25.02?

nv-rliu (approved):

acostadon:
Question: did we get rid of all the pages within cugraph and decide to move everything over to docs.rapids.ai during 25.02?

There is some developer content remainingin the readme files. I don't think that needs to move over.
I did find 3 more things I need to move over while doing this latest link fixing. I am creating issues for them.
1) move over the GNN readme content to the docs
2) duplicate ( or maybe just move ) the detailed contributing.md page 
3) figure out if news should be a thing at the cugraph level. The stuff in there now is useful and needs to be preserved.

BradReesWork (approved):

PR #4905, opened 2025-01-30
Migrate to NVKS for amd64 CI runners

bdice:
This migrates amd64 CI jobs (PRs and nightlies) to use L4 GPUs from the NVKS cluster.

xref: https://github.com/rapidsai/build-infra/issues/184

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.04
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@nvks-runners
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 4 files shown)


AyodeAwe (approved):

cpp/tests/CMakeLists.txt
bdice, line 1: @ChuckHastings I see some C++ tests are failing here with `std::bad_alloc: out_of_memory: RMM failure at:/opt/conda/envs/test/include/rmm/mr/device/pool_memory_resource.hpp:276: Maximum pool size exceeded`.

The new NVKS CI will be almost entirely L4 GPUs, which have 24 GB of memory. Previously our V100 GPUs had 32 GB of memory. I pushed 5adbc4c to address this. Just wanted you to be aware of the change.

ChuckHastings: Thanks.  Will this also affect 25.02 CI?

ChuckHastings: It is failing in 25.02 PRs with the same error.  Can we target this to 25.02 instead?

bdice: No. We are only migrating to NVKS in 25.04. 25.02 will not be affected.

ChuckHastings (approved):

PR #4911, opened 2025-02-04
Use new rapids-logger library

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/104.


python/libcugraph/libcugraph/load.py
```diff
@@ -45,19 +45,17 @@ def _load_wheel_installation(soname: str):
 def load_library():
     """Dynamically load libcugraph.so and its dependencies"""
     try:
-        # libraft must be loaded before libcugraph because libcugraph
-        # references its symbols
+        # These libraries  must be loaded before libcugraph because libcugraph
```
bdice, line 48:
```suggestion
        # These libraries must be loaded before libcugraph because libcugraph
```

python/libcugraph/libcugraph/load.py
```diff
@@ -45,19 +45,17 @@ def _load_wheel_installation(soname: str):
...
+        # references their symbols
+        import librmm
         import libraft
```
bdice, line 51: Note: some of the logger PRs use alphabetical order here, while others use dependency tree order. Make this consistent.

bdice (approved):

jameslamb (approved):
Approving, assuming the CI-specific testing stuff will be removed.

python/libcugraph/libcugraph/load.py
vyasr: I tried to stick to dependency ordering.

ChuckHastings (approved):

rlratzel (approved):

PR #4923, opened 2025-02-06
Enable MTMG tests in CI

### LINKED ISSUE #4915: Enable MTMG tests in CI
Recent changes caused MTMG tests to fail to compile.  They are not tested in CI, so the compilation step is skipped.

Enable this in CI so we don't get code divergence.

ChuckHastings:
Update MTMG software to support temporal graphs and add MTMG compilation to CI testing.

Closes #4915


seunghwak (approved):
LGTM

conda/recipes/libcugraph/conda_build_config.yaml
```diff
@@ -62,3 +62,6 @@ cuda11_cuda_profiler_api_host_version:
 
 cuda11_cuda_profiler_api_run_version:
   - ">=11.4.240,<12"
+
+libucxx_version:
+  - "0.43.*"
```
jameslamb, line 67: This needs a corresponding entry in `update-version.sh`, so this version will be updated when new release branches are cut. Here's how RAFT does that: https://github.com/rapidsai/raft/blob/68d412a119f918dc003ecd0c430b6bde7499bc96/ci/release/update-version.sh#L77

Alternatively, since this is only being used in 1 place, you could forgo the `conda_build_config.yaml` entry and just directly put the version in `conda/recipes/libcugraph/meta.yaml`, like this:

```yaml
- libucxx =0.43.*
```

I recommend just doing that because it's simpler, but either would be fine.

jameslamb (approved):
This needs one change, but approving because that suggestion is something small and not time-sensitive. If you push another commit here for other reasons, please address it. Otherwise, I can do it in a follow-up PR (together with https://github.com/rapidsai/cugraph/pull/4935#discussion_r1956777486).

conda/recipes/libcugraph/conda_build_config.yaml
jakirkham: Went ahead and pushed the `update-version.sh` bits to Chuck's PR

jakirkham:
Since Kyle's PR ( https://github.com/rapidsai/cugraph/pull/4935 ) landed, merged that in here and pushed a fix for James' review comment

Hoping we can get CI moving more quickly here during a lull

KyleFromNVIDIA (approved):

PR #4926, opened 2025-02-06
Replace `thrust::tuple` with `cuda::std::tuple`

miscco:
The latter is more or less deprecated and we want to move towards standard types


miscco:
I just found some issues I need to address on the CCCL side first

PR #4930, opened 2025-02-10
Replace `thrust::tuple` with `cuda::std::tuple`

miscco:
The latter is more or less deprecated and we want to move towards standard types

While we are at it also clean up the pattern of `make_zip_iterator(make_tuple(...))`


miscco:
This will definitely need to wait until we are ready to move to CCCL 3.0

ChuckHastings:
I'm going to guess that this will be easier to redo than to catch up at this point.  Thoughts @miscco ?

I can look at this after I've finished some of our compile time improvements.

ChuckHastings:
Closing this PR, it will be much easier to redo this work from the latest code base than try and catch this PR up.  Created https://github.com/rapidsai/cugraph/issues/5141 to track this work.

PR #4931, opened 2025-02-10
Use raft::host_span instead of const reference to std::vector

seunghwak:
This PR updates functions to consistently take raft::host_span instead of std::vector const& (we have been mixing the two); except for public functions in graph_functions.hpp.

Marked as breaking as this PR updates functions under include/cugraph/utilities/device_comm.hpp,shuffle_comm.cuh, but we don't expect public users to directly call these utility functions.


ChuckHastings (approved):

jnke2016 (approved):

miscco:
Question: why dont we use `cuda::std::span` it is available in all supported standard modes

seunghwak:
> Question: why dont we use `cuda::std::span` it is available in all supported standard modes

Oh, is it available now? I was thinking about using raft::host_span/device_span first and switch to cuda::std::span (and std::span which is from C++20) once they become available.

I will replace raft span with cuda::std::span in a separate PR. How do we distinguish host_span and device_span in cuda::std::span?

miscco:
> I will replace raft span with cuda::std::span in a separate PR. How do we distinguish host_span and device_span in cuda::std::span?

There is currently no way, but AFAIK raft also only has type aliases which boil down to the same type

seunghwak:
> > I will replace raft span with cuda::std::span in a separate PR. How do we distinguish host_span and device_span in cuda::std::span?
> 
> There is currently no way, but AFAIK raft also only has type aliases which boil down to the same type

https://github.com/rapidsai/raft/blob/branch-25.04/cpp/include/raft/core/span.hpp#L50

In raft, `bool is_device` is a non-type template parameter, and `host_span` and `device_span` are different types. I feel like we need to distinguish host span and device span (unless GPUs can directly access host_span data; like GH200 or GB200).

I think we will stay little longer with raft::host_span/device_span till we get a solution for this (maybe cuda::std::span and std::span).

PR #4942, opened 2025-02-14
[BUG] Fix Empty Batch Fill by Allowing Specification of Empty Tensor

alexbarghi-nv:
The `DistSampler` does not handle empty edge index tensors correctly.  Instead, it creates a 1D tensor which cannot be 2D-indexed.  When encountering an empty batch, the incorrectly dimensioned tensor causes an indexing error.

This PR fixes this bug by adding an option to specify the tensor to use when filling empty batches.  This could be useful for resolving other issues as well (i.e. using it to pass in a default batch).  It specifies the correct tensor when calling `sample_from_edges`.


alexbarghi-nv:
OBE

PR #4934, opened 2025-02-12
Account for breaking changes in RAFT API

divyegala:
We inadvertently broke cugraph by merging this PR https://github.com/rapidsai/raft/pull/2541.

Raised this PR in RAFT to fix the issues https://github.com/rapidsai/raft/pull/2581.

PR #4943, opened 2025-02-14
Fix spectral clustering tests

miscco:
raft has changed their API slightly to allow for larger arrays.

We need to properly pass the index size otherwise we end up with type conflicts.

NOTE: I was a tiny bit lazy and just passed the previous default. We might get around with specifying less but ... :shrug:


miscco:
closing in favor of #4934

PR #4945, opened 2025-02-14
[WIP] Find the forest in an undirected graph

seunghwak:
Found a new bug, debugging ongoing.

This PR implements a function to find a forest in the input undirected graph (without multi-edges & self-loops).

This is a pre-requisite to implement forest pruning in BFS.


seunghwak:
This was initially introduced for BFS forest pruning, but efficient implementation of forest pruning for BFS is too BFS specific. Better not include in the public API.

PR #4946, opened 2025-02-18
Update primitives to better follow STL/thrust conventions in supporting predicate operations

seunghwak:
STL/thrust algorithms that conditionally process array elements have naming convention including `if` (e.g. copy_if, transform_if, ...) and takes a separate predicate operator.

cuGraph C++ primitives have been taking a single operator that returns cuda::std::nullopt (to discard the result) and a valid cuda::std::optional object for elements to be processed.

This PR updates cuGraph C++ primitives to take a separate predicate operator and use `if` in primitive names to better follow STL/thrust convention.


KyleFromNVIDIA (approved):
Approved trivial CMake changes

ChuckHastings (approved):

PR #4948, opened 2025-02-25
Replace `cub::Min` with `cuda::minimum`

miscco:
The former is deprecated and will be removed in a future CCCL release.


ChuckHastings (approved):

ChuckHastings:
Are we missing a header?  Everything's failing to build.

miscco:
Oh sorry that was my bad, we only added that in 2.8 and rapids still builds 2.7 🤦‍♂️

miscco:
@ChuckHastings I made it work, could you have a look again?

cpp/src/prims/detail/per_v_transform_reduce_e.cuh
```diff
@@ -589,6 +589,12 @@ __global__ static void per_v_transform_reduce_e_mid_degree(
       }
     }
 
+#if CCCL_MAJOR_VERSION >= 3
```
ChuckHastings, line 592: Could you add the following comment:

// FIXME: Remove CUB reference once upgraded to CCCL version 3.x

or something like that.  We use FIXME comments to remind ourselves of cleanup work to do later.

miscco:
@ChuckHastings This is now green, can we merge it?

PR #4950, opened 2025-02-25
Use conda-build instead of conda-mambabuild

bdice:
This changes from `conda mambabuild` to `conda build`. Conda now uses the mamba solver so no performance regressions are expected.

This is a temporary change as we plan to migrate to `rattler-build` in the near future. However, this is needed sooner to drop `boa` and unblock Python 3.13 migrations.

xref: https://github.com/rapidsai/build-planning/issues/149

### DIFF
--- a/ci/build_cpp.sh
+++ b/ci/build_cpp.sh
@@ -17,7 +17,7 @@ rapids-logger "Begin cpp build"
 
 sccache --zero-stats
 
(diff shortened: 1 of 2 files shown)


bdice:
Nothing stands out as a root cause. Even if there is a problem:
- we're migrating to rattler build and can prioritize cuvs/cugraph
- we can update our conda images to a newer conda once we finish this migration, which may also help

PR #4951, opened 2025-02-26
Reduce downloaded test data in CI

ChuckHastings:
Attempting to improve testing times in CI.  The tests run in CI do not require all of the test data that is downloaded.  We should be able to skip calling `get_test_data.sh` since we're not running any of the benchmark tests.

This PR also fixes a few missing annotations in the python tests discovered during test.

There is still more work to be done, but that will be in a follow-on PR.


gforsyth (approved):

rlratzel (approved):
Thanks!

PR #4952, opened 2025-02-27
[DO NOT MERGE] code annotation for performance measurement

seunghwak:
Just to work with @alexbarghi-nv for performance diagnosis.

Will be closed after diagnosis.


alexbarghi-nv:
Thanks!  I'll test this out tomorrow

PR #4958, opened 2025-03-05
Use CuPy for arrays & CUDA Python for low-level CUDA functions

jakirkham:
Switch to using CuPy for array creation when needed. Can still pass these to Numba or use `__cuda_array_interface__` for low-level operations.

Also switch to CUDA Python for low-level CUDA functions calls in Python.


ChuckHastings:
@jakirkham - looks like there are ripple effects into the python testing.

python/cugraph/cugraph/layout/force_atlas2_wrapper.pyx
rlratzel, line 1: Thanks. FYI this file will be going away in 25.06 due to support being added in `pylibcugraph` for 25.04. cc @jnke2016

rlratzel (approved):

PR #4960, opened 2025-03-07
Silence compiler warnings about host device destructor

miscco:
We are seeing spurious compiler warnings like

> error: calling a __host__ function("std::vector<...> ::~vector()") from a __host__ __device__ function("cugraph::edge_property_view_t<...> ::~edge_property_view_t") is not allowed

This is clearly a compiler bug that has been fixed but affects 12.5 which is what runs in CI

Example [here](https://github.com/NVIDIA/cccl/actions/runs/13720179287/job/38373997200?pr=3953#step:6:12629)

### DIFF
--- a/cpp/include/cugraph/edge_property.hpp
+++ b/cpp/include/cugraph/edge_property.hpp
@@ -52,6 +52,9 @@ class edge_property_view_t {
   {
   }
 
+  // nvcc 12.5 sometimes deduces this as a host device function, just defining it fixes that
+  ~edge_property_view_t() {}
+
   std::vector<ValueIterator> const& value_firsts() const { return edge_partition_value_firsts_; }
 
   std::vector<edge_t> const& edge_counts() const { return edge_partition_edge_counts_; }


seunghwak:
Do you know when we will switch to 12.8 in CI? I am wondering whether we should add a hack to silence a spurious warning or if we are switching to 12.8 sometime soon, just better wait. I am locally using CUDA 12.8 and I don't see this warning.

miscco:
> Do you know when we will switch to 12.8 in CI? I am wondering whether we should add a hack to silence a spurious warning or if we are switching to 12.8 sometime soon, just better wait. I am locally using CUDA 12.8 and I don't see this warning.

Probably only when moving towards the next major CTK version, so 🤷

miscco:
@ChuckHastings can we rerun CI on this one? I do not see any reason this would affect any python tests

seunghwak:
> @ChuckHastings can we rerun CI on this one? I do not see any reason this would affect any python tests

I think there is a CI issue affecting every PR in cugraph. I will rerun CI once other PRs start to pass CI again.

ChuckHastings (approved):

PR #4964, opened 2025-03-10
Fix datasets API (no longer experimental)

bdice:
We have some failing notebooks in the Docker image because `cugraph.experimental.datasets` doesn't exist. This appears to be promoted to `cugraph.datasets`, so I updated code referencing the old location.

This corrects failures observed in https://github.com/rapidsai/docker/pull/742.


acostadon (approved):
Looks good

PR #4966, opened 2025-03-11
Optimize check for vertex existence

### LINKED ISSUE #4956: [QST]: Why are we experiencing high execution times with the BFS algorithm?
### What is your question?

We're testing the BFS algorithm where the start is fixed in a large directed network and observing strange results with computation time on a single GPU. First, the BFS computation has significantly higher execution times compared to the betweenness centrality algorithm for the same network and starting node. Second, we don't see any significant change in time by lowering the depth limit for the BFS. Do you have any insights into what might be causing this issue? or are they usually expected? 

`cugraph version 24.10
`

```
import os
import random
from time import perf_counter

import cugraph
import cudf

from cugraph.generators import rmat
from cugraph.structure import NumberMap

def generate_data(scale, edgefactor=16, mg=False):
    _gdf = rmat(
        scale,
        (2**scale) * edgefactor,
        0.57,
        0.19,
        0.19,
        42,
        clip_and_flip=False,
        scramble_vertex_ids=True,
        create_using=None,  # return edgelist instead of Graph instance
        mg=False,  # determines whether generated data will be used on one or multiple GPUs
    )

    clean_coo = NumberMap.renumber(_gdf, src_col_names="src", dst_col_names="dst")[0]
    clean_coo.rename(
        columns={"renumbered_src": "src", "renumbered_dst": "dst"}, inplace=True
    )
    return clean_coo

# generate a random graph
gdf = generate_data(scale=23, edgefactor=16, mg=False)

# build a cugraph graph
t1 = perf_counter()
g = cugraph.Graph(directed=True)
g.from_cudf_edgelist(
    gdf,
    source="src",
    destination="dst",
    edge_attr=None,
    renumber=False,
)
t2 = perf_counter() - t1
print(f"Time to create graph: {t2} sec")

random_node = random.choice(g.nodes())
print(f"Random node: {random_node}")

# run bfs 
for depth_limit in range(1, 7):
    t1 = perf_counter()
    bfs_df = cugraph.bfs(g, start=random_node, depth_limit=depth_limit)
    t2 = perf_counter() - t1
    bfs_df = bfs_df[bfs_df["distance"] <= depth_limit]   
    print(f"BFS depth_limit: {depth_limit} # nodes reachable {bfs_df.shape[0]} time: {t2} sec")

t1 = perf_counter()
bc_df = cugraph.betweenness_centrality(g, k=[random_node])
t2 = perf_counter() - t1
print(f"Time to run BC: {t2} sec")

```

```
BFS depth_limit: 1 # nodes reachable 7 time: 28.294853757000965 sec
BFS depth_limit: 2 # nodes reachable 5607 time: 28.294678542999463 sec
BFS depth_limit: 3 # nodes reachable 1719914 time: 28.321973072997935 sec
BFS depth_limit: 4 # nodes reachable 3803979 time: 28.333574186999613 sec
BFS depth_limit: 5 # nodes reachable 3844515 time: 28.343743367997376 sec

Time to run BC: 0.5936050090022036 sec

```

### Code of Conduct

- [x] I have searched the [open issues](https://github.com/rapidsai/cugraph/issues) and have found no duplicates for this question

jnke2016:
This PR leverages the CAPI to optimize the check for vertex existence which can lead to +100x speedup compared to the cudf based version as it can be noticed from the performance figure below

Screenshot 2025-03-17 at 9 47 37 AM

closes #4956

### DIFF
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -117,6 +117,22 @@ void transform_increment_ints(raft::device_span<value_t> values,
 
 /**
  * @ingroup utility_wrappers_cpp
+ * @brief    Update a value in a device span to 0 if it matches the target_value or 1
+ *
+ * @tparam      value_t      type of the value to operate on. Must be either int32_t or int64_t.
+ *
+ * @param[out]  values       device span to update
+ * @param[in]   target_value        value to be querried
+ * @param[in]   stream_view  stream view
+ *
+ */
+template <typename value_t>
+void transform_binary(raft::device_span<value_t> values,
+                      value_t target_value,
+                      raft::handle_t const& handle);
+
+/**
+ * @ingroup utility_wrappers_cpp
  * @brief    Fill a buffer with a sequence of values
  *
  * Fills the buffer with the sequence:
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -118,18 +118,20 @@ void transform_increment_ints(raft::device_span<value_t> values,
 
 /**
  * @ingroup utility_wrappers_cpp
- * @brief    Update a value in a device span to 0 if it matches the target_value or 1
+ * @brief    Update the values of device span to 0 if it matches the compare value or 1
  *
  * @tparam      value_t      type of the value to operate on. Must be either int32_t or int64_t.
  *
- * @param[out]  values       device span to update
- * @param[in]   target_value        value to be querried
+ * @param[out]  values       device span with the values to compare
+ * @param[out]  result       device span with the result of the comparison
+ * @param[in]   compare      value to be querriedm in the values array
  * @param[in]   stream_view  stream view
  *
  */
 template <typename value_t>
-void transform_binary(raft::device_span<value_t> values,
-                      value_t target_value,
+void transform_not_equal(raft::device_span<value_t> values,
+                      raft::device_span<bool> result,
+                      value_t compare,
                       raft::handle_t const& handle);
 
 /**
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -269,5 +269,21 @@ size_t count_values(raft::handle_t const& handle,
                     raft::device_span<data_t const> span,
                     data_t value);
 
+/**
+ * @ingroup utility_wrappers_cpp
+ * @brief Verify if a value appears in a span
+ *
+ * @tparam data_t type of data in span
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param span The span of data to compare
+ * @param value The value to count
+ * @return true if found, false if not found
+ */
+template <typename data_t>
+size_t has_vertex(raft::handle_t const& handle,
+                  raft::device_span<data_t const> span,
+                  data_t value);
+
 }  // namespace detail
 }  // namespace cugraph
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -269,21 +269,5 @@ size_t count_values(raft::handle_t const& handle,
                     raft::device_span<data_t const> span,
                     data_t value);
 
-/**
- * @ingroup utility_wrappers_cpp
- * @brief Verify if a value appears in a span
- *
- * @tparam data_t type of data in span
- * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
- * handles to various CUDA libraries) to run graph algorithms.
- * @param span The span of data to compare
- * @param value The value to count
- * @return true if found, false if not found
- */
-template <typename data_t>
-size_t has_vertex(raft::handle_t const& handle,
-                  raft::device_span<data_t const> span,
-                  data_t value);
-
 }  // namespace detail
 }  // namespace cugraph
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -130,9 +130,9 @@ void transform_increment_ints(raft::device_span<value_t> values,
  */
 template <typename value_t>
 void transform_not_equal(raft::device_span<value_t> values,
-                      raft::device_span<bool> result,
-                      value_t compare,
-                      raft::handle_t const& handle);
+                         raft::device_span<bool> result,
+                         value_t compare,
+                         raft::handle_t const& handle);
 
 /**
  * @ingroup utility_wrappers_cpp
--- a/cpp/include/cugraph/detail/utility_wrappers.hpp
+++ b/cpp/include/cugraph/detail/utility_wrappers.hpp
@@ -132,7 +132,7 @@ template <typename value_t>
 void transform_not_equal(raft::device_span<value_t> values,
                          raft::device_span<bool> result,
                          value_t compare,
-                         raft::handle_t const& handle);
+                         rmm::cuda_stream_view const& stream_view);
 
 /**
  * @ingroup utility_wrappers_cpp
--- a/cpp/include/cugraph_c/graph_functions.h
+++ b/cpp/include/cugraph_c/graph_functions.h
@@ -110,7 +110,7 @@ cugraph_error_code_t cugraph_two_hop_neighbors(
  * @param [in]  vertex         Vertex to be queried
  * @param [in]  do_expensive_check
  *                             A flag to run expensive checks for input arguments (if set to true)
- * @param [out] result         Opaque pointer to resulting queried vertex
+ * @param [out] result         Opaque pointer to resulting queried vertices
  * @param [out] error          Pointer to an error object storing details of any error.  Will
  *                             be populated if error code is not CUGRAPH_SUCCESS
  * @return error code
@@ -119,7 +119,7 @@ cugraph_error_code_t cugraph_has_vertex(const cugraph_resource_handle_t* handle,
                                         cugraph_graph_t* graph,
                                         cugraph_type_erased_device_array_view_t* vertices,
                                         bool_t do_expensive_check,
-                                        bool_t* result,
+                                        cugraph_type_erased_device_array_t** result,
                                         cugraph_error_t** error);
 
 /**
--- a/cpp/include/cugraph_c/graph_functions.h
+++ b/cpp/include/cugraph_c/graph_functions.h
@@ -107,7 +107,7 @@ cugraph_error_code_t cugraph_two_hop_neighbors(
  *
  * @param [in]  handle         Handle for accessing resources
  * @param [in]  graph          Pointer to graph
- * @param [in]  vertex         Vertex to be queried
+ * @param [in]  vertices       Vertices to be queried
  * @param [in]  do_expensive_check
  *                             A flag to run expensive checks for input arguments (if set to true)
  * @param [out] result         Opaque pointer to resulting queried vertices
--- a/cpp/include/cugraph_c/graph_functions.h
+++ b/cpp/include/cugraph_c/graph_functions.h
@@ -102,7 +102,6 @@ cugraph_error_code_t cugraph_two_hop_neighbors(
   cugraph_vertex_pairs_t** result,
   cugraph_error_t** error);
 
-
 /**
  * @brief      Verify if a vertex exists in the graph
  *
@@ -116,13 +115,12 @@ cugraph_error_code_t cugraph_two_hop_neighbors(
  *                             be populated if error code is not CUGRAPH_SUCCESS
  * @return error code
  */
-cugraph_error_code_t cugraph_has_vertex(
-  const cugraph_resource_handle_t* handle,
-  cugraph_graph_t* graph,
-  const int vertex,
-  bool_t do_expensive_check,
-  bool_t* result,
-  cugraph_error_t** error);
+cugraph_error_code_t cugraph_has_vertex(const cugraph_resource_handle_t* handle,
+                                        cugraph_graph_t* graph,
+                                        const int vertex,
+                                        bool_t do_expensive_check,
+                                        bool_t* result,
+                                        cugraph_error_t** error);
 
 /**
  * @brief       Opaque induced subgraph type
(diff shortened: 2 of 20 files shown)


rlratzel:
@jnke2016 does this close #4956 by itself or are there other changes to the Python layer needed?

cpp/src/c_api/graph_functions.cpp
```diff
@@ -132,6 +132,7 @@ struct two_hop_neighbors_functor : public cugraph::c_api::abstract_functor {
             bool multi_gpu>
   void operator()()
   {
+    printf("\nin two_hop_neighbors \n");
```
seunghwak, line 135: Please delete debug print statements.

seunghwak (commented):
Is this PR ready for review?

I see just debug print statements and commented out code. If the PR is not ready for review, you may better create a draft PR (this also saves CI resources).

cpp/src/c_api/graph_functions.cpp
```diff
@@ -261,6 +261,74 @@ struct count_multi_edges_functor : public cugraph::c_api::abstract_functor {
...
+      }
+
+      cugraph::renumber_ext_vertices<vertex_t, multi_gpu>(
+        handle_,
+        vertex_array.data(),
+        vertex_array.size(),
+        number_map->data(),
+        graph_view.local_vertex_partition_range_first(),
+        graph_view.local_vertex_partition_range_last(),
+        do_expensive_check_);
```
seunghwak, line 316: I assume the same vertex_t value is passed to every GPU in multi-GPU, right? Or are you assuming that each GPU can call this function with different `vertex_` values?

seunghwak: If the former is the case, you don't need to really shuffle vertices, just need to call `compute_gpu_id_from_ext_vertex_t` to find the owning GPU. The owning GPU performs the check and broadcast the check result.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -263,6 +263,34 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+                             edgelist_dsts.size(),
+                             handle_.get_stream());
+
+        cugraph::detail::sort_ints(handle_.get_stream(),
+                                   raft::device_span<vertex_t>{vertices.data(), vertices.size()});
+
+        unique_vertices_size = cugraph::detail::unique_ints(
+          handle_.get_stream(), raft::device_span<vertex_t>{vertices.data(), vertices.size()});
+
+        vertices.resize(unique_vertices_size, handle_.get_stream());
```
seunghwak, line 291: To reduce memory footprint, you can sort & unique sources & destinations separately and merge instead of sort & unique sources, then copy all the destinations, and sort & unique.

You may further cut the memory footprint by first performing hash based group by (https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/include/cugraph/utilities/shuffle_comm.cuh#L761 with mem_frugal_threshold set), but this might be little bit of an overkill and you may defer this till this actually becomes a bottlenck (especially considering that renumber = false is here mainly for historical/debugging reasons).

If you just want to put working simple implementation here (as this won't be a case actual users will heavily use), just copy all sources & destinations to a single array and sort & unique.

seunghwak: And don't forget to call shrink_to_fit() to actually release memory after resize.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -292,11 +320,25 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+          handle_.get_stream(),
+          raft::device_span<vertex_t>{vertices.data(), vertices.size()},
+          raft::device_span<vertex_t>{number_map->data(), number_map->size()});
+
+        if (!is_consecutive) {
+          mark_error(
+            CUGRAPH_INVALID_INPUT,
+            "Vertex list must be numbered consecutively from 0 when 'renumber' is 'false'");
+          return;
+        }
```
seunghwak, line 341: Do you need to keep both `vertices` and `number_map`?

If you just need to keep `number_map`, you can just check `vertices` is an array of consecutive integers starting from 0 and `move` `vertices` to `number_map`.

cpp/include/cugraph_c/graph_functions.h
```diff
@@ -102,6 +102,26 @@ cugraph_error_code_t cugraph_two_hop_neighbors(
...
+ * @param [in]  do_expensive_check
+ *                             A flag to run expensive checks for input arguments (if set to true)
+ * @param [out] result         Opaque pointer to resulting queried vertex
+ * @param [out] error          Pointer to an error object storing details of any error.  Will
+ *                             be populated if error code is not CUGRAPH_SUCCESS
+ * @return error code
+ */
+cugraph_error_code_t cugraph_has_vertex(const cugraph_resource_handle_t* handle,
+                                        cugraph_graph_t* graph,
+                                        const int vertex,
```
ChuckHastings, line 120: Can't use an int here.  A vertex can be int32_t or int64_t.

For SSSP, we pass in a `size_t` and cast it to either int32_t or int64_t inside the C API implementation.

cpp/src/c_api/graph_functions.cpp
```diff
@@ -261,6 +261,74 @@ struct count_multi_edges_functor : public cugraph::c_api::abstract_functor {
   }
 };
 
+struct has_vertex_functor : public cugraph::c_api::abstract_functor {
+  raft::handle_t const& handle_{};
+  cugraph::c_api::cugraph_graph_t* graph_{nullptr};
+  const int vertex_{};
```
ChuckHastings, line 267: Can't be an int.

rlratzel (changes requested):
Looks good, thanks!

I just have a few minor requests.

cpp/src/c_api/graph_sg.cpp
jnke2016: > To reduce memory footprint, you can sort & unique sources & destinations separately and

Right. I now sort & unique sources & destinations separately. I also call `shrink_to_fit`.

cpp/src/c_api/graph_sg.cpp
jnke2016: > If you just need to keep number_map

Yes I only need to keep `number_map`.

> you can just check vertices is an array of consecutive integers starting from 0 and move vertices to number_map

That's essentially what the check `cugraph::detail::is_equal` is doing.

cpp/src/c_api/graph_sg.cpp
jnke2016: > You may further cut the memory footprint by first performing hash based group by ...

Right, we can add this optimization if this becomes a bottleneck which might not be for now. And as you mention, the case this code is targeting is for historical/debugging reasons

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -116,6 +116,22 @@ void transform_increment_ints(raft::device_span<value_t> values,
...
+ *
+ * @param[out]  values       device span to update
+ * @param[in]   target_value        value to be querried
+ * @param[in]   stream_view  stream view
+ *
+ */
+template <typename value_t>
+void transform_binary(raft::device_span<value_t> values,
+                      value_t target_value,
+                      raft::handle_t const& handle);
```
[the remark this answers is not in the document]
seunghwak: I asked this to chatgpt and got this.

```
A good name for this function should align with STL naming conventions, which are often verb-based and descriptive of their operation. Some well-known STL functions that perform element-wise transformations or checks include std::transform, std::count_if, and std::any_of.

Since your function checks whether each element differs from a given value and returns a boolean result for each, a name that follows STL style could be:

Suggested names:
mismatch_mask – Inspired by std::mismatch, emphasizing that it creates a mask of mismatches.
not_equal_mask – Mirrors std::not_equal_to, clearly indicating the operation.
transform_not_equal – Follows the std::transform pattern, emphasizing element-wise transformation.
compare_not_equal – Explicit about the comparison operation.
Recommended choice:
not_equal_mask
This name aligns well with STL conventions, is concise, and clearly communicates that the function generates a mask indicating inequality.
```

cpp/src/c_api/graph_sg.cpp
```diff
@@ -263,6 +263,60 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+
+        cugraph::detail::sort_ints(handle_.get_stream(),
+                                   raft::device_span<vertex_t>{vertices.data(), vertices.size()});
+
+        size_t unique_vertices_size = cugraph::detail::unique_ints(
+          handle_.get_stream(), raft::device_span<vertex_t>{vertices.data(), vertices.size()});
+
+        vertices.resize(unique_vertices_size, handle_.get_stream());
+        vertices.shrink_to_fit(handle_.get_stream());
+      }
```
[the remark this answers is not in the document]
jnke2016: > You can move this code inside the else path below.

I don't think I can unless I am misunderstanding your suggestion. Inside the else path you mentioned, `edgelist_srcs` and `edgelist_dsts` are out of context because of `move` in `create_graph_from_edgelist`.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -263,6 +263,60 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+          handle_.get_stream(),
+          raft::device_span<vertex_t>{tmp_edgelist_dsts.data(), tmp_edgelist_dsts.size()});
+
+        tmp_edgelist_dsts.resize(unique_edgelist_dsts_size, handle_.get_stream());
+
+        tmp_edgelist_dsts.shrink_to_fit(handle_.get_stream());
+
+        vertices.resize(unique_edgelist_srcs_size + unique_edgelist_dsts_size,
+                        handle_.get_stream());
+        vertices.shrink_to_fit(handle_.get_stream());
```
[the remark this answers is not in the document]
jnke2016: Right this is addressed

cpp/src/c_api/graph_functions.cpp
```diff
@@ -261,6 +261,72 @@ struct count_multi_edges_functor : public cugraph::c_api::abstract_functor {
...
+        vertices.size(),
+        number_map->data(),
+        graph_view.local_vertex_partition_range_first(),
+        graph_view.local_vertex_partition_range_last(),
+        do_expensive_check_);
+
+      cugraph::detail::transform_binary(
+        raft::device_span<vertex_t>{vertices.data(), vertices.size()},
+        cugraph::invalid_vertex_id<vertex_t>::value,
+        handle_.get_stream());
```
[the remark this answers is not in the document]
jnke2016: I created a `rmm::device_uvector<bool>` to store the result of the comparison.

cpp/include/cugraph/detail/utility_wrappers.hpp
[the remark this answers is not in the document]
jnke2016: Interesting. I might go for `chatgpt` next time I am trying to find a name for my `thrust-like` function. I chose `transform_not_equal`

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -116,6 +116,24 @@ void transform_increment_ints(raft::device_span<value_t> values,
...
+ * @param[out]  result       device span with the result of the comparison
+ * @param[in]   compare      value to be querriedm in the values array
+ * @param[in]   stream_view  stream view
+ *
+ */
+template <typename value_t>
+void transform_not_equal(raft::device_span<value_t> values,
+                         raft::device_span<bool> result,
+                         value_t compare,
+                         raft::handle_t const& handle);
```
seunghwak, line 135: Our convention is to pass handle as the first input argument (and stream as the last).

cpp/src/c_api/graph_sg.cpp
seunghwak: Yeah... what I meant is that you don't need to allocate memory for `vertices` twice nor call `sequence_fill`.

`thrust::is_equal(vertices.begin(), vertices.end(), thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first());` would be sufficient.

cpp/src/c_api/graph_sg.cpp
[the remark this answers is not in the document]
seunghwak: Oh, I see, you're right.

cpp/src/c_api/graph_sg.cpp
seunghwak: Disregard this comment. This code won't be called in normal use cases, so let's just waste little memory/computing here rather than creating another thrust wrapper function.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -292,11 +294,34 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
       if (renumber_) {
         *number_map = std::move(new_number_map.value());
       } else {
+
```
seunghwak, line 297: This empty line is unnecessary.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -292,11 +294,34 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
+            raft::device_span<vertex_t>{number_map->data(), number_map->size()});
+
+          if (!is_consecutive) {
+            mark_error(
+              CUGRAPH_INVALID_INPUT,
+              "Vertex list must be numbered consecutively from 0 when 'renumber' is 'false'");
+            return;
+          }
+        
+
```
seunghwak, line 323: These empty lines are unnecessary.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -263,6 +263,8 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
       std::optional<rmm::device_uvector<edge_time_t>> dummy_start_times{std::nullopt};
       std::optional<rmm::device_uvector<edge_time_t>> dummy_end_times{std::nullopt};
 
+        
+
```
seunghwak, line 267: No need for 3 empty lines. 1 is sufficient.

seunghwak (approved):
LGTM (except for unnecessary empty lines).

ChuckHastings (approved):

rlratzel (commented):
Recent changes look good but I have a question.

python/pylibcugraph/pylibcugraph/has_vertex.pyx
```diff
@@ -64,29 +66,42 @@ def has_vertex(ResourceHandle resource_handle,
...
 
+    cdef cugraph_type_erased_device_array_view_t* \
+        result_view_ptr = \
+            cugraph_type_erased_device_array_view(
+                result_ptr)
+
+    cupy_has_vertex = copy_to_cupy_array(c_resource_handle_ptr, result_view_ptr)
 
-    return True if result else False
+    return cupy_has_vertex
```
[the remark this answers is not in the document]
rlratzel: Talked to @jnke2016 offline: reason for returning the array is to facilitate passing/querying multiple vertices and getting individual bools back for each.  This also helps facilitate MG use cases.  The benchmarks were not impacted significantly with this change.
@jnke2016 is going to update the docstring and example now that the return value is different.

rlratzel (approved):

PR #4969, opened 2025-03-12
Refactor Force Atlas 2

### LINKED ISSUE #4881: Expose Force Atlas 2 through PLC
In support of https://github.com/rapidsai/nx-cugraph/issues/63

Expose the Force Atlas 2 functionality through the C API and PLC.

### LINKED ISSUE #4968: [FEA]: Expose Force Atlas 2 to the PLC API

New Feature

Medium

Exposing Force Atlas 2 to the PLC API enables

- An integration in the nx-cugraph API
- The python cuGraph legacy implementation to leverage the PLC graph
- Smooth integration at the PLC and python layer once a new implementation leveraging our primitives is available

Expose Force Atlas 2 to the PLC

_No response_

### Additional context

_No response_

### Code of Conduct

jnke2016:
This PR exposes FA2 to the PLC API

closes #4881


cpp/include/cugraph_c/layout_algorithms.h
```diff
@@ -0,0 +1,101 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/**
+ * @brief   Force Atlas 2
+ *
+ * NOTE: This currently wraps the legacy force atlas 2 implementation and is only
```
ChuckHastings, line 36: This is inconsistent with the C API general approach.  Specifically, it exposes assumptions about the internal implementation through the API.  We've generally tried to avoid this.  If we were to ever create a MG version or to refactor FA to use graph primitives these assumptions would no longer be true.

I'll highlight some ideas on changes to make things more consistent.

cpp/include/cugraph_c/layout_algorithms.h
```diff
@@ -0,0 +1,101 @@
...
+ * available in Single GPU implementation.
+ *
+ * @param [in]   handle          Handle for accessing resources
+ * @param [in]   graph           Pointer to graph.  NOTE: Graph might be modified if the storage
+ *                               needs to be transposed
+ * @param [out]  pos             Position of the vertices
```
ChuckHastings, line 42: Note that in the C++ code, pos is a 2-d array and assumes that the output is ordered by internal vertex ids.

Generally, the C API hides the renumbering of vertices from the user (we automatically renumber/unrenumber in the C API before calling the C++ API.  As such, this really needs include the vertex ids.

I'd suggest creating a `cugraph_layout_result_t` type and including vertex id, x position, y position so that you can easily create a data frame for the resulting python.

cpp/include/cugraph_c/layout_algorithms.h
```diff
@@ -0,0 +1,101 @@
...
+ * @param [in]   max_iter        Maximum number of iterations. Initial vertex positioning
+ * @param [in]   x_start         Initial vertex positioning (x-axis)
```
ChuckHastings, line 44: Same issue for `x_start` and `y_start`.  The values in the legacy C++ implementation are assumed to be in the order of the internal vertex ids... but the python code doesn't know this order.

I'd suggest using the `cugraph_layout_result_t` object as input here.  You can create one, passing in the vertex ids, x_start and y_start values and use it as input.

cpp/include/cugraph_c/layout_algorithms.h
```diff
@@ -0,0 +1,101 @@
...
+ *                               How much repulsion you want. More makes a more sparse graph.
+ *                               Switching from regular mode to LinLog mode needs a readjustment of the scaling parameter.
+ * @param [in]   strong_gravity_mode
+ *                               Sets a force that attracts the nodes that are distant from the
+ *                               center more. It is so strong that it can sometimes dominate other forces.
+ * @param [in]   gravity         Attracts nodes to the center. Prevents islands from drifting away.
+ * @param [in]   verbose         Output convergence info at each interation.
+ * @param [in]   do_expensive_check
+ *                               A flag to run expensive checks for input arguments (if set to true)
+ * @param [out]  result          Opaque object containing the clustering result
```
ChuckHastings, line 73: This result would be what `pos` above is.  I'd delete the `pos` parameter above and move it to this position.  By convention, the C API functions pass all input parameters before the output parameters.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,248 @@
...
+  void operator()()
+  {
+    if constexpr (!cugraph::is_candidate<vertex_t, edge_t, weight_t>::value) {
+      unsupported();
+    } else if constexpr (multi_gpu) {
+      unsupported();
+    } else if constexpr (!std::is_same_v<edge_t, int32_t>) {
+      unsupported();
+    } else {
+      // FIXME: Does FA2 expects store_transposed == false
```
ChuckHastings, line 119: FA2 is expecting a `legacy::GraphCOOView`, which you're creating below.  there's no reason to transpose the graph, you should be able to use it exactly as is.

I'd take out this logic.  You can then remove the annotation in the C API documentation that the graph object might be modified... because in the case of this algorithm it would not be.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,248 @@
...
+        const_cast<vertex_t*>(dsts.data()),
+        (edge_weights == nullptr)
+          ? tmp_weights.data()
+          : const_cast<weight_t*>(wgts->data()),
+        graph->number_of_vertices(),
+        edge_partition_view.number_of_edges());   
+
+      cugraph::internals::GraphBasedDimRedCallback* callback = nullptr;
+      
+      //cugraph::legacy::GraphBasedDimRedCallback() // FIXME Add support for callback
```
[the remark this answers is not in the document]
rlratzel: This is exposed in the current Python API too, meaning users could theoretically be using it now. However, I really doubt anyone is, and I'm in favor of making this a breaking change as a tradeoff for not having to support, test, and document it in the new PLC-based API.

I like the code and it could be useful for something, but it may mean that we can't release the GIL (which we've been wanting to do in more of our functions), and I suspect we could get most of the intended functionality with an option that enables debug prints if we really wanted to.

Maybe @hlinsen has more info about the intended use case though?

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,248 @@
...
+#include <cugraph/algorithms.hpp>
+#include <cugraph/detail/utility_wrappers.hpp>
+#include <cugraph/graph_functions.hpp>
+
+#include <optional>
+
+namespace cugraph {
+namespace c_api {
+
+struct cugraph_clustering_result_t {
```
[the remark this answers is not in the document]
jnke2016: I defined `cugraph_layout_result_t` instead

cpp/include/cugraph_c/layout_algorithms.h
jnke2016: Got it. I have seen this line`NOTE: This currently wraps...implementation and is only
 * available in Single GPU implementation` in several legacy C code (`community_algorithms.h`) and I was under the assumption that we should add this note

cpp/src/c_api/legacy_fa2.cpp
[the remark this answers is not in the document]
hlinsen: This was an asked by @trxcllnt and @exactlyallan for dynamic viz. There is maybe some cyber use cases or demo that could need it. I think it is also exposed in umap for cuml.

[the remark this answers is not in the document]
jnke2016: Thanks for the feedback @hlinsen . Do you have suggestions about ways to test FA2 apart from the smoke tests cuGraph and networkX have?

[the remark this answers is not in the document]
hlinsen: Some ideas:
- Checking that we converged by checking some of the values like speed, swinging values [here](https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/layout/legacy/barnes_hut.cuh#L332-L334)
- Trying to make barnes hut deterministic and compare positions
- Check consistency in nearest neighbor nodes, look at a top k. Some times the layout is the same but shifted and I think nearest nodes should still be the same ones.

cpp/include/cugraph_c/layout_algorithms.h
ChuckHastings: With your latest refactoring you did not address this concern.  The legacy C++ implementation assumes that `x_start` and `y_start` are sorted such that element 0 refers to the starting position of vertex 0 using the internal vertex id.  If the graph was created with renumbering, there is no way for the caller to know what order to put `x_start` and `y_start`.

Simplest solution would be to add a vertices parameter (optional like x_start and y_start) that identified the vertex id for each position.

Then in the implementation you can make a copy of this data into rmm device vectors, renumber the vertices and then sort the vertices/x_start/y_start by vertex id.  I think page rank can be used as a model for handling the intial guess.


Also, documentation for x_start and y_start (and the new vertices parameter) should indicate that they are optional.  And there should be a `CAPI_EXPECTS` call that verifies that if you specify x_start and y_start you must also specify vertices.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,291 @@
...
+      rmm::device_uvector<float> pos(2 * (edge_partition_view.offsets().size() - 1),
+                                     handle_.get_stream());
+
+      cugraph::force_atlas2<vertex_t, edge_t, weight_t>(
+        handle_,
+        // rng_state_->rng_state_, # FIXME: Add support
+        legacy_coo_graph_view,
+        pos.data(),
+        max_iter_,
+        x_start_ != nullptr ? x_start_->as_type<float>() : nullptr,
```
ChuckHastings, line 176: As mentioned above, this need to be sorted properly, so you'll need to make a copy and sort by vertex id after renumbering.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,291 @@
...
+      cugraph::legacy::GraphCOOView<vertex_t, edge_t, weight_t> legacy_coo_graph_view(
+        const_cast<vertex_t*>(srcs.data()),
+        const_cast<vertex_t*>(dsts.data()),
+        (edge_weights == nullptr) ? tmp_weights.data() : const_cast<weight_t*>(wgts->data()),
+        graph->number_of_vertices(),
+        edge_partition_view.number_of_edges());
+
+      cugraph::internals::GraphBasedDimRedCallback* callback = nullptr;
+
+      rmm::device_uvector<vertex_t> vertices(graph_view.local_vertex_partition_range_size(),
```
ChuckHastings, line 163: I would move this down to where you do x_axis and y_axis... I don't think you need this until constructing the result.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,291 @@
...
+      // result_ = new cugraph::c_api::cugraph_type_erased_device_array_t(pos,
+      // graph_->weight_type_);
+
+      rmm::device_uvector<float> x_axis(graph_view.local_vertex_partition_range_size(),
+                                        handle_.get_stream());
+
+      rmm::device_uvector<float> y_axis(graph_view.local_vertex_partition_range_size(),
+                                        handle_.get_stream());
+
+      raft::copy(x_axis.data(), pos.data(), x_axis.size(), handle_.get_stream());
```
ChuckHastings, line 200: Is this correct?  I thought the x and y values are interleaved.  You would need a strided copy (elements 0, 2, 4, ... go in x_axis 0,1,2... and elements 1, 3, 5, ... go in y_axis 0, 1, 2...).  You might double check that.

cpp/src/c_api/legacy_fa2.cpp
jnke2016: After our discussion, I sort both the `x_start` and `y_start` by key where the key refers to the position of the `number_map`

cpp/src/c_api/legacy_fa2.cpp
jnke2016: I double checked and I don't think they are interleaved. The array pos holds both the x and y coordinate and when called for instance in [apply_forces](https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/layout/legacy/exact_fa2.cuh#L173:#L174) (pos, pos+n) where n is the number of vertex, this translates to the beginning of the [x and y position](https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/layout/legacy/fa2_kernels.cuh#L162:L163)

I also made sure the result matches the non PLC version which was the case

cpp/src/c_api/legacy_fa2.cpp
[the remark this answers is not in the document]
jnke2016: > Is legacy referring to the C++ implementation not using primitives

I believe that is correct @ChuckHastings 

> If so, once it uses primitives, would this file need to be replaced? I figured this would just be called fa2.cpp and only the C++ impl would change once primitives are used.
 
Right. No changes to the C (except the file name), and PLC API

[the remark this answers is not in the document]
ChuckHastings: Yes, we targeted getting the API definition right for both legacy implementation and some future primitive based implementation.  The implementation is currently solely a legacy implementation.  The implementation of the C API will look totally different on a primitive based C++ implementation.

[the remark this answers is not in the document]
jnke2016: > The implementation of the C API will look totally different

However the function definition in the C API will most likely be unchanged right? and regarding C the implementation, we will remove the legacy COO logic and call the primitive based C++ implementation right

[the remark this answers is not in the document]
ChuckHastings: Correct.  Function definition would be the same.  The functor definition might be the same... the implementation inside the `operator()` method would likely be totally different (simpler) as we wouldn't have to worry about legacy graph objects.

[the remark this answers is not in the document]
rlratzel: > Function definition would be the same.

ok, that's what I figured too.  In that case, my preference would be to just call the file `fa2.cpp` since the `_legacy` prefix isn't needed and will be misleading later, but I don't feel strongly about it, and it can always be renamed later if needed.

ChuckHastings (commented):
A few documentation cleanup comments.

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -0,0 +1,316 @@
...
+  double barnes_hut_theta,
+  double scaling_ratio,
+  bool_t strong_gravity_mode,
+  double gravity,
+  bool_t verbose,
+  bool_t do_expensive_check,
+  cugraph::c_api::cugraph_layout_result_t** result,
+  cugraph_error_t** error)
+{
+  force_atlas2_functor functor(handle,
```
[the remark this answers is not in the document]
jnke2016: I updated the docs and in case x_start and y_start are not provided, it is now mentioned in the documentation that those will be selected randomly internally.

rlratzel (approved):

PR #4971, opened 2025-03-13
[CTK 12.5]: Avoid another compiler issue with host device detection

miscco:
NVCC seems to also incorrectly deduce this in 12.5

### DIFF
--- a/cpp/include/cugraph/edge_property.hpp
+++ b/cpp/include/cugraph/edge_property.hpp
@@ -54,6 +54,10 @@ class edge_property_view_t {
 
   // nvcc 12.5 sometimes deduces this as a host device function, just defining it fixes that
   ~edge_property_view_t() {}
+  edge_property_view_t(const edge_property_view_t&)            = default;
+  edge_property_view_t(edge_property_view_t&&)                 = default;
+  edge_property_view_t& operator=(const edge_property_view_t&) = default;
+  edge_property_view_t& operator=(edge_property_view_t&&)      = default;
 
   std::vector<ValueIterator> const& value_firsts() const { return edge_partition_value_firsts_; }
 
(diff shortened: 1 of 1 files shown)


miscco:
This is the last issue in getting rapids to build with CCCL main see https://github.com/NVIDIA/cccl/actions/runs/13822026368/job/38669398923?pr=4123#step:6:12069

bdice (approved):

cpp/include/cugraph/edge_property.hpp
```diff
@@ -54,6 +54,10 @@ class edge_property_view_t {
 
   // nvcc 12.5 sometimes deduces this as a host device function, just defining it fixes that
   ~edge_property_view_t() {}
+  edge_property_view_t(const edge_property_view_t&)            = default;
```
miscco, line 57: I forgot that we need to tell the compiler that we still want the other SMF.

cpp/include/cugraph/edge_property.hpp
```diff
@@ -113,6 +117,13 @@ class edge_property_t {
   {
   }
 
+  // nvcc 12.5 sometimes deduces this as a host device function, just defining it fixes that
+  ~edge_property_t() {}
+  edge_property_t(const edge_property_t&)            = default;
```
seunghwak, line 122: Copy constructor & operator should be deleted. This class is a wrapper (+ alpha) for rmm::device_uvector.

ChuckHastings: @miscco - we should correct this before merging this change.

miscco: @ChuckHastings @seunghwak Changed to delete

ChuckHastings (approved):

PR #4972, opened 2025-03-13
Fix update-version script

raydouglass:
Add two new dependencies to `update-version.sh`

Skipping CI because this script is not tested in CI

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -59,6 +59,8 @@ DEPENDENCIES=(
   dask-cudf
   libcudf
   libcugraph
+  libcugraph_etl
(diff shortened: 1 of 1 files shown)


ci/release/update-version.sh
```diff
@@ -59,6 +59,8 @@ DEPENDENCIES=(
   dask-cudf
   libcudf
   libcugraph
+  libcugraph_etl
+  libcugraph-tests
```
jameslamb, line 63: `snake_case` for one and `kebob-case` for the other makes me sad 😭

jameslamb (approved):

PR #4985, opened 2025-03-19
Revert hacks around compile issue with host device deduction

miscco:
We found the issue to be in some unclear compile issues within libcu++ `not_fn`, so we can revert all those hacks

Revert "Silence compiler warnings about host device destructor (#4960)"
Revert "[CTK 12.5]: Avoid another compiler issue with host device detection (#4971)"


seunghwak (approved):
LGTM

PR #4988, opened 2025-03-19
Refactor Python Neighborhood Sample

### LINKED ISSUE #3855: Remove Old Sampling C API and PLC Functions
These were deprecated in 23.10 and should be removed in 23.12.

### LINKED ISSUE #3856: Remove "With Edge Properties" Sampling Option from Python API
This option was deprecated in 23.10.  It should be removed in 23.12.

### LINKED ISSUE #4358: Refactor the Python Sampling Algorithms
The Python sampling algorithms (particularly `uniform_neighbor_sample`) are overly complicated and need to be redesigned to better suit user needs, and properly support the full suite of features available at the C++/C API/pylibcugraph level.  For instance, right now, they take a list of batch ids rather than a batch size parameter, which makes the code significantly more complicated and slower.

Ultimately, the primary means of interaction with the sampling algorithms going forward is probably going to be `pylibcugraph` and the C++ API, so those should drive the new Python API.  Some parameters, like `label_to_output_comm_rank` should probably be thrown out in favor of preserving the original seed location.

For the future in-memory sampling, we may implement something like `label_to_output_comm_rank` that is more clever, and evenly divides the samples from _N_ samplers to be distributed across _M_ trainers.

jnke2016:
A new sampling API supporting both homogeneous and heterogeneous fanout type was exposed to the PLC API along with a biased version of neighborhood sampling.

This PR exposes the new sampling API to the python layer and removes deprecated API/calls/arguments.

closes #3855 
closes #4358 
closes #3856

### DIFF
--- a/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
+++ b/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2024, NVIDIA CORPORATION.
+# Copyright (c) 2022-2025, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -97,9 +97,9 @@ def call_algo(sg_algo_func, G, **kwargs):
             kwargs_to_pass = {a: kwargs[a] for a in possible_args if a in kwargs}
             result_df = uniform_neighbor_sample(G, **kwargs_to_pass)
             # Convert DataFrame into CuPy arrays for returning to the client
-            sources = result_df["sources"].to_cupy()
-            destinations = result_df["destinations"].to_cupy()
-            indices = result_df["indices"].to_cupy()
+            sources = result_df["majors"].to_cupy()
+            destinations = result_df["minors"].to_cupy()
+            indices = result_df["weight"].to_cupy()
 
         return UniformNeighborSampleResult(
             sources=sources,


python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
```diff
@@ -0,0 +1,349 @@
...
+            f"'vertex' type. input graph's vertex type is: {vertex_dtype} and got "
+            f"'start_list' of type: {start_list_dtypes}."
+        )
+        warnings.warn(warning_msg, UserWarning)
+        start_list = start_list.astype(vertex_dtype)
+
+    return start_list
+
+
+def heterogeneous_uniform_neighbor_sample(
```
rlratzel, line 59: Should this still have `_uniform` in the function name?

python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
```diff
@@ -0,0 +1,349 @@
...
+        sampling_result_array_dict = pylibcugraph_heterogeneous_biased_neighbor_sample(
+            resource_handle=ResourceHandle(),
+            input_graph=G._plc_graph,
+            start_vertex_list=start_list,
+            starting_vertex_label_offsets=starting_vertex_label_offsets,
+            vertex_type_offsets=vertex_type_offsets,
+            h_fan_out=fanout_vals,
+            num_edge_types=num_edge_types,
+            with_replacement=with_replacement,
+            do_expensive_check=True,
```
rlratzel, line 302: Should this be a parameter or at least hardcoded to `False`?

python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
```diff
@@ -0,0 +1,349 @@
...
+    if G.renumbered is True:
+        if isinstance(start_list, cudf.DataFrame):
+            start_list = G.lookup_internal_vertex_id(
+                start_list, start_list.columns
+            )
+        else:
+            start_list = G.lookup_internal_vertex_id(start_list)
+
+    if with_biases:
+        sampling_result_array_dict = pylibcugraph_heterogeneous_biased_neighbor_sample(
```
rlratzel, line 293: Are the calls exactly the same except for the PLC function itself? If so, you could trim this down quite a bit like this:
```suggestion
    if with_biases:
        sampling_function = pylibcugraph_heterogeneous_biased_neighbor_sample
    else:
        sampling_function = pylibcugraph_heterogeneous_uniform_neighbor_sample
```
Then just call it once as 
```
    sampling_result_array_dict = sampling_function( ... )
```

run_algo.py
rlratzel, line 1: Was this committed by mistake?

run_dummy.py
rlratzel, line 1: Was this committed by mistake?

run_plc.py
rlratzel, line 1: Was this committed by mistake?

test.txt
rlratzel, line 1: Was this committed by mistake?

rlratzel (changes requested):

python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
jnke2016: Right this has been resolved

python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
jnke2016: I set it to True when debugging and didn't set it back to False. It is resolved now

python/cugraph/cugraph/sampling/heterogeneous_neighbor_sample.py
jnke2016: That will work for now since all arguments are the same however, once we start supporting `biases` input argument for `biased_neighbor_sample`, it won't.

run_algo.py
jnke2016: Right those were not supposed to be committed and were only meant for debugging purposes. I removed them

rlratzel (approved):

alexbarghi-nv:
@jnke2016 I think we can resume this PR now that the code causing issues has been removed.

PR #4994, opened 2025-03-20
Temporal sampling implementation

ChuckHastings:
Temporal sampling implementation.  Sampling considers the time stamp of edges, if we arrive at a vertex `v` with timestamp `t1`, then when we depart from that vertex to continue sampling we only consider edges that occur after time `t1`.

PR includes C++ implementation and tests.

This significantly increases C++ compile time.  We'll try to improve in 25.10.


seunghwak (commented):
I thought about efficient implementation of temporal sampling especially considering that some seed vertices can be reached from multiple different paths and we need to apply multiple different temporal windows for the same seed vertex.

This can lead to many vertex partitions especially for power-law graphs.

And applying & creating graph-wise temporal mask can be pretty expensive if we need to do this many times.

We can apply a graph-wise temporal mask to set temporal window including the lower and upper bound of the start/end times for the entire set of seeds in multiple batches.

For a seed specific time window, I think adjusting bias values will lead to more efficient implementation.

We can tag a seed vertex with a time-stamp (https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/prims/per_v_random_select_transform_outgoing_e.cuh#L1092C28-L1092C72).

And when we set  the bias value (https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/prims/per_v_random_select_transform_outgoing_e.cuh#L1096), we can set the bias value to 0 if the edge is outside the seed specific time window.

I think this can lead to more efficient implementation than the current approach.

What do you think about this?

seunghwak:
And for uniform sampling, we may use a uniform sampling primitive for seeds that appear no more than once and use a biased sampling primitive for seeds that appear two or more times.

ChuckHastings:
> I thought about efficient implementation of temporal sampling especially considering that some seed vertices can be reached from multiple different paths and we need to apply multiple different temporal windows for the same seed vertex.
> 
> This can lead to many vertex partitions especially for power-law graphs.
> 
> And applying & creating graph-wise temporal mask can be pretty expensive if we need to do this many times.
> 
> We can apply a graph-wise temporal mask to set temporal window including the lower and upper bound of the start/end times for the entire set of seeds in multiple batches.
> 
> For a seed specific time window, I think adjusting bias values will lead to more efficient implementation.
> 
> We can tag a seed vertex with a time-stamp (https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/prims/per_v_random_select_transform_outgoing_e.cuh#L1092C28-L1092C72).
> 
> And when we set the bias value (https://github.com/rapidsai/cugraph/blob/branch-25.04/cpp/src/prims/per_v_random_select_transform_outgoing_e.cuh#L1096), we can set the bias value to 0 if the edge is outside the seed specific time window.
> 
> I think this can lead to more efficient implementation than the current approach.
> 
> What do you think about this?

So something akin to what node2vec does... return a bias of 0 if the edge time is invalid, return a bias of 1 if the edge time is valid.  Because we're operating on the tagged vertex, each vertex would have its own timestamp... therefore its own computed bias.

If my interpretation is correct, I think that would be a much simpler implementation and would probably result in significantly better performance in the cases where we end up with a high degree vertex that appears multiple times in the frontier.

seunghwak:
> erpretation is correct, I think that would be a much simpler implementation and would probably result in significantly better performance in the cases w

Yes, your interpretation is correct. I agree that this will be simpler & faster. For uniform sampling and to avoid the overhead of evaluating bias for every edge, we can use just a default uniform sampling for seeds that appear only once, and use bias values & tagging for seeds that appear more than once.

alexbarghi-nv:
@ChuckHastings what is the status of this PR?  Will this be completed in release 25.06?

ChuckHastings:
> @ChuckHastings what is the status of this PR? Will this be completed in release 25.06?

Should be.  I'm currently debugging the case where you pass a fanout value of -1.  Then it should be functional and ready for review.

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -200,6 +200,33 @@ shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
...
+ */
+template <typename vertex_t, typename value1_t, typename value2_t>
+std::
+  tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<value1_t>, rmm::device_uvector<value2_t>>
+  shuffle_int_vertex_two_value_pairs_to_local_gpu_by_vertex_partitioning(
+    raft::handle_t const& handle,
+    rmm::device_uvector<vertex_t>&& vertices,
+    rmm::device_uvector<value1_t>&& values1,
+    rmm::device_uvector<value2_t>&& values2,
+    raft::host_span<vertex_t const> vertex_partition_range_lasts);
```
seunghwak, line 228: Should we really create this function? I think this can be handled by explicitly instantiating `shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning` with `value_t=thrust::tuple<value1_t, value2_t>`.

cpp/include/cugraph/edge_property.hpp
```diff
@@ -203,4 +204,16 @@ auto view_concat(edge_property_view_t<edge_t, Iters, Types> const&... views)
     edge_partition_concat_value_firsts, first_view.edge_counts());
 }
 
+template <typename... Ts, std::size_t... Is>
+auto view_concat_impl(std::tuple<Ts...> const& tuple, std::index_sequence<Is...>)
+{
+  return view_concat(std::get<Is>(tuple)...);
+}
```
seunghwak, line 211: We may better move this to the detail namespace.

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -602,6 +602,382 @@ heterogeneous_biased_neighbor_sample(
...
+  std::optional<edge_property_view_t<edge_t, edge_time_t const*>> edge_end_time_view,
+  edge_property_view_t<edge_t, bias_t const*> edge_bias_view,
+  raft::device_span<vertex_t const> starting_vertices,
+  std::optional<raft::device_span<int32_t const>> starting_vertex_labels,
+  std::optional<raft::device_span<int32_t const>> label_to_output_comm_rank,
+  raft::host_span<int32_t const> fan_out,
+  edge_type_t num_edge_types,
+  sampling_flags_t sampling_flags,
+  bool do_expensive_check = false);
+
```
seunghwak, line 980: Sorry for nitpicking,

but we are declaring functions in the order of

homogeneous_uniform
heterogeneous_uniform
homogeneous_biased
heterogeneous_biased

homogeneous_uniform_temporal
homogeneous_biased_temporal
heterogenous_uniform_temporal
hteterogeneous_biased_temporal.

If we exclude the temporal part from the last four, the order of first four and the last four do not match... This can possibly confuse users trying to find a function they need...

cpp/include/cugraph/utilities/dataframe_buffer.hpp
```diff
@@ -202,6 +202,15 @@ auto get_dataframe_buffer_begin(BufferType& buffer)
...
 }
 
+template <typename BufferType,
+          typename std::enable_if_t<is_std_tuple_of_arithmetic_device_spans<
+            std::remove_cv_t<BufferType>>::value>* = nullptr>
+auto get_dataframe_buffer_begin(BufferType& buffer)
+{
+  return detail::get_dataframe_buffer_begin_tuple_impl(
+    std::make_index_sequence<std::tuple_size<BufferType>::value>(), buffer);
+}
```
seunghwak, line 212: If we plan to add this to support std::tuple of device_span objects, we need to add for cbegin, end, and cend as well.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -192,6 +192,25 @@ OutputIterator copy_if_mask_set(raft::handle_t const& handle,
     is_equal_t<bool>{true});
 }
 
+template <typename InputIterator,
+          typename MaskIterator,  // should be packed bool
+          typename OutputIterator>
+OutputIterator copy_if_mask_not_set(raft::handle_t const& handle,
```
seunghwak, line 198: I was slightly unhappy with this name and asked chatgpt about naming.

```
In C++, we have a copy_if_mask_set function, which copies if the mask value is 1. What should be the name of the function that copies if the mask value is 0?


ChatGPT said:
A good, consistent name for the function that copies when the mask value is 0 (i.e., the logical inverse of copy_if_mask_set) could be:

copy_if_mask_unset

Reasoning:
Consistency: It mirrors the structure of copy_if_mask_set.

Clarity: “unset” clearly implies the mask bit is 0.

Readability: Someone reading both copy_if_mask_set and copy_if_mask_unset will immediately understand the opposite behaviors.

Other possible but less ideal alternatives (depending on your naming style) include:

copy_if_mask_clear

copy_if_mask_false

copy_if_mask_zero

Would you like suggestions for the complementary function names in a boolean or SIMD context too?
```

ChatGPT's suggestions sound reasonable to me.

cpp/include/cugraph/utilities/thrust_tuple_utils.hpp
```diff
@@ -169,6 +171,14 @@ struct is_std_tuple_of_arithmetic_vectors<std::tuple<rmm::device_uvector<Ts>...>
   static constexpr bool value = (... && std::is_arithmetic_v<Ts>);
 };
 
+template <typename T>
+struct is_std_tuple_of_arithmetic_device_spans : std::false_type {};
+
+template <typename... Ts>
+struct is_std_tuple_of_arithmetic_device_spans<std::tuple<raft::device_span<Ts>...>> {
+  static constexpr bool value = (... && std::is_arithmetic_v<Ts>);
+};
```
seunghwak, line 180: To be consistent, I think we should rename this function to "is_std_tuple_of_arithmetic_spans" or rename the above "is_std_tuple_of_vectors" to "is_std_tuple_of_device_vectors".

seunghwak: I am a bit inclined to renaming this function as in some years, the distinction between the host vector and device vector may get significantly blurred.

cpp/include/cugraph/utilities/thrust_tuple_utils.hpp
```diff
@@ -311,6 +321,7 @@ __host__ __device__
 
 // a temporary function to emulate thrust::tuple_cat (not supported) using std::tuple_cat (should
 // retire once thrust::tuple is replaced with cuda::std::tuple)
+
```
seunghwak, line 324: Why did you add this spacing?

cpp/src/prims/kv_store.cuh
```diff
@@ -172,7 +172,7 @@ struct kv_binary_search_store_device_view_t {
   {
   }
 
-  __device__ value_type find(key_type key) const
+  __host__ __device__ value_type find(key_type key) const
```
seunghwak, line 175: Why did you add __host__ here? I don't think this function is host callable, and did you add this to fix some build error? There might be a better way than marking this only device callable function to host device.

seunghwak (commented):
Review part 1

cpp/src/sampling/detail/gather_one_hop_edgelist_impl.cuh
```diff
@@ -48,400 +144,551 @@ struct return_edges_with_properties_e_op {
...
+  key_bucket_t<vertex_t, tag_t, multi_gpu, false> const& key_list;
+  std::optional<raft::host_span<uint8_t const>> gather_flags;
+  bool do_expensive_check{false};
+
+  template <typename TupleType>
+  auto concatenate_views(TupleType edge_properties)
+  {
+    if constexpr (std::tuple_size_v<TupleType> == 0) {
+      return edge_property_view_type_t<graph_view_t<vertex_t, edge_t, false, multi_gpu>,
+                                       cuda::std::nullopt_t>{};
```
seunghwak, line 339: Can't we return dummy property here? edge_property_view_type_t is designed to work with arithmetic or thrust tuple of arithmetic types only.

cpp/src/sampling/detail/gather_one_hop_edgelist_impl.cuh
```diff
@@ -48,400 +144,551 @@ struct return_edges_with_properties_e_op {
...
-      } else {
-        edge_value_view = *edge_weight_view;
-      }
+      return view_concat(edge_properties);
     }
-  } else {
-    if constexpr (has_edge_id) {
-      if constexpr (has_edge_type) {
-        edge_value_view = view_concat(*edge_id_view, *edge_type_view);
+  }
```
seunghwak, line 345: Should we replicate this concatentate_views() function in every functor? They look very similar and can't we avoid this replication?

seunghwak (commented):
Part 2

cpp/src/sampling/detail/gather_one_hop_edgelist_impl.cuh
```diff
@@ -16,414 +16,717 @@
...
+  if constexpr (std::tuple_size_v<TupleType> == 0) {
+    return edge_dummy_property_view_t{};
+  } else if constexpr (std::tuple_size_v<TupleType> == 1) {
+    return std::get<0>(edge_properties);
+  } else {
+    return view_concat(edge_properties);
+  }
+}
+
+// FIXME:  Duplicated across files...
```
seunghwak, line 70: Can't we move this to detail/sampling_utils.hpp to avoid duplication?

cpp/src/sampling/detail/shuffle_and_organize_output_impl.cuh
```diff
@@ -16,9 +16,11 @@
 
 #pragma once
 
+#include "cugraph/utilities/shuffle_comm.cuh"
 #include "prims/update_edge_src_dst_property.cuh"  // ??
```
seunghwak, line 20: This might be from a previous PR but why ?? If we are not using this, can't we delete?

cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
```diff
@@ -0,0 +1,97 @@
...
+  raft::device_span<edge_time_t const> vertex_times,
+  edge_property_view_t<edge_t, uint32_t*, bool> edge_time_mask_view)
+{
+  edge_time_t const STARTING_TIME{std::numeric_limits<edge_time_t>::min()};
+
+  edge_src_property_t<edge_t, edge_time_t, false> edge_src_times(handle, graph_view);
+
+#if 0
+  // FIXME:  This call to update_edge_src_property seems like what I want, but it
+  //         doesn't work.
```
seunghwak, line 50: Do you know why? What happened with this?

seunghwak (approved):
LGTM (besides few comments about minor cosmetic issues)

cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
ChuckHastings: This variation of the call is only used in a few places.  MG SSSP, and Random Walks are the only places I see it.

Not sure if it's broken there and we haven't noticed, or if I'm doing something wrong here.  But I printed the relevant contents of the arrays and they were incorrect.  Since I use them as indices into device arrays, I could clearly see that the wrong values affected the output.

I spent a little time trying to diagnose, but didn't have time to really dig in.  It's possible I've done something incorrect in the setup, but setting do_expensive_check didn't reveal anything wrong.

seunghwak: How can I reproduce this? I can dig in to see what's going awry here.

ChuckHastings: I'll try and create a reproducer for you.

cpp/include/cugraph/arithmetic_variant_types.hpp
```diff
@@ -45,9 +45,11 @@ using const_arithmetic_device_span_t = std::variant<std::monostate,
                                                     raft::device_span<int64_t const>,
                                                     raft::device_span<size_t const>>;
 
-template <typename edge_t>
+template <typename vertex_t, typename edge_t>
 using edge_arithmetic_property_view_t =
   std::variant<std::monostate,
+               cugraph::edge_dummy_property_view_t,
+               cugraph::edge_multi_index_property_view_t<edge_t, vertex_t>,
```
seunghwak: I am not sure about including these special types here. "dummy property" is not an arithmetic property... And these types are missing in `edge_arithmetic_property_mutable_view_t`. This looks inconsistent.

And we may consider including (u)int8_t, (u)int16_t, uint32_t, and uint64_t in the future.

cpp/include/cugraph/arithmetic_variant_types.hpp
```diff
@@ -158,31 +160,18 @@ auto variant_type_dispatch(const_arithmetic_device_span_t& property, func_t func
...
-    return func(prop);
+  switch (property.index()) {
+    case 1: return func(std::get<1>(property));
+    case 2: return func(std::get<2>(property));
+    case 3: return func(std::get<3>(property));
+    case 4: return func(std::get<4>(property));
+    case 5: return func(std::get<5>(property));
+    case 6: return func(std::get<6>(property));
+    case 7: return func(std::get<7>(property));
+    default: CUGRAPH_FAIL("Variant not initialized");
```
seunghwak, line 127: So, we are mixing "if .. else if .. else if .. else" statements and switch statements... Should we better stick to one approach?

cpp/include/cugraph/arithmetic_variant_types.hpp
```diff
@@ -158,31 +160,18 @@ auto variant_type_dispatch(const_arithmetic_device_span_t& property, func_t func
...
-    auto& prop = std::get<cugraph::edge_property_view_t<edge_t, size_t const*>>(property);
-    return func(prop);
+  switch (property.index()) {
+    case 1: return func(std::get<1>(property));
+    case 2: return func(std::get<2>(property));
+    case 3: return func(std::get<3>(property));
+    case 4: return func(std::get<4>(property));
+    case 5: return func(std::get<5>(property));
+    case 6: return func(std::get<6>(property));
+    case 7: return func(std::get<7>(property));
```
seunghwak: And we may define an enum class for this purpose.

Something like
``
enum class arithmetic_property_variant_index_t {
  INVALID = 0;  // monostate
  FLOAT = 1;
  DOUBLE = 2;
  INT32 = 3,
  INT64 = 4,
  UINT64 = 5
};
```
?

File: cpp/include/cugraph/sampling_functions.hpp [hunk 17]
```diff
@@ -602,6 +602,382 @@ heterogeneous_biased_neighbor_sample(
...
+  raft::device_span<vertex_t const> starting_vertices,
+  std::optional<raft::device_span<int32_t const>> starting_vertex_labels,
+  std::optional<raft::device_span<int32_t const>> label_to_output_comm_rank,
+  raft::host_span<int32_t const> fan_out,
+  sampling_flags_t sampling_flags,
+  bool do_expensive_check = false);
+
+/**
+ * @ingroup sampling_functions_cpp
+ * @brief Heterogeneous Temporal Uniform Neighborhood Sampling.
```
seunghwak: Sorry for nitpicking but,

"Homogeneous Uniform Temporal Neighborhood Sampling."
https://github.com/rapidsai/cugraph/pull/4994/files#diff-2cdbfa97ec1f07c1f2c343715d235dd701988aaf562097994dcd25072816026aR607

and

"Heterogeneous Temporal Uniform Neighborhood Sampling."

sounds inconsistent. We should constantly use "Uniform Temporal" (as this better matches the function name as well).

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -602,6 +602,382 @@ heterogeneous_biased_neighbor_sample(
...
+  raft::device_span<vertex_t const> starting_vertices,
+  std::optional<raft::device_span<int32_t const>> starting_vertex_labels,
+  std::optional<raft::device_span<int32_t const>> label_to_output_comm_rank,
+  raft::host_span<int32_t const> fan_out,
+  sampling_flags_t sampling_flags,
+  bool do_expensive_check = false);
+
+/**
+ * @ingroup sampling_functions_cpp
+ * @brief Heterogeneous Temporal Biased Neighborhood Sampling.
```
seunghwak: "Temporal Biased"=>"Biased Temporal" for consistency.

cpp/include/cugraph/utilities/dataframe_buffer.hpp
seunghwak: We added for cbegin & cend, but I don't see one for end().

cpp/src/detail/shuffle_wrappers.hpp
```diff
@@ -135,22 +136,22 @@ rmm::device_uvector<vertex_t> shuffle_int_vertices_to_local_gpu_by_vertex_partit
  * memory resource is used). The shuffled vertex value pairs will also be stored in the buffer type
  * dictated by this parameter.
  *
- * @return tuple containing device vector of shuffled vertices and device vector of corresponding
+ * @return tuple containing device vectors of shuffled vertices and corresponding
  *         values
  */
-template <typename vertex_t, typename value_t>
-std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<value_t>>
+template <typename vertex_t, typename value_vector_t>
```
seunghwak: Is `value_vector_t` `dataframe_buffer_type_t<value_t>`?

cpp/src/sampling/detail/common_utilities.cuh
```diff
@@ -0,0 +1,56 @@
...
+    return std::get<0>(edge_properties);
+  } else {
+    return view_concat(edge_properties);
+  }
+}
+
+template <size_t input_tuple_pos,
+          size_t output_tuple_pos,
+          bool Flag,
+          bool... Flags,
```
seunghwak: Should these values be template parameters?

I think these can be input parameters...

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+struct return_edges_with_label_op {
+  template <typename vertex_t, typename label_t>
+  cuda::std::tuple<vertex_t, vertex_t, label_t> __device__
+  operator()(cuda::std::tuple<vertex_t, label_t> tagged_src,
+             vertex_t dst,
+             cuda::std::nullopt_t,
+             cuda::std::nullopt_t,
+             cuda::std::nullopt_t) const
+  {
+    return cuda::std::make_tuple(cuda::std::get<0>(tagged_src), dst, cuda::std::get<1>(tagged_src));
```
seunghwak, line 59: Is this different from `return tagged_src`?

seunghwak: And isn't `return_edge_with_label_op` a misnomer?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+  }
+};
+
+struct return_edges_with_index_and_position_op {
```
seunghwak, line 63: Is `index` here `multi-edge index`? `position` => `label`?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+  rmm::device_uvector<vertex_t> result_dsts(0, handle.get_stream());
+  std::vector<arithmetic_device_uvector_t> result_properties{};
+  std::optional<rmm::device_uvector<int32_t>> result_labels{std::nullopt};
+
+  if (edge_property_views.size() == 0) {
+    CUGRAPH_EXPECTS(!edge_type_view,
+                    "Can't specify type filtering without edge type as a property");
+
+    // Don't care if graph is multigraph, since we don't need to transform_gather_e any properties
+    std::tie(result_srcs, result_dsts, result_labels) = simple_gather_one_hop_not_multigraph(
```
seunghwak: In this context,
`simple_gather_one_hop_not_multigraph` sounds like a misnomer.

Should we better rename the functions to "simple_gather_one_hop_with_multi_edge_indices" and "simple_gather_one_hop_without_multi_edge_indices"?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+                                                      edge_src_dummy_property_t{}.view(),
+                                                      edge_dst_dummy_property_t{}.view(),
+                                                      property_view,
+                                                      return_edge_property_t{},
+                                                      tmp.begin());
+
+                          result_properties.push_back(arithmetic_device_uvector_t{std::move(tmp)});
+                        }
+                      });
+                  });
```
seunghwak: Should we better move this to a separate function? I guess this routine should be same for both gathering one hop & random sampling.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+};
+
+}  // namespace
+
+template <typename vertex_t, typename edge_t, typename tag_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<edge_t>,
+           std::optional<rmm::device_uvector<tag_t>>>
+simple_gather_one_hop_multigraph(
```
seunghwak: Should we better move simple gather functions to the anonymous namespace? I assume these are not public functions?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+          handle, std::move(*tmp_edge_indices), marked_entry_span, keep_count);
+      }
+
+      edge_list.clear();
+      edge_list.insert(result_srcs.begin(),
+                       result_srcs.end(),
+                       result_dsts.begin(),
+                       tmp_edge_indices ? std::optional<edge_t*>{tmp_edge_indices->begin()}
+                                        : std::optional<edge_t*>{std::nullopt});
+    }
```
seunghwak: Should we better move this to a separate function? (so we can use this for both temporal & non-temporal cases).

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+{
+  constexpr bool store_transposed = false;
+
+  rmm::device_uvector<vertex_t> result_srcs(0, handle.get_stream());
+  rmm::device_uvector<vertex_t> result_dsts(0, handle.get_stream());
+  std::vector<arithmetic_device_uvector_t> result_properties{};
+  std::optional<rmm::device_uvector<int32_t>> result_labels{std::nullopt};
+
+  // FIXME:  Modify sampling to take raft::host_span of arithmetic edge property views directly
+  // rather than creating and unpacking them
```
seunghwak, line 374: Is this FIXME still relevant?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+  if (edge_property_views.size() == 0) {
+    CUGRAPH_FAIL("Temporal requires at least one edge property - the time");
```
seunghwak: I guess just adding CUGRAPH_EXPECTS(edge_property_views.size() > 0, "..."); will be more concise than placing all the remaining code under the else block (creating just one additional level of indentation).

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+          handle, std::move(*tmp_edge_indices), marked_entry_span, keep_count);
+      }
+
+      edge_list.clear();
+      edge_list.insert(result_srcs.begin(),
+                       result_srcs.end(),
+                       result_dsts.begin(),
+                       tmp_edge_indices ? std::optional<edge_t*>{tmp_edge_indices->begin()}
+                                        : std::optional<edge_t*>{std::nullopt});
+    }
```
seunghwak: This code looks identical to the non-temporal case.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+      } else {
+        kv_store = kv_store_t<size_t, thrust::tuple<edge_time_t, label_t>, true>(
+          vertex_label_time_positions.begin(),
+          vertex_label_time_positions.end(),
+          thrust::make_zip_iterator(active_major_times.begin(),
+                                    active_major_labels->begin()),  // multi_gpu is different
+          thrust::make_tuple(edge_time_t{-1}, label_t{-1}),
+          true,
+          handle.get_stream());
+      }
```
seunghwak: I think we can defer creation of kv_store after simple gather.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+  if (edge_property_views.size() == 0) {
+    CUGRAPH_FAIL("Temporal requires at least one edge property - the time");
+  } else {
+    using label_t = int32_t;
+
+    cugraph::edge_bucket_t<vertex_t, edge_t, !store_transposed, true, false> edge_list(
+      handle, graph_view.is_multigraph());
+
+    // Only used if active_major_labels is set
+    kv_store_t<size_t, thrust::tuple<edge_time_t, label_t>, true> kv_store{handle.get_stream()};
```
seunghwak: So, I assume we can avoid this complication if we update the primitives to take cuda::std::tuple as a label.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+                                edge_dst_dummy_property_t{}.view(),
+                                edge_time_view,
+                                return_edge_property_t{},
+                                edge_times.begin());
+
+    auto [keep_count, marked_entries] = detail::mark_entries(
+      handle,
+      edge_times.size(),
+      [d_tmp           = edge_times.data(),
+       d_tmp_positions = tmp_positions->data(),
```
seunghwak: tmp_positions.has_value() is false here if labels are not used, no? (tmp_times is valid instead).

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -0,0 +1,662 @@
...
+    result_labels = rmm::device_uvector<label_t>(keep_count, handle.get_stream());
+    thrust::transform(
+      handle.get_thrust_policy(),
+      tmp_positions->begin(),
+      tmp_positions->end(),
+      result_labels->begin(),
+      [kv_store_view = kv_binary_search_store_device_view_t<decltype(kv_store.view())>{
+         kv_store.view()}] __device__(auto pos) {
+        return cuda::std::get<1>(kv_store_view.find(pos));
+      });
```
seunghwak: This code is valid only when tmp_positions.has_value() is true, right?

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -66,57 +82,137 @@ prepare_next_frontier(
...
 
-  if (frontier_vertex_labels) {
-    thrust::copy(handle.get_thrust_policy(),
-                 sampled_dst_vertex_labels->begin(),
-                 sampled_dst_vertex_labels->end(),
-                 frontier_vertex_labels->begin());
+  std::for_each(sampled_dst_vertices.begin(),
+                sampled_dst_vertices.end(),
+                [&handle, &frontier_vertices, &current_pos](auto& list) {
+                  if (list.size() > 0)
```
seunghwak: This check is not necessary for thrust calls.

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -66,57 +82,137 @@ prepare_next_frontier(
...
     }
   }
 
   if (frontier_vertex_labels) {
-    auto begin_iter =
-      thrust::make_zip_iterator(frontier_vertices.begin(), frontier_vertex_labels->begin());
-    thrust::sort(handle.get_thrust_policy(), begin_iter, begin_iter + frontier_vertices.size());
+    if (frontier_vertex_times) {
+      auto begin_iter =
+        thrust::make_zip_iterator(frontier_vertices.begin(), frontier_vertex_labels->begin());
```
seunghwak: begin_iter is same for both the if and else cases, so it can be moved outside the if - else blocks.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -32,44 +39,121 @@
...
+
+    if constexpr (std::is_same_v<key_t, vertex_t>)
+      src = optionally_tagged_src;
+    else
+      src = thrust::get<0>(optionally_tagged_src);
+
+    if constexpr (std::is_same_v<edge_properties_t, cuda::std::nullopt_t>) {
+      return thrust::make_tuple(src, dst);
+    } else {
+      return thrust::make_tuple(src, dst, edge_properties);
```
seunghwak: thrust::make_tuple => cuda::std::make_tuple here?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -84,387 +168,991 @@ struct segmented_fill_t {
...
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
            rmm::device_uvector<vertex_t>,
-           std::optional<rmm::device_uvector<weight_t>>,
-           std::optional<rmm::device_uvector<edge_t>>,
-           std::optional<rmm::device_uvector<edge_type_t>>>
-sample_edges_with_edge_values(
+           arithmetic_device_uvector_t,
+           std::optional<rmm::device_uvector<size_t>>>
+sample_with_one_property(
```
seunghwak, line 176: One property or "0 or 1" property? Are we doing this for performance? (compare to a sampling function that just returns src, dst, (optional multi-edge index)). I guess the difference may not be negligible for simple uniform sampling but for more complex sampling, we may not see significant difference in performance?

And completely separating edge sampling from property gathering will also have smaller peak memory usage (and may be faster in some cases as well for non-trivial sampling methods. We extract all the neighbors of seeds with edge property values. For high degree vertices, only the subset of extracted edge property values will actually be used... So, completely separating edge sampling from edge property gathering might be actually faster in certain cases).

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -84,387 +168,991 @@ struct segmented_fill_t {
...
+                                                      edge_src_dummy_property_t{}.view(),
+                                                      edge_dst_dummy_property_t{}.view(),
+                                                      property_view,
+                                                      return_edge_property_t{},
+                                                      tmp.begin());
+
+                          edge_properties.push_back(arithmetic_device_uvector_t{std::move(tmp)});
+                        }
+                      });
+                  });
```
seunghwak: Is this code any different from the gather case?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -84,387 +168,991 @@ struct segmented_fill_t {
...
+    thrust::make_zip_iterator(active_majors.begin(), active_major_times.begin()),
+    thrust::make_zip_iterator(active_majors.end(), active_major_times.end()));
+
+  rmm::device_uvector<vertex_t> majors(0, handle.get_stream());
+  rmm::device_uvector<vertex_t> minors(0, handle.get_stream());
+  std::vector<arithmetic_device_uvector_t> edge_properties{};
+  std::optional<rmm::device_uvector<int32_t>> labels{std::nullopt};
+  std::optional<rmm::device_uvector<size_t>> sample_offsets{std::nullopt};
+
+  if (edge_property_views.size() == 1) {
```
seunghwak, line 915: No need to consider "if (edge_property_views.size() == 0) {"?

cpp/src/sampling/detail/common_utilities.cuh
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  if constexpr (sizeof...(Flags) > 0) {
+    if constexpr (Flag) {
+      move_results<input_tuple_pos + 1, output_tuple_pos + 1, Flags...>(input_tuple, output_tuple);
+    } else {
+      move_results<input_tuple_pos, output_tuple_pos + 1, Flags...>(input_tuple, output_tuple);
+    }
+  }
+}
```
seunghwak: Are we using this function anywhere in the code base? Can we just delete this?

cpp/src/sampling/detail/common_utilities.cuh
```diff
@@ -0,0 +1,56 @@
...
+auto constexpr concatenate_views(TupleType edge_properties)
+{
+  if constexpr (std::tuple_size_v<TupleType> == 0) {
+    return cugraph::edge_dummy_property_view_t{};
+  } else if constexpr (std::tuple_size_v<TupleType> == 1) {
+    return std::get<0>(edge_properties);
+  } else {
+    return view_concat(edge_properties);
+  }
+}
```
seunghwak: Are we using this anywhere in the codebase? (as we are now separately calling transform_gather_e for each property). If not, can we just delete this? And if we can delete these, we can also delete the view_concat related functions added in edge_property.hpp.

cpp/src/sampling/neighbor_sampling_impl.hpp
```diff
@@ -200,19 +229,132 @@ neighbor_sample_impl(raft::handle_t const& handle,
...
+      result_sizes.push_back(srcs.size());
+      result_src_vectors.push_back(std::move(srcs));
+      result_dst_vectors.push_back(std::move(dsts));
+
+      if (weights) { (*result_weight_vectors).push_back(std::move(*weights)); }
+      if (edge_ids) { (*result_edge_id_vectors).push_back(std::move(*edge_ids)); }
+      if (edge_types) { (*result_edge_type_vectors).push_back(std::move(*edge_types)); }
+      if (labels) { (*result_label_vectors).push_back(std::move(*labels)); }
+
+#else
```
seunghwak: Delete the dead code.

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -0,0 +1,1108 @@
...
+            .push_back(std::move(
+              std::get<rmm::device_uvector<edge_time_t>>(sampled_edge_properties[pos++])));
+          if (edge_end_time_view) {
+            (*result_edge_end_time_vectors)
+              .push_back(std::move(
+                std::get<rmm::device_uvector<edge_time_t>>(sampled_edge_properties[pos++])));
+          }
+          if (labels) { (*result_label_vectors).push_back(std::move(*labels)); }
+        }
+#else
```
seunghwak: Delete the dead code

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -0,0 +1,1108 @@
...
```
seunghwak: Delete the dead code.

cpp/src/utilities/tuple_with_optionals_dispatching.hpp
```diff
@@ -0,0 +1,49 @@
...
+                              f, std::tuple_cat(tuple, std::make_tuple(std::move(*t1))), ts...)
+                          : tuple_with_optionals_dispatch_impl<Flags..., false>(f, tuple, ts...);
+  }
+}
+
+template <typename Functor, typename... Ts>
+auto tuple_with_optionals_dispatch(Functor f, Ts... ts)
+{
+  return tuple_with_optionals_dispatch_impl(f, std::make_tuple(), ts...);
+}
```
seunghwak: Are we using this in any place?

cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
seunghwak: We should debug this, the else path is clearly more expensive.

And in case we have both edge start & end time, can we just pass both the start & end time as edge property values and check (seed_time > edge start time) && (seed time < edge end time)?

Then, we don't need to perform temporal_partition_vertices nor update_temporal_edge_mask; much simpler code. Or aren't we doing this as this doubles the cases we need to consider (so double the binary size in temporal sampling edges?)

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -0,0 +1,1108 @@
...
+      has_duplicates_size = frontier_vertices_has_duplicates.size();
+
+      if constexpr (multi_gpu) {
+        no_duplicates_size = host_scalar_allreduce(
+          handle.get_comms(), no_duplicates_size, raft::comms::op_t::SUM, handle.get_stream());
+        has_duplicates_size = host_scalar_allreduce(
+          handle.get_comms(), no_duplicates_size, raft::comms::op_t::SUM, handle.get_stream());
+      }
+
+      if (no_duplicates_size > 0) {
```
seunghwak, line 256: Have you considered the case `graph_view` already has an attached mask?

seunghwak: In this case, you need to clear edge mask before attaching a new one. Also, edge_time_mask's values for masked-out edges won't be updated in primitive calls, so they may have undefined values.

cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
seunghwak: Or we may set the mask based on the minimum time attached with a seed vertex (so we can reduce the size of the extracted subgraph in the sampling primitive).

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -0,0 +1,1108 @@
...
+        update_temporal_edge_mask(
+          handle,
+          graph_view,
+          edge_start_time_view,
```
seunghwak, line 260: No need to consider edge_end_time_view here?

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -0,0 +1,1108 @@
...
+        if (edge_end_time_view) edge_property_views.push_back(*edge_end_time_view);
+
+        auto [srcs, dsts, sampled_edge_properties, labels] =
+          temporal_sample_edges<vertex_t, edge_t, edge_time_t, multi_gpu>(
+            handle,
+            rng_state,
+            graph_view,
+            raft::host_span<edge_arithmetic_property_view_t<edge_t, vertex_t>>{
+              edge_property_views.data(), edge_property_views.size()},
+            edge_start_time_view,
```
seunghwak, line 368: What about edge_end_time_view here?

seunghwak: Or are we filtering based on edge_end_time later?

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: I'm thinking I'd like to move each of these to the switch approach.  It's a bit more concise.  I forgot to go back and update the previous functions.  I can do that.

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: There's one place in the current implementation that requires this.  I'll see about refactoring that.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Will fix

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: I think I'll add a CUGRAPH_EXPECTS and a FIXME on this one.  Fast fail at the beginning if the graph_view has an attached max, FIXME to indicate that we should support that.

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: The temporal sampling only needs to support a single time for guaranteeing edges are selected in time sequence.  By convention we're using the start time as that single time.

The end time was added for different temporal use cases that need both start/stop time.

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: Removing them in the next push.

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: Added in next push.

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: Changed to switch in next push.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Fixed in next push

cpp/src/detail/shuffle_wrappers.hpp
ChuckHastings: Yes.  But I'm not quite sure how to switch that.  We can chat tomorrow.

cpp/include/cugraph/utilities/dataframe_buffer.hpp
ChuckHastings: Added in next push

cpp/src/sampling/detail/common_utilities.cuh
ChuckHastings: This whole file is now obsolete... deleted in next push.

Also deleted view_concat changes from edge_property.hpp

cpp/src/sampling/detail/common_utilities.cuh
ChuckHastings: I think they were necessary to be resolved at compile time.

Irrelevant now... deleted this whole file.

cpp/src/sampling/detail/common_utilities.cuh
ChuckHastings: Deleted this whole file.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: > Is this different from `return tagged_src`?

Yes, `dst` is inserted between the two elements.

> And isn't `return_edge_with_label_op` a misnomer?

It's returning `(src, dst, label)`.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yes and maybe.

* `index` is the particular multi-edge index for the edge in question.
* `position` might be the label, but it also might be the key for the time/label combination stored in the kvstore.

If you have better suggestions for names, I'm happy to modify.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Fixed in next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yes, although perhaps not in the ideal location.

As a potential future change I want to explore having the high level sampling functions take `  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views` instead of explicitly enumerating each edge property view as an optional parameter.  That will improve the compile time for those functions (and support further edge properties if any get added).

This FIXME might be better located in the neighbor_sampling and temporal_sampling files.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Will explore

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Will explore

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yes, this and some of the trickiness in the edge ops above are to get around that limitation.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: We still need to populate `vertex_label_time_positions` first, since that's what we use as the tag.  But I could defer the creation/population of the kvstore until later.  I did them together to keep the hack all in one place.  If you'd rather I can split that up.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yes, I'll look at moving that into a common function.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I don't have a test case where we don't use labels.  I'm not sure how common that case is.

Fixed code in next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yes, moved inside above if block in next push.

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Fixed in next push.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Technically, 0 or 1 property.  The function takes one property view, I pass in the dummy edge property view, I pass in a single edge property, or I pass in the multi-edge index property view.

I didn't measure performance, I assumed that if there was only one property then just sampling with that one property would be faster than sampling and then gathering the edges for that one property.  If you think that's not the case the code would certainly be a bit simpler if we just sampled with or without the multi-edge index.

I'm inclined to defer this to a later optimization rather than delay merging this, as that would take a bit more effort to refactor (unless I'm misunderstanding your point).

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: No, I'll follow up by making that a separate function and calling it here.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: In the temporal case, there must at least be a time property.

I added a CUGRAPH_EXPECTS to verify this in the next push.

cpp/src/sampling/neighbor_sampling_impl.hpp
ChuckHastings: Fixed in next push

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: As mentioned above, edge_end_time is not used for filtering.

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: Fixed in next push

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: Fixed in next push

cpp/src/utilities/tuple_with_optionals_dispatching.hpp
ChuckHastings: Not any longer.  Deleted file in next push.

cpp/include/cugraph/arithmetic_variant_types.hpp
```diff
@@ -26,6 +26,15 @@
...
 namespace cugraph {
 
+enum class arithmetic_property_variant_index_t {
+  INVALID = 0,  // monostate
+  FLOAT   = 1,
+  DOUBLE  = 2,
+  INT32   = 3,
+  INT64   = 4,
+  UINT64  = 5
+};
```
seunghwak: Yeah, this is not used, so you can just delete this.

seunghwak (approved):
LGTM

PR #4999, opened 2025-03-21
Port all conda recipes to `rattler-build`

gforsyth:
Port all condabuild recipes over to use `rattler-build` instead.

Contributes to rapidsai/build-planning#47

- To satisfy `rattler-build`, this changes all the licenses in the `pyproject.toml` files to the SPDX-compliant `Apache-2.0` instead of `Apache 2.0`


conda/recipes/cugraph/recipe.yaml
```diff
@@ -0,0 +1,116 @@
...
+  version: ${{ version }}
+
+source:
+  path: ../../..
+
+build:
+  string: cuda${{ cuda_major }}_py${{ py_buildstring }}_${{ date_string }}_${{ head_rev }}
+  files:
+    exclude:
+      - '*libarrow.so.*gdb.py'
```
bdice, line 25: Where does cuGraph end up with libarrow bits?

conda/recipes/cugraph/recipe.yaml
```diff
@@ -0,0 +1,116 @@
...
+  prefix_detection:
+    # See https://github.com/rapidsai/build-planning/issues/160
+    # Blanket ignore here as there are quite a few shared objects shipped in cuml
```
bdice, line 28:
```suggestion
    # Blanket ignore here as there are quite a few shared objects shipped in cugraph
```

conda/recipes/libcugraph/recipe.yaml
```diff
@@ -0,0 +1,268 @@
...
+            - libcusparse-dev ${{ cuda11_libcusparse_host_version }}
+          else:
+            - cuda-cudart-dev
+            - libcublas-dev
+            - libcurand-dev
+            - libcusolver-dev
+            - libcusparse-dev
+      run:
+        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+        - libraft =${{ minor_version }}
```
bdice, line 126: A couple checks on the RAPIDS dependencies:

Do we need a `libraft` or `librmm` in `host` too? Or just in the top-level cache?

And are we relying on `rapids-logger`'s run-exports to put it into `run`?

conda/recipes/libcugraph/recipe.yaml
```diff
@@ -0,0 +1,268 @@
...
+          # See https://github.com/rapidsai/build-planning/issues/160
+          - lib/libcugraph_etl.so
+      string: cuda${{ cuda_major }}_${{ date_string }}_${{ head_rev }}
+    requirements:
+      build:
+        - cmake ${{ cmake_version }}
+        - ${{ stdlib("c") }}
+      host:
+        - cuda-version =${{ cuda_version }}
+        - libcudf =${{ minor_version }}
```
bdice, line 185: Does this need `libcudf` in `run`? Check the linking, I think `libcugraph_etl.so` links to libcudf but I could be wrong.

conda/recipes/libcugraph/recipe.yaml
gforsyth: There were no linking errors, so I think we're good?  

`rapids-logger` comes with `librmm` (which is also how `cugraph` loads it, so that seems consistent?)

conda/recipes/cugraph/recipe.yaml
gforsyth: I'll have to rebuild to find it again, I'll start that going in the background

conda/recipes/libcugraph/recipe.yaml
bdice: Currently we need `librmm` in `run` for every package using it at build time, to act as an ABI constraint. This will become a real runtime constraint with https://github.com/rapidsai/rmm/issues/1779 when RMM becomes a shared library.

conda/recipes/cugraph/recipe.yaml
gforsyth: Here's an example of this file getting included in the package: https://github.com/rapidsai/cugraph/actions/runs/14196224989/job/39773072531?pr=4999#step:9:3094

bdice: Weird. I don't fully understand but I'll roll with it.

bdice (approved):

conda/recipes/cugraph/recipe.yaml
vyasr: https://github.com/prefix-dev/rattler-build/issues/979

PR #5004, opened 2025-03-31
Fix sampling primitive build error.

seunghwak:
This PR fixes a sampling primitive build error when an edge property passed to an input `bias_op` is either thrust::nullopt_t or thrust::tuple.


ChuckHastings (commented):
This looks good.  I'll try testing it in my PR.

PR #5013, opened 2025-04-03
Replace `Thrust` iterator facilities with `libcu++` ones

miscco:
They do the same and we should use the standard ones


ChuckHastings (approved):
Thanks for the updates!

PR #5015, opened 2025-04-03
Do not pretend to default initialize a device lambda

miscco:
Found in CCCL CI


bdice (approved):

cpp/src/prims/count_if_v.cuh
```diff
@@ -35,9 +35,9 @@ namespace detail {
 
 template <typename vertex_t, typename VertexValueInputIterator, typename VertexOp>
 struct count_if_call_v_op_t {
-  vertex_t local_vertex_partition_range_first{};
-  VertexValueInputIterator vertex_value_input_first{};
-  VertexOp v_op{};
+  vertex_t local_vertex_partition_range_first;
+  VertexValueInputIterator vertex_value_input_first;
```
seunghwak, line 39: Can't we default initialize iterators as well? Or is this because of possible thrust::transform_iterator with lambda?

cpp/src/prims/count_if_v.cuh
```diff
@@ -35,9 +35,9 @@ namespace detail {
 
 template <typename vertex_t, typename VertexValueInputIterator, typename VertexOp>
 struct count_if_call_v_op_t {
-  vertex_t local_vertex_partition_range_first{};
-  VertexValueInputIterator vertex_value_input_first{};
-  VertexOp v_op{};
+  vertex_t local_vertex_partition_range_first;
```
seunghwak, line 38: I think we can default initializing vertex_t (which is either int32_t or int64_t) or can this be also problematic?

cpp/src/prims/count_if_v.cuh
miscco: My reasoning with removing all default initializers is that partially initializing the struct does not make any sense and it is only ever initialized in a single place, so it feels "cleaner" to initialize all or nothing

Nothing I feel strongly about though

bdice:
@miscco @seunghwak Checking status here. Can we land this in 25.06?

seunghwak:
@bdice Sorry I just found that I've been neglecting my reviewer duty, just approved.

PR #5025, opened 2025-04-10
Pass dtype objects to cuDF as_column

mroeschke:
With https://github.com/rapidsai/cudf/pull/18331, cuDF now requires values passed to `dtype` in `as_column` to be supported data type objects recognized by cuDF (cuDF's custom types or `np.dtype`).

I only found this usage which was passing `np.int32` instead of `np.dtype(np.int32)` which should fix the failures on the Python build

### DIFF
--- a/python/cugraph/cugraph/structure/hypergraph.py
+++ b/python/cugraph/cugraph/structure/hypergraph.py
@@ -630,7 +630,9 @@ def _str_scalar_to_category(size, val):
         mask=None,
         offset=0,
         null_count=0,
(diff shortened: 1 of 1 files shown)


alexbarghi-nv (approved):
👍

jakirkham:
This is getting puled into PR: https://github.com/rapidsai/cugraph/pull/5023

Reason being we have C++ and Python failures. So to get things passing it is simpler to do in 1 PR

mroeschke:
Thanks, make sense. Closing this PR then.

PR #5026, opened 2025-04-11
Add missing thrust includes

trxcllnt:


jakirkham:
Thanks Paul! 🙏

Allison from the CCCL team caught this earlier today and submitted PR: https://github.com/rapidsai/cugraph/pull/5023

Also Matthew from the cuDF team added a fix that we pulled in too

If you have a moment, could you please review that PR and make sure we are not missing anything?

Once we are all happy with it, we can merge 🙂

PR #5028, opened 2025-04-11
Avoid private cuDF Python APIs

mroeschke:
From the discussion in https://github.com/rapidsai/cugraph/pull/5023, cugraph appears to be using some private cuDF Python APIs (from `cudf.core`). It appears their usage can be wholly replaced with public APIs which will prevent breakage as cuDF should be at liberty to modify private functionality.

This PR will require https://github.com/rapidsai/cudf/pull/18485 to be merged first.


rlratzel (approved):
Thanks, the new code is more clear too.

Issue #5029, opened 2025-04-14
ForceAtlas2 threads returning prematurely?


MathisHammel:
I'm currently working on cuGraph's FA2 implementation to add support for the `prevent_overlapping` parameter (#5022 ), which is documented but not supported.

I found something confusing in the implementation of the associated kernels:
- https://github.com/rapidsai/cugraph/blob/e7c9915c5d29ee1390c511c7170b728d6611fa59/cpp/src/layout/legacy/fa2_kernels.cuh#L50
- https://github.com/rapidsai/cugraph/blob/e7c9915c5d29ee1390c511c7170b728d6611fa59/cpp/src/layout/legacy/exact_repulsion.cuh#L37

Shouldn't these `return` statements be written as `continue` instead? I'm far from a CUDA expert, but it seems like the threads would end prematurely without processing the whole graph.

ChuckHastings:
Good catch.  I imagine the initial development focused on smaller graphs where it's easier to debug.  This error will only trigger when a device thread has more than one edge to process and it encounters an edge in the upper triangle before it encounters edges in the lower triangle.  The odds of this happening increase with the size of the graph.

I noticed you submitted a PR... thanks for the update!

PR #5030, opened 2025-04-14
Fix ForceAtlas2 incorrect loop conditions

MathisHammel:
As noted in #5029, it appears that ForceAtlas2 kernels will sometimes return without computing all forces.

Any of the two following conditions will result in an incorrect computation:
- graphs with more than 8k vertices, if `barnes_hut_optimize` is false (repulsion kernel)
- graphs with more than 16M edges (attraction kernel)

### DIFF
--- a/cpp/src/layout/legacy/exact_repulsion.cuh
+++ b/cpp/src/layout/legacy/exact_repulsion.cuh
@@ -30,11 +30,10 @@ __global__ static void repulsion_kernel(const float* restrict x_pos,
                                         const float scaling_ratio,
                                         const vertex_t n)
 {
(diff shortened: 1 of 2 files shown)


ChuckHastings (approved):
Thanks for the bug fix.

MathisHammel:
@ChuckHastings I think I granted that right already ("Allow edits and access to secrets by maintainers"), but I went ahead and edited the license in both files.

ChuckHastings:
/ok to test https://github.com/rapidsai/cugraph/pull/5030/commits/6d13bb123f13488c0d3a8b513a1d17f03d558ea4

PR #5032, opened 2025-04-15
Replace `Thrust` sequential algorithms with `libcu++` ones

miscco:
This might be a bit controversial because CCCL currently does not officially expose std::algorithms

On the other hand it lets us replace a ton of thrust algorithms that are much heavier than the std::ones, so that might be a win.

There are two I could not replace, `thrust::sequence` and `thrust::copy_if` with a second input range


miscco:
yeah, I will investigate what we can do for the sorting algorithms

PR #5036, opened 2025-04-15
Use public pylibcudf APIs in graph_primtypes

mroeschke:
`graph_primtypes` currently uses private cuDF Python APIs to construct a cuDF Python column from a pointer to an RMM buffer. After https://github.com/rapidsai/cudf/pull/18502 is merged, this can be constructed using public pylibcudf methods.


jameslamb (commented):
Leaving an early comment... since `pylibcudf` is being added to `dependencies.yaml`, it should be added to this list for updating release branches as well: https://github.com/rapidsai/cugraph/blob/9e1211ace9d631bb19bae3cab5bbad554b7560e0/ci/release/update-version.sh#L47

`@` me back when CI is passing and I can review again.

mroeschke:
@jameslamb pinging for another CI review (adding pylibcudf in the correct places as a runtime dependency)

python/cugraph/cugraph/structure/graph_primtypes.pyx
```diff
@@ -16,61 +16,93 @@
 # cython: embedsignature = True
 # cython: language_level = 3
 
-import numpy as np
 from libc.stdint cimport uintptr_t
 from libcpp.utility cimport move
 
 from rmm.pylibrmm.device_buffer cimport DeviceBuffer
-from cudf.core.buffer import as_buffer
+import pylibcudf as plc
```
rlratzel, line 23: This is minor but we've been using "plc" to mean "pylibcugraph" for 3 or more years within cugraph team and code, and I'm thinking this could be confusing. Can this be changed to "plcudf" or just kept as "pylibcudf"?

rlratzel (changes requested):
Hopefully just a minor change request, otherwise LGTM.

python/cugraph/cugraph/structure/graph_primtypes.pyx
mroeschke: Yup definitely, changed. (cuDF is starting to abbreviate pylibcudf as plc as well 😅 )

rlratzel (approved):
Thank you!

python/cugraph/pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
     "libcugraph==25.6.*,>=0.0.0a0",
     "numba>=0.59.1,<0.61.0a0",
     "numpy>=1.23,<3.0a0",
+    "pylibcudf==25.6.*,>=0.0.0a0",
```
jameslamb, line 35: This change should be reflected in the `cugraph` conda recipe as well:

Add `pylibcudf` to both the `host:` and `run:` environments, similar to how `pylibraft` is handled:

https://github.com/rapidsai/cugraph/blob/c3c19e39070638c1bc03320990772cb9d4c60cb2/conda/recipes/cugraph/recipe.yaml#L63

That dependency is getting satisfied transitively because it comes through `cudf`, but now that you're introducing a direct usage of `pylibcudf`, we should make that explicit.

jameslamb (approved):
Left one more suggestion, but approving so you don't need another review from me after it's implemented.

python/cugraph/pyproject.toml
mroeschke: Ah sure thing. Yeah it's easy to forget this since it's not modified by the dependency file generator

jameslamb: Yeah for sure. We're hoping that now that we're using `rattler-build`, it might be possible soon to have `rapids-dependency-file-generator` update these files.

In short... the main challenge was that all the Jinja templating stuff in `conda-build` recipes meant that they weren't valid YAML, which made automatic updates hard. `rattler-build` recipes are guaranteed to be valid YAML files 😁 

@vyasr wrote up a nice summary here: https://github.com/rapidsai/dependency-file-generator/issues/7#issuecomment-2766675431

vyasr:
@mroeschke do we know why tests are failing here?

mroeschke:
> do we know why tests are failing here?

It looks like a tolerance miss e.g.

```python
 =========================== short test summary info ============================
FAILED tests/layout/test_force_atlas2.py::test_force_atlas2[False-500-graph_file4-0.5] - assert np.float64(0.4897435897435898) > 0.5
FAILED tests/layout/test_force_atlas2.py::test_force_atlas2[True-500-graph_file4-0.5] - assert np.float64(0.4897435897435898) > 0.5
```

I don't think it's related to the changes here. I can push an empty commit to rerun the jobs unless you have permissions to rerun jobs in cuGraph

vyasr:
I do, and done. Odd that the failure happened on so many different runners though. Makes me suspicious of the underlying test.

jameslamb:
> Makes me suspicious of the underlying test.

Hi saw this go by in my notifications, and I want to add some evidence in favor of "this is flaky test".

I saw that exist test fail in the exact way for RAPIDS 24.06 DLFW builds. I patched around it there and didn't report it here because I thought it must be something related to that build environment, but seeing it show up on this PR targeting 25.06 suggests it might actually be a problem with the test or the underlying code.

Here's the patch I used to work around this in those builds (`-graph_file4` in the test ID refers to this `dining_prefs` dataset):

```diff
---
 python/cugraph/cugraph/tests/layout/test_force_atlas2.py | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/python/cugraph/cugraph/tests/layout/test_force_atlas2.py b/python/cugraph/cugraph/tests/layout/test_force_atlas2.py
index 1adfa1042..6a6bddf6d 100644
--- a/python/cugraph/cugraph/tests/layout/test_force_atlas2.py
+++ b/python/cugraph/cugraph/tests/layout/test_force_atlas2.py
@@ -108,7 +108,7 @@ DATASETS = [
     (polbooks, 0.75),
     (dolphins, 0.66),
     (netscience, 0.66),
-    (dining_prefs, 0.50),
+    (dining_prefs, 0.40),
 ]
 
 DATASETS2 = [
```

PR #5037, opened 2025-04-16
Avoid unnecessary libcugraph wheel builds

VenkateshJaya:
Adding a matrix filter to the `build.yaml` to fix the issue with artifact name collisions in `libcugraph` builds

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -71,7 +71,7 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.06
     with:
-      # build for every combination of arch and CUDA version, but only for the latest Python 
+      # build for every combination of arch and CUDA version, but only for the latest Python version
       matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -71,6 +71,7 @@ jobs:
     secrets: inherit
     uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.06
     with:
+      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
```
jameslamb, line 74:
```suggestion
      # build for every combination of arch and CUDA version, but only for the latest Python 
      matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
```

Can you please also include a comment with the plain-language description of what this is trying to accomplish? I've found that really helpful for these filters.

We can use the same one from the corresponding `pr.yaml` line:

https://github.com/rapidsai/cugraph/blob/9e1211ace9d631bb19bae3cab5bbad554b7560e0/.github/workflows/pr.yaml#L158-L159

jameslamb (approved):
Looks good! I hope you don't mind, I've changed the title to "Avoid unnecessary libcugraph wheels builds". I think that'll be a bit more informative in a changelog than the original title.

PR #5039, opened 2025-04-20
Added  #include <cuda/std/cmath> for correct cuda::std::log reference

SagaciousFish:
1. Added  #include <cuda/std/cmath> for correct cuda::std::log reference in sample_and_compute_local_nbr_indices.cuh
Without proper include, the following error would be raised during building libcugraph in my environment:

```
cpp/src/prims/detail/sample_and_compute_local_nbr_indices.cuh([[1994|2006|2260|2272]]):
error: namespace "cuda::std" has no member "log"
	return cuda::std::min(-cuda::std::log(r) / b, std::numeric_limits<bias_t>::max());
```
2. Update od_shortest_distances_impl.cuh, fix missing required parameter: invalid_distance in struct e_op_t
Without proper include, the following error would be raised during building libcugraph in my environment:

```
cpp/src/traversal/od_shortest_distances_impl.cuh(143):
	Build log is lost, but the line 143 would complain on missing variable invalid_distance.
```

> Apologies for missing the labels earlier, and now I'm not able to edit them myself. Could someone with the appropriate permissions kindly add the 'bug' and 'non-breaking' labels? Thanks for your help!

### DIFF
--- a/cpp/src/prims/detail/sample_and_compute_local_nbr_indices.cuh
+++ b/cpp/src/prims/detail/sample_and_compute_local_nbr_indices.cuh
@@ -37,6 +37,7 @@
 #include <cub/cub.cuh>
 #include <cuda/atomic>
 #include <cuda/functional>
+#include <cuda/std/cmath>
 #include <cuda/std/iterator>
 #include <cuda/std/optional>
 #include <thrust/adjacent_difference.h>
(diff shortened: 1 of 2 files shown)


ChuckHastings:
/ok to test https://github.com/rapidsai/cugraph/pull/5039/commits/0790e6304ea90f209336e10fef5507adc933df4b

ChuckHastings:
Thanks for the bug fix!

PR #5046, opened 2025-04-23
Fix failing test with new cccl

ChuckHastings:
Newest CCCL is causing a test to fail.

This fix adds a sync stream call which seems to address the issue.

### DIFF
--- a/cpp/tests/sampling/sampling_heterogeneous_post_processing_test.cpp
+++ b/cpp/tests/sampling/sampling_heterogeneous_post_processing_test.cpp
@@ -497,6 +497,9 @@ class Tests_SamplingHeterogeneousPostProcessing
 
         // Check the invariants in vertex renumber_map
 
+        // FIXME:  Adding this in https://github.com/rapidsai/cugraph/pull/5046 to
+        //         work around a test failure.  Need to investigate the root cause
+        //         of the failure rather than mask it.
         handle.sync_stream();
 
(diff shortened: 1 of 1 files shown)


cpp/tests/sampling/sampling_heterogeneous_post_processing_test.cpp
```diff
@@ -497,6 +497,8 @@ class Tests_SamplingHeterogeneousPostProcessing
 
         // Check the invariants in vertex renumber_map
 
+        handle.sync_stream();
```
seunghwak, line 500: We should add a FIXME statement saying something like this shouldn't be necessary but this fixes a segment fault that happens only in specific configurations and this needs further investigation if we want to quickly merge this PR.

seunghwak (approved):

cpp/tests/sampling/sampling_heterogeneous_post_processing_test.cpp
seunghwak: This needs more investigation (we should better find the root cause of any bugs we encounter when we can reproduce those)... If this manifests later in a much more difficult/expensive to reproduce situations, it can be a nightmare...

bdice (approved):
Thanks for the fix @ChuckHastings. I agree with @seunghwak that a comment could be worthwhile. If we can merge this and then merge branch-25.06 into the CCCL testing PR to see it pass, that would be ideal.

PR #5049, opened 2025-04-25
[BUGFIX] Fixed quoting in wheel paths in cugraph and pylibcugraph tests

VenkateshJaya:
Fixes the `libcugraph*.whl is not a valid wheel filename.` bug in the wheel test steps of `pylibcugraph` and `cugraph`on Github CI

### DIFF
--- a/ci/test_wheel_cugraph.sh
+++ b/ci/test_wheel_cugraph.sh
@@ -13,7 +13,7 @@ PYLIBCUGRAPH_WHEELHOUSE=$(RAPIDS_PY_WHEEL_NAME="pylibcugraph_${RAPIDS_PY_CUDA_SU
 # echo to expand wildcard before adding `[extra]` requires for pip
 rapids-pip-retry install \
     "$(echo "${CUGRAPH_WHEELHOUSE}"/cugraph*.whl)[test]" \
-    "${PYLIBCUGRAPH_WHEELHOUSE}/pylibcugraph*.whl" \
-    "${LIBCUGRAPH_WHEELHOUSE}/libcugraph*.whl"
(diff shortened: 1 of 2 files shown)


VenkateshJaya:
@jameslamb I am trying to investigate how this went through CI as well.

Here's the logs for the job that is failing currently, on the last commit I made on my PR was merged. 
https://github.com/rapidsai/cugraph/actions/runs/14583909159/job/40907050641

There's this line in the logs of the pylibcugraph test job `rapids-pip-retry install '/tmp/tmp.3wp5kNKwKB/pylibcugraph_cu12-25.6.0a47-cp310-cp310-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl[test]' '/tmp/tmp.JZGEHfhjDX/libcugraph*.whl'`

Essentially, the path expansion error occurred there as well, however it just went through CI and installed the packages. I am not exactly sure how that happened.

jameslamb:
Ok thanks for looking into that! It's surprising to not even see a *warning* on that PR.

Try this locally:

```shell
pip install '/tmp/nonsense-*.whl'
```

I see a warning and an error

```text
WARNING: Requirement '/tmp/nonsense-*.whl' looks like a filename, but the file does not exist
ERROR: nonsense-*.whl is not a valid wheel filename.
```

And a non-0 exit code

```shell
echo $?
# 1
```

Maybe there is something wrong with how we're passing arguments through `rapids-pip-retry`: https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-pip-retry

I don't know :/

This does not need to block this PR, to be clear... merge this whenever you want.

bdice (approved):
This has a cyclic dependency with https://github.com/rapidsai/cugraph/pull/5047. I will ask for admin-merges.

PR #5051, opened 2025-04-28
Fix MSBFS Test Failure

jnke2016:
The current MSBFS test creates a graph from multiple disjoint edgelist from the rmat generator where one seed from each disjoint graph is selected. Based on the current test configuration, it is possible that all seeds selected may have no outgoing edge hence failing the test. 
This PR

- Properly handles the case where all seeds are isolated vertices
- Refactor the test to follow our current test structure


cpp/tests/traversal/ms_bfs_test.cu
```diff
@@ -267,7 +267,36 @@ class Tests_MsBfs : public ::testing::TestWithParam<MsBfs_Usecase> {
...
+
+    auto seeds_degree_size = thrust::distance(thrust::make_zip_iterator(d_vertex_degree.begin(), d_vertices.begin()), seeds_degree_last);
+    
+    d_vertex_degree.resize(seeds_degree_size, handle.get_stream());
+
+    auto seeds_degree_sum = thrust::reduce(rmm::exec_policy(handle.get_stream()),
+                                           d_vertex_degree.begin(),
+                                           d_vertex_degree.end(),
+                                           static_cast<vertex_t>(0),
+                                           thrust::plus<vertex_t>());
```
seunghwak, line 295: Are you just summing the degree of all the sources?
```
auto seed_degree_first = thrust::make_transform_iterator(d_sources.begin(), cuda::proclaim_return_type<edge_t>([degrees = raft::devie_span<edge_t const>(d_vertex_degree.data(), d_vertex_degree.size())]__device__(auto v) { return degrees[v]; }));
auto seed_degree_sum = thrust::reduce(handle.get_thrust_policy(), seed_degree_first, seed_degree_first + d_sources.size());
```
will be sufficient.

seunghwak: And if you have time, could you update this test to use a same mechanism as other C++ tests?

We don't use
```
    hr_timer.start("msbfs");
    cudaProfilerStart();
``` (performance measurement should be enabled with --perf command line option).

We check with edge masking and we take check_correctness flag. We have rmat_small tests for correctness check and rmat benchmark tests without correctness check and so on.

This test is written in its own way... making our codebase less consistent and harder to maintain.

Comment (jnke2016) [reply cpp/tests/traversal/ms_bfs_test.cu:295]: Got it. Let me refactor this

Comment (jnke2016) [reply cpp/tests/traversal/ms_bfs_test.cu:295]: > Are you just summing the degree of all the sources?
Yep. Ok I will use `thrust::make_transform_iterator` as you suggested

Comment (jnke2016) [reply cpp/tests/traversal/ms_bfs_test.cu:295]: Actually I am summing the degree of all seeds, not sources

Comment (jnke2016) [reply cpp/tests/traversal/ms_bfs_test.cu:295]: > And if you have time, could you update this test to use a same mechanism as other C++ tests?

Wouldn't this break the current structure of the test? In fact, this test relies on creating multiple disjoint graphs while selecting one seed per graph which is possible by generating multiple edgelist. I am not sure how we can adapt this with `file_test` and `rmat_small_test` in a way that matches the other C++ tests. Do I just select random seeds from one edgelist?

Comment (ChuckHastings) [reply cpp/tests/traversal/ms_bfs_test.cu:295]: You could mimic the structure of the other tests but only use RMAT.  Perhaps a comment as to why the file tests are omitted.

Within the test you could still use the same "create a series of RMAT graphs" approach.  There are several tests that don't just create a graph but create the edge list, do some things and then create the graph from the resulting edge list.

But if the overall structure was more consistent that would be beneficial for maintenance.

Comment (jnke2016) [reply cpp/tests/traversal/ms_bfs_test.cu]: > Perhaps a comment as to why the file tests are omitted.

With the approach you suggested which relies on finding the connected components of a graph and selecting the MsBFS seeds from the representative of each component (min vertex IDs), we no longer need to omit the file tests

File: cpp/tests/traversal/ms_bfs_test.cpp [hunk 2]
```diff
@@ -0,0 +1,296 @@
...
+      cugraph::test::sort_by_key<vertex_t, vertex_t>(handle, d_labels, d_vertices);
+
+    rmm::device_uvector<vertex_t> d_tmp_sorted_labels(d_sorted_labels.size(), handle.get_stream());
+
+    raft::copy(d_tmp_sorted_labels.data(),
+               d_sorted_labels.data(),
+               d_sorted_labels.size(),
+               handle.get_stream());
+
+    auto num_unique_labels = cugraph::test::unique_count<vertex_t>(handle, d_tmp_sorted_labels);
```
ChuckHastings, line 118: Probably should add a check to make sure num_unique_labels > 1.  Otherwise it's not really multi-seed.

cpp/tests/traversal/ms_bfs_test.cpp
```diff
@@ -0,0 +1,296 @@
...
+    override_Rmat_Usecase_with_cmd_line_arguments(GetParam()));
+}
+
+INSTANTIATE_TEST_SUITE_P(
+  simple_test,
+  Tests_MsBfs_File,
+  ::testing::Combine(
+    // enable correctness checks
+    ::testing::Values(MsBfs_Usecase{2, 5, false, true}, MsBfs_Usecase{4, 9, true, true}),
+    ::testing::Values(cugraph::test::File_Usecase("test/datasets/netscience.mtx"),
```
ChuckHastings, line 274: Does netscience have multiple components?

jnke2016: Yes it has 396 components

cpp/tests/traversal/ms_bfs_test.cpp
jnke2016: Right, I added a check

cpp/tests/traversal/ms_bfs_test.cpp
```diff
@@ -0,0 +1,298 @@
...
+    auto d_vertices = cugraph::test::sequence<vertex_t>(
+      handle, graph_view.number_of_vertices(), size_t{1}, vertex_t{0});
+
+    rmm::device_uvector<vertex_t> d_sorted_vertices{0, handle.get_stream()};
+    rmm::device_uvector<vertex_t> d_sorted_labels{0, handle.get_stream()};
+
+    std::tie(d_sorted_labels, d_sorted_vertices) =
+      cugraph::test::sort_by_key<vertex_t, vertex_t>(handle, d_labels, d_vertices);
+
+    rmm::device_uvector<vertex_t> d_tmp_sorted_labels(d_sorted_labels.size(), handle.get_stream());
```
seunghwak, line 111: Why did you create d_tmp_sorted_labels? Can we just use d_sorted_labels? We are just making a copy...

cpp/tests/utilities/thrust_wrapper.cu
```diff
@@ -374,6 +375,151 @@ template rmm::device_uvector<int32_t> unique<int32_t>(raft::handle_t const& hand
...
+                                      rmm::device_uvector<int32_t> const& values);
+
+template size_t unique_count<int64_t>(raft::handle_t const& handle,
+                                      rmm::device_uvector<int64_t> const& values);
+
+template <typename key_t, typename value_t>
+std::tuple<cugraph::dataframe_buffer_type_t<key_t>, cugraph::dataframe_buffer_type_t<value_t>>
+reduce_by_key(raft::handle_t const& handle,
+              cugraph::dataframe_buffer_type_t<key_t>&& keys,
+              cugraph::dataframe_buffer_type_t<value_t>&& values,
```
seunghwak, line 397: Not sure these should be R-values. thrust::reduce_by_key does not destroy the inputs, I think this can just be const& similar to unique_count.

cpp/tests/utilities/thrust_wrapper.hpp
```diff
@@ -63,10 +63,41 @@ sort_by_key(raft::handle_t const& handle,
...
+cugraph::dataframe_buffer_type_t<value_t> replace(
+  raft::handle_t const& handle,
+  cugraph::dataframe_buffer_type_t<value_t>&& values,
+  value_t old_value,
+  value_t new_value);
+
+template <typename value_t>
+value_t reduce(raft::handle_t const& handle,
+               cugraph::dataframe_buffer_type_t<value_t>&& values,
+               value_t init_value);
```
seunghwak, line 83: Similar here, values here can be const& (no need for this to be an R-value).

cpp/tests/traversal/ms_bfs_test.cpp
```diff
@@ -0,0 +1,298 @@
...
+#include <limits>
+#include <numeric>
+#include <random>
+#include <set>
+#include <vector>
+
+struct MsBfs_Usecase {
+  size_t radius;
+  size_t max_seeds;
+  bool test_weighted_{false};
```
seunghwak, line 48: Not a must for this PR, but if you have time, you can add edge_masking option as well (similar to https://github.com/rapidsai/cugraph/blob/branch-25.06/cpp/tests/traversal/bfs_test.cpp#L86).

cpp/tests/utilities/thrust_wrapper.cu
jnke2016: Right `thrust::reduce_by_key` stores the result in different arrays

cpp/tests/traversal/ms_bfs_test.cpp
jnke2016: Ok I did though the legacy implementation of `weakly connected components` does not support `edge masking` hence I set it to `false` for all tests

seunghwak (approved):
LGTM

cpp/tests/traversal/ms_bfs_test.cpp
jnke2016: > Have you tested this with edge masking and found that it is not supported

Yep I get an `unimplemented` error
```
unknown file: Failure
C++ exception with description "cuGraph failure at file=/home/nfs/jnke/fix_msbfs/cugraph/cpp/src/components/weakly_connected_components_impl.cuh line=1009: unimplemented.
Obtained 13 stack frames
#1 in /home/nfs/jnke/miniconda3/envs/msbfs/lib/libcugraph.so: cugraph::logic_error::logic_error(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) +0x5a [0x7ff385776a3a]
#2 in /home/nfs/jnke/miniconda3/envs/msbfs/lib/libcugraph.so: void cugraph::weakly_connected_components<int, int, false>(raft::handle_t const&, cugraph::graph_view_t<int, int, false, false, void> const&, int*, bool) +0x101 [0x7ff38948a051]
```

jnke2016: There is an explicit check for that in WCC:`CUGRAPH_EXPECTS(!graph_view.has_edge_mask(), "unimplemented.");`

jnke2016: I removed it and the test passed.

ChuckHastings (approved):

PR #5053, opened 2025-04-30
Simplify edge_property_t

ChuckHastings:
Simplify edge_property_t and edge_*_property_t to depend directly on edge_t instead of graph_t<...>.

There's no real reason to have the graph view be the templated type, we only need it as a template on the constructor, nothing is saved in the object other than `edge_t` values.

This will make it easier to accelerate some compilation issues.


seunghwak (approved):
LGTM, we should eventually remove store_transposed in edge_src|dst_property_t's template parameter list as well.

PR #5063, opened 2025-05-08
Build and test with CUDA 12.9.0

bdice:
This PR uses CUDA 12.9.0 to build and test.

xref: https://github.com/rapidsai/build-planning/issues/173

PR #5064, opened 2025-05-12
CUDA 12.9 used updated compression flags

robertmaynard:
Set more fatbin compression flags to maintain existing binary sizes


jakirkham:
Thanks Rob! 🙏

Think we want to rebase this change on PR: https://github.com/rapidsai/cugraph/pull/5063

bdice (approved):

PR #5068, opened 2025-05-12
Pin cython to avoid version 3.1.0 because of bug (cython issue 6841)

rlratzel:
@TomAugspurger pointed out that recent CI failures are likely due to a change in Cython documented [here](https://github.com/cython/cython/issues/6841), which cause these breakages [here](https://github.com/rapidsai/cugraph/actions/runs/14923270406/job/41922482100)

This PR pins Cython to a version >3.0.0 but not 3.1.0 to avoid that version.


bdice (approved):
If failures persist, be aware that [alpha/beta/rc builds of Cython do exist](https://pypi.org/project/Cython/#history). You may want to use `<3.1.0a0` to prevent those from being found.

alexbarghi-nv:
Is the conda recipe getting updated properly here?

bdice:
> Is the conda recipe getting updated properly here?

Yes, that is the problem. I noticed that earlier as well and just pushed a fix.

rlratzel:
> > Is the conda recipe getting updated properly here?
> 
> Yes, that is the problem. I noticed that earlier as well and just pushed a fix.

Sorry, and thanks. I thought the codegen from `dependencies.yaml` covered those too but I guess not.

bdice:
I saw tests were failing with Python 3.13.
- [Python 3.13 test jobs appeared to be pulling in Python 3.12](https://github.com/rapidsai/cugraph/actions/runs/15004937641/job/42162489775?pr=5068#step:9:379)
- cugraph [25.06.00a61 is built](https://github.com/rapidsai/cugraph/actions/runs/15004937641/job/42161932027?pr=5068#step:9:4748), but [25.06.00a57 is tested](https://github.com/rapidsai/cugraph/actions/runs/15004937641/job/42162489775?pr=5068#step:9:192)

It seems like https://github.com/rapidsai/cugraph/pull/5057/ did not correctly update `dependencies.yaml`'s `py_version` section. I fixed that in 3fcbdf8.

PR #5027, opened 2025-04-11
Use pytest-xdist for parallel testing

bdice:
We use `pytest-xdist` in almost every RAPIDS repository but haven't enabled it for cuGraph yet. Test times are long, so we should try to parallelize Python tests.

The [last successful CI run](https://github.com/rapidsai/cugraph/actions/runs/14391805813/job/40361124281#step:9:8233) took a long time:
```
3427 passed, 905 skipped, 51 deselected, 833 warnings in 5767.53s (1:36:07)
```

I successfully ran with `--numprocesses=8 --dist=worksteal` on my local machine:
```
3456 passed, 909 skipped, 1186 warnings in 881.19s (0:14:41)
```
and [CI got similar results](https://github.com/rapidsai/cugraph/actions/runs/14409723669/job/40415873062?pr=5027#step:9:13339):
```
3427 passed, 905 skipped, 887 warnings in 901.83s (0:15:01)
```

This suggests that there isn't too much time lost in the overhead of parallelizing the test jobs -- it's about 6.4x faster overall.

Our lowest-memory CI node has 24 GB of GPU memory. There are some tests that appear to spawn multiple processes (probably dask or MG tests) and the highest overall GPU memory usage I saw was ~18 GB -- still safely below the cap of 24 GB.

The maximum memory usage I saw was ~1830 MB for any given worker. I saw 3 workers at ~1830 MB and 5 workers at ~450-600 MB for most of the test time. The memory usage seemed to be monotonically increasing for each worker through the test run, which makes me think that resources (perhaps large graphs from datasets?) are not being cleaned up after they are used for a test.

PR #5072, opened 2025-05-14
Remove get_test_data before tests run.

bdice:
Follow-up to #5027.


jameslamb (approved):
Seems ok to me, based on the context from https://github.com/rapidsai/cugraph/pull/5027#discussion_r2085608710

I looked through the logs and didn't see any skips or warnings that seemed to be of the form "skipping this test because data it needs isn't present", and hopefully test coverage would catch stuff like that anyway.

Approving.

PR #5074, opened 2025-05-15
Convert non-type template parameter store_transposed in edge_src|dst_property_t to a runtime parameter

### LINKED ISSUE #4916: Convert store_transposed from template parameter to boolean member variable on graph
To reduce our compile time and binary size, we should convert the `store_transposed` template parameter on the graph object to be a boolean member of the graph.  The cost of the if statements to check at runtime are minimal.

seunghwak:
* `edge_src|dst_property_t` takes `vertex_t`, `T`, and `bool store_transposed` while `edge_property_t takes` just `edge_t` and `T`. This PR resolves this inconsistency and removes a non-type template parameter `bool store_transposed` from `edge_src|dst_property_t`.

Closes #4916


cpp/src/components/mis_impl.cuh
```diff
@@ -188,11 +188,12 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
     per_v_transform_reduce_outgoing_e(
       handle,
       graph_view,
-      multi_gpu
-        ? src_rank_cache.view()
-        : detail::edge_major_property_view_t<vertex_t, vertex_t const*>(temporary_ranks.data()),
+      multi_gpu ? src_rank_cache.view()
+                : detail::edge_endpoint_property_view_t<vertex_t, vertex_t const*>(
```
ChuckHastings, line 192: And here

cpp/tests/components/mg_mis_test.cu
```diff
@@ -154,10 +154,11 @@ class Tests_MGMaximalIndependentSet
         *handle_,
         mg_graph_view,
         multi_gpu ? src_inclusion_cache.view()
-                  : cugraph::detail::edge_major_property_view_t<vertex_t, vertex_t const*>(
-                      inclusiong_flags.data()),
+                  : cugraph::detail::edge_endpoint_property_view_t<vertex_t, vertex_t const*>(
```
ChuckHastings, line 157: And here

ChuckHastings (approved):

PR #5085, opened 2025-05-20
Fix pylibcudf dependency

vyasr:
The 25.6 dependency probably stuck in through a forward merge.


gforsyth (approved):
:shipit:

vyasr:
This PR is currently blocked by the fact that [cugraph-service-client has a dependency on cugraph-pyg](https://github.com/rapidsai/cugraph/blob/branch-25.06/python/cugraph-service/client/pyproject.toml#L42), which creates a circular dependency that is causing old cugraph wheels to be pulled into the devcontainer environment where we are building cugraph. @rlratzel is working on fixing that issue in a separate PR, then we can get this one merged next.

PR #5097, opened 2025-05-27
Require Cython>=3.1.2

vyasr:
We are now compatible with this version and we want to start supporting using it for various new features.


gforsyth:
Two things:
1. I think we should edit the title of this PR to also reflect raising the lower bound of Cython
2. RAPIDS-wise, is this considered a breaking change?

bdice:
1. Make the title "Update to cython >=3.1.2"
2. No, not a breaking change. Cython versions are just build-time constraints, and don't really affect downstream users.

vyasr:
Done

Issue #3047, opened 2022-12-06
[FEA]: Enable running tests in parallel using pytest-xdist


vyasr:
### Is this a new feature, an improvement, or a change to existing functionality?

Improvement

### How would you describe the priority of this feature request

High

### Please provide a clear description of problem this feature solves

I would like to be able to run cugraph tests in parallel using pytest-xdist. Currently this works fine with pylibcugraph, but cugraph tests fail to run because different pytest-xdist workers collect different tests. Enabling parallelism would allow cugraph tests to run much faster, improving the developer experience and reducing CI workloads. Even if only a subset of tests can be run in parallel (I assume that there are is least some subset of tests that uses enough memory/GPU compute that those tests would not be parallelizable), that could still significantly improve test times.

### Describe your ideal solution

I am not certain of the exact cause of the problem at present, but it is definitely due to the usage of the `Dataset` class as a parameter. Normally this pytest-xdist issue is the result of nondeterminism in the ordering of parametrization, but in the case of cugraph it seems like the different instances of this class on different workers are thought to be different. This results in an error since pytest-xdist validates tests on different workers being identical before running anything. I attempted to implement `Dataset.__eq__` thinking that pytest-xdist might be comparing the values in that way, but that does not appear to be sufficient. If the current structure is particularly desirable for some reason it may be worth investigating further, but otherwise the easiest solution that I see is to delay construction of the Dataset objects until the parametrized fixtures are evaluated. This results in a consistent set of tests collected and enables the usage of pytest-xdist.

### Additional context

I've created a POC in #3046 to show what the changes would look like for a single file.

### Code of Conduct

- [X] I agree to follow cuGraph's Code of Conduct
- [X] I have searched the [open feature requests](https://github.com/rapidsai/cugraph/issues?q=is%3Aopen+is%3Aissue+label%3A%22feature+request%22%2Cimprovement%2Cenhancement) and have found no duplicates for this feature request

rlratzel:
This should be closed now that #5102 is merged.
We'll work on additional improvements to Python tests to remove redundancy and possibly improve MG efficiency in future PRs.

PR #5102, opened 2025-05-29
Use `pytest-xdist` on Python Tests

nv-rliu:
Continuing off from https://github.com/rapidsai/cugraph/pull/5027. Addresses issue https://github.com/rapidsai/cugraph/issues/3047

> We use pytest-xdist in almost every RAPIDS repository but haven't enabled it for cuGraph yet. Test times are long, so we should try to parallelize Python tests.

### The issue

It appears that a handful of tests may fail when tests are distributed among workers because `test_dataset.py` will override a global variable that determines where datasets are downloaded to. If this gets overwritten and then another test tries to access the csv file, it will complain that the file does not exist.

### Proposed Solution

Run the rest of the SG/MG Python tests normally while using `pytest-xdist`, but a separate section in `test_python.sh` to run the datasets tests. This may be a quick and efficient way to speed up the majority of CI tests while not having to refactor too much of the datasets unit test. (perhaps this can be addressed in a later PR if deemed necessary)


python/cugraph/cugraph/dask/common/mg_utils.py
```diff
@@ -56,7 +56,7 @@ def is_single_gpu():
     status, count = runtime.cudaGetDeviceCount()
     if status != runtime.cudaError_t.cudaSuccess:
         raise RuntimeError("Could not get CUDA device count.")
-    return count > 1
```
nv-rliu, line 59: `is_single_gpu()` should return True if there's *only* one GPU... Right? I hope I'm not misinterpreting this, but I was kind of dumbfounded that tests designed for MG systems were skipping until I saw this.

ci/run_cugraph_pytests.sh
```diff
@@ -11,6 +11,5 @@ DASK_DISTRIBUTED__SCHEDULER__WORKER_TTL="${DASK_DISTRIBUTED__SCHEDULER__WORKER_T
 DASK_DISTRIBUTED__COMM__TIMEOUTS__CONNECT="${DASK_DISTRIBUTED__COMM__TIMEOUTS__CONNECT:-1000s}" \
 DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT="${DASK_CUDA_WAIT_WORKERS_MIN_TIMEOUT:-1000s}" \
 pytest --cache-clear --import-mode=append --benchmark-disable \
-  -k "not test_property_graph_mg" \
```
nv-rliu, line 14: I don't think this should be here. Let's keep this script more general and merge this arg with the others in `test_python.sh`

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2628,11 +2628,12 @@ def func():
 
 @pytest.mark.slow
 @pytest.mark.parametrize("N", [1, 3, 10, 30])
+@pytest.mark.benchmark(warmup=True, warmup_iterations=10, max_time=0.005)
```
nv-rliu: Noticed that a `bench_` function was being called during testing due to lack of a marker.

python/cugraph/cugraph/tests/data_store/test_property_graph.py
```diff
@@ -2628,11 +2628,12 @@ def func():
...
 def bench_add_edges_cyber(benchmark, N):
     from cugraph.experimental import PropertyGraph
 
     # Partition the dataframe to add in chunks
-    cyber_df = cyber.get_edgelist()
+    cyber_df = cyber.get_edgelist(download=True)
```
nv-rliu: The `download=True` behavior will check if the file exists first before downloading. This helps prevent `FileNotFoundErrors`

python/cugraph/pyproject.toml
```diff
@@ -103,3 +103,6 @@ select = [
 
 # PyPI limit is 100 MiB, fail CI before we get too close to that
 max_allowed_size_compressed = '75M'
+
+[tool.coverage.run]
+disable_warnings = ["module-not-measured"]
```
nv-rliu, line 108: Suggested to me as the long-term fix for using `pytest>=8.4.0`

nv-rliu:
The SG tests appear to be passing and working with xdist. 

Running cugraph tests (SG with xdist, minus datasets)
```
========= 1764 passed, 776 skipped, 514 warnings in 358.01s (0:05:58) ==========
```

cugraph tests (datasets)
```
========= 145 passed, 47 skipped, 4135 deselected in 172.79s (0:02:52) =========
```

MG tests passed once `dist_sampler` and `uniform_neighbor_sampler` were moved out into a separate run. These cannot be run with xdist in their current state. 

MG
```
========= 1007 passed, 318 skipped, 317 warnings in 376.00s (0:06:16) ==========
```

`dist_sampler_mg` and `uniform_neighbor_sampler` MG
```
========= 124 passed, 8 skipped, 4195 deselected in 1568.24s (0:26:08) =========
```

ci/test_python.sh
```diff
@@ -82,44 +111,60 @@ rapids-logger "pytest cugraph (not mg, with xdist)"
...
+  -m "mg" \
+  -k "not test_dataset and not test_property_graph_mg and not test_dist_sampler_mg and not test_uniform_neighbor_sample_mg" \
   --cov-config=../../.coveragerc \
   --cov=cugraph \
   --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cugraph-coverage.xml" \
   --cov-report=term
 
-rapids-logger "pytest cugraph (mg)"
+  # excludes known failures that will always fail when run in combination
+rapids-logger "pytest cugraph (mg dist_sampler and uns)"
```
nv-rliu, line 112: NOTE: It is guaranteed that running `dist_sampler` and `uniform_neighbor_sample` in combination for MG with `pytest-xdist` will fail.

The code-owner has been notified and this is an item on their agenda.

python/cugraph/cugraph/dask/common/mg_utils.py
rlratzel: I believe you are correct.

python/cugraph/cugraph/tests/data_store/test_property_graph.py
rlratzel: This is probably better as a comment in the code for future maintainers wondering why `download=True` is being set.

ci/test_python.sh
```diff
@@ -82,31 +82,43 @@ rapids-logger "pytest cugraph (not mg, with xdist)"
   --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cugraph-coverage.xml" \
   --cov-report=term
 
-# Some tests fail with pytest-xdist enabled.
-# See https://github.com/rapidsai/cugraph/issues/5048
-rapids-logger "pytest cugraph (not mg, without xdist)"
+# The datasets tests may modify global states and interfere with tests running on other workers
+# these tests are typically only limited by network speed and run fairly quickly, so they can be
+# run separately
+rapids-logger "pytest cugraph (datasets)"
```
rlratzel: This is minor but since many other RAPIDS devs not familiar with our library often read the CI logs, can you change this to make it obvious that "datasets" is a Python module in our library and not just tests with datasets or some other misinterpretation?
```suggestion
rapids-logger "pytest cugraph (datasets APIs)"
```

bdice:
Instead of running dataset tests separately in serial, could we make them parallel safe? Using pytest facilities for temporary directories should ensure that workers have unique paths and won’t overwrite or conflict with each other.

nv-rliu:
> Instead of running dataset tests separately in serial, could we make them parallel safe? Using pytest facilities for temporary directories should ensure that workers have unique paths and won’t overwrite or conflict with each other.

Yep! I was going to have a separate PR shortly where the Datasets tests no longer uses the custom tmpdir fixtures and uses the built-in pytest ones instead

python/cugraph/cugraph/tests/data_store/test_property_graph.py
rlratzel: Can you explain this a bit more?  None of the other benchmark functions have this marker, so are they all running or just this one?
Since our [pytest.ini](https://github.com/rapidsai/cugraph/blob/1ab044a68bc5e4c20218c30f3226a147ec9b7e3b/python/cugraph/pytest.ini#L46-L56) has `bench_` as a test name prefix, we should probably consider moving the benchmarks out of this file to a separate benchmarks file (probably in the `benchmarks` dir) and/or changing the pytest command to use `--benchmark-skip` if we don't want them run at all (but note that we do want to run them as tests if possible to ensure they don't break).  Currently we're using `--pytest-disable` which just runs them like tests (one run, no captured run time) to ensure they stay working.

rlratzel (commented):
Looks good so far.

nv-rliu:
Replying to https://github.com/rapidsai/cugraph/pull/5102#discussion_r2132520548 (can't in thread for some reason)

Sorry this may have been intentionally left in when I was trying something. Removing this now.

rlratzel (approved):

bdice (approved):
Very nice work @nv-rliu. This cuts 42-49% off of the typical test runtime. I did some comparisons to another recent PR. Here is the breakdown:

image

Job | Startup | pylibcugraph | SG | MG | Benchmarks | Service
-- | -- | -- | -- | -- | -- | --
12.9 H100 AMD64 before | 341 | 12 | 739 | 3922 | 6 | 26
12.9 H100 AMD64 after | 340 | 12 | 691 | 1855 | 6 | 26
12.2 L4 AMD64 before | 386 | 13 | 524 | 4329 | 6 | 26
12.2 L4 AMD64 after | 396 | 14 | 486 | 1781 | 6 | 28
12.0 A100 ARM before | 264 | 20 | 844 | 4868 | 6 | 39
12.0 A100 ARM after | 241 | 21 | 754 | 2125 | 6 | 39

Here, the "startup" block is fetching CI images, downloading artifacts, and creating a conda environment for tests. Downloading artifacts is the longest portion of that.

As noted above, the non-xdist MG tests take the bulk of the time.

Two follow-up items we need to track:
- `wheel-tests-cugraph` still takes ~2 hours to run, it needs the same parallel treatment!
- See if we can use more than 8 parallel runners successfully. Maybe try 12 (or 10) and see if CI passes or fails.

PR #5104, opened 2025-05-30
Add edge_multi_index_property_t

seunghwak:
This PR add edge_multi_index_property_t which generates multi-edge index for each edge. One current limitation is that this class object's view() class object can't be concatenated (using view_concat) with other edge_property_t objects' view objects.

We can use (source, destination, multi-edge index) to uniquely identify edges in multi-graphs; (source, destination) pairs are sufficient for non-multi-graphs.

edge_bucket_t & transform_e and other files are updated to reflect these changes.

In addition to these changes, this PR deletes the deprecated function graph_view_t::number_of_edges() (this function is replaced with graph_view_t::compute_number_of_edges()).


cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -359,34 +380,41 @@ void accumulate_edge_results(
...
+                                                             dst_properties.view(),
+                                                             edge_multi_indices.view(),
+                                                             extract_edge_e_op_t<vertex_t>{},
+                                                             extract_edge_pred_op_t<vertex_t>{d},
+                                                             do_expensive_check);
+
+      auto triplet_first = thrust::make_zip_iterator(srcs.begin(), dsts.begin(), indices->begin());
+      thrust::sort(handle.get_thrust_policy(), triplet_first, triplet_first + srcs.size());
+    } else {
+      auto [srcs, dsts] = extract_transform_if_e(handle,
```
ChuckHastings: Shouldn't this be `std::tie(srcs, dsts)` ? 

This code will create a temporary srcs/dsts array and when we exit the else block the resulting arrays will be empty.

seunghwak: Yes, you're correct :-)

ChuckHastings (approved):

PR #5112, opened 2025-06-09
Compilation optimizations - batch 1

ChuckHastings:
Breaking previous PR into smaller chunks.  Should be easier to diagnose issues causing failures of tests in CI.

Batch 1: add arithmetic variant types, first use in groupby and count functions.

2% reduction in size of libcugraph.so

### DIFF
--- a/cpp/include/cugraph/arithmetic_variant_types.hpp
+++ b/cpp/include/cugraph/arithmetic_variant_types.hpp
@@ -1,0 +1,249 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


cpp/include/cugraph/arithmetic_variant_types.hpp
```diff
@@ -0,0 +1,249 @@
...
+  {
+    return sizeof(T);
+  }
+};
+
+inline arithmetic_device_span_t make_arithmetic_device_span(arithmetic_device_uvector_t& v)
+{
+  return variant_type_dispatch(v, [](auto& v) {
+    using T = typename std::remove_reference<decltype(v)>::type::value_type;
+    return static_cast<arithmetic_device_span_t>(raft::device_span<T>(v.data(), v.size()));
```
seunghwak, line 236: Minor thing but isn't `arithmetic_device_span_t(raft::device_span<T>(v.data(), v.size()))` sufficient? And assigning a span to a variant of span might be more intuitive than statically casting a span to a variant of a span.

seunghwak (approved):

cpp/include/cugraph/arithmetic_variant_types.hpp
ChuckHastings: Yes, that would be better.  I'll fix that in batch 2 of the changes.

jnke2016 (approved):

PR #5114, opened 2025-06-09
Refactor `cugraph.datasets` tests to be `xdist` compatible

### LINKED ISSUE #5113: [FEA]:  Use updated pytest fixtures for `cugraph.datasets` unit tests

Improvement

Medium

The current `cugraph.datasets` unit tests may interfere with the usage of `pytest-xdist` when running cugraph tests in CI, due to its usage of TempDirs in its custom written fixtures.

These fixtures are obsolete and can be replaced with pytest's built-in fixtures ([link to its docs](https://docs.pytest.org/en/stable/how-to/tmp_path.html#tmpdir-and-tmpdir-factory))

These changes should allow more SG/MG tests to be run in parallel in CI and also clean-up the unit test code.

The tempdirs are used for testing functionality with the `download=True` argument, as well as explicitly setting the `download_dir` location on the CI runner. It was designed this way back when the module was written, since tempdirs ensure that no existing paths would be used, which could introduce strange behavior.

### Additional context

_No response_

### Code of Conduct

nv-rliu:
Closes #5113 
Part of #3047 

## Proposed Changes

This PR removes the usage of custom fixtures in favor of pytest's built-in `tmp_path` fixtures. It also removes unnecessary parts of the datasets API tests where functions are trying to download datasets.

Along with these changes, the tests are now added back to the `(SG, with xdist)` section of `test_python.sh`.

### DIFF
--- a/python/cugraph/cugraph/tests/utils/test_dataset.py
+++ b/python/cugraph/cugraph/tests/utils/test_dataset.py
@@ -103,13 +103,6 @@ def test_modified_env_var(tmp_path):
     datasets.set_download_dir(None)
 
 
-def test_home_dir():
-    # passing None is the same as resetting download_dir
-    datasets.set_download_dir(None)
-
-    assert datasets.get_download_dir() == (Path.home() / ".cugraph/datasets")
-
-
 def test_set_download_dir(tmp_path):
     datasets.set_download_dir(tmp_path.name)
 


python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -128,45 +85,48 @@ def is_symmetric(dataset):
 # Tests
 
 
-# setting download_dir to None effectively re-initialized the default
-def test_env_var():
-    os.environ["RAPIDS_DATASET_ROOT_DIR"] = "custom_storage_location"
+def test_modified_env_var(tmp_path):
+    # changing env var to tmp_path's location
+    os.environ["RAPIDS_DATASET_ROOT_DIR"] = str(tmp_path)
```
rlratzel, line 90: I'm wondering if it would be better to honor the `RAPIDS_DATASET_ROOT_DIR` env var a user/test env has set. You could save it so it can be restored later:
```suggestion
    orig_rdrd = os.environ.get("RAPIDS_DATASET_ROOT_DIR")
    os.environ["RAPIDS_DATASET_ROOT_DIR"] = str(tmp_path)
```

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -128,45 +85,48 @@ def is_symmetric(dataset):
...
+    # calling this with None will reset to whatever the env var is set to
     datasets.set_download_dir(None)
 
-    expected_path = Path("custom_storage_location").absolute()
-    assert datasets.get_download_dir() == expected_path
+    assert datasets.get_download_dir() == tmp_path
 
-    del os.environ["RAPIDS_DATASET_ROOT_DIR"]
+    # reset to default
+    os.environ["RAPIDS_DATASET_ROOT_DIR"] = str(Path.home()) + "/.cugraph/datasets"
```
rlratzel: This will clobber the env var setting for all tests after this, even those in other test files that may be trying to use `RAPIDS_DATASET_ROOT_DIR` for something in particular (eg. testing with datasets in a custom location).  Using the suggestion above, this could be done instead:
```suggestion
    # restore to the initial value
    if orig_rdrd is None:
        del os.environ["RAPIDS_DATASET_ROOT_DIR"]
    else:
        os.environ["RAPIDS_DATASET_ROOT_DIR"] = orig_rdrd
```

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -128,45 +85,48 @@ def is_symmetric(dataset):
...
+    datasets.set_download_dir(None)
```
rlratzel: Do tests assume the default download dir is being set specifically to `~/.cugraph/datasets`?  If so, could you just call this instead?
```suggestion
    datasets.set_download_dir(str(Path.home()) + "/.cugraph/datasets")
```

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -128,45 +85,48 @@ def is_symmetric(dataset):
...
 
 
 def test_home_dir():
+    # passing None is the same as resetting download_dir
     datasets.set_download_dir(None)
-    expected_path = Path.home() / ".cugraph/datasets"
 
-    assert datasets.get_download_dir() == expected_path
+    assert datasets.get_download_dir() == (Path.home() / ".cugraph/datasets")
```
rlratzel: Won't this be a problem if RAPIDS_DATASET_ROOT_DIR is set by the test env?

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -128,45 +85,48 @@ def is_symmetric(dataset):
...
     assert E is not None
     assert dataset.get_path().is_file()
 
 
-@pytest.mark.skipif(is_single_gpu(), reason="skipping MG testing on Single GPU system")
-@pytest.mark.skip(reason="MG not supported on CI")
+@pytest.mark.mg
 @pytest.mark.parametrize("dataset", ALL_DATASETS)
+@pytest.mark.skip(reason="MG not supported on CI")
+@pytest.mark.skipif(is_single_gpu(), reason="skipping MG testing on Single GPU system")
```
rlratzel: Is there a reason both skip markers are being used?  I see they were here before, but I'd think we would only keep the `skipif`.

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -195,116 +155,126 @@ def test_reader(dataset):
...
     assert M is not None
     assert isinstance(M, dict)
 
 
 @pytest.mark.parametrize("dataset", ALL_DATASETS)
-def test_get_path(dataset):
-    tmpd = TemporaryDirectory()
-    datasets.set_download_dir(tmpd.name)
+def test_get_path(dataset, tmp_path):
+    print(tmp_path)
```
rlratzel: Is this debug print supposed to be left in?

rlratzel (changes requested):
Thanks, Ralph. I have a few questions and suggested changes.

python/cugraph/cugraph/tests/utils/test_dataset.py
nv-rliu: That would be the case _unless_ specified otherwise by `RAPIDS_DATASET_ROOT_DIR`. By calling it with `None`, we let `download_dir` go back to whatever was the default (could be something else that's set by the testing environment).

python/cugraph/cugraph/tests/utils/test_dataset.py
nv-rliu: Seems like L128 can be removed. We skip MG via the markers we specify when running the tests in CI.

python/cugraph/cugraph/tests/utils/test_dataset.py
nv-rliu: Oops

python/cugraph/cugraph/tests/utils/test_dataset.py
nv-rliu: I don't think this test is needed. It doesn't add anything.

rlratzel (approved):
LGTM, thanks Ralph!

gforsyth (approved):

PR #5115, opened 2025-06-10
[BUG] Fix Comms Shutdown Function

alexbarghi-nv:
The comms shutdown function in `cugraph.gnn` deletes the `__nccl_comms` and `__raft_handle` variables, which makes it impossible to later create a new communicator.  This causes issues running the multi-GPU tests in cuGraph-PyG, which were recently fixed to properly setup/teardown communicators.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -555,6 +555,7 @@ add_library(cugraph_c
         src/c_api/renumber_arbitrary_edgelist.cu
         src/c_api/legacy_fa2.cpp
         src/c_api/legacy_mst.cpp
+        src/c_api/extract_vertex_list.cpp
         )
 add_library(cugraph::cugraph_c ALIAS cugraph_c)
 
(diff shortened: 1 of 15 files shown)


python/cugraph/cugraph/gnn/comms/cugraph_nccl_comms.py
```diff
@@ -78,10 +78,11 @@ def cugraph_comms_shutdown():
     __nccl_comms.destroy()
     setDevice(__old_device)
 
-    del __raft_handle
-    del __nccl_comms
     del __old_device
 
+    __nccl_comms = None
+    __raft_handle = None
```
rlratzel: Any reason you wouldn't use the same mechanism here and just assign `__old_device` to `None` for consistency?

alexbarghi-nv: I think that's a reasonable suggestion.  I'll make the change.

PR #5116, opened 2025-06-11
Optionally store the generated edge list in pinned memory buffer

seunghwak:
Breaking as multiple functions take an additional input parameter (i.e. std::optional<large_buffer_type_t> large_buffer_type).


alexbarghi-nv:
@seunghwak how are we thinking of exposing this to the C API?

seunghwak:
> @seunghwak how are we thinking of exposing this to the C API?

This is a part 1 of the two PRs. This PR is about generating edges and the next  PR will be about creating a graph from the edge list.

In your workflow, I assume you will get edge list from elsewhere. And to handle graphs larger than GPU memory size, this edge list should be stored in the pinned memory buffer.

* **Initializing large memory buffer**

In managing a rmm::mr::pinned_memory_resource, we can think of two cases. Python owning this memory resource or C++ owning this memory resource (for rmm::mr::pinned_memory_resource, I guess Python and C++ creating separating ones won't be a problem as they just call cudaHostAlloc or cudaMallocHost but this can be a problem for other memory resources such as a pool memory resource backed by pinned memory, we don't want to maintain two pools).

In case we want C++ to own the pinned memory resource,

we can call `large_buffer_manager::create_memory_buffer_resource` (https://github.com/rapidsai/cugraph/pull/5116/files#diff-5ea2450fd4d69c9ee59c2d62e87c08c3e0fe835df8ce75be547bf181b8925421R94) and pass this to the `large_buffer_manager::init` (https://github.com/rapidsai/cugraph/pull/5116/files#diff-5ea2450fd4d69c9ee59c2d62e87c08c3e0fe835df8ce75be547bf181b8925421R83) function.

In case we want Python to own the pinned memory resource,

we can call the constructor of `large_memory_buffer_resource_t` (https://github.com/rapidsai/cugraph/pull/5116/files#diff-5ea2450fd4d69c9ee59c2d62e87c08c3e0fe835df8ce75be547bf181b8925421R38) with the memory resource passed from Python, and pass this to the `large_memory_buffer::init()` function.

We can call `large_buffer_manager::init()` function in a similar way to we call `partition_manager::init_subcomm` (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/include/cugraph/partition_manager.hpp#L157).

* **Creating a graph object from edge list (edge list may or may not be stored in pinned memory buffer)**

We will need to send flags to indicate whether to use large memory buffers to store per-vertex and per-edge data.

We have these flags for a cugraph C++ test utility function to generate vertex & edge list. These will be added to a C++ function to create a graph (e.g. https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/include/cugraph/graph_functions.hpp#L745). C-API functions need to be updated accordingly to reflect this change.

We have functions like `cugraph_create_graph_sg` (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/include/cugraph_c/graph.h#L76) and these functions need to be updated to take additional flags.

Let me know if you have any additional questions, concerns, or comments.

seunghwak:
And in designing C, pylibcugraph, and python layer APIs, we need to consider that, in the future, we may support storage (e.g. SSD) backed buffer as well in addition to the current (host) memory buffer.

jnke2016 (approved):

PR #5117, opened 2025-06-11
Compile optimizations - batch 2

ChuckHastings:
Second batch of updates to use the new variant structures.  This should reduce the compile time and compiler memory usage.

Another 1% savings from the baseline size of libcugraph.so.

### DIFF
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -215,7 +215,7 @@ shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
  * the sparse 2D matrix using sources as major indices) or destinations (otherwise)
  * @param[in,out] edgelist_minors Vertex IDs for destinations (if we are internally storing edges
  * in the sparse 2D matrix using sources as major indices) or sources (otherwise)
- * @param[in,out] edgelist_properties Vector of edge properties, each element a device span of an
+ * @param[in,out] edgelist_properties Span of edge properties, each element a device span of an
  * edge property.
  * @param[in] groupby_and_count_local_partition_by_minor If set to true, groupby and count edges
  * based on (local partition ID, GPU ID) pairs (where GPU IDs are computed by applying the
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -205,25 +205,18 @@ shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
 /**
  * @ingroup shuffle_wrappers_cpp
  * @brief Groupby and count edgelist using the key function which returns the target local partition
- * ID for an edge.
+ * ID for an edge.  The specified spans are reordered in place.
  *
  * @tparam vertex_t Type of vertex identifiers. Needs to be an integral type.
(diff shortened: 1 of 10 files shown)


robertmaynard (approved):

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -205,25 +205,18 @@ shuffle_int_vertex_value_pairs_to_local_gpu_by_vertex_partitioning(
...
- * @param[in,out] d_edgelist_minors Vertex IDs for destinations (if we are internally storing edges
+ * @param[in,out] edgelist_minors Vertex IDs for destinations (if we are internally storing edges
  * in the sparse 2D matrix using sources as major indices) or sources (otherwise)
- * @param[in,out] d_edgelist_weights Optional edge weights
- * @param[in,out] d_edgelist_ids Optional edge ids
- * @param[in,out] d_edgelist_types Optional edge types
- * @param[in,out] d_edgelist_start_times Optional edge start times
- * @param[in,out] d_edgelist_end_times Optional edge end times
+ * @param[in,out] edgelist_properties Vector of edge properties, each element a device span of an
+ * edge property.
```
seunghwak, line 219: I am not sure we can call host_span of edge properties as "Vector of" edge properties.

When I asked to ChatGPT about the meaning of vector in data structures in general, it said

```
In data structures, a vector is a type of dynamic array that:

Stores elements in contiguous memory (like an array),
Automatically resizes when elements are added or removed,
Supports random access to elements in constant time (O(1)).
```

host_span is not a "dynamic" array and does not support resizing. I feel like "Vector" here is a misnomer.

Should we better say "Array" of edge properties or "View" of edge property array or "Span" of edge properties or something like these?

cpp/include/cugraph/detail/shuffle_wrappers.hpp
```diff
@@ -277,4 +270,45 @@ rmm::device_uvector<value_t> collect_local_vertex_values_from_ext_vertex_value_p
...
+ *
+ * @return tuple of device vector of keys and vector of device vector of properties associated with
+ * the key.
+ */
+template <typename key_t, typename key_to_gpu_op_t>
+std::tuple<rmm::device_uvector<key_t>, std::vector<arithmetic_device_uvector_t>>
+shuffle_keys_with_properties(raft::handle_t const& handle,
+                             rmm::device_uvector<key_t>&& keys,
+                             std::vector<arithmetic_device_uvector_t>&& properties,
+                             key_to_gpu_op_t key_to_gpu_op);
```
seunghwak, line 296: We have cugraph/shuffle_functions.hpp and cugraph/detail/shuffle_wrappers.hpp, should we keep both?

I am thinking about moving non-detail space functions to shuffle_functions.hpp and create a non-detail space functions for functions in the detail/shuffle_wrappers.hpp and eventually removing shuffle_wrappers.hpp.

What do you think?

ChuckHastings: I was thinking along similar lines.  I was also thinking of moving the detail functions out of the include directory and into the src directory to remove them more completely from the public API.  I have more shuffle changes coming in the next batch of changes... I'll try moving some files around then.

cpp/include/cugraph/detail/shuffle_wrappers.hpp
ChuckHastings: Updating to Span in the next push.

seunghwak (approved):

PR #5123, opened 2025-06-17
Allow Graph.from_numpy_array to accept nodes that are types not supported by cupy

mroeschke:
We're seeing failures in `cudf` exercising `Graph.from_pandas_adjacency` where the `pandas.DataFrame` columns are strings due to https://github.com/rapidsai/cugraph/pull/5121.

In https://github.com/rapidsai/cugraph/pull/5121, there is an assumption that the `nodes` that are passed to `Graph.from_numpy_array` are cupy convertible, but cupy does not support certain types that numpy supports (e.g. strings, datetimes, etc.)

### DIFF
--- a/cpp/include/cugraph/detail/shuffle_wrappers.hpp
+++ b/cpp/include/cugraph/detail/shuffle_wrappers.hpp
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <cugraph/arithmetic_variant_types.hpp>
+#include <cugraph/large_buffer_manager.hpp>
 
 #include <raft/core/handle.hpp>
 #include <raft/core/host_span.hpp>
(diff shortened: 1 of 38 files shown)


rlratzel:
LGTM, thanks for fixing this and thanks for the test!

jnke2016:
Thanks for the PR. I didn't realize cupy doesn't support strings and datetimes like `cudf` and `numpy`

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -521,7 +521,13 @@ def from_numpy_array(self, np_array, nodes=None):
         """
 
         if nodes is not None:
-            nodes = cupy.array(nodes)
+            try:
+                nodes = cupy.array(nodes)
```
jnke2016: This is a very minor suggestion but in this case we can just have
```
if nodes is not None:
       nodes = cudf.Series(np.array(nodes))
```

python/cugraph/cugraph/structure/graph_classes.py
```diff
@@ -537,7 +543,7 @@ def from_numpy_array(self, np_array, nodes=None):
             df["src"] = src
             df["dst"] = dst
         df["weight"] = weight
-        self.from_cudf_edgelist(df, "src", "dst", edge_attr="weight", vertices=nodes)
+        self.from_cudf_edgelist(df, "src", "dst", edge_attr="weight", vertices=s_nodes)
```
jnke2016: and here
`self.from_cudf_edgelist(df, "src", "dst", edge_attr="weight", vertices=nodes)`

PR #5129, opened 2025-06-25
Remove nvidia and dask channels

vyasr:
Now that we have dropped support for CUDA 11 we no longer require the nvidia channel.
With the changes in https://github.com/rapidsai/rapids-dask-dependency/pull/85, RAPIDS now only uses released versions of dask, so we no longer need the dask channel either.
Contributes to https://github.com/rapidsai/build-planning/issues/184

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,7 +61,7 @@ repos:
       - id: sphinx-lint
         args: ["--enable=all", "--disable=line-too-long"]
   - repo: https://github.com/rapidsai/dependency-file-generator
-    rev: v1.17.0
+    rev: v1.19.0
     hooks:
         - id: rapids-dependency-file-generator
           args: ["--clean"]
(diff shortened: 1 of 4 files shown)


jakirkham:
Thanks Vyas! 🙏

Do we want to update this doc as well?

https://github.com/rapidsai/cugraph/blob/d49fb012d1fcd6c5406e7330cee2e5e362b95580/python/cugraph-service/README.md?plain=1#L13

Also see `-c nvidia` in this `Dockerfile`. Though don't have a good handle on how this is being used. So think it would be ok to punt and raise an issue for follow up

https://github.com/rapidsai/cugraph/blob/d49fb012d1fcd6c5406e7330cee2e5e362b95580/benchmarks/shared/build_cugraph_ucx/cugraph_ucx.dockerfile#L49-L55

vyasr:
Yeah let's save that for a follow-up where we do docs more thoroughly. I did it in the cuxfilter PR but in hindsight it would be better to keep the PRs a bit more focused so that we can do similar updates to all repos across RAPIDS simultaneously in sets of PRs.

jakirkham (approved):

PR #5132, opened 2025-06-26
Compile Optimizations - batch 4

ChuckHastings:
Next round of compile time optimizations.

Focus is on shuffle_and_organize_output to make it a non-templated function.

Also fixed a C API test that started failing.

About a 2% reduction in library size from this PR.


cpp/src/detail/graph_partition_utils.cuh
```diff
@@ -322,5 +322,17 @@ struct compute_local_edge_partition_minor_range_vertex_partition_id_t {
...
+template <typename label_t>
+struct shuffle_to_output_comm_rank_t {
+  raft::device_span<int32_t const> output_rank_;
+
+  template <typename key_t>
+  __device__ int32_t operator()(key_t key) const
+  {
+    return output_rank_[key];
+  }
+};
```
seunghwak: Isn't this indirection?

We already have a functor for this.
https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/include/cugraph/utilities/device_functors.cuh#L70

cpp/src/sampling/detail/shuffle_and_organize_output.cu
```diff
@@ -0,0 +1,138 @@
...
+
+namespace cugraph {
+namespace detail {
+
+std::tuple<std::vector<cugraph::arithmetic_device_uvector_t>,
+           std::optional<rmm::device_uvector<int32_t>>,
+           std::optional<rmm::device_uvector<size_t>>>
+shuffle_and_organize_output(
+  raft::handle_t const& handle,
+  std::vector<cugraph::arithmetic_device_uvector_t>&& edges_with_properties,
```
seunghwak: Just my 2 cents, `property_edges` to be little more succinct?

cpp/src/sampling/detail/sampling_utils.hpp
```diff
@@ -272,28 +273,31 @@ shuffle_sampling_results(raft::handle_t const& handle,
...
+ *
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param edges_with_properties vector of arithmetic device vectors of the edge data and the
+ * properties.  This should include the src and dst vertices, any edge properties that exist for the
+ * sampled edge and optionally the hop where the edge was sampled
+ * @param hop_index Optional value, if the hop was specified then this should specify with index
+ * that is located in the vector
+ * @param labels Optional labels associated with each edge.  If labels are not specified this
+ * function is a noop.
```
seunghwak: Why are we using hop_index for optional edge hop values, but using std::optional<rmm::device_uvector<int32_t>> for optional edge label values?

cpp/src/detail/graph_partition_utils.cuh
jnke2016: I believe I saw the same function when working on the neighborhood sampling API

ChuckHastings: Thanks.  I couldn't find what I was looking for.  Deleted this and switched to indirection_t.

cpp/src/sampling/detail/sampling_utils.hpp
ChuckHastings: Exposing a subtlety of implementation.  I see the concern, it's probably best not to expose implementation.  Let me describe the reason, some options and we can decide which we prefer.

**Reason it is currently this way**
The label is the key for shuffling.  `groupby_gpu_id_and_shuffle_kv_pairs` requires the key to be a separate iterator from the values iterator.  The tools I built upon use `groupby_gpu_id_and_shuffle_kv_pairs` to shuffle using the label as the key.  Further, the tools I built upon want all of the non-key fields to be part of a vector of `arithmetic_device_uvector_t`.  The `hop_index` was my workaround to get the functionality.

**Some options**
1. I could pass the label in the vector as well and pass both a `label_index` and a `hop_index`.  I could refactor the implementation to use `groupby_gpu_id_and_shuffle_values` and use the `label_index` as part of the functor to do the computation.
2. I could pass both `labels` and `hops` as optional parameters and capture the rx_counts from the current shuffling to make a separate `shuffle_values` call for hops.

cpp/src/sampling/detail/shuffle_and_organize_output.cu
ChuckHastings: I'll make that change.

cpp/src/sampling/detail/sampling_utils.hpp
ChuckHastings: I'm inclined toward option 1.  It feels, with the new variant approach, like we'll see more examples of passing a vector of variants and wanting to treat one of them as special.  So intuitively it feels like this will be a recurring pattern.  But I have no strong feelings either way.

seunghwak: Either is fine but I am slightly more inclined to option 2.

I am basically just considering the intuitiveness of the API completely ignoring the implementation issue (assuming that this doesn't make implementation too complex or too inefficient).

Label or hop is not an edge property. So, including these in `edges_with_properties` sound like a misnomer.

And before seeing the documentation & implementation, I thought hop_index is the hop number (starting from 0 when we are dealing with the seeds), so can be a bit confusing.

ChuckHastings: Went with option 2.  When I started looking at the implementation and the context for the function it made more sense to hops out of the vector.

jnke2016 (approved):

cpp/src/sampling/detail/sampling_utils.hpp
```diff
@@ -272,28 +273,32 @@ shuffle_sampling_results(raft::handle_t const& handle,
...
+ * @brief Organize sampling results by shuffling to the proper GPU (if necessary as identified by
+ * labels and label_to_output_comm_rank) and sorting the vertices by label and hop
+ *
+ * @param handle RAFT handle object to encapsulate resources (e.g. CUDA stream, communicator, and
+ * handles to various CUDA libraries) to run graph algorithms.
+ * @param property_edges vector of arithmetic device vectors of the edge data and the
+ * properties.  This should include the src and dst vertices, any edge properties that exist for the
+ * sampled edge and optionally the hop where the edge was sampled
+ * @param labels Optional labels associated with each edge.  If labels are not specified this
+ * function is a noop.
```
seunghwak, line 238: What do you mean by "this function is a noop"?

This is not a function, do you mean something like "if labels are specified the result is sorted by label (and hop, if hops.has_value() is also true)"?

cpp/src/sampling/detail/shuffle_and_organize_output.cu
```diff
@@ -0,0 +1,169 @@
...
+namespace cugraph {
+namespace detail {
+
+std::tuple<std::vector<cugraph::arithmetic_device_uvector_t>,
+           std::optional<rmm::device_uvector<int32_t>>,
+           std::optional<rmm::device_uvector<int32_t>>,
+           std::optional<rmm::device_uvector<size_t>>>
+shuffle_and_organize_output(
+  raft::handle_t const& handle,
+  std::vector<cugraph::arithmetic_device_uvector_t>&& sampled_edges,
```
seunghwak: `sampled_edges`=>`property_edges` to match the declaration in the header?

seunghwak (approved):
LGTM (except for minor cosmetic issues).

cpp/src/sampling/detail/sampling_utils.hpp
ChuckHastings: If labels is `std::nullopt` then the function `shuffle_and_organize_output` simply returns its input without modification.

cpp/src/sampling/detail/shuffle_and_organize_output.cu
ChuckHastings: Done

PR #5133, opened 2025-06-26
First set of updates - Centrality and community

BradReesWork:
Starting to remove the deprecated code that accepted NX graphs as input.

Also trying to find the issue that introduces. circular dependency. so changed file order in __init__ files


python/cugraph/cugraph/tests/centrality/test_betweenness_centrality.py
```diff
@@ -522,24 +522,8 @@ def test_betweenness_centrality_nx(graph_file, directed, edgevals):
...
-            abs(cugraph_bc[i][1] - networkx_bc[i][1]) > 0.01
-            and cugraph_bc[i][0] == networkx_bc[i][0]
-        ):
-            err = err + 1
-            print(f"{cugraph_bc[i][1]} and {networkx_bc[i][1]}")
-            print(f"{cugraph_bc[i][0]} and {networkx_bc[i][0]}")
-    print("Mismatches:", err)
-    assert err < (0.01 * len(cugraph_bc))
+    with pytest.raises(AttributeError):
+        cu_bc = cugraph.betweenness_centrality(Gnx)
```
rlratzel: I haven't finished reviewing this PR but this did catch my eye so I'll comment now...
For this and all others with similar changes, we should just remove the entire test since it's no longer testing anything we support.

BradReesWork: That is the plan.   

first pass is to remove the code from the functions and keep the tests. 
Then cleanup the util files and remove nx functions
Then fix all the tests so that it checks for invalid input

BradReesWork: trying to not have one massive PR that touches everything

BradReesWork: but dropping unsupported test is easy

rlratzel (approved):

jnke2016 (approved):

python/cugraph/cugraph/tests/utils/test_dataset.py
```diff
@@ -165,7 +165,7 @@ def test_reader_dask(dask_client, dataset):
 
 @pytest.mark.parametrize("dataset", ALL_DATASETS)
 def test_get_edgelist(dataset):
-    E = dataset.get_edgelist()
+    E = dataset.get_edgelist(download=True)
```
rlratzel, line 168: We can do this for now if it allows CI to pass, but we should look into why this is needed later.  [Here](https://github.com/rapidsai/cugraph/issues/5157)'s the issue for it.

rlratzel (approved):

PR #5146, opened 2025-07-02
Improve betweenness centrality performance

HowardHuang1:
Currently has issues with delta values being updated on mismatched vertex indices. In other words, the delta update calculations look correct but they are assigned to indices that are incorrect. This causes the final betweenness centrality scores to deviate more and more across iterations.


cpp/src/centrality/BC_pseudo.cuh
```diff
@@ -0,0 +1,109 @@
...
+                delta[u] += sigma[u] / sigma[v] * (1 + delta[v]);
+            }
+            if(v != s){
+                CB[v] += delta[v];
+            }
+        }
+    }
+    return CB;
+}
+
```
seunghwak: Is this just a sequential test code for your own testing? In that case, you don't need to include this in the PR?

cpp/CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ message("-- Building for GPU_ARCHS = ${CMAKE_CUDA_ARCHITECTURES}")
 
 list(APPEND CUGRAPH_CUDA_FLAGS --expt-extended-lambda --expt-relaxed-constexpr)
 list(APPEND CUGRAPH_CUDA_FLAGS -Werror=cross-execution-space-call -Wno-deprecated-declarations -DRAFT_HIDE_DEPRECATION_WARNINGS -Xptxas=--disable-warnings)
-list(APPEND CUGRAPH_CUDA_FLAGS -Xcompiler=-Wall,-Wno-error=sign-compare,-Wno-error=unused-but-set-variable)
+list(APPEND CUGRAPH_CUDA_FLAGS -Xcompiler=-Wall,-Wno-error=sign-compare,-Wno-error=unused-but-set-variable,-Wno-unused-but-set-variable)
```
seunghwak: Did you make this update? or you pulled updates from another PR?

cpp/libcugraph_etl/CMakeLists.txt
```diff
@@ -84,7 +84,7 @@ message("-- Building for GPU_ARCHS = ${CMAKE_CUDA_ARCHITECTURES}")
 
 list(APPEND CUGRAPH_ETL_CUDA_FLAGS --expt-extended-lambda --expt-relaxed-constexpr)
 list(APPEND CUGRAPH_ETL_CUDA_FLAGS -Werror=cross-execution-space-call -Wno-deprecated-declarations -Xptxas=--disable-warnings)
-list(APPEND CUGRAPH_ETL_CUDA_FLAGS -Xcompiler=-Wall,-Wno-error=sign-compare,-Wno-error=unused-but-set-variable)
+list(APPEND CUGRAPH_ETL_CUDA_FLAGS -Xcompiler=-Wall,-Wno-error=sign-compare,-Wno-error=unused-but-set-variable,-Wno-unused-but-set-variable)
```
seunghwak: Did you make this update? or you pulled updates from another PR?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,202 @@ void accumulate_vertex_results(
...
+  raft::copy(h_initial_centralities.data(), centralities.data(), centralities.size(), handle.get_stream());
+  handle.sync_stream();
+  for (size_t i = 0; i < std::min(size_t(10), h_initial_centralities.size()); ++i) {
+    std::cout << h_initial_centralities[i];
+    if (i < std::min(size_t(10), h_initial_centralities.size()) - 1) std::cout << ", ";
+  }
+  if (h_initial_centralities.size() > 10) std::cout << " ...";
+  std::cout << std::endl;
+  std::cout << "DEBUG: Starting backward pass..." << std::endl;
+
```
seunghwak, line 329: You don't need to push the debug code to PR.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,202 @@ void accumulate_vertex_results(
...
+    std::cout << std::endl;
+    
+    // Create vertex list of vertices at distance d-1 (previous frontier)
+    // Count vertices at distance d-1 first
+    auto num_vertices_at_distance = count_if_v(
+      handle,
+      graph_view,
+      distances.begin(),
+      [d] __device__(auto, auto distance) { return distance == (d - 1); },
+      do_expensive_check);
```
seunghwak: You may better sort the vertex, distance pairs outside the for loop, then, find the boundaries.

"And here, you start with the vertices at distance d and find the neighbors at distance d - 1."
=> This comment is actually wrong, please ignore.

seunghwak: Look for thrust::sort_by_key (to sort vertex, distance pairs by distance) and thrust::upper_bound (or thrust::lower_bound). Then, you don't need to scan the entire vertex list in every loop. This can make a significant difference in large diameter graphs.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,79 @@ void accumulate_vertex_results(
...
+        thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last()),
+        vertices_at_distance_d_minus_1.begin(),
+        [distances = distances.begin(), d, v_first] __device__(vertex_t vertex) {
+          auto offset = vertex - v_first;
+          return distances[offset] == (d - 1);
+        });
+    }
+    
+    // Count edges that will be processed (where dst at distance d and src at distance d-1)
+    auto num_edges_to_process = count_if_e(
```
seunghwak: Where are you using `num_edges_to_process`?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,79 @@ void accumulate_vertex_results(
...
+            return (sigma_v / sigma_w) * (1 + delta_w);
+          } else {
+            return weight_t{0};
+          }
+        },
+        weight_t{0},
+        reduce_op::plus<weight_t>{},
+        deltas.begin(),
+        do_expensive_check);
+    }
```
seunghwak: See
```
 * @param vertex_value_output_first Iterator pointing to the (tagged-)vertex property variables for
 * the first (inclusive) (tagged-)vertex in @p key_list. `vertex_value_output_last` (exclusive) is
 * deduced as @p vertex_value_output_first + @p key_list.size().
```
https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/src/prims/per_v_transform_reduce_incoming_outgoing_e.cuh#L362

Here, the output array size should be the size of the vertex_list (not the number of local vertices), and the first output value is for the first vertex in the vertex list.

cpp/tests/centrality/betweenness_centrality_test.cpp
```diff
@@ -87,6 +91,11 @@ class Tests_BetweennessCentrality
     auto edge_weight_view =
       edge_weights ? std::make_optional((*edge_weights).view()) : std::nullopt;
 
+    std::cout << "DEBUG: Number of vertices: " << graph_view.number_of_vertices() << std::endl;
+    std::cout << "DEBUG: Number of edges: " << graph_view.compute_number_of_edges(handle) << std::endl;
+    std::cout << "DEBUG: Is symmetric: " << graph_view.is_symmetric() << std::endl;
+    std::cout << "DEBUG: Has edge weights: " << (edge_weight_view.has_value() ? "yes" : "no") << std::endl;
```
seunghwak: Don't push the debug statements.

newyork.graphml
seunghwak, line 1: You don't push the test files you just locally use.

testCuGraph.py
seunghwak, line 1: Same here, you don't push the test files you only use for local testing. You push only the updates that will be merged to the main code branch.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,94 @@ void accumulate_vertex_results(
...
+        vertices_at_d.begin(),
+        [distances = distances.begin(), d, v_first] __device__(vertex_t vertex) {
+          auto offset = vertex - v_first;
+          return distances[offset] == d;
+        });
+      vertices_by_distance.push_back(std::move(vertices_at_d));
+    } else {
+      vertices_by_distance.push_back(rmm::device_uvector<vertex_t>(0, handle.get_stream()));
+    }
+  }
```
seunghwak, line 325: This code doesn't look efficient.

You are still scanning the entire vertex list `diameter` times.

```
rmm::device_uvector<vertex_t> frontier_vertices(...);
std::vector<size_t> frontier_offsets(...);
{
  thrust::copy(handle.get_thrust_policy(),
thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first()), 
thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last()),
frontier_vertices.begin());
rmm::device_uvector<vertex_t> tmp_distances(...);
copy distances;
thrust::stable_sort_by_key(...);  // key = distance, value = vertex IDs.
thrust::lower_bound(...);  // to update frontier offsets.
... // finish updating offsets.
}
```

seunghwak: Need to do something like this.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,94 @@ void accumulate_vertex_results(
...
 
+  // Based on Brandes algorithm, we want to follow back pointers in non-increasing
+  // distance from S to compute delta
+  //
+  for (vertex_t d = diameter; d > 1; --d) {
+    
+    // Clear deltas array for this iteration
+    detail::scalar_fill(handle, deltas.data(), deltas.size(), weight_t{0});
```
seunghwak: Why are you resetting deltas in every iteration?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,39 +277,94 @@ void accumulate_vertex_results(
...
+    
+    // Use pre-computed vertices at distance d-1
+    auto& vertices_at_distance_d_minus_1 = vertices_by_distance[d - 1];
+    
+
+    
+    if (vertices_at_distance_d_minus_1.size() > 0) {
+      // Create key_bucket_t from the vertex list
+      key_bucket_t<vertex_t, void, multi_gpu, true> vertex_list(
+        handle, std::move(vertices_at_distance_d_minus_1));
```
seunghwak: Move this outside the above if statement.

You can call aggregate_size() (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/src/prims/vertex_frontier.cuh#L523) to find the aggregate vertex frontier size (in multi-GPU, in single-GPU, this isn't necessary).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -319,6 +376,7 @@ void accumulate_vertex_results(
       thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()),
       dst_properties.mutable_view());
```
seunghwak: There are versions taking a vertex list (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/src/prims/update_edge_src_dst_property.cuh#L1133). You really don't need to update deltas for the entire range.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -319,6 +376,7 @@ void accumulate_vertex_results(
...
 
+    // Accumulate deltas into centralities
```
seunghwak: You are scanning the entire vertex range in every loop. This is inefficient. You need to update just the values for the frontier.

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: I reset deltas because deltas is the single dependency of the vertex of interest v (one that paths must pass through) and the current vertex s on the frontier. In the next iteration, when we advance the frontier, based on my understanding the new delta should be independent of the old one. If I leave out this delta resetting step, old values carry over and the code fails the correctness checks.

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: Oh I see I think this might be the source of the inefficiency since we're updating all edge properties rather than only the ones at the current frontier.

cpp/src/centrality/betweenness_centrality_impl.cuh
ChuckHastings: Just to log this... as mentioned in our meeting today...  `vertices_by_distance` could be computed inside the function `brandes_bfs` almost for free (the cost of a device vector copy), returned from that function and passed in here.

cpp/src/centrality/betweenness_centrality_impl.cuh
ChuckHastings: There might be something else wrong in this logic, if that's the case.  The original implementation did not have to fill deltas with 0 inside of the loop.

It should be the case that a particular value of `delta[v]` is only modified when vertex `v` is at this distance from the seed.  I'd suggest taking this `scalar_fill` call out and investigate the changes of the `delta` vector between iterations if the results are incorrect.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +280,127 @@ void accumulate_vertex_results(
...
-  //
+  // Pre-allocate reusable buffers to avoid repeated allocations
+  auto max_frontier_size = transform_reduce_v(
+    handle,
+    graph_view,
+    distances.begin(),
+    [] __device__(auto, auto d) { return (d == invalid_distance) ? vertex_t{0} : vertex_t{1}; },
+    vertex_t{0},
+    reduce_op::plus<vertex_t>{},
+    do_expensive_check);
```
seunghwak: This code doesn't look like doing what you intended.

This just counts the total number of vertices with the distance != invalid_distance.

seunghwak: I assume you want to compute frontier size per each d (where d is between 0 and max_diameter), and find the maximum frontier size.

For this, you can copy distance & vertex ID vectors. Sort by distance, then, perform vectorized binary seasrch (e.g. thrust::lower_bound) to find the boundaries for each distance values.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +280,127 @@ void accumulate_vertex_results(
...
+  rmm::device_uvector<weight_t> reusable_delta_buffer(max_frontier_size, handle.get_stream());
+  rmm::device_uvector<weight_t> reusable_centrality_buffer(max_frontier_size, handle.get_stream());
+
   // Based on Brandes algorithm, we want to follow back pointers in non-increasing
   // distance from S to compute delta
   //
   for (vertex_t d = diameter; d > 1; --d) {
-    per_v_transform_reduce_outgoing_e(
+    // Clear deltas array for this iteration
+    detail::scalar_fill(handle, deltas.data(), deltas.size(), weight_t{0});
```
seunghwak: deltas.size() == graph_view.local_vertex_partition_range_size(), right?

You don't need to reset deltas in every iteration. This wastes a lot of computing. You need to find the deltas only for the vertices in the frontier.

cpp/src/centrality/betweenness_centrality_impl.cuh
seunghwak: Again here, you're scanning the entire distances array here (size = graph_view.local_vertex_partition_range_size()). This is not necessary. If we sort the vertex list by distance outside the loop, no need to iterate over the entire vertex in every iteration.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +280,127 @@ void accumulate_vertex_results(
...
+      auto v_first = graph_view.local_vertex_partition_range_first();
+      thrust::copy_if(
+        handle.get_thrust_policy(),
+        thrust::make_counting_iterator(v_first),
+        thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last()),
+        reusable_vertex_buffer.begin(),
+        [distances = distances.begin(), d, v_first] __device__(vertex_t vertex) {
+          auto offset = vertex - v_first;
+          return distances[offset] == (d - 1);
+        });
```
seunghwak: This shouldn't be necessary if we sort the vertices by distances outside the loop.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +280,127 @@ void accumulate_vertex_results(
...
+      // Compute deltas for frontier vertices
+      per_v_transform_reduce_outgoing_e(
+        handle,
+        graph_view,
+        vertex_list,
+        src_properties.view(),
+        dst_properties.view(),
+        cugraph::edge_dummy_property_t{}.view(),
+        [d] __device__(auto, auto, auto src_props, auto dst_props, auto) {
+          if ((thrust::get<0>(dst_props) == d) && (thrust::get<0>(src_props) == (d - 1))) {
```
seunghwak: I assume thrust::get<0>(src_props) == d-1 is guaranteed to be true.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +280,127 @@ void accumulate_vertex_results(
...
+        reusable_centrality_buffer.begin(),
+        thrust::plus<weight_t>()
+      );
+      
+      thrust::scatter(
+        handle.get_thrust_policy(),
+        reusable_centrality_buffer.begin(),
+        reusable_centrality_buffer.begin() + frontier_count,
+        reusable_vertex_buffer.begin(),
+        centralities.begin()
```
seunghwak: You can combine this to a single thrust call.

thrust::for_each(handle.get_thrust_policy(),
thrust::make_zip_iterator(vertex_list.begin(), resuable_delta_buffer.begin()),
thrust::make_zip_iterator(vertex_list.begin(), resuable_delta_buffer.begin()) + vertex_list.size(),
[...]__device__(auto pair) {
  centralities[thrust::get<0>(pair)] += thrust::get<1>(pair);
}

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: Ok so instead of allocating memory for all vertices that are reachable, we can optimize this by just allocating memory equal to the size of the largest frontier

HowardHuang1: Was also thinking about 2 options

1) moving the largest frontier calculation outside of the loop entirely. So I would run 2 BFS to find diameter of the graph and then run a max aggregation to find the largest frontier. This would be a fixed buffer size for the reusable_vertex_buffer which contains frontier vertices in every iteration (every iteration would have same size buffer which might be wasteful in some situations but we would only have to compute diameter once outside loop as opposed to a 2 binary searches every iteration). This would be a tradeoff of memory for speed. 

2) Alternatively, using resizable buffers but not sure how expensive this would be if resizing is done in every iteration.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+    transform_begin,
+    transform_end,
+    search_keys.begin(),
+    search_keys.end(),
+    bounds.data());
+  
+  // Copy bounds to host for use in delta loop
+  std::vector<vertex_t> h_bounds(bounds.size());
+  raft::update_host(h_bounds.data(), bounds.data(), bounds.size(), handle.get_stream());
+  handle.sync_stream();
```
seunghwak: In this code, what we need after this is just `vertices_at_distance` and `h_bounds`.

Other variables will just eat-up memory and makes code harder to analyze.

We may make this as a separate function or the code is not large (complex enough) to justify creating another function, you can do something like the following.

```
rmm::device_uvector<vertex_t> vertices_at_distance(...);
std::vector<vertex_t> h_bounds{};
{
  // place the code to update vertices_at_distance & h_bounds.
}
```

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+  rmm::device_uvector<vertex_t> vertices_at_distance(graph_view.local_vertex_partition_range_size(), handle.get_stream());
+  thrust::copy(
+    handle.get_thrust_policy(),
+    thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first()),
+    thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last()),
+    vertices_at_distance.begin());
+  
+  // Create distance-vertex pairs for sorting
+  rmm::device_uvector<thrust::tuple<vertex_t, vertex_t>> distance_vertex_pairs(
+    graph_view.local_vertex_partition_range_size(), handle.get_stream());
```
seunghwak: We can consider two different options in allocating vectors for distances & vertices.

One is creating separate vectors for distances & vertices (either explicitly creating two rmm::device_uvector objects or calling an utility function allocate_dataframe_buffer<thrust::tuple<vertex_t, vertex_t>(...)).

Second is the approach you're doing. This wasn't allowed in the past (rmm::device_uvector did not allow tuple type elements), and generally not recommended for GPU architectures (there are exceptions but here, it is not the case).

And you really don't need to create two vectors (vertices_at_distance and distance_vertex_pairs) to store vertices.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+    });
+
+  // Sort by distance, then by vertex ID for stable ordering
+  thrust::sort(
+    handle.get_thrust_policy(),
+    distance_vertex_pairs.begin(),
+    distance_vertex_pairs.end());
+
+  // Single vectorized thrust call to compute all bounds for distances 0 to diameter
+  vertex_t max_distance = diameter;
```
seunghwak: Why are you creating an alias of diameter? The readers need to now remember two different variable names and also they are actually same. I am not sure what we are getting from this.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+      auto distance = thrust::get<0>(pair);
+      auto vertex = thrust::get<1>(pair);
+      return thrust::make_tuple(distance, vertex);
+    });
+
+  // Sort by distance, then by vertex ID for stable ordering
+  thrust::sort(
+    handle.get_thrust_policy(),
+    distance_vertex_pairs.begin(),
+    distance_vertex_pairs.end());
```
seunghwak: You can just create another distance array & thrust::copy distances.

Then, call stable_sort_by_key where key is the distance. And call stable_sort to maintain the vertices sorted (if the distances are same).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+  vertex_t max_distance = diameter;
+  
+  rmm::device_uvector<vertex_t> search_keys(max_distance + 1, handle.get_stream());
+  
+  // Use thrust::copy with counting_iterator instead of thrust::sequence
+  thrust::copy(
+    handle.get_thrust_policy(),
+    thrust::make_counting_iterator<vertex_t>(0),
+    thrust::make_counting_iterator<vertex_t>(max_distance + 1),
+    search_keys.begin());
```
seunghwak: This is not necessary. You can just pass thrust::make_counting_iterator to thrust::lower_bound as search keys.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+  
+  rmm::device_uvector<vertex_t> bounds(max_distance + 1, handle.get_stream());
```
seunghwak: If we are dealing with both host and device vectors, naming the device_vector as d_bounds and the host vector as h_bounds is our convention.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+  
+  // Vectorized lower_bound to compute all bounds at once
+  auto vertex_proj = [] __device__(auto pair) { return thrust::get<0>(pair); };
+  auto transform_begin = thrust::make_transform_iterator(distance_vertex_pairs.begin(), vertex_proj);
+  auto transform_end = thrust::make_transform_iterator(distance_vertex_pairs.end(), vertex_proj);
```
seunghwak: This is not necessary if we have a separate vector for distance & vertices.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+  
+  // Calculate max frontier size using the precomputed bounds
+  vertex_t max_frontier_size = 0;
+  for (vertex_t d = 0; d < h_bounds.size() - 1; ++d) {
+    vertex_t frontier_count = h_bounds[d + 1] - h_bounds[d];
+    max_frontier_size = std::max(max_frontier_size, frontier_count);
+  }
+
+  rmm::device_uvector<vertex_t> reusable_vertex_buffer(max_frontier_size, handle.get_stream());
+  rmm::device_uvector<weight_t> reusable_delta_buffer(max_frontier_size, handle.get_stream());
```
seunghwak, line 328: If you are using a rmm pool allocator (this is the default), memory allocation is not that really expensive, and the speedup you can from doing pre-allocation (at the expense of increase in code complexity) is pretty minimal. Just FYI.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
-
-    thrust::transform(handle.get_thrust_policy(),
-                      centralities.begin(),
-                      centralities.end(),
-                      deltas.begin(),
-                      centralities.begin(),
-                      thrust::plus<weight_t>());
+    // Use precomputed bounds for O(1) lookup instead of binary search
+    vertex_t first_d_minus_1 = h_bounds[d - 1];
+    vertex_t first_d = h_bounds[d];
```
seunghwak: I am not sure we should create aliases here. h_bounds[] is concise and easy to understand enough.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+    vertex_t frontier_count = first_d - first_d_minus_1;
+
+    if (frontier_count > 0) {      
+      // Combined operation: clear deltas and extract frontier vertices in one kernel
+      thrust::transform(
+        handle.get_thrust_policy(),
+        distance_vertex_pairs.begin() + first_d_minus_1,
+        distance_vertex_pairs.begin() + first_d,
```
seunghwak: You're just using vertices, no reason to read both vertices and distances.

seunghwak: And really no reason to copy to reusable_vertex_buffer.
```
vertex_list.insert(frontier_vertices.begin() + h_bounds[d - 1], frontier_vertices.begin() + h_bounds[d]);
```
should be sufficient.
To reset the reusable_delta_buffer, calling thrust::fill(...); will be faster.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -275,56 +290,170 @@ void accumulate_vertex_results(
...
+          
+          // Update source properties (major)
+          src_props_first[v] = thrust::make_tuple(distances[v], sigmas[v], deltas[v]);
+          
+          // Update destination properties (minor)
+          dst_props_first[v] = thrust::make_tuple(distances[v], sigmas[v], deltas[v]);
+          
+          // Accumulate centralities
+          centralities[v] += delta;
+        });
```
seunghwak: Don't do this. This code won't work in multi-GPU.

Call thrust::for_each to update just centralities.

To update edge properties, you just need to update deltas for the vertices in vertex_list. No need to read/write distances & sigmas.

Call update_edge_src|dst_property with vertex_list. To do this, you may create a separate edge_src|dst_property objects for distances & sigmas and deltas.

When you need all three, you can use view_concat (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/include/cugraph/edge_src_dst_property.hpp#L829).

seunghwak: And later, for final last level tuning, we can even skip this for SG, but we will revisit this later.

seunghwak (approved):

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -221,8 +233,10 @@ void accumulate_vertex_results(
   bool with_endpoints,
   bool do_expensive_check)
 {
+
```
seunghwak: This empty line is unecessary.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -221,8 +233,10 @@ void accumulate_vertex_results(
   bool with_endpoints,
   bool do_expensive_check)
 {
+  
   constexpr vertex_t invalid_distance = std::numeric_limits<vertex_t>::max();
 
+  RAFT_CUDA_TRY(cudaDeviceSynchronize());
```
seunghwak: I guess you added this for performance measurement, don't forget to delete this before merging this PR.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -274,57 +288,118 @@ void accumulate_vertex_results(
     graph_view,
     thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()),
     dst_properties.mutable_view());
+  
+  // Pre-allocate reusable buffers to avoid repeated allocations (optimized for max frontier size)
+  // Use binary search method to find frontier boundaries more efficiently
+  
+  std::vector<vertex_t> h_bounds{};
+  // Create separate arrays for sorting (preserve original distances)
+  rmm::device_uvector<vertex_t> distance_keys(distances.size(), handle.get_stream());
```
seunghwak: You can move this inside the 

{
...
}

block starting from line 315 (https://github.com/rapidsai/cugraph/pull/5146/files#diff-40a2c8b86c13c482d7a1e63034a033959887f8b82c9d43522832866c2203a2b1R315). You don't need this after computing `h_bounds`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -274,57 +288,118 @@ void accumulate_vertex_results(
     graph_view,
     thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()),
     dst_properties.mutable_view());
+  
+  // Pre-allocate reusable buffers to avoid repeated allocations (optimized for max frontier size)
```
seunghwak: This comment is relevant to line 340 (https://github.com/rapidsai/cugraph/pull/5146/files#diff-40a2c8b86c13c482d7a1e63034a033959887f8b82c9d43522832866c2203a2b1R340). Move it to the relevant place.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -274,57 +288,118 @@ void accumulate_vertex_results(
...
+        handle.get_thrust_policy(),
+        reusable_delta_buffer.begin(),
+        reusable_delta_buffer.begin() + frontier_count,
+        vertices_sorted.begin() + h_bounds[d - 1],
+        deltas.begin()
+      );
+      
+      // Update properties for vertices in vertex_list
+      update_edge_src_property(handle, graph_view, vertex_list.begin(), vertex_list.end(), thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()), src_properties.mutable_view());
+      update_edge_dst_property(handle, graph_view, vertex_list.begin(), vertex_list.end(), thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()), dst_properties.mutable_view());
```
seunghwak: Only deltas have been updated, no need to update all distances, sigmas, and deltas. When you create edge_src|dst_property_t (https://github.com/rapidsai/cugraph/pull/5146/files#diff-40a2c8b86c13c482d7a1e63034a033959887f8b82c9d43522832866c2203a2b1R277), create separate ones for distances & sigmas and deltas.

When you need to pass all three, you can use `view_concat`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -274,57 +288,118 @@ void accumulate_vertex_results(
...
+      
+      // Update centralities - both vertices_sorted and centralities use local vertex IDs
+      thrust::for_each(
+        handle.get_thrust_policy(),
+        thrust::make_zip_iterator(vertices_sorted.begin() + h_bounds[d - 1], reusable_delta_buffer.begin()),
+        thrust::make_zip_iterator(vertices_sorted.begin() + h_bounds[d], reusable_delta_buffer.begin()),
+        [centralities = centralities.data()] __device__(auto pair) {
```
seunghwak: To be multi-GPU ready,

include `v_first = graph_view.local_vertex_partition_range_first()` in the capture list, and compute `v_offset = v - v_first`. Then, use `centralities[v_offset]`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -274,57 +288,118 @@ void accumulate_vertex_results(
...
-      thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()),
-      dst_properties.mutable_view());
-
-    thrust::transform(handle.get_thrust_policy(),
-                      centralities.begin(),
-                      centralities.end(),
-                      deltas.begin(),
-                      centralities.begin(),
-                      thrust::plus<weight_t>());
+    vertex_t frontier_count = h_bounds[d] - h_bounds[d - 1];
```
seunghwak, line 333: To be multi-GPU ready,

you need to sum frontier_count over all the GPUs using host_scalar_allreduce().

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -259,72 +271,135 @@ void accumulate_vertex_results(
...
+  edge_dst_property_t<vertex_t, edge_t> dst_sigmas(handle, graph_view);
+  edge_dst_property_t<vertex_t, weight_t> dst_deltas(handle, graph_view);
+
+  // Update distances and sigmas initially (deltas start as 0)
+  update_edge_src_property(handle, graph_view, distances.begin(), src_distances.mutable_view());
+  update_edge_src_property(handle, graph_view, sigmas.begin(), src_sigmas.mutable_view());
+  update_edge_dst_property(handle, graph_view, distances.begin(), dst_distances.mutable_view());
+  update_edge_dst_property(handle, graph_view, sigmas.begin(), dst_sigmas.mutable_view());
+  update_edge_src_property(handle, graph_view, deltas.begin(), src_deltas.mutable_view());
+  update_edge_dst_property(handle, graph_view, deltas.begin(), dst_deltas.mutable_view());
```
seunghwak: You can use view_concat here as well, so instead of making 6 primitive calls, it can be just 2.

cpp/src/centrality/betweenness_centrality_impl.cuh
seunghwak: Something minor but,

```
auto v_first = graph_view.local_vertex_partition_range_first();
...
[centralities = centralities.data(), v_first]
```

can just be
```
[centrealities = centralities.data(), v_first = graph_view.local_vertex_partition_range_first()]
```

as this minimizes the scope of v_first.

cpp/src/centrality/oldBC.cuh
seunghwak, line 1: This file is no long relevant, right? Then, just delete.

cpp/tests/centrality/betweenness_centrality_test.cpp
```diff
@@ -34,6 +34,14 @@
...
 
 #include <gtest/gtest.h>
+#include <iostream>
+#include <iomanip>
+#include <fstream>
+#include <sstream>
+#include <algorithm>
+
+// Debug flag - set to true to enable debug output
+#define DEBUG_BETWEENNESS_CENTRALITY 1
```
seunghwak: Delete debug code before merging

testCuGraph.py
seunghwak, line 1: Delete if this is just for internal testing.

testNX.py
seunghwak, line 1: Delete if this is just for internal testing.

testPylibCuGraph.py
seunghwak, line 1: Delete if this is just for internal testing.

test/datasets/manhattan.csv
seunghwak, line 1: Check with @rlratzel that we want to include this file in our datasets.

test/datasets/manhattan_nodes.csv
seunghwak, line 1: Check with @rlratzel that we want to include this file in our datasets.

seunghwak (commented):
The updates in the backward path look good for me.

cpp/tests/centrality/betweenness_centrality_test.cpp
```diff
@@ -229,4 +305,41 @@ INSTANTIATE_TEST_SUITE_P(
...
+  Tests_BetweennessCentrality_File,
+  ::testing::Combine(
+    // disable correctness checks for large dataset, use fewer seeds
+    ::testing::Values(BetweennessCentrality_Usecase{50, false, false, false, false, false},
+                      BetweennessCentrality_Usecase{100, false, false, false, false, false},
+                      BetweennessCentrality_Usecase{200, false, false, false, false, false},
+                      BetweennessCentrality_Usecase{500, false, false, false, false, false},
+                      BetweennessCentrality_Usecase{1000, false, false, false, false, false},
+                      BetweennessCentrality_Usecase{std::numeric_limits<size_t>::max(), false, false, false, false, false}),
+    ::testing::Values(cugraph::test::File_Usecase("/home/nfs/howhuang/cugraph/roadNet-CA.csv"))));
```
seunghwak: You don't need to create new test suits to test a new file.

See

https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/tests/link_analysis/pagerank_test.cpp#L390

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -259,72 +271,138 @@ void accumulate_vertex_results(
...
+    
+    // Copy bounds to host for use in delta loop
+    h_bounds.resize(d_bounds.size());
+    raft::update_host(h_bounds.data(), d_bounds.data(), d_bounds.size(), handle.get_stream());
+    handle.sync_stream();
+  }
+  
+  // Calculate max frontier size using the precomputed bounds
+  vertex_t max_frontier_size = 0;
+  for (vertex_t d = 0; d < h_bounds.size() - 1; ++d) {
```
seunghwak: Please don't ignore compiler warnings.

You will see "comparison of integer expressions of different signedness". Fix it.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -39,6 +42,15 @@
 #include <cuda/std/optional>
 #include <thrust/functional.h>
 #include <thrust/reduce.h>
+#include <thrust/sort.h>
+#include <thrust/transform.h>
+#include <thrust/copy.h>
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/device_vector.h>
```
seunghwak: Why are you including this?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -39,6 +42,15 @@
...
+#include <thrust/execution_policy.h>
```
seunghwak: Add an empty line here, we place thrust include statements in one group and std include statements in another group.

seunghwak: Or just delete these std include statements, these are just for debugging, and irrelevant to the code to be merged.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -104,7 +116,8 @@ struct extract_edge_pred_op_t {
                              thrust::tuple<vertex_t, edge_t, weight_t> dst_props,
                              cuda::std::nullopt_t) const
   {
-    return ((thrust::get<0>(dst_props) == d) && (thrust::get<0>(src_props) == (d - 1)));
+    // For outgoing edges: we want edges where src is at distance d and dst is at distance d-1
+    return ((thrust::get<0>(src_props) == d) && (thrust::get<0>(dst_props) == (d - 1)));
```
seunghwak: Double check with @ChuckHastings that this change is legitimate.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -114,7 +127,8 @@ struct extract_edge_pred_op_t {
                              thrust::tuple<vertex_t, edge_t, weight_t> dst_props,
                              edge_t edge_multi_index) const
   {
-    return ((thrust::get<0>(dst_props) == d) && (thrust::get<0>(src_props) == (d - 1)));
+    // For outgoing edges: we want edges where src is at distance d and dst is at distance d-1
+    return ((thrust::get<0>(src_props) == d) && (thrust::get<0>(dst_props) == (d - 1)));
```
seunghwak: Double check with @ChuckHastings that this change is legitimate.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -39,6 +42,15 @@
 #include <cuda/std/optional>
 #include <thrust/functional.h>
 #include <thrust/reduce.h>
+#include <thrust/sort.h>
+#include <thrust/transform.h>
```
seunghwak, line 49: We no longer call thrust::transform, and I assume this include statement is no longer necessary.

seunghwak (approved):
LGTM except for minor cosmetic issues.s

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: I believe there are 2 more places that use thrust::transform one of them towards the end of the betweenness_centrality() function call

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: Thanks for catching this. This is actually incorrect. Didn't catch this earlier because it seems to only affect accumulate_edge_results() and not accumulate_vertex_results().

ChuckHastings (approved):
Looks good except for file format issues (sent instructions in slack).

PR #5147, opened 2025-07-07
Fix calls to Client.scheduler_info

TomAugspurger:
This fixes hangs when there are more than 5 workers in the dask cluster (as a consequence of https://github.com/dask/distributed/pull/9045; previously that returned all the worker addresses in 'workers', now it's limited to the top 5 by default).


rlratzel (approved):
Thank you!

PR #5149, opened 2025-07-07
Add test for Cosine Similarity algorithm

jnke2016:
This PR adds tests for Cosine Similarity algorithm and update docstrings


jnke2016:
Pushed the wrong PR which was meant to be this [one](https://github.com/rapidsai/cugraph/pull/5150)

PR #5151, opened 2025-07-08
Re-enable spectral clustering

jnke2016:


cpp/include/cugraph/algorithms.hpp
```diff
@@ -879,6 +879,7 @@ std::unique_ptr<legacy::GraphCOO<VT, ET, WT>> extract_subgraph_vertex(
  * 32-bit)
  * @tparam WT                        Type of edge weights. Supported values : float or double.
  *
+ * @param[in]  handle                Library handle (RAFT). If a communicator is set in the handle,
```
BradReesWork, line 882: Where is the line for" raft::random::RngState& rng_state

cpp/include/cugraph/algorithms.hpp
```diff
@@ -913,6 +916,7 @@ void balancedCutClustering(legacy::GraphCSRView<VT, ET, WT> const& graph,
...
```
BradReesWork, line 920: raft::random::RngState& rng_state entry missing

BradReesWork (approved):

PR #5152, opened 2025-07-08
[IMP] Allow Metadata Specification in Distributed Sampler

alexbarghi-nv:
Allows specification of metadata unique to the underlying GNN framework (i.e. PyG) in order to ensure the output minibatches can be processed correctly.  Required to fix a bug in cuGraph-PyG where the wrong input type is selected for the output minibatches.


python/cugraph/cugraph/gnn/data_loading/dist_sampler.py
```diff
@@ -78,6 +78,7 @@ def sample_batches(
         batch_id_offsets: TensorType,
         random_state: int = 0,
         assume_equal_input_size: bool = False,
+        metadata: Optional[Dict[str, Any]] = None,
```
rlratzel, line 102: I don't know the use case for this change, but do we really support a mapping of str to any type?  Mostly just asking, but I'm wondering if that could cause problems later if users expect support for metadata consisting of complex objects. Would it be better to start by supporting more limited types at first then opening it up later?

alexbarghi-nv: That's a fair point.  I could restrict this to the two supported types in `cugraph-pyg`, `str` and `tuple[str, str, str]`.

alexbarghi-nv: Fixed

PR #5159, opened 2025-07-15
Relax PyTorch pinning to >=2.3.

bdice:
This removes the PyTorch upper bound to ensure we can test with current versions.


alexbarghi-nv (approved):
Looks good, thanks for doing this.

Issue #4407, opened 2024-05-09
Investigate Migrating cugraph.gnn Subpackage


alexbarghi-nv:
cuGraph-PyG and cuGraph-DGL are being decoupled from cuGraph to rely on pylibcugraph instead, but we still have a cugraph.gnn subpackage within cugraph that both libraries need.  We should find a better long-term home for that subpackage, especially since we are thinking of changing the release cycle of cuGraph-PyG and cuGraph-DGL and those packages are very tightly coupled with cugraph.gnn.

alexbarghi-nv:
This is complicated by the distributed sampler's reliance on cuDF - maybe there could be an alternative where pytorch tensors are written instead so that cuDF could be an optional dependency.

alexbarghi-nv:
Pushing to 24.12 for now.

alexbarghi-nv:
Postponed to 25.06

alexbarghi-nv:
Current plan is to migrate remaining code to `cugraph-pyg` once `cugraph-dgl` has been removed.

nv-rliu:
I believe to close out this issue, the remaining code in the cugraph-pyg should now point to `comms` inside PLC

alexbarghi-nv:
> I believe to close out this issue, the remaining code in the cugraph-pyg should now point to `comms` inside PLC

Yes

rlratzel:
> > I believe to close out this issue, the remaining code in the cugraph-pyg should now point to `comms` inside PLC
> 
> Yes

How is this issue and https://github.com/rapidsai/cugraph-gnn/issues/249 related? Should one be closed as redundant, or are they tracking unique things?

PR #5166, opened 2025-07-17
Move `gnn.comms` to PLC

nv-rliu:
Part of https://github.com/rapidsai/cugraph/issues/4407


BradReesWork (approved):

alexbarghi-nv:
This is still calling code in `cugraph` that needs to be moved: https://github.com/rapidsai/cugraph/blob/c7a8abe2f7ffcde1edc0b478a507fd373a218fef/python/pylibcugraph/pylibcugraph/comms/cugraph_nccl_comms.py#L22

alexbarghi-nv:
Portions of the code here need to be migrated to PLC: https://github.com/rapidsai/cugraph/blob/branch-25.08/python/cugraph/cugraph/dask/comms/comms.py#L30

alexbarghi-nv:
Oh, and another issue here is that the usage in other files is not updated, but I can fix that too.

alexbarghi-nv (approved):
Approved based on my changes.

alexbarghi-nv:
We still need to migrate the code in the GNN packages - https://github.com/rapidsai/cugraph-gnn/issues/249

alexbarghi-nv:
> This comment and import handling in `cugraph.dask.comms.comms` is concerning in re: adding `raft-dask` as a runtime dependency:
> 
> ```python
> # FIXME: these raft imports break the library if ucx-py is
> # not available. They are necessary only when doing MG work.
> from cugraph.dask.common.read_utils import MissingUCXPy
> try:
>     from raft_dask.common.comms import Comms as raftComms
>     from raft_dask.common.comms import get_raft_comm_state
> except ImportError as err:
>     # FIXME: Generalize since err.name is arr when
>     # libnuma.so.1 is not available
>     if err.name == "ucp" or err.name == "arr":
>         raftComms = MissingUCXPy()
>         get_raft_comm_state = MissingUCXPy()
>     else:
>         raise
> ```
> 
> It seems like this is currently handled as an optional dependency -- if the hard dependency is being added, should there also be an addition of `ucx-py` (or `ucxx` as `ucx-py` is also on the way out)

We should not add any new runtime dependencies with this PR.  I'll make sure we keep it as optional in `pylibcugraph`.

alexbarghi-nv:
@gforsyth sorry for the confusion here - I have a fix ready locally, but I think we have a more long-term issue of how we're handling raft comms.  The functions in `cugraph_nccl_comms.py` aren't supposed to use UCX at all, and in fact they don't.  Ideally I'd like to be able to call `inject_comms_on_handle_coll_only` or something similar with only NCCL installed, and not have to worry about UCX.  If I could do that, then I wouldn't have to add a UCX dependency to `cugraph-pyg` once its `cugraph` dependency gets dropped.

But I also understand that isn't a high priority for the raft team for obvious reasons...

alexbarghi-nv:
@gforsyth please re-review

gforsyth (approved):

PR #5173, opened 2025-07-21
Update update_edge_src|dst_property behavior (with vertex list)

seunghwak:
update_edge_src|dst_property has two versions.

One that updates the values for the entire vertex range, and the second one that takes a vertex list to update.

In the second version, it currently asks users to provide the property values for the entire vertex partition range (same to the first version, the primitives internally gather the values). This behavior is counter intuitive and became the sources of bugs recently.

In this PR, the primitive is updated to take property values in an array which has the size of the vertex list.

### DIFF
--- a/cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
+++ b/cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
@@ -52,22 +52,6 @@ void update_temporal_edge_mask(
   // for each value so that we can reset everything back more efficiently.
   fill_edge_src_property(handle, graph_view, edge_src_times.mutable_view(), STARTING_TIME);
 
-  rmm::device_uvector<edge_time_t> local_frontier_vertex_times(
-    graph_view.local_vertex_partition_range_size(), handle.get_stream());
-
-  thrust::scatter(
-    handle.get_thrust_policy(),
-    vertex_times.begin(),
-    vertex_times.end(),
-    thrust::make_transform_iterator(
-      vertices.begin(),
-      cuda::proclaim_return_type<vertex_t>(
-        [vertex_partition = vertex_partition_device_view_t<vertex_t, multi_gpu>(
-           graph_view.local_vertex_partition_view())] __device__(auto v) {
-          return vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(v);
-        })),
-    local_frontier_vertex_times.begin());
-
   update_edge_src_property(handle,
                            graph_view,
                            vertices.begin(),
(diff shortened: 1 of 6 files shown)


seunghwak:
@ChuckHastings @jnke2016 FYI, this primitive behavior should now better match your intuitive expectation of the primitive behavior.

seunghwak:
@HowardHuang1

Once this PR gets merged

```
      // Scatter deltas back to main deltas array
      thrust::scatter(
        handle.get_thrust_policy(),
        reusable_delta_buffer.begin(),
        reusable_delta_buffer.begin() + frontier_count,
        vertices_sorted.begin() + h_bounds[d - 1],
        deltas.begin()
      );

      // Update properties for vertices in vertex_list
      update_edge_src_property(handle, graph_view, vertex_list.begin(), vertex_list.end(), thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()), src_properties.mutable_view());
      update_edge_dst_property(handle, graph_view, vertex_list.begin(), vertex_list.end(), thrust::make_zip_iterator(distances.begin(), sigmas.begin(), deltas.begin()), dst_properties.mutable_view());
```
https://github.com/rapidsai/cugraph/pull/5146/files#diff-40a2c8b86c13c482d7a1e63034a033959887f8b82c9d43522832866c2203a2b1R379-R390

needs to be updated. No need to call the scatter anymore.

cpp/src/sampling/detail/update_temporal_edge_mask_impl.cuh
```diff
@@ -52,22 +52,6 @@ void update_temporal_edge_mask(
...
-    vertex_times.end(),
-    thrust::make_transform_iterator(
-      vertices.begin(),
-      cuda::proclaim_return_type<vertex_t>(
-        [vertex_partition = vertex_partition_device_view_t<vertex_t, multi_gpu>(
-           graph_view.local_vertex_partition_view())] __device__(auto v) {
-          return vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(v);
-        })),
-    local_frontier_vertex_times.begin());
-
```
seunghwak, line 70: @ChuckHastings Please double check that this change is correct. It seems like the original implementation updated local_frontier_vertex_times but didn't use it.

ChuckHastings: Your update is correct.

That was careless of me.  This was one of the last changes I made to the C++ PR was to delete the old implementation and replace it with this.  The below call to `update_edge_src_property` should be using `local_frontier_vertex_times.begin()` as the fifth parameter and would be switched back to `vertex_times.begin()` in your update.

ChuckHastings (approved):

PR #5194, opened 2025-07-28
cleanup and dropping NX

BradReesWork:
Dropping NX and addressing TODO/FIXMEs

### DIFF
--- a/python/cugraph/cugraph/link_prediction/cosine.py
+++ b/python/cugraph/cugraph/link_prediction/cosine.py
@@ -12,13 +12,11 @@
 # limitations under the License.
 
 from cugraph.utilities import (
-    ensure_cugraph_obj_for_nx,
-    df_edge_score_to_dictionary,
     renumber_vertex_pair,
+    ensure_valid_dtype
 )
 import cudf
-import warnings
-from typing import Union, Iterable
+from warnings import deprecated
 
 from pylibcugraph import (
     cosine_coefficients as pylibcugraph_cosine_coefficients,
@@ -27,41 +25,13 @@ from pylibcugraph import (
 from pylibcugraph import ResourceHandle
 
 from cugraph.structure import Graph
-from cugraph.utilities.utils import import_optional
-
-# FIXME: the networkx.Graph type used in type annotations is specified
-# using a string literal to avoid depending on and importing networkx.
-# Instead, networkx is imported optionally, which may cause a problem
-# for a type checker if run in an environment where networkx is not installed.
-networkx = import_optional("networkx")
-
-
-# FIXME: Move this function to the utility module so that it can be
-# shared by other algos
-def ensure_valid_dtype(input_graph, vertex_pair):
-    vertex_dtype = input_graph.edgelist.edgelist_df.dtypes.iloc[0]
-    vertex_pair_dtypes = vertex_pair.dtypes
-
-    if (
-        vertex_pair_dtypes.iloc[0] != vertex_dtype
-        or vertex_pair_dtypes.iloc[1] != vertex_dtype
-    ):
-        warning_msg = (
-            "Cosine requires 'vertex_pair' to match the graph's 'vertex' type. "
-            f"input graph's vertex type is: {vertex_dtype} and got "
-            f"'vertex_pair' of type: {vertex_pair_dtypes}."
-        )
-        warnings.warn(warning_msg, UserWarning)
-        vertex_pair = vertex_pair.astype(vertex_dtype)
-
-    return vertex_pair
 
 
 def cosine(
     input_graph: Graph,
     vertex_pair: cudf.DataFrame = None,
     use_weight: bool = False,
-):
+) -> cudf.DataFrame:
     """
     Compute the Cosine similarity between each pair of vertices connected by
     an edge, or between arbitrary pairs of vertices specified by the user.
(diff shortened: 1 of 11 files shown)


nv-rliu:
Have we decided to drop interchangeable support for nx Graph objects in favor of using nx-cugraph?

acostadon (approved):
Looks good

BradReesWork:
@nv-rliu all unit tests have been either updated or removed

python/cugraph/cugraph/components/connectivity.py
```diff
@@ -35,7 +33,7 @@ def _ensure_args(api_name, G, directed, connection, return_labels):
     """
     G_type = type(G)
     # Check for Graph-type inputs and set defaults if unset
-    if (G_type in [Graph]) or is_nx_graph_type(G_type):
+    if G_type == Graph:
```
rlratzel: I would replace these lines with:
```python
if isinstance(G, Graph):
```

python/cugraph/cugraph/link_prediction/cosine.py
```diff
@@ -135,7 +102,7 @@ def cosine(
 
     if isinstance(vertex_pair, cudf.DataFrame):
         vertex_pair = renumber_vertex_pair(input_graph, vertex_pair)
-        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair)
+        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair, "Cosine")
```
rlratzel: should match the function name exactly:
```suggestion
        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair, "cosine")
```

python/cugraph/cugraph/link_prediction/cosine.py
```diff
@@ -229,19 +185,14 @@ def cosine_coefficient(
...
-    if isNx is True and ebunch is not None:
-        vertex_pair = cudf.DataFrame(ebunch)
-
-    df = cosine(G, vertex_pair)
+    warnings.warn(
+        "deprecated as of 25.10. This function was to match "
+        "NetworkX which is no longer needed, use networkx directly with"
+        "the ``nx-cugraph`` backend. See:  https://rapids.ai/nx-cugraph/",
+        DeprecationWarning,
+    )
```
rlratzel: The other alternative is to just call `cugraph.cosine()`. 
```suggestion
    warnings.warn(
        "Deprecated as of 25.10. Use `cosine()` instead. If calling with a NetworkX Graph object, use networkx with the nx-cugraph backend. See: https://rapids.ai/nx-cugraph",
        DeprecationWarning,
    )
```

python/cugraph/cugraph/link_prediction/jaccard.py
```diff
@@ -135,7 +105,7 @@ def jaccard(
 
     if isinstance(vertex_pair, cudf.DataFrame):
         vertex_pair = renumber_vertex_pair(input_graph, vertex_pair)
-        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair)
+        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair, "Jaccard")
```
rlratzel: Same suggestion as above, and same for all other calls to `ensure_valid_dtype`
```suggestion
        vertex_pair = ensure_valid_dtype(input_graph, vertex_pair, "jaccard")
```

python/cugraph/cugraph/link_prediction/jaccard.py
```diff
@@ -229,20 +190,17 @@ def jaccard_coefficient(
...
     """
-    vertex_pair = None
-
-    G, isNx = ensure_cugraph_obj_for_nx(G)
+    warnings.warn(
+        "deprecated as of 25.10. This function was to match "
+        "NetworkX which is no longer needed, use networkx directly with"
+        "the ``nx-cugraph`` backend. See:  https://rapids.ai/nx-cugraph/",
+        DeprecationWarning,
+    )
```
rlratzel: I'd follow the same pattern as used for cosine() and suggest either networkx or simply call jaccard().

Same for all other similar deprecations.

python/cugraph/cugraph/traversal/bfs.py
```diff
@@ -42,8 +41,9 @@ def _ensure_args(G, start, i_start, directed):
     start = start if start is not None else i_start
 
     G_type = type(G)
+
     # Check for Graph-type inputs
-    if G_type is Graph or is_nx_graph_type(G_type):
+    if G_type is Graph:
```
rlratzel: Probably also more correct to use `isinstance()` and remove `G_type = type(G)`, here and in all other places.

python/cugraph/cugraph/utilities/utils.py
```diff
@@ -205,12 +195,26 @@ def get_traversed_path_list(df, id):
     return answer
 
 
-# FIXME: if G is a Nx type, the weight attribute is assumed to be "weight", if
-# set. An additional optional parameter for the weight attr name when accepting
-# Nx graphs may be needed.  From the Nx docs:
-# |      Many NetworkX algorithms designed for weighted graphs use
-# |      an edge attribute (by default `weight`) to hold a numerical value.
-def ensure_cugraph_obj(obj, nx_weight_attr=None, matrix_graph_type=None):
+def ensure_valid_dtype(input_graph, vertex_pair, func_name):
```
rlratzel: An alternative to having the caller pass in `func_name` could be to use `inspect` to find the calling function name:
```python
import inspect
func_name = inspect.stack()[1].function
```
Since calling `inspect` is a bit expensive, only do it if the check fails right before issuing the warning. The downside here is that it assumes the calling function name should be in the message which may be wrong if an API calls a helper which calls this.

Or just continue to require `func_name` be passed in, but this also has downsides (similar problem if an API calls a helper which calls this).

python/cugraph/cugraph/components/connectivity.py
BradReesWork: updated

python/cugraph/cugraph/link_prediction/cosine.py
BradReesWork: fixed

PR #5197, opened 2025-07-29
remove references to DGL

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/203

Removes references to DGL and `cugraph-dgl`, now that we are no longer publishing `cugraph-dgl` packages.

## Notes for reviewers

Found these like this:

```shell
git grep -i dgl
```

### DIFF
--- a/ci/build_docs.sh
+++ b/ci/build_docs.sh
@@ -3,8 +3,6 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


jameslamb:
Ah missed that we need `cugraph-python-codeowners` approval too. Will go ask.

alexbarghi-nv (approved):
👍

PR #5199, opened 2025-07-30
Fix sampling bug introduced by compile time optimizations

ChuckHastings:
As part of the compile time optimizations, we started using transform_gather_e.  Unfortunately, the relevant C API tests were not properly returning error codes, so we never detected that there was a bug in this code.

The new primitive `transform_gather_e` expects input edges to be shuffled to the proper GPU, but `per_v_random_select_transform_outgoing_e` does not shuffle them that way, they are shuffled by the major vertex.

This PR adds a shuffle prior to the `transform_gather_e` to get the sampled edges on the proper GPU to retrieve their properties, and then shuffles them back to the GPU of the major vertex.

Also added a FIXME to consider a more optimal solution in the future.

### DIFF
--- a/cpp/src/sampling/detail/gather_one_hop_impl.cuh
+++ b/cpp/src/sampling/detail/gather_one_hop_impl.cuh
@@ -213,7 +213,7 @@ filter_edge_by_type(raft::handle_t const& handle,
   // Filter by type
   using edge_type_t = int32_t;
 
-  const bool store_transposed = false;
+  constexpr bool store_transposed = false;
 
   rmm::device_uvector<edge_type_t> edge_type(srcs.size(), handle.get_stream());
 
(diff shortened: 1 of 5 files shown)


cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -199,23 +199,35 @@ simple_gather_one_hop_without_multi_edge_indices(
...
                     rmm::device_uvector<vertex_t>&& dsts,
-                    std::optional<rmm::device_uvector<edge_t>>&& edge_indices,
+                    arithmetic_device_uvector_t&& edge_indices,
                     std::optional<rmm::device_uvector<int32_t>>&& labels,
                     raft::device_span<uint8_t const> gather_flags)
 {
   // Filter by type
   using edge_type_t = int32_t;
 
+  const bool store_transposed = false;
```
seunghwak: Something minor but, "const" => "constexpr".

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -526,97 +497,101 @@ temporal_gather_one_hop_edgelist(
...
+    result_labels = rmm::device_uvector<label_t>(keep_count, handle.get_stream());
+    thrust::transform(
+      handle.get_thrust_policy(),
+      tmp_positions->begin(),
+      tmp_positions->end(),
+      result_labels->begin(),
+      [kv_store_view = kv_binary_search_store_device_view_t<decltype(kv_store.view())>{
+         kv_store.view()}] __device__(auto pos) {
+        return cuda::std::get<1>(kv_store_view.find(pos));
+      });
```
seunghwak: Why not just call kv_store_view.find()? (https://github.com/rapidsai/cugraph/blob/branch-25.08/cpp/src/prims/kv_store.cuh#L257).

seunghwak (approved):
LGTM (few complaints about minor cosmetic issues).

PR #5206, opened 2025-08-04
Concurrent multisource backwardpass

### LINKED ISSUE #5221: Multi-source BFS for Betweenness Centrality
Improve betweenness centrality implementation performance by running multiple BFS seeds concurrently.

HowardHuang1:
Update the Betweenness Centrality implementation to perform multi-source backwards pass concurrently.

Closes #5221

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,7 +34,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -45,7 +45,7 @@ jobs:
   python-build:
     needs: [cpp-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -55,7 +55,7 @@ jobs:
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -65,7 +65,7 @@ jobs:
     if: github.ref_type == 'branch'
     needs: python-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.10
     with:
       arch: "amd64"
       branch: ${{ inputs.branch }}
@@ -77,7 +77,7 @@ jobs:
       sha: ${{ inputs.sha }}
   wheel-build-libcugraph:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       # build for every combination of arch and CUDA version, but only for the latest Python version
       matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
@@ -92,7 +92,7 @@ jobs:
   wheel-publish-libcugraph:
     needs: wheel-build-libcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -103,7 +103,7 @@ jobs:
   wheel-build-pylibcugraph:
     needs: wheel-build-libcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -115,7 +115,7 @@ jobs:
   wheel-publish-pylibcugraph:
     needs: wheel-build-pylibcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -126,7 +126,7 @@ jobs:
   wheel-build-cugraph:
     needs: wheel-build-pylibcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -138,7 +138,7 @@ jobs:
   wheel-publish-cugraph:
     needs: wheel-build-cugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-publish.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -28,7 +28,7 @@ jobs:
       - telemetry-setup
       - devcontainer
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@branch-25.10
     if: always()
     with:
       needs: ${{ toJSON(needs) }}
@@ -53,7 +53,7 @@ jobs:
   changed-files:
     secrets: inherit
     needs: telemetry-setup
-    uses: rapidsai/shared-workflows/.github/workflows/changed-files.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/changed-files.yaml@branch-25.10
     with:
       files_yaml: |
         test_cpp:
@@ -87,14 +87,14 @@ jobs:
   checks:
     secrets: inherit
     needs: telemetry-setup
-    uses: rapidsai/shared-workflows/.github/workflows/checks.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/checks.yaml@branch-25.10
     with:
       enable_check_generated_files: false
       ignored_pr_jobs: telemetry-summarize
   conda-cpp-build:
     needs: checks
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.10
     with:
       build_type: pull-request
       node_type: cpu32
@@ -102,7 +102,7 @@ jobs:
   conda-cpp-tests:
     needs: [conda-cpp-build, changed-files]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-tests.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-tests.yaml@branch-25.10
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_cpp
     with:
       build_type: pull-request
@@ -110,21 +110,21 @@ jobs:
   conda-cpp-checks:
     needs: conda-cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@branch-25.10
     with:
       build_type: pull-request
       symbol_exclusions: (cugraph::ops|hornet|void writeEdgeCountsKernel|void markUniqueOffsetsKernel)
   conda-python-build:
     needs: conda-cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.10
     with:
       build_type: pull-request
       script: ci/build_python.sh
   conda-python-tests:
     needs: [conda-python-build, changed-files]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@branch-25.10
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
@@ -132,7 +132,7 @@ jobs:
   conda-notebook-tests:
     needs: [conda-python-build, changed-files]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.10
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_notebooks
     with:
       build_type: pull-request
@@ -143,7 +143,7 @@ jobs:
   docs-build:
     needs: conda-python-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.10
     with:
       build_type: pull-request
       node_type: "gpu-l4-latest-1"
@@ -153,7 +153,7 @@ jobs:
   wheel-build-libcugraph:
     needs: checks
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       # build for every combination of arch and CUDA version, but only for the latest Python
       matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
@@ -164,7 +164,7 @@ jobs:
   wheel-build-pylibcugraph:
     needs: wheel-build-libcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       build_type: pull-request
       script: ci/build_wheel_pylibcugraph.sh
@@ -173,7 +173,7 @@ jobs:
   wheel-tests-pylibcugraph:
     needs: [wheel-build-pylibcugraph, changed-files]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-25.10
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
@@ -181,7 +181,7 @@ jobs:
   wheel-build-cugraph:
     needs: wheel-build-pylibcugraph
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-build.yaml@branch-25.10
     with:
       build_type: pull-request
       script: ci/build_wheel_cugraph.sh
@@ -190,7 +190,7 @@ jobs:
   wheel-tests-cugraph:
     needs: [wheel-build-cugraph, changed-files]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-25.10
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python
     with:
       build_type: pull-request
@@ -198,7 +198,7 @@ jobs:
   devcontainer:
     secrets: inherit
     needs: telemetry-setup
-    uses: rapidsai/shared-workflows/.github/workflows/build-in-devcontainer.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/build-in-devcontainer.yaml@branch-25.10
     with:
       arch: '["amd64", "arm64"]'
       cuda: '["13.0"]'
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,7 +25,7 @@ on:
 jobs:
   conda-cpp-checks:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-post-build-checks.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
@@ -34,7 +34,7 @@ jobs:
       symbol_exclusions: (cugraph::ops|hornet|void writeEdgeCountsKernel|void markUniqueOffsetsKernel)
   conda-cpp-tests:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-tests.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-tests.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
@@ -43,7 +43,7 @@ jobs:
       sha: ${{ inputs.sha }}
   conda-python-tests:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-tests.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
@@ -52,7 +52,7 @@ jobs:
       sha: ${{ inputs.sha }}
   wheel-tests-pylibcugraph:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
@@ -61,7 +61,7 @@ jobs:
       script: ci/test_wheel_pylibcugraph.sh
   wheel-tests-cugraph:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/wheels-test.yaml@branch-25.10
     with:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
--- a/.github/workflows/trigger-breaking-change-alert.yaml
+++ b/.github/workflows/trigger-breaking-change-alert.yaml
@@ -12,7 +12,7 @@ jobs:
   trigger-notifier:
     if: contains(github.event.pull_request.labels.*.name, 'breaking')
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@cuda13.0
+    uses: rapidsai/shared-workflows/.github/workflows/breaking-change-alert.yaml@branch-25.10
     with:
       sender_login: ${{ github.event.sender.login }}
       sender_avatar: ${{ github.event.sender.avatar_url }}


cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -13,813 +13,1515 @@
...
+       distances + num_vertices,
+       vertex_t{0},
+       [invalid_distance] __device__(vertex_t a, vertex_t b) -> vertex_t {
+         if (a == invalid_distance && b == invalid_distance) return invalid_distance;
+         if (a == invalid_distance) return b;
+         if (b == invalid_distance) return a;
+         return thrust::max(a, b);
+       });
+     global_max_distance = std::max(global_max_distance, max_dist);
+   }
```
seunghwak:
```
{
auto d_first = thrust::make_transform_iterator(distances.begin(), cuda::procalim_return_type<vertex_t>([invalid_distance]__device__(auto d) { return d == invalid_distance ? vertex_t{0} : d; }));
global_max_distance = thrust::reduce(handle.get_thrust_policy(), d_first, d_first + distances.begin());
}
```
should work. No need to iterate over num_sources.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -13,813 +13,1515 @@
...
+   
+   printf("DEBUG: Global max distance: %ld\n", static_cast<long>(global_max_distance));
+   
+   // Process distance levels using parallel edge enumeration as recommended
+   for (vertex_t d = global_max_distance; d > 1; --d) {
+     auto distance_start_time = std::chrono::high_resolution_clock::now();
```
seunghwak: If you want to property measure execution time,

you need to call RAFT_CUAD_TRY(cudaDeviceSynchronize()); or RAFT_CUDA_TRY(cudaStreamSynchronize(handle.get_stream())); to ensure that all the previously launched CUDA kernels are completed.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -13,813 +13,1515 @@
...
+       });
+     
+     // Copy back to host and compute total
+     std::vector<size_t> vertices_per_source(num_sources);
+     
+     // Use explicit device-to-host copy
+     cudaMemcpy(vertices_per_source.data(), 
+                vertices_per_source_device.data(),
+                num_sources * sizeof(size_t),
+                cudaMemcpyDeviceToHost);
```
seunghwak: Use `raft::update_host`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -13,813 +13,1515 @@
...
+       [distances_2d = distances_2d.data(), num_vertices, d,
+        vertices_per_source = vertices_per_source_device.data()] 
+       __device__(size_t global_idx) {
+         size_t source_idx = global_idx / num_vertices;
+         vertex_t v = global_idx % num_vertices;
+         const vertex_t* distances = distances_2d + source_idx * num_vertices;
+         if (distances[v] == d - 1) {
+           atomicAdd(&vertices_per_source[source_idx], size_t(1));
+         }
+       });
```
seunghwak: We are scanning the entire distance array `global_max_distance` times. This is expensive (and we are scanning the distance array only once in `accumulate_results`; so this means more work compared to the non-concurrent case. If you are incurring any additional work, you need to make sure that is absolutely necessary and the gain is bigger than loss, you are doing `global_max_distance` times more work here... it will be very difficult to get this back).

seunghwak: Outside the loop, you can partition (v, origin) pairs based on distances. Then, you don't need to scan the entire distance array `global_max_distance` times.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -601,38 +601,31 @@ rmm::device_uvector<weight_t> betweenness_centrality(
   // expand from multiple sources concurrently. The challenge is managing
   // the memory explosion.
   //
```
seunghwak: You can delete the above FIXME statement. This is no longer relevant.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -601,38 +601,31 @@ rmm::device_uvector<weight_t> betweenness_centrality(
...
+  multisource_backward_pass(
+    handle,
+    graph_view,
+    edge_weight_view,
+    raft::device_span<weight_t>{centralities.data(), centralities.size()},
+    std::move(distances_2d),
+    std::move(sigmas_2d),
+    raft::device_span<vertex_t const>{sources_buffer.data(), sources_buffer.size()},
+    include_endpoints,
+    do_expensive_check);
```
seunghwak: Here, we need to run all sources concurrently.

To be more future-proof, you may better restructure this code.

```
auto max_concurrent_sources = num_sources;  // FIXME: we should set this based on the memory footprint vs concurrency trade-offs
size_t num_processed{0};
while (num_processed < num_sources) {
  auto num_concurrent_sources = std::min(max_concurrent_sources, num_sources - num_processed);
  rmm::devcie_uvector<vertex_t> sources(num_concurrent_sources, handle.get_stream());
  thrust::copy(handle.get_thrust_policy(), vertices.begin() + num_processed, vertices.begin() + (num_processed + num_concurrent_sources, sources.begin());
  // process num_concurrent_sources concurrently.
  num_processed += num_concurrent_sources; 
}
```

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+  auto num_vertices = graph_view.local_vertex_partition_range_size();
+  auto num_sources  = sources.size();
+
+  rmm::device_uvector<edge_t> sigmas_2d(num_sources * num_vertices, handle.get_stream());
+  rmm::device_uvector<vertex_t> distances_2d(num_sources * num_vertices, handle.get_stream());
+  detail::scalar_fill(handle, distances_2d.data(), distances_2d.size(), invalid_distance);
+  detail::scalar_fill(handle, sigmas_2d.data(), sigmas_2d.size(), edge_t{0});
+
+  // Create tagged frontier with origin indices
+  vertex_frontier_t<vertex_t, origin_t, multi_gpu, true> vertex_frontier(handle, 2);
```
seunghwak: Instead of using 2,

we can define

```
  constexpr size_t bucket_idx_cur  = 0;
  constexpr size_t bucket_idx_next = 1;
  constexpr size_t num_buckets     = 2;
```
( https://github.com/rapidsai/cugraph/blob/branch-25.10/cpp/src/traversal/bfs_impl.cuh#L369) and use num_buckets.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
 
 }  // namespace detail
 
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<edge_t>> multisource_bfs(
```
seunghwak: This function is called by the above detail::betweenness_centrality. The detail space function calling a public space function is a bit odd. We should move this function to the detail space and define this before the above  detail::betweenness_centrality.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+  if (sources.size() > 0) {
+    rmm::device_uvector<origin_t> origins(sources.size(), handle.get_stream());
+
+    // Create origins
+    thrust::sequence(handle.get_thrust_policy(), origins.begin(), origins.end(), origin_t{0});
+
+    // Insert tagged sources into frontier
+    vertex_frontier.bucket(bucket_idx_cur)
+      .insert(thrust::make_zip_iterator(sources.begin(), origins.begin()),
+              thrust::make_zip_iterator(sources.end(), origins.end()));
```
seunghwak: We can use
```
auto pair_first = thrust::make_zip_iterator(vertex_first, thrust::make_counting_iterator(origin_t{0}));
```
And if we modify this function to take VertexIterator vertex_first & vertex_last, we don't need to create a temporary rmm::device_uvector object store sources. This saves both memory & copy time.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                       auto idx           = origin * num_vertices + offset;
+                       d_distance_2d[idx] = 0;
+                       d_sigma_2d[idx]    = 1;
+                     });
+  }
+
+  edge_t hop{0};
+
+  while (vertex_frontier.bucket(bucket_idx_cur).aggregate_size() > 0) {
+    // Step 1: Extract ALL edges from frontier (no filtering)
```
seunghwak: no filtering? Aren't we using a predicate here?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+
+    // Reduce by key to sum sigmas for identical (destination, origin) pairs
+    auto num_unique = thrust::count_if(handle.get_thrust_policy(),
+                                       thrust::make_counting_iterator(size_t{0}),
+                                       thrust::make_counting_iterator(frontier_vertices.size()),
+                                       [vertices = frontier_vertices.data(),
+                                        origins  = frontier_origins.data()] __device__(size_t i) {
+                                         return (i == 0) || (vertices[i] != vertices[i - 1]) ||
+                                                (origins[i] != origins[i - 1]);
+                                       });
```
seunghwak: You can just call thrust::unique_count.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+      cuda::proclaim_return_type<bool>(
+        [d_distances_2d = distances_2d.begin(),
+         num_vertices,
+         vertex_partition,
+         invalid_distance] __device__(auto tagged_src, auto dst, auto, auto, auto) {
+          auto origin     = thrust::get<1>(tagged_src);
+          auto dst_offset = vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(dst);
+          auto dst_idx    = origin * num_vertices + dst_offset;
+          return d_distances_2d[dst_idx] == invalid_distance;
+        }));
```
seunghwak: Why are you using the detail space function?

Can't you just call the public space function? (https://github.com/rapidsai/cugraph/blob/branch-25.10/cpp/src/prims/extract_transform_if_v_frontier_outgoing_e.cuh#L86)

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+
+    thrust::transform(handle.get_thrust_policy(),
+                      frontier_tuples.begin(),
+                      frontier_tuples.end(),
+                      thrust::make_zip_iterator(
+                        frontier_vertices.begin(), frontier_origins.begin(), sigmas.begin()),
+                      [] __device__(auto tuple) {
+                        return thrust::make_tuple(
+                          thrust::get<0>(tuple), thrust::get<1>(tuple), thrust::get<2>(tuple));
+                      });
```
seunghwak: Why are you making copies here?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+
+  return std::make_tuple(std::move(distances_2d), std::move(sigmas_2d));
+}
+
+// Remove the problematic device functor - use simpler lambda approach instead
+
+// Parallelized backward pass
+
+template <typename vertex_t, typename edge_t, typename weight_t, bool multi_gpu>
+void multisource_backward_pass(
```
seunghwak: Similar here, move this to the detail space (and before detail::betweenness_centrality)

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                                                d_first + distances_2d.size(),
+                                                vertex_t{0},
+                                                thrust::maximum<vertex_t>());
+
+  // PRE-COMPUTE: Partition all (vertex, source) pairs by distance ONCE
+  // This eliminates the need to scan the distance array global_max_distance times
+
+  // Create buckets for each distance level
+  std::vector<rmm::device_uvector<vertex_t>> distance_buckets_vertices;
+  std::vector<rmm::device_uvector<size_t>> distance_buckets_sources;
```
seunghwak: We defines `origin_t` above. Don't mix origin_t and size_t, use either origin_t or size_t consistently.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                      num_vertices,
+                      distance_counts = distance_counts.data(),
+                      global_max_distance] __device__(size_t global_idx) {
+                       size_t source_idx         = global_idx / num_vertices;
+                       vertex_t v_offset         = global_idx % num_vertices;
+                       const vertex_t* distances = distances_2d + source_idx * num_vertices;
+                       vertex_t dist             = distances[v_offset];
+
+                       if (dist >= 0 && dist <= global_max_distance) {
+                         atomicAdd(&distance_counts[dist], size_t{1});
```
seunghwak: `atomicAdd` is old style. Use cuda::atomic_ref and fetch_add (e.g. https://github.com/rapidsai/cugraph/blob/branch-25.10/cpp/src/community/edge_triangle_count_impl.cuh#L79).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                      bucket_sources_ptrs  = device_bucket_sources_ptrs.data(),
+                      v_first,
+                      global_max_distance] __device__(size_t global_idx) {
+                       size_t source_idx         = global_idx / num_vertices;
+                       vertex_t v_offset         = global_idx % num_vertices;
+                       const vertex_t* distances = distances_2d + source_idx * num_vertices;
+                       vertex_t dist             = distances[v_offset];
+
+                       if (dist >= 0 && dist <= global_max_distance) {
+                         size_t offset = atomicAdd(&distance_counts[dist], size_t{1});
```
seunghwak: atomicAdd=>atomic_ref + fetch_add

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                                                                           handle.get_stream());
+      thrust::transform(
+        handle.get_thrust_policy(),
+        frontier_vertices.begin(),
+        frontier_vertices.end(),
+        frontier_sources.begin(),
+        tagged_vertices.begin(),
+        [] __device__(auto vertex, auto source) { return thrust::make_tuple(vertex, source); });
+
+      frontier.bucket(0).insert(tagged_vertices.begin(), tagged_vertices.end());
```
seunghwak: Why are you creating temporaries here?

```
auto pair_first = thrust::make_zip_iterator(frontier_vertices.begin(), frontier_sources.begin());
frontier.bucket(0).insert(pair_first, pair_firswt + frontier_vertices.size());
```
```

File: cpp/src/centrality/betweenness_centrality_impl.cuh [hunk 19]
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+
+                  // Check if dst is at distance d for this source
+                  const vertex_t* distances = distances_2d + source_idx * num_vertices;
+                  const edge_t* sigmas      = sigmas_2d + source_idx * num_vertices;
+                  const weight_t* deltas    = delta_buffer + source_idx * num_vertices;
+
+                  auto src_offset = src - v_first;
+                  auto dst_offset = dst - v_first;
+
+                  if (distances[dst_offset] == d) {
```
seunghwak: This has already been checked in the predicate, no need to check again.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+        handle.get_thrust_policy(),
+        frontier_vertices.begin(),
+        frontier_vertices.end(),
+        frontier_sources.begin(),
+        tagged_vertices.begin(),
+        [] __device__(auto vertex, auto source) { return thrust::make_tuple(vertex, source); });
+
+      frontier.bucket(0).insert(tagged_vertices.begin(), tagged_vertices.end());
+
+      auto result = detail::extract_transform_if_v_frontier_e<
```
seunghwak: Why are you using a detail space function? Use a public space one.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+                          get_dataframe_buffer_begin(edge_tuples_buffer),
+                          get_dataframe_buffer_end(edge_tuples_buffer),
+                          thrust::make_zip_iterator(edge_keys.begin(), edge_deltas.begin()),
+                          [] __device__(auto tuple) {
+                            return thrust::make_tuple(
+                              thrust::make_tuple(thrust::get<0>(tuple),
+                                                 thrust::get<1>(tuple)),  // (src, source_idx)
+                              thrust::get<3>(tuple)                       // delta
+                            );
+                          });
```
seunghwak: I don't think you need to produce temporaries and copy to the temporaries.

You can directly use `result`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+           num_vertices,
+           v_first = graph_view.local_vertex_partition_range_first()] __device__(auto tuple) {
+            auto key        = thrust::get<0>(tuple);
+            auto delta      = thrust::get<1>(tuple);
+            auto src        = thrust::get<0>(key);
+            auto source_idx = thrust::get<1>(key);
+
+            // Update centrality
+            auto src_offset = src - v_first;
+            atomicAdd(&centralities[src_offset], delta);
```
seunghwak: atomicAdd=>atomic_ref + fetch_add

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -827,6 +820,560 @@ edge_property_t<edge_t, weight_t> edge_betweenness_centrality(
...
+
+            // Accumulate delta for next iteration
+            weight_t* source_deltas = delta_buffer + source_idx * num_vertices;
+            atomicAdd(&source_deltas[src_offset], delta);
```
seunghwak: Same here.

seunghwak: And double check you really need to update delta_buffer here. You aren't using this anywhere else, no?

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: The extract_transform_if_v_frontier_e fucntion rreturns a dataframe buffer which to my knowledge is not designed to be modified in-place safely. However, most thrust operations including sort_by_key reduce_by_key require mutable data. Thus, we copy to a mutable data structure first since summing/combining sigmas across same (vertex, origin) pairs requires mutability.

Let me know if there is a different workaround that can let me operate directly on the returned buffer safely!

seunghwak: >is not designed to be modified in-place safely

They can be modified. dataframe_buffer_t<T> where T is a tuple is just a std tuple of rmm::device_uvector objects.

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: I've made this change but it ends up slowing things down. With 500 sources atomicAdd on California road network takes 4.78 seconds but with atomic_ref + fetch_add it takes just over 5 seconds for same number of sources. Is there a reason why atomic_ref + fetch_add is preferred over atomicAdd? Does this have to do with thread safety or something other than performance?

seunghwak: atomicAdd is old style, this is more modern. You get slowdown after just replacing atomicAdd here or everywhere? I should ask this in the slack channel next week, but it seems like atomic_ref + fetch_add has some issues with floating point numbers pre-C++20. If you don't see any slow down with ints but see some slowdowns with floats, it might be related to this.

Let's ask this in slack next week.

HowardHuang1: I get slowdown after replacing atomicAdd everywhere. Haven't tested with just replacing this single line. Ok sounds good thanks.

seunghwak: Try just replacing atomic on integers (but use atomicAdd for floating point numbers) and let me know the results. If you don't see slowdown in this case, this might be related to the issues in floating point numbers (pre C++20).

HowardHuang1: I tried replacing atomic on integers and using atomicAdd for floating point numbers but slowdown is still very noticeable. It is consistently a bit faster using atomicAdd for all operations (int & floating point).

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: delta_buffer update is needed because each iteration of the brandes algo relies on the delta values in the previous distance level to compute the deltas in the current distance level. Thus, delta_buffer needs to store all the delta values from previous levels. Otherwise, without this line delta_buffer will not be updated and will just remain 0.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+  VertexIterator vertex_first,
+  VertexIterator vertex_last,
+  bool do_expensive_check)
+{
+  constexpr vertex_t invalid_distance = std::numeric_limits<vertex_t>::max();
+  constexpr size_t bucket_idx_cur     = 0;
+  constexpr size_t bucket_idx_next    = 1;
+  constexpr size_t num_buckets        = 2;
+
+  using origin_t = uint32_t;  // Source index type
```
seunghwak: Add
CUGRAPH_EXPECTS(cuda::std::distance(vertex_first, vertex_last) <= std::numeric_limits<origin_t>::max(), "error message here");

seunghwak: Otherwise, origin_t will overflow.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+
+  // Use 2D arrays to track per-source distances and sigmas
+  // Layout: [source_idx * num_vertices + vertex_idx]
+  auto num_vertices = graph_view.local_vertex_partition_range_size();
+  auto num_sources  = cuda::std::distance(vertex_first, vertex_last);
+
+  rmm::device_uvector<edge_t> sigmas_2d(num_sources * num_vertices, handle.get_stream());
+  rmm::device_uvector<vertex_t> distances_2d(num_sources * num_vertices, handle.get_stream());
+  detail::scalar_fill(handle, distances_2d.data(), distances_2d.size(), invalid_distance);
+  detail::scalar_fill(handle, sigmas_2d.data(), sigmas_2d.size(), edge_t{0});
```
seunghwak: Sorry for nitpicking, but why do you define `sigmas_2d` first but fill `distances_2d` before `sigmas_2d`? This hurts code readability.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+  auto vertex_partition =
+    vertex_partition_device_view_t<vertex_t, multi_gpu>(graph_view.local_vertex_partition_view());
+
+  // Initialize sources with their origins using zip iterator approach
+  if (num_sources > 0) {
+    // Create zip iterator for (vertex, origin) pairs
+    auto pair_first =
+      thrust::make_zip_iterator(vertex_first, thrust::make_counting_iterator(origin_t{0}));
+    auto pair_last = thrust::make_zip_iterator(
+      vertex_last, thrust::make_counting_iterator(origin_t{static_cast<uint32_t>(num_sources)}));
```
seunghwak: `pair_first + num_sources` is simpler.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+         vertex_partition,
+         invalid_distance] __device__(auto tagged_src, auto dst, auto, auto, auto) {
+          auto origin     = thrust::get<1>(tagged_src);
+          auto dst_offset = vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(dst);
+          auto dst_idx    = origin * num_vertices + dst_offset;
+          return d_distances_2d[dst_idx] == invalid_distance;
+        }));
+
+    // Step 2: Work directly with the dataframe buffer components (no temporaries needed)
+    auto new_frontier_tagged_vertex_buffer = std::move(result);
```
seunghwak: Just name `result` to `new_frontier_tagged_vertex_buffer`. No need for this line.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+      // Step 3: Use extract_transform_if_v_frontier_e to enumerate all qualifying edges
+      // This extracts (src, tag, dst) triplets as recommended
+
+      // Create a proper frontier object for the tagged vertices
+      vertex_frontier_t<vertex_t, origin_t, multi_gpu, true> frontier(handle, 1);
+
+      // Insert tagged vertices directly using zip iterator (no temporary needed)
+      auto pair_first =
+        thrust::make_zip_iterator(frontier_vertices.begin(), frontier_sources.begin());
+      frontier.bucket(0).insert(pair_first, pair_first + frontier_vertices.size());
```
seunghwak: I don't think the forntier.bucket(0) elements are sorted. Make sure they are sorted (as you set `sorted_unique` to true in `vertex_frontier_t<vertex_t, origin_t, multi_gpu, true> frontier(handle, 1);`

seunghwak: If you need to sort each partition anyways, it might be better to first sort (v, origin, distance) triplets (primary key: dist, secondary key: v, tertiary key: origin), and run binary search to find partition boundaries.

Compare the performance.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+          [d, distances_2d = distances_2d.data(), num_vertices, v_first] __device__(
+            auto tagged_src, auto dst, auto, auto, auto) {
+            auto source_idx           = thrust::get<1>(tagged_src);
+            const vertex_t* distances = distances_2d + source_idx * num_vertices;
+            auto dst_offset           = dst - v_first;
+            return distances[dst_offset] == d;
+          }));
+
+      // Step 4: Work directly with the result buffer (no temporaries needed)
+      auto edge_tuples_buffer = std::move(result);
```
seunghwak: Just renumber result to `edge_tuples_buffer`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,550 @@ void accumulate_edge_results(
...
+          handle.get_thrust_policy(),
+          srcs.begin(),
+          srcs.end(),
+          thrust::make_zip_iterator(source_indices.begin(), deltas.begin()));
+
+        // Then sort by source indices within each source group (stable sort preserves source order)
+        thrust::stable_sort_by_key(handle.get_thrust_policy(),
+                                   source_indices.begin(),
+                                   source_indices.end(),
+                                   thrust::make_zip_iterator(srcs.begin(), deltas.begin()));
```
seunghwak: ???
Can you just sort using (src, source_index) as a key?

seunghwak (commented):
Have you tested this in multi-GPU? 

The updated code won't work in multi-GPU, are you calling the old code in multi-GPU and invoking the new code in single-GPU?

cpp/src/centrality/betweenness_centrality_impl.cuh
HowardHuang1: The frontier.bucket gets frontier_vertices inserted into it. Since frontier_vertices was already sorted in the forward pass by distance levels, frontier.bucket must also be sorted since we're inserting sorted elements into it.

seunghwak:
@ChuckHastings Any thoughts on why CI is not running here?

HowardHuang1:
Is this a particular config I need to setup?

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +539,543 @@ void accumulate_edge_results(
...
+  std::vector<rmm::device_uvector<origin_t>> distance_buckets_sources;
+
+  // Reserve space and create empty buckets
+  distance_buckets_vertices.reserve(global_max_distance + 1);
+  distance_buckets_sources.reserve(global_max_distance + 1);
+
+  for (vertex_t d = 0; d <= global_max_distance; ++d) {
+    distance_buckets_vertices.emplace_back(0, handle.get_stream());
+    distance_buckets_sources.emplace_back(0, handle.get_stream());
+  }
```
seunghwak: Minor code readability issue, but in general, it is better to minimize the scope of a variable.

distance_buckets_vertices & distance_buckets_sources won't be used till we allocate the exact sized buckets (line 812). No need to define here. And no need to first call emplace_back here and resize later. We can just directly call emplace_back with the size we know. This also cuts 4 lines of code.

cpp/src/centrality/betweenness_centrality_impl.cuh
seunghwak:
```
        cuda::atomic_ref<size_t, cuda::thread_scope_device> counter(distance_counts[dist]);
        size_t offset = counter.fetch_add(size_t{1}, cuda::std::memory_order_relaxed);
        bucket_vertices_ptrs[dist][offset] = v_first + v_offset;
        bucket_sources_ptrs[dist][offset]  = source_idx;
```
Here, for the (vertex, source_idx) pairs having the same distance, they can be inserted in an arbitrary order, right?

So,
```
    auto& frontier_vertices            = distance_buckets_vertices[d - 1];
    auto& frontier_sources             = distance_buckets_sources[d - 1];
```
won't be sorted based on (vertex, source_idx);

But
```
      vertex_frontier_t<vertex_t, origin_t, multi_gpu, true> frontier(handle, 1);
```
says the keys inserted to the buckets will be sorted (the last template parameter).

This looks like a bug, and am I missing something?

seunghwak:
> /ok to test [257375a](https://github.com/rapidsai/cugraph/commit/257375af914443c6beb3e7495cc1f8269ca65586)

So should I specify a specific commit?

ChuckHastings:
Yes, if the commits aren't signed than the ok to test message can reference the most recent commit to authorize testing.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -48,6 +49,9 @@
 #include <thrust/sort.h>
 #include <thrust/transform.h>
 
+// Add CUB include for better sorting performance
```
seunghwak: I think this comment is an overkill.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+  // Sort vertices by vertex ID within each distance level using chunked CUB segmented sort
+  if (total_vertices > 0) {
+    // Calculate target chunk size based on GPU hardware (like manager's code)
+    auto approx_vertices_to_sort_per_iteration =
+      static_cast<size_t>(handle.get_device_properties().multiProcessorCount) * (1 << 20);
+
+    printf("DEBUG: GPU has %d SMs, target chunk size: %zu vertices\n",
+           handle.get_device_properties().multiProcessorCount,
+           approx_vertices_to_sort_per_iteration);
```
seunghwak: Delete this before merging the code

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+  // Count total vertices across all distance levels
+  size_t total_vertices = 0;
+  for (vertex_t d = 1; d <= global_max_distance; ++d) {
+    total_vertices += distance_buckets_vertices[d].size();
+  }
+
+  // Sort vertices by vertex ID within each distance level using chunked CUB segmented sort
+  if (total_vertices > 0) {
+    // Calculate target chunk size based on GPU hardware (like manager's code)
```
seunghwak: manager's code? Here, you aren't writing comments for you, for others reading this code later.

Just saying

// Calculate target chunk size based on GPU hardware

will be sufficient

seunghwak: We are basically setting the chunk size large enough to saturate a GPU but small enough to avoid memory allocation failure.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+    auto approx_vertices_to_sort_per_iteration =
+      static_cast<size_t>(handle.get_device_properties().multiProcessorCount) * (1 << 20);
```
seunghwak: Add /* tuning parameter */ after (1 << 20) to inform that this value can be modified without impacting code correctness.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+  // Allocate exact-sized buckets
+  for (vertex_t d = 0; d <= global_max_distance; ++d) {
+    if (host_distance_counts[d] > 0) {
+      distance_buckets_vertices.emplace_back(host_distance_counts[d], handle.get_stream());
+      distance_buckets_sources.emplace_back(host_distance_counts[d], handle.get_stream());
+    } else {
+      distance_buckets_vertices.emplace_back(0, handle.get_stream());
+      distance_buckets_sources.emplace_back(0, handle.get_stream());
+    }
```
seunghwak: Do you need this if - else statement? The if and the else paths execute the same code.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+                      host_bucket_vertices_ptrs.data(),
+                      global_max_distance + 1,
+                      handle.get_stream());
+  raft::update_device(device_bucket_sources_ptrs.data(),
+                      host_bucket_sources_ptrs.data(),
+                      global_max_distance + 1,
+                      handle.get_stream());
+
+  // Ensure pointer arrays are copied before kernel launch
+  handle.sync_stream();
```
seunghwak: No need here as this host to copy operation and the following device kernel launch both run on the same stream.

seunghwak: In general, you need to call handle.sync_stream() when you copy data to host and manipulate the data in host (as the host code is not stream ordered) but you don't need this when you copy data to device and launch a device kernel (unless they execute on different streams)

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+           handle.get_device_properties().multiProcessorCount,
+           approx_vertices_to_sort_per_iteration);
+
+    // Build distance level offsets array for chunking
+    std::vector<size_t> distance_level_offsets;
+    distance_level_offsets.push_back(0);
+
+    size_t cumulative_vertices = 0;
+    for (vertex_t d = 1; d <= global_max_distance; ++d) {
+      if (d < distance_buckets_vertices.size() && distance_buckets_vertices[d].size() > 0) {
```
seunghwak: Why do you need this? distance_bucket_vertices.size() == global_max_distance + 1, so the first condition can never be false. If the distance_bucekt_vertices[d] = 0, we just cumulative vertices just stay same. And in single GPU, I guess distance_bucket_vertices[d] can't be 0.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+      auto [h_distance_level_chunk_offsets, h_vertex_chunk_offsets] =
+        detail::compute_offset_aligned_element_chunks(
+          handle,
+          raft::device_span<size_t const>(d_distance_level_offsets.data(),
+                                          d_distance_level_offsets.size()),
+          total_vertices,
+          approx_vertices_to_sort_per_iteration);
+
+      auto num_chunks = h_distance_level_chunk_offsets.size() - 1;
+      printf("DEBUG: Created %zu chunks for distance levels\n", num_chunks);
```
seunghwak: Delete this.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+      // Find max chunk size for memory allocation
+      size_t max_chunk_size = 0;
+      for (size_t i = 0; i < num_chunks; ++i) {
+        max_chunk_size =
+          std::max(max_chunk_size,
+                   static_cast<size_t>(h_vertex_chunk_offsets[i + 1] - h_vertex_chunk_offsets[i]));
+      }
+      printf("DEBUG: Max chunk size: %zu vertices\n", max_chunk_size);
```
seunghwak: Delete.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+      // Allocate reusable arrays for chunks
+      rmm::device_uvector<int32_t> chunk_vertices_int32(max_chunk_size, handle.get_stream());
```
seunghwak: int32_t => vertex_t.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+        size_t chunk_vertex_end      = h_vertex_chunk_offsets[chunk_i + 1];
+        size_t chunk_distance_start  = h_distance_level_chunk_offsets[chunk_i];
+        size_t chunk_distance_end    = h_distance_level_chunk_offsets[chunk_i + 1];
+        size_t chunk_size            = chunk_vertex_end - chunk_vertex_start;
+        size_t num_segments_in_chunk = chunk_distance_end - chunk_distance_start;
+
+        printf("DEBUG: Processing chunk %zu: %zu vertices, %zu distance levels\n",
+               chunk_i,
+               chunk_size,
+               num_segments_in_chunk);
```
seunghwak: Delete.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+        std::vector<size_t> chunk_segment_offsets;
+        chunk_segment_offsets.push_back(0);
+
+        // Map distance level indices to actual distance values
+        std::vector<vertex_t> distance_levels_in_chunk;
+        for (vertex_t d = 1; d <= global_max_distance; ++d) {
+          if (d < distance_buckets_vertices.size() && distance_buckets_vertices[d].size() > 0) {
+            distance_levels_in_chunk.push_back(d);
+          }
+        }
```
seunghwak: This code is not necessary.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+          // Copy sources
+          thrust::copy(handle.get_thrust_policy(),
+                       distance_buckets_sources[d].begin(),
+                       distance_buckets_sources[d].end(),
+                       chunk_sources.begin() + write_offset);
+
+          write_offset += level_size;
+          chunk_segment_offsets.push_back(write_offset);
+        }
```
seunghwak: This as well.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+
+        if (num_segments_in_chunk > 0) {
+          // Copy segment offsets to device
+          rmm::device_uvector<size_t> d_chunk_segment_offsets(chunk_segment_offsets.size(),
+                                                              handle.get_stream());
+          raft::update_device(d_chunk_segment_offsets.data(),
+                              chunk_segment_offsets.data(),
+                              chunk_segment_offsets.size(),
+                              handle.get_stream());
```
seunghwak: This too.

See this.
https://github.com/rapidsai/cugraph/blob/branch-25.10/cpp/src/structure/detail/structure_utils.cuh#L508

You can just use distance_level_offsets (just need to shift the offset values properly).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +542,720 @@ void accumulate_edge_results(
...
+                              distance_buckets_vertices[d].begin(),
+                              [] __device__(int32_t v32) { return static_cast<vertex_t>(v32); });
+
+            thrust::copy(handle.get_thrust_policy(),
+                         chunk_sources.begin() + read_offset,
+                         chunk_sources.begin() + read_offset + level_size,
+                         distance_buckets_sources[d].begin());
+
+            read_offset += level_size;
+          }
```
seunghwak: You don't need to copy in loops. Just copy for the entire chunk.

seunghwak (commented):
Please ping me in slack if you have any questions.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+  constexpr size_t bucket_idx_cur     = 0;
+  constexpr size_t bucket_idx_next    = 1;
+  constexpr size_t num_buckets        = 2;
+
+  using origin_t = uint32_t;  // Source index type
+
+  // Check that number of sources doesn't overflow origin_t
+  CUGRAPH_EXPECTS(
+    cuda::std::distance(vertex_first, vertex_last) <= std::numeric_limits<origin_t>::max(),
+    "Number of sources exceeds maximum value for origin_t (uint32_t), would cause overflow");
```
seunghwak: You may set origin_t to uint16_t to cut memory capacity & bandwidth requirements.

I don't think speedup will be dramatic but it can be non-negligible...

uint16_t can cover up to 65535 sources, and I assume for most practical use cases (except for tiny graphs), limiting the maximum number of concurrent sources to this number will be sufficient.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+                       d_distance_2d[idx] = 0;
+                       d_sigma_2d[idx]    = 1;
+                     });
+  }
+
+  edge_t hop{0};
+
+  while (vertex_frontier.bucket(bucket_idx_cur).aggregate_size() > 0) {
+    // Step 1: Extract ALL edges from frontier (filtered by unvisited vertices)
+    using bfs_edge_tuple_t = thrust::tuple<vertex_t, origin_t, edge_t>;
```
seunghwak: We are migrating from thrust::tuple to cuda::std::tuple. Don't use thrust::tuple anymore.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+      graph_view,
+      vertex_frontier.bucket(bucket_idx_cur),
+      edge_src_dummy_property_t{}.view(),
+      edge_dst_dummy_property_t{}.view(),
+      edge_dummy_property_t{}.view(),
+      cuda::proclaim_return_type<bfs_edge_tuple_t>(
+        [d_sigma_2d = sigmas_2d.begin(), num_vertices, vertex_partition] __device__(
+          auto tagged_src, auto dst, auto, auto, auto) {
+          auto src        = thrust::get<0>(tagged_src);
+          auto origin     = thrust::get<1>(tagged_src);
```
seunghwak: thrust::get => cuda::std::get

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+          auto src_offset = vertex_partition.local_vertex_partition_offset_from_vertex_nocheck(src);
+          auto src_idx    = origin * num_vertices + src_offset;
+          auto src_sigma  = static_cast<edge_t>(d_sigma_2d[src_idx]);
+
+          return thrust::make_tuple(dst, origin, src_sigma);
```
seunghwak: thrust::make_tuple=>cuda::std::make_tuple

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+      thrust::make_zip_iterator(frontier_vertices.begin(), frontier_origins.begin()),
+      thrust::make_zip_iterator(frontier_vertices.end(), frontier_origins.end()),
+      sigmas.begin());
+
+    // Reduce by key to sum sigmas for identical (destination, origin) pairs
+    auto num_unique = thrust::unique_count(
+      handle.get_thrust_policy(),
+      thrust::make_zip_iterator(frontier_vertices.begin(), frontier_origins.begin()),
+      thrust::make_zip_iterator(frontier_vertices.end(), frontier_origins.end()),
+      [] __device__(auto const& a, auto const& b) { return a == b; });
```
seunghwak: I assume you don't need this. You are not using this anywhere.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+                       d_distances_2d[idx] = hop + 1;
+                       d_sigmas_2d[idx]    = sigma;
+                     });
+
+    // Step 4: Update frontier for next iteration using in-place reduced data
+    vertex_frontier.bucket(bucket_idx_cur).clear();
+    vertex_frontier.bucket(bucket_idx_next)
+      .insert(thrust::make_zip_iterator(frontier_vertices.begin(), frontier_origins.begin()),
+              thrust::make_zip_iterator(frontier_vertices.begin() + num_reduced,
+                                        frontier_origins.begin() + num_reduced));
```
seunghwak: Something minor  but

thrust::make_zip_iterator(frontier_vertices.begin(), frontier_origins.begin()) + num_reduced is slightly more concise.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+          typename edge_t,
+          typename weight_t,
+          bool multi_gpu,
+          typename VertexIterator>
+std::tuple<rmm::device_uvector<vertex_t>, rmm::device_uvector<edge_t>> multisource_bfs(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  std::optional<edge_property_view_t<edge_t, weight_t const*>> edge_weight_view,
+  VertexIterator vertex_first,
+  VertexIterator vertex_last,
```
seunghwak: We are using source_first & source_last for the backward path.

https://github.com/rapidsai/cugraph/pull/5206/files#diff-40a2c8b86c13c482d7a1e63034a033959887f8b82c9d43522832866c2203a2b1R733

Better be consistent, and we may better rename this to source_first and source_last.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+  rmm::device_uvector<edge_t>&& sigmas_2d,
+  VertexIterator sources_first,
+  VertexIterator sources_last,
+  bool include_endpoints,
+  bool do_expensive_check)
+{
+  auto num_vertices = static_cast<size_t>(graph_view.local_vertex_partition_range_size());
+  auto num_sources  = cuda::std::distance(sources_first, sources_last);
+
+  using origin_t = uint32_t;
```
seunghwak: Same here. Check for overflow, and we may use uint16_t.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+      static_cast<size_t>(handle.get_device_properties().multiProcessorCount) *
+      (1 << 20); /* tuning parameter */
+
+    // Build distance level offsets array for chunking
+    std::vector<size_t> distance_level_offsets;
+    distance_level_offsets.push_back(0);
+
+    size_t cumulative_vertices = 0;
+    for (vertex_t d = 0; d <= global_max_distance; ++d) {
+      if (distance_buckets_vertices[d].size() > 0) {
```
seunghwak: distance_buckets_vertices[d] can't be 0, this if statement is not necessary.

As long as total_vertices > 0, distance_buckets_vertices[global_max_distance] >= 1 and there should be at least one vertex in each d.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+        cumulative_vertices += distance_buckets_vertices[d].size();
+        distance_level_offsets.push_back(cumulative_vertices);
+      }
+    }
+
+    if (distance_level_offsets.size() > 1) {
```
seunghwak: Is this necessary? I assume distance_level_offsets.size() is always >= 2.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -538,6 +540,681 @@ void accumulate_edge_results(
...
+
+          // Copy sources
+          thrust::copy(handle.get_thrust_policy(),
+                       distance_buckets_sources[d].begin(),
+                       distance_buckets_sources[d].end(),
+                       chunk_sources.begin() + write_offset);
+
+          write_offset += level_size;
+          chunk_segment_offsets.push_back(write_offset);
+        }
```
seunghwak: You can skip this copy if you maintain a (vertex, origin) pairs in a consecutive array (instead of creating separate vector pairs for every d).

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -594,44 +1271,55 @@ rmm::device_uvector<weight_t> betweenness_centrality(
...
-  for (size_t source_idx = 0; source_idx < num_sources; ++source_idx) {
-    //
-    //  BFS
-    //
-    constexpr size_t bucket_idx_cur = 0;
-    constexpr size_t num_buckets    = 2;
-
-    vertex_frontier_t<vertex_t, void, multi_gpu, true> vertex_frontier(handle, num_buckets);
+  if constexpr (multi_gpu) {
+    // Multi-GPU: Use sequential brandes_bfs (more reliable for cross-GPU)
```
seunghwak: "more reliable" doesn't sound right here. Our multi-source code is not ready for multi-GPU yet. Once it is updated, it should be as reliable as the sequential version.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -594,44 +1271,55 @@ rmm::device_uvector<weight_t> betweenness_centrality(
...
-    accumulate_vertex_results(handle,
-                              graph_view,
-                              edge_weight_view,
-                              raft::device_span<weight_t>{centralities.data(), centralities.size()},
-                              std::move(distances),
-                              std::move(sigmas),
-                              include_endpoints,
-                              do_expensive_check);
+    auto [distances_2d, sigmas_2d] = detail::multisource_bfs(
+      handle, graph_view, edge_weight_view, vertices_begin, vertices_end, do_expensive_check);
```
seunghwak: Where are you limiting the number of concurrent origins? You need to limit the number of origins within the size of origin_t. If the total number exceeds the limit, you need to call multisource_bfs (and backward_pass) in multiple rounds.

seunghwak (approved):
LGTM, and great job!!!

seunghwak:
/ok to test [95764a4](https://github.com/rapidsai/cugraph/pull/5206/commits/95764a4cd8efd98a9e5c2cbd125ff70def864c32)

ChuckHastings (approved):

Issue #4680, opened 2024-09-29
[QST]: force_atlas2 Prevent Node Overlaps


WXXXQXXXCXX:
### What is your question?

Hi,
I tried to set `prevent_overlapping` to true in `cugraph.force_atlas2()` to minimise node overlap. However, this gives feature not supported exception. May I ask if there is plan to add this feature, or is there other ways to prevent node overlaps? Thanks!

### Code of Conduct

- [x] I agree to follow cuGraph's Code of Conduct
- [x] I have searched the [open issues](https://github.com/rapidsai/cugraph/issues) and have found no duplicates for this question

chaitanyawho:
I'm also very curious about this feature. Is it the case that it hasn't been implemented yet or is there another way to use it? If so, could you please point me to the documentation for it?

ChuckHastings:
We haven't done anything with the force_atlas2 code in a while.  No changes are on the current road map.

We'll have some discussion about this, and I will update this issue after our discussion.

MathisHammel:
Implemented in #5213 :)

PR #5213, opened 2025-08-05
ForceAtlas2 - support for prevent_overlapping and mobility

MathisHammel:
Rolled into:  #5260

--------------

This PR adds two new features to ForceAtlas2:

## prevent_overlapping

This allows specifying the radius of each vertex, taken into account when computing attraction and repulsion forces. The implementation follows the specification from the [original ForceAtlas2 paper](https://medialab.sciencespo.fr/publications/Jacomy_Heymann_Venturini-Force_Atlas2.pdf).

The parameter was already present and documented in cuGraph, but raised a `NotImplementedError` when enabled.

## mobility

It can sometimes be useful to tweak the speed of each node individually, e.g. to freeze the position of large nodes while the smaller ones take more time to settle.

The `mobility` parameter is a DataFrame giving each vertex a "mobility factor", which is multiplied with its speed when node positions are updated. This makes it possible to lock a subset of nodes by setting their mobility to 0.0 while the rest stay at 1.0.

The choice of making mobility a float and not a boolean gives finer control on the speed of each node, for cases where it is desirable to slow down or speed up their movement based on specific constraints.

----

closes https://github.com/rapidsai/cugraph/issues/4680


ChuckHastings:
Thanks for the pull request.  Will review this week.

ChuckHastings (approved):
C/C++ code looks good to me.  There are some formatting issues.  You should be able to use `pre-commit` to correct these.

Thanks for the new feature!

jnke2016 (approved):

MathisHammel:
Thanks for the heads up, I pushed the correct style using `pre-commit`.

MathisHammel:
Sorry to bother again with beginner issues, could someone walk me through why the [build](https://github.com/rapidsai/cugraph/actions/runs/16834759280/job/47697365972) of this branch is failing?

I understand it has something to do with mismatched column types on [the tests I added](https://github.com/MathisHammel/cugraph/blob/fa2_overlap/python/cugraph/cugraph/tests/layout/test_force_atlas2.py#L224) (and it only happens on the `dining_prefs` dataset, the only one where node ids are strings and not ints). It runs without issue on my local build and the [error](https://github.com/rapidsai/cugraph/actions/runs/16834759280/job/47697365972#step:10:7291) is raised from `cudf`, which I'm not super familiar with.

The way I use `input_graph.add_internal_vertex_id(` on the `vertex_radius` and `mobility` DataFrames was copied from the [existing implementation](https://github.com/MathisHammel/cugraph/blob/fa2_overlap/python/cugraph/cugraph/layout/force_atlas2.py#L151) for `pos_list`, which makes me think the bug is rather in the implementation of my tests (ironic!)

rlratzel:
/ok to test [d5494ce](https://github.com/rapidsai/cugraph/pull/5213/commits/d5494ce456a66e55fb0ddf7194c4982bcfb606cf)

rlratzel:
Thanks @MathisHammel for your PR. We really appreciate it.

I spent some time looking at the failure, and while I don't have a fully working set of changes locally, I did notice that there may be an issue with how input data seems to be assumed to be renumbered in some places, and unrenumbered in others.  For example, if I change the test code of the failing test to create `vertex_radius` with unrenumbered vertices (which I believe is the expectation users will have, especially for a user with a graph like dining prefs), I can get the failing test to pass. Here's the code I changed in the test:
```python
    #vertex_radius = cudf.DataFrame(
    #    {
    #        "vertex": G.nodes() if G.is_renumbered() else range(G.nodes().max() + 1),
    #        "radius": radius,
    #    }
    #)
    vertex_radius = cudf.DataFrame(
        {
            "vertex": G.extract_vertex_list(return_unrenumbered_vertices=True),
            "radius": radius,
        }
    )
```
However, other tests then fail due to (I think) assumptions about the vertices in `vertex_radius` being contiguous starting from zero, such as this code:
```python
    if prevent_overlapping:
        ...
        vertex_radius['radius'][vertex_radius['vertex']] = vertex_radius['radius']
```
So I think there's probably a few changes still to be made related to renumbered vs. unrenumbered vertices used in input values like `vertex_radius`. Sorry I don't have anything more specific yet.

> ...and it only happens on the `dining_prefs` dataset, the only one where node ids are strings and not ints). It runs without issue on my local build and the [error](https://github.com/rapidsai/cugraph/actions/runs/16834759280/job/47697365972#step:10:7291) is raised from `cudf`, which I'm not super familiar with.

I'm surprised you have passing tests locally since my local build of your PR branch failed exactly as the CI run did.  Can you verify you're running the same failing test?

> The way I use `input_graph.add_internal_vertex_id(` on the `vertex_radius` and `mobility` DataFrames was copied from the [existing implementation](https://github.com/MathisHammel/cugraph/blob/fa2_overlap/python/cugraph/cugraph/layout/force_atlas2.py#L151) for `pos_list`, which makes me think the bug is rather in the implementation of my tests (ironic!)

I've never looked at the force_atlas2 code in detail before, but I think you're using `input_graph.add_internal_vertex_id` correctly in the sense that you're converting external to internal vertex IDs (which I believe `pos_list` is also expected to use but it looks like we don't test it), but the test was using "internal" vertex IDs instead of "external" in `vertex_radius`. I think the other datasets using numerical vertex IDs may have also had a problem but not one that raised an exception (meaning we may not be checking outputs thoroughly).

rlratzel:
Just wanted to thank you again @MathisHammel for all your work here.  I think we should merge it once the tests are passing.  I do want to mention that the plan will be for us to submit a followup PR after this is merged to take care of a few things slightly outside the scope of what you're trying to do:
1. We have another package that uses `pylibcugraph.force_atlas2`, and since the signature changes in this PR in a non-backwards compatible way, it will break. I'm referring to `nx-cugraph` - [our GPU-accelerated cugraph backend for NetworkX](https://rapids.ai/nx-cugraph/) - and how it calls `plc.force_atlas2` [here](https://github.com/rapidsai/nx-cugraph/blob/20af3d8e06639c0b5efea656f0691421c7d70702/nx_cugraph/drawing/layout.py#L129-L148).  Rather than ask you to make even more contributions (although, you're always welcome to!) we'll take care of it, especially given point 2 below....
2. You undoubtedly noticed that there's two cython implementations of force_atlas2: one in `pylibcugraph` and the other in the `cugraph` package that accesses the C++ interface directly, supports a callback, and is what the higher-level cugraph Python library uses. We want to consolidate those into a single implementation in `pylibcugraph` and update the cugraph Python code to call that one (which is also called by `nx-cugraph` currently), so we'll do that shortly after the PR is merged too. In the process, we'll be removing support for the callback feature since we realized it's no longer used by the users that requested it years ago, it's hard to test, etc.

ChuckHastings:
> ...
>
> I've never looked at the force_atlas2 code in detail before, but I think you're using `input_graph.add_internal_vertex_id` correctly in the sense that you're converting external to internal vertex IDs (which I believe `pos_list` is also expected to use but it looks like we don't test it), but the test was using "internal" vertex IDs instead of "external" in `vertex_radius`. I think the other datasets using numerical vertex IDs may have also had a problem but not one that raised an exception (meaning we may not be checking outputs thoroughly).
>

So, do we want to fix the FA2 python tests as part of this PR, or should we remove `dining_prefs` from the graph list for now so it will pass CI and merge this, and create a separate issue to fix the FA2 python tests.  Seems like the latter approach gets this new feature added and is no worse than before from a test perspective.

rlratzel:
> So, do we want to fix the FA2 python tests as part of this PR, or should we remove dining_prefs from the graph list for now so it will pass CI and merge this, and create a separate issue to fix the FA2 python tests. Seems like the latter approach gets this new feature added and is no worse than before from a test perspective.

I think the tests are uncovering issues in how the code is not handling the new inputs to the function properly, specifically for use cases that require vertex IDs to be renumbered.  @MathisHammel do you have any thoughts?  I may be able to help by pushing updates to the PR to fix this, but my week is looking busy so it might not be soon.

MathisHammel:
I'm happy to remove dining_prefs from my tests then, and the relabel issue will probably block my projects to the point I'll need to fix it later anyway

rlratzel:
Hey @MathisHammel , thanks for the update but @ChuckHastings and I were chatting and we're thinking that we'll want that test left in afterall if it is indeed uncovering a bug in the PR since ideally we don't want to merge something that's broken.  Would you mind re-enabling the test? Sorry for the miscommunication here.

Also, have you had a chance to look at the reason for the failure, and does my analysis above seem correct?  I'll plan on working on this more this week and if I get everything working I'll push an update to your PR.

eriknw:
Hi @MathisHammel, I am going to be taking a closer look at this PR this week and will try to fix the failing tests that @rlratzel and @ChuckHastings said should not be skipped b/c they probably reveal actual issues. Is it okay with you if I push the fixes to this PR, or do you prefer to do it yourself perhaps with guidance? If I come up with a fix, I plan to push it.

Thank you for the amazing work, we are eager to get this in!

MathisHammel:
Hi @eriknw, by all means feel free to push directly to this branch!
Sorry for the lack of updates, my schedule has been quite intense lately so I really appreciate the help.

rlratzel:
I added the `DO NOT MERGE` label to prevent accidental merging now that CI is passing. @eriknw is still working on some updates it needs, see comment [here](https://github.com/rapidsai/cugraph/pull/5213#issuecomment-3245984258).

eriknw:
I don't have fixes yet (although I have several attempted fixes that partially work, and I think there are multiple confounding bugs), but here are a few updates from my investigations

I suspect there are several issues here, and most of them existed before this PR. This PR _may_ have introduced more issues, but I'm not confident either way yet. I read >95% of the PR very carefully, and it looks solid to me and it's not obvious where a bug impacting prior behavior could have been introduced. This PR also closely matches established patterns that may themselves be faulty. For adding a new feature, I think it's appropriate to integrate into and mimic existing code, and, if non-trivial, fixing pre-existing bugs should not block the PR from being merged. So, if we're confident that this PR doesn't introduce any new bugs, I would argue to merge it and make new issues--and PRs if appropriate--for pre-existing bugs.

One of the ways I'm investigating FA2 is to run with the same arguments (such as providing `pos_list`), but with different vertex labels (contiguous integers, non-contiguous integers, and strings) and see if results are the same. Also, to run with cugraph and with pylibcugraph. I am also comparing between this PR and branch-25.10. Another hypothesis that I'm _hoping_ is more-or-less valid is that positions should change very little if the starting positions are already stable. That is, that running FA2 is approximately idempotent when `jitter_tolerance` is small. This seems to be possibly true, but I'm not certain that there aren't "gotchas". I'm curious what @ChuckHastings thinks about this.
<br>
The first experiment here is to establish a baseline. It uses contiguous int vertex IDs with `cugraph.force_atlas`. No worries here, and it looks like our approach of using a stable `pos_list` may work:
<details>
<summary>Contiguous int vertex ids, cugraph.force_atlas2: PR and branch-25.10 both good</summary>

```python
# Hypothesis: force_atlas2 is approx idempotent when start positions are stable.
#             Also, hope to get the same results when only vertex label values or types change.
# Contiguous int vertex ids, cugraph.force_atlas2
# PR 5213 result: looks good!
# branch-25.10 result: looks good!
# Comparison: the same
import cudf
import cugraph

df = cudf.DataFrame({
    'source': [0, 1],
    'destination': [1, 2],
    'weight': [1.0, 2.0],
})
G = cugraph.Graph(directed=False)
G.from_cudf_edgelist(df, edge_attr='weight', symmetrize=True, renumber=False)
pos_list = cudf.DataFrame({
    'vertex': [0, 1, 2],
    'x': [0.02769389934837818, -2.183491687901551e-06, -0.017176803201436996],
    'y': [2.5444934368133545, -0.0001994451304199174, -1.5421199798583984],
})
print(f"{pos_list=}")
fa2 = cugraph.force_atlas2(G, pos_list=pos_list, jitter_tolerance=0.01)
print(f"{fa2=}")

# PR 5213 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       1 -0.000002 -0.000199
2       2 -0.017177 -1.542120
fa2=   vertex         x         y
0       0  0.027712  2.544403
1       1 -0.000002 -0.000201
2       2 -0.017165 -1.543966
"""

# branch-25.10 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       1 -0.000002 -0.000199
2       2 -0.017177 -1.542120
fa2=   vertex         x         y
0       0  0.027712  2.544403
1       1 -0.000002 -0.000201
2       2 -0.017165 -1.543966
"""
```

</details>

The next experiment uses non-contiguous int vertices. This one is very wrong, but at least this PR and branch-25.10 both behave the same. I believe this is one of the issues that @rlratzel pointed out. First, the result of FA2 has additional vertices in the results, which I think is undesirable and wrong (it assumes there are `G.nodes().max() + 1` number of vertices). The `pos_list` needs to include the additional, spurious vertices. The position also changed, so it may be that FA2 runs differently in this case (maybe there are isolated vertices floating around?). I also noticed that `pos_list` gets modified sometimes. This experiment is one of those times. So, lots to fix.

<details>
<summary>Non-contiguous int vertex ids, cugraph.force_atlas2: PR and branch-25.10 both bad, the same</summary>

```python
# Hypothesis: force_atlas2 is approx idempotent when start positions are stable
#             Also, hope to get the same results when only vertex label values or types change.
# Non-contiguous int vertex ids, cugraph.force_atlas2
# PR 5213: incorrect. pos_list needs additional vertex ids. final positions moved. pos_list was modified.
# branch-25.10: same as PR
# Comparison: same
import cudf
import cugraph

df = cudf.DataFrame({
    'source': [0, 2],
    'destination': [2, 4],
    'weight': [1.0, 2.0],
})
G = cugraph.Graph(directed=False)
G.from_cudf_edgelist(df, edge_attr='weight', symmetrize=True, renumber=True)
pos_list = cudf.DataFrame({
    'vertex': [0, 2, 4, 1, 3],
    'x': [0.02769389934837818, -2.183491687901551e-06, -0.017176803201436996, 0, 0],
    'y': [2.5444934368133545, -0.0001994451304199174, -1.5421199798583984, 0, 0],
})
print(f"{pos_list=}")
fa2 = cugraph.force_atlas2(G, pos_list=pos_list, jitter_tolerance=0.01)
print(f"{pos_list=}")
print(f"{fa2=}")

# PR 5213 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       2 -0.000002 -0.000199
2       4 -0.017177 -1.542120
3       1  0.000000  0.000000
4       3  0.000000  0.000000
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       2  0.000000  0.000000
2       4 -0.000002 -0.000199
3       1  0.000000  0.000000
4       3 -0.017177 -1.542120
fa2=   vertex         x         y
0       0  0.221937  3.500401
1       1 -3.040798 -6.254920
2       2  0.109285  0.100606
3       3 -3.227196 -2.101423
4       4  1.576791 -0.791426
"""

# branch-25.10 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       2 -0.000002 -0.000199
2       4 -0.017177 -1.542120
3       1  0.000000  0.000000
4       3  0.000000  0.000000
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       2  0.000000  0.000000
2       4 -0.000002 -0.000199
3       1  0.000000  0.000000
4       3 -0.017177 -1.542120
fa2=   vertex         x         y
0       0  0.221937  3.500401
1       1 -3.040798 -6.254920
2       2  0.109285  0.100606
3       3 -3.227196 -2.101423
4       4  1.576791 -0.791426
"""
```

</details>

The next experiment uses string vertices, so the graph is renumbered. Interestingly, branch-25.10 behaves well (positions don't move), but this PR does not (positions move _a lot_). This is the main signal that suggests to me that this PR may have a bug. However, I don't have any possible explanations for this yet, so it also increases my doubts about this approach. I would like to spend a little bit of time digging deeper to see if this PR is doing anything different here. I think we can probably merge this PR after this is understood and either fixed or determined it's a non-issue.

<details>
<summary>str vertex ids, cugraph.force_atlas2: PR is bad, branch-25.10 is good</summary>

```python
# Hypothesis: force_atlas2 is approx idempotent when start positions are stable
#             Also, hope to get the same results when only vertex label values or types change.
# str vertex ids, cugraph.force_atlas2
# PR 5213 result: very different results
# branch-25.10 result: looks good!
# Comparison: different
import cudf
import cugraph

df = cudf.DataFrame({
    'source': ['A', 'B'],
    'destination': ['B', 'C'],
    'weight': [1.0, 2.0],
})
G = cugraph.Graph(directed=False)
G.from_cudf_edgelist(df, edge_attr='weight', symmetrize=True, renumber=True)
pos_list = cudf.DataFrame({
    'vertex': ["A", "B", "C"],
    'x': [0.02769389934837818, -2.183491687901551e-06, -0.017176803201436996],
    'y': [2.5444934368133545, -0.0001994451304199174, -1.5421199798583984],
})
print(f"{pos_list=}")
fa2 = cugraph.force_atlas2(G, pos_list=pos_list, jitter_tolerance=0.01)
print(f"{fa2=}")

# PR 5213 result
"""
pos_list=  vertex         x         y
0      A  0.027694  2.544493
1      B -0.000002 -0.000199
2      C -0.017177 -1.542120
fa2=          x         y vertex
0 -0.000026 -0.000462      B
1 -0.016360 -1.483836      C
2  0.047882  4.389587      A
"""

# branch-25.10 result
"""
pos_list=  vertex         x         y
0      A  0.027694  2.544493
1      B -0.000002 -0.000199
2      C -0.017177 -1.542120
fa2=          x         y vertex
0 -0.000002 -0.000201      B
1 -0.017165 -1.543966      C
2  0.027712  2.544403      A
"""
```

</details>

Another experiment I tried is to run `pylibcugraph.force_atlas2` on the same graphs with the same exact input. In principle, I think this should be calling the same exact C++/CUDA code with the same data, so it should give the same result. However, that's not what we see. Results are different.

<details>
<summary>Contiguous int vertex ids, plc.force_atlas2: PR and branch-25.10 both bad, differently</summary>

```python
# Hypothesis: force_atlas2 is approx idempotent when start positions are stable
#             Also, hope to get the same results when only vertex label values or types change.
# Contiguous int vertex ids, plc.force_atlas2
# PR 5213 result: position moved significantly
# branch-25.10: moved a little bit, could be significant
# Comparison: different
import cudf
import cugraph
import pylibcugraph as plc

df = cudf.DataFrame({
    'source': [0, 1],
    'destination': [1, 2],
    'weight': [1.0, 2.0],
})
df["source"] = df["source"].astype("int32")
df["destination"] = df["destination"].astype("int32")
df["weight"] = df["weight"].astype("float32")

G = cugraph.Graph(directed=False)
G.from_cudf_edgelist(df, edge_attr='weight', symmetrize=True, renumber=False)
pos_list = cudf.DataFrame({
    'vertex': [0, 1, 2],
    'x': [0.02769389934837818, -2.183491687901551e-06, -0.017176803201436996],
    'y': [2.5444934368133545, -0.0001994451304199174, -1.5421199798583984],
})
print(f"{pos_list=}")
vertices, x_axis, y_axis = plc.force_atlas2(
    random_state=None,
    graph=G._plc_graph,
    max_iter=500,
    x_start=pos_list["x"].astype("float32"),
    y_start=pos_list["y"].astype("float32"),
    outbound_attraction_distribution=True,
    lin_log_mode=False,
    prevent_overlapping=False,
    vertex_radius=None,  # Comment out for branch-25.10
    overlap_scaling_ratio=100.0,  # Comment out for branch-25.10
    edge_weight_influence=1.0,
    jitter_tolerance=0.01,
    barnes_hut_optimize=True,
    barnes_hut_theta=0.5,
    scaling_ratio=2.0,
    strong_gravity_mode=False,
    gravity=1.0,
    mobility=None,  # Comment out for branch-25.10
    verbose=False,
    do_expensive_check=False,
)
fa = cudf.DataFrame({"vertex": vertices, "x": x_axis, "y": y_axis})
print(f"{fa=}")

# PR 5213 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       1 -0.000002 -0.000199
2       2 -0.017177 -1.542120
fa=   vertex         x         y
0       0  0.030644  2.811744
1       1  0.000047  0.000107
2       2 -0.024886 -2.240005
"""

# branch-25.10 result
"""
pos_list=   vertex         x         y
0       0  0.027694  2.544493
1       1 -0.000002 -0.000199
2       2 -0.017177 -1.542120
fa=   vertex         x         y
0       0  0.027325  2.507750
1       1 -0.000010  0.000063
2       2 -0.018884 -1.699808
"""
```

</details>

This last experiment that I'm sharing is similar to the previous one, except I use for the initial positions the results from running `plc.force_atlas2` on this PR. When using that start_pos, the positions change very little (as hoped) on this PR, but they change significantly when run on branch-25.10. I don't know why, and this gives me further doubts.

<details>
<summary>Contiguous int vertex ids, plc.force_atlas2: PR is good, branch-25.10 is bad</summary>

```python
# Hypothesis: force_atlas2 is approx idempotent when start positions are stable
#             Also, hope to get the same results when only vertex label values or types change.
# Note: this uses positions stable-ish when using PLC in PR 5213
# Contiguous int vertex ids, plc.force_atlas2
# PR 5213 result: looks good!
# branch-25.10 result: moved, probably significant
# Comparison: different
import cudf
import cugraph
import pylibcugraph as plc

df = cudf.DataFrame({
    'source': [0, 1],
    'destination': [1, 2],
    'weight': [1.0, 2.0],
})
df["source"] = df["source"].astype("int32")
df["destination"] = df["destination"].astype("int32")
df["weight"] = df["weight"].astype("float32")

G = cugraph.Graph(directed=False)
G.from_cudf_edgelist(df, edge_attr='weight', symmetrize=True, renumber=False)
pos_list = cudf.DataFrame({
    'vertex': [0, 1, 2],
    'x': [-0.04411527141928673, -8.09331177151762e-05, 0.006547126453369856],
    'y': [2.9845144748687744, -0.0008206764468923211, -2.62831449508667],
})
print(f"{pos_list=}")
vertices, x_axis, y_axis = plc.force_atlas2(
    random_state=None,
    graph=G._plc_graph,
    max_iter=500,
    x_start=pos_list["x"].astype("float32"),
    y_start=pos_list["y"].astype("float32"),
    outbound_attraction_distribution=True,
    lin_log_mode=False,
    prevent_overlapping=False,
    vertex_radius=None,  # Comment out for branch-25.10
    overlap_scaling_ratio=100.0,  # Comment out for branch-25.10
    edge_weight_influence=1.0,
    jitter_tolerance=0.1,
    barnes_hut_optimize=True,
    barnes_hut_theta=0.5,
    scaling_ratio=2.0,
    strong_gravity_mode=False,
    gravity=1.0,
    mobility=None,  # Comment out for branch-25.10
    verbose=False,
    do_expensive_check=False,
)
fa = cudf.DataFrame({"vertex": vertices, "x": x_axis, "y": y_axis})
print(f"{fa=}")

# PR 5213 result
"""
pos_list=   vertex         x         y
0       0 -0.044115  2.984514
1       1 -0.000081 -0.000821
2       2  0.006547 -2.628314
fa=   vertex         x         y
0       0 -0.041424  2.983171
1       1 -0.000047 -0.000032
2       2  0.009416 -2.633636
"""

# branch-25.10 result
"""
pos_list=   vertex         x         y
0       0 -0.044115  2.984514
1       1 -0.000081 -0.000821
2       2  0.006547 -2.628314
fa=   vertex         x         y
0       0 -0.034180  2.530714
1       1  0.000151  0.000745
2       2  0.008329 -1.773781
"""
```

</details>

I'll share more experiments that I find useful (and want reviewed for sanity)--and possible fixes that I have isolated--once I have them distilled into a something comprehensible.

eriknw:
@rlratzel and @jnke2016 discussed this today, and we think it may be worth trying to update `cugraph.force_atlas2` to use `pylibcugraph.force_atlas2` instead of the Cython wrapper. I am going to work on this. Since `pylibcugraph.force_atlas2` doesn't have `callback` argument, this will mean removing `callback` from `cugraph.force_atlas2` (at least for the next step).

ChuckHastings:
@MathisHammel - your changes were combined into #5260 and merged today.

Thanks for your contribution!  Please let us know how the updated code works for you!

PR #5230, opened 2025-08-15
Raise wheel size threshold for libcugraph

ChuckHastings:
RAPIDS added building for more GPU architectures, this increased our library size by about 10%, raising threshold accordingly.

We are generally working toward decreasing the library size, but this will take a while.


jameslamb (approved):
> RAPIDS added building for more GPU architectures, this increased our library size by about 10%, raising threshold accordingly.

Linking the related change: https://github.com/rapidsai/rapids-cmake/pull/897

PR #5232, opened 2025-08-16
Temporal Sampling C API

### LINKED ISSUE #5219: Update sampling post-processing C++ code to support temporal edge properties
With temporal edge properties available in the output of C++ sampling, we need to be able to do the post-processing (sorting as appropriate) with the edge properties as well.

### LINKED ISSUE #5220: Add temporal sampling to C API
Temporal sampling is currently available via C++.  Make this available to the C API.  Needs #5219 to complete.+

ChuckHastings:
Add C API for temporal sampling.

This PR currently only has C API tests for SG uniform temporal sampling.

Closes #5220
Closes #5219


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -1299,29 +1277,28 @@ renumber_and_sort_sampled_edgelist(
...
 template <typename vertex_t,
-          typename weight_t,
           typename edge_id_t,
           typename edge_type_t>
 std::tuple<
   rmm::device_uvector<vertex_t>,                  // srcs
   rmm::device_uvector<vertex_t>,                  // dsts
-  std::optional<rmm::device_uvector<weight_t>>,   // weights
+  std::vector<arithmetic_device_uvector_t>,       // edge properties
   std::optional<rmm::device_uvector<edge_id_t>>,  // edge IDs
```
seunghwak, line 1297: Do we still need edge IDs? (or can't it be returned as edge properties?)

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -1333,7 +1310,7 @@ heterogeneous_renumber_and_sort_sampled_edgelist(
   raft::handle_t const& handle,
   rmm::device_uvector<vertex_t>&& edgelist_srcs,
   rmm::device_uvector<vertex_t>&& edgelist_dsts,
-  std::optional<rmm::device_uvector<weight_t>>&& edgelist_weights,
+  std::vector<arithmetic_device_uvector_t>&& edgelist_edge_properties,
   std::optional<rmm::device_uvector<edge_id_t>>&& edgelist_edge_ids,
   std::optional<rmm::device_uvector<edge_type_t>>&& edgelist_edge_types,
```
seunghwak, line 1310: Do we still need edgelist_edge_ids & edgelist_edge_types here? Aren't they edgelist_edge_properties?

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: They need to be separate in input in order to renumber the edge ids, which occurs inside this function.  It made sense (to me) to leave them as a separate output parameters also.

I could put them into the input vector `edgelist_edge_properties`  and return them in the output vector and have an optional index indicating which element of the vector was the edge id for renumbering (and the same with the edge type).  But we discussed that in a previous PR and I thought we decided it was cleaner to keep them separate.

I could make them arithmetic_device_uvector_t instances to reduce the number of instantiated variations.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: As mentioned above, if we put them in edgelist_edge_properties we would need to know their indices.  At the moment I'm keeping them separate in the calling functions and constructing the edgelist_edge_properties on demand (although I ultimately want to avoid that also).

It might make sense to put them together and pass optional indices.  Perhaps this example make the motivation for that approach more clear than the last PR?  I'm fine either way.

cpp/include/cugraph/sampling_functions.hpp
seunghwak: I have thought more about this and what about
```
edgelist_edge_ids,
edgelist_edge_types,
edgelist_edge_(extra|additional|...)_properties
```
?

Will this better communicate that ID, types are also edge properties but they are separate for a reason?

cpp/src/sampling/sampling_post_processing_impl.cuh
```diff
@@ -158,18 +159,12 @@ struct optionally_compute_label_index_t {
...
-          typename edge_type_t>
+template <typename label_index_t, typename vertex_t, typename vertex_type_t>
 void check_input_edges(raft::handle_t const& handle,
                        rmm::device_uvector<vertex_t> const& edgelist_majors,
                        rmm::device_uvector<vertex_t> const& edgelist_minors,
-                       std::optional<rmm::device_uvector<weight_t>> const& edgelist_weights,
-                       std::optional<rmm::device_uvector<edge_id_t>> const& edgelist_edge_ids,
-                       std::optional<rmm::device_uvector<edge_type_t>> const& edgelist_edge_types,
+                       std::vector<arithmetic_device_uvector_t> const& edgelist_edge_properties,
+                       std::optional<const_arithmetic_device_span_t> edgelist_edge_types,
```
seunghwak, line 167: If we decide to use edgelist_edge_(extra|additional|..)_properties naming convention, we need to update here (and other places) as well for consistency.

seunghwak (approved):
LGTM

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Seems like a better name.  Renamed to `edgelist_edge_additional_properties`

PR #5238, opened 2025-08-25
Move from UCX-Py to UCXX

pentschev:
UCX-Py is deprecated and being archived. This change moves all UCX-Py code, dependencies and docs to UCXX. See https://github.com/rapidsai/ucx-py/issues/1160 for details.


BradReesWork (approved):

pentschev:
Thanks everyone!

PR #5236, opened 2025-08-22
WIP: Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* adds CUDA 13 devcontainers
* moves some dependency pins:
  - `cuda-python`: `>=12.9.2` (CUDA 12), `>=13.0.1` (CUDA 13)
  - `cupy`: `>=13.6.0`

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

PR #5244, opened 2025-09-02
fix pytorch requirement for CUDA 13 builds

jameslamb:
Follow-up to #5236

There, I intended to introduce a dependency on `torch` nightly wheels for CUDA 13, but made a typo in the `dependencies.yaml` filter. This fixes that, so that wheels for the correct `torch` variant + version will be installed in wheel-testing CI jobs here.

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -582,9 +582,9 @@ dependencies:
           - matrix:
               cuda: "13.*"
             packages:
-              # TODO: remove this 'a0' (which allows nightlies) once there are pytorch CUDA 13 released packages
+              # TODO: remove this '.dev0' (which allows nightlies) once there are pytorch CUDA 13 released packages
               #  ref: https://github.com/pytorch/pytorch/issues/159779
-              - &pytorch_pip torch>=2.9.0a0
+              - &pytorch_pip torch>=2.9.0.dev0
               - &tensordict tensordict>=0.1.2
           - {matrix: null, packages: [*pytorch_pip, *tensordict]}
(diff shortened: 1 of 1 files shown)


dependencies.yaml
```diff
@@ -580,11 +580,11 @@ dependencies:
...
               - tensordict>=0.1.2
           - matrix:
-              cuda: "12.*"
+              cuda: "13.*"
             packages:
-              # TODO: remove this 'a0' (which allows nightlies) once there are pytorch CUDA 13 released packages
+              # TODO: remove this '.dev0' (which allows nightlies) once there are pytorch CUDA 13 released packages
               #  ref: https://github.com/pytorch/pytorch/issues/159779
-              - &pytorch_pip torch>=2.9.0a0
+              - &pytorch_pip torch>=2.9.0.dev0
```
jameslamb, line 587: Guess this needs to be `.dev` instead of `a0`.

> ERROR: Could not find a version that satisfies the requirement torch>=2.9.0a0 (from versions: 2.6.0, 2.7.0, 2.7.1, 2.8.0, 2.9.0.dev20250828+cu130, 2.9.0.dev20250829+cu130, 2.9.0.dev20250830+cu130, 2.9.0.dev20250831+cu130, 2.9.0.dev20250901+cu130, 2.9.0.dev20250902+cu130)

([pip devcontainer build link](https://github.com/rapidsai/cugraph/actions/runs/17417959519/job/49450499051?pr=5244#step:11:1834))

gforsyth (approved):

PR #5245, opened 2025-09-04
check that optional is set before resizing

### LINKED ISSUE #5155: [BUG]: Attempt to resize() an object in a std::optional variable that has not been initialized
### Version

25.02

_No response_

### Describe the bug.

Note this is the same in 25.08.
In the file cpp/tests/sampling/detail/sampling_post_processing_validate.cu, There is a a resize() operation a line 1350:
```
(*this_type_unique_minor_hops)
            .resize(this_type_unique_minors.size(), handle.get_stream());
```
The problem here is that when running the test SAMPLING_HETEROGENEOUS_POST_PROCESSING_TEST, the first pass through the loop the std::optional object this_type_unique_minor hops is initialized to std::nullopt (as is this_type_unique_major_hops). When this happens, the attemped resize of the underlying device_uvector seg faults because of the lack of an previously allocated device_uvector. The fix is something like

```
          if (this_type_unique_minor_hops) {
            .resize(this_type_unique_minors.size(), handle.get_stream());
            (*this_type_unique_minor_hops)
              .resize(this_type_unique_minors.size(), handle.get_stream());
          }
```

But probably better to be explicit

```
          if ((*this_type_unique_minor_hops).has_value()) {
            .resize(this_type_unique_minors.size(), handle.get_stream());
            (*this_type_unique_minor_hops)
              .resize(this_type_unique_minors.size(), handle.get_stream());
          }
```

```shell

```

### Relevant log output

```shell

```

### Environment details

```shell

```

### Other/Misc.

_No response_

### Code of Conduct

- [x] I have searched the [open bugs](https://github.com/rapidsai/cugraph/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

ChuckHastings:
Closes #5155

### DIFF
--- a/cpp/tests/sampling/detail/sampling_post_processing_validate.cu
+++ b/cpp/tests/sampling/detail/sampling_post_processing_validate.cu
@@ -1079,7 +1079,10 @@ bool check_vertex_renumber_map_invariants(
                                                   this_label_unique_minors.end(),
                                                   (*this_label_unique_minor_hops).begin()))),
         handle.get_stream());
-      (*this_label_unique_minor_hops).resize(this_label_unique_minors.size(), handle.get_stream());
+      if (this_label_unique_minor_hops) {
+        (*this_label_unique_minor_hops)
+          .resize(this_label_unique_minors.size(), handle.get_stream());
+      }
     } else {
       thrust::sort(handle.get_thrust_policy(),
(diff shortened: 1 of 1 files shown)


seunghwak (approved):
LGTM

PR #5248, opened 2025-09-04
Update create graph to use arithmetic types

### LINKED ISSUE #5239: Refactor Graph Creation to use variant types
Modify the graph creation functions to pass a vector of variant types for creating edge properties and return a vector of variant types of the edge_property objects.

This will reduce the number of variations of graph construction functions that are required, reducing compile time and memory utilization during compilation as well as reducing the size of libcugraph.so.

ChuckHastings:
This PR modifies create_graph_from_edgelist to use the new variant arithmetic types as input.

There's also a small change to validation of sampling results because the edge weight was not used in one step but was being passed through the validation functions.

This is part of the effort to reduce library size and compile time.

Change is breaking to the C++ API, the C API and PLC API are preserved.

Closes #5239


cpp/CMakeLists.txt
```diff
@@ -400,18 +400,10 @@ set(CUGRAPH_SOURCES
...
     src/components/vertex_coloring_mg_v64_e64.cu
     src/components/vertex_coloring_mg_v32_e32.cu
-    src/structure/create_graph_from_edgelist_sg_v64_e64.cu
-    src/structure/create_graph_from_edgelist_sg_v32_e32.cu
-    src/structure/create_graph_from_edgelist_mg_v64_e64.cu
-    src/structure/create_graph_from_edgelist_mg_v32_e32.cu
     src/structure/create_graph_from_edgelist_sg_v64_e64_t32.cu
     src/structure/create_graph_from_edgelist_sg_v32_e32_t32.cu
     src/structure/create_graph_from_edgelist_mg_v64_e64_t32.cu
     src/structure/create_graph_from_edgelist_mg_v32_e32_t32.cu
```
seunghwak: Should we better rename these files (i.e. remove t32 or t64)?

Time stamps are not just additional edge properties. I guess no reason to include them in the file names while we don't include weights, edge ID, edge types in the file names.

cpp/src/community/detail/refine_impl.cuh
```diff
@@ -642,32 +642,23 @@ refine_clustering(
...
 
     std::optional<rmm::device_uvector<vertex_t>> renumber_map{std::nullopt};
-    std::optional<edge_property_t<edge_t, weight_t>> coarse_edge_weights{std::nullopt};
 
     if constexpr (multi_gpu) {
-      std::vector<cugraph::arithmetic_device_uvector_t> edge_properties{};
-      if (d_weights) edge_properties.push_back(std::move(*d_weights));
-
-      std::tie(d_srcs, d_dsts, edge_properties, std::ignore) =
+      std::tie(d_srcs, d_dsts, std::ignore, std::ignore) =
```
seunghwak, line 637: Why aren't we no longer shuffling d_weights? Is it because we are not using it after shuffle? In that case, why are we still allocating & updating d_weights in the code above?

cpp/CMakeLists.txt
```diff
@@ -400,18 +400,10 @@ set(CUGRAPH_SOURCES
...
-    src/structure/create_graph_from_edgelist_sg_v64_e64_t64.cu
-    src/structure/create_graph_from_edgelist_sg_v32_e32_t64.cu
-    src/structure/create_graph_from_edgelist_mg_v64_e64_t64.cu
-    src/structure/create_graph_from_edgelist_mg_v32_e32_t64.cu
```
seunghwak, line 414: Similar here. Better remove _t64 from the file names?

cpp/include/cugraph/mtmg/graph.hpp
```diff
@@ -101,47 +87,31 @@ void create_graph_from_edgelist(
...
-      : std::nullopt,
-    my_edgelist.get_edge_end_time()
-      ? std::make_optional(std::move((*my_edgelist.get_edge_end_time())[0]))
-      : std::nullopt,
-    graph_properties,
-    renumber,
-    std::nullopt,
-    std::nullopt,
-    do_expensive_check);
+  std::vector<cugraph::arithmetic_device_uvector_t> edge_properties_buffers;
```
seunghwak: Sorry for nitpicking but
```
edge_properties_buffers; => edge_properties_buffers{};
```

cpp/include/cugraph/mtmg/graph_view.hpp
```diff
@@ -26,9 +26,9 @@ namespace mtmg {
 /**
  * @brief Graph view for each GPU
  */
-template <typename vertex_t, typename edge_t, bool store_transposed, bool multi_gpu>
+template <typename vertex_t, bool store_transposed, bool multi_gpu>
 class graph_view_t : public detail::device_shared_wrapper_t<
-                       cugraph::graph_view_t<vertex_t, edge_t, store_transposed, multi_gpu>> {
+                       cugraph::graph_view_t<vertex_t, vertex_t, store_transposed, multi_gpu>> {
```
seunghwak: Why are we removing edge_t from here? We are not explicitly instantiating for vertex_t = int32_t and edge_t = int64_t but we still allow users to compile for this combination. Are we changing anything related to this policy?

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -418,6 +418,79 @@ std::vector<rmm::device_uvector<T>> split_edge_chunk_elements_to_local_edge_part
...
+        }
+        edge_partition_elements.push_back(std::move(edge_partition_buffer));
+      });
+  }
+
+  edgelist_elements.clear();
+
+  return edge_partition_elements;
+}
+
```
seunghwak, line 438: This function is very similar to the function above; the only difference is that the function above takes a std::vector of rmm::device_uvector while this function takes std::vector of arithemtic_device_uvector.

Should we keep both?

Can we call the function above using variant_type_dispatch or delete the function above just calling this? I feel like keeping both is a sort of code bloat.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -1451,11 +1114,14 @@ create_graph_from_edgelist_impl(
...
-      if (edgelist_edge_types) { element_size += sizeof(edge_type_t); }
-      if (edgelist_edge_start_times) { element_size += sizeof(edge_time_t); }
-      if (edgelist_edge_end_times) { element_size += sizeof(edge_time_t); }
+
+      if (edgelist_edge_properties.size() == 1) {
+        element_size +=
+          variant_type_dispatch(edgelist_edge_properties[0][0], sizeof_arithmetic_element{});
+      } else {
+        element_size += sizeof(edge_t);
+      }
```
seunghwak: I don't think this code is correct. This compares the total edge list data size vs global memory size. So, even when edgelist_edge_properties.size() > 1, we need to add the size of every edge property.

cpp/CMakeLists.txt
ChuckHastings: Changed

cpp/CMakeLists.txt
ChuckHastings: These files were deleted... so take care of :-)

cpp/include/cugraph/mtmg/graph.hpp
ChuckHastings: Fixed

cpp/include/cugraph/mtmg/graph_view.hpp
ChuckHastings: I'll change that back.

cpp/src/community/detail/refine_impl.cuh
ChuckHastings: The weight is used to filter positive gain on line 623 above, which is before the shuffling.  We never use the weights again in the refinement phase, so there's no need to shuffle them.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: Merged the functions with some constexpr blocks to handle the differences.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: I'll change this to a reduce to match the old logic.

Do we really need to fit `num_edges * total_element_size` where `total_element_size` is the sum of the sizes of both vertices and each property comfortably in the available memory?  Aren't we doing them one array at a time?  First the srcs, then dsts, then each property?  It feels like this might be overestimating the memory usage.

jnke2016 (approved):

cpp/src/community/detail/refine_impl.cuh
seunghwak: OK, so you write to d_weights but you don't need to use d_weights. We are creating d_weights just as edge_begin() and edge_end() have gains as well.

We can avoid creating and writing to d_weights by using thrust::copy_if that takes a stencil iterator.

See
```
template<typename DerivedPolicy, typename InputIterator1, typename InputIterator2, typename OutputIterator, typename Predicate>
OutputIterator thrust::copy_if(const thrust::detail::execution_policy_base &exec,
                                InputIterator1 first, InputIterator1 last,
                                InputIterator2 stencil,
                                OutputIterator result,
                                Predicate pred);
```
```
auto d_src_dst_iterator =
      thrust::make_zip_iterator(d_srcs.begin(), d_dsts.begin());
auto edge_begin = thrust::make_zip_iterator(vertex_begin,
                                cuda::std::get<1>(gain_and_dst_first.get_iterator_tuple()));
auto edge_end =
      thrust::make_zip_iterator(vertex_end,
                                cuda::std::get<1>(gain_and_dst_last.get_iterator_tuple()));
thrust::copy_if(handle.get_thrust_policy(), edge_begin, edge_end, gain_and_dst_first, d_src_dst_iterator, []__device__(auto pair) {
  auto gain = cuda::std::get<0>(pair);
  auto dsat = cuda::std::get<1>(pair);
  return (gain > POSITIVE_GAIN) && (dst >= 0); });
```

Also, no need to resize d_srcs & d_dsts after copy_if???

cpp/src/structure/create_graph_from_edgelist_impl.cuh
seunghwak: Are we using the original function anywhere? Can we just delete the first one?

cpp/src/community/detail/refine_impl.cuh
ChuckHastings: This is all pre-existing code... but yes, I'll fix.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
seunghwak: This is mainly to decide whether to compress source & destination vertices or not (mainly introduced to run largest possible R-mat scale in Graph 500).

And we can take graph edges in multiple chunks (e.g. think about edge data stored in multiple files).

So, the first thing is we need to re-organize edges in multiple chunks to edges in multiple edge partitions. During this transition we will use temporary memory either in thrust calls or in some places, and there are some places we need to keep both (data in multiple chunks and data in edge partitions).

\# edges * (data size per edge) is basically to estimate the current memory usage at the beginning. So, if we are  using more than threshold % of the total global memory, so the remaining memory is less than the total GPU memory * (1-threshold), we try to take a slower but more memory frugal path.

This is the rationale behind this logic, and in this context, I think deciding based on the memory footprint of the total edge list is more appropriate.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: I see.  I missed that there were 3 functions in the original PR that were similar.  I combined 2 of them... but not the 2 you were thinking of.

I combined the `rmm::device_uvector<T>` and `rmm::device_uvector<std::byte>` versions into the same function and just added constexpr blocks to special case the logic when the `T` is `std::byte`.

Combining the versions of `rmm::device_uvector<T>` and `arithmetic_device_uvector_t` is a bit more complicated, because the compiler won't just convert `std::vector<rmm::device_uvector<T>>` into `std::vector<arithmetic_device_uvector_t>`.  But I think I see how to combine those together as well.

seunghwak: Yeah... sorry for the ambiguity.

I didn't mean to combine the one working on compressed vertices (one taking std::byte) and the one working on uncompressed vertices and edge properties.

We now have two versions working on uncompressed values. One taking rmm::device_uvector and the other taking arithmetic_device_uvector.

My first question is whether we are using the one taking rmm::device_uvector at all; you updated this code to use arithmetic device_uvector. If we are not using the first version at all, we can just simply delete the first one and just keep the second one.


So, are we using the first one in anyplace?

seunghwak: Is it because we are calling the one taking rmm::device_uvector for sources and destinations and using the arithmetic_device_uvector version for edge properties?

In this case, for uncompressed sources/destinations, we can just move the sources and destinations and store them in arithmetic_device_uvector objects and call the second version. So, we can delete the first version.

This is much less increase in code size, compile time, and binary size than replicating two nearly identical functions and explicitly instantiate both for all different type combinations.

ChuckHastings: I just pushed an update with a single function that addresses compressed, uncompressed vertices and edge properties.  All 3 functions were very similar in structure, now there's only one function.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -1684,82 +1278,71 @@ create_graph_from_edgelist_impl(
...
-        edge_partition_edge_counts,
-        edge_partition_intra_partition_segment_offset_vectors,
-        edge_partition_intra_segment_copy_output_displacement_vectors,
-        large_edge_buffer_type);
+    std::vector<arithmetic_device_uvector_t> tmp_edgelist_srcs{};
+    tmp_edgelist_srcs.reserve(edgelist_srcs.size());
+    std::for_each(
+      edgelist_srcs.begin(), edgelist_srcs.end(), [&tmp_edgelist_srcs](auto&& edgelist_src) {
+        tmp_edgelist_srcs.push_back(std::move(edgelist_src));
+      });
```
seunghwak, line 1286: A minor suggestion, may better call edgelist_srcs.clear(); (and possibly shrink_to_fit as well) to prevent edgelist_srcs holding moved-out objects. Moved out objects have valid but unknown states... I guess it is better to clear() so put them in a known state.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
```diff
@@ -1684,82 +1278,71 @@ create_graph_from_edgelist_impl(
...
-        edge_partition_intra_partition_segment_offset_vectors,
-        edge_partition_intra_segment_copy_output_displacement_vectors,
-        large_edge_buffer_type);
-  }
+    std::for_each(tmp_edgelist_srcs.begin(),
+                  tmp_edgelist_srcs.end(),
+                  [&edge_partition_edgelist_srcs](auto&& edgelist_src) {
+                    edge_partition_edgelist_srcs.push_back(
+                      std::move(std::get<rmm::device_uvector<vertex_t>>(edgelist_src)));
+                  });
```
seunghwak, line 1304: Similar here, tmp_edgelist_srcs.clear(); And we may better do this for destinations as well.

seunghwak (approved):
LGTM besides minor complaints.

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: Changed

cpp/src/structure/create_graph_from_edgelist_impl.cuh
ChuckHastings: Changed

PR #5260, opened 2025-09-18
ForceAtlas2: continue #5213 for `prevent_overlapping` and `mobility`

eriknw:
This PR is a continuation of #5213 by @MathisHammel and aims to address issues that were identified in that PR and pay off some technical debt.

Some changes:
- Add support for passing vertex ids for initial positions, vertex_radius, and mobility.
- Add support for passing random_state.
- Add support for int64_t vertex/edge types.
- Remove support for `callback=` argument in `cugraph.force_atlas2` (it now gives informative error).
- Remove `cugraph/layout/force_atlas2_wrapper.pyx` (use PLC instead).
- Fix mutating input arrays (we now copy and use `const` in C++).
- Fix inputs and result when integral vertex ids weren't contiguous.
  - For example, if vertices are `[0, 2]`, inputs and result would contain values for `1`.
- Fixing renumbering of vertices.

This is getting close, only a couple failing tests to investigate, and I should add a new test or two.


cpp/include/cugraph/algorithms.hpp
```diff
@@ -192,7 +192,8 @@ void overlap_list(legacy::GraphCSRView<VT, ET, WT> const& graph,
  * or double.
  *
  * @param[in] handle                            Library handle (RAFT). If a communicator is set in
- * the handle, the multi GPU version will be selected.
+ * @param[in] rng_state                         The RngState instance holding pseudo-random number
+ * generator state. the handle, the multi GPU version will be selected.
```
ChuckHastings: This comment looks garbled.  The `the handle, the multi GPU version will be selected` is a continuation of the handle parameter comment, the `generator state.` is the second part of the `rng_state` comment.

cpp/include/cugraph_c/layout_algorithms.h
```diff
@@ -47,21 +47,28 @@ typedef struct {
...
  *                               are selected randomly.
- *                               y_start is sorted by vertex id
  * @param [in]   outbound_attraction_distribution
  *                               Distributes attraction along outbound edges
  *                               Hubs attract less and thus are pushed to the borders.
  * @param [in]   lin_log_mode    Switch Force Atlas model from lin-lin to lin-log.
  *                               Makes clusters more tight.
  * @param [in]   prevent_overlapping
  *                               Prevent nodes to overlap.
+ * @param [in]   vertex_radius_vertices
```
ChuckHastings, line 65: Do we want to make these separate from `start_vertices`?

Is there a use case where you specify x/y values and don't specify a vertex radius value, or vice verse?  Same question for mobility values.

I guess I'm less concerned with things that are O(number_of_vertices), but it does cost us GPU memory, and since they're managed by python the memory will be used for the duration of the call.

cpp/src/layout/legacy/barnes_hut.cuh
```diff
@@ -93,17 +96,19 @@ void barnes_hut(raft::handle_t const& handle,
...
-  thrust::fill(handle.get_thrust_policy(), d_massl.begin(), d_massl.end(), 1);
+  rmm::device_uvector<edge_t> d_massl_et(nnodes + 1, stream_view);
+  thrust::fill(handle.get_thrust_policy(), d_massl_et.begin(), d_massl_et.end(), 1);
 
   rmm::device_uvector<float> d_maxxl(blocks * FACTOR1, stream_view);
   rmm::device_uvector<float> d_maxyl(blocks * FACTOR1, stream_view);
   rmm::device_uvector<float> d_minxl(blocks * FACTOR1, stream_view);
   rmm::device_uvector<float> d_minyl(blocks * FACTOR1, stream_view);
 
   // Actual mallocs
```
ChuckHastings: This old comment is not correct anymore.  The allocation happens in the RAII container construction above.  I'd suggest just deleting the comment.

cpp/src/layout/legacy/barnes_hut.cuh
```diff
@@ -162,9 +165,19 @@ void barnes_hut(raft::handle_t const& handle,
   sort(graph, stream_view.value());
   RAFT_CHECK_CUDA(stream_view.value());
 
-  graph.degree(massl, cugraph::legacy::DegreeDirection::OUT);
+  graph.degree(massl_et, cugraph::legacy::DegreeDirection::OUT);
   RAFT_CHECK_CUDA(stream_view.value());
 
+  // Cast massl from edge_t to int.
```
ChuckHastings: We should probably look at updating the functions to use `edge_t`.  While it is uncommon, it is possible for an edge in a 64-bit graph to have > 2^31-1 neighbors, in which case we'll generate an incorrect answer by down casting this into a 32-bit integer.  Not sure the ripple effects here... my first glance is that it seems tractable.

cpp/src/layout/legacy/exact_fa2.cuh
```diff
@@ -96,9 +100,19 @@ void exact_fa2(raft::handle_t const& handle,
   sort(graph, stream_view.value());
   RAFT_CHECK_CUDA(stream_view.value());
 
-  graph.degree(d_mass, cugraph::legacy::DegreeDirection::OUT);
+  graph.degree(d_mass_et, cugraph::legacy::DegreeDirection::OUT);
   RAFT_CHECK_CUDA(stream_view.value());
 
+  // Cast mass from edge_t to int.
+  // ERIK: this is a bandaid since `graph.degree` uses edge_t and other
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -83,16 +106,18 @@ void apply_attraction(const vertex_t* restrict row,
...
+                      bool prevent_overlapping,
+                      const float* restrict vertex_radius,
                       cudaStream_t stream)
 {
   // 0 edge graph.
   if (!e) return;
 
   dim3 nthreads, nblocks;
-  nthreads.x = min(e, CUDA_MAX_KERNEL_THREADS);
+  nthreads.x = min(static_cast<int>(e), CUDA_MAX_KERNEL_THREADS);
```
ChuckHastings: Would this be better as:
`nthreads.x = static_cast<int>(e, CUDA_MAX_KERNEL_THREADS)`?

If e doesn't fit in in a 32-bit integer we'll want to be sure we pick CUDA_MAX_KERNEL_THREADS.

If the above doesn't work, something like this might work:

`nthreads.x = static_cast<int>(e, static_cast<edge_t>(CUDA_MAX_KERNEL_THREADS))`?

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -83,16 +106,18 @@ void apply_attraction(const vertex_t* restrict row,
...
   nthreads.y = 1;
   nthreads.z = 1;
-  nblocks.x  = min((e + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
+  nblocks.x  = min((static_cast<int>(e) + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
```
ChuckHastings: Same concern

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -168,10 +195,10 @@ void apply_gravity(const float* restrict x_pos,
                    cudaStream_t stream)
 {
   dim3 nthreads, nblocks;
-  nthreads.x = min(n, CUDA_MAX_KERNEL_THREADS);
+  nthreads.x = min(static_cast<int>(n), CUDA_MAX_KERNEL_THREADS);
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -168,10 +195,10 @@ void apply_gravity(const float* restrict x_pos,
...
   nthreads.y = 1;
   nthreads.z = 1;
-  nblocks.x  = min((n + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
+  nblocks.x  = min((static_cast<int>(n) + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -222,10 +249,10 @@ void compute_local_speed(const float* restrict repel_x,
                          cudaStream_t stream)
 {
   dim3 nthreads, nblocks;
-  nthreads.x = min(n, CUDA_MAX_KERNEL_THREADS);
+  nthreads.x = min(static_cast<int>(n), CUDA_MAX_KERNEL_THREADS);
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -222,10 +249,10 @@ void compute_local_speed(const float* restrict repel_x,
...
   nthreads.y = 1;
   nthreads.z = 1;
-  nblocks.x  = min((n + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
+  nblocks.x  = min((static_cast<int>(n) + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -306,20 +335,31 @@ void apply_forces(float* restrict x_pos,
...
                   float* restrict old_dy,
                   const float* restrict swinging,
+                  const float* restrict mobility,
                   const float speed,
                   const vertex_t n,
                   cudaStream_t stream)
 {
   dim3 nthreads, nblocks;
-  nthreads.x = min(n, CUDA_MAX_KERNEL_THREADS);
+  nthreads.x = min(static_cast<int>(n), CUDA_MAX_KERNEL_THREADS);
```
ChuckHastings: Same

cpp/src/layout/legacy/fa2_kernels.cuh
```diff
@@ -306,20 +335,31 @@ void apply_forces(float* restrict x_pos,
...
   nthreads.y = 1;
   nthreads.z = 1;
-  nblocks.x  = min((n + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
+  nblocks.x  = min((static_cast<int>(n) + nthreads.x - 1) / nthreads.x, CUDA_MAX_BLOCKS);
```
ChuckHastings: Same

cpp/include/cugraph/algorithms.hpp
eriknw: Good catch! Fixed locally.

cpp/include/cugraph_c/layout_algorithms.h
eriknw: I don't have a strong opinion.

I made them separate because they come from separate dataframes when using cugraph. This also allows each parameter (pos, vertex_radius, and mobility) to more easily evolve independently if users want to improve the API. For example, perhaps it's reasonable to provide a subset of vertices for some of these, and each parameter can decide what to do for values of the missing vertices.

And, yeah, I think it's reasonable to sometimes provide initial x/y or a vertex radius but not both.

Thanks for sharing your thoughts. I'm also not super sensitive about O(V) memory in general, but I do want to be sensitive to cugraph norms and to not be too wasteful of memory.

cpp/src/layout/legacy/barnes_hut.cuh
eriknw: Done.

cpp/src/layout/legacy/barnes_hut.cuh
eriknw: Thanks, working on this now! It's straightforward so far

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -147,9 +175,13 @@ struct force_atlas2_functor : public cugraph::c_api::abstract_functor {
...
-          (number_map != nullptr) ? std::make_optional(raft::device_span<vertex_t const>{
-                                      number_map->data(), number_map->size()})
-                                  : std::nullopt);
+
+          // ERIK: we want srcs and dsts to be renumbered, but providing the number_map
+          //       appears to unrenumber them. Is not passing number_map okay?
+          std::nullopt);  // ERIK:option 1
+      //(number_map != nullptr) ? std::make_optional(raft::device_span<vertex_t const>{
+      //                            number_map->data(), number_map->size()})
+      //                        : std::nullopt);
```
eriknw: It's unclear to me how `number_map` is supposed to be used in `decompress_to_edgelist`. We want the internal ids here, so 0..N-1. I found two ways to do this: don't provide `number_map` to `decompress_to_edgelist`, or renumber using `renumber_ext_vertices`. @ChuckHastings, what is preferred?

I do see that `decompress_to_edgelist` _unrenumbers_ things here:

https://github.com/rapidsai/cugraph/blob/3bde56e23589c074341e064613f4ab41f4daf12d/cpp/src/structure/decompress_to_edgelist_impl.cuh#L296

and here:

https://github.com/rapidsai/cugraph/blob/3bde56e23589c074341e064613f4ab41f4daf12d/cpp/src/structure/decompress_to_edgelist_impl.cuh#L420

python/cugraph/cugraph/layout/force_atlas2.py
```diff
@@ -129,43 +187,109 @@ def on_train_end(self, positions):
...
-            pos_list = input_graph.add_internal_vertex_id(pos_list, "vertex", cols)
+
+        if input_graph.renumbered:
+            pos_list = renumber_vertices(input_graph, pos_list, 2)
+        # Ensure dtypes are valid, warn if we need to cast
+        initial_pos_vertices = ensure_vertex_dtype(
+            input_graph, pos_list["vertex"], 'pos_list["vertex"]'
+        )
+        initial_pos_x = ensure_float32_dtype(pos_list["x"], 'pos_list["x"]')
+        initial_pos_y = ensure_float32_dtype(pos_list["y"], 'pos_list["y"]')
```
eriknw, line 224: What is best practice for checking or converting dtypes lower in the stack? Many inputs for `force_atlas2` must be float32 regardless of edge dtype. For `cugraph.force_atlas2`, I follow existing precedent and convert the inputs to the appropriate type and emit a warning to the user if the type is incorrect.

Should we also check types in PLC and/or C++?

cpp/src/c_api/legacy_fa2.cpp
```diff
@@ -167,48 +219,123 @@ struct force_atlas2_functor : public cugraph::c_api::abstract_functor {
...
+                   start_vertices_->size_,
+                   handle_.get_stream());
         raft::copy(
-          cp_number_map->data(), number_map->data(), number_map->size(), handle_.get_stream());
+          x_start.data(), x_start_->as_type<float>(), x_start_->size_, handle_.get_stream());
+        raft::copy(
+          y_start.data(), y_start_->as_type<float>(), y_start_->size_, handle_.get_stream());
 
-        number_map_pos = rmm::device_uvector<vertex_t>{number_map->size(), handle_.get_stream()};
+        cugraph::renumber_ext_vertices<vertex_t, multi_gpu>(
```
eriknw, line 220: Does it matter whether we use `renumber_ext_vertices` or `renumber_local_ext_vertices` in `force_atlas2`, which is SG-only? What is best practice?

python/cugraph/cugraph/tests/layout/test_force_atlas2.py
```diff
@@ -188,3 +218,136 @@ def test_force_atlas2(graph_file, score, max_iter, barnes_hut_optimize):
...
+            ((pairs["x_1"] - pairs["x_2"]) ** 2 + (pairs["y_1"] - pairs["y_2"]) ** 2)
+            < (2 * radius) ** 2
+        ]
+        return len(overlaps)
+
+    cu_M = graph_file.get_edgelist(download=True)
+    G = graph_file.get_graph()
+    vertex_radius = cudf.DataFrame(
+        {
+            "vertex": G.extract_vertex_list(return_unrenumbered_vertices=True),
```
eriknw, line 251: To get `dining_prefs` dataset to work, I needed to change `"vertex"` from `G.nodes()` to the above. Is there a better way?

I was surprised `G.nodes()` returned internal vertex IDs. Can somebody explain the intuition behind this? I want to better understand the cugraph API, object model, and expectations and best practices. Why does `G.edges()` return external vertex IDs but `G.nodes()` returns internal?

eriknw:
This PR is ready for review. I left a few questions inline.

I plan to update nx-cugraph next, and maybe add another test or two here. Are there any other changes we should make while I'm familiar with the code for FA2? Some possibilities:
- Re-add support for the `callback` parameter (by adding it to PLC)
- Allow `vertex_radius_vertices` (as used by PLC) to be a subset of vertices; unspecified can default to 0
- Allow `mobility_vertices` (as used by PLC) to be a subset of vertices; unspecified can default to 1
- Allow `start_vertices` (as used by PLC) to be a subset of vertices; unsure what to do for unspecified--random?

cpp/src/c_api/legacy_fa2.cpp
ChuckHastings: Don't provide number_map to `decompress_to_edgelist` is the behavior you want.  There's no point in unrenumbering so that you can then renumber again :-)

python/cugraph/cugraph/layout/force_atlas2.py
ChuckHastings: Yes, I notice a bug in the original C API code where it checks to make sure that the types of xstart and ystart match, but then assumes that they are of type `float`.

I would modify the CAPI_EXPECTS call to verify that all of the parameters are of type `float`.'

FYI - edge dtype is either int32_t or int64_t and must match vertex dtype.  The weight would be the only floating point type in the legacy graph structure.  Does FA2 use weights?  If so, it would probably be necessary to make sure that weight_t is also `float`... although if that were the case I would expect you would have hit a compile error.

cpp/src/c_api/legacy_fa2.cpp
ChuckHastings: SG-only shouldn't matter, I tend to use `renumber_ext_vertices` in SG code, because there's really no concept of a local vertex in SG.

In MG the `renumber_local_ext_vertices` is more efficient because you are asserting that all of the vertices you are renumbering are local to this GPU and you don't have to do any shuffling of data.  But sometimes the results are partitioned on the GPU where you can't guarantee that, so if you call `renumber_ext_vertices` it will shuffle the necessary data to the right GPU to get the renumbering done properly.

python/cugraph/cugraph/tests/layout/test_force_atlas2.py
ChuckHastings: That sounds like a bug to me.  The intention in the C API is to always communicate with external vertex IDs.  Leaking internal ids into python is a bad plan, because there really should be no way to use them.

Although I'd just create a ticket and we can fix that separately.

ChuckHastings:
> This PR is ready for review. I left a few questions inline.
> 
> I plan to update nx-cugraph next, and maybe add another test or two here. Are there any other changes we should make while I'm familiar with the code for FA2? Some possibilities:
> 
> * Re-add support for the `callback` parameter (by adding it to PLC)
> * Allow `vertex_radius_vertices` (as used by PLC) to be a subset of vertices; unspecified can default to 0
> * Allow `mobility_vertices` (as used by PLC) to be a subset of vertices; unspecified can default to 1
> * Allow `start_vertices` (as used by PLC) to be a subset of vertices; unsure what to do for unspecified--random?

I'm unaware of any other FA2 user requests.

ChuckHastings (commented):
I'll approve C++ changes after you remove the alternative options that I commented on.

cpp/src/c_api/legacy_fa2.cpp
eriknw: Perfect, that's what I was hoping for 👍 . Removing the alternative and unnecessary comments

cpp/src/c_api/legacy_fa2.cpp
eriknw: Great, thanks for explaining 👍

python/cugraph/cugraph/tests/layout/test_force_atlas2.py
eriknw: Gotcha, that's what makes sense to me too. I created #5280 to capture this.

python/cugraph/cugraph/layout/force_atlas2.py
```diff
@@ -23,15 +64,20 @@ def force_atlas2(
...
     barnes_hut_optimize=True,
     barnes_hut_theta=0.5,
     scaling_ratio=2.0,
     strong_gravity_mode=False,
     gravity=1.0,
+    mobility=None,
     verbose=False,
     callback=None,
+    *,
+    random_state=None,
```
rlratzel, line 80: Since we're breaking position-only arg passing starting at `vertex_radius`, perhaps we should move the `*` up the list? I'm not opposed to breaking position-only a bit further in the interest of improving things more long-term and moving `*` to right after `input_graph`, making `input_graph` the only non-kwarg arg.  For this function, I can't imagine anyone using it position-only for anything other than `input_graph` anyway...  Thoughts?

python/cugraph/cugraph/layout/force_atlas2.py
```diff
@@ -129,43 +187,109 @@ def on_train_end(self, positions):
...
+    mobility_vertices = None
+    mobility_values = None
+    do_expensive_check = False
+
     if pos_list is not None:
         if not isinstance(pos_list, cudf.DataFrame):
             raise TypeError("pos_list should be a cudf.DataFrame")
-        if set(pos_list.columns) != set(["x", "y", "vertex"]):
+
+        if set(pos_list.columns) != {"x", "y", "vertex"}:
```
rlratzel: Now that we have additional args taking DataFrames all describing attributes for vertices, as a user I might expect to be able to create a single DataFrame that includes `vertex`, `x`, `y`, `mobility`, `radius`, etc. specifically for layout, and pass it to each of the args in this function.  Is this the only thing preventing that?  If so, can we just do this instead?
```suggestion
        if not {"x", "y", "vertex"} <= set(pos_list.columns):
```

rlratzel (commented):
Thanks, Erik! I'm excited to have less tech debt and the legacy Python API removed. I had a few suggestions/questions.

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: No preference. I was doing minimal, safe changes, but am happy to improve UX. NetworkX also includes `pos` as a positional argument:
- https://github.com/networkx/networkx/blob/bf928712c3c6db298691a9b15c4b48e6b7949b35/networkx/drawing/layout.py#L1604-L1608

so maybe `cugraph.force_atlas2` could do the same.

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: That's a good idea! I actually started changing this early on, but stopped b/c other things needed more attention. I'll change it so a single dataframe can be used for all.

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: I added more CAPI_EXPECTS checks. Should we check anything else such as array sizes?

Yes, edge weights _are_ used, and results seem reasonable to me when they are integral. Not sure how that works 🤷‍♂️ . Is this something we should check or try to understand?

python/cugraph/cugraph/layout/force_atlas2.py
```diff
@@ -23,15 +64,20 @@ def force_atlas2(
...
+    vertex_radius=None,
+    overlap_scaling_ratio=100.0,
     edge_weight_influence=1.0,
     jitter_tolerance=1.0,
     barnes_hut_optimize=True,
     barnes_hut_theta=0.5,
     scaling_ratio=2.0,
     strong_gravity_mode=False,
     gravity=1.0,
+    mobility=None,
```
eriknw: Why are the new parameters `vertex_radius` and `mobility`? Would e.g. `vertex_mobility` be more clear and consistent?

nv-rliu:
/ok to test [abf4c66](https://github.com/rapidsai/cugraph/pull/5260/commits/abf4c66fd7e4c0f79a89ed1d2a4f56f6ff9a6bc8)

python/cugraph/cugraph/layout/force_atlas2.py
ChuckHastings: Are you saying that the edge types in C++ are integral, or that you're creating a graph with integral edge types from python and the results are reasonable?

I seem to recall that the python code, when creating a graph with weights, converts the weights into float if they are not of that input type...

You could probably check by displaying the output of `view_edge_list` on the python graph object.  I believe that will be a data frame containing the src, dst, and wgt of the C++ graph structure.  If that data structure is floats, then that's why it works.

ChuckHastings (approved):

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: Aha, yes indeed I'm "creating a graph with integral edge types from python and the results are reasonable", and yes indeed the weights are converted to floating point.

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: Done.

python/cugraph/cugraph/layout/force_atlas2.py
rlratzel: > Why are the new parameters `vertex_radius` and `mobility`? Would e.g. `vertex_mobility` be more clear and consistent?

Sounds good to me!

rlratzel (approved):

python/cugraph/cugraph/layout/force_atlas2.py
rlratzel: @eriknw shall I `/merge` or do you want to make this change?

eriknw: I'm +0.5 to use `vertex_mobility`, so if you're also in favor, I'll make the change tonight, then maybe we can merge tomorrow morning.

rlratzel: I agree that it seems more consistent so I'm in favor.
I'll add a `DO_NOT_MERGE` since this has been approved so it doesn't get merged by someone else before you get that change in.

rlratzel:
Added `DO_NOT_MERGE` since there's one more change to be made: https://github.com/rapidsai/cugraph/pull/5260#discussion_r2380624038

eriknw:
/ok to test [8a12d44](https://github.com/rapidsai/cugraph/pull/5260/commits/8a12d440bbede266df5a85468a65cb2fd1bd8f41)

python/cugraph/cugraph/layout/force_atlas2.py
eriknw: I pushed the change and removed the `DO_NOT_MERGE` label

python/cugraph/cugraph/layout/force_atlas2.py
```diff
@@ -129,43 +187,117 @@ def on_train_end(self, positions):
...
+        # Ensure dtypes are valid, warn if we need to cast
+        vertex_mobility_vertices = ensure_vertex_dtype(
+            input_graph, vertex_mobility["vertex"], 'vertex_mobility["vertex"]'
+        )
+        vertex_mobility_values = ensure_float32_dtype(
+            vertex_mobility["mobility"], 'vertex_mobility["mobility"]'
+        )
 
     if input_graph.is_directed():
         input_graph = input_graph.to_undirected()
```
eriknw, line 272: Would it be better to symmetrize the edges in C++?

eriknw:
One more note: allowing the vertex masses to be set by the user would be useful for nx-cugraph to better match NetworkX.

PR #5269, opened 2025-09-23
deprecate cugraph-service

BradReesWork:
first step in dropping cugraph-service

### DIFF
--- a/python/cugraph-service/client/cugraph_service_client/client.py
+++ b/python/cugraph-service/client/cugraph_service_client/client.py
@@ -160,9 +160,9 @@ class CugraphServiceClient:
         >>> client = CugraphServiceClient()
         """
         warnings.warn(
-            "cugraph-service-client is deprecated and will be removed in a future release. "
-            "If cugraph_service is critical for your work, please submit a GitHub "
-            "issue at https://github.com/rapidsai/cugraph/issues",
+            "cugraph-service-client is deprecated and will be removed in a future "
+            "release. If cugraph_service is critical for your work, please submit "
+            "a GitHub issue at https://github.com/rapidsai/cugraph/issues",
             FutureWarning,
         )
 
--- a/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
+++ b/python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
@@ -163,9 +163,9 @@ class CugraphHandler:
 
     def __init__(self):
         warnings.warn(
-            "cugraph-service-server is deprecated and will be removed in a future release. "
-            "If cugraph_service is critical for your work, please submit a GitHub "
-            "issue at https://github.com/rapidsai/cugraph/issues",
+            "cugraph-service-server is deprecated and will be removed in a future "
+            "release. If cugraph_service is critical for your work, please submit "
+            "a GitHub issue at https://github.com/rapidsai/cugraph/issues",
             FutureWarning,
         )
         self.__next_graph_id = defaults.graph_id + 1


python/cugraph-service/README.md
```diff
@@ -1,5 +1,12 @@
 # cugraph_service
 
+## NOTICE: cuGraph Service is deprecated
+
+cugraph_service, both client and server, is deprecated and will be remove in release 25.12
```
rlratzel: I think we should follow the RAPIDS / cuDF guidelines [here](https://docs.rapids.ai/api/cudf/nightly/developer_guide/contributing_guide/#deprecating-and-removing-code) and not put releases in the message.
```suggestion
## NOTICE: cugraph-service is deprecated

The cugraph-service client and server are deprecated and will be removed in a future release.
```

python/cugraph-service/client/cugraph_service_client/client.py
```diff
@@ -158,6 +159,13 @@ def __init__(
         >>> from cugraph_service_client import CugraphServiceClient
         >>> client = CugraphServiceClient()
         """
+        warnings.warn(
+            "deprecated as of 25.10 and will be remove in release 25.12. "
+            "If cugraph_service is critical for your work, please submit a GitHub "
+            "issue at https://github.com/rapidsai/cugraph/issues",
+            DeprecationWarning,
+        )
```
rlratzel, line 167: More updates based on the deprecation guidelines.  We may also want to drop the "If cugraph-service is critical..." sentence to keep the message to a single line and just leave it in the README.
```suggestion
        warnings.warn(
            "cugraph-service-client is deprecated and will be removed in a future release. "
            "If cugraph_service is critical for your work, please submit a GitHub "
            "issue at https://github.com/rapidsai/cugraph/issues",
            FutureWarning,
        )
```

python/cugraph-service/server/cugraph_service_server/cugraph_handler.py
```diff
@@ -160,6 +162,12 @@ class CugraphHandler:
     __server_facade_extension_param_name = "server"
 
     def __init__(self):
+        warnings.warn(
+            "deprecated as of 25.10 and will be remove in release 25.12. "
+            "If cugraph_service is critical for your work, please submit a GitHub "
+            "issue at https://github.com/rapidsai/cugraph/issues",
+            DeprecationWarning,
+        )
```
rlratzel, line 170:
```suggestion
        warnings.warn(
            "cugraph-service-server is deprecated and will be removed in a future release. "
            "If cugraph_service is critical for your work, please submit a GitHub "
            "issue at https://github.com/rapidsai/cugraph/issues",
            FutureWarning,
        )
```

python/cugraph-service/pytest.ini
```diff
@@ -37,3 +37,8 @@ python_files = bench_*
 
 python_functions = bench_*
                    test_*
+
+filterwarnings =
+          error::DeprecationWarning
+          # TODO
+          ignore:deprecated as of 25.10 and will be remove in release 25.12. If cugraph_service is critical for your work, please submit a GitHub issue at https://github.com/rapidsai/cugraph/issues:DeprecationWarning
```
rlratzel: It doesn't hurt to leave the line in to treat DeprecationWarnings as errors (even though the next step is to remove cugraph-service entirely so it's probably not needed), but we'll have to update the `ignore` line to match the suggestions (I haven't tested this to see if the regex is right).
```suggestion

filterwarnings =
          error::DeprecationWarning
          # TODO
          ignore:cugraph-service-.* is deprecated and will be removed .*:FutureWarning
```

python/cugraph-service/README.md
BradReesWork: Typically I would agree with this. But in this case we are eager to get this package dropped and need to let people know it is soon

python/cugraph-service/client/cugraph_service_client/client.py
BradReesWork: Having it just in the README is fine.   
The problem is that there are no cugraph-service customer _that we know of_.  I would to force any customer to reach out know

rlratzel (approved):
LGTM

PR #5273, opened 2025-09-23
Clarify supported type combination

### LINKED ISSUE #4587: [IMP] Clarify Supported Type Combinations when creating a SGGraph or MGGraph
When passing an unsupported type combination, the user gets currently gets  an undescriptive error message that is meaningless to them unless they understand libcugraph.  Instead, we should raise a more descriptive error in pylibcugraph (i.e. `int64` is not supported for edge weights).

jnke2016:
This PR internally matches graph' attribute's types if the type combination is not supported and throws a warning.

closes #4587


rlratzel:
I met with @jnke2016 and @ChuckHastings yesterday. We agreed to move the array widening to C/C++ and have the new Python utility in this PR simply check the input dtypes and issue a warning if they'll be widened to 64-bit. I'm just waiting on those changes before review/approval.

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -116,6 +116,23 @@ void transform_increment_ints(raft::device_span<value_t> values,
...
+ * @tparam      old_vertex_t     type of the value to operate on. Must be either int32_t or int64_t.
+ *
+ * @param[out]  new_vertices      device span to update with new data type
+ * @param[in]   old_vertices      device span with initial data type
+ * @param[in]   stream_view  stream view
+ *
+ */
+template <typename new_vertex_t, typename old_vertex_t>
+void transform_cast_ints(raft::device_span<new_vertex_t> new_vertices,
+                         raft::device_span<old_vertex_t> old_vertices,
```
ChuckHastings: Should be `old_vertex_t const`, we can't modify the input.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -109,19 +109,74 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
                   : std::nullopt;
 
       if (vertex_list) {
-        raft::copy<vertex_t>(vertex_list->data(),
-                             vertices_->as_type<vertex_t>(),
-                             vertices_->size_,
-                             handle_.get_stream());
+        if (((vertices_->type_ == cugraph_data_type_id_t::INT32) &&
```
ChuckHastings: I think this looks correct.  I'd suggest a slightly different approach.

Instead of adding a function `transform_cast_ints` and replicating all of this type checking logic wherever it's required, how about creating the following function:

```
template <typename new_type>
void copy_or_transform(
  raft::device_span<new_type_t> output,
  cugraph_type_erased_device_array_view_t const* input,
  rmm::cuda_stream_view const& stream_view);
```

This function would have your if statements (determining type combinations), perform the raft::copy when appropriate and do the transform when appropriate.

I think we'll end up with cleaner and more readable code.

I'd also include logic in this function to support float/double combinations, although we don't have a current example where we need that, I could see where upcasting floats to doubles might be a thing we need to do in the C API.

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -115,23 +117,21 @@ template <typename value_t>
...
+ * @param[in]   input       cugraph_type_erased_device_array_view_t with initial data type
  * @param[in]   stream_view  stream view
  *
  */
-template <typename new_vertex_t, typename old_vertex_t>
-void transform_cast_ints(raft::device_span<new_vertex_t> new_vertices,
-                         raft::device_span<old_vertex_t> old_vertices,
-                         rmm::cuda_stream_view const& stream_view);
+template <typename new_type_t>
+void copy_or_transform(raft::device_span<new_type_t> output,
```
ChuckHastings: This function should be defined in the c_api directory.  I'd suggest `graph_helper.hpp`.  Putting it here brings C API things into the C++ code path.

Putting it in `graph_helper.hpp` will also put it in the `cugraph::c_api` namespace which will make some of my other comments possible.

cpp/src/c_api/graph_mg.cpp
```diff
@@ -46,10 +46,10 @@ rmm::device_uvector<value_t> concatenate(
   size_t concat_pos{0};
 
   for (size_t i = 0; i < num_arrays; ++i) {
-    raft::copy<value_t>(results.data() + concat_pos,
-                        values[i]->as_type<value_t>(),
-                        values[i]->size_,
-                        handle.get_stream());
+    cugraph::detail::copy_or_transform(
+      raft::device_span<value_t>{results.data(), results.size()},
+      reinterpret_cast<cugraph_type_erased_device_array_view_t const*>(values[i]),
```
ChuckHastings: If we move the function, the reinterpret_cast should not be necessary here.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -109,74 +109,24 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
-          } else {
-            cugraph::detail::transform_cast_ints(
-              raft::device_span<vertex_t>{(*vertex_list).data(), (*vertex_list).size()},
-              raft::device_span<int64_t const>{vertices_->as_type<int64_t>(), vertices_->size_},
-              handle_.get_stream());
-          }
-        }
+        cugraph::detail::copy_or_transform(
+          raft::device_span<vertex_t>{(*vertex_list).data(), (*vertex_list).size()},
+          reinterpret_cast<cugraph_type_erased_device_array_view_t const*>(vertices_),
```
ChuckHastings: If we move the function, the reinterpret cast should not be necessary here.

cpp/src/detail/utility_wrappers_32.cu
```diff
@@ -81,13 +81,18 @@ template void transform_increment_ints(raft::device_span<int32_t> values,
...
                                        rmm::cuda_stream_view const& stream_view);
 
-template void transform_cast_ints(raft::device_span<int32_t> new_vertices,
-                                  raft::device_span<int32_t const> old_vertices,
-                                  rmm::cuda_stream_view const& stream_view);
-
-template void transform_cast_ints(raft::device_span<int32_t> new_vertices,
-                                  raft::device_span<int64_t const> old_vertices,
-                                  rmm::cuda_stream_view const& stream_view);
+template void copy_or_transform(raft::device_span<int32_t> output,
```
ChuckHastings: These would move to `graph_helper_sg.cu`.  Won't need a copy in _mg file, only one is necessary.

cpp/src/detail/utility_wrappers_64.cu
```diff
@@ -79,13 +79,13 @@ template void transform_increment_ints(raft::device_span<int64_t> values,
                                        int64_t value,
                                        rmm::cuda_stream_view const& stream_view);
 
-template void transform_cast_ints(raft::device_span<int64_t> new_vertices,
-                                  raft::device_span<int64_t const> old_vertices,
-                                  rmm::cuda_stream_view const& stream_view);
+template void copy_or_transform(raft::device_span<int64_t> output,
```
ChuckHastings: These would move to `graph_helper_sg.cu`. Won't need a copy in _mg file, only one is necessary.

cpp/src/detail/utility_wrappers_impl.cuh
```diff
@@ -102,18 +104,55 @@ void transform_increment_ints(raft::device_span<value_t> values,
                     }));
 }
 
-template <typename new_vertex_t, typename old_vertex_t>
-void transform_cast_ints(raft::device_span<new_vertex_t> new_vertices,
-                         raft::device_span<old_vertex_t> old_vertices,
-                         rmm::cuda_stream_view const& stream_view)
+template <typename new_type_t>
```
ChuckHastings: This would move into `graph_helper_impl.cuh`.

cpp/src/detail/utility_wrappers_impl.cuh
```diff
@@ -102,18 +104,55 @@ void transform_increment_ints(raft::device_span<value_t> values,
...
 {
-  thrust::transform(rmm::exec_policy(stream_view),
-                    old_vertices.begin(),
-                    old_vertices.end(),
-                    new_vertices.begin(),
-                    cuda::proclaim_return_type<new_vertex_t>([] __device__(old_vertex_t value) {
-                      return static_cast<new_vertex_t>(value);
-                    }));
+  auto input_ =
+    reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_view_t const*>(input);
```
ChuckHastings: With the code in the c_api namespace, the reinterpret cast would not be necessary.

ChuckHastings (commented):
Logic looks correct again.  I think a quick move around of some things will let you git rid of some reinterpret_casts and wrap this up.

cpp/src/c_api/graph_sg.cpp
```diff
@@ -164,23 +169,23 @@ struct create_graph_functor : public cugraph::c_api::abstract_functor {
...
-                                edge_end_times_->size_,
-                                handle_.get_stream());
+      if (edgelist_edge_start_times) {
+        cugraph::c_api::copy_or_transform(
+          raft::device_span<edge_time_t>{edgelist_edge_start_times->data(),
+                                         edgelist_edge_start_times->size()},
+          edge_start_times_,
+          handle_.get_stream());
+
+        cugraph::c_api::copy_or_transform(
```
ChuckHastings, line 186: Seems like you're missing an if block here.

The original code had an if block for edgelist_edge_start_times and a separate if block for edgelist_edge_end_times.  It's possible for one to be specified without the other.

PR #5283, opened 2025-09-25
Add support for other temporal comparisons

### LINKED ISSUE #5040: Temporal Sampling - Support `>` and `>=` Temporal Constraints in C/C++ API
Support adjusting the behavior to allow vertices to be selected with timestamps `>` or `>=` the previous timestamp depending on a flag provided by the user.

ChuckHastings:
Closes #5040.

Update temporal sampling to support the following variations:
* STRICTLY_INCREASING - time of next edge > time of previous edge
* MONOTONICALLY_INCREASING - time of next edge >= time of previous edge. ** OLD default behavior
* STRICTLY_DECREASING - time of next edge < time of previous edge
* MONOTONICALLY_DECREASING - time of next edge <= time of previous edge

This PR includes C++ tests and C API tests.

Marked as breaking as we add a parameter to the C++ temporal sampling function, although it can be defaulted so it's only breaking for code that specifies `do_expensive_check`.


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -254,6 +254,16 @@ biased_neighbor_sample(
...
   bool do_expensive_check                         = false);
 
+enum class temporal_sampling_comparison_t {
+  STRICTLY_INCREASING = 0,  /** Time strictly increasing (each time is after the previous one) */
+  MONOTONICALLY_INCREASING, /** Time monotonically increasing (could have multiple edges with same
+                              time) */
+  STRICTLY_DECREASING,      /** Time strictly decreasing (each time is before the previous one) */
+  MONOTONICALLY_DECREASING, /** Time monotonically decreasing (could have multiple edges with same
+                                time) */
+  LAST                      /** Support last n behavior */
```
seunghwak, line 264: `LAST                      /** Support last n behavior */`

This sounds vague to me.

Isn't this something like `NUM_COMPARISON_TYPES`? (basically the number of supported temporal sampling comparison methods/modes/types... and so o n)

cpp/src/c_api/temporal_neighbor_sampling.cpp
```diff
@@ -279,6 +279,27 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
...
+          break;
+        case cugraph_temporal_sampling_comparison_t::STRICTLY_DECREASING:
+          temporal_sampling_comparison =
+            cugraph::temporal_sampling_comparison_t::STRICTLY_DECREASING;
+          break;
+        case cugraph_temporal_sampling_comparison_t::MONOTONICALLY_DECREASING:
+          temporal_sampling_comparison =
+            cugraph::temporal_sampling_comparison_t::MONOTONICALLY_DECREASING;
+          break;
+        default: CUGRAPH_FAIL("Invalid temporal sampling comparison type");
```
seunghwak, line 300: `auto temporal_sampling_comparison = options_.temporal_sampling_comparison_;`

Won't this work?

cpp/src/c_api/temporal_neighbor_sampling.cpp
```diff
@@ -321,7 +342,8 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
                                         options_.return_hops_,
                                         options_.dedupe_sources_,
-                                        options_.with_replacement_},
+                                        options_.with_replacement_,
+                                        temporal_sampling_comparison},
```
seunghwak, line 346: Or can we just call this with `options_.temporal_sampling_comparison_`? Why are we creating a temporary?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -532,24 +534,48 @@ temporal_gather_one_hop_edgelist(
...
+                                 switch (temporal_sampling_comparison) {
+                                   case temporal_sampling_comparison_t::STRICTLY_INCREASING:
+                                     return (edge_time > key_time);
+                                   case temporal_sampling_comparison_t::MONOTONICALLY_INCREASING:
+                                     return (edge_time >= key_time);
+                                   case temporal_sampling_comparison_t::STRICTLY_DECREASING:
+                                     return (edge_time < key_time);
+                                   case temporal_sampling_comparison_t::MONOTONICALLY_DECREASING:
+                                     return (edge_time <= key_time);
+                                 }
```
seunghwak, line 556: `assert(false);  // never be reached` I assume this part should never be reached...

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -532,24 +534,48 @@ temporal_gather_one_hop_edgelist(
...
```
seunghwak, line 576: Same here.

`assert(false);  // never be reached`

seunghwak (approved):
LGTM except for few cosmetic issues.

jnke2016 (approved):

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: This is not the end of the enum.  It's actually to support a feature that will be implemented later called "last n".

If `LAST` is specified, the software will (eventually, not part of this PR) use the fanout value (call it `n`) and instead of randomly selecting `n` edges that leave this vertex it will select the `n` edges with the largest time stamps.  You can think of this as sorting the outgoing edges by time and selecting the "last n" of them.

This is a feature that we'll add later (perhaps later in 25.12, perhaps not until 26.01, not sure of the priorities and how far we'll progress).

cpp/src/c_api/temporal_neighbor_sampling.cpp
ChuckHastings: We're converting between types here.  I'm hesitant to assume that translating from the C enum definition in the `.h` file to the C++ enum class definition in the .hpp file is guaranteed to work by assignment.  This was my simple solution.

cpp/src/c_api/temporal_neighbor_sampling.cpp
ChuckHastings: Type conversion as mentioned above

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I'll make that change.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I'll make that change

PR #5291, opened 2025-10-01
Use Docker CUDA major-version tags

bdice:
We need to update some Docker image tags in the docs to use only CUDA major versions.

For example: `rapidsai/base:25.10-cuda12-py3.13` instead of `rapidsai/base:25.10-cuda12.9-py3.13`.

See the following for more information:
- https://docs.rapids.ai/notices/rsn0053/
- https://github.com/rapidsai/docker/issues/781

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -59,17 +59,17 @@ Running the example in these notebooks requires:
 
 ### QuickStart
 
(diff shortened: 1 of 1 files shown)


acostadon (approved):
Looks good

PR #5293, opened 2025-10-01
`force_atlas2`: allow vertex mass to be given by the user

eriknw:
This is a "quick" follow-up to https://github.com/rapidsai/cugraph/pull/5260 (and https://github.com/rapidsai/cugraph/pull/5213) and https://github.com/rapidsai/nx-cugraph/pull/191 to add vertex mass to the API. This will allow `nx-cugraph` to better match the networkx API.

This required changing the dtype of the mass arrays from integral to float32.

As time permits, I plan to add a test to show that increasing mass of a vertex increases its distance to other vertices. The plan is to also update `nx-cugraph` to use this parameter.

### DIFF
--- a/python/pylibcugraph/pylibcugraph/force_atlas2.pyx
+++ b/python/pylibcugraph/pylibcugraph/force_atlas2.pyx
@@ -84,6 +84,8 @@ def force_atlas2(ResourceHandle resource_handle,
                  double gravity,
                  vertex_mobility_vertices,
                  vertex_mobility_values,
+                 vertex_mass_vertices,
+                 vertex_mass_values,
                  bool_t verbose,
                  bool_t do_expensive_check,
                 ):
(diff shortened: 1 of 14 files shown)


python/pylibcugraph/pylibcugraph/force_atlas2.pyx
```diff
@@ -196,14 +205,16 @@ def force_atlas2(ResourceHandle resource_handle,
...
+    ...     None, None, False, False)
     >>> vertices
     [   0  1   2   3   4   5    ]
     >>> x_axis
-    [ 5.444471    0.4794112   1.2495936  -0.01039529 -1.1892298  -1.5889403 ]
+    [-7.7015705   7.6763854  -2.7651896  -0.02446137  1.6971487  -1.2822613 ]
     >>> y_axis
-    [-1.4304754e+01 -3.8182523e+00  3.8365445e+00  8.3183739e-03 -8.3009762e-01 -1.9155006e-01
+    [ 23.276543     7.9067745   13.618961    -0.07172047  -6.8321953  -11.90544   ]
+
```
rlratzel, line 217: Thanks for remembering the docstring example.

I'm assuming these values are drastically different because the call in the example is also different/updated (and not simply because support for `mass` was added)?

eriknw: Yeah, the call is slightly different, and also includes RNG seed, which hopefully makes the result more reproducible.

PR #5300, opened 2025-10-03
Update symmetrization to use variant structures to reduce library size

ChuckHastings:
So far this just defines the proposed new symmetrization function signatures.


cpp/include/cugraph/graph_functions.hpp
```diff
@@ -441,39 +453,56 @@ decompress_to_edgelist(
...
-                    rmm::device_uvector<vertex_t>&& edgelist_srcs,
-                    rmm::device_uvector<vertex_t>&& edgelist_dsts,
-                    std::optional<rmm::device_uvector<weight_t>>&& edgelist_weights,
-                    std::optional<rmm::device_uvector<edge_t>>&& edgelist_edge_ids,
-                    std::optional<rmm::device_uvector<edge_type_t>>&& edgelist_edge_types,
-                    std::optional<rmm::device_uvector<edge_time_t>>&& edgelist_edge_start_times,
-                    std::optional<rmm::device_uvector<edge_time_t>>&& edgelist_edge_end_times,
-                    bool reciprocal);
+           std::vector<arithmetic_device_uvector_t>>
+add_reciprocal_edges(
```
seunghwak: I think this name is confusing.

This sounds like we are adding reciprocal edges to another edge list. But isn't this inspecting the input edge list and keeping only the reciprocal edges?

I think we need a clearer name.

And the function below...

Isn't removing non-reciprocal edges same to the previous symmetrize_edgelist function with reciprocal = true. The comment is saying the opposite.

So, there are two ways to symmetrize.

1) Only keep the reciprocal edges. Remove all non-reciprocal edges.
2) Add inverse edges for all non-reciprocal edges.

I think we'd better rename the functions to make this more intuitive...

I asked this to ChatGPT, and these are the names ChatGPT suggested...

make_reciprocal_edges() <= But I think these names are bad... Still confusing.
make_symmetric_edges()

symmetrize_edgelist_union() <= Sounds better... not sure reciprocal & intersection aligns 100%
symmetrize_edgelist_intersection()

keep_reciprocal_edges() <= If I need to pick one out of the three.. I may go with this.
add_inverse_edges()

But there might be better names than this... Just a starting  point...

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -423,15 +423,27 @@ decompress_to_edgelist(
 
 /**
  * @ingroup graph_functions_cpp
- * @brief Symmetrize edgelist.
+ * @brief Rule for combining edge properties when adding reciprocal edges.
+ */
+enum class edge_property_combining_rule_t {
+  PART_OF_UNIQUENESS,  /// This property is part of the uniqueness of the edge.
```
seunghwak: Yeah... I am not sure about this... This is not really a combining rule...

And if this is selected for an edge property... how do we consider that in symmetrization.

Say edge ID is selected as "PART_OF_UNIQUENESS" property.

 In (src, dst, ID) triplets, how do we symmetrize (3, 5, 1) and (5, 3, 1) with reciprocal = true/false? What about (3, 5, 1) and (5, 3, 2)?
 
 I think we should make this cleaner and better document the expected behavior. Just reading the names, I can't properly guess how these functions will work.

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -412,15 +412,40 @@ decompress_to_edgelist(
...
+ */
+enum class edge_value_combining_rule_t {
+  SUM,   /** Sum the values of the two edge properties. */
+  MAX,   /** Take the maximum of the values of the two edge properties. */
+  MIN,   /** Take the minimum of the values of the two edge properties. */
+  MEAN,  /** Take the mean of the values of the two edge properties. */
+  FIRST, /** Use the value of the first edge property. */
+  LAST,  /** Use the value of the last edge property. */
+  ANY,   /** Use the value of any edge property. */
+  NONE   /** Do not combine the values. */
```
seunghwak: What is NONE here? Isn't it pretty much identical to edge_key_selector_t::KEY?

seunghwak: If set to edge_key_selector::VALUE, I guess we need to combine in one way or another. If we don't want to combine, we need to set to edge_key_selector::KEY.

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -412,15 +412,40 @@ decompress_to_edgelist(
...
+};
+
+/**
+ * @ingroup graph_functions_cpp
+ * @brief Add reciprocal edges and merge the edges.  This was formerly called symmetrize_edgelist
+ * with reciprocal set to false.
+ *
+ * The input edge list is replicated as reciprocal edges.  Then the original edges and the
```
seunghwak, line 462: I am wondering we should use "reciprocal" here or use "inverse" instead. I am not sure what "reciprocal" exactly means. Some quick search says it is a pair of edges going in both directions between a pair of vertices.

cpp/include/cugraph/graph_functions.hpp
```diff
@@ -430,39 +455,63 @@ decompress_to_edgelist(
...
+ */
+template <typename vertex_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::vector<arithmetic_device_uvector_t>>
+drop_nonreciprocal_edge_tuples(raft::handle_t const& handle,
+                               rmm::device_uvector<vertex_t>&& edgelist_srcs,
+                               rmm::device_uvector<vertex_t>&& edgelist_dsts,
+                               std::vector<arithmetic_device_uvector_t>&& edgelist_edge_properties,
+                               std::vector<edge_key_selector_t> const& edge_key_selector,
```
seunghwak: We don't need `std::vector<edge_value_combining_rule_t> const& edge_value_combining_rule` here? If we have a reciprocal edges, are there no use-cases we need to combine edge properties? (e.g. update the edge weights to the average)

cpp/include/cugraph/graph_functions.hpp
seunghwak, line 454: This is just to keep the original values, right? Should we explain this in little more detail?

I initially thought we don't need NONE here as we now have `edge_key_selector_t`. But now I think this is still necessary as we may consider two different edges with the matching source/destination pairs as reciprocal but in some cases, we don't want to combine there edge properties. This may not be clear to everyone from the beginning.

seunghwak (commented):
I like `edge_key_selector_t`. I still few some questions.

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: I'm fine switching to inverse here if that's more intuitive to you.  I suppose by the definition you found we're really creating reciprocal edges here by adding the inverse edge.  So that's probably more correct.

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: I'll add some text to clarify the different use cases.

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: Yes, I suppose we do.  I hadn't considered that case.  But if we have (1,2,1.0) and (2,1,2.0) then we would need to have the `edge_value_combining_rule_t` for that edge property to determine what value to assign to the pair of edges.  In that case `NONE` would leave the edges unmodified (and different in each direction), `ANY` would arbitrarily pick one of them to make them match.

PR #5305, opened 2025-10-10
[CRITICAL BUG] Fix sample result destructor

alexbarghi-nv:
Addresses a severe memory leak in sampling caused by a bug in the C API.  The destructor for the sample result object was not properly updated to cover additional fields that were added later.  This resulted in these fields never being cleaned up, and eventually to an OOM error.


alexbarghi-nv:
Looks like some issue with rmm is breaking the python build

bdice:
I think the failed jobs can be rerun once a new `pylibcudf` build is available. A PR was just merged to cuDF so hopefully this will be fixed soon.

alexbarghi-nv:
Will retarget to 25.10

alexbarghi-nv:
Superseded by https://github.com/rapidsai/cugraph/pull/5306

PR #5310, opened 2025-10-13
Require C++20

bdice:
This PR enables C++20 in cuGraph builds.


seunghwak:
Oh, so are we migrating to C++20?

bdice:
@seunghwak I am investigating where various libraries are with respect to C++20 support. This isn't a super high priority but I would like to better understand what our constraints might be. It would help us in RMM if we could use C++20 features, but that requires other RAPIDS libraries to migrate to C++20. cuDF and some other libraries already use C++20. xref: https://github.com/rapidsai/rmm/issues/2034

PR #5316, opened 2025-10-16
Fix warning in unreachable device code

bdice:
cuGraph C++ builds have been issuing a warning in a dispatch that is unreachable. This fixes the warning by marking the code as unreachable with a macro borrowed from cuDF.

Warnings look like:

```cpp
/home/coder/cugraph/cpp/src/sampling/detail/gather_one_hop_impl.cuh(558): warning #940-D: missing return statement at end of non-void function "lambda [](auto)->auto [with <auto-1>=unsigned long]"
```

### DIFF
--- a/cpp/src/sampling/detail/gather_one_hop_impl.cuh
+++ b/cpp/src/sampling/detail/gather_one_hop_impl.cuh
@@ -28,6 +28,7 @@
 #include <cugraph/arithmetic_variant_types.hpp>
 #include <cugraph/edge_property.hpp>
 #include <cugraph/sampling_functions.hpp>
+#include <cugraph/utilities/assert.cuh>
 #include <cugraph/utilities/mask_utils.cuh>
 
 #include <raft/util/cudart_utils.hpp>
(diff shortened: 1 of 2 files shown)


cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -554,7 +555,7 @@ temporal_gather_one_hop_edgelist(
                                    case temporal_sampling_comparison_t::MONOTONICALLY_DECREASING:
                                      return (edge_time <= key_time);
                                  }
-                                 assert(false);
+                                 CUGRAPH_UNREACHABLE("Unsupported temporal sampling comparison");
```
bdice, line 558: I am assuming this is _meant_ to be unreachable from the `assert(false)` but I notice there is an unhandled case, `temporal_sampling_comparison_t::LAST`. I am unsure if this is a bug.

ChuckHastings: Thanks.  The `temporal_sampling_comparison_t::LAST` is a place holder for an as of yet unimplemented feature.  This is meant to be unreachable.

PR #5326, opened 2025-10-26
Update API to leverage cuVS Spectral clustering

jnke2016:
This PR integrates cuVS spectral clustering from upstream into cuGraph while maintaining full backward compatibility. The integration deprecates the legacy balanced cut clustering C++ and Python APIs in favor of spectral_modularity_maximization. In fact, cuVS consolidated the spectral clustering API to assign vertices to clusters that by default maximizes the modularity score

**Static Linking with Conda Fallback:**
- cuGraph is configured to statically link libcuvs for optimal performance
- The build system first attempts to use conda-installed libcuvs if available
- If libcuvs is not present in the conda environment or static linking is required, the build system automatically clones the main cuVS repository (rapidsai/cuvs) and builds it from source
- Uses the standard RAPIDS release branch (${rapids-cmake-checkout-tag}) ensuring version compatibility

Note: The header files `raft/spectral/partition.cuh` and `raft/spectral/modularity_maximization.cuh` are still necessary in `spectraly_clustering.cu` because APIs (like `raft::spectral::analyzePartition`, `raft::spectral::analyzeModularity`) are still being called in cuGraph and are not supported in cuVS. This is not an issue because those APIs are not deprecated in raft in contrary to `kmeans` and `solvers`

### DIFF
--- a/cpp/include/cugraph/sampling_functions.hpp
+++ b/cpp/include/cugraph/sampling_functions.hpp
@@ -73,6 +73,11 @@ struct sampling_flags_t {
    */
   temporal_sampling_comparison_t temporal_sampling_comparison{
     temporal_sampling_comparison_t::STRICTLY_INCREASING};
+
+  /**
+   * Specifies if disjoint sampling should be enforced. Default is false.
+   */
+  bool disjoint_sampling{false};
 };
 
 /**
--- a/cpp/include/cugraph_c/sampling_algorithms.h
+++ b/cpp/include/cugraph_c/sampling_algorithms.h
@@ -311,6 +311,17 @@ void cugraph_sampling_set_temporal_sampling_comparison(
 
 /**
  * @ingroup samplingC
+ * @brief   Set flag to perform disjoint sampling
+ *
+ * Note: This flag is not supported in the current implementation.
+ *
+ * @param options - opaque pointer to the sampling options
+ * @param value - Boolean value to assign to the option
+ */
+void cugraph_sampling_set_disjoint_sampling(cugraph_sampling_options_t* options, bool_t value);
+
+/**
+ * @ingroup samplingC
  * @brief     Free sampling options object
  *
  * @param [in]   options   Opaque pointer to sampling object
--- a/cpp/src/c_api/neighbor_sampling.cpp
+++ b/cpp/src/c_api/neighbor_sampling.cpp
@@ -279,10 +279,13 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               num_edge_types_,
-              cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_ == TRUE,
-                                        options_.dedupe_sources_ == TRUE,
-                                        options_.with_replacement_ == TRUE},
+              cugraph::sampling_flags_t{
+                options_.prior_sources_behavior_,
+                options_.return_hops_ == TRUE,
+                options_.dedupe_sources_ == TRUE,
+                options_.with_replacement_ == TRUE,
+                cugraph::temporal_sampling_comparison_t::STRICTLY_INCREASING,
+                options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         } else {
           std::tie(sampled_srcs,
@@ -309,10 +312,13 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               num_edge_types_,
-              cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_ == TRUE,
-                                        options_.dedupe_sources_ == TRUE,
-                                        options_.with_replacement_ == TRUE},
+              cugraph::sampling_flags_t{
+                options_.prior_sources_behavior_,
+                options_.return_hops_ == TRUE,
+                options_.dedupe_sources_ == TRUE,
+                options_.with_replacement_ == TRUE,
+                cugraph::temporal_sampling_comparison_t::STRICTLY_INCREASING,
+                options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         }
       } else {
@@ -342,10 +348,13 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
                                      (*label_to_comm_rank).data(), (*label_to_comm_rank).size()})
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
-              cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_ == TRUE,
-                                        options_.dedupe_sources_ == TRUE,
-                                        options_.with_replacement_ == TRUE},
+              cugraph::sampling_flags_t{
+                options_.prior_sources_behavior_,
+                options_.return_hops_ == TRUE,
+                options_.dedupe_sources_ == TRUE,
+                options_.with_replacement_ == TRUE,
+                cugraph::temporal_sampling_comparison_t::STRICTLY_INCREASING,
+                options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         } else {
           std::tie(sampled_srcs,
@@ -371,10 +380,13 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
                                      (*label_to_comm_rank).data(), (*label_to_comm_rank).size()})
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
-              cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_ == TRUE,
-                                        options_.dedupe_sources_ == TRUE,
-                                        options_.with_replacement_ == TRUE},
+              cugraph::sampling_flags_t{
+                options_.prior_sources_behavior_,
+                options_.return_hops_ == TRUE,
+                options_.dedupe_sources_ == TRUE,
+                options_.with_replacement_ == TRUE,
+                cugraph::temporal_sampling_comparison_t::STRICTLY_INCREASING,
+                options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         }
       }
--- a/cpp/src/c_api/sampling_common.hpp
+++ b/cpp/src/c_api/sampling_common.hpp
@@ -23,6 +23,7 @@ struct cugraph_sampling_options_t {
   bool_t retain_seeds_{FALSE};
   cugraph_temporal_sampling_comparison_t temporal_sampling_comparison_{
     cugraph_temporal_sampling_comparison_t::STRICTLY_INCREASING};
+  bool_t disjoint_sampling_{FALSE};
 };
 
 struct sampling_flags_t {
--- a/cpp/src/c_api/sampling_result.cpp
+++ b/cpp/src/c_api/sampling_result.cpp
@@ -99,6 +99,13 @@ extern "C" void cugraph_sampling_set_temporal_sampling_comparison(
   internal_pointer->temporal_sampling_comparison_ = value;
 }
 
+extern "C" void cugraph_sampling_set_disjoint_sampling(cugraph_sampling_options_t* options,
+                                                       bool_t value)
+{
+  auto internal_pointer = reinterpret_cast<cugraph::c_api::cugraph_sampling_options_t*>(options);
+  internal_pointer->disjoint_sampling_ = value;
+}
+
 extern "C" void cugraph_sampling_options_free(cugraph_sampling_options_t* options)
 {
   auto internal_pointer = reinterpret_cast<cugraph::c_api::cugraph_sampling_options_t*>(options);
--- a/cpp/src/c_api/temporal_neighbor_sampling.cpp
+++ b/cpp/src/c_api/temporal_neighbor_sampling.cpp
@@ -364,10 +364,11 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               num_edge_types_,
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_,
-                                        options_.dedupe_sources_,
-                                        options_.with_replacement_,
-                                        temporal_sampling_comparison},
+                                        options_.return_hops_ == TRUE,
+                                        options_.dedupe_sources_ == TRUE,
+                                        options_.with_replacement_ == TRUE,
+                                        temporal_sampling_comparison,
+                                        options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         } else {
           std::tie(sampled_edge_srcs,
@@ -404,10 +405,11 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               num_edge_types_,
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_,
-                                        options_.dedupe_sources_,
-                                        options_.with_replacement_,
-                                        temporal_sampling_comparison},
+                                        options_.return_hops_ == TRUE,
+                                        options_.dedupe_sources_ == TRUE,
+                                        options_.with_replacement_ == TRUE,
+                                        temporal_sampling_comparison,
+                                        options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         }
       } else {
@@ -447,10 +449,11 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_,
-                                        options_.dedupe_sources_,
-                                        options_.with_replacement_,
-                                        temporal_sampling_comparison},
+                                        options_.return_hops_ == TRUE,
+                                        options_.dedupe_sources_ == TRUE,
+                                        options_.with_replacement_ == TRUE,
+                                        temporal_sampling_comparison,
+                                        options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         } else {
           std::tie(sampled_edge_srcs,
@@ -486,10 +489,11 @@ struct temporal_neighbor_sampling_functor : public cugraph::c_api::abstract_func
                                  : std::nullopt,
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_,
-                                        options_.dedupe_sources_,
-                                        options_.with_replacement_,
-                                        temporal_sampling_comparison},
+                                        options_.return_hops_ == TRUE,
+                                        options_.dedupe_sources_ == TRUE,
+                                        options_.with_replacement_ == TRUE,
+                                        temporal_sampling_comparison,
+                                        options_.disjoint_sampling_ == TRUE},
               do_expensive_check_);
         }
       }
--- a/python/pylibcugraph/pylibcugraph/_cugraph_c/algorithms.pxd
+++ b/python/pylibcugraph/pylibcugraph/_cugraph_c/algorithms.pxd
@@ -341,6 +341,12 @@ cdef extern from "cugraph_c/algorithms.h":
         )
 
     cdef void \
+        cugraph_sampling_set_disjoint_sampling(
+            cugraph_sampling_options_t* options,
+            bool_t value,
+        )
+
+    cdef void \
         cugraph_sampling_options_free(
             cugraph_sampling_options_t* options,
         )
--- a/python/pylibcugraph/pylibcugraph/heterogeneous_biased_neighbor_sample.pyx
+++ b/python/pylibcugraph/pylibcugraph/heterogeneous_biased_neighbor_sample.pyx
@@ -45,6 +45,7 @@ from pylibcugraph._cugraph_c.algorithms cimport (
     cugraph_sampling_set_compress_per_hop,
     cugraph_sampling_set_compression_type,
     cugraph_sampling_set_retain_seeds,
+    cugraph_sampling_set_disjoint_sampling,
 )
 from pylibcugraph._cugraph_c.sampling_algorithms cimport (
     cugraph_heterogeneous_biased_neighbor_sample,
@@ -85,6 +86,7 @@ def heterogeneous_biased_neighbor_sample(ResourceHandle resource_handle,
                                          bool_t do_expensive_check,
                                          prior_sources_behavior=None,
                                          deduplicate_sources=False,
+                                         disjoint_sampling=False,
                                          return_hops=False,
                                          renumber=False,
                                          retain_seeds=False,
@@ -173,6 +175,10 @@ def heterogeneous_biased_neighbor_sample(ResourceHandle resource_handle,
         entire batch.
         If True, will create a separate compressed edgelist per hop within
         a batch.
+
+    disjoint_sampling: bool (Optional)
+        If True, enables disjoint sampling between seeds per hop when supported.
+        Defaults to False.
 
     random_state: int (Optional)
         Random state to use when generating samples.  Optional argument,
@@ -357,6 +363,7 @@ def heterogeneous_biased_neighbor_sample(ResourceHandle resource_handle,
     cugraph_sampling_set_compression_type(sampling_options, compression_behavior_e)
     cugraph_sampling_set_compress_per_hop(sampling_options, c_compress_per_hop)
     cugraph_sampling_set_retain_seeds(sampling_options, retain_seeds)
+    cugraph_sampling_set_disjoint_sampling(sampling_options, disjoint_sampling)
 
     error_code = cugraph_heterogeneous_biased_neighbor_sample(
         c_resource_handle_ptr,
--- a/python/pylibcugraph/pylibcugraph/heterogeneous_biased_temporal_neighbor_sample.pyx
+++ b/python/pylibcugraph/pylibcugraph/heterogeneous_biased_temporal_neighbor_sample.pyx
@@ -45,6 +45,7 @@ from pylibcugraph._cugraph_c.algorithms cimport (
     cugraph_sampling_set_compress_per_hop,
     cugraph_sampling_set_compression_type,
     cugraph_sampling_set_retain_seeds,
+    cugraph_sampling_set_disjoint_sampling,
 )
 from pylibcugraph._cugraph_c.sampling_algorithms cimport (
     cugraph_heterogeneous_biased_temporal_neighbor_sample,
@@ -87,6 +88,7 @@ def heterogeneous_biased_temporal_neighbor_sample(ResourceHandle resource_handle
                                                    bool_t do_expensive_check,
                                                    prior_sources_behavior=None,
                                                    deduplicate_sources=False,
+                                                   disjoint_sampling=False,
                                                    return_hops=False,
                                                    renumber=False,
                                                    retain_seeds=False,
@@ -196,6 +198,10 @@ def heterogeneous_biased_temporal_neighbor_sample(ResourceHandle resource_handle
         Random state to use when generating samples.  Optional argument,
         defaults to a hash of process id, time, and hostname.
         (See pylibcugraph.random.CuGraphRandomState)
+
+    disjoint_sampling: bool (Optional)
+        If True, enables disjoint sampling between seeds per hop when supported.
+        Defaults to False.
 
     Returns
     -------
@@ -396,6 +402,7 @@ def heterogeneous_biased_temporal_neighbor_sample(ResourceHandle resource_handle
     cugraph_sampling_set_compression_type(sampling_options, compression_behavior_e)
     cugraph_sampling_set_compress_per_hop(sampling_options, c_compress_per_hop)
     cugraph_sampling_set_retain_seeds(sampling_options, retain_seeds)
+    cugraph_sampling_set_disjoint_sampling(sampling_options, disjoint_sampling)
 
     error_code = cugraph_heterogeneous_biased_temporal_neighbor_sample(
         c_resource_handle_ptr,
--- a/python/pylibcugraph/pylibcugraph/heterogeneous_uniform_neighbor_sample.pyx
+++ b/python/pylibcugraph/pylibcugraph/heterogeneous_uniform_neighbor_sample.pyx
@@ -42,6 +42,7 @@ from pylibcugraph._cugraph_c.algorithms cimport (
     cugraph_sampling_set_compress_per_hop,
     cugraph_sampling_set_compression_type,
     cugraph_sampling_set_retain_seeds,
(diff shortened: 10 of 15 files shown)


cpp/src/community/legacy/spectral_clustering.cu
```diff
@@ -57,54 +62,61 @@ void balancedCutClustering_impl(raft::handle_t const& handle,
...
-  rmm::device_uvector<unsigned long long> d_seed(1, handle.get_stream());
-
-  raft::random::uniformInt<unsigned long long>(
-    rng_state, d_seed.data(), 1, 0, std::numeric_limits<vertex_t>::max() - 1, handle.get_stream());
-
-  raft::update_host(&seed1, d_seed.data(), d_seed.size(), handle.get_stream());
-  unsigned long long seed2 = seed1 + 1;
+  // Use cuVS for float precision until double precision is supported
+  if constexpr (std::is_same_v<weight_t, float>) {
+    // Convert CSR to COO using thrust
```
aamijar: You might be able to use `raft::sparse::convert::csr_to_coo` here if its more convenient.

cpp/src/community/legacy/spectral_clustering.cu
```diff
@@ -57,54 +62,61 @@ void balancedCutClustering_impl(raft::handle_t const& handle,
...
+    params.n_init       = 10;  // Multiple initializations for better results
+    params.n_neighbors  = std::min(static_cast<int>(graph.number_of_vertices) - 1, 15);  // Adaptive neighbor count
+
+    cuvs::cluster::spectral::fit_predict(
+      handle,
+      params,
+      coo_matrix,
+      raft::make_device_vector_view<vertex_t, vertex_t>(clustering, graph.number_of_vertices));
+  } else {
+    CUGRAPH_FAIL("Double precision spectral clustering not yet supported with cuVS");
```
aamijar: I added `double` support for the cuvs PR so we should be able to try with that.

jnke2016: Thanks. I updated the PR to support double and after merging your latest changes it worked

cpp/src/community/legacy/spectral_clustering.cu
jnke2016: Thanks. I leveraged this raft utils function to convert from `csr` to `coo`

aamijar (approved):
Thanks for incorporating the suggestions!

ChuckHastings (commented):
Looks good, will do a final review after it passes CI

cpp/src/community/legacy/spectral_clustering.cu
```diff
@@ -57,54 +63,54 @@ void balancedCutClustering_impl(raft::handle_t const& handle,
...
 
-  raft::spectral::cluster_solver_config_t<index_type, value_type> clust_cfg{
-    n_clusters, kmean_max_it, kmean_tol, seed2};
-  raft::spectral::kmeans_solver_t<index_type, value_type> cluster_solver{clust_cfg};
+  params.rng_state    = rng_state;
+  params.n_clusters   = n_clusters;
+  params.n_components = n_eig_vects;
+  params.n_init       = 10;  // Multiple initializations for better results
+  params.n_neighbors =
+    std::min(static_cast<int>(graph.number_of_vertices) - 1, 15);  // Adaptive neighbor count
```
aamijar, line 107: Hi @jnke2016, I don't think you are setting the `params.tolerance` parameter anywhere.

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
+#=============================================================================
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
```
bdice: New file, new copyright.
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
```

cpp/CMakeLists.txt
```diff
@@ -41,8 +43,12 @@ option(BUILD_TESTS "Configure CMake to build tests" ON)
 option(USE_RAFT_STATIC "Build raft as a static library" OFF)
 option(CUGRAPH_COMPILE_RAFT_LIB "Compile the raft library instead of using it header-only" ON)
 option(CUDA_STATIC_RUNTIME "Statically link the CUDA toolkit runtime and libraries" OFF)
+option(CUGRAPH_USE_CUVS_STATIC "Build and statically link the cuVS library" ON)
+option(CUGRAPH_COMPILE_CUVS "Always compile cuVS from source" ON)
```
bdice: Match the order of the log messages and the order in which these are used.
```suggestion
option(CUGRAPH_COMPILE_CUVS "Always compile cuVS from source" ON)
option(CUGRAPH_USE_CUVS_STATIC "Build and statically link the cuVS library" ON)
```

python/cugraph/cugraph/tests/community/test_balanced_cut.py
```diff
@@ -98,3 +98,17 @@ def test_edge_cut_clustering_directed_graph(partitions):
     )
     with pytest.raises(ValueError):
         cugraph_call(G_edge, partitions)
```
rlratzel: `cugraph_call` is still calling `cugraph.spectralBalancedCutClustering()` - shouldn't it be updated to call `cugraph.spectralModularityMaximizationClustering()` now?

python/cugraph/pytest.ini
```diff
@@ -65,6 +65,7 @@ filterwarnings =
           ignore:node2vec is deprecated and will be removed in a future release in favor of node2vec_random_walks:FutureWarning
           ignore:random_walks is deprecated and will be removed in the next release in favor of uniform_random_walks:FutureWarning
           ignore:Coalesced path results, returned when setting legacy_result_type=True, is deprecated and will no longer be supported:FutureWarning
+          ignore:spectralBalancedCutClustering is deprecated and will be removed in a future release. Use spectralModularityMaximizationClustering instead.:DeprecationWarning
```
bdice: We always use `FutureWarning` for deprecations, because `DeprecationWarning` is hidden from the user by default.
```suggestion
          ignore:spectralBalancedCutClustering is deprecated and will be removed in a future release. Use spectralModularityMaximizationClustering instead.:FutureWarning
```

python/cugraph/cugraph/community/spectral_clustering.py
```diff
@@ -77,6 +78,11 @@ def spectralBalancedCutClustering(
     >>> df = cugraph.spectralBalancedCutClustering(G, 5)
 
     """
+    warnings.warn(
+        "spectralBalancedCutClustering is deprecated and will be removed in a future "
+        "release. Use spectralModularityMaximizationClustering instead.",
+        DeprecationWarning,
```
bdice: We always use `FutureWarning` for deprecations, because `DeprecationWarning` is hidden from the user by default.
```suggestion
        FutureWarning,
```

python/cugraph/cugraph/tests/community/test_balanced_cut.py
```diff
@@ -98,3 +98,17 @@ def test_edge_cut_clustering_directed_graph(partitions):
...
+def test_spectral_balanced_cut_clustering_deprecation_warning():
+    G = karate.get_graph(
+        create_using=cugraph.Graph(directed=False), ignore_weights=True
+    )
+    warning_msg = (
+        "spectralBalancedCutClustering is deprecated and will be removed in a future "
+        "release. Use spectralModularityMaximizationClustering instead."
+    )
+
+    with pytest.warns(DeprecationWarning, match=warning_msg):
```
bdice: We always use `FutureWarning` for deprecations, because `DeprecationWarning` is hidden from the user by default.
```suggestion
    with pytest.warns(FutureWarning, match=warning_msg):
```

python/libcugraph/pyproject.toml
```diff
@@ -27,6 +27,7 @@ classifiers = [
     "Environment :: GPU :: NVIDIA CUDA",
 ]
 dependencies = [
+    "libcuvs==25.12.*,>=0.0.0a0",
```
bdice: I don't think we need this as a dependency if we're statically linking the library.
```suggestion
```

python/libcugraph/pyproject.toml
```diff
@@ -66,6 +67,7 @@ dependencies-file = "../../dependencies.yaml"
 matrix-entry = "cuda_suffixed=true"
 requires = [
     "cmake>=3.30.4",
+    "libcuvs==25.12.*,>=0.0.0a0",
```
bdice: I don't think we need this as a dependency if we're statically linking the library.
```suggestion
```

python/pylibcugraph/pyproject.toml
```diff
@@ -75,6 +75,7 @@ requires = [
     "cmake>=3.30.4",
     "cython>=3.1.2,<3.2.0a0",
     "libcugraph==25.12.*,>=0.0.0a0",
+    "libcuvs==25.12.*,>=0.0.0a0",
```
bdice: I don't think we need this as a dependency if we're statically linking the library.
```suggestion
```

dependencies.yaml
```diff
@@ -508,6 +513,30 @@ dependencies:
             packages:
               - libraft-cu13==25.12.*,>=0.0.0a0
           - {matrix: null, packages: [*libraft_unsuffixed]}
+  depends_on_libcuvs:
+    common:
+      - output_types: conda
+        packages:
+          - &libcuvs_unsuffixed libcuvs==25.12.*,>=0.0.0a0
```
bdice: Is the intent here to use shared libraries in conda developer environments?

We want static libraries in conda & wheel package builds. Developer environments and devcontainers would depend on shared libraries with the current state here -- but we're setting `option(CUGRAPH_COMPILE_CUVS "Always compile cuVS from source" ON)`, which makes it sound like we won't even use those shared libraries...?

python/cugraph/pytest.ini
rlratzel: Ideally we shouldn't need to do this since it will hide deprecation warnings we need to see (eg. if we call the old API in places we didn't intend to).

python/cugraph/cugraph/tests/community/test_balanced_cut.py
rlratzel: chatted offline with @jnke2016 - this test file has redundant coverage with `test_modularity.py` now, so all but the deprecation warning test will be removed and the ignore line in `pytest.ini` will also be removed.

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
...
+      set(CPM_DOWNLOAD_cuvs ON)
+    endif()
+
+    if(PKG_USE_CUVS_STATIC)
+      set(CUVS_LIB cuvs::cuvs_static PARENT_SCOPE)
+      message(STATUS "CUGRAPH: Using static cuVS library")
+    else()
+      set(CUVS_LIB cuvs::cuvs PARENT_SCOPE)
+      message(STATUS "CUGRAPH: Using shared cuVS library")
+    endif()
```
robertmaynard: No point to this whole block, since the root CMakeLists.txt overwrites the content of `CUVS_LIB`

cpp/CMakeLists.txt
```diff
@@ -41,8 +43,12 @@ option(BUILD_TESTS "Configure CMake to build tests" ON)
 option(USE_RAFT_STATIC "Build raft as a static library" OFF)
 option(CUGRAPH_COMPILE_RAFT_LIB "Compile the raft library instead of using it header-only" ON)
 option(CUDA_STATIC_RUNTIME "Statically link the CUDA toolkit runtime and libraries" OFF)
+option(CUGRAPH_USE_CUVS_STATIC "Build and statically link the cuVS library" ON)
```
robertmaynard: This should be a depedent option https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html on `CUGRAPH_COMPILE_CUVS`

cpp/CMakeLists.txt
```diff
@@ -474,6 +485,15 @@ if(CUGRAPH_COMPILE_RAFT_LIB)
   endif()
 endif()
 
+set(CUVS_LIB )
```
robertmaynard:
```suggestion
unset(CUVS_LIB)
```

cpp/CMakeLists.txt
```diff
@@ -474,6 +485,15 @@ if(CUGRAPH_COMPILE_RAFT_LIB)
...
+if(CUGRAPH_COMPILE_CUVS)
+  if(CUGRAPH_USE_CUVS_STATIC)
+    set(CUVS_LIB cuvs::cuvs_static)
+  else()
+    set(CUVS_LIB cuvs::cuvs)
+  endif()
+endif()
+
```
robertmaynard:
```suggestion
if(CUGRAPH_COMPILE_CUVS AND CUGRAPH_USE_CUVS_STATIC)
  set(CUVS_LIB cuvs::cuvs_static)
elseif(CUGRAPH_COMPILE_CUVS)
  set(CUVS_LIB cuvs::cuvs)
endif()

```

jnke2016:
@robertmaynard right. When I started this PR, I was testing @aamijar fork to make sure the new API is backward compatible with cuGraph with no tests failure before getting his branch merged

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
...
+    if(PKG_CLONE_ON_PIN AND NOT PKG_PINNED_TAG STREQUAL "${rapids-cmake-checkout-tag}")
+        message(STATUS "CUGRAPH: Pinned tag found: ${PKG_PINNED_TAG}. Cloning cuVS locally.")
+        set(CPM_DOWNLOAD_cuvs ON)
+    elseif(PKG_USE_CUVS_STATIC AND (NOT CPM_cuvs_SOURCE))
+      message(STATUS "CUGRAPH: Cloning cuVS locally to build static libraries.")
+      set(CPM_DOWNLOAD_cuvs ON)
+    endif()
+
+    if(PKG_USE_CUVS_STATIC)
+      set(CUVS_LIB cuvs::cuvs_static PARENT_SCOPE)
```
robertmaynard: the `cuvs::cuvs_static` target is a guarded behind the `cuvs_static` component when using a prebuilt package ( aka conda ). So you need to pass `COMPONENTS  cuvs_static` when calling `rapids_cpm_find` when looking for the static version.

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
...
+      set(CUVS_LIB cuvs::cuvs PARENT_SCOPE)
+      message(STATUS "CUGRAPH: Using shared cuVS library")
+    endif()
+
+    # Disable multi-GPU algorithms to avoid NCCL dependency
+    set(CUVS_BUILD_MG_ALGOS OFF)
+
+    rapids_cpm_find(cuvs ${PKG_VERSION}
+      GLOBAL_TARGETS      cuvs::cuvs
+      BUILD_EXPORT_SET    cugraph-exports
```
robertmaynard:
```suggestion
```

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
...
+find_and_configure_cuvs(VERSION          ${CUGRAPH_MIN_VERSION_cuvs}
+                        FORK             rapidsai
+                        PINNED_TAG       ${rapids-cmake-checkout-tag}
+
+                        # When PINNED_TAG above doesn't match cugraph,
+                        # force local cuVS clone in build directory
+                        # even if it's already installed.
+                        CLONE_ON_PIN     ON
+                        USE_CUVS_STATIC  ${CUGRAPH_USE_CUVS_STATIC}
+                        COMPILE_LIBRARY  ${CUGRAPH_COMPILE_CUVS}
```
robertmaynard: Unused variable
```suggestion
```

cpp/cmake/thirdparty/get_cuvs.cmake
```diff
@@ -0,0 +1,71 @@
...
+# cmake-format: off
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+# cmake-format: on
+#=============================================================================
+
+set(CUGRAPH_MIN_VERSION_cuvs "${CUGRAPH_VERSION_MAJOR}.${CUGRAPH_VERSION_MINOR}.00")
+
+function(find_and_configure_cuvs)
+    set(oneValueArgs VERSION FORK PINNED_TAG EXCLUDE_FROM_ALL USE_CUVS_STATIC COMPILE_LIBRARY CLONE_ON_PIN)
```
robertmaynard:
```suggestion
    set(oneValueArgs VERSION FORK PINNED_TAG EXCLUDE_FROM_ALL USE_CUVS_STATIC CLONE_ON_PIN)
```

Unused option

bdice:
I verified that package contents do not contain `libcuvs.so` or `libcuvs_static.a`. That's great!

At first I was going to recommend using conda packages of `libcuvs-static` instead of building libcuvs from source in our conda builds, but it appears that we no longer produce the `libcuvs-static` package. This might be the best we can do?

dependencies.yaml
jnke2016: Yes, the intent is to provide shared libcuvs libraries for conda developer environments. I agree there is a contradiction in the CMake file which has it set to ON by default. As a result those libraries installed but never used.
I think we can change CUGRAPH_COMPILE_CUVS default from ON to OFF, and remove the conditional guard so get_cuvs.cmake always runs. This lets developers use conda's shared libcuvs, while CI/package builds explicitly set the option to ON for static compilation from source.

jnke2016: Does this sound good?

cpp/CMakeLists.txt
```diff
@@ -474,6 +486,16 @@ if(CUGRAPH_COMPILE_RAFT_LIB)
...
+unset(CUVS_LIB)
+if(CUGRAPH_COMPILE_CUVS AND CUGRAPH_USE_CUVS_STATIC)
+  set(CUVS_LIB cuvs::cuvs_static)
+elseif(CUGRAPH_COMPILE_CUVS)
+  set(CUVS_LIB cuvs::cuvs)
+else()
+  # When using prebuilt packages (CUGRAPH_COMPILE_CUVS=OFF), link to shared library
+  set(CUVS_LIB cuvs::cuvs)
+endif()
+
```
robertmaynard:
```suggestion
set(CUVS_LIB cuvs::cuvs)
if(CUGRAPH_USE_CUVS_STATIC)
  set(CUVS_LIB cuvs::cuvs_static)
endif()

```

robertmaynard (changes requested):

rlratzel (approved):
Thanks, @jnke2016  - approving Python changes

cpp/CMakeLists.txt
jnke2016: Right. I believe this is the cause to the `overlinking` error reported in CI. Because of the ambiguity of my branches, it is calling `cuvs::cuvs` (second branch) when it should be calling `cuvs::cuvs_static` resulting in a shared binary dependency. By simplifying the logic and making `CUGRAPH_USE_CUVS_STATIC` the single source of truth for choosing static vs shared. this should resolve the issue and remove the ambiguity.

jnke2016:
@robertmaynard @bdice . I am still getting an `overlinking` error during the conda build.  Can the error be caused by the fact that `rapids_cpm_find` only discovers shared library (libcuvs.so) which forces CMake to fallback to `cuvs::cuvs` (dynamic linking? If it is the case, would building `libcuvs` statically with `BUILD_STATIC_LIBS ON` in rapids_cpm_find enable `libcuvs` to be statically linked to libcugraph through `cuvs::cuvs_static`?

jnke2016:
I am still trying to figure out the source of the `overlinking` issue, and I think it arose when we introduced the COMPONENTS element. Unless I am wrong, there is no `cuvs_static` component defined, and both `cuvs::cuvs` (shared) and `cuvs::cuvs_static` (static) targets are always loaded together in the main/default package (no component specification needed). `cuvs_c` is the only component, which provides the C API [set(cuvs_comp_names c_api)](https://github.com/rapidsai/cuvs/blob/branch-25.12/cpp/CMakeLists.txt#L782). I am not sure what the expected behavior is in this case, but I can speculate. Since the `cuvs_static` component doesn't exist, it falls back to shared → overlinking because libcuvs becomes a runtime dependency but is not listed in the conda package metadata (which expects static linking).

robertmaynard:
> I am still trying to figure out the source of the `overlinking` issue, and I think it arose when we introduced the COMPONENTS element. Unless I am wrong, there is no `cuvs_static` component defined, and both `cuvs::cuvs` (shared) and `cuvs::cuvs_static` (static) targets are always loaded together in the main/default package (no component specification needed). `cuvs_c` is the only component, which provides the C API [set(cuvs_comp_names c_api)](https://github.com/rapidsai/cuvs/blob/branch-25.12/cpp/CMakeLists.txt#L782). I am not sure what the expected behavior is in this case, but I can speculate. Since the `cuvs_static` component doesn't exist, it falls back to shared → overlinking because libcuvs becomes a runtime dependency but is not listed in the conda package metadata (which expects static linking).

Yeah I was wrong about the components. The `cuvs::cuvs_static` target will be brought in with `find_package(cuvs)` whenever it is available

ci/build_wheel.sh
```diff
@@ -30,14 +30,15 @@ rapids-pip-retry wheel \
 sccache --show-adv-stats
 
 EXCLUDE_ARGS=(
+  --exclude "libcuvs.so"
```
bdice: This shouldn't be needed. Wheels should always be statically linking, so shared libraries shouldn't be present in the first place. See cuML's wheel building script, which doesn't need this exclusion. cuGraph should be the same. https://github.com/rapidsai/cuml/blob/main/ci/build_wheel_libcuml.sh

python/cugraph/cugraph/community/spectral_clustering.py
```diff
@@ -74,9 +75,14 @@ def spectralBalancedCutClustering(
     --------
     >>> from cugraph.datasets import karate
     >>> G = karate.get_graph(download=True)
-    >>> df = cugraph.spectralBalancedCutClustering(G, 5)
+    >>> df = cugraph.spectralModularityMaximizationClustering(G, 5)
```
bdice: This is an example for the docs of `spectralBalancedCutClustering`. We shouldn't change the example code here. We could add a deprecation notice in the docs, though.
```suggestion
    >>> df = cugraph.spectralBalancedCutClustering(G, 5)
```

dependencies.yaml
```diff
@@ -103,6 +104,7 @@ files:
     includes:
       - common_build
       - depends_on_libcugraph
+      - depends_on_libcuvs
```
bdice: Wheel builds shouldn't depend on libcuvs libraries, it should be fetched and built statically by CMake.
```suggestion
```

dependencies.yaml
```diff
@@ -152,13 +154,15 @@ files:
     includes:
       - common_build
       - depends_on_librmm
+      - depends_on_libcuvs
```
bdice: Wheel builds shouldn't depend on libcuvs libraries, it should be fetched and built statically by CMake.
```suggestion
```

dependencies.yaml
```diff
@@ -152,13 +154,15 @@ files:
...
       - depends_on_libraft
   py_run_libcugraph:
     output: pyproject
     pyproject_dir: python/libcugraph
     extras:
       table: project
     includes:
+      - depends_on_libcuvs
```
bdice: We do not want a runtime dependency on libcuvs, that's the point of statically linking.
```suggestion
```

dependencies.yaml
```diff
@@ -177,6 +181,7 @@ files:
     includes:
       - common_build
       - depends_on_libcugraph
+      - depends_on_libcuvs
```
bdice: Wheel builds shouldn't depend on libcuvs libraries, it should be fetched and built statically by CMake.
```suggestion
```

dependencies.yaml
bdice: Here's what I think you want:
- conda developer environments: **depend on shared library**
- conda devcontainers: **depend on shared library**
- conda recipes: **fetch with CMake, build and link libcuvs statically**
  - No dependency in `host` or `run`
  - Note: this differs from cuML, which depends on a shared library here. This can go either way in my opinion.
- wheel builds: **fetch with CMake, build and link libcuvs statically**
- wheel at runtime: **no dependency (it's statically linked)**
- pip devcontainers: **???**
  - This one we might want to discuss. Probably fetch with CMake, build and link libcuvs statically? I haven't checked what this PR does right now.

python/cugraph/cugraph/community/spectral_clustering.py
jnke2016: > We shouldn't change the example code here. We could add a deprecation notice in the docs, though.

Right but this is what I did initially and the pytest doctests would fail unless that warning failure get skipped. @rlratzel any thoughts?

bdice: You need something like this: https://github.com/rapidsai/cudf/blob/cff81d86667c1a0fbcbf7887c28e640a29bc7dfd/python/cudf/cudf/tests/test_doctests.py#L16

rlratzel: We should probably just remove the example entirely.  I think an example is useful for someone trying to learn how to add this function to their code, which we/they don't want since it's going away. Make sure the docstring is updated to indicate it's deprecated and direct them to the new API they should call.

cpp/CMakeLists.txt
jnke2016: @robertmaynard, @bdice  , just for my understanding, is there a reason why we don't statically link when using prebuilt packages (cuML also adopted this path). However, I believe conda libcuvs provides both static and shared libraries.

python/cugraph/cugraph/community/spectral_clustering.py
bdice: You should still add the ignores I mentioned above because this will apply to _all_ deprecated APIs in the future.

cpp/CMakeLists.txt
bdice: Oh gosh. It appears `libcuvs` conda packages do ship both `libcuvs.so` and `libcuvs_static.a`. This seems undesirable from a binary size perspective.

@robertmaynard Do you know if is this intentional? Most of RAPIDS does not ship static libraries in conda packages.

```bash
$ wget https://anaconda.org/rapidsai-nightly/libcuvs/25.12.00a105/download/linux-64/libcuvs-25.12.00a105-cuda13_251113_9953f9b8.conda
$ cph extract libcuvs-25.12.00a105-cuda13_251113_9953f9b8.conda
$ ls -alh libcuvs-25.12.00a105-cuda13_251113_9953f9b8/lib/
total 1.1G
drwxrwx--- 3 bdice bdice 4.0K Nov 17 12:29 .
drwxrwx--- 5 bdice bdice 4.0K Nov 17 12:29 ..
drwxrwx--- 4 bdice bdice 4.0K Nov 17 12:29 cmake
-rwxr-x--- 1 bdice bdice 465M Nov 13 16:18 libcuvs.so
-rwxr-x--- 1 bdice bdice 1.2M Nov 13 16:18 libcuvs_c.so
-rw-r----- 1 bdice bdice 612M Nov 13 16:18 libcuvs_static.a
```

bdice: Update: the static libraries are being split out into a new package in https://github.com/rapidsai/cuvs/pull/1494.

ci/build_wheel.sh
jnke2016: I am confused. Aren't wheels using prebuilt packages and link dynamically (hence we have libcuvs.so in the build artifact). Am I missing something?

jnke2016: > Wheels should always be statically linking, so shared libraries shouldn't be present in the first place

In that case, I should update the wheels to compile and build from source instead of using prebuilt packages?

jnke2016: Got it now after reviewing cuML. Thanks

dependencies.yaml
jnke2016: Got it. I understand what you mean here based on [this](https://github.com/rapidsai/cugraph/pull/5326#discussion_r2534679567)

robertmaynard:
@jnke2016 Are you still planning on getting this into 25.12? If so please don't merge in `main` anymore and re-target `release/25.12` as soon as possible.

jnke2016:
> Are you still planning on getting this into 25.12? If so please don't merge in main anymore and re-target release/25.12 as soon as possible.

Got it @robertmaynard . @ChuckHastings  can you retarget it please?

ChuckHastings:
> > Are you still planning on getting this into 25.12? If so please don't merge in main anymore and re-target release/25.12 as soon as possible.
> 
> Got it @robertmaynard . @ChuckHastings can you retarget it please?

Done

jnke2016:
> pip devcontainers: ???
This one we might want to discuss. Probably fetch with CMake, build and link libcuvs statically? I haven't checked what this PR does right now.

@bdice Good catch thanks. I didn't have a devcontainers section. I just added it and it installs pre-built libcuvs and link dynamically

python/cugraph/cugraph/community/spectral_clustering.py
rlratzel: > You should still add the ignores I mentioned above because this will apply to _all_ deprecated APIs in the future.

I'm thinking that ignoring errors from deprecated APIs in doctests is not good though, and perhaps we should just adopt the policy we're doing here (remove example docstring, ensure new API is referenced) for all deprecated APIs in the future.

I'm concerned the ignores will mask other API deprecations we need to know about (eg. an example docstring uses a deprecated cuDF API, or a cugraph docstring for function A uses deprecated cugraph API B as part of the example and should be updated).

dependencies.yaml
```diff
@@ -37,6 +38,31 @@ files:
       - test_python_common
       - test_python_cugraph
       - test_python_pylibcugraph
+  devcontainers:
```
bdice: This `devcontainers` key isn't doing anything right now. To make it take effect, you'd need to update the `devcontainers` repo like this: https://github.com/rapidsai/devcontainers/blob/f9a516d1e4cd9b669845527d4ae13a4fdc720933/features/src/rapids-build-utils/opt/rapids-build-utils/manifest.yaml#L186-L187

However, why do we need this? I think there should be a solution that doesn't require a new dependency file key.

Look at how cuML is solving this: https://github.com/rapidsai/cuml/blob/6b42da514957a909a6d9d8921bad9d705161d8d8/dependencies.yaml#L684-L688

cuML only defines `libcuvs` for `conda` output types and not `pyproject` or `requirements`. I think that's probably what you want here too. Following the other discussion where we broke down the [expected behavior by build type](https://github.com/rapidsai/cugraph/pull/5326/files#r2534712915), you probably never want libcuvs wheels because you'll always build and statically link libcuvs. The only cases where you want `libcuvs` packages are for conda developer environments and devcontainers (but not conda recipes!).

I propose:
- Delete the `devcontainers` key here
- Modify `depends_on_libcuvs` below to only have conda outputs, matching cuML

python/cugraph/cugraph/community/spectral_clustering.py
bdice: > I'm concerned the ignores will mask other API deprecations we need to know about

If it helps, `ignore::FutureWarning:cugraph` will ignore just the deprecations in the cugraph module, or you can use `ignore:MESSAGE_REGEX:FutureWarning:cugraph` with some regex string to ignore specific (known) deprecations from cugraph.

dependencies.yaml
jnke2016: So if I understand correctly, you are suggesting to modify depends_on_libcuvs to only output for conda (not pyproject or requirements) because:

- Conda developer environments & devcontainers automatically get libcuvs from the all: section
- Wheel builds & pip devcontainers don't get libcuvs because pyproject outputs won't include it (statically link)

jnke2016: Also:

- Conda recipes don't get libcuvs as a runtime dependency (they also statically link cuVS)

bdice (approved):
Perfect. Approved!

PR #5328, opened 2025-10-31
Time filtering hop 0

### LINKED ISSUE #5263: [FEA] Support Time Filtering in Hop 0 During Temporal Sampling
Currently, hop 0 vertices are allowed to select any edge during sampling.  However, this conflicts with the PyG API, especially when performing edge-based sampling, since the API allows for overrides in hop 0.  Without support for hop 0 filtering, the accuracy of edge-based sampling in cuGraph-PyG will probably worse than the accuracy in native PyG.

Supporting this involves the following changes to the API:
1. Allow specifying a "time override" for each seed in hop 0
2. When a time override is specified, use those times as the constraints for edge selection in hop 0.  After hop 0, the overrides should be ignored.

ChuckHastings:
Added support for time filtering in hop 0 of temporal sampling.  If starting times are provided in the API for temporal sampling then we do filtering in that hop, if they are not provided then time filtering only occurs later (which is the previous behavior).

Refactored the C API testing for sampling as things were getting to complicated.

Closes #5263 

Change is breaking since we are changing the public API for temporal sampling and deleting some previously deprecated functions.


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -690,6 +692,7 @@ homogeneous_uniform_temporal_neighbor_sample(
   edge_property_view_t<edge_t, edge_time_t const*> edge_start_time_view,
   std::optional<edge_property_view_t<edge_t, edge_time_t const*>> edge_end_time_view,
   raft::device_span<vertex_t const> starting_vertices,
+  std::optional<raft::device_span<edge_time_t const>> starting_vertex_times,
```
seunghwak, line 486: Something very minor, but `"edge_time"_t` here and `starting_"vertex_time"s` here sounds mismatching...

As we use the same data type for both vertices and edges, should we better rename eddge_time_t to something like time_t or time_stamp_t?

cpp/src/c_api/neighbor_sampling.cpp
```diff
@@ -341,19 +341,19 @@ struct uniform_neighbor_sampling_functor : public cugraph::c_api::abstract_funct
...
+          cugraph::sort_sampled_edgelist(
+            handle_,
+            std::move(sampled_srcs),
+            std::move(sampled_dsts),
+            std::move(output_edge_properties),
+            std::move(hop),
+            offsets ? std::make_optional(
+                        raft::device_span<size_t const>{offsets->data(), offsets->size()})
+                    : std::nullopt,
+            edge_label ? edge_label->size() : size_t{1},  // wrong value...
```
seunghwak: What do you mean by `// wrong value...` here? I think this needs little more explanation for anyone who will read this code in the future.

alexbarghi-nv (approved):
👍

cpp/src/c_api/neighbor_sampling.cpp
```diff
@@ -1072,9 +280,9 @@ struct neighbor_sampling_functor : public cugraph::c_api::abstract_functor {
               raft::host_span<const int>(fan_out_->as_type<const int>(), fan_out_->size_),
               num_edge_types_,
               cugraph::sampling_flags_t{options_.prior_sources_behavior_,
-                                        options_.return_hops_,
-                                        options_.dedupe_sources_,
-                                        options_.with_replacement_},
+                                        options_.return_hops_ == TRUE,
+                                        options_.dedupe_sources_ == TRUE,
+                                        options_.with_replacement_ == TRUE},
```
seunghwak, line 285: Another tedious thing... but why are we using `TRUE` instead of the standard `true`? Aren't these variables bool type variables?

cpp/src/c_api/sampling_result.cpp
```diff
@@ -258,6 +258,11 @@ extern "C" cugraph_type_erased_device_array_view_t* cugraph_sample_result_get_re
   const cugraph_sample_result_t* result)
 {
   auto internal_pointer = reinterpret_cast<cugraph::c_api::cugraph_sample_result_t const*>(result);
+
+  if (internal_pointer->renumber_map_ != nullptr) {
+    auto xxx = reinterpret_cast<cugraph::c_api::cugraph_type_erased_device_array_t*>(
+      internal_pointer->renumber_map_);
```
seunghwak: What's the purpose of this code? We are not using `xxx` anywhere...

cpp/src/sampling/detail/shuffle_and_organize_output.cu
```diff
@@ -130,25 +134,31 @@ shuffle_and_organize_output(
...
-                          offsets->begin());
+    thrust::transform(
+      handle.get_thrust_policy(),
+      unique_labels.begin(),
+      unique_labels.end(),
+      offsets->begin(),
+      [d_labels = labels->data(), labels_size = labels->size()] __device__(int32_t label) {
+        return thrust::distance(
+          d_labels, thrust::lower_bound(thrust::seq, d_labels, d_labels + labels_size, label));
+      });
```
seunghwak: Can't we just call the vectorized version of thrust::lower_bound?

cpp/src/sampling/detail/shuffle_and_organize_output.cu
```diff
@@ -130,25 +134,31 @@ shuffle_and_organize_output(
...
 
-    thrust::exclusive_scan(
-      handle.get_thrust_policy(), offsets->begin(), offsets->end(), offsets->begin());
-    labels = std::move(unique_labels);
+    size_t last_offset = labels->size();
+    offsets->set_element_async(num_unique_labels, last_offset, handle.get_stream());
```
seunghwak, line 158: You need to call handle.sync_stream() here. `last_offset` will go out-of-scope before set_element_async finishes and the memory can be reclaimed.

cpp/src/sampling/temporal_sampling_impl.hpp
```diff
@@ -751,12 +790,18 @@ temporal_neighbor_sample_impl(
 
   std::optional<rmm::device_uvector<size_t>> result_offsets{std::nullopt};
 
+  // OK... got a problem here.  For the result_offsets, we need to store the offset for each label,
+  // but if a label stops early we're not computing this properly.  Need to investigate.
```
seunghwak: I am not fully understanding this. What really is the problem here? And don't we need to solve this now? And better add FIXME: if we are not going to solve this in this PR?

seunghwak (commented):
Mostly looks good but it seems like there are few things to address before merging this.

cpp/src/c_api/neighbor_sampling.cpp
ChuckHastings: Crossing between C and C++.  The C API defines its own bool_t as an enum taking values TRUE and FALSE.  So this is comparing against the enum values.

I'm honestly not sure about compatibility between the C99 addition of _Bool, the C23 addition of bool and C++ bool variables (or if we have C23 support in all of our compilers yet).

If we want to explore that, we should do that as part of a separate issue/PR as it would affect all of the C API functions.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: I'd be afraid of using `time_t` and getting it confused with the time_t in the std namespace.  We could use `time_stamp_t`.  Should probably make that change universally through the code since it would need to be consistent everywhere.  I'd be inclined to do that as a separate PR, but I can work on it quickly after this one.

cpp/src/c_api/neighbor_sampling.cpp
ChuckHastings: This comment is obsolete, the code changed since your first review.  There is a spurious comment that's obsolete in the latest code a bit further down in the file, and I'll push an update to remove that as well.

cpp/src/c_api/sampling_result.cpp
ChuckHastings: Debugging code.  Will delete.

cpp/src/sampling/detail/shuffle_and_organize_output.cu
ChuckHastings: Yes, I think I tried some different things before settling on this, didn't make the leap of using the vectorized version.  Will fix.

cpp/src/sampling/detail/shuffle_and_organize_output.cu
ChuckHastings: Fixed

cpp/src/sampling/temporal_sampling_impl.hpp
ChuckHastings: Deleted.  This was a comment to me while debugging.  The changes above to use the thrust::lower_bound above fixed that problem, I missed deleting the comment.

jnke2016 (approved):
Looks good to me

seunghwak (approved):
LGTM

PR #5333, opened 2025-11-04
Fixes networkx version checks for skipping certain tests.

rlratzel:
Updates minimum networkx version check on various BC tests to use `packaging.version` to provide more robust checks.

(also updates utility used by PropertyGraph to accommodate [cudf breaking change](https://github.com/rapidsai/cudf/pull/20306) in order to pass CI - perhaps that should be a separate PR?)

### DIFF
--- a/python/cugraph/cugraph/utilities/utils.py
+++ b/python/cugraph/cugraph/utilities/utils.py
@@ -7,6 +7,7 @@ import shutil
 
 import cudf
 from cudf.core.column import as_column
+import pylibcudf as plcudf
 
 from warnings import warn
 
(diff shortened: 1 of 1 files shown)


python/cugraph/cugraph/tests/centrality/test_batch_betweenness_centrality_mg.py
```diff
@@ -15,12 +16,12 @@
     compare_scores,
 )
 
+nx_version_3_5 = packaging.version.parse("3.5")
+nx_version = packaging.version.parse(nx.__version__)
```
KyleFromNVIDIA: This logic gets repeated across multiple files. I would instead suggest using a [custom PyTest marker](https://docs.pytest.org/en/stable/example/markers.html).

Add the following to `pytest.ini`:

```
[pytest]
markers =
    requires_nx_3_5: requires nx version 3.5
```

Add the following to `conftest.py`:

```python
import packaging.version


nx_version_3_5 = packaging.version.parse("3.5")
nx_version = packaging.version.parse(nx.__version__)

def pytest_runtest_setup(item):
    if nx_version < nx_version_3_5 and any(item.iter_markers(name="requires_nx_3_5")):
        pytest.skip(f"Requires networkx >= {nx_version_3_5}")
```

And then finally use this custom marker in your tests:

```python
@pytest.mark.requires_nx_3_5
def test_stuff(): ...
```

KyleFromNVIDIA: Actually, here's a more general approach. Add the following to `pytest.ini`:

```
[pytest]
markers =
    requires_nx: requires networkx version
```

Add the following to `conftest.py`:

```python
import packaging.version


nx_version = packaging.version.parse(nx.__version__)

def pytest_runtest_setup(item):
    for mark in filter(lambda mark: packaging.version.parse(mark.kwargs["version"]) > nx_version, item.iter_markers(name="requires_nx")):
        pytest.skip(f"Requires networkx >= {mark.kwargs['version']}")
```

And use it on your tests:

```python
@pytest.mark.requires_nx(version="3.5")
def test_stuff(): ...
```

rlratzel: Thanks, @KyleFromNVIDIA
I applied a slightly modified version of your suggestion that uses a post collection hook instead of a setup hook since we have all the info we need to determine if we're skipping a test at collection time, to avoid repeated calls. I tested it with both v3.5 and older versions of NX installed and it worked as expected.

PR #5334, opened 2025-11-05
Use ruff-check, ruff-format instead of flake8, black

KyleFromNVIDIA:
Also bump pre-commit-hooks to version 1.2.1.

Issue: https://github.com/rapidsai/build-planning/issues/130
Ops-Bot-Merge-Barrier: true


python/cugraph/cugraph/structure/graph_implementation/simpleDistributedGraph.py
```diff
@@ -257,7 +257,7 @@ def __from_edgelist(
                 ]
                 value_col_names = [self.edgeWeightCol, self.edgeIdCol, self.edgeTypeCol]
             else:
-                raise ValueError("Only 1 or 3 values may be provided" "for edge_attr")
+                raise ValueError("Only 1 or 3 values may be providedfor edge_attr")
```
acostadon: space missing between provided and for

python/cugraph/cugraph/tests/centrality/test_katz_centrality.py
```diff
@@ -123,14 +123,12 @@ def test_katz_centrality_toy(graph_file):
         expected_score = centralities[vertex]
         actual_score = ck["katz_centrality"].iloc[vertex]
         assert pytest.approx(expected_score, abs=1e-2) == actual_score, (
-            f"Katz centrality score is {actual_score}, should have"
-            f"been {expected_score}"
+            f"Katz centrality score is {actual_score}, should havebeen {expected_score}"
```
acostadon: missing space between words

python/cugraph/cugraph/datasets/dataset.py
```diff
@@ -147,7 +147,7 @@ def __download_csv(self, url):
 
         else:
             raise RuntimeError(
-                f"The directory {self._dl_path.path.absolute()}" "does not exist"
+                f"The directory {self._dl_path.path.absolute()}does not exist"
```
bdice:
```suggestion
                f"The directory {self._dl_path.path.absolute()} does not exist"
```

python/cugraph/cugraph/tests/community/test_leiden.py
```diff
@@ -38,43 +54,402 @@
     "data_2": {
         "graph": {
             # fmt: off
```
bdice: Seems like this `# fmt: off` was ignored?

bdice (approved):
Two suggestions, otherwise LGTM.

acostadon (approved):
Looks good. You just fixed the comments I had.
I don't like the global replacement if single ticks with backslash double quote but I guess that is a standard.

python/cugraph/cugraph/tests/community/test_leiden.py
KyleFromNVIDIA: From the [ruff-format docs](https://docs.astral.sh/ruff/formatter/#format-suppression):

> `# fmt: on` and `# fmt: off` comments are enforced at the statement level:
> 
> ```python
> # fmt: off
> not_formatted=3
> also_not_formatted=4
> # fmt: on
> ```
>
> As such, adding `# fmt: on` and `# fmt: off` comments within expressions will have no effect. In the following example, both list entries will be formatted, despite the `# fmt: off`:
>
> ```python
> [
>     # fmt: off
>     '1',
>     # fmt: on
>     '2',
> ]
> ```
>
> Instead, apply the `# fmt: off` comment to the entire statement:
>
> ```python
> # fmt: off
> [
>     '1',
>     '2',
> ]
> # fmt: on
> ```

Updating accordingly.

KyleFromNVIDIA:
I think it comes from the fact that Python prefers double quotes over single combined with the fact that notebooks aren't really meant to be read and edited by a human, at least not at the JSON level. If you look at the notebooks in a notebook editor, they look fine - no backslashes.

rlratzel:
Thanks @KyleFromNVIDIA . What's the policy when `ruff` finds issues `ruff-format` can't fix?  I browsed https://github.com/rapidsai/build-planning/issues/130 but didn't see.  Just asking because I see changes like this:
```python
   names = (["src", "dst", "weight"],)  # noqa: F841
```
(unused variable) where I might have just removed the unused variable instead of adding the `noqa` comment.  Should we (library devs) just take care of all the `noqa` comments we don't want to keep in a separate PR?

KyleFromNVIDIA:
My approach is to fix things if there's a clear and obvious fix, and if not, maintain the status quo to avoid accidentally breaking something or tripping someone up. If that `names` variable is truly not needed (I assumed it was, if you look a few lines below, you can see it was once used but was commented out) , I can remove it in this PR, or you can remove it in a follow-up.

rlratzel:
> My approach is to fix things if there's a clear and obvious fix, and if not, maintain the status quo to avoid accidentally breaking something or tripping someone up. If that `names` variable is truly not needed (I assumed it was, if you look a few lines below, you can see it was once used but was commented out) , I can remove it in this PR, or you can remove it in a follow-up.

Gotcha, I trust your judgement, I just wanted to know if the approach was to simply `noqa` everything for later, in which case we should plan for a significant followup PR ;)
 
Since it sounds like the egregious/obvious ones are all taken care of, I'll just plan on addressing any others as part of other PRs over time. And now that I look more, I don't see many at all.

rlratzel (approved):
Nothing is jumping out as a problem, LGTM, thanks!

PR #5339, opened 2025-11-07
rename edge_time_t to time_stamp_t

### LINKED ISSUE #5338: Rename edge_time_t to time_stamp_t
Something very minor, but `"edge_time"_t` here and `starting_"vertex_time"s` here sounds mismatching...

As we use the same data type for both vertices and edges, should we better rename eddge_time_t to something like time_t or time_stamp_t?

_Originally posted by @seunghwak in https://github.com/rapidsai/cugraph/pull/5328#discussion_r2482856106_

ChuckHastings:
Closes #5338


cpp/include/cugraph/graph_functions.hpp
```diff
@@ -418,7 +418,7 @@ decompress_to_edgelist(
  * @tparam edge_t Type of edge identifiers. Needs to be an integral type.
  * @tparam weight_t Type of edge weights. Needs to be a floating point type.
  * @tparam edge_type_t Type of edge type identifiers. Needs to be an integral type.
- * @tparam edge_time_t Type of edge time. Needs to be an integral type.
+ * @tparam time_stamp_t Type of edge time. Needs to be an integral type.
```
seunghwak: "Type of edge time"=>Shouldn't we update this as well?

seunghwak (approved):
Otherwise, LGTM

cpp/include/cugraph/graph_functions.hpp
ChuckHastings: Fixed the documentation.

PR #5345, opened 2025-11-14
[FEA] Update pylibcugraph to support different temporal comparisons

alexbarghi-nv:
Updates pylibcugraph to support different temporal comparisons (i.e. >=, <, <=, >, last).


jnke2016 (approved):

ChuckHastings (approved):

python/pylibcugraph/pylibcugraph/heterogeneous_uniform_temporal_neighbor_sample.pyx
```diff
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+    # SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
```
rlratzel: I don't think this is supposed to be indented?

rlratzel (changes requested):
Can you add at least one test that uses the new arguments?

Also, is using strings `>`, `<`, etc. as the arg value a common convention that is already familiar to callers?  Otherwise, it seems like it would be much more self-documenting to take strings like "strictly_increasing", "monotonically_decreasing", etc. (or some abbreviation of those).

alexbarghi-nv:
I think that's fair, I'll add a test and change the argument names.

rlratzel (approved):
Thanks

PR #5354, opened 2025-11-25
Main merge release/25.12 (unblock forward merge)

alexbarghi-nv:
Unblocks the forward merge.

PR #5355, opened 2025-11-26
Requires NX 3.5 for edge BC tests

rlratzel:
* Requires that edge BC tests use NX 3.5 for comparisons, otherwise those tests are XFAIL'd.
  * Changes to NX 3.6 caused some comparisons in tests to fail when NX 3.6 is used as the reference. We're essentially skipping for now but will look into those changes (likely [here](https://github.com/networkx/networkx/pull/8256))
* This also updates the `requires_nx` marker to take up to 2 args for min and max required version, and a flag for skip vs. xfail.

### DIFF
--- a/notebooks/algorithms/centrality/Degree.ipynb
+++ b/notebooks/algorithms/centrality/Degree.ipynb
@@ -150,7 +150,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "G = karate.get_graph()"
+    "G = karate.get_graph(download=True)"
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


BradReesWork (approved):

alexbarghi-nv:
@rlratzel will this unblock https://github.com/rapidsai/cugraph/pull/5354?

ChuckHastings:
This is consistently failing a notebook test.

acostadon (approved):
looks good

rlratzel:
> @rlratzel will this unblock #5354?

Yes I believe it should.

PR #5356, opened 2025-11-26
CUDA 13: use releases, not nightlies, for PyTorch wheels

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

Follow-up to #5236

Similar to https://github.com/rapidsai/cugraph-gnn/pull/355

PyTorch now has official releases with CUDA 13 support. This switches from nightlies to those releases for wheels in CI here.


jameslamb:
Moving this to draft since it's non-urgent and doesn't have reviews yet anyway.

I'll use it to test https://github.com/rapidsai/ci-imgs/pull/333

jameslamb:
Re-ran `check-nightly-ci` now that the nightly tests are passing ([build from 12ish hours ago](https://github.com/rapidsai/cugraph/actions/runs/19923235894)), I'm going to merge this 😁

PR #5375, opened 2025-12-17
Proclaim return types and other fixes needed for CCCL 3.2

bdice:
Fixes needed for CCCL 3.2 compatibility.

### DIFF
--- a/cpp/src/traversal/od_shortest_distances_impl.cuh
+++ b/cpp/src/traversal/od_shortest_distances_impl.cuh
@@ -273,8 +273,12 @@ __global__ static void multi_partition_copy(
       if (tmp_idx < num_elems) {
         auto partition    = partition_op(*(input_first + tmp_idx));
         tmp_partitions[i] = partition;
-        tmp_offsets[i]    = tmp_counts[partition];
-        ++tmp_counts[partition];
+        // Skip count update for discarded elements (partition == max_num_partitions)
+        // to avoid out-of-bounds access on tmp_counts array
+        if (partition != static_cast<uint8_t>(max_num_partitions)) {
+          tmp_offsets[i] = tmp_counts[partition];
+          ++tmp_counts[partition];
+        }
       }
       tmp_idx += gridDim.x * blockDim.x;
     }


cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -795,18 +796,21 @@ void multisource_backward_pass(
   auto v_first = graph_view.local_vertex_partition_range_first();
 
   // Calculate offsets for each distance level in the consecutive arrays
-  std::vector<size_t> h_distance_offsets(global_max_distance + 1);
+  // Need global_max_distance + 2 elements: one for each distance level (0 to global_max_distance)
+  // plus a sentinel at the end for CUB segmented sort end offsets
+  std::vector<size_t> h_distance_offsets(global_max_distance + 2);
```
bdice, line 801: This appears to be an off-by-one error that results in an illegal memory access in `SortPairs`.

cpp/src/centrality/betweenness_centrality_impl.cuh
```diff
@@ -873,7 +877,11 @@ void multisource_backward_pass(
     // Allocate temporary storage for CUB segmented sort
     rmm::device_uvector<std::byte> d_tmp_storage(0, handle.get_stream());
 
-    // Process each chunk - sort consecutive arrays directly in-place
+    // Allocate output buffers for CUB sort (input/output cannot overlap)
+    rmm::device_uvector<vertex_t> sorted_vertices(total_vertices, handle.get_stream());
+    rmm::device_uvector<origin_t> sorted_sources(total_vertices, handle.get_stream());
```
bdice, line 882: Tests were failing so I was investigating this file, and I think this comment points to a violation of the CUB API preconditions. The inputs and outputs of the `cub::DeviceSegmentedSort::SortPairs` call overlapped. See [the docs for cub::DeviceSegmentedSort::SortPairs](https://nvidia.github.io/cccl/cub/api/structcub_1_1DeviceSegmentedSort.html#_CPPv4I0000EN3cub19DeviceSegmentedSort9SortPairsE11cudaError_tPvR6size_tPK4KeyTP4KeyTPK6ValueTP6ValueTN4cudaSt7int64_tEN4cudaSt7int64_tE20BeginOffsetIteratorT18EndOffsetIteratorT12cudaStream_t).

As a workaround, I am initializing an output buffer for the sorted results, and then moving it back to the original variable once computation is complete.

alexbarghi-nv (approved):
Approved but should be re-reviewed by @ChuckHastings or @seunghwak when they're back from vacation.

bdice:
Thanks! I agree -- these should be reviewed more but I don't want to break cuGraph with the CCCL 3.2 update. Thanks for taking a look and discussing offline @alexbarghi-nv and @rlratzel.

PR #5379, opened 2025-12-22
Specify argument types inside proclaim_return_type

vyasr:
This seems to be necessary to make the type inference happy.


jameslamb:
Admin-merging this based on offline discussion with @vyasr and @bdice . We think this will get nightlies working here in `cugraph` again, which should hopefully unblock nightlies of everything downstream of it.

cc @rapidsai/cugraph-cpp-codeowners please do take a look when you're available, we have plenty of time to revert any of these changes before the next release if you see issues.

PR #5386, opened 2026-01-07
Fix negative sampling with biases

ChuckHastings:
Original implementation required the biases to match the type of the edge weights in the graph.  But creating an unweighted graph defaults the weight type to float.  There are plenty of use cases where we don't need these to match.

This PR changes the implementation to allow for biases to be specified either float or double regardless of the weight type - since the weight is never used as part of this algorithm it's irrelevant if they match or not.


alexbarghi-nv (approved):
👍

jnke2016 (approved):

seunghwak (approved):
LGTM

PR #5394, opened 2026-01-13
Replace deprecated thrust facilities in favor of `libcu++` ones

miscco:
We have cleaned up a lot of header includes in thrust recently and it seems that cugraph was missing some includes.

Take that opportunity and replace some deprecated thrust features with libcu++ ones


miscco:
🙀 https://github.com/rapidsai/cugraph/actions/runs/20986959852/job/60325392604?pr=5394#step:12:957

I am a bit unsure how that would be related

PR #5383, opened 2026-01-05
build and test against CUDA 13.1.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/236

Tests that CI here will work with the changes from https://github.com/rapidsai/shared-workflows/pull/483,
switches CUDA 13 builds to CUDA 13.1.0 and adds some CUDA 13.1.0 test jobs.

PR #5395, opened 2026-01-13
Avoid modifying a std::move-ed out object

seunghwak:
Modifying a std::move-ed out object is an undefined behavior. To re-use the move-ed out object, it should be re-initialized. We may pull these updates to #5383 and just close this PR.

### DIFF
--- a/cpp/src/traversal/od_shortest_distances_impl.cuh
+++ b/cpp/src/traversal/od_shortest_distances_impl.cuh
@@ -912,8 +912,8 @@ rmm::device_uvector<weight_t> od_shortest_distances(
                                   h_split_thresholds.size(),
                                   handle.get_stream());
 
-              rmm::device_uvector<key_t> tmp_buffer = std::move(far_buffers.back());
-              far_buffers.back() = rmm::device_uvector<key_t>(0, handle.get_stream());
+              auto tmp_buffer = rmm::device_uvector<key_t>(0, handle.get_stream());
+              std::swap(tmp_buffer, far_buffers.back());
               std::vector<key_t*> h_buffer_ptrs(h_split_thresholds.size() + 1);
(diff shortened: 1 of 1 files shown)


cpp/src/traversal/od_shortest_distances_impl.cuh
```diff
@@ -913,6 +913,7 @@ rmm::device_uvector<weight_t> od_shortest_distances(
                                   handle.get_stream());
 
               rmm::device_uvector<key_t> tmp_buffer = std::move(far_buffers.back());
+              far_buffers.back() = rmm::device_uvector<key_t>(0, handle.get_stream());
```
bdice: This is the cleanest way I know to write this:
```suggestion
              auto tmp_buffer = rmm::device_uvector<key_t>(0, handle.get_stream());
              std::swap(tmp_buffer, far_buffers.back());
```

cpp/src/traversal/od_shortest_distances_impl.cuh
```diff
@@ -995,6 +996,7 @@ rmm::device_uvector<weight_t> od_shortest_distances(
                                      handle.get_stream());
             } else {
               far_buffers[i - num_near_q_insert_buffers] = std::move(far_buffers[i]);
+              far_buffers[i] = rmm::device_uvector<key_t>(0, handle.get_stream());
```
bdice: Similar suggestion to the above. Reuse the existing `device_uvector`. Swap them and then resize.
```suggestion
              std::swap(far_buffers[i - num_near_q_insert_buffers], far_buffers[i]);
              far_buffers[i].resize(0, handle.get_stream());
```

seunghwak:
#5383 includes this fix, just closing the PR

PR #5399, opened 2026-01-15
wheel builds: react to changes in pip's handling of build constraints

mmccarty:
## Description
Contributes to https://github.com/rapidsai/build-planning/issues/242

Modifying `ci/build_wheel.sh` to

- pass`--build-constraint="${PIP_CONSTRAINT}"` unless build isolation is enabled.
- unset `PIP_CONSTRAINT` (set by rapids-init-pip)... it doesn't affect builds as of pip 25.3, and results in an error from `pip wheel` when set and `--build-constraint` is also passed

### DIFF
--- a/ci/test_python.sh
+++ b/ci/test_python.sh
@@ -86,19 +86,18 @@ rapids-logger "pytest cugraph (not mg, with xdist)"
   --cov-report=term
 
 # excludes known failures that will always fail when run in combination
-# FIXME: temporarily disables MG tests due to hang in CI with CUDA 13.1.0
-rapids-logger "pytest cugraph (mg, with xdist) - SKIPPING DUE TO HANG IN CI WITH CUDA 13.1.0"
-#./ci/run_cugraph_pytests.sh \
-#  --verbose \
-#  --junitxml="${RAPIDS_TESTS_DIR}/junit-cugraph.xml" \
-#  --numprocesses=8 \
-#  --dist=worksteal \
-#  -m "mg" \
-#  -k "not test_dist_sampler_mg" \
-#  --cov-config=../../.coveragerc \
-#  --cov=cugraph \
-#  --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cugraph-coverage.xml" \
-#  --cov-report=term
+rapids-logger "pytest cugraph (mg, with xdist)"
+./ci/run_cugraph_pytests.sh \
+  --verbose \
+  --junitxml="${RAPIDS_TESTS_DIR}/junit-cugraph.xml" \
+  --numprocesses=8 \
+  --dist=worksteal \
+  -m "mg" \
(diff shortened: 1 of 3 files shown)


ci/build_wheel.sh
```diff
@@ -22,12 +22,26 @@ sccache --stop-server 2>/dev/null || true
 
 rapids-logger "Building '${package_name}' wheel"
 
+RAPIDS_PIP_WHEEL_ARGS=(
+  -w dist
+  -v
+  --no-deps
+  --disable-pip-version-check
```
jameslamb, line 29: Looks like an `--extra-index-url https://pypi.nvidia.com` was lost when these arguments were moved up.

At this point let's just see from CI if that ends up mattering, and if not let's leave it out. One less thing to update in https://github.com/rapidsai/build-planning/issues/241 😁

jameslamb (changes requested):
We want to also update comments like these, which are no longer accurate:

https://github.com/rapidsai/cugraph/blob/52c02ab049fa65c17df3b09c7cd111fbcc395c41/ci/build_wheel_pylibcugraph.sh#L16-L17

Similar to what you did in the RMM PR: https://github.com/rapidsai/rmm/pull/2212/files#diff-616eb8e64286a9a59a2cd9c77d817fba66ce78ecfdde635db9f4028b74349585R25

ci/build_wheel.sh
jameslamb: Ok I see it's been put back. That's fine, removing it wasn't a part of this anyway.

ci/build_wheel_pylibcugraph.sh
```diff
@@ -13,8 +13,8 @@ RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen "${RAPIDS_CUDA_VERSION}")"
 # Download the libcugraph wheel built in the previous step and make it
 # available for pip to find.
 #
-# Using env variable PIP_CONSTRAINT (initialized by 'rapids-init-pip') is necessary to ensure the constraints
-# are used when creating the isolated build environment.
+# env variable 'PIP_CONSTRAINT' is set up by rapids-init-pip. It constrains all subsequent
+# 'pip install', 'pip download', etc. calls (except those used in 'pip wheel', handled separately in build scripts)
```
jameslamb, line 17: There's one of these comments in `ci/build_wheel_cugraph.sh` as well:

https://github.com/rapidsai/cugraph/blob/52c02ab049fa65c17df3b09c7cd111fbcc395c41/ci/build_wheel_cugraph.sh#L16-L17

That also needs to be updated.

mmccarty: Man, these are everywhere. I'm just going to start grepping

jameslamb: Yeah, got to.

PR #5414, opened 2026-01-29
C++ Disjoint sampling implementation

ChuckHastings:
This PR adds the disjoint sampling feature to sampling in C++.  C++ tests exist for homogeneous uniform and biased sampling, both for SG and MG.

This should get us started on the disjoint feature, we can add tests for heterogeneous and temporal variations as well, I plan to do that in a follow on activity.  We also need to test that the C API level, which I will add to a later pull request.


cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
                          std::move(result_labels));
 }
 
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::vector<arithmetic_device_uvector_t>,
+           std::optional<rmm::device_uvector<int32_t>>>
+gather_one_hop_edgelist_with_visited(
```
seunghwak: I am not sure about this name, this function finds all one hop edges and filters out all the edges with visited destinations. This function name some what implies that it keeps the edges with visited destinations. Can we rename this function to easily find what this function is doing?

What about something like `gather_edgelist_to_unvisited_neighbors` or `gather_one_hop_edgelist_to_unvisited_neighbors`?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
+  std::optional<edge_property_view_t<edge_t, int32_t const*>> edge_type_view,
+  raft::device_span<vertex_t const> active_majors,
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  std::optional<raft::device_span<uint8_t const>> gather_flags,
+  std::optional<rmm::device_uvector<vertex_t>>& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>& visited_vertex_labels,
```
seunghwak: Better take `visited_vertices` and `visited_vertex_lables` as R-value references and return the new ones to be more functional.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+  bool do_expensive_check)
+{
+  CUGRAPH_EXPECTS(visited_vertices, "Visited vertices must be provided");
```
seunghwak: In this case, why are we taking std::optional? Better to detect this in compile time than run-time.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+           std::vector<arithmetic_device_uvector_t>,
+           std::optional<rmm::device_uvector<int32_t>>>
+gather_one_hop_edgelist_with_visited(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
+  std::optional<edge_property_view_t<edge_t, int32_t const*>> edge_type_view,
+  raft::device_span<vertex_t const> active_majors,
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  std::optional<raft::device_span<uint8_t const>> gather_flags,
```
seunghwak: I know we are using uint8_t in `gather_one_hop_edgelist` as well, but should we better use bool instead of uint8_t here?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+  {
+    auto [keep_count, marked_entries] =
+      visited_vertex_labels
+        ? detail::mark_entries(
+            handle,
+            result_dsts.size(),
+            [dsts                  = result_dsts.data(),
+             labels                = result_labels->data(),
+             visited_vertices      = visited_vertices->data(),
+             visited_labels        = visited_vertex_labels->data(),
```
seunghwak: Any assumptions about how `visited_vertices` and `visited_vertex_labels` will be partitioned in multi-GPU? Will this code work in multi-GPU? (especially in extreme-scale?)

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+      handle, remove_dsts.size(), detail::is_first_in_run_t<decltype(tuple_first)>{tuple_first});
+
+    positions = detail::keep_marked_entries(
+      handle,
+      std::move(positions),
+      raft::device_span<uint32_t const>{keep_flags.data(), keep_flags.size()},
+      keep_count);
+  } else {
+    rmm::device_uvector<vertex_t> remove_srcs(result_srcs.size(), handle.get_stream());
+    rmm::device_uvector<vertex_t> remove_dsts(result_dsts.size(), handle.get_stream());
```
seunghwak: Why do we need `srcs` here? Aren't we just removing destination vertices that appear more than once?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+    auto tuple_first =
+      thrust::make_zip_iterator(remove_labels.begin(), remove_dsts.begin(), remove_srcs.begin());
+    auto [keep_count, keep_flags] = detail::mark_entries(
+      handle, remove_dsts.size(), detail::is_first_in_run_t<decltype(tuple_first)>{tuple_first});
+
+    positions = detail::keep_marked_entries(
+      handle,
+      std::move(positions),
+      raft::device_span<uint32_t const>{keep_flags.data(), keep_flags.size()},
+      keep_count);
```
seunghwak: I assume this code is same for both the if and else cases? Should we replicate the code?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: And this code won't work across multi-GPUs, won't this?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +333,233 @@ gather_one_hop_edgelist(
...
+                       positions.end(),
+                       property.begin(),
+                       keep_property.begin());
+        property = std::move(keep_property);
+      });
+    });
+
+  // Finally, I update the visited vertices and vertex labels to include the new result set
+  rmm::device_uvector<vertex_t> new_visited_vertices(visited_vertices->size() + result_dsts.size(),
+                                                     handle.get_stream());
```
seunghwak: Better sort `result_dsts` first and call `thrust::merge`. Sorting the entire array again and again is expensive.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -83,12 +85,53 @@ struct sample_edge_biases_op_t {
   }
 };
 
+template <typename vertex_t, typename bias_t>
+struct sample_unvisited_edge_biases_op_t {
+  raft::device_span<vertex_t const> visited_vertices{};
+  cuda::std::optional<raft::device_span<int32_t const>> visited_vertex_labels{};
```
seunghwak: Any assumption about partitioning here? will this follow edge partitioning or vertex partitioning? If this follows edge partitioning and if this is for edge destinations, we may better use minors instead of vertices in the naming.

seunghwak (commented):
Review part 1.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -230,255 +510,480 @@ sample_with_one_property(
...
+                               *edge_type_view))
+                         : std::nullopt,
+          rng_state,
+          Ks,
+          active_major_labels,
+          with_replacement);
+    }
+
+    // Check for duplicates in the sampled minor vertices
+    rmm::device_uvector<vertex_t> local_majors(sampled_majors.size(), handle.get_stream());
```
seunghwak: I assume this code to remove duplicates in the sampled minor vertices is same for both this function and the gather-one-hop-edgelist function. Can't we merge the two to a single utility function?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -1020,16 +1321,234 @@ temporal_sample_edges(raft::handle_t const& handle,
...
+  return std::make_tuple(
+    std::move(majors), std::move(minors), std::move(edge_properties), std::move(labels));
+}
+
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           std::vector<arithmetic_device_uvector_t>,
+           std::optional<rmm::device_uvector<int32_t>>>
+sample_edges_with_visited(
```
seunghwak: Similar to the gather one-hop edge list function, should we better rename this function?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -1020,16 +1321,234 @@ temporal_sample_edges(raft::handle_t const& handle,
...
+  raft::random::RngState& rng_state,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_type_view,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_bias_view,
+  raft::device_span<vertex_t const> active_majors,
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  raft::host_span<size_t const> Ks,
+  std::optional<rmm::device_uvector<vertex_t>>& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>& visited_vertex_labels,
```
seunghwak: Similar here, better take R-value references.

cpp/src/sampling/detail/update_visited_utils.cu
```diff
@@ -0,0 +1,175 @@
...
+
+namespace cugraph {
+namespace detail {
+
+template <typename vertex_t, typename edge_t, bool multi_gpu>
+std::tuple<std::optional<rmm::device_uvector<vertex_t>>,
+           std::optional<rmm::device_uvector<int32_t>>>
+update_dst_visited_vertices_and_labels(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
```
seunghwak, line 30: Any assumptions about MG partitioning here?

cpp/src/sampling/detail/update_visited_utils.cu
```diff
@@ -0,0 +1,175 @@
...
+  std::optional<rmm::device_uvector<vertex_t>>&& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>&& visited_vertex_labels,
+  raft::device_span<vertex_t const> sampled_vertices,
+  std::optional<raft::device_span<int32_t const>> sampled_vertex_labels)
+{
+  CUGRAPH_EXPECTS(visited_vertices.has_value(), "Invalid input: visited_vertices must be provided");
```
seunghwak: Then, why are we taking std::optional here?

cpp/src/sampling/detail/update_visited_utils.cu
```diff
@@ -0,0 +1,175 @@
...
+    thrust::copy(handle.get_thrust_policy(),
+                 sampled_vertex_labels->begin(),
+                 sampled_vertex_labels->end(),
+                 new_sample_labels->begin());
+    props.push_back(std::move(*new_sample_labels));
+  }
+
+  if constexpr (multi_gpu) {
+    std::tie(new_samples, props) = cugraph::shuffle_int_vertices(
+      handle, std::move(new_samples), std::move(props), graph_view.vertex_partition_range_lasts());
```
seunghwak, line 61: What about labels here? No need to shuffle labels as well?

cpp/src/sampling/detail/update_visited_utils.cu
seunghwak: If sampled_vertices follow the vertex partitioning and visited_vertices/visited_vertex follows edge partitioning (for minors), we should better rename accordingly.

cpp/tests/sampling/detail/nbr_sampling_validate.cu
```diff
@@ -426,5 +428,73 @@ template bool validate_temporal_integrity(
   raft::device_span<int64_t const> source_vertices,
   cugraph::temporal_sampling_comparison_t temporal_sampling_comparison);
 
+template <typename vertex_t>
+bool validate_disjoint_sampling(
```
seunghwak, line 431: Is this an internal test utility function or something general users will call?

seunghwak (commented):
Review part 2

cpp/tests/sampling/detail/nbr_sampling_validate.cu
ChuckHastings: Internal test utility function.

seunghwak: In that case, should we include this in libcugraph.so (considering the pressure on binary size)? Better move to the test directory?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Updated to `gather_one_hop_edgelist_to_unvisited_neighbors` in next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Yeah, I started with this as a change to gather_one_hop_vertices, and it would be optional... if specified it added the new logic.  Then I realized I needed to restructure the main flow, seemed too complex, so I made it a separate function.

I'll change it in the next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Will change in the next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: This is managed by `update_dst_visited_vertices_and_labels`.  They are replicated (allgatherv) across the minor communicator.  This is necessary for the logic to work.

Size should be reasonably manageable.  `Number of hops * fanout / p_row` (or something like that) would be the expected number of entries per GPU.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I could pull it out, but right now the scope of keep_flags and keep_count is inside the code block and they are automatically freed when we exit.

If I pull it out I'll need to define them explicitly and then resize and shrink to fit.  Not sure which is better.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I use srcs in the sort so that I select the edge with the lowest source as the one that is selected.  Perhaps not important in the gather_one_hop path, but the other path needs to break the ties in a way that guarantees that at least one source is fully sampled in each iteration of the loop.  I kept that in place here more for consistency, but I can drop it if you think we shouldn't worry about that consistency.

The multi-GPU issue is a defect, I'll fix that in the next push.  I have an extra shuffle and check in the sample edges path that I should replicate after this to check for duplicates across GPUs.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: This should be calling the update_dst_visited_vertices_and_labels function which handles the MG support as well.

Now your thrust::merge comment might be relevant there.

ChuckHastings: Next push will call the function instead.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: visited_vertices/visited_vertex_labels is an allgatherv across the minor communicator, so all elements are replicated across the minor communicator.  This allows any GPU that might include a vertex as a destination to have that information.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Will look at this as I fix the MG portion of gather-one-hop-edgelist

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Renamed to `sample_edges_to_unvisited_neighbors` in the latest push.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/update_visited_utils.cu
ChuckHastings: Fixed in next push.

cpp/src/sampling/detail/update_visited_utils.cu
ChuckHastings: Suggestions on names?

Yes, sampled_vertices are partitioned by vertex partitioning,
visited_vertices/visited_vertex_labels are replicated (allgatherv) across the minor communicator.

cpp/src/sampling/detail/update_visited_utils.cu
ChuckHastings: Labels are in props.  I reorganized in the next push, the props.push_back above and the std::move below are all in the multi_gpu block which makes it clear what's happening.

cpp/tests/sampling/detail/nbr_sampling_validate.cu
ChuckHastings: This file is in the `cpp/tests` directory and is compiled into `libcugraphtest_util.a`.

ChuckHastings (commented):
Will address many of these in my next push.  A few comments/questions that aren't corrected yet.  A few things not commented on I'll follow up with later.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +334,212 @@ gather_one_hop_edgelist(
...
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  std::optional<raft::device_span<uint8_t const>> gather_flags,
+  rmm::device_uvector<vertex_t>&& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>&& visited_vertex_labels,
+  bool do_expensive_check)
+{
+  // Implement this, should be a little easier than sample_edges_to_unvisited_neighbors, since we
+  // don't need to compute the probability of sampling for an edge based on the label/tag.  We can
+  // just extract everything and then filter the results based on the visited vertices and vertex
+  // labels.
```
seunghwak: "Implement this,"=>Isn't this now an outdated comment? You already implemented this.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +334,212 @@ gather_one_hop_edgelist(
...
+           std::optional<rmm::device_uvector<int32_t>>>
+gather_one_hop_edgelist_to_unvisited_neighbors(
+  raft::handle_t const& handle,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
+  std::optional<edge_property_view_t<edge_t, int32_t const*>> edge_type_view,
+  raft::device_span<vertex_t const> active_majors,
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  std::optional<raft::device_span<uint8_t const>> gather_flags,
+  rmm::device_uvector<vertex_t>&& visited_vertices,
```
seunghwak: In multi-GPU, is this visited_minors? If this just stores the "visited_(local_)vertices, this code won't work.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +334,212 @@ gather_one_hop_edgelist(
...
+  std::optional<rmm::device_uvector<int32_t>>&& visited_vertex_labels,
+  bool do_expensive_check)
+{
+  // Implement this, should be a little easier than sample_edges_to_unvisited_neighbors, since we
+  // don't need to compute the probability of sampling for an edge based on the label/tag.  We can
+  // just extract everything and then filter the results based on the visited vertices and vertex
+  // labels.
+  auto [result_srcs, result_dsts, result_properties, result_labels] =
```
seunghwak: And we may consistently use majors & minors in the detail namespace (even though here, store_transposed == false, so minors are always destinations).

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: Oh, I see, yeah... those are comparable... so not worth the additional work.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: Can't we break ties with positions?

> but the other path needs to break the ties in a way that guarantees that at least one source is fully sampled in each iteration of the loop.

I am having hard time interpreting this. You mean in the sampling path, you want each active major to have at least one sampled edges (so, if there are multiple (different active_major, same minor) pairs, you prefer to select the one with no currently sampled edges?).

cpp/src/sampling/detail/sample_edges.cuh
seunghwak: Then, should we better name this as `visited_minors` & `visited_minor_labels`?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -230,255 +510,482 @@ sample_with_one_property(
...
+sample_unvisited_with_one_property(
+  raft::handle_t const& handle,
+  raft::random::RngState& rng_state,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  property_view_t edge_property_view,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_type_view,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_bias_view,
+  cugraph::vertex_frontier_t<vertex_t, tag_t, multi_gpu, false>& vertex_frontier,
+  rmm::device_uvector<vertex_t>&& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>&& visited_vertex_labels,
```
seunghwak: Should better be visited_minors & visited_minor_labels (or visited_dsts and visited_dst_labels if this function is in the public namespace)?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -1020,20 +1323,245 @@ temporal_sample_edges(raft::handle_t const& handle,
...
+  raft::random::RngState& rng_state,
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_type_view,
+  std::optional<edge_arithmetic_property_view_t<edge_t>> edge_bias_view,
+  raft::device_span<vertex_t const> active_majors,
+  std::optional<raft::device_span<int32_t const>> active_major_labels,
+  raft::host_span<size_t const> Ks,
+  rmm::device_uvector<vertex_t>&& visited_vertices,
+  std::optional<rmm::device_uvector<int32_t>>&& visited_vertex_labels,
```
seunghwak: visited_minors & visited_minor_labels (or visited_dsts & visisted_dst_labels if in the public namespace)?

cpp/src/sampling/detail/update_visited_utils.cu
seunghwak: visited_minors & visited_minor_labels (if in the detail namespace) or visited_dsts and visited_dst_lables (if in the public namespace)?

cpp/tests/sampling/detail/nbr_sampling_validate.cu
seunghwak: Oh, I see... my bad :-(

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: Yeah... I think bool will better convey the meaning of this parameter.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: Yes, in this context, majors & minors will make more sense. Mixing majors/minors and srcs/dsts can be confusing.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -333,6 +336,110 @@ gather_one_hop_edgelist(
...
+                                                         do_expensive_check);
+
+  // Remove any edges that lead to a visited vertex
+  {
+    auto [keep_count, marked_entries] =
+      visited_minor_labels
+        ? detail::mark_entries(
+            handle,
+            result_dsts.size(),
+            [dsts                = result_dsts.data(),
```
seunghwak: dsts=>minors, better not mix src/dst and major/minor. We use src/dst in cugraph, major/minor in cugraph::detail

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
```diff
@@ -0,0 +1,496 @@
...
+remove_duplicate_edges(raft::handle_t const& handle,
+                       graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+                       rmm::device_uvector<vertex_t>&& result_majors,
+                       rmm::device_uvector<vertex_t>&& result_minors,
+                       std::vector<arithmetic_device_uvector_t>&& result_properties,
+                       std::optional<rmm::device_uvector<int32_t>>&& result_labels,
+                       rmm::device_uvector<vertex_t>&& visited_minors,
+                       std::optional<rmm::device_uvector<int32_t>>&& visited_minor_labels,
+                       bool call_from_sampling)
+{
```
seunghwak, line 44: In case we don't have labels, isn't this remove_multi_edges?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -216,11 +258,248 @@ struct segmented_fill_t {
   }
 };
 
+template <typename vertex_t,
```
seunghwak, line 280: Adding a comment saying this function is to handle both homogeneous/heterogeneous sampling might be helpful in understanding why this wrapper is here.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -216,11 +258,248 @@ struct segmented_fill_t {
...
+      segmented_fill_t{*active_major_labels,
+                       raft::device_span<size_t const>{offsets->data(), offsets->size()},
+                       raft::device_span<int32_t>{labels->data(), labels->size()}});
+  }
+
+  return std::make_tuple(
+    std::move(labels), std::move(majors), std::move(minors), std::move(sampled_properties));
+}
+
+template <typename vertex_t,
```
seunghwak, line 413: Adding a comment saying this function is to handle both homogeneous/heterogeneous sampling might be helpful in understanding why this wrapper is here.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -216,11 +258,248 @@ struct segmented_fill_t {
...
+                           edge_dst_dummy_property_t{}.view(),
+                           edge_property_view,
+                           edges_op,
+                           *edge_type_view,
+                           rng_state,
+                           Ks,
+                           with_replacement,
+                           std::optional<return_type>{std::nullopt},
+                           false);
+  }
```
seunghwak, line 390: Aren't we separating sampling from edge property gathering?

We added edge_multi_index_property_view_t (https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/edge_property.hpp#L57) to support this.

seunghwak: Once we have edge list, we can use a single edge property gathering function for all types of sampling AND gathering functions.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -216,11 +258,248 @@ struct segmented_fill_t {
...
+      rmm::device_uvector<int32_t>(offsets->back_element(handle.get_stream()), handle.get_stream());
+    auto num_segments = offsets->size() - size_t{1};
+    thrust::for_each(
+      handle.get_thrust_policy(),
+      thrust::make_counting_iterator(size_t{0}),
+      thrust::make_counting_iterator(num_segments),
+      segmented_fill_t{*active_major_labels,
+                       raft::device_span<size_t const>{offsets->data(), offsets->size()},
+                       raft::device_span<int32_t>{labels->data(), labels->size()}});
+  }
```
seunghwak, line 515: Once we separate edge sampling from edge property gathering, do we still need to replicate this segmented fill logic?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -216,11 +258,248 @@ struct segmented_fill_t {
...
+    std::move(labels), std::move(majors), std::move(minors), std::move(sampled_properties));
+}
+
 template <typename vertex_t, typename edge_t, typename property_view_t, bool multi_gpu>
 std::tuple<rmm::device_uvector<vertex_t>,
            rmm::device_uvector<vertex_t>,
            arithmetic_device_uvector_t,
-           std::optional<rmm::device_uvector<size_t>>>
+           std::optional<rmm::device_uvector<int32_t>>>
 sample_with_one_property(
```
seunghwak, line 544: If we separate edge sampling from property gathering, do we still need this function?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -230,255 +509,298 @@ sample_with_one_property(
...
+          typename tag_t,
+          typename property_view_t,
+          bool multi_gpu>
+std::tuple<rmm::device_uvector<vertex_t>,
+           rmm::device_uvector<vertex_t>,
+           arithmetic_device_uvector_t,
+           std::optional<rmm::device_uvector<int32_t>>,
+           rmm::device_uvector<vertex_t>,
+           std::optional<rmm::device_uvector<int32_t>>>
+sample_unvisited_with_one_property(
```
seunghwak, line 638: Similar here, if we separate edge sampling from edge property gathering, we just need different sampling functions for heterogenous/homogeneous temporal/non-temporal disjoint/non-disjoint cases. All the remaining steps can be a single function I think.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Switched to a bool

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Fixed all src and dst references throughout this file to be majors/minors

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Changed

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
ChuckHastings: As discussed, this name is not clear, I have renamed the function to be `deduplicate_edges_by_minor`

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Added some comments.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: I have moved this out in to the calling functions.

We can cleanly separate this for non-temporal implementations and I have done that with the latest push.  Temporal sampling needs some of the properties in order to function properly, I could not find a good way to delay for that code path.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Added a comment block here as well.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: I think we do.  I could move it to the calling location, but this helper function is called in many places, so it would be called in more places.

I could create a function to do the computation so that we don't recompile it multiple times.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: This function is currently designed to be called with a dummy edge property, a single edge property if there is only one edge property for the graph, or a multi-edge property if we have a multi-graph.

I created it (and now the call_{biased, unbiased}_per_v_random_select_transform_outgoing_e functions) primarily to avoid replicating the calls to per_v_random_select_transform_outgoing_e that had complex nesting inside if statements for all of the combinations that we needed to support.

It might be cleaner to refactor this function to take an optional multi-edge property view and either use it or a dummy edge property.  I'd be inclined to leave that for a future improvement.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Same discussion... I was trying to avoid replicating all of the different calls so many places.

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
```diff
@@ -0,0 +1,501 @@
...
+                           std::move(resample_majors),
+                           std::move(resample_major_labels));
+  }
+
+  size_t total_edges = result_majors.size();
+
+  if constexpr (multi_gpu) {
+    total_edges = host_scalar_allreduce(
+      handle.get_comms(), total_edges, raft::comms::op_t::SUM, handle.get_stream());
+  }
```
seunghwak: result_majors.size() != result_minors.size()? (or result_size != total edges?)

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
```diff
@@ -0,0 +1,501 @@
...
+    shuffle_properties.push_back(std::move(keep_majors));
+    shuffle_properties.push_back(std::move(*keep_ranks));
+    shuffle_properties.push_back(std::move(keep_positions));
+    if (result_labels) { shuffle_properties.push_back(std::move(*keep_labels)); }
+
+    std::tie(keep_minors, shuffle_properties) =
+      shuffle_int_vertices(handle,
+                           std::move(keep_minors),
+                           std::move(shuffle_properties),
+                           raft::host_span<vertex_t const>{},
```
seunghwak: Why are we passing an empty span here?

seunghwak: Shouldn't this be `graph_view.vertex_partition_range_lasts()`?

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
```diff
@@ -0,0 +1,501 @@
...
+  size_t global_remove_count{keep_minors.size() - keep_count};
+  if constexpr (multi_gpu) {
+    global_remove_count = host_scalar_allreduce(handle.get_comms(),
+                                                (keep_minors.size() - keep_count),
+                                                raft::comms::op_t::SUM,
+                                                handle.get_stream());
+  }
+
+  // TODO: think about this.  If global_remove_count is 0, we can just return the current
+  // result_majors and result_minors?
```
seunghwak: Sounds right to me.

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
```diff
@@ -0,0 +1,501 @@
...
+
+  // 4. If we have any duplicates on any GPU we need to remove them
+  if (global_remove_count > 0) {
+    if (call_from_sampling) {
+      // FIXME: After we refactor sample_edges_to_unvisited_neighbors to use the partial results we
+      // can eliminate this if
```
seunghwak: Better move this FIXME to the function definition? First seeing `call_from_sampling`, I was puzzled why this parameter is necessary.

cpp/src/utilities/swap_marked_and_unmarked_entries.cu
```diff
@@ -0,0 +1,39 @@
...
+namespace detail {
+
+void swap_marked_and_unmarked_entries(raft::handle_t const& handle,
+                                      raft::device_span<uint32_t> flags,
+                                      size_t num_entries)
+{
+  uint32_t full_mask    = packed_bool_full_mask();
+  uint32_t partial_mask = packed_bool_partial_mask(num_entries % packed_bools_per_word());
+
+  thrust::for_each(handle.get_thrust_policy(),
```
seunghwak: thrust::tabulate should work better.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -256,6 +256,9 @@ rmm::device_uvector<T> keep_marked_entries(
   return result;
 }
 
-}  // namespace detail
+void swap_marked_and_unmarked_entries(raft::handle_t const& handle,
+                                      raft::device_span<uint32_t> flags,
+                                      size_t num_entries);
```
seunghwak: flip(invert)_keep_flags? swap somewhat implies that we are changing values of two variables (e.g. std::swap).

seunghwak (approved):
I have few questions/recommendations, but otherwise, LGTM.

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
ChuckHastings: Fixed in next push.

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
ChuckHastings: Yes

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
ChuckHastings: I refactored this slightly.  I had combined the update of visited vertices into this function, which means there's still work in this function.  I think it makes sense for them to be separate so I removed the call to update visited vertices and call it separately at each call site.

Removed the TODO.

cpp/src/sampling/detail/deduplicate_edges_by_minor.cu
ChuckHastings: Moved the FIXME and reworded a bit.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Changed it to invert_flags.

cpp/src/utilities/swap_marked_and_unmarked_entries.cu
ChuckHastings: Updated to thrust::tabulate

PR #5482, opened 2026-04-02
Remove a couple of variant branches to shrink libcugraph_mg about 6%

ChuckHastings:
Part of the effort for #5473 

A couple of the variant paths add a significant amount to our binary size, don't expect there to be a big performance difference.

### DIFF
--- a/cpp/src/detail/groupby_and_count.cuh
+++ b/cpp/src/detail/groupby_and_count.cuh
@@ -109,38 +109,6 @@ rmm::device_uvector<size_t> groupby_and_count_edgelist_by_local_partition_id(
                                           handle.get_stream(),
                                           large_buffer_type);
     }
-  } else if (edgelist_properties.size() == 1) {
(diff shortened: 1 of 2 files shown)


seunghwak:
Have you checked the performance difference when the property type is float?

Instead of grouping with 4B float, we perform grouping with 8B size_t, then gather as well. I expect to see some performance differences. Not sure it is rather moderate (e.g. 10-20% difference) or pretty prominent (100%+).

LGTM otherwise.

jnke2016 (approved):

ChuckHastings:
> Have you checked the performance difference when the property type is float?
> 
> Instead of grouping with 4B float, we perform grouping with 8B size_t, then gather as well. I expect to see some performance differences. Not sure it is rather moderate (e.g. 10-20% difference) or pretty prominent (100%+).
> 
> LGTM otherwise.

I did not measure it.  I can try that.

seunghwak (approved):
LGTM

PR #5483, opened 2026-04-03
Migrate RMM usage to CCCL MR design

bdice:
## Summary

- Replace removed `rmm::mr::device_memory_resource` base class, `owning_wrapper`, `shared_ptr`-based resource management, and deprecated per-device resource APIs with CCCL-native memory resource types
- Use `cuda::mr::any_resource<cuda::mr::device_accessible>` for owning type-erased storage, `rmm::device_async_resource_ref` for non-owning references, and value-typed resources (`cuda_memory_resource`, `pinned_host_memory_resource`)
- Pass the memory resource to `raft::handle_t` as the `workspace_resource` (3rd) constructor argument, matching the new raft API (`stream_view`, `stream_pool`, `std::optional<raft::mr::device_resource>`)

Depends on https://github.com/rapidsai/rmm/pull/2361.
Depends on https://github.com/rapidsai/ucxx/pull/636.
Depends on https://github.com/rapidsai/raft/pull/2996.
Depends on https://github.com/rapidsai/cuvs/pull/1990.

## Files changed

**Headers:**
- `algorithms.hpp`, `dendrogram.hpp`, `legacy/graph.hpp`, `legacy/functions.hpp`: `get_current_device_resource()` → `get_current_device_resource_ref()` in default argument expressions
- `host_staging_buffer_manager.hpp`: Remove `owning_wrapper`, store `pool_memory_resource` by value in a `std::optional`, accept `pinned_host_memory_resource` by value in `init()`
- `large_buffer_manager.hpp`: Store `pinned_host_memory_resource` by value (not `shared_ptr`), return `device_async_resource_ref` from `get()`, `std::move` the resource into storage
- `mtmg/resource_manager.hpp`: Use `cuda::mr::any_resource<device_accessible>` instead of `shared_ptr<device_memory_resource>` for `per_device_rmm_resources_`, use non-deprecated `set_per_device_resource`, pass resource as `workspace_resource` to `raft::handle_t`

**Tests:**
- `base_fixture.hpp`: Return `any_resource<device_accessible>` from `create_memory_resource()`, use value-typed MR factory helpers (`make_cuda`, `make_managed`, `make_pool`, `make_binning`), switch to non-deprecated `set_current_device_resource` / `get_current_device_resource_ref`
- `multi_node_threaded_test.cpp`: Switch to non-deprecated `set_current_device_resource(resource)`
[...]

PR #5486, opened 2026-04-08
Reduce binary size part2

ChuckHastings:
Add some constexpr keywords to some multi_gpu if blocks to reduce library size of libcugraph_mg.  Restructured Leiden refinement code to mirror other algorithms template parameters, then a few changes to reduce library size in Leiden.


cpp/src/community/detail/decision_graph_mis.cuh
```diff
@@ -0,0 +1,96 @@
...
+
+template <typename vertex_t, bool multi_gpu>
+rmm::device_uvector<vertex_t> vertices_in_mis_from_decision_edgelist(
+  raft::handle_t const& handle,
+  raft::random::RngState& rng_state,
+  raft::host_span<vertex_t const> vertex_partition_range_lasts,
+  rmm::device_uvector<vertex_t>&& d_srcs,
+  rmm::device_uvector<vertex_t>&& d_dsts)
+{
+  using edge_t = vertex_t;
```
seunghwak, line 33: Should we better avoid this?

We currently don't build for edge_t != vertex_t cases in our regular release but we still allow users to manually build for these cases.

seunghwak: Or is there a guarantee that # edges <= # vertices in this use case? In that case, we should add a comment here to avoid future confusion.

cpp/src/community/detail/decision_graph_mis.cuh
```diff
@@ -0,0 +1,96 @@
...
+
+  constexpr bool decision_store_transposed = false;
+
+  cugraph::graph_t<vertex_t, edge_t, decision_store_transposed, multi_gpu> decision_graph(handle);
+
+  std::optional<rmm::device_uvector<vertex_t>> renumber_map{std::nullopt};
```
seunghwak: Something very tedious, but in general, it is better to minimize the scope of a variable. We can define this after the shuffle operation below.

seunghwak (approved):
LGTM except for minor cosmetic issues.

cpp/src/community/detail/decision_graph_mis.cuh
ChuckHastings: Yes, I've added a comment.  The number of entries in d_srcs/d_dsts starts as the number of vertices in the graph and then gets filtered down, so it's guaranteed to fit in a vertex_t.

cpp/src/community/detail/decision_graph_mis.cuh
ChuckHastings: I moved it down.

ChuckHastings:
Fix is coming in #5483

PR #5490, opened 2026-04-13
Addressing Issue 5397 5384

### LINKED ISSUE #5384: Drop dependence on python-louvain in testing

### LINKED ISSUE #5397: [BUG]: `batched_ego_graphs` (deprecated and removed) still used in an MG test
### Version

dev (26.02)

Source

### Describe the bug.

Tests in `python/cugraph/cugraph/tests/sampling/test_egonet_mg.py` use `cugraph.batched_ego_graphs`, which was removed in #5133. See also #4708.

I encountered this when running _all_ cugraph tests locally on a multi-gpu system.

The easiest solution would be to remove the test file, but a better solution may be to update it to not use `batched_ego_graphs`. Also, I'm surprised I'm the first to catch this since #5133 was merged last July!

```shell
Run `test_egonet_mg.py` tests on a multi-gpu system.
```

### Relevant log output

```shell
_ ERROR at setup of test_dask_mg_ego_graphs[graph_file:/home/nfs/erwelch/.cugraph/datasets/karate.csv-directed:True-seeds:0-radius:1] _
                                                                                                                                        input_combo = {'directed': True, 'graph_file': PosixPath('/home/nfs/erwelch/.cugraph/datasets/karate.csv'), 'radius': 1, 'seeds': 0}
    def input_expected_output(input_combo):                                                                                                     """
        This fixture returns the inputs and expected results from the egonet algo.                                                              (based on cuGraph batched_ego_graphs) which can be used for validation.
        """                                                                                                                             ····
        input_data_path = input_combo["graph_file"]                                                                                             directed = input_combo["directed"]
        seeds = input_combo["seeds"]                                                                                                            radius = input_combo["radius"]
        G = utils.generate_cugraph_graph_from_file(                                                                                                 input_data_path, directed=directed, edgevals=True
        )                                                                                                                               ····
>       sg_cugraph_ego_graphs = cugraph.batched_ego_graphs(G, seeds=seeds, radius=radius)                                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: module 'cugraph' has no attribute 'batched_ego_graphs'                                                          
tests/sampling/test_egonet_mg.py:73: AttributeError
```

### Environment details

```shell

```

### Other/Misc.

_No response_

### Code of Conduct

- [x] I have searched the [open bugs](https://github.com/rapidsai/cugraph/issues?q=is%3Aopen+is%3Aissue+label%3Abug) and have found no duplicates for this bug report

BradReesWork:
1) Addressed issues 5397 that batch ego graph was still being called after being drooped

2) Dropped batch betweenness centrality 

3) Dropped use of python-louvain 

4) added some benchamrk tweaks to help ensure memory is collected

Close #5397
Close #5384


acostadon (approved):
Looks good

benchmarks/cugraph/pytest-based/bench_algos.py
```diff
@@ -197,31 +199,41 @@ def dataset(request, rmm_config):
 @pytest.fixture(scope="module")
 def edgelist(request, dataset):
     df = dataset.get_edgelist()
-    return df
+    yield df
+    del df
+    gc.collect()
```
rlratzel, line 204: I'm curious if you observed this helping or if it was added just in case.  I'm thinking we shouldn't have to do this, and if we do then we either have a bug or a usability problem that should be addressed.  Or, if there's a technical limitation that requires us to do this, then it should be documented so users are aware.

I'm also curious if these changes are actually doing anything, given that these fixtures are scoped to "module".  The post-yield teardown is still only going to run *once*, and only *after* every benchmark that uses this fixture in the module is run.  The point of making it "module" was to share a single instance of the common fixture values to prevent creating and destroying the same objects over and over.  If this is actually causing a problem (possibly by having >1 dataset in memory at a time while pytest is running several combinations), then we should change the scope.  Better yet, we can educate users on this so they can take advantage of module-level efficiency by running benchmarks on fewer datasets at a time.  Either approach would be preferred IMO.

Possibly related note: I remember hearing that running these benchmarks resulted in OOM errors for some, but then I later learned that some users were trying to run benchmarks using *every* available dataset. There will likely be some datasets that are too large for some algos on some machines, in which case those OOMs wouldn't be prevented by manually deleting objs and forcing the GC like this.  It would be nice to know more about the circumstances around those OOMs so we can properly root cause them.

python/cugraph/cugraph/tests/sampling/test_egonet_mg.py
```diff
@@ -70,13 +72,46 @@ def input_expected_output(input_combo):
...
+    n = seeds.astype(n_type)
+    do_expensive_check = False
+
+    source, destination, weight, offset = pylibcugraph_ego_graph(
+        resource_handle=ResourceHandle(),
+        graph=G._plc_graph,
+        source_vertices=n,
+        radius=radius,
+        do_expensive_check=do_expensive_check,
+    )
```
rlratzel, line 95: I'd rather not have this test here at all if it's actually a PLC test.  Can you move this to PLC, or just remove it if we have the proper coverage elsewhere?

python/cugraph/cugraph/tests/community/test_louvain.py
```diff
@@ -1,31 +1,17 @@
...
 import cugraph
 import cupyx
 import cudf
-from cugraph.testing import utils, UNDIRECTED_DATASETS
+from cugraph.testing import UNDIRECTED_DATASETS
 from cugraph.datasets import karate_asymmetric
 
 
-try:
-    import community
```
rlratzel, line 17: Please also remove the `python-louvain` dependency [here](https://github.com/rapidsai/cugraph/blob/e45b2a36d733195ab7a2c3603c72a4035f6b64be/dependencies.yaml#L497).

rlratzel (changes requested):

rlratzel:
Sorry, I submitted my PR review without a summary:

Thanks, @BradReesWork for taking care of these items.  I have a few requests which I already posted, LMK if I can help.

python/cugraph/cugraph/tests/sampling/test_egonet_mg.py
ChuckHastings: Yes, let's move this test to PLC.  That makes more sense.

benchmarks/cugraph/pytest-based/bench_algos.py
BradReesWork: I did see this keeping the memory use lower.  It forced garbage collection more often, but the end result was still the same OOM.

BradReesWork: You are correct that we shouldn't have to do this.  I was diving into the OOM error and wanted to force a clean up after each algorithm test.  The forced garbage collection had a very slight impact on memory, but was not enough to fix the issue

I was going off this suggestion: 
"Rule of thumb: Use return only for simple, stateless values (a config dict, a number). Use yield whenever the fixture allocates anything that needs cleanup."

So I will leave the "yield" call but removed the gc calls

python/cugraph/cugraph/tests/sampling/test_egonet_mg.py
BradReesWork: It is not a test (PLC or otherwise) - what the "input_expected_output" is make a call that produces the expected results.  The batched egonet API produced a different result tupal then what plan egonet does.  Since I did not want to modify the egonet API to produce something just for a test case, I called the PLC code to get the needed values.  

The "test" is actually "test_dask_mg_ego_graphs" - now that could be modified to not have the "offset" value, but that would have been a lot more work

rlratzel: Ah, thanks, my fault for not looking at the full context here.  I'll make this as resolved.

KyleFromNVIDIA (approved):

rlratzel (approved):

jameslamb:
Thanks so much for doing this @BradReesWork !

Issue #5494, opened 2026-04-16
[BUG]: Python k-truss test fails on Blackwell with CUDA 13.0 due to thrust::reduce_by_key


jnke2016:
### Version

26.06 (main branch)

### Which installation method(s) does this occur on?

Pip

### Describe the bug.

The test_ktruss_subgraph_Graph test fails on Blackwell (RTX PRO 6000, sm_120) when built with nvcc 13.0 + CCCL 3.4.0. The bug was caught on CI through the wheel test configuration which uses nvcc 13.0, but the issue is not wheel-specific and affects any build using that nvcc version on Blackwell. The failure manifests as CUGRAPH_UNKNOWN_ERROR: edge_list contains edges that do not exist in the input graph during the k-truss peeling loop. The root cause is thrust::reduce_by_key on zip_iterator<tuple<int32_t, int32_t>> silently corrupting its output.

The bug is in nvcc 13.0 and is fixed in nvcc 13.2. Both use the exact same CCCL 3.4.0 headers; the only delta is the nvcc binary.

A workaround has been applied in k_truss_impl.cuh [MENTION PR NUMBER] replacing thrust::reduce_by_key with thrust::unique_copy + thrust::lower_bound / thrust::upper_bound.

The original reduce_by_key call should be restored once nvcc 13.0 is no longer supported or the nvcc fix is confirmed backported.

### Minimum reproducible example

```shell
Run the python `test_ktruss_subgraph_Graph` with the above configurations.
```

### Relevant log output

```shell
RuntimeError: non-success value returned from cugraph_k_truss_subgraph: CUGRAPH_UNKNOWN_ERROR
cuGraph failure at file=cpp/include/cugraph/prims/transform_e.cuh line=589:
Invalid input arguments: edge_list contains edges that do not exist in the input graph.
```

### Environment details

```shell
GPU: NVIDIA RTX PRO 6000 Blackwell Server Edition (sm_120)
Driver: 580.95.05
CUDA Toolkit: 13.0.88 (fails) / 13.2.51 (passes)
CCCL: 3.4.0 (same for both)
OS: Ubuntu 24.04 (container)
CI config: wheel-tests-cugraph / 13.1.1, 3.14, amd64, ubuntu24.04, rtxpro6000
```

### Code of Conduct

- [x] I agree to follow cuGraph's Code of Conduct
[...]

PR #5495, opened 2026-04-16
Fix k-truss test failure on Blackwell

jnke2016:
The test_ktruss_subgraph_Graph test was failing on Blackwell (RTX PRO 6000) with nvcc 13.0 due to thrust::reduce_by_key silently corrupting its output when operating on zip_iterator<tuple<int32_t, int32_t>>. The bug is in nvcc 13.0 and is fixed in nvcc 13.2.

This PR:

- Applies a workaround in k_truss_impl.cuh replacing the buggy thrust::reduce_by_key call with thrust::unique_copy + thrust::lower_bound / thrust::upper_bound
- Re-enables the k-truss wheel tests that were previously skipped.

See #5494  for details. The original reduce_by_key call should be restored once nvcc 13.0 is no longer supported.


cpp/src/community/k_truss_impl.cuh
```diff
@@ -476,13 +478,51 @@ k_truss(raft::handle_t const& handle,
...
+                          d_upper_bounds.begin());
+
+      thrust::transform(
+        handle.get_thrust_policy(),
+        d_lower_bounds.begin(),
+        d_lower_bounds.end(),
+        d_upper_bounds.begin(),
+        decrease_count.begin(),
+        cuda::proclaim_return_type<edge_t>(
+          [] __device__(edge_t d_lower, edge_t d_upper) { return d_upper - d_lower; }));
```
seunghwak, line 525: I think using
```
#if 1
work_around;
#else
original;
#endif
```
is easier to identify the workaround code over the original code. In the current code, it takes little time to figure out the boundary.

seunghwak: And instead of calling 3 thrust calls (lower_bound, upper_bound, and transform) + creating two intermediate vectors (d_lower_bounds & d_upper_bounds),

```
thrust::transform(
  handle.get_thrust_policy(),
  get_dataframe_buffer_begin(vertex_pair_buffer_unique),
  get_dataframe_buffer_begin(vertex_pair_buffer_unique) + unique_pair_count,
  decrease_counts.begin(),
  cuda::proclaim_return_type<edge_t>([
    edgelist_first = get_dataframe_buffer_begin(edge_list_to_update_count),
    edgelist_last = get_dataframe_buffer_end(edge_list_to_update_count)
  ]__device__(auto pair) {
    return static_cast<edge_t>(cuda::std::distance(
      thrust::lower_bound(thrust::seq, edgelist_first, edgelist_last, pair),
      thrust::upper_bound(thrust::seq, edgelist_first, edgelist_last, pair)));
  }));
```

might be simpler unless you see a noticeable performance difference (and there is a very good chance that this code is actually faster).

jnke2016: Right. I tested the suggested approach on the standalone reproducer with nvcc 13.0 on Blackwell and it passes (the device-side thrust::lower_bound/upper_bound(thrust::seq) on zip_iterator is not affected by the same bug). This is simpler and perhaps faster because it doesn't pay the cost of 3 separate thrust calls (`lower_bound`, `upper_bound`, `transform`), each launching its own kernel, and avoids allocating two intermediate device vectors. I'll updated the workaround accordingly.

jnke2016: And regarding the `#if 1 / #else` guard, I remember seeing it int he cuGraph codebase for temporary workarounds. I will stick to that format.

ChuckHastings (approved):

seunghwak (approved):
LGTM

cpp/src/community/k_truss_impl.cuh
```diff
@@ -476,13 +478,38 @@ k_truss(raft::handle_t const& handle,
 
       rmm::device_uvector<edge_t> decrease_count(unique_pair_count, handle.get_stream());
 
+      // FIXME: thrust::reduce_by_key produces corrupted output when compiled with
+      // nvcc 13.0 + CCCL 3.4.0 on Blackwell (sm_120). The bug does not occur in nvcc 13.2.
+      // Remove this workaround (set to #if 0) once nvcc 13.0 is no longer supported.
+      // See https://github.com/rapidsai/cugraph/issues/5494
+#if 1
```
bdice: Can we replace this with something that correctly identifies the compiler? e.g.

```cpp
#if defined(__CUDACC_VER_MAJOR__) && defined(__CUDACC_VER_MINOR__) && \
    (__CUDACC_VER_MAJOR__ == 13) && (__CUDACC_VER_MINOR__ == 0)
```

Adapt as needed for CUDA 13.1, I'm not sure if 13.1 works or fails.

jnke2016: Thanks for the suggestion. I have confirmed that the bug is only in `nvcc 13.0`. I tested the standalone reproducer on `nvcc 13.1.115` and `13.2.51` and both pass. I'll update the `#if 1` guard to use a compile-time check on the `nvcc` version so the workaround only activates for the buggy compiler version and automatically uses the original `reduce_by_key` on fixed versions.

bdice (approved):

rlratzel (approved):

PR #5499, opened 2026-04-23
Link libcugraph dynamically against libcuvs

bdice:
## Summary
- remove the cuVS static-link configuration path and always link `libcugraph` against shared `libcuvs`
- add `libcuvs` as a runtime dependency for conda and wheel builds, and exclude `libcuvs.so` from wheel bundling
- re-enable the previously filtered CUDA test matrix entries now that `libcugraph.so` no longer carries the cuVS JIT LTO `cudaLibrary*` symbols

## Validation
- `build-cugraph -j0`
- `ldd cpp/build/conda/cuda-13.1/release/libcugraph.so` shows `libcuvs.so`
- `nm -D --undefined-only cpp/build/conda/cuda-13.1/release/libcugraph.so | rg cudaLibrary` returns no matches


divyegala (approved):

KyleFromNVIDIA (approved):

python/libcugraph/libcugraph/load.py
```diff
@@ -38,9 +38,11 @@ def load_library():
         # their symbols
         import librmm
         import libraft
+        import libcuvs
 
         librmm.load_library()
         libraft.load_library()
+        libcuvs.load_library()
     except ModuleNotFoundError:
```
jakirkham, line 46: This is the Python change for context

jakirkham (approved):
Thanks Bradley! 🙏

rlratzel (approved):
LGTM from the Python perspective.

PR #5502, opened 2026-04-29
cuGraph common refactor

### LINKED ISSUE #5503: STREAM_TEST fails on ARM due to unresolved sequence_fill<int> in libcugraph_mg.so
## Summary

The `conda-cpp-tests` ARM job on `main` fails immediately when launching `STREAM_TEST` because `libcugraph_mg.so` has an unresolved dependency on a helper instantiation that is not available at load time.

Failing job:
- https://github.com/rapidsai/cugraph/actions/runs/25106712190/job/73600759201

Failure signature:

```text
STREAM_TEST: symbol lookup error:
/opt/conda/envs/test/bin/gtests/libcugraph/../../../lib/libcugraph_mg.so:
undefined symbol: _ZN7cugraph6detail13sequence_fillIiEEvRKN3rmm16cuda_stream_viewEPT_mS6_
```

Demangled symbol:

```text
void cugraph::detail::sequence_fill<int>(
  rmm::cuda_stream_view const&,
  int*,
  unsigned long,
  int)
```

## Observed behavior

`STREAM_TEST` itself is a single-GPU test, but it loads `libcugraph.so`, which currently pulls in `libcugraph_mg.so` through the runtime dependency graph. `libcugraph_mg.so` then fails to load because it references `cugraph::detail::sequence_fill<int>` without that symbol being resolvable in this environment.

## Likely root cause

`libcugraph_mg.so` depends on helper/template instantiations that are owned outside the MG library. This creates a fragile cross-DSO symbol dependency that can fail depending on link/load behavior and platform.

## Possible fix direction

#5502 addresses this class of failure by introducing `libcugraph_common.so` to own shared helper/explicit-instantiation symbols and by separating single-GPU, multi-GPU, and MTMG source ownership.

The specific fix pattern is:

- move shared helper instantiations such as `sequence_fill<int>` into a common DSO
- make both `libcugraph.so` and `libcugraph_mg.so` depend on that common DSO
- avoid relying on `libcugraph_mg.so` resolving helpers from `libcugraph.so`

bdice:
## Summary

This PR separates cuGraph's shared helper/template-instantiation ownership from the single-GPU and multi-GPU algorithm libraries.

At a high level, it:

- introduces `libcugraph_common.so` for helper and explicit-instantiation symbols shared by SG, MG, and MTMG code paths
- moves MTMG sources into a dedicated `libcugraph_mtmg.so` target instead of hiding them in the SG/MG source lists
- removes the direct runtime dependency from `libcugraph.so` to `libcugraph_mg.so`; `libcugraph_c.so` still links both because the C API exposes both SG and MG functionality
- keeps SG, MG, MTMG, and common source ownership explicit in CMake

This avoids fragile cross-DSO symbol resolution where `libcugraph_mg.so` could depend on helper instantiations owned only by `libcugraph.so`.

Closes: #5503


cpp/src/sampling/detail/update_visited_utils_impl.cuh
```diff
@@ -38,7 +40,6 @@ update_dst_visited_vertices_and_labels(
     visited_minor_labels.has_value() == sampled_vertex_labels.has_value(),
     "Invalid input: visited_minor_labels and sampled_vertex_labels must have the same presence");
 
-  // 1) Shuffle sampled items to the owning GPU by vertex partitioning (minor path context)
   rmm::device_uvector<vertex_t> new_samples(sampled_vertices.size(), handle.get_stream());
```
ChuckHastings, line 44: Let's add these comments back in.

cpp/src/sampling/detail/remove_visited_vertices_from_frontier_common_v32_e32.cu
```diff
@@ -0,0 +1,42 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

cpp/src/sampling/detail/remove_visited_vertices_from_frontier_common_v64_e64.cu
```diff
@@ -0,0 +1,42 @@
+/*
```
ChuckHastings, line 1: Same question here.

cpp/src/sampling/detail/update_visited_utils_impl.cuh
```diff
@@ -38,7 +40,6 @@ update_dst_visited_vertices_and_labels(
     visited_minor_labels.has_value() == sampled_vertex_labels.has_value(),
     "Invalid input: visited_minor_labels and sampled_vertex_labels must have the same presence");
 
-  // 1) Shuffle sampled items to the owning GPU by vertex partitioning (minor path context)
```
bdice, line 41: We need to restore these comments.

cpp/src/structure/graph_weight_utils_mg_v32_e32.cu
```diff
@@ -8,44 +8,44 @@ namespace cugraph {
 
 // SG instantiation
 
-// compute_in_weight_sums
```
bdice, line 11: Follow conventions on what does/doesn't have comments here and in related files.

cpp/src/structure/graph_weight_utils_mg_v32_e32.cu
```diff
@@ -8,44 +8,44 @@ namespace cugraph {
...
-
-template rmm::device_uvector<float> compute_in_weight_sums<int32_t, int32_t, float, false, true>(
+template rmm::device_uvector<float> compute_out_weight_sums<int32_t, int32_t, float, false, true>(
```
ChuckHastings: This seems like a mistake.  We need both compute_in_weight_sums and compute_out_weight_sums.

cpp/src/structure/graph_weight_utils_mg_v64_e64.cu
```diff
@@ -8,50 +8,50 @@ namespace cugraph {
 
 // SG instantiation
 
-// compute_in_weight_sums
-
-template rmm::device_uvector<float> compute_in_weight_sums<int64_t, int64_t, float, false, true>(
+template rmm::device_uvector<float> compute_out_weight_sums<int64_t, int64_t, float, false, true>(
```
ChuckHastings: Same

cpp/src/utilities/shuffle_properties_common.cu
```diff
@@ -0,0 +1,120 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

cpp/src/utilities/shuffle_vertex_pairs_common_v32_e32.cu
```diff
@@ -0,0 +1,31 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

cpp/src/utilities/shuffle_vertex_pairs_common_v64_e64.cu
```diff
@@ -0,0 +1,31 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

cpp/src/utilities/shuffle_vertices_common_v32.cu
```diff
@@ -0,0 +1,23 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

cpp/src/utilities/shuffle_vertices_common_v64.cu
```diff
@@ -0,0 +1,23 @@
+/*
```
ChuckHastings, line 1: Should this be a rename?

ChuckHastings (commented):
A batch of comments.

I wonder if this will compile with some of the libcuvs dynamic changes in this PR.  Might need to back some of them out.

ChuckHastings (approved):

cpp/src/sampling/detail/conversion_utilities_sg.cu
seunghwak, line 1: Shouldn't this be `conversion_utilities_sg.cu`?

cpp/CMakeLists.txt
```diff
@@ -158,16 +158,68 @@ endif()
...
+    src/utilities/shuffle_vertex_pairs_common_v32_e32.cu
+    src/utilities/shuffle_vertex_pairs_common_v64_e64.cu
+)
+
+set(CUGRAPH_COMMON_SAMPLING_SOURCES
+    src/sampling/detail/shuffle_and_organize_output.cu
+    src/sampling/detail/remove_visited_vertices_from_frontier_common_v32_e32.cu
+    src/sampling/detail/remove_visited_vertices_from_frontier_common_v64_e64.cu
+    src/sampling/detail/temporal_partition_vertices_v64.cu
+    src/sampling/detail/temporal_partition_vertices_v32.cu
```
seunghwak: Why some files have `_common_` and some files don't?

cpp/CMakeLists.txt
```diff
@@ -158,16 +158,68 @@ endif()
 #       since ninja will run them in parallel in this order,
 #       which should give us a better parallel schedule.
 
+set(CUGRAPH_COMMON_UTILITY_SOURCES
+    src/detail/utility_wrappers_32.cu
+    src/detail/utility_wrappers_64.cu
+    src/detail/permute_range_v32.cu
+    src/detail/permute_range_v64.cu
+    src/utilities/invert_flags.cu
```
seunghwak: Shouldn't these files have `_common_` in file names?

seunghwak (approved):
LGTM except for some file naming issues.

Another question, should we better rename libcugraph.so to libcugraph_sg.so? Or do we have a plan to move back to a single `libcugraph.so` and we want to keep this name?

ChuckHastings:
> LGTM except for some file naming issues.
> 
> Another question, should we better rename libcugraph.so to libcugraph_sg.so? Or do we have a plan to move back to a single `libcugraph.so` and we want to keep this name?

I'll address the file naming issues.

I think there are some external dependencies that expect the SG library to be called libcugraph.so.  This is a fair naming question to consider down the road, but we would need to put out a deprecation warning to anyone using libcugraph.so if we decide to go that route and give everyone a release to prepare, perhaps providing a symbolic link to the old name in the interim.

cpp/CMakeLists.txt
ChuckHastings: Renaming to be consistent.

cpp/CMakeLists.txt
ChuckHastings: Renaming to be consistent.

jameslamb (approved):
I've somewhat been following along the library-splitting work, this reorganization makes sense to me.

seunghwak:
>I think there are some external dependencies that expect the SG library to be called libcugraph.so. This is a fair naming question to consider down the road, but we would need to put out a deprecation warning to anyone using libcugraph.so if we decide to go that route and give everyone a release to prepare, perhaps providing a symbolic link to the old name in the interim.

Sounds reasonable. Just a question is that, aren't we breaking existing users anyways? I mean they should add libcugraph_comm.so in addition anyways... A question might be whether should we call libcugraph_common.so as libcugraph.so or libcugraph_sg.so as libcugraph.so... Not a major issue if we have any plan to come back to a single libcugraph.so but if this our longer term solution, I think this needs little more discussion.

rlratzel (approved):

PR #5520, opened 2026-05-15
add support for multigraph to disjoint sampling

### LINKED ISSUE #5500: [FEA] Support Disjoint Sampling on Multigraphs
Disjoint sampling is currently manually restricted, automatically failing if the disjoint flag is set and the graph is a multigraph.

However, the problem is that there's no such limitation at the PyG API level.  All graphs are currently assumed to potentially include multi-edges, and so during the graph creation stage, the pylibcugraph graph object is constructed as a multigraph.  Adding enforcement for this at the Python API level would be tricky, requiring several extra checks and increasing the runtime of the graph creation process, which adds significant overhead on MNMG systems.

I'm requesting that we add a workaround, where there is some expected behavior for multi-edges (i.e. it will reject duplicates and resample if they're encountered).  That way, we don't have to add the check and leave it to users (with appropriate warning) to ensure they're aware of this behavior or that they're checking for multi-edges prior to graph construction if this behavior is unacceptable to them.

ChuckHastings:
Adds support for multigraph to the disjoint sampling implementation.  Changes include:
* Use of partial results from the first sampling attempt rather than discarding all sampled edges - should result in faster convergence in edge cases
* Removal of fast-failure checks for multi-graph
* A couple of additional tests that verify that it works

Closes #5500


alexbarghi-nv (approved):
Tests are passing for disjoint sampling now - approved.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
seunghwak: Something very tedious, but why define `keep_labels` before `keep_minors`? This violates (majors, minors, labels) ordering before here.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
```diff
@@ -114,26 +119,21 @@ deduplicate_edges_by_minor(raft::handle_t const& handle,
   }
 
   // 2. Now all edges that lead to a visited minor are on this GPU, we can
-  //    Sort by minor vertex id and identify duplicates.
+  //    Sort by minor vertex id and identify duplicates.  local_positions is a unique per-row
+  //    index so it tie-breaks rows that share the same (label, minor).
   rmm::device_uvector<vertex_t> local_positions(keep_minors.size(), handle.get_stream());
   thrust::sequence(
     handle.get_thrust_policy(), local_positions.begin(), local_positions.end(), size_t{0});
```
seunghwak: In single GPU, local_positions == keep_positions, so isn't this just a duplicate?

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
```diff
@@ -165,246 +165,157 @@ deduplicate_edges_by_minor(raft::handle_t const& handle,
...
-                           keep_minors.size());
+    // Gather to reflect the original positions of the edges (sorted row -> original local index).
+    {
+      rmm::device_uvector<size_t> tmp(local_positions.size(), handle.get_stream());
+      thrust::gather(handle.get_thrust_policy(),
+                     local_positions.begin(),
+                     local_positions.end(),
+                     keep_positions.data(),
+                     tmp.begin());
+      keep_positions = std::move(tmp);
```
seunghwak: In SG, this is just same as keep_positions = std::move(local_positions), right?

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
```diff
@@ -35,19 +35,17 @@ std::tuple<rmm::device_uvector<vertex_t>,
...
            std::optional<rmm::device_uvector<int32_t>>>
 deduplicate_edges_by_minor(raft::handle_t const& handle,
                            graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
                            rmm::device_uvector<vertex_t>&& result_majors,
                            rmm::device_uvector<vertex_t>&& result_minors,
                            arithmetic_device_uvector_t&& tmp_edge_indices,
-                           std::optional<rmm::device_uvector<int32_t>>&& result_labels,
-                           bool call_from_sampling)
+                           std::optional<rmm::device_uvector<int32_t>>&& result_labels)
 {
```
seunghwak, line 47: If I understood correctly, we want to keep just one edge per (label, minor) or minor (if no label).

I think this code is overly complicated for this purpose.

In SG, what we need is
```
auto key_pair_first = thrust::make_zip_iterator(result_labels->begin(), result_minors.begin());
thrust::sort_by_key(
  handle.get_thrust_policy(),
  key_pair_first, key_pair_first + result_labels->size(),
  thrust::make_zip_iterator(result_majors.begin(), tmp_edge_indices->begin()));
auto [keep_count, keep_flags] =
    detail::mark_entries(handle,
                           result_labels->size(),
                           detail::is_first_in_run_t<decltype(key_pair_first)>{key_pair_first});
thrust::partition(...);  // use keep_flags as stencil
copy the second half to discard vectors, then the first half becomes keep vectors.
```

We need an additional step to discard duplicate (label, minor) edges across GPUs.

If we want to minimize the communication volume,

we can sort the keep (label, minor) pairs by the owning GPU ID.

We call `shuffle_values` which return `rx_counts` as well. We update keep_flags, send the keep flags back using `rx_counts` as `tx_counts`. Using the keep flags, discard the original (label, minor) pairs. Go back to the first half, keep in the first half only if the pair is in the survived (label, minor) pair. Move to the discard partition otherwise.

Or if you want something simpler,

shuffle (label, minor, rank) triplets based on the minor's vertex partition ID. Run, sort and unique to keep only one rank value per (label, minor). Shuffle back based on the rank. Now you have survived (label, minor) pairs. This involves additional rank values, and need to shuffle back triplets instead of just flags, but might be simpler.

seunghwak (commented):
Review the de-duplicate function first.

If I understood the problem correctly, I think we can simplify this code quite a bit. Let me know if I misunderstood the problem.

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -658,16 +671,21 @@ sample_unvisited_with_one_property(
...
   bool sample_and_append{true};
-  std::optional<rmm::device_uvector<vertex_t>> resample_active_majors{std::nullopt};
-  std::optional<rmm::device_uvector<int32_t>> resample_active_major_labels{std::nullopt};
+  rmm::device_uvector<vertex_t> carryover_frontier_majors(0, handle.get_stream());
+  std::optional<rmm::device_uvector<int32_t>> carryover_frontier_labels{std::nullopt};
+  std::optional<rmm::device_uvector<int32_t>> carryover_frontier_types{std::nullopt};
+  rmm::device_uvector<size_t> carryover_frontier_capacity(0, handle.get_stream());
 
   cugraph::key_bucket_view_t<vertex_t, tag_t, multi_gpu, false> active_bucket_view =
     key_bucket_view;
```
seunghwak: auto active_key_bucket_view = key_bucket_view;

cpp/src/sampling/detail/sample_edges.cuh
seunghwak, line 785: `else {}` here can't happen (as bias should be float or double), right?
Better document this,

```
else {
  CUGRAPH_FAIL("should not be reached.");
}
```

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -745,19 +766,355 @@ sample_unvisited_with_one_property(
...
+                                                                  std::move(sampled_labels));
+
+    carryover_frontier_labels   = std::nullopt;
+    carryover_frontier_types    = std::nullopt;
+    carryover_frontier_majors   = rmm::device_uvector<vertex_t>(0, handle.get_stream());
+    carryover_frontier_capacity = rmm::device_uvector<size_t>(0, handle.get_stream());
+
+    if (discarded_majors.size() != 0) {
+      size_t const num_types = Ks.size();
+      CUGRAPH_EXPECTS(num_types >= 1, "Ks must be non-empty.");
```
seunghwak: Should we better check this at the beginning of this function?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -745,19 +766,355 @@ sample_unvisited_with_one_property(
...
+          agg_majors.resize(
+            static_cast<size_t>(cuda::std::distance(agg_majors.begin(), ends.first)),
+            handle.get_stream());
+          agg_counts.resize(agg_majors.size(), handle.get_stream());
+          carryover_frontier_majors = std::move(agg_majors);
+        }
+        carryover_frontier_capacity = std::move(agg_counts);
+      } else {
+        CUGRAPH_EXPECTS(
+          std::holds_alternative<rmm::device_uvector<int32_t>>(discarded_tmp_indices),
```
seunghwak: We defined edge_type_t at the beginning of this function, shouldn't we use it?

cpp/src/sampling/detail/sample_edges.cuh
```diff
@@ -745,19 +766,355 @@ sample_unvisited_with_one_property(
...
+
+      if (num_types == size_t{1}) {
+        if (agg_labels) {
+          thrust::sort(handle.get_thrust_policy(),
+                       thrust::make_zip_iterator(agg_labels->begin(), agg_majors.begin()),
+                       thrust::make_zip_iterator(agg_labels->end(), agg_majors.end()));
+        } else {
+          cugraph::detail::sort_ints(
+            handle, raft::device_span<vertex_t>{agg_majors.data(), agg_majors.size()});
+        }
```
seunghwak, line 1084: This code is common regardless of num_types, should we better sort before the if else block.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
ChuckHastings: Yes, I cleaned up the more complex code to make it function.  But really with the simpler task we should have a simpler implementation.

I have refactored this to be much simpler.  I did not use thrust::partition, I instead used keep_marked_entries and copy_if_mask_unset.  I think this will save a bit of memory because I can do one array at a time (it's essentially the same approach as the mem_frugal partition implementation.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
ChuckHastings: Obsolete with refactoring, but I made sure in the refactored logic to keep the ordering consistent.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
ChuckHastings: New logic eliminates all of this.

cpp/src/sampling/detail/deduplicate_edges_by_minor_impl.cuh
ChuckHastings: New logic eliminates this.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Done

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Done.  I reviewed several CUGRAPH_EXPECTS that were in the middle of this function, most of them were actually while I was debugging and I removed the unnecessary ones.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: Not quite the same.  In the else clause we add types.begin()/types.end() to the zip iterator.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: I'll push an update with this change.

cpp/src/sampling/detail/sample_edges.cuh
ChuckHastings: I think this is now gone.

seunghwak (approved):
LGTM

jnke2016 (approved):

PR #5523, opened 2026-05-20
Improve devcontainer startup times

trxcllnt:
This PR builds and publishes devcontainers to `ghcr.io` and uses the new images to improve CI and local devcontainer startup times.

In `rapidsai/devcontainers`, devcontainer startup times went from [`3m28s`](https://github.com/rapidsai/devcontainers/actions/runs/26123239225/job/76830349936#step:9:2006) to [`34s`](https://github.com/rapidsai/devcontainers/actions/runs/26132932976/job/76861774810?pr=706#step:9:305).

Contributes to https://github.com/rapidsai/build-planning/issues/280.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -223,3 +223,12 @@ jobs:
       id-token: write
       packages: read
       pull-requests: read
+  devcontainers:
+    name: Build devcontainers
+    secrets: inherit  # zizmor: ignore[secrets-inherit]
+    uses: rapidsai/shared-workflows/.github/workflows/build-devcontainers.yaml@main
+    permissions:
+      packages: write
+    with:
+      push: true
+      cuda: '["12.9", "13.2"]'
--- a/.devcontainer/cuda12.9-conda/devcontainer.json
+++ b/.devcontainer/cuda12.9-conda/devcontainer.json
@@ -6,7 +6,10 @@
       "CUDA": "12.9",
       "PYTHON_PACKAGE_MANAGER": "conda",
       "BASE": "rapidsai/devcontainers:26.08-cpp-mambaforge"
-    }
+    },
+    "cacheFrom": [
+      "ghcr.io/rapidsai/cugraph/devcontainer:26.08-cuda12.9-conda"
+    ]
   },
   "runArgs": [
     "--rm",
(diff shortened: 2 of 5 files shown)


.github/workflows/build.yaml
```diff
@@ -223,3 +223,12 @@ jobs:
...
       pull-requests: read
+  devcontainers:
+    name: Build devcontainers
+    secrets: inherit  # zizmor: ignore[secrets-inherit]
+    uses: rapidsai/shared-workflows/.github/workflows/build-devcontainers.yaml@main
+    permissions:
+      packages: write
+    with:
+      push: true
+      cuda: '["12.9", "13.2"]'
```
jameslamb, line 234: The issue and PR title are about changing the caching strategy, but this looks like a second, unrelated change... we're now looking to run devcontainer builds on nightlies and merges to `main`?

Just double-checking, is that intentional?

One effect will be that if the devcontainer builds are failing for a repo, it'll cause the nightly test jobs to not be triggered (they're only triggered on a successful `build.yaml` run). Do we want that?

jameslamb (changes requested):
Left one question I'd like an answer to before we merge this.

.github/workflows/build.yaml
trxcllnt: This job doesn't run the C++ and Python builds in the devcontainer, it builds the devcontainer _images_ and pushes it to `ghcr.io` so they can be used as the cache when launching a devcontainer (in both CI and local dev).

Compare two `rapidsai/devcontainers` "Build all RAPIDS" workflow runs from before and after this change:
* In the [first one](https://github.com/rapidsai/devcontainers/actions/runs/26123239225/job/76830349936#step:9:389), the CI machine has to build the devcontainer image first (which involves building the layer with the `ghcr.io/rapidsai/devcontainers/features/cuda` feature). 
* The [second one](https://github.com/rapidsai/devcontainers/actions/runs/26132932976/job/76861774810?pr=706#step:9:256) can skip this step because it loads that layer from the image cache via `docker buildx build --cache-from ghcr.io/rapidsai/devcontainers/devcontainer:26.08-cuda12.9-conda`

jameslamb: Ohhhhh got it! I totally misread the PR description (my mistake, not yours... you clearly said "This PR builds and publishes devcontainers" 🙃 ).

Thanks for explaining that, I'm good with this.

jameslamb (approved):

trxcllnt:
Test failures are unrelated to the changes in this PR, I'm going to merge so we can close this out.

PR #5527, opened 2026-05-23
Further reduce file size

ChuckHastings:
Continuing the work to reduce the binary size.

This PR makes the following changes:
1. A few residual compute_number_of_edges updates
2. A little cleanup of the utility_wrapper functions, they should pass stream at the end of the parameters
3. thrust::sort moved to be compiled in libcugraph_common.so for most cases, added new function device_sort which either calls thrust::sort inline or calls one of the pre-compiled variants

## Size reduction breakdown

|---|---|---|---|---|
| CUDA 12 Before | 97 MB | 803 MB | 1037 MB | 1937 MB |
| CUDA 12 After |  100 MB (+3, +3.1%) | 756 MB (-47, -5.9%) | 940 MB (-97, -10.3) | 1796 MB (-141, -7.28%) |
|---|---|---|---|---|
| CUDA 13 Before | 50 MB | 406 MB | 510 MB | 966 MB |
| CUDA 13 After |  51 MB (+1, +0.2%)| 384 MB (-22, -5.4%)| 472  MB (-38, -7.5%) | 907 MB (-48, -5.0%)|


cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -30,66 +36,98 @@ namespace detail {
...
  * @param[in]   value        value
+ * @param[in]   stream_view  CUDA stream
  *
  */
 template <typename value_t>
-void scalar_fill(raft::handle_t const& handle, value_t* d_value, size_t size, value_t value);
+void scalar_fill(value_t* d_value,
+                 size_t size,
+                 value_t value,
+                 rmm::cuda_stream_view const& stream_view);
```
seunghwak, line 71: This is a wrapper function for thrust::fill while the above uniform_random_fill is not.

How should we handle thrust wrappers versus non-thrust wrapper functions?

For thrust wrappers (to avoid multiple instantiations), we may better maintain the API as close as possible to thrust (so we can just replace thrust:: to cugraph::), and define them in thrust wrapper specific files.

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -30,66 +36,98 @@ namespace detail {
...
+ * @tparam      value_t      element type; must satisfy @ref device_sort_scalar_value_v
+ *
+ * @param[in]   policy       Thrust execution policy (e.g. @c handle.get_thrust_policy())
+ * @param[in]   values       device span to sort in-place
+ *
+ */
+template <typename ExecutionPolicy,
+          typename value_t,
+          std::enable_if_t<device_sort_scalar_value_v<value_t>, int> = 0>
+void device_sort(ExecutionPolicy const& policy, raft::device_span<value_t> values)
```
seunghwak, line 113: I am debating between defining a separate device_sort function versus maintain `sort` same to thrust. Calling device_sort with thrust::host execution policy may look a bit awkward. And for modern/future GPU architectures, distinctions between device memory and host memory are getting blurred. It is becoming more like smaller&faster memory versus larger&slower memory.

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -30,66 +36,98 @@ namespace detail {
...
- * and handles to various CUDA libraries) to run graph algorithms.
- * @param[in]  values      device span of unique elements.
+ * @param[in]  values       device span of sorted elements; consecutive duplicates are collapsed.
+ * @param[in]  stream_view  CUDA stream
  * @return the number of unique elements.
  *
  */
 template <typename value_t>
-size_t unique_ints(raft::handle_t const& handle, raft::device_span<value_t> values);
+size_t unique_ints(raft::device_span<value_t> values, rmm::cuda_stream_view const& stream_view);
```
seunghwak, line 130: May better call just unique (so can be a better drop-in replacement of thrust::unique?)

cpp/include/cugraph/detail/utility_wrappers.hpp
```diff
@@ -136,17 +174,17 @@ void transform_not_equal(raft::device_span<value_t> values,
...
  *
  */
 template <typename value_t>
-void sequence_fill(rmm::cuda_stream_view const& stream_view,
-                   value_t* d_value,
+void sequence_fill(value_t* d_value,
                    size_t size,
-                   value_t start_value);
+                   value_t start_value,
+                   rmm::cuda_stream_view const& stream_view);
```
seunghwak, line 187: May better call just sequence (so can be a better drop-in replacement of thrust::sequence?)

cpp/include/cugraph/detail/utility_wrappers.hpp
ChuckHastings: Working on this for the next pull request.  I can add a FIXME here.

cpp/include/cugraph/detail/utility_wrappers.hpp
ChuckHastings: Updated to sequence, small change to the function signature in latest.

ChuckHastings:
Splitting this into smaller chunks.

First : #5532

ChuckHastings:
Next: #5540

ChuckHastings:
Closing.  All of these efforts are now in other PRs.

PR #5536, opened 2026-06-01
fix(ci): fix configuration for breaking change notification workflow

jameslamb:
Contributes to https://github.com/rapidsai/shared-workflows/issues/560

Fixes misconfigurations in the GitHub Actions workflow that generates Notifications when PRs are labeled `breaking`.

Any other changes come from making the configuration for this workflow identical across all RAPIDS repos.


jakirkham (approved):
Thanks James! 🙏

Looks like the cuGraph case was mostly correct and so the automated tooling is merely shifting things around (with the exception of the webhook rename).

PR #5540, opened 2026-06-04
Define a wrapper for thrust::sort to reduce binary size

ChuckHastings:
Continuing the work to reduce the binary size.

This PR wraps thrust::sort so it can be compiled in libcugraph_common.so for most cases, added new function cugraph::sort_wrapper which either calls thrust::sort inline or calls one of the pre-compiled variants.  Note that we'd like to name the function cugraph::sort, however there's a problem getting that to work, so we're using cugraph::sort_wrapper for the time being.

## Size reduction breakdown

|---|---|---|---|---|
| CUDA 12 Before | 97 MB | 783 MB | 967 MB | 1847 MB |
| CUDA 12 After |  98 MB (+1, +1.1%) | 756 MB (-27, -3.4%) | 940 MB (-27, -2.8%) | 1794 MB (-53, -2.87%) |
|---|---|---|---|---|
| CUDA 13 Before | 49 MB | 383 MB | 466 MB | 898 MB |
| CUDA 13 After |  49 MB (+0, +0%)| 370 MB (-13, -3.4%)| 452  MB (-14, -3.0%) | 871 MB (-27, -3.0%)|

### DIFF
--- a/cpp/include/cugraph/utilities/thrust_wrappers.hpp
+++ b/cpp/include/cugraph/utilities/thrust_wrappers.hpp
@@ -1,0 +1,155 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cugraph/export.hpp>
+#include <cugraph/utilities/thrust_wrappers_zip_types.hpp>
+
+#include <rmm/exec_policy.hpp>
+
+#include <cuda/std/iterator>
+#include <thrust/sort.h>
+
+#include <cstdint>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace CUGRAPH_EXPORT cugraph {
+namespace detail {
+
+/** @brief True for value types dispatched to @ref sort_impl without Thrust in headers. */
+template <typename T>
+inline constexpr bool sort_scalar_value_v =
+  std::disjunction_v<std::is_same<std::remove_cv_t<T>, std::int32_t>,
+                     std::is_same<std::remove_cv_t<T>, std::uint32_t>,
+                     std::is_same<std::remove_cv_t<T>, std::int64_t>>;
+
+/** @brief Whether iterator type @p T has an out-of-line @ref sort_impl lexicographic sort.
+ *
+ *  Supported @c thrust::make_zip_iterator(...) iterator types are listed in
+ *  thrust_wrappers_zip_types.hpp and instantiated in thrust_wrappers.cu. Scalar element sorts use
+ *  @ref sort_scalar_value_v and @ref sort_impl for @c rmm::exec_policy and
+ *  @c rmm::exec_policy_nosync.
+ *
+ *  Keep this in lockstep with those explicit instantiations: add a disjunct only when you add the
+ *  matching explicit instantiation (otherwise a call through @ref cugraph::sort_wrapper can fail at
+ *  link
+ *  time).
+ */
+template <typename T>
+inline constexpr bool sort_supported_v =
+  std::disjunction_v<std::is_same<std::remove_cv_t<T>, zip_i32_i32>,
+                     std::is_same<std::remove_cv_t<T>, zip_i64_i64>,
+                     std::is_same<std::remove_cv_t<T>, zip_i64_i32>,
+                     std::is_same<std::remove_cv_t<T>, zip_i32_i64>,
+                     std::is_same<std::remove_cv_t<T>, zip_sz_i32>,
+                     std::is_same<std::remove_cv_t<T>, zip_sz_i64>,
+                     std::is_same<std::remove_cv_t<T>, zip_f_sz>,
+                     std::is_same<std::remove_cv_t<T>, zip_d_sz>,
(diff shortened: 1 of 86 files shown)


cpp/include/cugraph/utilities/thrust_wrappers.hpp
seunghwak, line 1: This file calls thrust::sort and I assume this file won't compile with g++. The extension should be `.cuh`.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace CUGRAPH_EXPORT cugraph {
+namespace detail {
+
+/** @brief True for value types dispatched to @ref sort_impl without Thrust in headers. */
+template <typename T>
+inline constexpr bool sort_scalar_value_v =
```
seunghwak: `sort_supported_(arithmetic_)scalar_v`? to be consistent with the below `sort_supported_v` (and something like `is_arithmetic_scalar_v` in cugraph?

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -0,0 +1,155 @@
...
+ *  @ref sort_scalar_value_v and @ref sort_impl for @c rmm::exec_policy and
+ *  @c rmm::exec_policy_nosync.
+ *
+ *  Keep this in lockstep with those explicit instantiations: add a disjunct only when you add the
+ *  matching explicit instantiation (otherwise a call through @ref cugraph::sort_wrapper can fail at
+ *  link
+ *  time).
+ */
+template <typename T>
+inline constexpr bool sort_supported_v =
```
seunghwak: `sort_supported_zip_v` as this is for just zip iterators?

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include <cstdint>
+#include <iterator>
+#include <type_traits>
+#include <utility>
+
+namespace CUGRAPH_EXPORT cugraph {
+namespace detail {
+
+/** @brief True for value types dispatched to @ref sort_impl without Thrust in headers. */
```
seunghwak: Keep this comment and the comment for zip iterators in sync?

```
True for value types dispatched to @ref sort_impl without Thrust in headers.

Whether iterator type @p T has an out-of-line @ref sort_impl lexicographic sort.
```
Whether iterator type @p T has an out-of-line @ref sort_impl?

seunghwak (commented):
Looks good besides some minor cosmetic/consistency issues.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: CCCL has designed thrust to be useable from g++, with the proviso that you don't call thrust::sort (or any of these functions) on device pointers when you're coming from a .cpp file.  I ran a simple test that verified that we can create an std::vector and call thrust::sort on it from a .cpp file perfectly fine.

So, for any .cpp file that calls cugraph::sort_wrapper, as long as it is passing iterators that translate to a cugraph::detail::sort_impl that is actually instantiated in thrust_wrappers.cu (all of the scalar types or one of the zip iterator types that are precompiled) it will compile perfectly fine.

We would not be able to call cugraph::sort_wrapper with a custom comparator or for a combination that's not precompiled from a .cpp file, it would need to be from a .cu file.

So I think it's safe to leave this as .hpp.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Fixed

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Fixed

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Fixed

cpp/include/cugraph/utilities/thrust_wrappers.hpp
seunghwak: Oh, I see, this is a news, so in that case, should we also rename our existing .cuh files to .hpp and make it the caller's responsibility to not call thrust with a device side execution policy?

seunghwak (approved):
LGTM

ChuckHastings:
> Oh, I see, this is a news, so in that case, should we also rename our existing .cuh files to .hpp and make it the caller's responsibility to not call thrust with a device side execution policy?

I think we have a mixed bag.  We have .cuh files that clearly define device-only features.  If we wanted to make them .hpp files we would need to add all of the C preprocessor syntax to include certain parts of the implementation if you're using nvcc and not include it if you're using g++.  I'm not sure there's value to convert the files in those situations.  I don't think, for example, that we want the primitives (at least not at the moment) to be able to operate on std::vector objects.

But I think there are some .cuh files that could probably benefit from being .hpp files and being able to include them from .cpp files and compile them with g++.  I think it's just something to keep an eye on for the future.

PR #5541, opened 2026-06-04
Improve Louvain determinism

### LINKED ISSUE #5516: [FEA]: Non-deterministic behavior from Louvain due to numerical instability

New Feature

Medium

`cugraph.louvain` produces non-deterministic partitions across repeated runs on the *same* graph, same RAPIDS version, same GPU. `cugraph.leiden` already accepts a `random_state` argument, but `cugraph.louvain` exposes no equivalent, so there is no way to obtain reproducible Louvain results.

 cugraph version: 26.04.00 (also seen on earlier releases)

**Reproducibility evidence.** Running Louvain 6× on a fixed ~2k-node weighted graph (community-detection step of a gene-program pipeline): - 5/6 runs produced an identical partition - 1/6 collapsed ~15 communities into their neighbors (different community count)

The same workload under `cugraph.leiden(random_state=42)` was bit-identical across all 6 runs. The downstream effect in our case was a ~20% swing in the number of detected communities from one Louvain run to the next.

  A way to make `cugraph.louvain` reproducible, ideally:

  1. **`random_state` parameter** on `cugraph.louvain` (and `pylibcugraph.louvain`), matching the existing `cugraph.leiden` signature, to seed any RNG-driven tie-breaking; and/or
  2. **A documented deterministic mode** — since Louvain's variation also stems from parallel vertex-move ordering and non-associative floating-point reductions of modularity gains, a seed alone may not be sufficient. An optional flag that fixes traversal order / uses deterministic reductions (accepting a performance cost) would let users trade speed for reproducibility when needed.

API parity with `cugraph.leiden` is the minimum bar; even just documenting which sources of non-determinism a seed would and would not address would help.

_No response_

### Additional context

Use case: community detection on weighted similarity graphs in a single-cell genomics pipeline, where reproducible runs are required for published results. Happy to share a minimal reproducer graph if useful.

### Code of Conduct

ChuckHastings:
Update the threshold parameter to also use it for comparing to decide if a delta modularity is worth moving to a new cluster.

We occasionally see non-deterministic results across runs with the same graph due to numerical instability.  The fundamental issue arises when the delta modularity gets small, and when that happens we occasionally move vertices between clusters when the change in modularity is in fact not significant.

This PR uses the existing threshold parameter, which was previously used solely to decide whether we had converged or not, to decide whether a delta modularity is worth moving a vertex as well.  The consequence of this change is that the user has the ability to adjust the threshold to increase the determinism (a larger threshold causes movement of vertices between clusters to stop earlier in each level of the hierarchical clustering).

Closes #5516

### DIFF
--- a/cpp/include/cugraph/algorithms.hpp
+++ b/cpp/include/cugraph/algorithms.hpp
@@ -528,7 +528,11 @@ weight_t hungarian(raft::handle_t const& handle,
  *                               are assumed to be 1.0.
   @param[out] clustering         Pointer to device array where the clustering should be stored
  * @param[in]  max_level         (optional) maximum number of levels to run (default 100)
- * @param[in]  threshold         (optional) threshold for convergence at each level (default 1e-7)
+ * @param[in]  threshold         (optional) Minimum modularity gain threshold for each level.
+ *                               A vertex move is accepted only if its delta modularity exceeds
+ *                               @p threshold, and an iteration is committed / the level loop
+ *                               continues only if the global modularity gain exceeds @p threshold.
+ *                               (default 1e-7)
  * @param[in]  resolution        (optional) The value of the resolution parameter to use.
  *                               Called gamma in the modularity formula, this changes the size
  *                               of the communities.  Higher resolutions lead to more smaller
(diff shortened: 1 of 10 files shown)


seunghwak (approved):
LGTM

PR #5546, opened 2026-06-08
Reduce binary size scan

ChuckHastings:
Next phase of the reduce binary size effort.

This PR creates cugraph::inclusive_scan and cugraph::exclusive_scan as wrappers around the thrust analogs, pre-compiling variations that make sense to reduce the binary size.

The improvement in binary size here is small, only about 0.3%.

### DIFF
--- a/cpp/include/cugraph/utilities/thrust_wrappers.hpp
+++ b/cpp/include/cugraph/utilities/thrust_wrappers.hpp
@@ -82,8 +82,7 @@ template <typename ExecutionPolicy>
 inline constexpr bool is_rmm_exec_policy_v =
   std::is_same_v<std::remove_cv_t<ExecutionPolicy>, rmm::exec_policy>;
 
-/** Input iterator value type for @ref cugraph::inclusive_scan_wrapper / @ref
- * cugraph::exclusive_scan. */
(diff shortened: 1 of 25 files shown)


cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -152,4 +205,155 @@ void sort_wrapper(ExecutionPolicy const& policy,
   thrust::sort(policy, first, last, compare);
 }
 
+/**
```
ChuckHastings, line 208: Note... this is a workaround that allows us to use `cugraph::inclusive_scan` as our name instead of the `cugraph::inclusive_scan_wrapper` which was the technique we used for wrapping thrust::sort.

Fundamentally the issue is that thrust calls (like sort and inclusive_scan) use ADL to select the proper implementation, resulting in a call to inclusive_scan without a specifically scoped namespace.  If there's an alternative function in a different namespace that is in scope, the compiler might chose that function and we get incorrect behavior.

Making this a functor and then instantiating the functor below tricks ADL into doing the right thing.

seunghwak (approved):
This is neat!

PR #5551, opened 2026-06-09
Add per-edge triplet neighbor-intersection transform-reduce primitive

jnke2016:


cpp/include/cugraph/prims/transform_reduce_triplet_of_src_dst_nbr_intersection_of_e_endpoints_by_e.cuh
```diff
@@ -0,0 +1,1263 @@
...
+          typename VertexPairIterator,
+          typename T>
+std::tuple<rmm::device_uvector<typename GraphViewType::vertex_type>,
+           rmm::device_uvector<typename GraphViewType::vertex_type>,
+           dataframe_buffer_type_t<T>>
+transform_reduce_triplet_of_dst_nbr_intersection_of_e_endpoints_by_e(
+  raft::handle_t const& handle,
+  GraphViewType const& graph_view,
+  VertexPairIterator vertex_pair_first,
+  VertexPairIterator vertex_pair_last,
```
seunghwak, line 1917: If we are performing reduction "by_e", this means vertex pairs should be edges. Then, we should better use an edge bucket than general vertex pairs.

See

https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/prims/transform_gather_e.cuh#L151

cpp/include/cugraph/prims/detail/nbr_intersection.cuh
```diff
@@ -386,6 +388,1177 @@ __device__ edge_t set_intersection_by_key_with_mask(InputKeyIterator0 input_key_
...
+template <typename vertex_t, typename edge_t, bool multi_gpu, typename ElementToIdxMap>
+__device__ void nbr_intersection_nbr_list(
+  edge_partition_device_view_t<vertex_t, edge_t, multi_gpu> const& edge_partition,
+  ElementToIdxMap element_to_idx_map,
+  raft::device_span<edge_t const> element_offsets,
+  raft::device_span<vertex_t const> element_indices,
+  vertex_t v,
+  vertex_t const*& nbr_indices,
+  edge_t& nbr_offset,
+  edge_t& nbr_degree)
```
seunghwak, line 428: Better return cuda::std::tuple<vertex_t const*, edge_t, edge_t> rather than taking `nbr_indices`, `nbr_offset` and `nbr_degree` as L-value references. Returning a value is more functional and generally preferable unless there is a strong reason to take L-value references.

cpp/include/cugraph/prims/detail/nbr_intersection.cuh
```diff
@@ -386,6 +388,1177 @@ __device__ edge_t set_intersection_by_key_with_mask(InputKeyIterator0 input_key_
...
+{
+  auto p = short_off + s;
+  auto w = short_indices[p];
+  // First half of the multigraph fix: if w is repeated, only the first copy in the run does anything
+  // and every later copy leaves immediately. The list is sorted, so a repeat is recognized by
+  // comparing against the entry just before it. Without this, k parallel edges to w would run this
+  // whole function k times and emit k times over.
+  if (s > edge_t{0} && short_indices[p - 1] == w) { return; }
+
+  // Binary search for w in the longer list. Could equivalently use the sequential thrust version:
```
seunghwak, line 521: Why aren't you using `thrust::lower_bound(thrust::seq, ...)`?

cpp/include/cugraph/prims/detail/nbr_intersection.cuh
seunghwak: And you can actually return just nbr_indices + nbr_offset and nbr_degree.

cpp/include/cugraph/prims/detail/nbr_intersection.cuh
seunghwak: And first, you can skip all the complexity if the graph is not a multi-graph (`graph_view.is_multigraph()`), and you may be able to update the caller site to call this only once per unique v in the shorter neighbor list (assuming that there is very few edges with a huge multiplicity value).

You can run thrust::lower_bound first and just scan from there.

cpp/include/cugraph/prims/detail/nbr_intersection.cuh
```diff
@@ -386,6 +388,1177 @@ __device__ edge_t set_intersection_by_key_with_mask(InputKeyIterator0 input_key_
...
+    nbr_intersection_nbr_list(edge_partition,
+                              second_element_to_idx_map,
+                              second_element_offsets,
+                              second_element_indices,
+                              v1,
+                              v1_indices,
+                              v1_off,
+                              v1_deg);
+
+    // Scan the shorter neighbor list, binary-search the longer one.
```
seunghwak, line 739: You may use

`__ballot_sync` (set to 1 if its the first in the run) followed by `__popc` then `__fns` to find n'th unique neighbor in the short neighbor list.

There can be a trade-off. If multi-edges are infrequent, you may just let few threads in a warp. If multi-edges are common, we may allow few edges to process more than one neighbor in the short neighbor list.

Or if we assume that invoking intersection_op is expensive, we can separate finding common neighbor indices in the long and short arrays (each thread may process more than one neighbor in the short neighbor list) and calling intersection_op.

We use all threads in a warp to update the array indices. Then, for the identified indices, call intersection_op in the second round. This may lead to faster & simpler code.

PR #5556, opened 2026-06-10
Reduce memory size by restructuring copy_if use

ChuckHastings:
copy_if is the next source of the kernel usage explosion.

This PR restructures some of our copy_if usage.  Biggest changes:
1. Refactor the mask utilities to create explicit instantiations of copy_if_mask_set and copy_if_mask_unset for common scalar types
2. Refactor some obvious places in the code to use copy_if_mask_set when we can
3. Restructure the mem_frugal_partition function to use mark_entries and copy_if_mask_* on the components of our iterator

|---|---|---|---|---|
| CUDA 12 Before | 98 MB | 753 MB | 936 MB | 1786 MB |
| CUDA 12 After |  93 MB (-4.3, -4.5%) | 739 MB (-13.4 -1.77%) | 884 MB (-52.0, -5.6) | 1716 MB (-70, -3.9%) |
|---|---|---|---|---|
| CUDA 13 Before | 50 MB | 383 MB | 469 MB | 902 MB |
| CUDA 13 After |  46 MB (-3.6, -7.3%)| 375 MB (-7.1, -1.9%)| 452  MB (-17.4, -3.71%) | 873 MB (-28.1, -3.1%)|


KyleFromNVIDIA (approved):
Approved trivial CMake changes

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -8,21 +8,30 @@
...
 #include <thrust/transform_reduce.h>
 
+#include <cstdint>
+#include <type_traits>
+
 namespace CUGRAPH_EXPORT cugraph {
 
 namespace detail {
 
+#ifdef __CUDACC__
```
seunghwak: Do we need this? This file is .cuh, so only expected to be used in .cu files.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
       is_equal_t<bool>{true})));
 }
 
-template <typename MaskIterator>  // should be packed bool
-size_t count_set_bits(raft::handle_t const& handle, MaskIterator mask_first, size_t num_bits)
+#endif  // __CUDACC__
+
+size_t count_set_bits(raft::handle_t const& handle, uint32_t const* mask_first, size_t num_bits);
+
+size_t count_set_bits(rmm::cuda_stream_view stream_view,
```
seunghwak: stream_view should be the last parameter than first to be consistent.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+template <typename Iterator>
+inline constexpr bool is_thrust_zip_iterator_v =
+  is_thrust_tuple_of_arithmetic_v<copy_if_mask_iterator_value_t<std::remove_cv_t<Iterator>>>;
+
+template <typename ZipIterator>
+using zip_iterator_tuple_t = typename ZipIterator::iterator_tuple;
+
+template <typename ZipIterator>
+inline constexpr size_t zip_iterator_arity_v =
+  cuda::std::tuple_size<zip_iterator_tuple_t<ZipIterator>>::value;
```
seunghwak: ?

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+}
+
+template <typename InputIterator, typename OutputIterator>
+OutputIterator copy_if_mask_set_scalar(raft::handle_t const& handle,
+                                       InputIterator input_first,
+                                       InputIterator input_last,
+                                       uint32_t const* mask_first,
+                                       OutputIterator output_first)
+{
+  if constexpr (copy_if_mask_supported_scalar_v<InputIterator, OutputIterator>) {
```
seunghwak: This should better be `copy_if_mask_supported_iterator_v` to be consistent and also correct for fancy iterators.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+}
+
+template <typename InputIterator, typename OutputIterator>
+OutputIterator copy_if_mask_unset_scalar(raft::handle_t const& handle,
+                                         InputIterator input_first,
+                                         InputIterator input_last,
+                                         uint32_t const* mask_first,
+                                         OutputIterator output_first)
+{
+  if constexpr (copy_if_mask_supported_scalar_v<InputIterator, OutputIterator>) {
```
seunghwak: This should better be `copy_if_mask_supported_iterator_v` to be consistent and also correct for fancy iterators.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+  }
+  return output_first + (output_end - cuda::std::get<0>(output_tuple));
 }
 
 template <typename InputIterator,
-          typename MaskIterator,  // should be packed bool
-          typename OutputIterator>
+          typename OutputIterator,
+          std::enable_if_t<is_thrust_zip_iterator_v<InputIterator>, bool> = true>
+OutputIterator copy_if_mask_set(raft::handle_t const& handle,
```
seunghwak, line 334: I think we should move

`copy_if_mask_set`, `copy_if_mask_unset` and `mark_entries` outside the detail namespace.

And create `partition_mask` here and move `partition_mask` implementation functions from `shuffle_comm.cuh` to this file.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
     cuda::make_transform_iterator(thrust::make_counting_iterator(size_t{0}),
-                                  check_bit_set_t<MaskIterator, size_t>{mask_first, size_t{0}}),
+                                  check_bit_set_t<uint32_t const*, size_t>{mask_first, size_t{0}}),
     output_first,
-    is_equal_t<bool>{true});
+    is_equal_t<bool>{false});
+}
+
+template <typename InputIterator, typename OutputIterator>
+OutputIterator copy_if_mask_set_scalar(raft::handle_t const& handle,
```
seunghwak: Just a recommendation, but we are doing something similar in device_comm.cuh (supporting arithmetic types and thrust tuple of arithmetic types).

https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/utilities/device_comm.hpp#L40
https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/utilities/device_comm.hpp#L54

I think it might be better to use the same/similar mechanism/naming convention to be consistent (either update here or update the existing functions).

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +151,296 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
-                                InputIterator input_first,
-                                InputIterator input_last,
-                                MaskIterator mask_first,
-                                OutputIterator output_first)
+template <typename InputIterator, typename OutputIterator>
+OutputIterator copy_if_mask_unset_inline(raft::handle_t const& handle,
+                                         InputIterator input_first,
+                                         InputIterator input_last,
+                                         uint32_t const* mask_first,
+                                         OutputIterator output_first)
```
seunghwak: Do we need this additional level of indirection or we can do something like
```
OutputIterator copy_if_mask_unset(...) {
  if constexpr (copy_if_mask_supported_iterator_v<...>) {
    detail::copy_if_mask_unset_impl(...);
  }
  else {
    thrust::copy_if(...);
  }
}
```

cpp/include/cugraph/utilities/shuffle_comm.cuh
```diff
@@ -453,50 +539,31 @@ ValueIterator mem_frugal_partition(
...
+                 stream_view,
+                 large_buffer_type);
+
+  auto const first_size      = marked_count;
+  auto const second_size     = num_elements - first_size;
+  uint32_t const* mask_first = marked_entries.data();
+
+  if constexpr (is_thrust_zip_iterator_v<ValueIterator>) {
+    partition_mask_zip_split(
+      value_first, value_last, mask_first, first_size, second_size, stream_view, large_buffer_type);
```
seunghwak: We may better update to call cugraph::partition_mask or cugraph::copy_if_mask_set|unset rather than directly calling detail space functions.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: So I added an include of `mask_utils.cuh` to `shuffle_comm.cuh`.

It turns out that `shuffle_comm.cuh` is included by `partition_manager.hpp`.  It apparently didn't matter before, but adding `mask_utils.cuh` caused a compilation failure.  Adding these allowed me to defer chasing down why.

I tested a fix, it looks like `partition_manager.hpp` doesn't need to include `shuffle_comm.cuh`.  I've removed that include and I'll remove these from the file.

I did notice a few other examples like this.  I'll create a small PR to make everything else consistent with our design (no .cuh files include in a .hpp or .cpp file).

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Consolidate these and changed to use exec policy instead, more consistent with other approaches.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Removed this expression and did the tuple_size calls directly in place.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Renamed

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Got rid of the two inline functions and just inlined them as you suggested.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Fixed

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Moved partition_mask here, renamed to partition_by_mask to be a bit more clear.  Also made the function signatures match the copy_if_mask_set approach (single function, different SFINAE implementations).

cpp/include/cugraph/utilities/shuffle_comm.cuh
ChuckHastings: Yes, when I moved this I renamed it to partition_by_mask and put it in the cugraph namespace.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Refactored this to just be `copy_if_mask_{set, unset}`, using SFINAE to select the proper version.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +114,295 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
       is_equal_t<bool>{true})));
 }
 
+/** Value type of @p Iterator for @ref copy_if_mask_supported_iterator_v (C++17-friendly). */
+template <typename Iterator>
+using copy_if_mask_iterator_value_t =
+  std::remove_cv_t<typename thrust::iterator_traits<std::remove_cv_t<Iterator>>::value_type>;
```
seunghwak: `is_thrust_zip_iterator_v` as well. There is nothing specific to "copy_if" here. Should we better move this to https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/utilities/iterator_utils.hpp?

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +114,295 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+      input_last,
+      cuda::make_transform_iterator(
+        thrust::make_counting_iterator(size_t{0}),
+        detail::check_bit_set_t<uint32_t const*, size_t>{mask_first, size_t{0}}),
+      output_first,
+      detail::is_equal_t<bool>{false});
+  }
+}
+
+namespace detail {
```
seunghwak:
```
namespace cugraph {
namespace detail {
}
...
namespace detail {
}
}
```

Should we better merge the two detail namespace sections? This makes it easier to separate public API from implementation details.

And this separates copy_if_mask_(set|unset) functions on non-zip iterators from the functions on zip iterators. Merging the two detail sections will fix this.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +114,295 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+          typename OutputIterator,
+          std::enable_if_t<detail::is_thrust_zip_iterator_v<InputIterator>, bool> = true>
+OutputIterator copy_if_mask_set(raft::handle_t const& handle,
+                                InputIterator input_first,
+                                InputIterator input_last,
+                                uint32_t const* mask_first,
+                                OutputIterator output_first)
+{
+  static_assert(detail::is_thrust_zip_iterator_v<OutputIterator>,
+                "copy_if_mask_set zip overload requires a zip output iterator.");
```
seunghwak: We are already checking this in `enable_if`. Sounds redundant.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -142,79 +114,295 @@ __device__ size_t copy_if_mask_set(InputIterator input_first,
...
+          typename OutputIterator,
+          std::enable_if_t<detail::is_thrust_zip_iterator_v<InputIterator>, bool> = true>
+OutputIterator copy_if_mask_unset(raft::handle_t const& handle,
+                                  InputIterator input_first,
+                                  InputIterator input_last,
+                                  uint32_t const* mask_first,
+                                  OutputIterator output_first)
+{
+  static_assert(detail::is_thrust_zip_iterator_v<OutputIterator>,
+                "copy_if_mask_unset zip overload requires a zip output iterator.");
```
seunghwak, line 355: Similar here, looks redundant.

cpp/include/cugraph/utilities/mask_utils.cuh
```diff
@@ -249,18 +438,118 @@ rmm::device_uvector<T> keep_marked_entries(
...
+  copy_if_mask_unset(
+    handle, input_first, input_last, mask_first, get_dataframe_buffer_begin(tmp_buffer));
+  copy_if_mask_set(handle, input_first, input_last, mask_first, input_first);
+  thrust::copy(rmm::exec_policy(stream_view),
+               get_dataframe_buffer_cbegin(tmp_buffer),
+               get_dataframe_buffer_cend(tmp_buffer),
+               input_first + first_size);
+}
+
+namespace detail {
```
seunghwak: Here as well, better place detail space functions with other detail space functions.

cpp/src/utilities/mask_utils.cu
```diff
@@ -0,0 +1,101 @@
...
+      if ((i + 1) * packed_bools_per_word() > num_bits) {
+        word &= packed_bool_partial_mask(num_bits % packed_bools_per_word());
+      }
+      return static_cast<size_t>(__popc(word));
+    }),
+    size_t{0},
+    cuda::std::plus<size_t>{});
+}
+
+namespace detail {
```
seunghwak, line 45: Better place detail namespace before any public space function code.

seunghwak (approved):
LGTM besides few minor cosmetic issues.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: I made a more generic name (`iterator_value_t`) , used it everywhere in this file and moved it to `iterator_utils.hpp`.  Also moved is_thrust_zip_iterator_v to the same file.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Restructured.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Restructured.

cpp/include/cugraph/utilities/mask_utils.cuh
ChuckHastings: Technically the enable_if is on the input iterator, and this assert is checking the output iterator.   Presumably this would result in a more helpful compile error if the output iterator doesn't match the expected structure.

cpp/src/utilities/mask_utils.cu
ChuckHastings: I waffled on keeping all things together by function or grouping all detail functions first, then all non-detail functions.  I constructed this order to allow the zip implementations to call the scalar implementations that are in the cugraph namespace.  I have restructured this so that the cugraph namespace variation simply calls the proper detail method instead of this which is a bit convoluted.

PR #5559, opened 2026-06-11
Create unique, sequence, and fill wrappers in cugraph to reduce binary size

ChuckHastings:
Next effort in binary size reduction.

This PR:
* Renames the sort_wrapper back to sort using the workaround we tried for scan
* Creates wrappers for thrust::unique, thrust::sequence, thrust::fill

|---|---|---|---|---|
| CUDA 12 Before | 98 MB | 752 MB |  935 MB | 1785 MB |
| CUDA 12 After |  99 MB (+1, +1.2%) | 736 MB (-16, -2.1%) | 919 MB (-16, -1.8%) | 1754 MB (-31, -1.8%) |
|---|---|---|---|---|
| CUDA 13 Before | 50 MB | 382 MB | 469 MB | 901 MB |
| CUDA 13 After |  50 MB (+0.5, +0.96%)| 374 MB (-8, -2.1%)| 461  MB (-8, -1.8%) | 885 MB (-16, -1.8%)|

### DIFF
--- a/cpp/include/cugraph/utilities/thrust_wrappers.hpp
+++ b/cpp/include/cugraph/utilities/thrust_wrappers.hpp
@@ -25,6 +25,24 @@
 namespace CUGRAPH_EXPORT cugraph {
 namespace detail {
 
+/** @brief True when @p ExecutionPolicy is @c rmm::exec_policy.
+ */
+template <typename ExecutionPolicy>
+inline constexpr bool is_rmm_exec_policy_v =
+  std::is_same_v<std::remove_cv_t<std::remove_reference_t<ExecutionPolicy>>, rmm::exec_policy>;
+
+/** @brief True when @p ExecutionPolicy is @c rmm::exec_policy_nosync (e.g. @c
+ * handle.get_thrust_policy()).
+ */
+template <typename ExecutionPolicy>
+inline constexpr bool is_rmm_exec_policy_nosync_v =
+  std::is_same_v<std::remove_cv_t<std::remove_reference_t<ExecutionPolicy>>,
+                 rmm::exec_policy_nosync>;
+
+}  // namespace detail
+
+namespace detail {
+
 /** @brief True for scalar value types dispatched to @ref sort_impl / @ref unique_impl. */
 template <typename T>
 inline constexpr bool sort_supported_arithmetic_scalar_v =
@@ -41,8 +59,7 @@ inline constexpr bool sort_supported_arithmetic_scalar_v =
  *
  *  Keep this in lockstep with those explicit instantiations: add a disjunct only when you add the
  *  matching explicit instantiation (otherwise a call through @ref cugraph::sort can fail at
- *  link
- *  time).
+ *  link time).
  */
 template <typename T>
 inline constexpr bool sort_supported_zip_v =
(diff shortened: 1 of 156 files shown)


cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -356,4 +465,106 @@ struct exclusive_scan_t {
...
+ */
+struct fill_t {
+  template <typename ExecutionPolicy, typename ForwardIterator, typename T>
+  void operator()(ExecutionPolicy const& policy,
+                  ForwardIterator first,
+                  ForwardIterator last,
+                  T const& value) const
+  {
+    using value_t = detail::fill_iterator_value_t<ForwardIterator>;
+    if constexpr (detail::fill_supported_iterator_v<ForwardIterator>) {
```
seunghwak: Don't we need to check for `exec_policy` (i.e. detail::is_rmm_exec_policy_v<ExecutionPolicy>)?

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -155,55 +226,93 @@ using sort_iterator_value_t =
...
+ * scalar and zip-iterator ranges.
+ */
+struct unique_t {
+  template <typename ExecutionPolicy, typename RandomAccessIterator>
+  RandomAccessIterator operator()(ExecutionPolicy const& policy,
+                                  RandomAccessIterator first,
+                                  RandomAccessIterator last) const
+  {
+    using value_t = detail::sort_iterator_value_t<RandomAccessIterator>;
+    if constexpr (detail::sort_supported_arithmetic_scalar_v<value_t> ||
```
seunghwak: Why are we using sort_supported_arithmetic_scalar_v here but checking for iterator in other places? (e.g. fill_supported_iterator_v). And no need to check for exec_policy?

Will this work if RandomAccessIterator is not a pointer?

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: We currently instantiate both of the exec policies that we currently use (rmm::exec_policy, rmm::exec_policy_nosync), so everything currently compiles and executes properly.  If we were to try and use a different exec policy, the compilation would fail - so at least it wouldn't be a runtime error.

We could add a safety check in case we want things to keep compiling in those cases (if we somehow add a different exec policy).  But if we do that I suspect we should do it for all of the functions in this file.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Yes, this was an oversight in the original implementation.  I will fix this.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -32,12 +35,12 @@ inline constexpr bool sort_supported_arithmetic_scalar_v =
...
- *  @ref sort_supported_arithmetic_scalar_v and @ref sort_impl for @c rmm::exec_policy and
+ *  thrust_wrappers_zip_types.hpp and instantiated in thrust_wrappers.cu. Scalar sorts use
+ *  @ref sort_supported_scalar_iterator_v and @ref sort_impl for @c rmm::exec_policy and
  *  @c rmm::exec_policy_nosync.
  *
  *  Keep this in lockstep with those explicit instantiations: add a disjunct only when you add the
- *  matching explicit instantiation (otherwise a call through @ref cugraph::sort_wrapper can fail at
+ *  matching explicit instantiation (otherwise a call through @ref cugraph::sort can fail at
  *  link
  *  time).
```
seunghwak: Something very minor but "link" and "time" can go to the same line. No need to break line here.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -155,63 +274,102 @@ using sort_iterator_value_t =
...
+ */
+struct unique_t {
+  template <typename ExecutionPolicy, typename RandomAccessIterator>
+  RandomAccessIterator operator()(ExecutionPolicy const& policy,
+                                  RandomAccessIterator first,
+                                  RandomAccessIterator last) const
+  {
+    if constexpr ((detail::is_rmm_exec_policy_v<ExecutionPolicy> ||
+                   detail::is_rmm_exec_policy_nosync_v<ExecutionPolicy>) &&
+                  detail::sort_supported_iterator_v<RandomAccessIterator>) {
```
seunghwak: Isn't this a misnomer?

If sort & unique support same types of iterators, should we better define
```
using iterator_supported_iterator_v = sort_supported_iterator_v;
```
?

cpp/include/cugraph/utilities/thrust_wrappers.hpp
```diff
@@ -22,7 +25,7 @@
 namespace CUGRAPH_EXPORT cugraph {
 namespace detail {
 
-/** @brief Whether iterator type @p T has an out-of-line @ref sort_impl lexicographic sort. */
+/** @brief True for scalar value types dispatched to @ref sort_impl / @ref unique_impl. */
```
seunghwak, line 46: Something minor but what's the rule in placing these type trait templates?

We have

sort related templates
scan related templates
rmm_exec_policy related templates
more scan related templates
scan_impl
sort_impl
unique_impl
more sort related templates
fill related templates
sequence related templates
(and they alternate few times)
fill_impl
sequence_impl
...

This looks inconsistent and makes it a bit difficult to navigate.

We may do something like

all type trait templates
all implementations

or

sort related templates
sort impl
unique related templates
unique impl
scan related templates
scan impl
...

and so on.

seunghwak (approved):
LGTM besides few minor cosmetic issues

jnke2016 (approved):

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Added additional using.

cpp/include/cugraph/utilities/thrust_wrappers.hpp
ChuckHastings: Reordered

PR #5560, opened 2026-06-12
[DO NOT MERGE] Test RAFT 3052

divyegala:


ci/test_cpp.sh
```diff
@@ -13,6 +13,7 @@ rapids-logger "Configuring conda strict channel priority"
 conda config --set channel_priority strict
 
 CPP_CHANNEL=$(rapids-download-from-github "$(rapids-artifact-name conda_cpp libcugraph cugraph --cuda "$RAPIDS_CUDA_VERSION")")
+source ./ci/use_conda_packages_from_prs.sh
```
gforsyth, line 16: The `RAPIDS_PREPRENDED_CONDA_CHANNELS` variable only impacts builds (because we always `source rapids-rattler-channel-string`) -- on the testing side, you'll want to additional add something like the following to the `rapids-dependency-file-generator` call below:

```sh
rapids-dependency-file-generator \
  --output conda \
  --file-key test_cpp \
  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch)" \
  --prepend-channel "${LIBRAFT_CHANNEL}" \ 
  --prepend-channel "${RAFT_CHANNEL}" \ 
  --prepend-channel "${LIBCUVS_CHANNEL}" \ 
  --prepend-channel "${CPP_CHANNEL}" \
  | tee env.yaml
```

PR #5564, opened 2026-06-24
Fix cuh in hpp files

ChuckHastings:
PR #5556 identified some issues where we are including .cuh files inside of .hpp files.

Our convention is to avoid that, so that we don't need to litter our headers with ifdef's identifying parts that should be ignored if we're compiling with gcc.

This PR addresses those examples that weren't already part of #5556.  For these cases, they are all header files that define implementations that are instantiated in .cu files, so renaming them to .cuh is sufficient.


seunghwak (approved):
LGTM

PR #5570, opened 2026-07-02
switch from 'cuda-python' to specific components (e.g. 'cuda-bindings')

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/285

`cuda-python` is a metapackage and sometimes pulls in components that RAPIDS libraries don't need. This PR is part of a series dropping direct use of `cuda-python` in favor of depending only on those specific components.

* drops dependency on `cuda-python`
* adds explicit dependencies on components:
  - `cuda-bindings >=12.9.2,<13.0` (CUDA 12), `cuda-bindings >=13.0.1,<14.0` (CUDA 13)
* adds explicit testing of oldest `cuda-bindings` in CI

### Why these specific bounds?

* these roughly match what we were getting with `cuda-python`
* worked through finding compatible ranges in https://github.com/rapidsai/rmm/pull/2460#discussion_r3514148466
* `cuda-bindings` major version follows the CTK major version, so kept a ceiling just like we currently have for `cuda-python`

### DIFF
--- a/conda/environments/all_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-129_arch-aarch64.yaml
@@ -10,11 +10,11 @@ dependencies:
 - c-compiler
 - certifi
 - cmake>=4.0
+- cuda-bindings==12.*,>=12.9.2
 - cuda-nvcc
 - cuda-nvrtc-dev
 - cuda-nvtx-dev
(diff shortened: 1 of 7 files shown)


jakirkham (approved):
Thanks James! 🙏

PR #5572, opened 2026-07-06
Update RAPIDS.cmake to log source of rapids-cmake

arhag23:
Issue: https://github.com/rapidsai/build-planning/issues/225


KyleFromNVIDIA (approved):

gforsyth (approved):
Test failures look flakey, and in any case, they wouldn't be triggered by additional status messages from cmake.  Admin merging

PR #5574, opened 2026-07-07
Reduce binary size random select primitive

ChuckHastings:
The random select primitive is our most complex primitive, and many portions of it do not vary based on the graph, the bias operator or the edge operator.  This PR splits some of these common pieces out into separate helper functions that can be compiled once for each variation that occurs rather than being compiled redundantly with the combinatorial explosion of template parameters of the primitive.

On CUDA 12 (the CUDA version driving our work), here is the savings:
|---|---|---|---|---|
| CUDA 12 Before | 89.2 MB | 702.6 MB | 822.3 MB | 1614.1 MB |
| CUDA 12 After |  174.9 MB (+85.6, +95.9%) | 584.6 MB (-117.9 -16.8%) | 698.9 MB (-123.4, -15.0%) | 1458.3 MB (-155.7, -9.65%) |


cpp/include/cugraph/prims/detail/heterogeneous_biased_sample.cuh
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * @brief Algorithm A-Res biased sampling index selection for heterogeneous graphs.
+ *
+ * Template dependence is limited to @p edge_t, @p edge_type_t, and @p bias_t.
+ */
+template <typename edge_t, typename edge_type_t, typename bias_t>
+void compute_heterogeneous_biased_sampling_index_without_replacement(
+  raft::handle_t const& handle,
+  std::optional<raft::device_span<size_t const>>
+    input_frontier_indices,  // input_per_tyep_degree_offsets & input_biases are already packed if
```
seunghwak, line 32: tyep=>type

cpp/src/prims/detail/compact_nonzero_aggregate_local_frontier_biases_impl.cuh
```diff
@@ -0,0 +1,131 @@
...
+      aggregate_local_frontier_biases.begin(),
+      aggregate_local_frontier_biases.end(),
+      check_out_of_range_t<bias_t>{bias_t{0.0}, std::numeric_limits<bias_t>::max()});
+    if (multi_gpu) {
+      num_invalid_biases = host_scalar_allreduce(
+        handle.get_comms(), num_invalid_biases, raft::comms::op_t::SUM, handle.get_stream());
+    }
+    CUGRAPH_EXPECTS(num_invalid_biases == 0,
+                    "invalid_input_argument: bias_e_op return values should be non-negative and "
+                    "should not exceed std::numeirc_limits<bias_t>::max().");
```
seunghwak, line 55: `numeirc`=>`numeric`

bdice (approved):
Approving CMake.

seunghwak:
I am still thinking about the best approach, but I am thinking more in the line of

splitting `per_v_random_select_transform_outgoing_e` to `per_v_random_select_e` and `transform_gather_e` (this already exists).

I guess the edge sampling part varies a lot for different sampling methods while the follow up transform and gather part is largely same (IIRC, this is the main motivation for creating `transform_gather_e`. Then, we can perform transform & gather in a single wrapper function and avoid instantiating this part multiple times while keeping the primitives header only and flexible for any template type combinations.

Let me think more about this.

cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh
seunghwak, line 231: Yeah... and this is an easy target, not sure how big will be the gain, but we have

`count_invalid_vertex_pairs`
https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/utilities/error_check_utils.cuh#L59

We may create
`count_invalid_vertices` as well.

And explicitly instantiate these two functions, and use them everywhere we check for invalid vertices or invalid vertex pairs.

cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh
```diff
@@ -322,19 +269,21 @@ per_v_random_select_transform_e(raft::handle_t const& handle,
...
+      aggregate_local_key_list =
+        allocate_dataframe_buffer<key_t>(local_key_list_offsets.back(), handle.get_stream());
+      device_allgatherv(
+        minor_comm,
+        key_list.begin(),
+        get_dataframe_buffer_begin(*aggregate_local_key_list),
+        raft::host_span<size_t const>(local_key_list_sizes.data(), local_key_list_sizes.size()),
+        raft::host_span<size_t const>(local_key_list_offsets.data(),
+                                      local_key_list_offsets.size() - 1),
+        handle.get_stream());
```
seunghwak, line 285: And another thing to experiment is that how much we can save by explicitly instantiating device_comm functions (device_allgatherv, device_bcast, and so on).

Maybe not worth the effort as most heavy lifting happens inside the NCCL functions, so the saving might be negligible but we might be able to quickly experiment (e.g. just delete few device comm calls and see whether we see any meaningful change in the resulting binary size).

seunghwak: If this turns out to be non-negligible, we can do something similar to what we are doing for thrust algorithms.

cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh
seunghwak, line 407: Need to double check but `sample_key_indices.has_value()` is true only when minor_comm_size > 1.

If we accept some overhead when `GraphViewType::is_multi_gpu` is true and `minor_comm_size == 1`, we may create `std::conditional_t<GraphViewType::is_multi_gpu, rmm::device_uvector<size_t>, std::byte /* dummy */> sample_key_indices` and here we can use `if constexpr (GraphViewType::is_multi_gpu)` instead of `if`. So, only one of the two transform calls will be instantiated. Try to delete the if pass or the else pass and see how much this reduces the binary size. If the difference is non-negligible, this might be another place to cut binary size.

seunghwak (commented):
I am still reviewing this and thinking more about the best approach but few quick early comments.

cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh
ChuckHastings: I didn't look extensively through the code, but I did add some constexpr blocks in a few places where we were just checking minor_comm_size so that in SG code we wouldn't even compile the block.  I'm sure there are other examples like this, I can take a look.

seunghwak:
A data point to consider in this work.

`per_v_random_select_transform_outgoing_e` primitives are instantiated in these files.

Via random_walks_impl.cuh
* cpp/src/sampling/random_walks_sg_v32_e32.cu
* cpp/src/sampling/random_walks_mg_v32_e32.cu
* cpp/src/sampling/random_walks_sg_v64_e64.cu
* cpp/src/sampling/random_walks_mg_v64_e64.cu
These instantiate uniform/biased/node2vec random walks; those call outgoing primitive mainly in K=1 homogeneous forms (biased and unbiased variants).

Via sample_edges.cuh
* cpp/src/sampling/detail/sample_edges_sg_v32_e32.cu
* cpp/src/sampling/detail/sample_edges_mg_v32_e32.cu
* cpp/src/sampling/detail/sample_edges_sg_v64_e64.cu
* cpp/src/sampling/detail/sample_edges_mg_v64_e64.cu
* cpp/src/sampling/detail/temporal_sample_edges_sg_v32_e32.cu
* cpp/src/sampling/detail/temporal_sample_edges_mg_v32_e32.cu
* cpp/src/sampling/detail/temporal_sample_edges_sg_v64_e64.cu
* cpp/src/sampling/detail/temporal_sample_edges_mg_v64_e64.cu
These are where you get the broadest variation coverage (biased/unbiased, homogeneous/heterogeneous, K scalar vs Ks span, with/without edge types, temporal bias ops).

Merged these files to four, SG/v32/e32, SG/v64/e64, MG/v32/e32, MG/v64/e64.

After fixing a minor build error (`/home/seunghwak/RAPIDS/cugraph/cpp/src/sampling/detail/sample_edges.cuh(65): error: class template "cugraph::detail::sample_edges_op_t" has already been defined (previous definition at line 50 of /home/seunghwak/RAPIDS/cugraph/cpp/src/sampling/random_walks_impl.cuh)`, just renumbered `sample_edges_op_t` in `random_walks_impl.cuh` to `rw_sample_edges_op_t` as a quick and dirty fix), I got roughly 5% cut in the binary size.

-rwxr-xr-x 1 seunghwak domain-users  21814664 Jul  8 11:16 libcugraph_common.so
-rwxr-xr-x 1 seunghwak domain-users   8396160 Jul  8 11:20 libcugraph_c.so
-rwxr-xr-x 1 seunghwak domain-users 208718048 Jul  8 11:20 libcugraph_mg.so
-rwxr-xr-x 1 seunghwak domain-users   7210536 Jul  8 11:18 libcugraph_mtmg.so
-rwxr-xr-x 1 seunghwak domain-users 175323672 Jul  8 11:17 libcugraph.so
to
-rwxr-xr-x 1 seunghwak domain-users  21814664 Jul  8 11:16 libcugraph_common.so
-rwxr-xr-x 1 seunghwak domain-users   8396160 Jul  8 12:08 libcugraph_c.so
-rwxr-xr-x 1 seunghwak domain-users 198574200 Jul  8 12:08 libcugraph_mg.so (4.86% cut)
-rwxr-xr-x 1 seunghwak domain-users   7210536 Jul  8 11:18 libcugraph_mtmg.so
-rwxr-xr-x 1 seunghwak domain-users 166253072 Jul  8 12:08 libcugraph.so (5.17% cut)

We may re-structure the primitive sub-functions little more to further facilitate a function re-use within a TU (e.g. create functions with only edge_t template type). I think we might be able to get similar 10% cut in this way without sacrificing primitives' capability to support arbitrary type combinations.

seunghwak:
I thought more about splitting `per_v_random_select_transform_outgoing_e` to `per_v_random_select_outgoing_e` and the follow-up transform primitive but it seems like the gain is not sufficient enough to justify some API complications.

The main motivation for this was

1. Recover build parallelism (max 2x) by building multiple `per_v_random_select_transform_outgoing_e` calls in a single TU.
2. Possibly remove some redundancy in `per_v_random_select_transform_outgoing_e` and `transform_gather_e`.

But it seems like the build overhead of the follow-up transform part is not very significant and removing redundancy requires additional plumbing code largely offsetting the code size reduction while also introducing some performance overhead.

More details here.

`detail::per_v_random_select_transform_e` is the main function here.
https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh#L213

This works in the following steps.

1. Random sample edges
2. Shuffle the edges to the edge owning GPU
3. Transform
4. Shuffle back the transformed output to the key owning GPU.

1 & 2 can go to `per_v_random_select_outgoing_e`
3 can go to `transform_gather_e`
and 4 could be a separate function possibly outside the primitive.

In step 3,

`if () { thrust::transform(...); } else { thrust::transform(...); }` is the main part.
https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/prims/per_v_random_select_transform_outgoing_e.cuh#L458

They will be compiled only once if template parameters are same (mainly `EdgeOp` and `T`). This won't have too much impact on build time (I guess different sampling methods have more impact on the edge selection part rather than `EdgeOp` or `T` in many cases).

In step 4,

we are mainly using cugraph thrust wrappers (except for one thrust::transform and one thrust::tabulate) or `shuffle_values` (I guess most heavy lifting happens in NCCL) so this will be cheap to build. And we can place this under `if constexpr (GraphViewType::is_multi_gpu)` so we don't need to build this for SG.

Steps 3 & 4 look cheap to build, so splitting may not be very helpful in build time by increasing parallelism (this is most effective when steps 1&2 and 3&4 took comparable amount of time, but it seems like steps 1&2 will be much more expensive than 3&4).

Complication part.

We are not really gathering edges here but gathering key frontier (major vertex can be tagged) and their sampled outgoing edges. The key distinction here is that the edge major can be tagged. This is not a simple gathering of edges. This can cause several complications. Another source of complication is that `transform_gather_e` uses (src, dst, multi_edge_index) to index an edge but here, we use local edge index. The latter can be more efficient but it is more of an implementation detail and properly exposing this in the public API can be a bit tricky.

So, I am leaning towards not splitting the primitive for now and focusing more on building in a single TU, use if constexpr if it helps cutting binary size/build time, refactor some sub-routines if it can prevent multiple explicit instantiations within a single TU.

ChuckHastings:
Superseded by #5583

PR #5576, opened 2026-07-08
Add dlpack conversion to capi

### LINKED ISSUE #5567: Add support for DLPack's DLDataType to the C API
pylibcugraph should support passing pinned host memory tensors, especially PyTorch pinned tensors, without forcing an unnecessary copy to device memory.  Today, PLC generally expects CuPy-/NumPy-style array objects and uses CuPy/NumPy mechanisms to infer dtype. That works for device arrays and some existing paths, but PyTorch host tensors do not expose __array_interface__, and converting through cupy.asarray forces a device copy for pinned host memory. DLPack appears to be the most generic path because PyTorch, CuPy, cuDF, and other array libraries support it.

The goal here is to add a function to the C API (and PLC) to convert a DLDataType structure into the data type expected by the C API.

ChuckHastings:
Closes #5567 

dlpack.h defines the structure for DLPack data types.  This function will convert the structure to the dtype we use in the C API.

I kept the changes minimal.

### DIFF
--- a/cpp/cmake/thirdparty/get_dlpack.cmake
+++ b/cpp/cmake/thirdparty/get_dlpack.cmake
@@ -1,6 +1,6 @@
 #=============================================================================
 # cmake-format: off
-# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 # cmake-format: on
 #=============================================================================
--- a/cpp/cmake/thirdparty/get_dlpack.cmake
+++ b/cpp/cmake/thirdparty/get_dlpack.cmake
@@ -5,38 +5,26 @@
 # cmake-format: on
 #=============================================================================
 
-# dlpack is a header-only library used by the cugraph_c DLPack interop API.
-# In conda environments it is provided as an installed CMake package; in other
-# environments (e.g. pip / devcontainer builds) it may not be installed, so we
-# fall back to downloading the headers via CPM.
+# dlpack is a header-only library used by the cugraph_c DLPack interop API
+# (cugraph_c/dlpack_interop.h). In conda environments it is provided as an
+# installed package; in other environments (e.g. pip / devcontainer / wheel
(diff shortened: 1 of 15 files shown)


dependencies.yaml
```diff
@@ -378,6 +378,7 @@ dependencies:
           - c-compiler
           - cuda-nvcc
           - cxx-compiler
+          - dlpack
```
alexbarghi-nv, line 381: Shouldn't we be pulling `dlpack` from `cudf` or `cupy`?  Do we need to specify it explicitly?

alexbarghi-nv (changes requested):
This looks good for the most part; I just need to verify that I can actually pass the capsule.  I'm going to try that tonight.

alexbarghi-nv (commented):
@ChuckHastings I just finished confirming that this does work as expected - just wanted to verify that we're handling the `dlpack` dependency correctly.  I think I may do something a bit radical in my follow-up PR and try to use DLPack everywhere, while still enforcing that host arrays are in host memory and that device arrays are in either device or pinned memory.

alexbarghi-nv (approved):

dependencies.yaml
alexbarghi-nv: Sorry, I just realized this was because you need the headers at compile time. I resolved the comment.

cpp/cmake/thirdparty/get_dlpack.cmake
```diff
@@ -0,0 +1,34 @@
...
+#=============================================================================
+
+# dlpack is a header-only library used by the cugraph_c DLPack interop API
+# (cugraph_c/dlpack_interop.h). In conda environments it is provided as an
+# installed package; in other environments (e.g. pip / devcontainer / wheel
+# builds) it may not be installed, so we fall back to downloading the headers.
+function(find_and_configure_dlpack VERSION)
+
+    include(${rapids-cmake-dir}/find/generate_module.cmake)
+    rapids_find_generate_module(DLPACK HEADER_NAMES dlpack.h)
```
KyleFromNVIDIA: Should this be the same case as `rapids_cpm_find()`? Like:


```suggestion
    rapids_find_generate_module(dlpack HEADER_NAMES dlpack.h)
```

KyleFromNVIDIA (approved):

seunghwak (approved):
LGTM

PR #5580, opened 2026-07-09
dnm: test wheel ci on cuda base

gforsyth:
xref rapidsai/build-planning#143
xref rapidsai/ci-imgs#400
xref rapidsai/shared-workflows#521

Test run using the smaller `cuda-base` image as the parent image of `citestwheel`

This shouldn't be merged. If all projects can run wheel tests with this image, we'll swap out the image upstream in `shared-workflows`.


gforsyth:
Done testing, thanks for riding along

PR #5583, opened 2026-07-10
Combine sampling primitive calls into fewer TUs to decrease size

ChuckHastings:
Supercedes #5574.

We get a pretty good reduction in binary size by combining the TUs that call the sampling primitives.  We'll start there, since this preserves the header-only design aspect of the primitives.

|---|---|---|---|---|
| CUDA 12 Before | 89.5 MB | 702.7 MB | 822.4 MB | 1614.5 MB |
| CUDA 12 After |  253.5 MB (+164.1, +183.4%) | 561.5 MB (-141.2 -20.1%) | 673.2 MB (-149.1, -18.1%) | 1488.3 MB (-126.2, -7.8%) |


cpp/.clang-format
```diff
@@ -56,7 +56,7 @@ BreakConstructorInitializers: BeforeColon
 BreakInheritanceList: BeforeColon
 BreakStringLiterals: true
 ColumnLimit: 100
-CommentPragmas: '^ IWYU pragma:'
+CommentPragmas: '^ ?[*]? ?(IWYU pragma:|SPDX-)'
```
seunghwak: What is this? Did you pull updates from another unmerged PR?

cpp/src/sampling/detail/sample_edges_one_property.hpp
seunghwak, line 1: Should this be .cuh? Or can be .hpp?

seunghwak (approved):
LGTM besides the minor issues below.

cpp/.clang-format
bdice: We probably want the version in https://github.com/rapidsai/cugraph/pull/5582. Please see https://github.com/rapidsai/build-planning/issues/303 and that PR's description for details.

ChuckHastings: A fix for an update to our copyright rules.  The copyrights are now too long for the format string.  This allows them to be.  I'm going to tweak it to be standardized across RAPIDS (there's a slack thread discussing the exact string)

ChuckHastings: I merged main which uses the new version @bdice

cpp/src/sampling/detail/sample_edges_one_property.hpp
ChuckHastings: Yes, that should be .hpp.  I'll rename it.

Issue #4191, opened 2024-02-23
[FEA]: Topic: SG Egograph (Batched vs. Single Vertex)


nv-rliu:
### Is this a new feature, an improvement, or a change to existing functionality?

Change

### How would you describe the priority of this feature request

High

### Please provide a clear description of problem this feature solves

Upon closer examination of [Egonet.py](https://github.com/rapidsai/cugraph/blob/branch-24.04/python/cugraph/cugraph/community/egonet.py), it has been discovered that there are a few inconsistencies with the behavior of `ego_graph` and `batched_ego_graph` based on what the documentation says. 

According to the docs, the `ego_graph` function is supposed to compute the induced sub-graph of neighbors centered at a single node, `n`, and return the sub-graph. The `batched_ego_graph` function (in congruence with its design in the legacy API) is supposed to accept a List of `n` values, or `seeds`, and compute each induced sub-graph in "batch" mode (parallel). 

However, currently both Python functions just call the same PLC algorithm, which raises two concerns:

1. The PLC function doesn't actually run ego_graph in "batch" mode like it's designed to
2. The single `ego_graph` function can be called with multiple `n` values and returns a graph that doesn't specify the cut-offs (offsets). 

### Describe your ideal solution

The proposed solution for the Python API is to combined both functions into a single `ego_graph` function, which will recognize when it is given a single `n` or multiple `seeds` to compute. 

Down the line, another parameter can be added which allows for proper "batch" mode, similar to how cugraph handles batched betweenness centrality. 
```
def ego_graph(
    G,
    n,.
    radius=1,
    center=True,
    undirected=None,
    distance=None,
    batch_mode=False
):

New Parameters
    ----------
    n : integer or List, cudf.Series, cudf.DataFrame
    < update docstring to specify that it can compute a single n or multiple seeds >

    batch_mode : Boolean
    Can be set to True or False

Returns
    ----------
A Graph or Edge Lists with Offsets
```

And the current `batched_betweenness_centrality` algorithm can be given a `DeprecationWarning`.

### Describe any alternatives you have considered

The two functions could also remain the same, but then `ego_graph` would need to not allow multiple `seeds` in order to be differentiated from `batched_ego_graph`, aka, multiple ego_graph calls. 

### Additional context

This was discovered while looking into [this bug](https://github.com/rapidsai/cugraph/issues/4190) on the MG implementation of `ego_graph`

Further context may be included in the discussion down below. 

### Code of Conduct

[...]

PR #5584, opened 2026-07-11
Fix multi-seed ego_graph offset handling

sauravsingla:
## Summary

This draft addresses #4191.

The lower-level `pylibcugraph.ego_graph` API already returns subgraph offsets for
multiple seed vertices, but the high-level Python wrapper currently discards
them.

This proposal:

- preserves existing single-seed behavior;
- prevents ambiguous composite output for multiple seeds;
- exposes offsets through `return_offsets=True`;
- adds regression tests covering multi-seed behavior;
- adds weighted graph and renumbered graph coverage.

## Testing

Completed without GPU:

- Python syntax validation
- py_compile
- mocked wrapper control-flow tests
- input validation tests

GPU validation is still required for:

- cuDF integration
- pylibcugraph execution
- weighted graph behavior
- renumbered graph behavior
- SG/MG CI

This is opened as a draft to discuss the preferred public API before finalizing the implementation.


sauravsingla:
This PR was closed accidentally while I was trying to convert it to draft. The branch and commits are still available.

Could a maintainer please reopen this PR, or advise whether I should open a new draft PR from the same branch?

The change addresses #4191 and is intended for API discussion and GPU CI validation before being marked ready for review.

sauravsingla:
Hi @alexbarghi-nv, just checking whether the CI run for commit 62e035b needs to be retriggered. Please let me know if anything is needed from my side.

ChuckHastings:
Always excited for external contributions, thanks!

There are format issues with your PR that should be fixed, you should be able to follow the links to see them.  I think if you enable pre-commit hooks you should be able to let the tools do much of the work for you.

sauravsingla:
Thank you for the guidance. I have applied the repository Python formatting fixes in the latest commit `b2454a1`. Please let me know if any further changes are needed.

sauravsingla:
Hi, thanks for the review. I addressed the requested changes and updated the PR. The style checks and build jobs are now passing. I’m currently investigating the remaining Python 3.12 CI failure (conda-python-tests / 13.0.3, 3.12, amd64, ubuntu24.04) and will update the PR once I identify the root cause. Please let me know if you notice anything specific that I should look at.

sauravsingla:
I reviewed the Python 3.12 CI logs. The remaining failure is in tests/utils/test_dataset.py::test_benchmarking_datasets[dataset4] and is caused by urllib.error.URLError: No address associated with hostname. This appears to be an external dataset download/DNS issue and does not seem related to the changes in this PR. All other tests in the job passed.

sauravsingla:
Hi @rlratzel, just checking in on the review when convenient. Please let me know if any additional information or changes are needed from my side.

rlratzel:
Thanks for your contribution @sauravsingla .  I think the PR changes look technically fine, but I'm wondering if having a completely different return type is expected when `return_offsets=True` - I think a user of this API would be surprised if a `Graph` object is not returned.

That said, I'm not opposed to this change, I just want to make sure these aspects have been discussed.  Also, would simply using `pylibcugraph` directly for your use case be an option instead?

sauravsingla:
Thanks for the thoughtful review!

For my own use case, calling pylibcugraph directly is certainly an option. My motivation for this PR was broader: to make the same capability available through the high-level cugraph API so users don’t need to drop down to the lower-level interface to work with multi-seed ego graphs.

Regarding the return type, I agree that returning something other than a Graph when return_offsets=True could be surprising. I chose (edge_dataframe, offsets) because the offsets returned by pylibcugraph directly delimit contiguous ranges in the returned edge output, so that representation preserves the meaning of the offsets naturally.

My intention wasn’t to introduce a different API pattern, but to expose the existing functionality while preserving the per-seed boundaries. If maintaining a Graph return type is the preferred direction for the public API, I’m happy to revise the implementation to align with the project’s conventions. I’d appreciate your guidance on the API shape you think would best fit cuGraph.

sauravsingla:
Hi @rlratzel, just following up on the API direction when convenient. I’m ready to update the implementation once the preferred return shape is confirmed.

rlratzel:
Thank you again @sauravsingla for this PR!  I have some thoughts below:

> For my own use case, calling pylibcugraph directly is certainly an option. My motivation for this PR was broader: to make the same capability available through the high-level cugraph API so users don’t need to drop down to the lower-level interface to work with multi-seed ego graphs.

I think that's the right goal, but IMO the new return type still requires users to effectively drop down to a lower-level interface anyway (edge dataframes and lists vs. higher-level cuGraph Graph objects).

> My intention wasn’t to introduce a different API pattern, but to expose the existing functionality while preserving the per-seed boundaries. If maintaining a Graph return type is the preferred direction for the public API, I’m happy to revise the implementation to align with the project’s conventions. I’d appreciate your guidance on the API shape you think would best fit cuGraph.

I don't have a strong opinion, but I would think users expect to either get a single Graph object returned (the case of `n` being an int) or a list of Graph objects (the case of `n` being a list). The list could either be ordered based on the ordering of `n`, or instead it could be a dict with values from `n` being the keys and the corresponding return Graphs as the values.  IMO that's the most consistent/expected behavior as the higher-level interface cuGraph provides.

I realize that could have a (minor?) performance impact though if there are many Graph objs to create, but I would think if users need the highest performance possible in Python that they'd be using pylibcugraph directly anyway?  As a user of this API, what do you think?

sauravsingla:
Thank you for the detailed feedback. I agree that the high-level cugraph API should continue to return Graph objects rather than expose the lower-level edge dataframe and offsets representation.

From a user perspective, I think the clearest behaviour would be:

* when n is a single vertex, return one Graph, preserving the existing single-seed behaviour;
* when n contains multiple vertices, return a list[Graph] in the same order as the supplied seeds.

I would favour a list over a dictionary because it preserves the exact input ordering, supports duplicate seeds, and works consistently for multi-column vertex identifiers without requiring a separate dictionary-key representation.

The offsets returned by pylibcugraph can remain an internal implementation detail: the combined edge output can be partitioned using those offsets, and each partition can then be used to construct the corresponding Graph.

I agree that creating multiple Graph objects introduces some additional overhead, particularly for a large number of seeds. However, that seems like a reasonable trade-off for the usability and consistency expected from the high-level API. Users who need the most efficient batched representation can continue to call pylibcugraph directly and work with the combined output and offsets.

My preference is therefore to update this PR so that single-seed input returns one Graph, while multi-seed input returns an ordered list of Graph objects.

One compatibility point I would like to confirm: the current wrapper accepts multiple seeds and returns one combined Graph, although the individual boundaries are lost. Would changing that multi-seed result to list[Graph] be acceptable as the correction for issue #4191? If so, I will update the implementation, tests, and documentation accordingly.

rlratzel:
> One compatibility point I would like to confirm: the current wrapper accepts multiple seeds and returns one combined Graph, although the individual boundaries are lost. Would changing that multi-seed result to list[Graph] be acceptable as the correction for issue #4191? If so, I will update the implementation, tests, and documentation accordingly.

I think changing the multi-seed result to list[Graph] would not only be acceptable, but also an improvement.

Thank you again for this work, @sauravsingla !

sauravsingla:
Thank you again for confirming the API direction.

I have updated the implementation so that:

- a single seed continues to return one `cugraph.Graph`;
- multiple seeds return an ordered `list[cugraph.Graph]` aligned with the input seed order;
- the offsets returned by `pylibcugraph` are now used internally only to partition the combined edge output;
- the public `return_offsets` option has been removed.

I also updated the regression coverage for ordered multi-seed output, weighted graphs, renumbered graphs and external vertex IDs, directed graphs, multi-column seed vertices, single-seed compatibility, and empty-seed validation.

The current PR description still refers to the earlier `return_offsets=True` design. I was unable to edit the PR description through my GitHub integration, but the implementation and tests now follow the agreed `Graph` / `list[Graph]` API.

The latest branch head is `1edd932`. GPU and cuDF integration validation is still required through RAPIDS CI because my local environment is CPU-only.

Please let me know if you would like any further changes to the API or test coverage.

rlratzel (approved):
Thanks for the updates!

sauravsingla:
Thank you for the approval. The branch is now updated with the latest `main`, and the repository checks are passing. The required `pr-builder`/GPU CI has not started yet for commit `35bb3fc`. Could a maintainer please enable or retrigger it when convenient?

alexbarghi-nv (approved):

sauravsingla:
Thank you @ChuckHastings, @alexbarghi-nv, and @rlratzel for the guidance, reviews, and support throughout this contribution. I appreciate the opportunity to contribute to RAPIDS cuGraph.

PR #5590, opened 2026-07-14
Start refactoring C API and pylibcugraph to Use DLPack

alexbarghi-nv:
## Overview

Refactor pylibcugraph to use DLPack for array exchange instead of calling NumPy and CuPy APIs directly. The C API owns DLPack interpretation and validation, while the Python bindings route array conversions through that common boundary.

## Changes

- add and use C API helpers for DLPack tensor conversion
- update pylibcugraph algorithms and graph construction paths to exchange arrays through DLPack
- validate managed tensors, data types, dimensions, contiguity, and ownership behavior
- add C and Python coverage for DLPack conversion and graph construction
- The goal is to eventually get to require DLPack `>=1.0,<1.1` but staying at 0.8 in `dependencies.yaml`
- use the DLPack v1.0 CMake fallback and regenerate the checked-in 26.10 conda environments

## Release target

This PR targets `main` for the 26.10 release. The DLPack dependency updates previously tracked in #5602 are incorporated here so the API refactor and its required dependency range land together.

The cross-project DLPack migration is tracked in rapidsai/build-planning#308 and related to NVIDIA/cuvs#2354.

## Validation

- ran `rapids-dependency-file-generator --clean --warn-all --strict`
- verified dependency generation is idempotent
- parsed `dependencies.yaml` and the generated `all_*.yaml` environments with PyYAML
- ran `git diff --check`
- verified all generated conda environments use `dlpack>=1.0,<1.1`

Broader C++ and Python tests are handled by the PR's CI workflows.


BradReesWork:
This is just the first PR and not the final version.  There are a number of dependencies on other packages that prevents fully supporting DLPAck version 1.0 at this point.   But this sets up the code for simply updating the dependency file when cuDF and others have moved to DLPack 1.0

BradReesWork (approved):

cpp/cmake/thirdparty/get_dlpack.cmake
```diff
@@ -29,6 +28,6 @@ function(find_and_configure_dlpack VERSION)
 
 endfunction()
 
-set(CUGRAPH_MIN_VERSION_dlpack 0.8)
+set(CUGRAPH_MIN_VERSION_dlpack 1.0)
```
alexbarghi-nv, line 31: Do we need to change this to 0.8?  Or maybe it's safe to build with 1.0+ here as long as we set our runtime version to 0.8+.  That would make sense I think.

alexbarghi-nv (commented):
@BradReesWork just 1 comment; otherwise looks ok.

cpp/cmake/thirdparty/get_dlpack.cmake
msarahan: I'm not familiar with dlpack's version guarantees. This assumption usually does not hold, and I would set it to 0.8 to be safe - unless you know better about dlpack's behavior.

msarahan (approved):
I don't think this is safe, and I don't understand the thinking. I read https://github.com/rapidsai/build-planning/issues/308 and I understand that an earlier effort was closed for 26.08 because cudf wasn't ready. I think the correct approach is to remain pinned to 0.8 until all of the projects are ready to change over.

I approve of the added pinnings, but I think the cmake file needs to revert back to 0.8.

PR #5599, opened 2026-07-22
Fix dask cudf failure

jnke2016:
Dask single-GPU tests were intermittently failing on CI with a `ZeroDivisionError` in dask's `_repartition.py`. The cause is a worker-registration race in `simpleDistributedGraphImpl.__from_edgelist`: the worker count is queried from `client.scheduler_info()` and used to set the repartition size `(npartitions = len(workers) * 2)`, but if the workers have not registered yet, that count is 0, so repartition receives `npartitions=0` and `das`k divides by zero.

This PR addresses this issue by waiting for at least one worker to register (`_client.wait_for_workers(n_workers=1)`) before querying `scheduler_info()`, so the worker count is never transiently zero.

### DIFF
--- a/python/cugraph/cugraph/dask/centrality/betweenness_centrality.py
+++ b/python/cugraph/cugraph/dask/centrality/betweenness_centrality.py
@@ -81,6 +81,16 @@ def _mg_call_plc_betweenness_centrality(
     endpoints: bool = False,
     edge_bc: bool = False,
 ) -> dask_cudf.DataFrame:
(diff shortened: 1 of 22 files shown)


alexbarghi-nv (approved):
Thanks for getting this fixed.  This approach looks reasonable - using the stored list to get the number of workers when possible and otherwise properly calling `get_workers()` which access the raft-dask comms directly.  I assume that object persists the actual number of workers registered upon initialization.

PR #5600, opened 2026-07-22
empty commit to trigger a new build

jameslamb:
#5594 was intended to produce new nightlies with new versions in response to an ABI break in dependencies, but it looks like it picked up not-yet-rebuilt versions of dependencies, so `libcugraph.so` in the latest nightlies is still using an older RMM ABI.

```shell
docker run \
  --gpus all \
  --rm \
  -it rapidsai/citestwheel:26.08-cuda12.9.2-ubuntu22.04-py3.11 \
  bash

pip install \
  --prefer-binary \
  'pylibcugraph-cu12==26.8.*,>=26.8.0a50'

python -c "import pylibcugraph.comms"
# ImportError: libcuml.so: cannot open shared object file: No such file or directory

python -c "import libcugraph; import libraft; import libcuvs; libraft.load_library(); libcuvs.load_library(); libcugraph.load._load_wheel_installation('libcugraph.so')"
# Traceback (most recent call last):
#   File "<string>", line 1, in <module>
#   File "/pyenv/versions/3.11.15/lib/python3.11/site-packages/libcugraph/load.py", line 30, in _load_wheel_installation
#     return ctypes.CDLL(lib, PREFERRED_LOAD_FLAG)
#            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
#   File "/pyenv/versions/3.11.15/lib/python3.11/ctypes/__init__.py", line 376, in __init__
#     self._handle = _dlopen(self._name, mode)
#                    ^^^^^^^^^^^^^^^^^^^^^^^^^
# OSError: /pyenv/versions/3.11.15/lib/python3.11/site-packages/libcugraph/lib64/libcugraph_common.so: undefined symbol: _ZTIN3rmm9bad_allocE
```

This pushes another empty commit, to rebuild with newer versions of everything.

That should hopefully resolve failures observed in CI for projects depending on cuGraph, like over in https://github.com/rapidsai/cugraph-gnn


jameslamb:
cancelling CI (to save resources) and admin-merging (to immediately begin a new build)

PR #5602, opened 2026-07-22
Update DLPack to v1.0

alexbarghi-nv:
## Summary

- require DLPack `>=1.0,<1.1` in `dependencies.yaml`
- update the CMake fallback tag/minimum version to v1.0
- regenerate all checked-in conda environment files with `rapids-dependency-file-generator` 1.21.0

## Impact

cuGraph builds and downstream development environments now consistently use the DLPack 1.0 release series, while avoiding unvalidated changes from 1.1 and later.

## Release target

This PR now targets `main` for the 26.10 release. It was retargeted from `release/26.08` so the DLPack 1.x migration can be coordinated across RAPIDS projects outside the 26.08 code freeze. The cross-project support plan is tracked in rapidsai/build-planning#308.

## Validation

- reran `rapids-dependency-file-generator` and verified the output is idempotent
- parsed `dependencies.yaml` and all generated `all_*.yaml` environment files with PyYAML
- ran `git diff --check`
- verified all generated conda environments use the `>=1.0,<1.1` constraint

## Constraints

- coordinate with rapidsai/build-planning#308 before merging


bdice (changes requested):
This is blocked without a DLPack update in cuDF.

```
error    libmamba Could not solve for environment specs
    The following packages are incompatible
    ├─ cuda-version =13.3 * is requested and can be installed;
    ├─ dlpack >=1.0,<1.1 * is requested and can be installed;
    └─ libcudf =26.8,>=0.0.0a0 * is not installable because there are no viable options
       ├─ libcudf [26.08.00a943|26.08.00a944|...|26.08.00a988] would require
       │  └─ cuda-version >=12,<13.0a0 *, which conflicts with any installable versions previously reported;
       └─ libcudf [26.08.00a943|26.08.00a944|...|26.08.00a988] would require
          └─ dlpack >=0.8,<1.0 *, which conflicts with any installable versions previously reported.
```

However, cuDF is frozen for 26.08. The version discrepancy is blocking our CCCL-RAPIDS CI, which is one of the places where we need to support "unified" environments with all of RAPIDS.

We need to revert https://github.com/NVIDIA/cuvs/pull/2354 for 26.08 and retarget this DLPack update to 26.10. As a part of that, we also need to make sure we are following DLPack practices for version stability:

> Starting DLPack 1.0, the DLManagedTensorVersioned struct should be used and the existing DLManagedTensor struct is considered deprecated, though a library should try to support both during the transition period if possible.

For the future, it would be helpful to document the motivation for cross-library dependency changes like this with an issue in https://github.com/rapidsai/build-planning/issues. The motivation for this update is not specified in this PR or in the cuVS PR.

alexbarghi-nv:
Superseded by #5590, which now includes the DLPack v1.0 CMake fallback, the `dlpack>=1.0,<1.1` dependency constraint, and regenerated 26.10 environment files.

PR #5603, opened 2026-07-22
Time window sampling

### LINKED ISSUE #5573: Time Window Sampling
Add a new feature to sampling to provide time windows.  This is to support https://github.com/rapidsai/cugraph-gnn/issues/450.

The objective is to allow each seed to be passed with a time window.  Some notes:
1. The start time would be used as the start time just like in temporal sampling
2. The end time would restrict end times for edges in temporal sampling, the end time does not change as we traverse the graph
3. At least initially we would start with requiring disjoint sampling to be set, since edge times if a vertex appears in multiple places would be problematic.

ChuckHastings:
Add a new feature - time window sampling - to our sampling algorithms.

Closes #5573


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
seunghwak, line 2: What is this? When should we add " & AFFILIATES. All rights reserved."?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -448,8 +448,13 @@ heterogeneous_biased_neighbor_sample(
  * @param edge_end_time_view Optional view object holding edge end times for @p graph_view.
  * @param starting_vertices Device span of starting vertex IDs for the sampling.
  * In a multi-gpu context the starting vertices should be local to this GPU.
- * @param starting_vertex_times Optional device span of times associated with each starting
- * vertex for the sampling.
+ * @param starting_vertex_times Optional device span of start times defining the lower bound of the
```
seunghwak: Should we better rename this to `starting_vertex_start_times` (to be consistent with `starting_vertex_end_times`)?

seunghwak: Now we have both start & end times, in the decreasing run (no window), the caller may just provide the end time, in the increasing run, the caller provides just the start time, and for time window sampling (either decreasing or increasing runs), the caller provides both?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -540,8 +546,13 @@ homogeneous_uniform_temporal_neighbor_sample(
...
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -640,8 +652,13 @@ heterogeneous_uniform_temporal_neighbor_sample(
...
+ * time window for each starting vertex. When specified with @p starting_vertex_end_times, must be
+ * less than or equal to the corresponding end time. For increasing walks the frontier begins at
+ * this time; for decreasing walks the frontier begins at @p starting_vertex_end_times.
+ * @param starting_vertex_end_times Optional device span of end times defining the upper bound of
+ * the time window for each starting vertex. Edge start times must fall within
+ * [@p starting_vertex_times, @p starting_vertex_end_times] when both bounds are specified.
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -739,8 +757,13 @@ homogeneous_biased_temporal_neighbor_sample(
...
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -548,8 +548,13 @@ CUGRAPH_EXPORT cugraph_error_code_t cugraph_heterogeneous_biased_neighbor_sample
...
  * @param [in]  start_vertices Device array of start vertices for the sampling
- * @param [in]  starting_vertex_times Optional device span of times associated with each starting
- * vertex for the sampling.
+ * @param [in]  starting_vertex_times Optional device span of start times defining the lower bound
+ * of the time window for each starting vertex. When specified with @p starting_vertex_end_times,
+ * must be less than or equal to the corresponding end time. For increasing walks the frontier
+ * begins at this time; for decreasing walks the frontier begins at @p starting_vertex_end_times.
+ * @param [in]  starting_vertex_end_times Optional device span of end times defining the upper bound
+ * of the time window for each starting vertex. Edge start times must fall within
+ * [@p starting_vertex_times, @p starting_vertex_end_times] when both bounds are specified.
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -603,8 +609,13 @@ CUGRAPH_EXPORT cugraph_error_code_t cugraph_homogeneous_uniform_temporal_neighbo
...
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -657,8 +669,13 @@ CUGRAPH_EXPORT cugraph_error_code_t cugraph_homogeneous_biased_temporal_neighbor
...
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/include/cugraph_c/sampling_algorithms.h
```diff
@@ -718,8 +736,13 @@ CUGRAPH_EXPORT cugraph_error_code_t cugraph_heterogeneous_uniform_temporal_neigh
...
```
seunghwak: Similar here, should we better rename this to `starting_vertex_start_times`?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -441,16 +451,16 @@ gather_one_hop_edgelist_to_unvisited_neighbors(
...
+           arithmetic_device_uvector_t,
            std::optional<rmm::device_uvector<int32_t>>>
 temporal_gather_one_hop_edgelist(
   raft::handle_t const& handle,
   graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
-  raft::host_span<edge_arithmetic_property_view_t<edge_t>> edge_property_views,
   edge_property_view_t<edge_t, time_stamp_t const*> edge_time_view,
   std::optional<edge_property_view_t<edge_t, int32_t const*>> edge_type_view,
   raft::device_span<vertex_t const> active_majors,
   raft::device_span<time_stamp_t const> active_major_times,
```
seunghwak: Should this better be `active_major_window_starts`?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -460,114 +470,161 @@ temporal_gather_one_hop_edgelist(
...
-  // FIXME:  Modify sampling to take raft::host_span of arithmetic edge property views directly
-  // rather than creating and unpacking them
-  CUGRAPH_EXPECTS(edge_property_views.size() > 0,
-                  "Temporal requires at least one edge property - the time");
   using label_t = int32_t;
 
-  // Only used if active_major_labels is set
-  kv_store_t<size_t, cuda::std::tuple<time_stamp_t, label_t>, true> kv_store{handle.get_stream()};
-  std::optional<rmm::device_uvector<size_t>> tmp_positions{std::nullopt};
+  // Sorted per-source side tables keyed by (major, label).  Used when labels are present.
```
seunghwak: What does side mean here? This is basically a KV store with K = (major, label) and V = (start time, end_time), correct?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -460,114 +470,161 @@ temporal_gather_one_hop_edgelist(
...
+  std::optional<rmm::device_uvector<vertex_t>> labeled_side_majors{std::nullopt};
+  std::optional<rmm::device_uvector<label_t>> labeled_side_labels{std::nullopt};
+  std::optional<rmm::device_uvector<time_stamp_t>> labeled_side_times{std::nullopt};
+  std::optional<rmm::device_uvector<time_stamp_t>> labeled_side_window_ends{std::nullopt};
+  std::optional<rmm::device_uvector<label_t>> edge_src_labels{std::nullopt};
 
   // Only used if active_major_labels is not set
   std::optional<rmm::device_uvector<time_stamp_t>> tmp_times{std::nullopt};
```
seunghwak: `tmp_window_starts`?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -460,114 +470,161 @@ temporal_gather_one_hop_edgelist(
...
+  std::optional<rmm::device_uvector<time_stamp_t>> tmp_window_ends{std::nullopt};
 
   // Only used if graph_view is a multi_graph
   arithmetic_device_uvector_t tmp_edge_indices{std::monostate{}};
```
seunghwak: Are these "edge indices" or "multi-edge indices"?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -460,114 +470,161 @@ temporal_gather_one_hop_edgelist(
...
-        handle.get_comms(), active_majors.size(), handle.get_stream());
-      std::exclusive_scan(sizes.begin(), sizes.end(), sizes.begin(), size_t{0});
-      starting_pos = sizes[handle.get_comms().get_rank()];
+    // Key the per-source (time, window_end) by (major, label).  Under disjoint sampling that key is
+    // unique in the frontier; on MG, allgather across minor_comm and keep the time extremum
+    // required by the temporal mode so gather filtering matches the sample path.
+    rmm::device_uvector<vertex_t> kv_majors(active_majors.size(), handle.get_stream());
+    rmm::device_uvector<label_t> kv_labels(active_major_labels->size(), handle.get_stream());
+    rmm::device_uvector<time_stamp_t> kv_times(active_major_times.size(), handle.get_stream());
+    rmm::device_uvector<time_stamp_t> kv_window_ends(active_majors.size(), handle.get_stream());
```
seunghwak: Any reason that we are using std::optional elsewhere but setting this to the extreme value here when `active_major_window_ends.has_value()` is false here?

cpp/src/sampling/detail/gather_one_hop_impl.cuh
```diff
@@ -694,21 +766,143 @@ temporal_gather_one_hop_edgelist(
...
+           std::ignore,
+           std::ignore,
+           std::ignore,
+           std::ignore) = deduplicate_edges_by_minor(handle,
+                                                     graph_view,
+                                                     std::move(result_majors),
+                                                     std::move(result_minors),
+                                                     std::move(tmp_edge_indices),
+                                                     arithmetic_device_uvector_t{std::monostate{}},
+                                                     std::move(result_labels));
```
seunghwak, line 929: If I am not mistaken, we SG de-duplicate as a code block while MG de-duplicate in a function? Any reason for this? Isn't it more natural to handle both SG & MG de-duplicate in a de-duplicate function?

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -46,9 +47,12 @@ prepare_next_frontier(
   raft::device_span<vertex_t const> sampled_src_vertices,
   std::optional<raft::device_span<label_t const>> sampled_src_vertex_labels,
   std::optional<raft::device_span<time_stamp_t const>> sampled_src_vertex_times,
```
seunghwak: Similar here, `sampled_src_vertex_times`=>`sampled_src_vertex_window_starts`?

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
```diff
@@ -46,9 +47,12 @@ prepare_next_frontier(
...
+  std::optional<raft::device_span<time_stamp_t const>> sampled_src_vertex_window_ends,
   raft::host_span<raft::device_span<vertex_t const>> sampled_dst_vertices,
   std::optional<raft::host_span<raft::device_span<label_t const>>> sampled_dst_vertex_labels,
   std::optional<raft::host_span<raft::device_span<time_stamp_t const>>> sampled_dst_vertex_times,
```
seunghwak: Similar here, `sampled_dst_vertex_times`=>`sampled_dst_vertex_window_start_spans`?

cpp/src/utilities/arithmetic_variant_types.cpp
```diff
@@ -0,0 +1,54 @@
...
+#include <variant>
+#include <vector>
+
+namespace cugraph {
+namespace detail {
+
+CUGRAPH_EXPORT arithmetic_device_uvector_t
+concatenate_spans(raft::handle_t const& handle,
+                  std::vector<arithmetic_device_span_t> const& spans,
+                  size_t result_size)
```
seunghwak, line 26: Concatenating spans is a much more generic feature (not sampling specific), should we better place this somewhere else? (e.g. where we define arithmetic_device_span_t?)

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: It's automatically getting added by the pre-commit checker.  The guidance is to add it as we change files rather than just updating copyright headers everywhere.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Renamed in latest push.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Renamed

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Renamed

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Renamed

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Renamed in next push

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Renamed in next push

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Renamed in next push

cpp/include/cugraph_c/sampling_algorithms.h
ChuckHastings: Renamed in next push

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Renamed in next push

cpp/src/utilities/arithmetic_variant_types.cpp
ChuckHastings: Moved definition to arithmetic_variant_types.h, moved implementation to new file arithmetic_variant_types.cpp in next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Renamed to kv_majors, kv_labels, kv_window_starts and kv_window_ends in next push.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: renamed in next push

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Renamed to tmp_multi_edge_indices in the next push

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
ChuckHastings: Fixed in next push

cpp/src/sampling/detail/prepare_next_frontier_impl.cuh
ChuckHastings: fixed in next push

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: I'm not sure I understand your comment here.  Perhaps the comment on line 886 above is confusing the issue?

deduplicate_edges_by_minor is called for both SG and MG mode, this function is shared for regular sampling and for temporal sampling to be part of the logic for implementing disjoint sampling which requires that we only visit each vertex (per label) once in our sampling flow.  For the gather_one_hop logic since we've already excluded vertices that arrive at an already visited vertex, but if multiple edges in this frontier arrive at the same vertex we need to pick one to win and the others to lose.  That's what this function does, and it is called for both SG and MG modes.

The comment above was essentially justifying why the gather_sampled_properties logic was changed (in one of the intermediate commits we gathered sample properties before deduplicating, which worked fine in SG but in MG mode we ended up with illegal memory references.

I think the above comment was helpful to me for debugging, but is probably not relevant to the final code.  I can remove that if that helps.  Or if I'm misunderstanding your question, please clarify.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: > Now we have both start & end times, in the decreasing run (no window), the caller may just provide the end time, in the increasing run, the caller provides just the start time, and for time window sampling (either decreasing or increasing runs), the caller provides both?

Yes, that is the behavior.

alexbarghi-nv (approved):
C API and PLC changes look good.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
ChuckHastings: Refactored the code in the next push to propagate the optionals all the way into the device functor so that we don't materialize the temporal views and populate them with the extreme values except for the one case where we are doing temporal sampling but no time values were specified with the seeds.  In that case, only on hop 0, we'll populate the extreme values to make the computations consistent for each hop.

cpp/src/sampling/detail/gather_one_hop_impl.cuh
seunghwak: Oh, my bad. I thought the above code removing visited minors is the SG part of de-duplication.

OK, so the first is removing already visited mionrs and de-duplication follows this.

We may consider updating the comments to make this clearer.

"Keep graph edge indices aligned so
  // subsequent MG deduplication and property gather remain valid." => somewhat misled me to think that the follow-up deduplication is for MG. I assume edge indices (or multi-edge indices?) are relevant for both SG & MG.

seunghwak (approved):
LGTM

PR #5605, opened 2026-07-23
enforce 'yamllint' checks

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/305

Proposes enforcing `yamllint` checks here. My primary motivation is to catch correctness issues in `dependencies.yaml` files, like duplicate entries silently resolving to the last one or indentation mistakes leading to filters being ignored.

But this also has some side benefits for consistency, which makes it a bit easier to write automation.


codecov.yml
```diff
@@ -2,5 +2,5 @@
 comment: false
 coverage:
   status:
-    project: off
-    patch: off
+    project: off  # yamllint disable-line rule:truthy
+    patch: off  # yamllint disable-line rule:truthy
```
jameslamb, line 6: similar to https://github.com/NVIDIA/cuopt/pull/1613#discussion_r3641189058 ... `codecov`'s docs explicitly say to use `off`, so let's not touch it.

conda/recipes/cugraph/recipe.yaml
```diff
@@ -30,7 +30,7 @@ build:
   prefix_detection:
     # See https://github.com/rapidsai/build-planning/issues/160
     # Blanket ignore here as there are quite a few shared objects shipped in cugraph
-    ignore_binary_files: True
+    ignore_binary_files: true
```
jameslamb, line 33: `yamllint` caught this.

YAML 1.1 treats `True` (capital `T`) as a boolean `true`.

YAML 1.2 only supports `true` / `false` and treats `True` as the string `"True"`. See https://yaml.org/spec/1.2.2/#10212-boolean

This maybe hasn't bitten us yet because a lot of YAML parsers still respect the 1.1 spec (example: https://github.com/yaml/pyyaml/issues/116), but it's a nice bit of future-proofing to just use the values supported by both versions.

KyleFromNVIDIA (approved):

.yamllint.yaml
```diff
@@ -0,0 +1,37 @@
...
+  comments-indentation: disable
+  document-end: disable
+  document-start: disable
+  key-duplicates:
+    forbid-duplicated-merge-keys: true
+  line-length: disable
+  truthy:
+    allowed-values: ['false', 'true']
+    # having problematic value in keys is rare... and also
+    # GitHub Actions' choie of 'on:' triggers this check
```
jameslamb:
```suggestion
    # GitHub Actions' choice of 'on:' triggers this check
```

bdice (approved):

dependencies.yaml
```diff
@@ -458,8 +458,8 @@ dependencies:
     common:
       - output_types: [conda, pyproject, requirements]
         packages:
-          - &dask rapids-dask-dependency==26.10.*,>=0.0.0a0
-          - &numba numba>=0.60.0,<0.65.0
+          - rapids-dask-dependency==26.10.*,>=0.0.0a0
+          - numba>=0.60.0,<0.65.0
           - &numpy numpy>=2.0,<3.0
```
alexbarghi-nv, line 463: How come we're no longer consistent here - I see `&numpy` is still unchanged but we dropped `&dask` and `&numba` here.  Is that because they're not referenced anywhere else or was this unintentional?

alexbarghi-nv (changes requested):
Looks good but I had one question before approving.

dependencies.yaml
jameslamb: Yeah exactly... they're not referenced any more. `yamllint` warns about unused anchors to help us clean up stuff like that.

`*numpy` is used here:

https://github.com/rapidsai/cugraph-gnn/blob/64241899dd43d32f2ae73e6746ee1d2f70d229c4/dependencies.yaml#L361

alexbarghi-nv (approved):

PR #5608, opened 2026-07-24
test: isolate library wheel smoke test

bdice:
## Description

Install the library wheel in a clean virtual environment and call its `load_library()` entry point before installing the high-level wheel or test extras. This exposes missing library-wheel runtime dependencies that the full test environment can otherwise mask.

xref: https://github.com/rapidsai/build-planning/issues/307


jameslamb (approved):

jameslamb:
CI got stuck here in the latest GitHub outage, just merged in latest `main` to trigger a new run. I'll check back in a bit.

ci/test_wheel_pylibcugraph.sh
```diff
@@ -13,6 +13,17 @@ LIBCUGRAPH_WHEELHOUSE=$(rapids-download-from-github "$(rapids-artifact-name whee
...
 
+python -m venv libcugraph-env
+. libcugraph-env/bin/activate
+
+rapids-pip-retry install \
+    -v \
+    --prefer-binary \
+    --constraint "${PIP_CONSTRAINT}" \
+    "${LIBCUGRAPH_WHEELHOUSE}"/libcugraph*.whl
+python -c "import libcugraph; assert (libraries := libcugraph.load_library()) and all(libraries)"
```
jameslamb, line 24: This is failing only for CUDA 12 right now.

```text
+ python -c 'import libcugraph; assert (libraries := libcugraph.load_library()) and all(libraries)'
    import libcugraph; assert (libraries := libcugraph.load_library()) and all(libraries)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/33093433018/job/98612623998?pr=5608#step:13:142))

I could reproduce that locally.

```shell
docker run \
  --rm \
  --gpus all \
  -it rapidsai/citestwheel:26.10-cuda12.9.2-rockylinux8-py3.13 \
  bash

pip install \
  --extra-index-url https://pypi.anaconda.org/rapidsai-wheels-nightly/simple/ \
  'libcugraph-cu12>=26.10.0a0'

python -c "import libcugraph; assert (libraries := libcugraph.load_library()) and all(libraries)"
```

Picking through the internals of `libcugraph.load_library()`, the root error is from loading `libcuvs.so`.

```text
>>> lib = "/pyenv/versions/3.13.15/lib/python3.13/site-packages/libcuvs/lib64/libcuvs.so"
... ctypes.CDLL(lib, ctypes.RTLD_LOCAL)
... 
  File "<python-input-2>", line 2, in <module>
    ctypes.CDLL(lib, ctypes.RTLD_LOCAL)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.13.15/lib/python3.13/ctypes/__init__.py", line 361, in __init__
    self._handle = self._load_library(name, mode, handle, winmode)
                   ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.13.15/lib/python3.13/ctypes/__init__.py", line 403, in _load_library
    return _dlopen(name, mode)
OSError: libnvrtc.so.12: cannot open shared object file: No such file or directory
```

`libcuvs` is missing `cuda-pathfinder`, which it needs to load nvRTC. I put up a fix here: https://github.com/NVIDIA/cuvs/pull/2524

Once that's merged, we should see CI pass on a re-run here... in my local environment, just running `pip install cuda-pathfinder` was sufficient for `libcugraph.load_library()` to succeed.

jameslamb: Seeing jobs pass now that that cuVS fix was merged. I'll queue this to merge.

PR #5609, opened 2026-07-24
Improve performance of identifying unique seeds for sampling frontier

### LINKED ISSUE #5604: [PERF] Finite-fanout neighbor sampling dominated by compute_unique_keys mapping
### Summary

We are benchmarking a relational temporal neighbor sampler backed by
`pylibcugraph` against the existing CPU `pyg-lib` implementation. On an NVIDIA
L4, cuGraph scales well for one hop and is approximately 2x faster than PyG at
batch size 8,192. For two-hop finite-fanout sampling, however, cuGraph becomes
slower than PyG. Nsight profiling attributes most of the additional second-hop
GPU time to the original-frontier-key to unique-key-index mapping inside
`cugraph::detail::compute_unique_keys`.

We would appreciate guidance on whether this is a known scaling limitation,
whether our use of the sampling API is triggering an unintended path, and
whether the mapping implementation could be improved.

### Workload and benchmark contract

- Dataset/task: RelBench `rel-arxiv` / `paper-citation`, test split.
- Operation: heterogeneous temporal neighbor sampling, uniform, without
  replacement.
- Fanout: `[16]` for one hop and `[16, 16]` for two hops.
- Batch: 8,192 task/query rows passed as one disjoint sampler request.
- Timing: median of 50 public sampler calls after 10 warmups.
- CUDA data and task inputs are already device-resident. Topology construction,
  initial host-to-device transfer, output materialization, correctness checks,
  and profiling overhead are excluded.
- Both implementations use the same task rows, cutoff semantics, seed, package
  environment, and output invariants.

Environment:

- NVIDIA L4, 58 SMs, 24 GB.
- Driver 580.126.09.
- CUDA 13 / PyTorch 2.11.0+cu130.
- `libcugraph-cu13==26.6.0`, `pylibcugraph-cu13==26.6.0`.
- `pyg-lib==0.7.0+pt211cu130`.
- Nsight Systems 2025.3.2 and Nsight Compute 2025.3.1.

### Uninstrumented results

| Fanout | Batch | PyG CPU median | cuGraph L4 median | Result |
|:-------|------:|---------------:|------------------:|:-------|
| `[16]` | 8,192 | 35.384 ms | 17.965 ms | cuGraph 1.97x faster |
| `[16, 16]` | 8,192 | 71.726 ms | 85.338 ms | PyG 1.19x faster |

At `[16, 16]`, cuGraph sampled approximately 3.9% more rows because finite
sampling is stochastic, but sampled-row throughput still favored PyG by about
1.15x. All correctness invariants passed.

The first hop expands the 8,192 seeds into an approximately 99,000-occurrence
mixed frontier. The cuGraph wrapper then invokes the temporal native sampler
for hop two using that device-resident frontier.

### Nsight Systems observation

We added NVTX ranges around each native hop and captured one request after five
warmups. The instrumentation does not add synchronization inside a hop.
GPU-metric tracing perturbs absolute latency, so the following numbers are used
only to attribute work:

| `[16, 16]` phase | Native wall time | Kernel count | Summed GPU kernel time |
|:-----------------|-----------------:|-------------:|-----------------------:|
| Hop 1 | 21.0 ms | 219 | 5.9 ms |
| Hop 2 | 68.5 ms | 298 | 42.2 ms |

Two CUB transform launches whose demangled names contain
`cugraph::detail::compute_unique_keys` account for 31.4 ms, or 74% of all
hop-two GPU kernel time.

The final explicit device synchronization after the complete request is only
about 0.04 ms. The native call performs only 301 bytes of device-to-host
copies, so the regression is not explained by materializing the frontier on
the CPU. The hop-two native call contains 183 `cudaMalloc` and 176 `cudaFree`
calls; synchronous frees occupy 45.3 ms of traced host API time in aggregate,
including waits coincident with the two dominant kernels. These API times
overlap GPU work and are not added to kernel time.

As a useful control, batch-8,192 exhaustive `[-1, -1]` returns more rows
(274,355 versus 214,631) but its profiled second native hop takes 23.5 ms with
only 6.4 ms of GPU kernel time. It avoids the dominant finite-fanout
`compute_unique_keys` transform path. This suggests that output size alone is
not the explanation.

### Nsight Compute observation

A targeted profile of one dominant mapping launch reports:

| Metric | Value |
|:-------|------:|
| Grid / block | 190 blocks / 256 threads |
| Waves per SM | 0.55 |
| Theoretical / achieved occupancy | 100% / 32.15% |
| Active / eligible warps per scheduler | 3.86 / 0.39 |
| Scheduler cycles with no eligible warp | 70.89% |
| Long-scoreboard stall cycles per issued instruction | 7.58 |
| Total warp cycles per issued instruction | 13.25 |
| Compute / memory throughput | 24.31% / 14.96% |
| Aggregate L1 / L2 hit rate | 97.71% / 98.60% |

Nsight attributes approximately 57% of the cycles between issued instructions
to long-scoreboard waits. The launch has too few waves, and too few eligible
warps, to hide the dependent lookup latency. It is not close to either compute
or memory-bandwidth saturation.

The Nsight Compute duration itself is not used because multi-pass kernel replay
perturbs timing; the Systems trace provides the time attribution.

### Suspected implementation issue

In the 26.06 implementation, `compute_unique_keys` first sorts and uniques the
frontier keys per local edge partition. It then builds
`aggregate_local_frontier_key_idx_to_unique_key_idx` using a transform over the
original keys. Each transform invocation performs a sequential
`thrust::find(thrust::seq, unique_key_first, unique_key_last, key)` over the
unique-key range:

https://github.com/rapidsai/cugraph/blob/v26.06.00/cpp/include/cugraph/prims/detail/sample_and_compute_local_nbr_indices.cuh#L425-L437

For `N` original frontier keys and `U` unique keys in a partition, this appears
to perform `O(N * U)` comparisons in the worst case, parallelized over `N`.
The observed kernel signature uses `const long*` keys, and the dominant
transform is the original source-vertex-ID to unique-source-index mapping,
rather than random edge-key generation.

This seems consistent with the profile: high aggregate cache hit rates but
large instruction count, dependent L1TEX scoreboard stalls, low eligible-warp
count, and an imbalanced tail.

We may be missing an assumption that keeps `U` small in expected workloads.
If not, could the inverse mapping be emitted while sorting/grouping the
`(key, original_index)` pairs, or use a logarithmic/hashed lookup instead of a
linear search? We would value maintainers' advice before attempting a change.

### Reproduction and retained profiles

Benchmark and NVTX profiler:

- Internal benchmark MR:
  https://gitlab-master.nvidia.com/kumo/sdm-benchmark/-/merge_requests/1
- Sampler implementation used by the cuGraph side:
  https://github.com/NVIDIA/structured-data-models/pull/256

The profiler command is:

```shell
export RELBENCH_CACHE_DIR=/path/to/relbench-cache

nsys profile \
  --trace=cuda,nvtx,osrt --sample=none --cpuctxsw=none \
  --capture-range=cudaProfilerApi --capture-range-end=stop \
  --gpu-metrics-devices=0 --gpu-metrics-set=ad10x \
  --gpu-metrics-frequency=10000 \
  --output=b8192_f16x16 \
  .venv/bin/python scripts/profile_cugraph_hops.py \
  --batch-size 8192 --fanout 16,16 --warmups 5

ncu \
  --nvtx --nvtx-include 'hop_2_pylibcugraph/' \
  --kernel-name-base demangled \
  --kernel-name 'regex:compute_unique_keys' \
  --launch-count 1 \
  --section SpeedOfLight --section LaunchStats --section Occupancy \
  --section ComputeWorkloadAnalysis --section MemoryWorkloadAnalysis \
  --section SchedulerStats --section WarpStateStats \
  --export=b8192_f16x16_hop2_unique \
  .venv/bin/python scripts/profile_cugraph_hops.py \
  --batch-size 8192 --fanout 16,16 --warmups 5
```

We are happy to reduce this to a standalone pylibcugraph reproducer or test a
candidate implementation if maintainers can advise on the intended API path.

### Code of Conduct

- [x] I agree to follow cuGraph's Code of Conduct.
- [x] I searched open issues and found no duplicate for this problem.

ChuckHastings:
Closes #5604 

Added C unit test to demonstrate slow sampling performance (used rmat rather than the referenced data sets)
Change from find_if to lower_bound in the sampling primitive since the data in the range searched is sorted right before this call.

|Fanout | Before | After | Speedup
|-- | -- | -- | --
|[16] | 20.725 ms | 13.177 ms | 1.57× (~36% less)
|[16, 16] | 142.424 ms | 98.129 ms | 1.45× (~31% less)


cpp/include/cugraph/prims/detail/sample_and_compute_local_nbr_indices.cuh
```diff
@@ -442,9 +443,7 @@ compute_unique_keys(raft::handle_t const& handle,
                            local_frontier_unique_key_offsets[i + 1]] __device__(key_t key) {
           return static_cast<size_t>(cuda::std::distance(
             unique_key_first,
-            thrust::find_if(thrust::seq, unique_key_first, unique_key_last, [key](auto unique_key) {
-              return unique_key == key;
-            })));
+            thrust::lower_bound(thrust::seq, unique_key_first, unique_key_last, key)));
         }));
```
seunghwak: Have you tried this?
```
thrust::lower_bound(
  handle.get_thrust_policy(),
  get_dataframe_buffer_begin(aggregate_local_frontier_unique_keys) +
                            local_frontier_unique_key_offsets[i],
  get_dataframe_buffer_begin(aggregate_local_frontier_unique_keys) +
                           local_frontier_unique_key_offsets[i + 1],
  
  aggregate_local_frontier_key_first + local_frontier_offsets[i],
  aggregate_local_frontier_key_first + local_frontier_offsets[i + 1],
  aggregate_local_frontier_key_idx_to_unique_key_idx.begin() + local_frontier_offsets[i]);
```

ChuckHastings: Changed to this.

Comparable performance improvement, simpler code.

seunghwak (approved):
LGTM

PR #5612, opened 2026-07-28
Fix flaky MG ZeroDivisionError in symmetrize_ddf

jnke2016:
Dask single-GPU tests were still intermittently failing on CI with a ZeroDivisionError in dask's _repartition.py. PR #5599  fixed this in the graph-build path by sizing the repartition from a race-free worker count, but one path was left unguarded: `symmetrize_ddf` in `symmetrize.py` still set `npartitions = len(scheduler_info()["workers"]) * 2`.

This PR resolves it by sizing the repartition (and drop-duplicates) from `get_n_workers() (scheduler_info()["n_workers"])`, the race-free worker count that is never emptied, matching the approach already used elsewhere

### DIFF
--- a/python/cugraph/cugraph/structure/symmetrize.py
+++ b/python/cugraph/cugraph/structure/symmetrize.py
@@ -160,9 +160,7 @@ def symmetrize_ddf(
         return result
     else:
         vertex_col_name = src_name + dst_name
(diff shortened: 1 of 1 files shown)


alexbarghi-nv (approved):
Thanks for the quick turnaround! Looks good!

PR #5613, opened 2026-07-28
Primitive MG tests not actually validating results

ChuckHastings:
thrust::equal calls were not actually doing anything because *_begin was called for both the start and end iterator.

Correct iterator end call, then handle renumbering discrepancy that this fix reveals.


seunghwak (approved):
LGTM

PR #5619, opened 2026-08-04
Update to rapids-logger 0.3

bdice:
Contributes to https://github.com/rapidsai/build-planning/issues/313.

Updates rapids-logger to version 0.3.


vyasr (approved):

vyasr:
It seems wrong that rapids-logger isn't in dependencies.yaml at all, but I did verify that is the case. Probably somewhere either we're not making ideal usage of dfg in CI for cugraph or we're relying on transitive deps. Out of scope for this PR though.

PR #5625, opened 2026-08-12
Fixed window temporal sampling

### LINKED ISSUE #5593: Temporal sampling: support a fixed per-seed cutoff across hops
## Problem

Temporal sampling currently propagates the sampled edge timestamp to the next hop. Some node-time workloads instead require every hop to use the original per-seed cutoff, without imposing temporal ordering between consecutive edges.

For example:

```text
seed cutoff = 10
hop 1: root -> A at time 3
hop 2: A -> B at time 8
```

Both edges should be eligible because both timestamps are at or before the original cutoff. With monotonically decreasing traversal, the first edge changes the next-hop bound to 3, so the edge at time 8 is excluded.

#5573 proposes per-seed time windows with an end time that remains fixed during traversal. That may overlap with this requirement, but it is unclear whether the changing start time would still enforce cross-hop ordering.

## Target behavior

Provide a temporal sampling policy where:

- each seed has its own cutoff;
- the same original cutoff is applied at every hop;
- sampled edge timestamps do not replace that cutoff;
- separate seed labels retain separate cutoffs, including when they reference the same source vertex.

For the example above, a two-hop call should return both edges at times 3 and 8.

This policy should be available for homogeneous and heterogeneous temporal neighbor sampling. It is independent of the neighbor selection strategy: eligible edges may subsequently be sampled uniformly or selected using latest-`k`.

ChuckHastings:
Add fixed window temporal sampling implementation, closes #5593 

Also lays the groundwork for implementing last-n sampling, updating the API framework temporal sampling to make the last-n choice orthogonal to the temporal parameters.


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -42,7 +43,17 @@ enum class temporal_sampling_comparison_t {
   STRICTLY_DECREASING,      /** Time strictly decreasing (each time is before the previous one) */
   MONOTONICALLY_DECREASING, /** Time monotonically decreasing (could have multiple edges with same
                                 time) */
-  LAST                      /** Support last n behavior */
+  FIXED_WINDOW,             /** Apply the original per-seed time window at every hop */
+  LAST = FIXED_WINDOW       /** Deprecated alias for FIXED_WINDOW */
```
seunghwak: What is this? Last-N is not a temporal sampling method, it is a neighbor selection method (as specified in `neighbor_selection_t`), right? Should we better delete this to avoid confusion?

seunghwak: Or last-N or first-N are based on time stamps, not priorities (biases)?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -42,7 +43,17 @@ enum class temporal_sampling_comparison_t {
...
+};
+
+/**
+ * @brief Selects neighbors from the set that satisfies the sampling filters.
+ */
+enum class neighbor_selection_t {
+  RANDOM = 0, /** Random selection. Uniform if no bias view is supplied, biased otherwise. */
+  FIRST,      /** Deterministically select the earliest edges. Not yet implemented. */
+  LAST        /** Deterministically select the latest edges. Not yet implemented. */
```
seunghwak: Similar here, last-N or first-N are based on time stamps?

In this case, does it make sense to set neighbor_selection_t to LAST and temporal_sampling_comparison_t to STRICTLY_INCREASING?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -889,6 +913,65 @@ heterogeneous_biased_temporal_neighbor_sample(
...
+  raft::device_span<vertex_t const> starting_vertices,
+  std::optional<raft::device_span<time_stamp_t const>> starting_vertex_start_times,
+  std::optional<raft::device_span<time_stamp_t const>> starting_vertex_end_times,
+  std::optional<raft::device_span<int32_t const>> starting_vertex_labels,
+  std::optional<raft::device_span<int32_t const>> label_to_output_comm_rank,
+  raft::host_span<int32_t const> fan_out,
+  std::optional<edge_type_t> num_edge_types,
+  neighbor_selection_t neighbor_selection,
+  std::optional<temporal_sampling_comparison_t> temporal_sampling_comparison,
+  sampling_flags_t sampling_flags,
```
seunghwak: Now we have `neighbor_selection_t`, `temporal_sampling_comparison_t`, and `sampling_flags_t`. But aren't the first two sampling flags (or sampling options)?

cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -889,6 +913,65 @@ heterogeneous_biased_temporal_neighbor_sample(
...
+ * @brief Sample a homogeneous or heterogeneous neighborhood, optionally using temporal filters.
+ *
+ * This unified API replaces the homogeneous/heterogeneous, uniform/biased, and
+ * temporal/non-temporal entry-point matrix above.
+ *
+ * Sampling is heterogeneous when @p num_edge_types is specified; in that case @p edge_type_view
+ * is required and @p fan_out contains one value per (hop, edge type). Otherwise, @p fan_out
+ * contains one value per hop.
+ *
+ * RANDOM selection samples uniformly when @p edge_bias_view is absent and samples according to
```
seunghwak, line 937: RANDOM=>neighbor_selection_t::RANDOM?

or something like

`neighbor_selection_t ` dictates a neighbor selection method. RANDOM selection samples ... ?

And documentation for template and input parameters and return values are missing here.

seunghwak (commented):
Quick reviews/questions about the API changes.

More reviews on the implementation part are in progress.

cpp/src/c_api/neighbor_sampling.cpp
```diff
@@ -876,53 +776,29 @@ cugraph_error_code_t cugraph_homogeneous_uniform_neighbor_sample(
...
-                                    nullptr,
-                                    fan_out,
-                                    1,  // num_edge_types
-                                    std::move(options_cpp),
-                                    FALSE,
-                                    do_expensive_check};
-  return cugraph::c_api::run_algorithm(graph, functor, result, error);
+  auto const* options_cpp =
+    reinterpret_cast<cugraph::c_api::cugraph_sampling_options_t const*>(options);
+  return cugraph_neighbor_sample(
```
seunghwak: Any reason the biased wrappers call run_neighbor_sample() (https://github.com/rapidsai/cugraph/pull/5625/changes#diff-d268da7d4ce7574a739c9818704be440c1d2d633f087c39f8a73736b2f057859R746) directly while the uniform wrappers go through `cugraph_neighbor_sample()`? Better be more consistent about call chains?

It is a bit challenging to understand the call stack hierarchy.

cpp/src/structure/graph_view_mg_v32_e32.cu
```diff
@@ -1,7 +1,12 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
+
+// Built into libcugraph_common.so: sampling common TUs (gather_one_hop / sample_outgoing_edges)
```
seunghwak: This seems to describe a build/linkage requirement rather than the implementation itself. Would it be better to keep this explanation in CMakeLists.txt next to where these sources are added to CUGRAPH_COMMON_*?

seunghwak (commented):
Additional reviews.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Last-N for the current use case is based on timestamps, but we already use timestamps to create biases, so we can easily implement with your new primitive.  I think having it based on biases is a more flexible design anyway.

The LAST enum value was already defined, leaving this in place as deprecated to allow this to be a non-breaking API change.  If a user previously called with LAST it would be valid, although it would have fast-failed as not implemented.

I'll change this so that it is not the same as FIXED_WINDOW (that's probably not correct) and instead returns a not implemented but deprecated error message.  We'll actually delete the value in the next release.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Yes, that might be a cleaner design.  I'll look at that.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Yes, I think combining them would make sense, let me look at that.

cpp/include/cugraph/sampling_functions.hpp
ChuckHastings: Will look at this

cpp/src/c_api/neighbor_sampling.cpp
ChuckHastings: This all should go through cugraph_neighbor_sample, I'll fix that.  I tried a few things and ended up with this model, missed getting rid of some of the run_neighbor_sample() calls.

The intention is to remove the old interfaces and have everything call cugraph_neighbor_sample, but I wanted to marek them as deprecated and leave the old methods in place for a release before we delete them.

cpp/src/structure/graph_view_mg_v32_e32.cu
ChuckHastings: I'll move that.

cpp/include/cugraph/sampling_functions.hpp
seunghwak: It seems like the documentation still needs updates.

cpp/src/c_api/neighbor_sample.cpp
```diff
@@ -0,0 +1,955 @@
...
+            break;
+          case cugraph_temporal_sampling_comparison_t::FIXED_WINDOW:
+            temporal_sampling_comparison = cugraph::temporal_sampling_comparison_t::FIXED_WINDOW;
+            break;
+          default: CUGRAPH_FAIL("Invalid temporal sampling comparison type");
+        }
+      }
+
+      // FIXME: For biased sampling, the user should pass either biases or edge weights,
+      // otherwised throw an error and suggest the user to call uniform neighbor sample instead
```
seunghwak: Any reason to delay addressing this?

cpp/src/c_api/neighbor_sample.cpp
```diff
@@ -0,0 +1,955 @@
...
+               CUGRAPH_INVALID_INPUT,
+               "sampling_options must not be NULL",
+               *error);
+  CAPI_EXPECTS(
+    num_edge_types >= 1, CUGRAPH_INVALID_INPUT, "num_edge_types must be at least 1", *error);
+
+  auto options =
+    *reinterpret_cast<cugraph::c_api::cugraph_sampling_options_t const*>(sampling_options);
+
+  auto const is_biased   = edge_biases != nullptr;
```
alexbarghi-nv: What happens when we're using the legacy functions (i.e. `homogeneous_biased_temporal_neighbor_sample`) and don't have a way to pass edge biases from PLC?  It seems like it will just fall back to uniform neighbor sampling which we don't want.  This PR needs to expose a way to pass the biases from PLC, or at least to preserve the old behavior while we work that out in a separate PR.

alexbarghi-nv (changes requested):

cpp/src/sampling/detail/sample_edges_impl.cuh
```diff
@@ -189,18 +190,32 @@ namespace {
...
+    if (fixed_window) { return a; }
     auto const time_a = cuda::std::get<0>(a);
     auto const time_b = cuda::std::get<0>(b);
     if (increasing) { return time_a >= time_b ? a : b; }
     return time_a <= time_b ? a : b;
   }
 };
 
+// FIXED_WINDOW analogue of thrust::maximum / thrust::minimum used to dedupe scalar side-table
+// times: duplicate entries for the same key are expected to be identical, so keep the first.
```
seunghwak, line 210: Minor documentation issue but we are migrating from thrust::minimum/maximum to cuda::minimum/maximum.

cpp/src/sampling/detail/sample_edges_impl.cuh
```diff
@@ -252,7 +268,19 @@ void dedupe_sorted_temporal_mg_side_table(
...
-    if (increasing) {
+    if (fixed_window) {
+      auto ends = thrust::reduce_by_key(
+        handle.get_thrust_policy(),
+        thrust::make_zip_iterator(sorted_majors.begin(), sorted_labels->begin()),
+        thrust::make_zip_iterator(sorted_majors.end(), sorted_labels->end()),
+        sorted_times.begin(),
+        thrust::make_zip_iterator(out_majors.begin(), out_labels.begin()),
+        out_times.begin(),
+        thrust::equal_to<cuda::std::tuple<vertex_t, int32_t>>{},
```
seunghwak: cuda::std::equal_to.

cpp/src/sampling/detail/sample_edges_impl.cuh
```diff
@@ -312,7 +340,17 @@ void dedupe_sorted_temporal_mg_side_table(
...
     size_t new_size{};
-    if (increasing) {
+    if (fixed_window) {
+      auto ends = thrust::reduce_by_key(handle.get_thrust_policy(),
+                                        sorted_majors.begin(),
+                                        sorted_majors.end(),
+                                        sorted_times.begin(),
+                                        out_majors.begin(),
+                                        out_times.begin(),
+                                        thrust::equal_to<vertex_t>{},
```
seunghwak: cuda::std::equal_to

cpp/src/structure/graph_view_mg_v32_e32.cu
seunghwak: Sounds great.

seunghwak (approved):
Comments about several minor cosmetic issues. Otherwise, LGTM.

cpp/src/c_api/neighbor_sample.cpp
ChuckHastings: Good catch.  I forgot that we don't yet have a good way of creating a biases array except as the edge weights when we create the graph.  I'll add that back.

cpp/src/c_api/neighbor_sample.cpp
ChuckHastings: Fixed.  The comment was mostly obsolete, but added another CUGRAPH_EXPECTS to catch the edge case that was still wrong.

cpp/src/structure/graph_view_mg_v32_e32.cu
ChuckHastings: Done

cpp/src/c_api/neighbor_sample.cpp
ChuckHastings: Added back, I also added the new function (cugraph_neighbor_sample) into the PLC layer.  We should plan on migrating to that, I marked the old functions as deprecated.

jnke2016 (approved):

alexbarghi-nv (approved):

PR #5629, opened 2026-08-15
Fix bug found in how we handle empty frontiers in sampling

ChuckHastings:
The sampling primitives don't handle empty frontiers cleanly.

Added checks in the primitives as well as a short-circuit in the sampling loops to avoid these situations.


seunghwak (approved):
LGTM

alexbarghi-nv (approved):

PR #5636, opened 2026-08-24
Last-n implementation using the new primitive

ChuckHastings:
Had to make a minor change to the API, last-n needs to be orthogonal to the direction (increasing/decreasing) in order to support first-n (which is basically free if we keep the parameters orthogonal).  I have marked this as non-breaking in spite of this change since we only defined the change within this release, so the only potential breakage would be developers using our nightlies.

Implements last-n using the new primitive.  For int64_t times, the result might be approximate if edge times exceed 2**53 due to the conversion of the time to a double.  If this becomes problematic we can explore alternative options.

### DIFF
--- a/cpp/include/cugraph/sampling_functions.hpp
+++ b/cpp/include/cugraph/sampling_functions.hpp
@@ -60,8 +60,13 @@ enum class neighbor_selection_t {
                   times for STRICTLY_INCREASING and MONOTONICALLY_INCREASING (same as PyG);
                   earlier times for STRICTLY_DECREASING and MONOTONICALLY_DECREASING (the
                   last elements in decreasing time order). Does not accept edge biases or
-                  with_replacement. Equal timestamps (and 64-bit times that are not uniquely
-                  representable as double) may tie in arbitrary order. */
+                  with_replacement. Start times are ranked via double conversion: int32
+                  ranks exactly over the full signed range; int64 ranks exactly on
+                  [-2^52, 2^52 - 1] (typical unix seconds/milliseconds/microseconds).
+                  Beyond that, ordering is preserved but values may tie, so fanout edges
+                  are still returned yet the chosen set may be wrong when more than fanout
+                  K eligible edges on one source share the same rank. Equal timestamps
+                  always tie. */
 };
 
 /**
@@ -111,6 +116,9 @@ struct sampling_options_t {
    * LAST (PyG temporal_strategy='last') keeps fanout-K edges ranked by start time
    * along the walk implied by @p temporal_sampling_comparison: later times for
    * increasing modes, earlier times (last in decreasing order) for decreasing modes.
+   * int32 start times rank exactly over the full signed range; int64 ranks exactly
+   * on [-2^52, 2^52 - 1]. Outside that int64 range, ordering is preserved but ties
+   * are possible (see neighbor_selection_t::LAST).
    */
   neighbor_selection_t neighbor_selection{neighbor_selection_t::RANDOM};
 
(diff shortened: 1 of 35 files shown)


cpp/include/cugraph/sampling_functions.hpp
```diff
@@ -52,8 +51,15 @@ enum class temporal_sampling_comparison_t {
...
-          temporal_sampling_comparison. Not yet implemented. */
+  LAST        /** Deterministically select the last-n eligible edges (temporal only).
+                  Eligible edges are those that pass the temporal window and disjoint
+                  unvisited-destination filter. Among them, keep fanout K (per edge type
+                  when heterogeneous) ranked by edge start time: later times for
+                  STRICTLY_INCREASING and MONOTONICALLY_INCREASING; earlier
+                  times for STRICTLY_DECREASING and MONOTONICALLY_DECREASING. Does not
+                  accept edge biases or with_replacement. Equal timestamps (and 64-bit
+                  times that are not uniquely representable as double) may tie in
+                  arbitrary order. */
```
seunghwak: I am wondering `LAST` here is a right terminology or not.

Without any background and just encountering `LAST`, I will conjecture that we will pick the neighbors of last (largest, least recent, latest) time stamps.

But `LAST` here can mean the opposite based on the value of the `temporal_sampling_comparison_t` parameter.

We may need a better name, something like `TOP_K`, `TOP_N`, `DETERMINISTIC`, or `ORDERED` UNLESS `LAST` (or `LAST_N`) historically means picking the top N ones with the best scores, so people in the field won't get confused.

ChuckHastings: I [technically the agent I asked] checked PyG, TGN, and DGL. There isn’t a single enum-name standard, but the closest ecosystem precedent is PyG’s temporal_strategy='last' (vs 'uniform'), added in [pyg-team/pytorch_geometric#5576](https://github.com/pyg-team/pytorch_geometric/pull/5576) (originally last_k). TGN uses --uniform as the inverse (default = most-recent), and DGL TGN examples use sampler_type='topk'. Literature calls it “most-recent sampling.” Names like TOP_K/DETERMINISTIC don’t appear as strategy enums in these libraries.

We used LAST to align with PyG and our legacy temporal_sampling_comparison_t::LAST. The intended meaning is “last-n along the temporal walk” (fanout-K edges ranked by start time), not unconditionally “latest timestamp.”

That said, cuGraph supports decreasing walks also, so the rank direction depends on temporal_sampling_comparison — an ambiguity PyG doesn’t expose. I’m happy to (a) keep LAST and make the docs explicit upfront, or (b) rename to LAST_N to signal “n neighbors” without implying a single time direction.

cpp/src/sampling/detail/sample_outgoing_edges_impl.cuh
```diff
@@ -705,13 +852,19 @@ rmm::device_uvector<T> gather_scalar_edge_property_for_edgelist(
...
+                   (std::holds_alternative<rmm::device_uvector<edge_t>>(multi_edge_index))
+                     ? std::make_optional([&]() {
+                         CUGRAPH_EXPECTS(
+                           std::holds_alternative<rmm::device_uvector<edge_t>>(multi_edge_index),
+                           "gather_scalar_edge_property_for_edgelist: multi_edge_index expected "
+                           "rmm::device_uvector<edge_t>, variant index %zu",
+                           multi_edge_index.index());
+                         return std::get<rmm::device_uvector<edge_t>>(multi_edge_index).begin();
+                       }())
+                     : std::nullopt);
```
seunghwak: Can you call the edge_bucket constructor taking R-values? (https://github.com/rapidsai/cugraph/blob/main/cpp/include/cugraph/prims/edge_bucket.cuh#L67)

Then, we don't need to pay the cost (memory/compute) of insert.

seunghwak (approved):
LGTM (one suggestions to improve performance)

jnke2016 (approved):

cpp/src/sampling/detail/sample_outgoing_edges_impl.cuh
ChuckHastings: Updated the code to do that.  I added a new method to the edge bucket to take back ownership of the device vectors.

I wonder if we would be better mimicking the key_bucket_view_t paradigm we use for the vertex buckets and create an edge_bucket_view_t that could just take the spans.  We do have read-only use cases like this as well as read-write use cases.

alexbarghi-nv (approved):
Local integration tests are now passing; thanks @ChuckHastings!

PR #5637, opened 2026-08-25
Don't use the r-value reference constructor for rmm::device_scalar

ChuckHastings:
https://github.com/rapidsai/rmm/pull/2527 eliminated the r-value reference constructors for rmm::device_scalar, since it did asynchronous updates and could have issues if the r-value went out of scope.

This PR removes all references to that constructor.

### DIFF
--- a/cpp/include/cugraph/prims/vertex_frontier.cuh
+++ b/cpp/include/cugraph/prims/vertex_frontier.cuh
@@ -187,9 +187,7 @@ void device_bcast_vertex_list(
     assert((comm.get_rank() != root) || (std::get<0>(v_list).size() == tmp_bitmap.size()));
     device_bcast(
       comm, std::get<0>(v_list).data(), tmp_bitmap.data(), tmp_bitmap.size(), root, stream_view);
-    size_t initial_dummy{0};
-    rmm::device_scalar<size_t> dummy(initial_dummy, stream_view);  // we already know the count
-    stream_view.synchronize();
+    rmm::device_scalar<size_t> dummy(stream_view);  // we already know the count
     detail::copy_if_nosync(
       thrust::make_counting_iterator(vertex_range_first),
       thrust::make_counting_iterator(vertex_range_last),
--- a/cpp/include/cugraph/prims/vertex_frontier.cuh
+++ b/cpp/include/cugraph/prims/vertex_frontier.cuh
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
@@ -187,7 +187,9 @@ void device_bcast_vertex_list(
     assert((comm.get_rank() != root) || (std::get<0>(v_list).size() == tmp_bitmap.size()));
     device_bcast(
       comm, std::get<0>(v_list).data(), tmp_bitmap.data(), tmp_bitmap.size(), root, stream_view);
-    rmm::device_scalar<size_t> dummy(size_t{0}, stream_view);  // we already know the count
+    size_t initial_dummy{0};
+    rmm::device_scalar<size_t> dummy(initial_dummy, stream_view);  // we already know the count
+    stream_view.synchronize();
     detail::copy_if_nosync(
       thrust::make_counting_iterator(vertex_range_first),
       thrust::make_counting_iterator(vertex_range_last),
--- a/cpp/tests/traversal/bfs_test.cpp
+++ b/cpp/tests/traversal/bfs_test.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -138,7 +138,8 @@ class Tests_BFS : public ::testing::TestWithParam<std::tuple<BFS_Usecase, input_
       hr_timer.start("BFS");
     }
 
-    rmm::device_scalar<vertex_t> const d_source(bfs_usecase.source, handle.get_stream());
+    vertex_t const source{static_cast<vertex_t>(bfs_usecase.source)};
+    rmm::device_scalar<vertex_t> const d_source(source, handle.get_stream());
 
     cugraph::bfs(handle,
                  graph_view,
--- a/cpp/tests/traversal/bfs_test.cpp
+++ b/cpp/tests/traversal/bfs_test.cpp
@@ -140,6 +140,8 @@ class Tests_BFS : public ::testing::TestWithParam<std::tuple<BFS_Usecase, input_
 
     vertex_t const source{static_cast<vertex_t>(bfs_usecase.source)};
     rmm::device_scalar<vertex_t> const d_source(source, handle.get_stream());
+    handle
+      .sync_stream();  // before source goes out-of-scope (async H2D copy from device_scalar ctor)
 
     cugraph::bfs(handle,
                  graph_view,
--- a/cpp/tests/traversal/extract_bfs_paths_test.cu
+++ b/cpp/tests/traversal/extract_bfs_paths_test.cu
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2021-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -104,8 +104,8 @@ class Tests_ExtractBfsPaths
     rmm::device_uvector<vertex_t> d_predecessors(graph_view.number_of_vertices(),
                                                  handle.get_stream());
 
-    rmm::device_scalar<vertex_t> const d_source(extract_bfs_paths_usecase.source,
-                                                handle.get_stream());
+    vertex_t const source{static_cast<vertex_t>(extract_bfs_paths_usecase.source)};
+    rmm::device_scalar<vertex_t> const d_source(source, handle.get_stream());
 
     cugraph::bfs(handle,
                  graph_view,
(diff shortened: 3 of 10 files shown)


cpp/include/cugraph/prims/vertex_frontier.cuh
```diff
@@ -187,7 +187,9 @@ void device_bcast_vertex_list(
     assert((comm.get_rank() != root) || (std::get<0>(v_list).size() == tmp_bitmap.size()));
     device_bcast(
       comm, std::get<0>(v_list).data(), tmp_bitmap.data(), tmp_bitmap.size(), root, stream_view);
-    rmm::device_scalar<size_t> dummy(size_t{0}, stream_view);  // we already know the count
+    size_t initial_dummy{0};
+    rmm::device_scalar<size_t> dummy(initial_dummy, stream_view);  // we already know the count
+    stream_view.synchronize();
```
bdice: You shouldn't need to sync here, since you're overwriting the value on the same stream immediately after. The example I gave you had a stream sync, but you should adapt it to the surrounding code to be properly stream-ordered, and remove or defer the sync as needed.

ChuckHastings: Reviewed all of the sync calls and removed the unnecessary ones.

bdice: Is this one safe to remove? I think it is, but maybe I'm missing something.

cpp/tests/utilities/property_generator_utilities_impl.cuh
```diff
@@ -167,7 +167,9 @@ generate<GraphViewType, property_t>::unique_edge_property(raft::handle_t const&
     CUGRAPH_EXPECTS(
       graph_view.compute_number_of_edges(handle) <= std::numeric_limits<property_t>::max(),
       "std::numeric_limits<property_t>::max() is smaller than the number of edges.");
-    rmm::device_scalar<property_t> counter(property_t{0}, handle.get_stream());
+    property_t initial_counter{0};
+    rmm::device_scalar<property_t> counter(initial_counter, handle.get_stream());
+    handle.sync_stream();
```
bdice: Why is this sync needed? The `transform_e` should be stream-ordered-after the `counter` initialization, so the kernel that does those atomic updates should see the zero value.

cpp/include/cugraph/prims/vertex_frontier.cuh
ChuckHastings: Yes, removed in latest push.

cpp/tests/utilities/property_generator_utilities_impl.cuh
ChuckHastings: Yes, removed this sync as well.

cpp/tests/traversal/mg_sssp_test.cpp
```diff
@@ -129,8 +129,8 @@ class Tests_MGSSSP : public ::testing::TestWithParam<std::tuple<SSSP_Usecase, in
         (*mg_renumber_map).data(),
         mg_graph_view.vertex_partition_range_lasts());
 
-      rmm::device_scalar<vertex_t> d_sg_source(static_cast<vertex_t>(sssp_usecase.source),
-                                               handle_->get_stream());
+      vertex_t sg_source{static_cast<vertex_t>(sssp_usecase.source)};
+      rmm::device_scalar<vertex_t> d_sg_source(sg_source, handle_->get_stream());
```
seunghwak, line 133: Here, which line following this guarantees that `d_sg_source` initialization will finish before `sg_source` goes out-of-scope?

cpp/tests/utilities/property_generator_utilities_impl.cuh
```diff
@@ -167,7 +167,8 @@ generate<GraphViewType, property_t>::unique_edge_property(raft::handle_t const&
     CUGRAPH_EXPECTS(
       graph_view.compute_number_of_edges(handle) <= std::numeric_limits<property_t>::max(),
       "std::numeric_limits<property_t>::max() is smaller than the number of edges.");
-    rmm::device_scalar<property_t> counter(property_t{0}, handle.get_stream());
+    rmm::device_scalar<property_t> counter(handle.get_stream());
+    counter.set_value_to_zero_async(handle.get_stream());
```
seunghwak, line 171: Instead of
```
  edge_t initial_value{0};
  rmm::device_scalar<edge_t> ret(initial_value, handle.get_stream());
```
Should we use this elsewhere as well? What are the trade-offs?

ChuckHastings: Feels like this is more intuitive as to what is happening to me.  I'm fine using this elsewhere, I started with the former based on what was suggested as the simplest solution in a slack thread.  The agent suggested this alternative once the sync was no longer required.  Thoughts?

seunghwak: Is set_value_to_zero_async internally calls cudaMemsetAsync? In that case, it sounds like a better and also a safer solution?

If we define "edge_t initial_value{0}", it will be a mental toll to check that `initial_value` out-lives the synchronization point.

cpp/tests/traversal/mg_sssp_test.cpp
ChuckHastings: Added the sync back in here.

cpp/tests/utilities/property_generator_utilities_impl.cuh
ChuckHastings: Alright, refactored to use this approach when we initialize to 0.

cpp/tests/traversal/mg_sssp_test.cpp
seunghwak: Maybe adding a comment explaining why this sync is necessary (like here https://github.com/rapidsai/cugraph/blob/main/cpp/src/components/weakly_connected_components_impl.cuh#L518) could be helpful. Some people may think this sync is unnecessary.

seunghwak (approved):
LGTM

cpp/tests/traversal/bfs_test.cpp
```diff
@@ -138,7 +138,8 @@ class Tests_BFS : public ::testing::TestWithParam<std::tuple<BFS_Usecase, input_
       hr_timer.start("BFS");
     }
 
-    rmm::device_scalar<vertex_t> const d_source(bfs_usecase.source, handle.get_stream());
+    vertex_t const source{static_cast<vertex_t>(bfs_usecase.source)};
+    rmm::device_scalar<vertex_t> const d_source(source, handle.get_stream());
```
seunghwak, line 142: Debating between whether we need to call `handle.sync_stream()` here or not. 

How do we know `source` will out-live `d_source` initialization? For someone who knows `bfs` internals, yes, we synchronize inside `bfs` but we cannot find that out just reading the `bfs` API (no value is returned to host, so just seeing the API, users may need to assume that `bfs` is asynchronous). It might be a safer practice to call sync_stream() unless we can clearly find that it is unnecessary without diving into implementation internals.

cpp/tests/traversal/extract_bfs_paths_test.cu
```diff
@@ -104,8 +104,8 @@ class Tests_ExtractBfsPaths
     rmm::device_uvector<vertex_t> d_predecessors(graph_view.number_of_vertices(),
                                                  handle.get_stream());
 
-    rmm::device_scalar<vertex_t> const d_source(extract_bfs_paths_usecase.source,
-                                                handle.get_stream());
+    vertex_t const source{static_cast<vertex_t>(extract_bfs_paths_usecase.source)};
+    rmm::device_scalar<vertex_t> const d_source(source, handle.get_stream());
```
seunghwak, line 108: Here,

```
    auto h_distances    = cugraph::test::to_host(handle, d_distances);
    auto h_predecessors = cugraph::test::to_host(handle, d_predecessors);
```

clearly synchronizes, so no need to call sync_stream().

cpp/tests/traversal/mg_bfs_test.cpp
```diff
@@ -102,10 +102,11 @@ class Tests_MGBFS : public ::testing::TestWithParam<std::tuple<BFS_Usecase, inpu
...
 
-    auto d_mg_source = mg_graph_view.in_local_vertex_partition_range_nocheck(bfs_usecase.source)
-                         ? std::make_optional<rmm::device_scalar<vertex_t>>(bfs_usecase.source,
-                                                                            handle_->get_stream())
-                         : std::nullopt;
+    vertex_t const source{static_cast<vertex_t>(bfs_usecase.source)};
+    auto d_mg_source =
+      mg_graph_view.in_local_vertex_partition_range_nocheck(source)
+        ? std::make_optional<rmm::device_scalar<vertex_t>>(source, handle_->get_stream())
+        : std::nullopt;
```
seunghwak, line 109: Similar here, we cannot tell `source` will out-live `d_mg_source` initialization unless someone knows `bfs` internals. Synchronization happens only when (bfs_usecase.check_correctness) is true and for root.

cpp/tests/traversal/mg_extract_bfs_paths_test.cu
```diff
@@ -104,10 +104,10 @@ class Tests_MGExtractBFSPaths
     rmm::device_uvector<vertex_t> d_mg_predecessors(
       mg_graph_view.local_vertex_partition_range_size(), handle_->get_stream());
 
+    vertex_t const source{static_cast<vertex_t>(extract_bfs_paths_usecase.source)};
     auto const d_mg_source =
-      mg_graph_view.in_local_vertex_partition_range_nocheck(extract_bfs_paths_usecase.source)
-        ? std::make_optional<rmm::device_scalar<vertex_t>>(extract_bfs_paths_usecase.source,
-                                                           handle_->get_stream())
+      mg_graph_view.in_local_vertex_partition_range_nocheck(source)
+        ? std::make_optional<rmm::device_scalar<vertex_t>>(source, handle_->get_stream())
```
seunghwak, line 110:
```
    auto h_mg_distances    = cugraph::test::to_host(*handle_, d_mg_distances);
    auto h_mg_predecessors = cugraph::test::to_host(*handle_, d_mg_predecessors);
```

clearly synchronizes, so, not calling sync_stream() is clearly safe.

cpp/tests/traversal/bfs_test.cpp
ChuckHastings: Since all of these are test codes, perhaps we just call `sync_stream()`, since we're not actually worried about performance.

ChuckHastings: Added a sync based on our slack conversation.  Both places where it's not obvious solely from the test code that synchronization has occurred.

cpp/tests/traversal/mg_bfs_test.cpp
ChuckHastings: Added sync here as well.

PR #3129, opened 2023-01-11
[WIP] Memory efficient cugraph  benchmarking

VibhuJawa:

PR #5639, opened 2026-08-28
Migrate stream APIs from rmm::cuda_stream_view to cuda::stream_ref

bdice:
## Summary

Track the coordinated migration of stream APIs and call sites from `rmm::cuda_stream_view` to CCCL's `cuda::stream_ref`. This propagates `cuda::stream_ref` through RMM containers and memory resources, RAFT resource and handle APIs, downstream C++ interfaces, Python/Cython bindings, benchmarks, tests, and documentation.

This migrates affected cuGraph API signatures and call sites, including the MTMG handle stream accessor, while preserving `cuda::stream_ref` until a raw stream is required.

Depends on rapidsai/rmm#2372 and NVIDIA/raft#3129.

Tracked in rapidsai/build-planning#318.

## Migrations

- Pass `cuda::stream_ref` through stream pools, resource accessors, conditionals, and downstream APIs without converting to `rmm::cuda_stream_view`
- Use `cuda::stream_ref` constructions for default/legacy/per-thread streams
  - `rmm::cuda_stream_default` ➡️ `cuda::stream_ref{cudaStream_t{cudaStreamDefault}}`
  - `rmm::cuda_stream_legacy` ➡️ `cuda::stream_ref{cudaStreamLegacy}`
  - `rmm::cuda_stream_per_thread` ➡️ `cuda::stream_ref{cudaStreamPerThread}`
- Use `.get()` when calling an API that requires a raw `cudaStream_t`, including CUDA runtime, library, CUB, and legacy API boundaries (previously `rmm::cuda_stream_view` used `value()`)
- Use `.sync()` when synchronizing a `cuda::stream_ref` (previously `rmm::cuda_stream_view` used `synchronize()`)
- Update Cython declarations and call sites to pass stream references directly where supported


ChuckHastings (approved):
LGTM

cpp/include/cugraph/mtmg/handle.hpp
```diff
@@ -51,19 +53,19 @@ class handle_t {
    *
    * @return cuda stream
    */
-  rmm::cuda_stream_view get_stream() const
+  cuda::stream_ref get_stream() const
   {
     return raft_handle_.is_stream_pool_initialized()
              ? raft_handle_.get_stream_from_stream_pool(thread_rank_)
-             : raft_handle_.get_stream();
+             : static_cast<cuda::stream_ref>(raft_handle_.get_stream());
```
seunghwak: This is a bit confusing to me.

`raft_handle_.get_stream_from_stream_pool()` returns `cuda::stream_ref` but `raft_handle_.get_stream()` returns something else and needs to be type-casted? Better return the same type?

cpp/include/cugraph/prims/update_edge_src_dst_property.cuh
```diff
@@ -707,11 +707,12 @@ void update_edge_minor_property(
...
 
-    auto local_v_list_sizes = host_scalar_allgather(major_comm, v_list_size, handle.get_stream());
+    auto local_v_list_sizes =
+      host_scalar_allgather(major_comm, v_list_size, handle.get_stream().get());
     auto local_v_list_range_firsts =
-      host_scalar_allgather(major_comm, v_list_range[0], handle.get_stream());
+      host_scalar_allgather(major_comm, v_list_range[0], handle.get_stream().get());
     auto local_v_list_range_lasts =
-      host_scalar_allgather(major_comm, v_list_range[1], handle.get_stream());
+      host_scalar_allgather(major_comm, v_list_range[1], handle.get_stream().get());
```
seunghwak: To be consistent with the rest of the API (use cuda::stream_ref as much as possible till raw cuda stream is absolutely necessary), should we better update `host_scalar_allgather` and other host scalar utility functions to take cuda::stream_ref and get the raw stream right before calling the NCCL functions in raft?

cpp/include/cugraph/utilities/groupby_and_count.cuh
```diff
@@ -445,7 +446,7 @@ void mem_frugal_groupby(
   int num_groups,
   size_t mem_frugal_threshold,  // take the memory frugal approach (instead of thrust::sort) if #
                                 // elements to groupby is no smaller than this value
-  rmm::cuda_stream_view stream_view,
+  cuda::stream_ref stream_view,
```
seunghwak, line 449: We have been mixing `stream_view` and `stream`. `stream_view` now sounds like a misnomer. Better use `stream` consistently?

cpp/include/cugraph/utilities/shuffle_comm.cuh
```diff
@@ -74,9 +75,9 @@ compute_tx_rx_counts_displs_ranks(raft::comms::comms_t const& comm,
   rmm::device_uvector<size_t> d_rx_value_counts(comm_size, stream_view);
   device_alltoall(comm, d_tx_value_counts.data(), d_rx_value_counts.data(), size_t{1}, stream_view);
 
-  raft::update_host(tx_counts.data(), d_tx_value_counts.data(), comm_size, stream_view.value());
-  raft::update_host(rx_counts.data(), d_rx_value_counts.data(), comm_size, stream_view.value());
-  stream_view.synchronize();
+  raft::update_host(tx_counts.data(), d_tx_value_counts.data(), comm_size, stream_view.get());
+  raft::update_host(rx_counts.data(), d_rx_value_counts.data(), comm_size, stream_view.get());
```
seunghwak: Should we better update raft::update_host to take cuda::stream_ref?

cpp/include/cugraph/prims/vertex_frontier.cuh
```diff
@@ -112,7 +113,7 @@ std::vector<size_t> compute_key_segment_offsets(KeyIterator sorted_key_first,
 
   std::vector<size_t> h_offsets(d_offsets.size() + 2);
   raft::update_host(h_offsets.data() + 1, d_offsets.data(), d_offsets.size(), stream_view);
-  RAFT_CUDA_TRY(cudaStreamSynchronize(stream_view));
+  RAFT_CUDA_TRY(cudaStreamSynchronize(stream_view.get()));
```
seunghwak: Instead of directly calling `cudaStreamSynchronize()`, should we better call `stream_view.sync()`? I assume the philosophy here is to avoid using raw CUDA stream unless it is absolutely necessary?

cpp/include/cugraph/utilities/shuffle_comm.cuh
```diff
@@ -296,7 +297,7 @@ auto shuffle_values(
                     stream_view);
   raft::update_host(
     rx_aligned_counts.data(), d_rx_aligned_counts.data(), d_rx_aligned_counts.size(), stream_view);
-  RAFT_CUDA_TRY(cudaStreamSynchronize(stream_view));
+  RAFT_CUDA_TRY(cudaStreamSynchronize(stream_view.get()));
```
seunghwak: Won't `stream_view.sync()` be better?

bdice:
Rebased onto preparation PR #5648 and validated against the rebased RMM/RAFT stack. Local `build-cugraph -j0` completed successfully, and `STREAM_TEST` passed. CI is rerunning at `3c519b896`.

bdice:
CI follow-up: both pip devcontainer failures are environmental, before compiling cuGraph. cuVS configuration fails because `cuda.tile` is absent from the Python 3.14 pip environment (`ModuleNotFoundError: No module named 'cuda.tile'`) on both amd64 and arm64. The full migration stack has already been validated locally with the cuTile/NVVM environment documented above.

bdice:
The conda devcontainer failures are dependency-order failures against pre-migration RAFT: `handle.get_stream()` is still `rmm::cuda_stream_view`, making conditionals with the migrated `cuda::stream_ref` branch ambiguous. With RAFT #3129 in the local stack, stream references are preserved on both branches and the `-j0` cuGraph build passes.

cpp/include/cugraph/mtmg/handle.hpp
bdice: Agreed. Removed the casts in 1ea1c7267; with the migrated RAFT API both accessors return `cuda::stream_ref`.

cpp/include/cugraph/prims/update_edge_src_dst_property.cuh
bdice: Agreed. 406f17cc2 migrates all host-scalar utility signatures to `cuda::stream_ref`, forwards stream references from callers, and unwraps only at the RAFT comms boundary.

cpp/include/cugraph/utilities/groupby_and_count.cuh
bdice: `cuda::stream_ref` is still a non-owning stream view, and `stream_view` remains common across this codebase (159 declarations). I retained existing parameter names to keep this API migration from becoming a broad naming-only change.

cpp/include/cugraph/utilities/shuffle_comm.cuh
bdice: Agreed. RAFT `update_host` is migrated to `cuda::stream_ref`; 406f17cc2 now forwards `stream_view` directly here.

cpp/include/cugraph/prims/vertex_frontier.cuh
bdice: Agreed. Replaced the raw runtime synchronization with `stream_view.sync()` in 1ea1c7267.

cpp/include/cugraph/utilities/shuffle_comm.cuh
bdice: Agreed. Replaced the raw runtime synchronization with `stream_view.sync()` in 1ea1c7267.

bdice:
Stack update: the three `.sync()` compatibility changes now live in preparation PR #5648 at a9d3409cb. This migration was rebased onto that head and is now b8ee95bdb; its source tree exactly matches the previously validated migration tree.

seunghwak (approved):
LGTM

PR #5640, opened 2026-08-28
Add synchronization to coo_to_csr

ChuckHastings:
This legacy code is only used in MST and in some legacy python code.  Performance is not critical.

We could add a handle so that we can all use the same stream, but that has breaking API implications.  Simplest is to just add a synchronization call before leaving the function.


bdice (approved):
Do you plan to add a parameter (stream/handle/etc.) so that this function can be properly stream-ordered?

seunghwak (approved):
LGTM

ChuckHastings:
> Do you plan to add a parameter (stream/handle/etc.) so that this function can be properly stream-ordered?

There's an update to MST in the works in raft (https://github.com/NVIDIA/raft/pull/3118).  Once that's complete we'll be doing some testing within cugraph.  I was going to use that as an opportunity to update some of the MST code in cugraph.

PR #5642, opened 2026-09-03
Initial draft of making cugraph a mono repo

BradReesWork:
The goal is to make the "cugraph" repo a monorepo for all graph work.

This first PR focuses just on the what is currently in the cugraph repo; cugraph, pylibcugraph, libcugraph (which includes libcugraph_c)

The goal is for the CI process to only run against the module related to the changed files.   This should improve CI run times.

When a file is changed, only the effected module is built and tested.  The other modules are downloaded and used for testing but are not rebuilt.


BradReesWork:
Needs to be updated after #5653 is merged

Issue #2462, opened 2022-07-28
[FEA] Support for simple_cycles


beckernick:
**Describe the solution you'd like and any additional context**

I'd like to be able to use [simple_cycles](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.cycles.simple_cycles.html) like I can with NetworkX.

```python
import networkx as nx
​
edges = [(0, 0), (0, 1), (0, 2), (1, 2), (2, 0), (2, 1), (2, 2)]
G = nx.DiGraph(edges)
sorted(nx.simple_cycles(G))
[[0], [0, 1, 2], [0, 2], [1, 2], [2]]
```

---
_For Developers below this line_

  - [] Code passes CI
  - [] Code uses Graph Primitives
  - [] Code in C++/CUDA layer
  - [] Code in C layer
  - [] Code in pylibcugraph Python layer
  - [] Code in cugraph Python layer
  - [] Documentation
  - [] Test cases at each layer (for MG tests, a note in the PR description indicating the new/existing MG tests were run and passed)

acostadon:
We currently do not support that functionality but it is on our to-do list.  Unfortunately it will not be until mid/late next year.

sauravsingla:
Would it be possible to prioritize GPU-accelerated support for bounded simple_cycles, particularly for workflows that already use cuGraph for graph construction and BFS/hop computation from a seed node?

In our current pipeline, graph construction and hop computation remain on GPU with cuGraph, but cycle detection requires falling back to NetworkX. On large graphs, this introduces significant CPU overhead and GPU-to-CPU conversion cost, reducing the benefit of keeping the rest of the workflow GPU-native.

A particularly useful first scope could be support equivalent to:

nx.simple_cycles(G, length_bound=k)

with an optional optimization for cycles within a seed vertex or N-hop neighborhood. Since unrestricted simple-cycle enumeration can be extremely expensive, bounded cycles may also be a more practical initial GPU implementation.

This would help enable end-to-end GPU graph traversal and cycle analysis and could eventually support nx-cugraph dispatch for bounded simple_cycles.

sauravsingla:
[...]

PR #5644, opened 2026-09-03
C++ simple_cycles 

seunghwak:
To address Issue #2462

This PR implements simple_cycles for single-GPU and multi-GPU.


cpp/src/components/simple_cycles_impl.cuh
```diff
@@ -0,0 +1,1387 @@
...
+  graph_view_t<vertex_t, edge_t, false, multi_gpu> const& graph_view,
+  std::optional<raft::device_span<vertex_t const>> seed_vertices,
+  vertex_t length_bound,
+  bool do_expensive_check)
+{
+  auto [cycle_vertices, cycle_lengths] =
+    detail::simple_cycles_impl(handle, graph_view, seed_vertices, length_bound, do_expensive_check);
+  rmm::device_uvector<size_t> cycle_offsets(cycle_lengths.size() + 1, handle.get_stream());
+  cycle_offsets.set_element_to_zero_async(size_t{0}, handle.get_stream());
+  thrust::inclusive_scan(
```
ChuckHastings: Should probably call cugraph::inclusive_scan, which will fall through to thrust::inclusive_scan since we don't have a specialization for transform_iterator.  But if we ever needed to add that we could.

This idiom is reused in other algorithms (creating offsets from sizes).  Would it make sense to create a helper function for this?

cpp/src/components/simple_cycles_impl.cuh
```diff
@@ -0,0 +1,1387 @@
...
+        std::forward_as_tuple(std::tie(nbrs, nbr_path_idxs), std::ignore) =
+          groupby_gpu_id_and_shuffle_values(
+            handle.get_comms(),
+            pair_first,
+            pair_first + nbrs.size(),
+            cuda::proclaim_return_type<int>(
+              [lasts = raft::device_span<size_t const>(
+                 d_path_idx_lasts.data(), d_path_idx_lasts.size())] __device__(auto pair) {
+                auto path_idx = cuda::std::get<1>(pair);
+                return static_cast<int>(cuda::std::distance(
```
ChuckHastings, line 568: Consider using compute_gpu_id_t from k_hop_nbrs_impl.cuh.  We could pull that out as a common functor to reuse here.  If we changed this to groupby_gpu_id_and_shuffle_kv_pairs I think you could just use the functor directly rather than needing a device lambda.

ChuckHastings (approved):
A couple of suggestions.  Otherwise looks good.

cpp/src/components/simple_cycles_impl.cuh
seunghwak: Now this file calls cugraph::inclusive_scan. For refactoring, we can do this in a separate PR if necessary. This will touch multiple files and it will clutter this PR.

cpp/src/components/simple_cycles_impl.cuh
seunghwak: One difference is that we are retrieving an index from a tuple ("auto path_idx = cuda::std::get<1>(pair);"); so we can't use compute_gpu_id_t from k_hop_nbrs as is. Yes, k_hop_nbrs functor can be cugraph::detail::segment_id_t (which already exists). This is a bit different. cuda::std::get first then segment_id_t. Sounds a bit too specific to be a general functor.

ChuckHastings: I think if you used groupby_gpu_id_and_shuffle_kv_pairs, you could use nbr_path_idxs as the key and nbrs as the value.  Then you could use the compute_gpu_id_t from k_hob_nbrs_impl.cuh, I think.

seunghwak: Nice catch, updated both k_hop_nbrs and simple_cycles to use segment_idx_t.

seunghwak:
/ok to test https://github.com/rapidsai/cugraph/pull/5644/commits/753d30ab0abce6679c721f11f7a94161a0862533

seunghwak:
/ok to test [d2dd5fc](https://github.com/rapidsai/cugraph/pull/5644/commits/d2dd5fc3844c291b914b4a33b7e75ee6f9ed846d)

seunghwak:
/ok to test https://github.com/rapidsai/cugraph/pull/5644/commits/19116bc56534fa441ed211b9bcfcb12a2672d1fe

jnke2016 (approved):

PR #5646, opened 2026-09-04
wheels: fix handling of cuda-tile dependency

ChuckHastings:
*(heavily edited by @jameslamb on Sep 8, 2026)*

Contributes to https://github.com/rapidsai/build-planning/issues/324

Wheel builds have been failing like this:

```text
  -- Found Python: /pyenv/versions/3.14.7/bin/python (found version "3.14.7") found components: Interpreter
  CMake Error at build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/cmake/modules/generate_cutile_kernels.cmake:61 (message):
    cuda.tile (cuTile Python) is required to build cuTile embedded kernels.
    Install cutile-python and cuda-tileiras (conda), or cuda-tile[tileiras]
    (pip).

    Interpreter: /pyenv/versions/3.14.7/bin/python

    Import error: Traceback (most recent call last):

        import cuda.tile

    ModuleNotFoundError: No module named 'cuda'
  Call Stack (most recent call first):
    build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/cmake/modules/generate_cutile_kernels.cmake:229 (_cutile_kernels_setup)
    build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/CMakeLists.txt:1161 (generate_cutile_kernels)


  -- Configuring incomplete, errors occurred!
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/34192266298/job/102086734598)

Because:

1. cuVS now needs `cuda-tile` (Python library) at build time, since https://github.com/NVIDIA/cuvs/pull/2548
2. `libcugraph` wheels build cuVS from source

This PR fixes that by adding a build-time dependency on `cuda-tile` for `libcugraph`.

### Why no conda changes?

`libcugraph` conda builds do not compile cuVS from source, they link against pre-built packages.


bdice (approved):

dependencies.yaml
```diff
@@ -317,6 +318,42 @@ dependencies:
...
+      - output_types: conda
+        matrices:
+          - matrix:
+              cuda: "12.*"
+            packages:
+          - matrix:
+              cuda: "13.3"
+            packages:
+              - cutile-python
+              - cuda-tileiras
```
bdice: This can probably be covered by the `13.*` below?
```suggestion
```

dependencies.yaml
```diff
@@ -13,6 +13,7 @@ files:
       - common_build
       - cpp_build
       - cuda
+      - cutile_python
```
jameslamb: > _corrects the broken nightly builds._

I went to the build history to figure out what that means specifically. It looks like the failure in CI is CUDA 13 `libcugraph` wheel builds

```text
  -- Found Python: /pyenv/versions/3.14.7/bin/python (found version "3.14.7") found components: Interpreter
  CMake Error at build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/cmake/modules/generate_cutile_kernels.cmake:61 (message):
    cuda.tile (cuTile Python) is required to build cuTile embedded kernels.
    Install cutile-python and cuda-tileiras (conda), or cuda-tile[tileiras]
    (pip).

    Interpreter: /pyenv/versions/3.14.7/bin/python

    Import error: Traceback (most recent call last):

        import cuda.tile

    ModuleNotFoundError: No module named 'cuda'
  Call Stack (most recent call first):
    build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/cmake/modules/generate_cutile_kernels.cmake:229 (_cutile_kernels_setup)
    build/py3-none-linux_x86_64/_deps/cuvs-src/cpp/CMakeLists.txt:1161 (generate_cutile_kernels)


  -- Configuring incomplete, errors occurred!
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/34192266298/job/102086734598)

Changing just the `all:` group in `dependencies.yaml` wouldn't fix that, as those wheel builds are built without isolation using the dependencies defined in the `py_build_libcugraph` and `py_rapids_build_libcugraph` groups: https://github.com/rapidsai/cugraph/blob/cb14128cb6743706062d84c0e3766b17b2da5e43/ci/build_wheel_libcugraph.sh#L15-L26

Short-term, if these wheels are needed at build time they need to be in the wheel-building dependency set.

Longer-term, it'd be great to switch `libcugraph` to building against `libcuvs` wheels to not need to keep making changes like this just for cuVS.

jameslamb (commented):
Started a thread with my comments from our offline conversation, so it's here in the GitHub history.

dependencies.yaml
jameslamb: I will push a fix here.

msarahan:
This is currently a blocker on the 26.10 release cycle. Please prioritize getting this merged @ChuckHastings. Thanks for your work on this, @jameslamb

ChuckHastings:
> This is currently a blocker on the 26.10 release cycle. Please prioritize getting this merged @ChuckHastings. Thanks for your work on this, @jameslamb

I think this is being superseded by https://github.com/rapidsai/cugraph/pull/5654.  But in either case, it is blocking all PRs planned for 26.10, so it's our highest priority to get this problem resolved.

jameslamb:
This is superseded by #5654 and #5656

PR #5653, opened 2026-09-08
CI: Use isolated build for all wheel packages

robertmaynard:
Replace pip wheel with isolated python -m build --wheel in the shared wheel-build helper used by every wheel build job


BradReesWork:
I was doing the same thing in my darft PR (#5642) but I'll defer to your PR and then update

vyasr (approved):
This repo is a good case where I think combining cugraph and pylibcugraph builds into the same CI job would likely be fruitful.

PR #5654, opened 2026-09-08
Use libcuvs wheels in wheel and pip-devcontainer builds

bdice:
## Summary

- add libcuvs to the libcugraph wheel build requirements
- discover only the libcuvs CMake package prefix exported by the installed wheel during direct C++ builds, so pip devcontainers use the installed libcuvs wheel instead of cloning cuVS
- remove direct NVRTC and nvJitLink dependencies that were added for the old source-built/static cuVS integration; libcuvs now owns those requirements

Part of https://github.com/rapidsai/build-planning/issues/325.

## Root causes

There were two independent gaps after #5499 switched libcugraph to dynamic libcuvs linkage:

1. The libcugraph wheel job installs only the py_build_libcugraph and py_rapids_build_libcugraph dependency sets before running pip wheel --no-deps. libcuvs was a runtime dependency but not a build requirement, so it was absent and rapids_cpm_find fell back to a source build.
2. Pip devcontainers already installed libcuvs-cu12/libcuvs-cu13, but build.sh invokes CMake directly. Unlike scikit-build-core, direct CMake did not consume the wheel cmake.prefix entry point, so it still cloned cuVS into _deps/cuvs-src.

The second gap is fixed with the RAPIDS CMake wheel-prefix discovery helper, filtered to libcuvs. RMM and RAFT must remain source-backed in this direct C++ build so the generated cugraph build-tree package carries complete transitive dependency targets for downstream pylibcugraph configuration.

## Dependency audit

- #5483 added cuda-nvrtc-dev because source-built cuVS compiled a file including nvrtc.h.
- #5479 and #5457 added explicit nvJitLink build/runtime requirements for source-built/static cuVS.
- Those direct requirements are no longer needed with dynamic linkage to the prebuilt libcuvs package. The release/26.10 libcuvs conda and wheel metadata already owns NVRTC and nvJitLink.
- The cutile additions proposed in #5646 are not needed by cuGraph once it stops building cuVS.

I also verified NVIDIA/cuML has the same pip-devcontainer issue: its current CUDA 13.3 pip job installs libcuvs-cu13, then logs CPM: Adding package cuvs and uses _deps/cuvs-src. cuML needs the analogous CMake prefix-discovery change and can then remove its direct cutile workaround separately.

## Validation

Using rapidsai/ci-wheel:26.10-cuda13.3.0-rockylinux8-py3.14:

- reproduced the original failure at e6cb49389: CMake cloned cuVS and failed under _deps/cuvs-src because cuda.tile was unavailable
- after this change, CMake reports CPM: Using local package cuvs@26.10.0 and cuvs_DIR points into site-packages/libcuvs
- no _deps/cuvs-src directory is created and CMake configuration completes
- generated CUDA 13.3 pip-devcontainer requirements contain libcuvs-cu13 and no direct cuda-tile, NVRTC, or nvJitLink requirements
- generated CUDA 12 requirements select libcuvs-cu12
- strict RAPIDS dependency generation passes
- pre-commit passes on all changed files

Using ghcr.io/rapidsai/cugraph/devcontainer:26.10-cuda13.3-pip with libcuvs, libraft, and librmm wheels installed:

- reproduced the CI regression caused by exposing every wheel prefix: downstream pylibcugraph generation could not resolve nvtx3::nvtx3-cpp or rapids_logger::rapids_logger from wheel-backed RMM/RAFT exports
- verified the filtered prefix discovery selects source RMM/RAFT and wheel libcuvs
- verified both libcugraph and downstream pylibcugraph CMake configure/generate successfully with all three wheel prefixes present


jameslamb (approved):
Fantastic, if this works it's exactly what we want.

ChuckHastings (approved):

jameslamb (approved):
re-approving, great! I'm glad to see this switch and for this project to no longer need to carry around copies of all of cuVS's build dependencies.

cpp/CMakeLists.txt
```diff
@@ -19,6 +19,18 @@ rapids_cuda_init_architectures(CUGRAPH)
...
 project(CUGRAPH VERSION "${RAPIDS_VERSION}" LANGUAGES C CXX CUDA)
 
+# Make libcuvs shipped in an installed wheel discoverable. This is needed for
+# direct C++ builds in pip environments, where scikit-build-core is not present
+# to add the wheel's `cmake.prefix` entry point to CMAKE_PREFIX_PATH.
+find_package(Python COMPONENTS Interpreter)
+if(Python_FOUND)
+  include(${rapids-cmake-dir}/cython-core/find_prefix_paths.cmake)
+  rapids_cython_find_prefix_paths("${Python_EXECUTABLE}" wheel_prefix_paths)
+  list(FILTER wheel_prefix_paths INCLUDE REGEX "/libcuvs$")
```
jameslamb, line 29: This is fine as a stop-gap to fix the immediate CI issues but... do we need some followups here?

It means devcontainers here are still rebuilding `librmm`, `libraft`, etc. from source instead of using packages.

```text
-- CPM: Adding package rmm@26.10 (release/26.10)
...
-o _deps/rmm-build/CMakeFiles/rmm.dir/src/aligned.cpp.o
...
```

([build link](https://github.com/rapidsai/cugraph/actions/runs/34296492001/job/102523165602?pr=5654))

jameslamb: We talked offline, there will be a follow-up PR. Merging for now to unblock CI.

PR #5667, opened 2026-09-11
Deprecate MTMG graph and result APIs for removal in 27.02

ChuckHastings:
Mark documented public MTMG edgelist, graph, property, renumber, and result types for removal while keeping resource manager and handle APIs.


seunghwak (approved):
LGTM

PR #5672, opened 2026-09-14
Optimize Maximal Independent Set with degree ordered priorities

jnke2016:
## Summary

Reworks `maximal_independent_set` to derive vertex priorities from degree information instead of sampling random vertices every iteration, and restricts the edge traversal to the undecided vertices. This yields both larger independent sets and fewer edge visits per iteration.

Ranks are assigned once, up front, as a random permutation *within* each degree segment. Keeping the degree bands intact preserves the low-degree-first priority that produces the larger set, while randomizing the order inside a band avoids the long dependency chains a monotone ordering creates on high diameter graphs. The permutation uses the `rng_state` the API already takes, so results stay reproducible. On multi-GPU the vertex ID is used as the rank directly: vertices are hashed to GPUs and each GPU renumbers its own by degree, so the ordering is already randomized at the GPU block level. Graphs built without renumbering have no segment offsets, so their vertices are sorted by degree and the band boundaries recovered from the sorted degrees.

## Directed graphs

A vertex is adjacent to both its incoming and its outgoing neighbors, so for a symmetric graph the outgoing edge traversal already covers both directions and the second pass over incoming edges is dropped. Asymmetric graphs instead push each undecided vertex's rank to its destinations, which reaches the incoming neighbors without materializing a transposed copy of the graph.

## Tests

Each result is verified against the two defining properties of a maximal independent set:

- **Independence**: no edge has both endpoints in the set, ignoring self loops. Checked by reducing over the outgoing edges of every vertex and requiring a count of zero.
- **Maximality**: every vertex outside the set has at least one neighbor inside it, so the set cannot be augmented. Checked by reducing over the outgoing edges of the excluded vertices and requiring each to find one.

On multi-GPU the independence count is summed across ranks, while maximality is verified per vertex partition. Both checks run on the real datasets and on small RMAT graphs, with and without renumbering, and are disabled only for the large benchmark graph.


cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +47,296 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+    h_segment_offsets.back()  = local_vtx_partition_size;
+    for (size_t i = 0; i < num_sparse_segments_per_vertex_partition; ++i) {
+      auto threshold = static_cast<edge_t>((i == 0)   ? mid_degree_threshold   // high, mid
+                                           : (i == 1) ? low_degree_threshold   // mid, low
+                                                      : size_t{1});            // low, zero
+      h_segment_offsets[i + 1] = static_cast<vertex_t>(
+        thrust::count_if(handle.get_thrust_policy(),
+                         out_degrees.begin(),
+                         out_degrees.end(),
+                         [threshold] __device__(auto out_degree) {
```
ChuckHastings: In device_functors.cuh there's a new functor is_greater_than_or_equal_to_t that we should use here.

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +47,296 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
-       v_first = graph_view.local_vertex_partition_range_first()] __device__(auto v) {
-        //
-        // if rank of a non-candidate vertex is not std::numeric_limits<vertex_t>::max() (i.e. the
-        // vertex is not already in MIS), set it to std::numeric_limits<vertex_t>::lowest()
-        //
-        auto v_offset = v - v_first;
-        if (temporary_ranks[v_offset] < std::numeric_limits<vertex_t>::max()) {
-          temporary_ranks[v_offset] = std::numeric_limits<vertex_t>::lowest();
-        }
+      thrust::make_counting_iterator(vertex_t{0}),
```
ChuckHastings: This entire thrust::for_each block might be easier to interpret as a thrust::scatter with a transform iterator.

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +47,296 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
-        handle, graph_view, temporary_ranks.begin(), dst_rank_cache.mutable_view());
-    }
+  remaining_vertices.resize(
+    cuda::std::distance(remaining_vertices.begin(),
+                        thrust::copy_if(handle.get_thrust_policy(),
+                                        vertex_begin,
+                                        vertex_end,
+                                        ranks.begin(),
+                                        remaining_vertices.begin(),
+                                        [] __device__(auto rank) {
```
ChuckHastings: use is_less_than_to_t from device_functors.

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+      thrust::transform(handle.get_thrust_policy(),
+                        degrees.begin(),
+                        degrees.end(),
+                        in_degrees.begin(),
+                        degrees.begin(),
+                        cuda::std::plus<edge_t>{});
+    }
+
+    sorted_vertices = rmm::device_uvector<vertex_t>(local_vtx_partition_size, handle.get_stream());
+    thrust::copy(handle.get_thrust_policy(), vertex_begin, vertex_end, (*sorted_vertices).begin());
```
seunghwak, line 93: `cugraph::sequence();`

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+    // levels. Perturbing the ranks any further only adds work.
+    //
+    if (sorted_vertices) {
+      thrust::copy(handle.get_thrust_policy(),
+                   (*sorted_vertices).begin(),
+                   (*sorted_vertices).end(),
+                   sorted_vertex_ranks.begin());
+    } else {
+      thrust::copy(
+        handle.get_thrust_policy(), vertex_begin, vertex_end, sorted_vertex_ranks.begin());
```
seunghwak, line 147: cugraph::sequence();

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+          remaining_vertices.begin() + nr_remaining_local_vertices_to_check;
+
+        rmm::device_uvector<vertex_t> processed_ranks(num_processed_vertices, handle.get_stream());
+        thrust::transform(handle.get_thrust_policy(),
+                          processed_vertex_first,
+                          processed_vertex_first + num_processed_vertices,
+                          processed_ranks.begin(),
+                          cuda::proclaim_return_type<vertex_t>(
+                            [ranks = raft::device_span<vertex_t const>(ranks.data(), ranks.size()),
+                             v_first = v_first] __device__(auto v) { return ranks[v - v_first]; }));
```
seunghwak, line 255: cugraph::gather() with transform_iterator with shift_left_t (shift amount: v_first).

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
 {
   using GraphViewType = cugraph::graph_view_t<vertex_t, edge_t, false, multi_gpu>;
```
seunghwak, line 51: What about masking out self-loops and multi-edges (except for the first one)? We are performing this in several places and we may refactor the code can create a utility function to create a mask for this.

Self-loops or multi-edges should not impact the rank of a vertex in MIS. I am wondering masking out self-loops and multi-edges are worth the computational cost (how much MIS quality improvement do we see?)

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+  auto symmetric = graph_view.is_symmetric();
 
-  rmm::device_uvector<vertex_t> remaining_vertices(local_vtx_partitoin_size, handle.get_stream());
+  vertex_t local_vtx_partition_size = graph_view.local_vertex_partition_range_size();
+  auto v_first                      = graph_view.local_vertex_partition_range_first();
 
-  auto vertex_begin =
-    thrust::make_counting_iterator(graph_view.local_vertex_partition_range_first());
+  auto vertex_begin = thrust::make_counting_iterator(v_first);
   auto vertex_end = thrust::make_counting_iterator(graph_view.local_vertex_partition_range_last());
```
seunghwak, line 61: This might be a personal style thing but just suggestions.

I generally don't like creating too many temporaries.

defining `v_first` and 'vertex_begin` and repeating thrust::make_transform_iterator(graph_view.local_vertex_partition_range_first()) won't make much difference once the code gets compiled.

Then, the question is more about which makes the code easier to read/maintain.

I am more on the side that a bit lengthier `thrust::make_transform_iterator(graph_view.local_vertex_partition_range_first())` is easier to understand then trying to renumber multiple temporaries like `v_first` or `vertex_begin`.

Just my preference and just think about this.

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+    for (size_t i = 0; i < num_sparse_segments_per_vertex_partition; ++i) {
+      auto threshold = static_cast<edge_t>((i == 0)   ? mid_degree_threshold  // high, mid
+                                           : (i == 1) ? low_degree_threshold  // mid, low
+                                                      : size_t{1});           // low, zero
+      h_segment_offsets[i + 1] =
+        static_cast<vertex_t>(thrust::count_if(handle.get_thrust_policy(),
+                                               degrees.begin(),
+                                               degrees.end(),
+                                               is_greater_than_or_equal_to_t<edge_t>{threshold}));
+    }
```
seunghwak, line 116: Use vectorized thrust::lower_bound (or upper_bound) instead of scanning degrees multiple times. few binary search (as degrees are sorted) will be much faster than scanning degrees multiple  times.

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+    // IDs, and each GPU owns one contiguous block of the renumbered ID space while sorting its
+    // own vertices by degree. A rank is therefore decided first by the (random) GPU block and
+    // only then by degree, which already amounts to a random permutation coarsened to comm_size
+    // levels. Perturbing the ranks any further only adds work.
+    //
+    if (sorted_vertices) {
+      thrust::copy(handle.get_thrust_policy(),
+                   (*sorted_vertices).begin(),
+                   (*sorted_vertices).end(),
+                   sorted_vertex_ranks.begin());
```
seunghwak, line 144: Doesn't this mean all vertices in GPU1 will have higher ranks than all vertices in GPU0? (regardless of their degrees/segments).

cpp/src/components/mis_impl.cuh
```diff
@@ -41,243 +50,363 @@ rmm::device_uvector<vertex_t> maximal_independent_set(
...
+      sorted_vertex_ranks.end(),
+      thrust::make_transform_iterator((*sorted_vertices).begin(), shift_left_t<vertex_t>{v_first}),
+      ranks.begin());
+
+    sorted_vertex_ranks.resize(0, handle.get_stream());
+    sorted_vertex_ranks.shrink_to_fit(handle.get_stream());
+    sorted_vertices.reset();
+  } else {
+    ranks = std::move(sorted_vertex_ranks);
+  }
```
seunghwak, line 196: OK, so the final outcome of the code till here is `ranks`. All the other variables are temporaries.

We may better reduce the scope of other temporaries.

rmm::device_uvector<vertex_t> ranks(0, handle.get_stream());
{
...
}

or

defining rmm::device_uvector<vertex_t> compute_ranks(...);

will make it cleaner to separate rank computations from MIS computations based on the computed ranks.

seunghwak: So, this is the flow if I am not mistaken.
```
rmm::device_uvector<vertex_t> ranks(0, handle.get_stream());
if (segment_offsets && symmetric) {
  ranks.resize(graph_view.local_vertex_partition_range_size(),
                         handle.get_stream());
  thrust::sequences(ranks.begin(), ranks.end(),
                                   graph_view.local_vertex_partition_range_first());
  if constexpr (!multi_gpu) {
    // random permute ranks within a segment.
  }
}
else {
  compute degrees
  compute sorted_vertices from degrees
  if constexpr (multi_gpu) {
    ranks = std::move(sorted_vertices);
  }
  else {
    // segment sorted_vertices based on degrees
    // for each segment, assign random (permuted) ranks in [graph_view.local_vertex_partition_range_first() + segment_offsets[segment_idx], graph_view.local_vertex_partition_range_first() + segment_offsets[segment_idx + 1],);
    // now update ranks from (sorted_vertices, sorted_vertex_ranks).
  }
}
```
You may restructure the rank computation code to make this flow clearer.

seunghwak: And in multi-GPU, every vertex in GPU1 will  get higher ranks than GPU0 regardless of the degree. Sounds undesirable. Degree 1000 vertex should have a lower rank than degree 1 vertex regardless of GPU ranks.

cpp/src/components/mis_impl.cuh
jnke2016: > I am wondering masking out self-loops and multi-edges are worth the computational cost (how much MIS quality improvement do we see?)

Just to confirm: you're asking whether the MIS quality gain justifies the cost of building those masks?

jnke2016: Both checks fit in a single `transform_e`, so masking costs one O(E) pass at setup. The only recurring cost is the mask lookup on every edge traversal in the main loop.

jnke2016: actually it seems like transform_e cannot mask multi edges, only self loops unless I am missing something?

seunghwak: See https://github.com/rapidsai/cugraph/blob/main/cpp/src/components/strongly_connected_components_impl.cuh#L2049 for masking out multi-edges.

seunghwak: And yes, I am asking masking out self-loops & multi-edges (as this is not free) justifies the cost. We are. prioritizing based on in-degrees and out-degrees but self-loops and multi-edges (except for the first one) should not impact the priority.

PR #5676, opened 2026-09-14
Update PLC Graph objects to eliminate unnecessary dangling pointers

### LINKED ISSUE #5193: Refactor pylibcugraph to avoid maintaining unnecessary pointers to weights, edge IDs, others
There's currently code in PLC for taking data provided by Python users to be used as edge weights and/or edge IDs (and possibly others) and creating device arrays managed by RMM for use by libcugraph. This data is kept as a member of the internal `_GPUGraph` class and occasionally referenced later to see if the graph was created with edge weights or IDs, but does not appear to be used anywhere else.

This task is to investigate if there's an alternate approach that avoids keeping the pointers to the data and making the change if possible.

ChuckHastings:
- Add C API helpers `cugraph_graph_has_edge_{weights,ids,types,start_times,end_times}` that query the opaque `cugraph_graph_t` (Closes #5193).
- Stop storing weight/edge-ID array views on `_GPUGraph`; use create-time locals and query the graph handle via the new C API from PLC (random walks, edge betweenness).
- Fix `MGGraph` construction leaks: `free()` all `malloc`’d per-property pointer arrays after graph create; run view/array cleanup in a `finally` block so failed creates do not leak.
- SG: free edge start/end time views after create (were previously leaked on the COO path).


alexbarghi-nv (approved):
Looks good; this should hopefully help further cut memory usage

jnke2016 (approved):
Looks good to me

seunghwak (approved):
LGTM

PR #5683, opened 2026-09-17
Add download=true to reduce number of retries in CI

ChuckHastings:
We see sporadic failures in CI tests running in xdist by files not being available.  Setting download=true should reduce that.  It won't download if the file already exists, so we should only get multiple downloads of a file if it's not there and two python tests start up at the same time needing that file.


jnke2016 (approved):
> so we should only get multiple downloads of a file if it's not there and two python tests start up at the same time needing that file.

And I think this is a safe and cheap workaround while trying to figure out why those datasets are missing or deleted in the first place.

